PR #9311, opened 2018-12-12
libG copy steps should have the same behavior as before.

mjkkirschner:
### Purpose

@QilongTang noticed a change in the libG folder names copied the bin directory.
I'm not a fan of the libG folders in extern being `LibG_` and those in bin being `libG_` - but I think this is the least risky change at this point. 

Normally this should not be a problem on standard windows, but it might be on mono, mac, linux platforms.

This PR just reverts the copy steps to their previous behavior.

Some alternatives:
 1.  change all the libG loading to use uppercase paths like 
`var libGFolderName = string.Format("Libg_{0}_{1}_{2}", major, minor, build);` instead of the lowercase libG. 
2. change the folders in extern to be lowerCase and request BRE team to change their names in autogenerated cross merge PR.
3. try to normalize the paths before checking if they exist on disk, using toLower() or some other mechanism - seems like the most work.
4. do nothing - likely will cause problems on other platforms or windows machines that have case sensitive filesystems enabled. 

Looking for some feedback/thoughts.

### Declarations

### Reviewers

@QilongTang 
@aparajit-pratap 
### FYIs

@smangarole


aparajit-pratap:
@mjkkirschner does this pose an issue in Windows? How and when does it surface?

QilongTang:
@aparajit-pratap I was testing ASM 225 on my machine locally and loadling libG failed for me because of case sensitive setting on my machine

mjkkirschner:
@aparajit-pratap on windows this will cause issues if a user has enabled file system case sensitivity which is now an option in windows 10. 😬

PR #9313, opened 2018-12-13
LibG Binaries Update

DynamoEngOps:
__LibG225__ 
CommitInfo:Aparajit Pratap - Skip failure category tests from running in memory leak checker (#651) (#654) 
SHA:[1bc999ac6ccb3fcbfe85fcff8b4ba193bd40f387](https://git.autodesk.com/Dynamo/LibG/commit/1bc999ac6ccb3fcbfe85fcff8b4ba193bd40f387)
Reviewer: Aparajit Pratap

### DIFF
--- a/src/DynamoCore/Search/SearchDictionary.cs
+++ b/src/DynamoCore/Search/SearchDictionary.cs
@@ -14,6 +14,8 @@ namespace Dynamo.Search
     {
         protected readonly Dictionary<V, Dictionary<string, double>> entryDictionary =
             new Dictionary<V, Dictionary<string, double>>();
+        
+        private List<IGrouping<string, Tuple<V, double>>> tagDictionary;
 
         /// <summary>
         ///     All the current entries in search.
@@ -48,6 +50,7 @@ namespace Dynamo.Search
         {
             var handler = EntryAdded;
             if (handler != null) handler(entry);
+            tagDictionary = null;
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


mjkkirschner:
hmm weird- 225 is getting copied but protogeo has changed size here... @aparajit-pratap any thoughts?

aparajit-pratap:
@mjkkirschner I'm not sure what you mean? Protogeometry.dll hasn't changed in size, I don't think. It shows `+0 bytes`.

mjkkirschner:
sorry, meant libG.Interface:
`BIN -512 Bytes (99%) extern/ProtoGeometry/LibG.Interface.dll`

aparajit-pratap:
Not that it should matter much but are the `ProtoGeometry` and `LibG.Interface` common binaries coming from the latest ASM branch, i.e. ASM225 branch? They could come from any of the branches, depending on whichever one is built last. They just get overwritten each time until the last branch is built. It would be best practice therefore to have the branches build starting from the lowest version to the latest, in that order.

aparajit-pratap:
@mjkkirschner can you figure out this order by looking at the BRE build script on Jenkins? 

On another note I haven't cherry-picked my binary versioning changes into `ASM223` and `ASM224` branches. Not sure which Revit version did the Revit team want this for?

mjkkirschner:
versioning changes were required for next release - but I would guess we should version any files that make it into the install directory - (ASM 224 branches I think)

I will checkout the build script.

aparajit-pratap:
Just got Minjie's message on the Revit JIRA task. They didn't really need the versioning for older Dynamo-Revit dll's so I guess we're okay.

PR #9305, opened 2018-12-11
UI fix for Dictionary label in node preview

aparajit-pratap:
### Purpose

This fixes the `Dictionary` top-level label in the node preview. The text was partially cut-off - more specifically the `D` in `Dictionary`. The reason was that the top-level label had a different margin setting than nested labels. This fix removes the special setting for top-level labels and keeps just one margin setting for both top and nested labels. 

Before:
image

After:
image

### Declarations

Check these if you believe they are true

- [x] The code base is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [x] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions), and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.

### Reviewers

@mjkkirschner 
@Racel - this requires your UX review as the top-level labels for both `List` and `Dictionary` texts are shifted 4 units to the right and are the same as nested labels. Note that I couldn't find an alternative way of fixing this UI issue.

PR #9320, opened 2018-12-17
Update preview for multi-output node when disconnecting it 

aparajit-pratap:
### Purpose

JIRA: https://jira.autodesk.com/browse/DYN-1211

This addresses an issue where for any multi-output node, the preview bubble wasn't getting cleared after disconnecting the node from an upstream node. The preview values would remain. 

The reason for this is that whenever a node is disconnected, its AST's are recomputed to look something like:
```
var_<node_guid> = __CreateFunctionObject(...);
```
where the node, identified by its output identifier, `var_<guid>`, now becomes a function object (or partial function). However for multi-output nodes, the variable identifier was erroneously being set to something else, `__partial_<guid>`. Since the original node identifier was not being replaced with a function object, it retained its original cached value and when it was queried in the preview bubble, it returned the original value as expected. The fix is to replace the new identifier with the original one so that it can be reset to a function object when the node is disconnected.

screenshot

### Declarations

Need to run self-serve and add tests - WIP

### Reviewers

@alfarok 

### FYIs

@Racel

### DIFF
--- a/src/DynamoCore/Graph/Nodes/FunctionCallNodeController.cs
+++ b/src/DynamoCore/Graph/Nodes/FunctionCallNodeController.cs
@@ -68,8 +68,7 @@ namespace Dynamo.Graph.Nodes
         {
             var missingAmt =
                 Enumerable.Range(0, model.InPorts.Count).Count(x => !model.InPorts[x].IsConnected);
-            var tmp =
-                AstFactory.BuildIdentifier("__partial_" + model.GUID.ToString().Replace('-', '_'));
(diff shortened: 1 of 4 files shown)


QilongTang:
LGTM but why are the data in the preview bubble not aligned in your gif? high resolution display bug?

aparajit-pratap:
> LGTM but why are the data in the preview bubble not aligned in your gif? high resolution display bug?

Yes, I tried to fix it here: https://github.com/DynamoDS/Dynamo/pull/9305 but apparently I'm the only one who sees it on my computer.

PR #9324, opened 2018-12-18
API for libG Preloader Lookup

QilongTang:
Please Note:

### Purpose

This is an improvement when working with a beta release of ASM release that we make our libG preloader lookup mechanism more tolerant than just search for the exact match.

The influence of this API is that for in-process Dynamo integration, client need to add this API call to their ASM preloading logic:

```
        internal static Version PreloadAsmFromRevit()
        {
            var asmLocation = AppDomain.CurrentDomain.BaseDirectory;
            Version libGversion = findRevitASMVersion(asmLocation);
            var dynCorePath = DynamoRevitApp.DynamoCorePath;
            var preloaderLocation = DynamoShapeManager.Utilities.GetLibGPreloaderLocation(libGversion, dynCorePath);
            DynamoShapeManager.Utilities.PreloadAsmFromPath(preloaderLocation, asmLocation);
            return libGversion;
        }
```

### Declarations

Running self-CI

### Reviewers

@angelowang @mjkkirschner 

### FYIs

@aparajit-pratap @alfarok

### DIFF
--- a/src/Tools/DynamoShapeManager/Utilities.cs
+++ b/src/Tools/DynamoShapeManager/Utilities.cs
@@ -203,6 +203,38 @@ namespace DynamoShapeManager
         }
 
         /// <summary>
+        /// Get the corresponding libG preloader location for the target ASM loading version.
+        /// </summary>
+        /// <param name="version">The target loading version of ASM.</param>
+        /// <param name="rootFolder">Full path of the directory that contains 
+        /// LibG_major_minor_build folder. In a 
+        /// typical setup this would be the same directory that contains Dynamo 
+        /// core modules. This must represent a valid directory.</param>
+        /// <returns></returns>
+        public static string GetLibGPreloaderLocation(Version version, string rootFolder)
+        {
+            int major = 0, minor = 0, build = 0;
+            if (version != null)
+            {
+                major = version.Major;
+                minor = version.Minor;
+                build = version.Build;
+            }
+
+            var libGFolderName = string.Format("libg_{0}_{1}_{2}", version.Major, version.Minor, version.Build);
+            var dir = new DirectoryInfo(Path.Combine(rootFolder, libGFolderName));
+            if (dir.Exists)
+                return dir.FullName;
+            else
+            {
+                // This usually means libG version is behind the target version, fallback to use another existing version 
+                var rootDir = new DirectoryInfo(rootFolder);
+                // TODO: Sort the existing libG folders somehow and find a folder below the target ASM version
+                return rootDir.GetDirectories().Where(x => x.Name.Contains("LibG_" + version.Major.ToString())).First().FullName;
+            }
+        }
+
+        /// <summary>
         /// Call this method to preload ASM binaries from a specific location. This 
         /// method does not have a return value, any failures in loading ASM binaries
         /// will result in an exception being thrown.
--- a/src/Tools/DynamoShapeManager/Utilities.cs
+++ b/src/Tools/DynamoShapeManager/Utilities.cs
@@ -5,7 +5,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
-
+using System.Text.RegularExpressions;
 
 namespace DynamoShapeManager
 {
@@ -204,6 +204,8 @@ namespace DynamoShapeManager
 
         /// <summary>
         /// Get the corresponding libG preloader location for the target ASM loading version.
+        /// If there is exact match preloader version to the target ASM version, use it, 
+        /// otherwise use the closet below.
         /// </summary>
         /// <param name="version">The target loading version of ASM.</param>
         /// <param name="rootFolder">Full path of the directory that contains 
@@ -227,10 +229,30 @@ namespace DynamoShapeManager
                 return dir.FullName;
             else
             {
-                // This usually means libG version is behind the target version, fallback to use another existing version 
+                // This usually means libG preloader version is behind the target version
                 var rootDir = new DirectoryInfo(rootFolder);
-                // TODO: Sort the existing libG folders somehow and find a folder below the target ASM version
-                return rootDir.GetDirectories().Where(x => x.Name.Contains("LibG_" + version.Major.ToString())).First().FullName;
+
+                // Use regex to get all the libG versions supported
+                var libgFolders = rootDir.EnumerateDirectories("libg_*", SearchOption.TopDirectoryOnly);
+                var regExp = new Regex(@"^libg_(\d\d\d)_(\d)_(\d)$", RegexOptions.IgnoreCase);
+                var preloaderVersions = new List<Version>();
+                foreach (var folder in libgFolders)
+                {
+                    var match = regExp.Match(folder.Name);
+                    if (match.Groups.Count == 4)
+                    {
+                        preloaderVersions.Add(new Version(
+                                Convert.ToInt32(match.Groups[1].Value),
+                                Convert.ToInt32(match.Groups[2].Value),
+                                Convert.ToInt32(match.Groups[3].Value)));
+                    }
+                }
+                preloaderVersions.Sort();
+                preloaderVersions.Reverse();
+                // Pick the closest preloader version below
+                var preloaderVersion = preloaderVersions.First();
+                libGFolderName = string.Format("libg_{0}_{1}_{2}", preloaderVersion.Major, preloaderVersion.Minor, preloaderVersion.Build);
+                return libGFolderName;
             }
         }
 
--- a/src/Tools/DynamoShapeManager/Utilities.cs
+++ b/src/Tools/DynamoShapeManager/Utilities.cs
@@ -215,12 +215,9 @@ namespace DynamoShapeManager
         /// <returns></returns>
         public static string GetLibGPreloaderLocation(Version version, string rootFolder)
         {
-            int major = 0, minor = 0, build = 0;
-            if (version != null)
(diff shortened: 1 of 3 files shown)


src/Tools/DynamoShapeManager/Utilities.cs
```diff
@@ -202,6 +202,38 @@ public static Version GetInstalledAsmVersion2(IEnumerable<Version> versions, ref
...
+            var libGFolderName = string.Format("libg_{0}_{1}_{2}", version.Major, version.Minor, version.Build);
+            var dir = new DirectoryInfo(Path.Combine(rootFolder, libGFolderName));
+            if (dir.Exists)
+                return dir.FullName;
+            else
+            {
+                // This usually means libG version is behind the target version, fallback to use another existing version 
+                var rootDir = new DirectoryInfo(rootFolder);
+                // TODO: Sort the existing libG folders somehow and find a folder below the target ASM version
+                return rootDir.GetDirectories().Where(x => x.Name.Contains("LibG_" + version.Major.ToString())).First().FullName;
```
angelowang: if target major version does not exist, then probably will crash or exception? i will prefer using the most closest one among all the libg_x_x_x folders.

QilongTang: Will update this part

src/Tools/DynamoShapeManager/Utilities.cs
```diff
@@ -202,6 +202,60 @@ public static Version GetInstalledAsmVersion2(IEnumerable<Version> versions, ref
...
+                                Convert.ToInt32(match.Groups[1].Value),
+                                Convert.ToInt32(match.Groups[2].Value),
+                                Convert.ToInt32(match.Groups[3].Value)));
+                    }
+                }
+                preloaderVersions.Sort();
+                preloaderVersions.Reverse();
+                // Pick the closest preloader version below
+                var preloaderVersion = preloaderVersions.First();
+                libGFolderName = string.Format("libg_{0}_{1}_{2}", preloaderVersion.Major, preloaderVersion.Minor, preloaderVersion.Build);
```
angelowang: this will not include the rootfolder:) which seems not the same as tthe other branch (`dir.Exists`)

src/Tools/DynamoShapeManager/Utilities.cs
```diff
@@ -202,6 +202,60 @@ public static Version GetInstalledAsmVersion2(IEnumerable<Version> versions, ref
...
+                var preloaderVersions = new List<Version>();
+                foreach (var folder in libgFolders)
+                {
+                    var match = regExp.Match(folder.Name);
+                    if (match.Groups.Count == 4)
+                    {
+                        preloaderVersions.Add(new Version(
+                                Convert.ToInt32(match.Groups[1].Value),
+                                Convert.ToInt32(match.Groups[2].Value),
+                                Convert.ToInt32(match.Groups[3].Value)));
```
angelowang: check if the version of the folder is less than or equal to the target version. e.g., if target is 225.3, we should not match to 225.4.

src/Tools/DynamoShapeManager/Utilities.cs
```diff
@@ -202,6 +202,60 @@ public static Version GetInstalledAsmVersion2(IEnumerable<Version> versions, ref
...
+                    }
+                }
+                preloaderVersions.Sort();
+                preloaderVersions.Reverse();
+                // Pick the closest preloader version below
+                var preloaderVersion = preloaderVersions.First();
```
angelowang: `FirstOrDefault` to avoid exceptions, and check null later. but that might be a little too complicated and not worth it.

src/Tools/DynamoShapeManager/Utilities.cs
QilongTang: Good catch, funny that I wrote unit test same way which failed to catch this :)

test/DynamoCoreTests/libGPreloaderTests.cs
```diff
@@ -167,6 +167,41 @@ public void GetInstalledASMVersions2_ASMVersionFallback()
...
+            var rootFolder = Path.Combine(Path.GetTempPath(), "LibGTest");
+            // both versions of libG exist
+            var libG22440path = System.IO.Directory.CreateDirectory(Path.Combine(rootFolder, "LibG_224_4_0"));
+            var libG22500path = System.IO.Directory.CreateDirectory(Path.Combine(rootFolder, "LibG_225_0_0"));
+
+            var foundVersion = DynamoShapeManager.Utilities.GetInstalledAsmVersion2(
+                versions, ref foundPath, rootFolder, (path) => { return mockedInstalledASMs; });
+
+            // The found version in this case is a fallback of lowest version within same major which should be 225.0.0
+            Assert.AreEqual(targetVersion, foundVersion);
```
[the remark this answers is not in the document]
QilongTang: Will do

src/Tools/DynamoShapeManager/Utilities.cs
QilongTang: Added check above

QilongTang:
@mjkkirschner @aparajit-pratap @angelowang Thank you guys for the review, I addressed all your comments and this is ready for another look

src/Tools/DynamoShapeManager/Utilities.cs
```diff
@@ -202,6 +202,58 @@ public static Version GetInstalledAsmVersion2(IEnumerable<Version> versions, ref
...
+                                Convert.ToInt32(match.Groups[3].Value)));
+                    }
+                }
+                preloaderVersions.Sort();
+                preloaderVersions.Reverse();
+                // Pick the closest preloader version below or use the default value when no libG folder found
+                var preloaderVersion = preloaderVersions.FirstOrDefault() == null ? version : preloaderVersions.First();
+                return Path.Combine(rootFolder, string.Format("libg_{0}_{1}_{2}", preloaderVersion.Major, preloaderVersion.Minor, preloaderVersion.Build));
+            }
+        }
```
aparajit-pratap: @QilongTang so I'm trying to understand this ... you're **not** taking into account the case here where an ASM installed version like ASM 224.5 is available on a user's machine and it still picks `LibG_224_4_0` folder to load LibG from. You're only looking at the latest ASM version (225 currently) and making sure that if the user has a version of ASM 225.X installed where X > Y, it still loads libg from `LibG_ASM225_Y_Z`? Am I right?

aparajit-pratap: @QilongTang I'm talking about 2 individual cases: one, where the user has ASM 224.X, where X > 4 and the other where the user has ASM 225.X, where X > 0 (what we have currently).

QilongTang: @aparajit-pratap As I said, it depends on if the user has both Revit 2020 and Revit 2019.X installed locally and choose to just launch DynamoSandbox without Revit. If that is the case, Sandbox prefer ASM 225.X, if user only have one version of ASM locally, either 225.X or 224.X, Sandbox only work with that version. Does that answer your question?

QilongTang:
@mjkkirschner @aparajit-pratap Ping

src/Tools/DynamoShapeManager/Utilities.cs
```diff
@@ -202,6 +202,58 @@ public static Version GetInstalledAsmVersion2(IEnumerable<Version> versions, ref
...
+                                Convert.ToInt32(match.Groups[1].Value),
+                                Convert.ToInt32(match.Groups[2].Value),
+                                Convert.ToInt32(match.Groups[3].Value)));
+                    }
+                }
+                preloaderVersions.Sort();
+                preloaderVersions.Reverse();
+                // Pick the closest preloader version below or use the default value when no libG folder found
+                var preloaderVersion = preloaderVersions.FirstOrDefault() == null ? version : preloaderVersions.First();
+                return Path.Combine(rootFolder, string.Format("libg_{0}_{1}_{2}", preloaderVersion.Major, preloaderVersion.Minor, preloaderVersion.Build));
```
aparajit-pratap: If there's no matching libg folder found for the target `version` passed in, then setting `preloaderVersion` to `version` and constructing a libg folder path from it will return a non-existent location.

aparajit-pratap:
@QilongTang I just have one more comment; the last one above that I think needs to be addressed. After that this is good to go.

src/Tools/DynamoShapeManager/Utilities.cs
QilongTang: @aparajit-pratap `Preload()` will fail in this case, do you prefer just fall back to whatever libG folder available even the version is larger?

mjkkirschner:
@QilongTang @aparajit-pratap is this good to go?

QilongTang:
@mjkkirschner Not yet, @aparajit-pratap wanted to add some better error message when libG preloader not found other than an invoke exception so I am looking into it

src/Tools/DynamoShapeManager/Utilities.cs
QilongTang: @aparajit-pratap If we end up returning an invalid LibG folder although unlikely, we will throw exception at https://github.com/DynamoDS/Dynamo/blob/7b885b76ea585128af0eb997e0f1567545952466/src/Tools/DynamoShapeManager/Utilities.cs#L289 which will end up in the crash dialog and analytics. I can make the exception message more explicit

aparajit-pratap:
Thanks @QilongTang, LGTM!

PR #9330, opened 2018-12-19
Avoid namespace collisions when collapsing graph to custom node

mjkkirschner:
### Purpose

Collapsing a graph to a custom node can create errors if there is a namespace conflict in types used in the generated input nodes.

Adds a new visitor to the shortQualifiedNameReplacer and uses it to replace names when creating custom node symbols when collapsing.

Will cherry pick this to 2.1 branch if no issues found.

### Declarations

### Reviewers

@aparajit-pratap 

### FYIs

@smangarole


src/DynamoCore/Core/CustomNodeManager.cs
```diff
@@ -674,7 +678,8 @@ internal bool TryGetInfoFromPath(string path, bool isTestMode, out CustomNodeInf
                 nodeGraph.ElementResolver,
```
aparajit-pratap: I see the nodeGraph is instantiated from an xml file? Is that why it doesn't get the correct element resolver as we are now using JSON?

mjkkirschner:
@aparajit-pratap I've updated this with one potential solution.

src/DynamoCore/Core/CustomNodeManager.cs
mjkkirschner: here is where we use the deserialized element resolver when loading a JSON graph and pass it to the home workspace or customNode workspace that was just deserialized.

https://github.com/DynamoDS/Dynamo/blob/ec10f936824152e7dd7d6d019efdcda0d78a5264/src/DynamoCore/Graph/Workspaces/SerializationConverters.cs#L532

src/DynamoCore/Core/CustomNodeManager.cs
```diff
@@ -1090,10 +1103,35 @@ internal bool TryGetNodeInfo(string name, out CustomNodeInfo info)
...
+                                        new List<AssociativeNode> { typedNode },
+                                        currentWorkspace.ElementResolver);
+
+                                    node.InputSymbol = $"{typedNode.Value} :{typedNode.TypeAlias}";
+                                }
+                                catch(Exception e)
+                                {
+                                    node.InputSymbol += ":" + typedParameter.Type.Name;
+                                    this.AsLogger().LogError($"{e.Message}: could not generate a short type name for {typedParameter.Type.Name}");
+                                }
```
aparajit-pratap: Does this set `InputSymbol` to the full name even though it is unable to find a short name? If not, it should.

src/DynamoCore/Engine/NodeToCode/NodeToCode.cs
```diff
@@ -320,6 +321,55 @@ public ShortestQualifiedNameReplacer(ClassTable classTable, ElementResolver reso
...
+            if (node == null)
+                return null;
+
+            // If type is primitive type
+            if (node.datatype.UID != (int)PrimitiveType.InvalidType &&
+                node.datatype.UID < (int)PrimitiveType.MaxPrimitive)
+                return node;
+            
+            // build an idlistnode from the full type name
+            var identListNode = CoreUtils.CreateNodeFromString(node.datatype.Name);
```
aparajit-pratap: We can just pass `node.TypeAlias` to  `CreateNodeFromString`.

src/DynamoCore/Engine/NodeToCode/NodeToCode.cs
```diff
@@ -320,6 +321,55 @@ public ShortestQualifiedNameReplacer(ClassTable classTable, ElementResolver reso
...
+            if (identListNode == null)
+                return null;
```
aparajit-pratap: It might be safer to return `node` here just in case it's still valid.

src/DynamoCore/Engine/NodeToCode/NodeToCode.cs
```diff
@@ -320,6 +321,55 @@ public ShortestQualifiedNameReplacer(ClassTable classTable, ElementResolver reso
...
+
+            // visit the idlist 
+            var shortNameNode = identListNode.Accept(this);
+            if(shortNameNode == null)
+                return null;
```
aparajit-pratap: It might be safer to return node here just in case it's still valid.

src/DynamoCore/Engine/NodeToCode/NodeToCode.cs
```diff
@@ -320,6 +321,55 @@ public ShortestQualifiedNameReplacer(ClassTable classTable, ElementResolver reso
...
+            {
+                var idListNode = (shortNameNode as IdentifierListNode);
+                resultingNode.TypeAlias = idListNode.ToString();
+            }
+
+            else if (shortNameNode is IdentifierNode)
+            {
+                var idNode = (shortNameNode as IdentifierNode);
+                resultingNode.TypeAlias = idNode.Value;
+            }
```
aparajit-pratap: I think we can just do: `resultingNode.TypeAlias = shortNameNode.ToString();`

src/DynamoCore/Engine/NodeToCode/NodeToCode.cs
```diff
@@ -320,6 +321,55 @@ public ShortestQualifiedNameReplacer(ClassTable classTable, ElementResolver reso
...
+                return null;
+
+            // visit the idlist 
+            var shortNameNode = identListNode.Accept(this);
+            if(shortNameNode == null)
+                return null;
+            
+
+            var resultingNode = new TypedIdentifierNode();
+            resultingNode.Name = node.Name;
```
aparajit-pratap:
```
var resultingNode = new TypedIdentifierNode{
Name = node.Name,
Value = node.Name,
...
}

File: src/DynamoCore/Engine/NodeToCode/NodeToCode.cs [hunk 8]
```diff
@@ -320,6 +321,55 @@ public ShortestQualifiedNameReplacer(ClassTable classTable, ElementResolver reso
...
+            {
+                var idNode = (shortNameNode as IdentifierNode);
+                resultingNode.TypeAlias = idNode.Value;
+            }
+
+            //modify the AST node that was passed to the visitor.
+            node.Name = resultingNode.Name;
+            node.Value = resultingNode.Value;
+            node.datatype = resultingNode.datatype;
+            node.TypeAlias = resultingNode.TypeAlias;
```
aparajit-pratap: Does `node` have to be passed as a `ref` parameter in order to directly assign it to `resultingNode`? Is that why you have modified its properties individually here?

src/DynamoCore/Core/CustomNodeManager.cs
mjkkirschner: yep, thats what it does.

src/DynamoCore/Engine/NodeToCode/NodeToCode.cs
mjkkirschner: will try that - thanks.

src/DynamoCore/Engine/NodeToCode/NodeToCode.cs
mjkkirschner: 👍

src/DynamoCore/Engine/NodeToCode/NodeToCode.cs
mjkkirschner: yep, I found that assigning `node = resultantNode` did not update the node I passed in, and using a ref parameter requires touching the ASTreplacer base class - so this is not ideal, but it does seem to work.

src/DynamoCore/Core/CustomNodeManager.cs
```diff
@@ -1077,10 +1083,17 @@ internal bool TryGetNodeInfo(string name, out CustomNodeInfo info)
                             var funcDesc = dsFunc.Controller.Definition;
                             parameters = funcDesc.Parameters.ToList();
 
+                            // if the node is an instance member the function won't contain a 
+                            // parameter for this type so we need to geneate a new typedParameter.
```
aparajit-pratap: typo: generate

aparajit-pratap:
@mjkkirschner I think this looks good overall. Couple of other comments you could address here. I will look at the test cases you add before we decide to merge this. Thanks for sticking with this and fixing it :)

mjkkirschner:
@aparajit-pratap I've moved the `ShortQualifiedNameReplacer` to it's own file but kept it in dynamo core because it was used by public methods in dynamoCore (ie `ReplaceWithShorestQualifiedName`) and I wanted to keep the underlying class internal. 

I've added a few tests for shortNameReplacer that test the `typedIdentifer` visitor and some customNodeCollapse tests that use FFITarget to create conflicts.

I have yet to run all the tests on self serve CI.

test/DynamoCoreTests/CustomNodes.cs
```diff
@@ -893,6 +893,85 @@ public void TestCustomNodeFromCollapsedNodeHasTypes()
...
+                    Name = "__CollapseTest__",
+                    Success = true
+                });
+
+            CurrentDynamoModel.AddCustomNodeWorkspace(ws);
+            CurrentDynamoModel.OpenCustomNodeWorkspace(ws.CustomNodeId);
+            var inputs = CurrentDynamoModel.CurrentWorkspace.Nodes.OfType<Symbol>().ToList();
+            Assert.IsNotNull(inputs);
+
+            var pointParams = inputs;
```
aparajit-pratap: where is `pointParams` used?

test/DynamoCoreTests/CustomNodes.cs
```diff
@@ -893,6 +893,85 @@ public void TestCustomNodeFromCollapsedNodeHasTypes()
...
+            inputs.ForEach(x => Assert.NotNull(x));
+
+            Assert.IsTrue(inputs.All(t =>
+            {
+                Console.WriteLine(t.InputSymbol);
```
aparajit-pratap: What's the `Console.WriteLine` for?

test/DynamoCoreTests/CustomNodes.cs
```diff
@@ -893,6 +893,85 @@ public void TestCustomNodeFromCollapsedNodeHasTypes()
...
+                    Name = "__CollapseTest__",
+                    Success = true
+                });
+
+            CurrentDynamoModel.AddCustomNodeWorkspace(ws);
+            CurrentDynamoModel.OpenCustomNodeWorkspace(ws.CustomNodeId);
+            var inputs = CurrentDynamoModel.CurrentWorkspace.Nodes.OfType<Symbol>().ToList();
+            Assert.IsNotNull(inputs);
+
+            var pointParams = inputs;
```
aparajit-pratap: same here.

aparajit-pratap:
@mjkkirschner excellent! Some minor comments and then LGTM!

PR #9332, opened 2018-12-19
Bring search improvements out from behind debug menu

scottmitchell:
### Purpose

This PR updates finalizes improvements to search made in [DYN-1326](https://jira.autodesk.com/browse/DYN-1326) and [DYN-1324](https://jira.autodesk.com/browse/DYN-1326). These improvements were initially placed behind the debug menu. This PR brings both improvements out from behind the debug menu and removes them from the debug menu.

### Declarations

### Reviewers
@ColinDayOrg 
### FYIs
@Racel


src/DynamoCore/Search/SearchDictionary.cs
```diff
@@ -372,11 +359,10 @@ internal IEnumerable<V> Search(string query, int minResultsForTolerantSearch = 0
...
             var searchResults = orderedSearchDict.Select(x => x.Key);
-            if (TruncateSearchResults)
-            {
-                searchResults = searchResults.Take(20);
-            }
 
+            // return only the top 20 search results
+            searchResults = searchResults.Take(20);
+            
 #if DEBUG
```
scottmitchell: @ColinDayOrg I'm assuming we want to leave the logging work you did?

Racel:
just doing another pass at all of this. i think we should change the number of search results to 30. will share the results in a bit. still going through the data

scottmitchell:
@Racel Just updated to return top 30 instead of top 20.

src/DynamoCore/Search/SearchDictionary.cs
```diff
@@ -372,11 +359,10 @@ internal IEnumerable<V> Search(string query, int minResultsForTolerantSearch = 0
             var orderedSearchDict = searchDict.OrderByDescending(x => x.Value);
 
             var searchResults = orderedSearchDict.Select(x => x.Key);
-            if (TruncateSearchResults)
-            {
-                searchResults = searchResults.Take(20);
-            }
 
+            // return only the top 30 search results
```
ColinDayOrg: Not that I'm against it, but why the change from 20 to 30 here?

src/DynamoCore/Search/SearchDictionary.cs
ColinDayOrg: I think so, it's not a lot of logging output and the output could be useful.

src/DynamoCore/Search/SearchDictionary.cs
ColinDayOrg: Just saw the conversation in the main comments, it didn't show up in the file comparison view. This is fine with me.

ColinDayOrg:
LGTM :shipit:

src/DynamoCore/Search/SearchDictionary.cs
ColinDayOrg: Odd that file comments don't show up here always and vice versa, so repeating here, it's not a lot of  logging output and it could be useful (and is only under debug anyways) so I would say yes.

Racel:
Hey guys- so, I did a thorough review of the work against the search analysis results. And, the results look like that people are likely to just type in a new search after not finding it in the top 10. Or on the first page, can we please just set the number back to 20. I think our instincts were correct on this one, and it looks like the data backs that up. See the Jira task for more details.

PR #9334, opened 2018-12-20
Heterogeneous list update gives infinite loop

reddyashish:
### Purpose

This PR fixes the bug when using a heterogeneous list in computing some results. It runs into an infinite loop and eventually crashes the Dynamo. 

There is a redundant 'for' loop on line: 217 in replicator.cs. This will recursively add unnecessary(duplicate) values to the ReducedParams list in there, which can be avoided. 
I did some manual testing for other workflows which involved cartesian and zip replication. 

### Declarations

### Reviewers

### FYIs
@alfarok @ColinDayOrg


aparajit-pratap:
Thanks @reddyashish, LGTM!

QilongTang:
LGTM

PR #9374, opened 2019-01-04
Use app switch to use old wpf grid layout algorithm to avoid hang on scaled displays.

mjkkirschner:
### Purpose

Adds an app config to DynamoCoreWPF assembly which contains a single key. This key is suggested in this discussion: https://developercommunity.visualstudio.com/content/problem/244615/setfinalsizemaxdiscrepancy-getting-stuck-in-an-inf.html

as a workaround to a bug introduced in the .net 4.7 timeframe into wpf grid control.

I verified I could reproduce the bug (which Aparajit initially found on bootcamp) with my VM using a high resolution 3000 x 2000 ish - at 300% scaling.

Adding this switch here or earlier prevents the hang. Using this we lose whatever enhancements were made to grid resizing in .net 4.7 - but I think a scaled laptop display these days is very common - and not hanging is more important.

with app switch false:

falsetest2

with app switch true:

truetest4

### Declarations

### Reviewers

@QilongTang 
@aparajit-pratap 
### FYIs

@Racel @smangarole @Dewb


aparajit-pratap:
So you added the config to DynamoCoreWpf and `ViewModel` since it is related to WPF grids? Minor but you could also add a comment to remind devs to remove the `App.config` file too once the bug is fixed in .NET 4.7. LGTM.

PR #9376, opened 2019-01-06
Update TestSessionConfiguration.cs

mjkkirschner:
@QilongTang @smangarole 
try to get the tests running again.


mjkkirschner:
I'm not confident this will fix anything, but it should be done anyway for consistency.

QilongTang:
@mjkkirschner I believe lookup logic always go from newer versions to older, otherwise as a typical dev setup, when running these unit tests locally, with R2019 or R2020, they will never pass right?

QilongTang:
@mjkkirschner Sure, thanks!

QilongTang:
Closing since we don't have to remove the legacy version anymore seems.. If we do, this probably should be part of that PR

PR #5431, opened 2015-10-04
MAGN-8735 Re-enable Visualization tests and move into shared assembly. Update job on CI to target new visualization assembly.

ikeough:
### Purpose

During the recent addition of an updated `HelixWatch3DViewModel` component, it was found that, on our CI systems, we could not create a rendering context during system testing due to the lack of DirectX 10 features on the test machines. A temporary solution was enacted to shut off the creation of the background preview when running tests. But for a number of tests, a proper rendering context is required. These tests were marked as failures in order to get the build running again.

This PR replaces the temporary solution, checking for `DynamoModel.IsTestMode` and conditionally creating the view, with a more robust solution where the type of `Watch3DViewModel` is passed into the constructor of the `DynamoViewModel` through the `StartupParams`. By default, a `DefaultWatch3DViewModel` will be created which allows for the running of system tests, but does not actually render anything. For sandbox, Revit, and the visualization tests, we inject a `HelixWatch3DViewModel` which does render.

A new static constructor has been added,`HelixWatch3DViewModel.TryCreateHelixWatch3DViewModel(...)`. This method of construction attempts to create a valid `HelixWatch3DViewModel`, and if it cannot be created due to the lack of resources on the client, a `DefaultWatch3DViewModel` is created, a message is logged to the console, and the exception is written to the log.

All tests requiring a full 3d rendering context are moved to a new assembly, `WpfVisualizationTests`, where a new base class `VisualizationTest` constructs the `DynamoViewModel` using a `HelixWatch3DViewModel` and thus establishing a proper rendering context.

To test this functionality, we can use the DirectX debugging tools. When we set the "Feature Level limit" to 9_3, and we start Dynamo.
image
Dynamo starts with the background preview disabled. The background preview selection is disabledin the View menu as well.
image
Turning off the feature level limit, and restarting Dynamo puts us, once again, into the mode where the background preview is disabled, as this was the last saved state. Enabling the background preview in the view model returns us to 3D:
image

### Related PRs:
https://github.com/DynamoDS/DynamoRevit/pull/681

### Declarations

Check these if you believe they are true

- [x] The code base is in a better state after this PR
- [x] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [x] The level of testing this PR includes is appropriate
- [x] User facing strings, if any, are extracted into `*.resx` files

### Reviewers

@mjkkirschner 

### FYIs
[...]

PR #9377, opened 2019-01-07
Toggling geometry edge preview setting should not require re-execution of the graph

alfarok:
### Purpose

[DYN-1025](https://jira.autodesk.com/browse/DYN-1025)

The PR ensures `RegenerateAllPackages` is only called on `HelixWatch3DViewModel` or a view model that inherits from `HelixWatch3DViewModel`, thus preventing `RegenerateAllPackages` from being called on `DefaultWatch3DViewModel`.  This doesn't seem to cause any disturbance to the behaviour of the `RevitBackgroundPreview` which implements  `DefaultWatch3DViewModel`.

`RegenerateAllPackages` is only called in 2 places, when toggling edge visibility and when toggling the active state of the specified view.  This change should result in the same current behavior for `DefaultWatch3DViewModel` which will require re-execuction of the graph to generate edges the render packages.

### Concerns

When looking into existing unit tests that appeared to be covering this functionality I noticed none of the `WpfVisualizationTests` are running on the self-serve CI or during the complete install tests.  Running them locally 50 of the 75 are failing with several exceptions I have outlined in [DYN-1462](https://jira.autodesk.com/browse/DYN-1462).  This being said it makes it difficult to determine if any regressions are occuring.

I don't think this is the best fix as more investigation is required into the deeper issues outlined in the spike above.  For example, if several places we are attempting to parse `DefaultWatch3DViewModel` to `HelixWatch3DViewModel` which should never work given the inheritance chain.  This bug has been in existence since Dynamo 1.0 (2015) it's hard to determine when this behavior was originally introduced or if it ever worked after the `RevitBackgroundPreview` was introduced.  See [5431](https://github.com/DynamoDS/Dynamo/pull/5431) for more info.

### Before
old_behavior

### After
new_behavior

### Declarations

### Reviewers

### FYIs

@Racel


Racel:
LGTM. But @mjkkirschner and @alfarok what do you think the risk is of this causing other regressions is since a bunch of tests are failing? Would you recommend putting this into the 2.1 RC?

mjkkirschner:
I'm not sure if this fix should be merged. I don't understand why regenerating default packages would cause an exception to be thrown.

I feel thats what should be addressed /investigated here - @alfarok did you ever determine the exact cause of the exception?

DefaultRenderPackages should still be generated / regenerated on platforms where helix doesn't exist, suppose we are running dynamo core on mac and using default render packages to hold tessellated data before passing it to some rendering extension (like yours) - I would still want regneration to be called when the view was toggled on and off.

I think the change in showEdges is okay because it's isolated - but the Active toggle is worrisome to me because it effects all DefaultRenderPackages as far as I can tell.

alfarok:
@Racel @mjkkirschner I don't think this should be merged either.  This task was originally story pointed a 2 because we were under the impression it was probably simple fix.  I have already exceeded this and haven't been able to get to the source.  It appears to me that it's more complicated and the failing tests should be investigated/resolved before making any WPF visualization changes to monitor regressions.   I can add my notes and add the task back to the backlog which can be revisited when the tests are resolved.

src/DynamoCoreWpf/ViewModels/Watch3D/DefaultWatch3DViewModel.cs
```diff
@@ -95,7 +95,7 @@ public bool Active
 
                 OnActiveStateChanged();
 
-                if (active)
+                if (active && this is HelixWatch3DViewModel)
```
aparajit-pratap: In such cases where you have derived classes needing to override behavior in the base class, it's better design to override the method/property in the derived class and let polymorphism do its job. So in this case it would be overriding `Active` in `HelixWatch3DViewModel`.

alfarok:
@mjkkirschner Do we have any examples of graphs in which nodes have multiple render packages per node?  Revit is the only case I can think of that does this today because there are multiple 3DViewModels. Calling regenerate calls `node.RequestVisualUpdateAsync` so for every node this gets called twice.  I need to double check if this method updates all the render packages at once.  If this is the case this operation is redundant and make sense why everything continues to work with this fix, however if the only render package is a `DefaultRenderPackage` this will break down (such as while using the CLI).  It is also possible that this scenario is already not working and requires the graph to be re-executed but no one has encountered it.  Maybe the black team while originally working with the CLI?

I am having difficulty understanding if this ever worked properly.   If you check out the PR from Ian in 2015 `HelixWatch3DViewmodel` and `DefaultWatch3DViewModel` were pretty much overhauled. Which is around the same time the `RevitBackgroundPreview` was introduced.

`HelixWatch3DViewmodel` : `DefaultWatch3DViewModel` :  `NotificationObject`, `IWatch3DViewModel`, `IDisposable`
`DefaultRenderPackage` : `IRenderPackage`
`HelixRenderPackage` : `IRenderPackage`, `ITransformable`

Several places in the code we try to dynamically cast a `DefaultRenderPackage` to `HelixRenderPackage` which given the inheritance chain above should not work.

PR #9387, opened 2019-01-08
DYN-972: Input symbols in custom nodes with spaces wreak havoc

scottmitchell:
### Purpose

This PR addresses jira task [DYN-972](https://jira.autodesk.com/browse/DYN-972). This PR fixes a bug where input symbols containing spaces were parsed incorrectly. This PR disallows input names with spaces. If a user tries to create an input name containing spaces, the input node inside the custom node workspace is put into an error state and the custom node workspace becomes read-only. The custom node inside the dynamo graph is also put into a warning state.

If a user opens an old graph containing a custom node with an invalid input, the custom node will show a warning. 

If a user opens an old custom node with an invalid input, a message will pop up explaining that there is an invalid input and that the workspace is read only.

### Declarations

dyn972_ui

### Reviewers

@mjkkirschner 

### FYIs

@Racel


src/DynamoCore/Core/CustomNodeManager.cs
```diff
@@ -282,7 +282,17 @@ private static void RegisterCustomNodeInstanceForUpdates(Function node, CustomNo
...
             {
-                node.ResyncWithDefinition(workspace.CustomNodeDefinition);
+                var symbols = workspace.Nodes.OfType<Symbol>().ToList();
+                var removeErrorState = true;
+                foreach (Symbol s in symbols)
+                {
+                    if (!s.Parameter.isValid)
+                    {
+                        node.Error("!");
+                        removeErrorState = false;
```
QilongTang: Should we break here?

scottmitchell: Yeah, good point.

Racel:
@scottmitchell Why "unnamed_input"? Maybe just leave it blank with the ">" ?

src/DynamoCore/Library/TypedParameter.cs
```diff
@@ -13,7 +13,8 @@ public class TypedParameter
         private string summary = null; // Indicating that it is not initialized.
         private readonly string defaultValueString;
 
-
+        public bool nameIsValid { get; set; } = true;
```
mjkkirschner: a public property needs a /// summary and should be named PascalCase 
https://en.wikipedia.org/wiki/Naming_convention_(programming)

mjkkirschner: this is going to be serialized because it's not ignored... I don't think we can do that easily at this point - it's a schema change from the cogs schema. This bool should be ignored if possible from serialization - please check if TypedParameter has a converter for save/load - I think it may , in which case you're done! - if it uses attributes then you may need to ignore this property with [JSONIGNORE]

mjkkirschner: indeed there is a typedParameter converter for serialization:
https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Graph/Workspaces/SerializationConverters.cs#L900

im not sure how deserialization works though - would be worth checking into I think.

src/DynamoCore/Graph/Nodes/CustomNodes/Function.cs
```diff
@@ -309,9 +309,9 @@ private void ValidateDefinition(CustomNodeDefinition def)
         ///     Validates passed Custom Node definition and synchronizes node with it.
         /// </summary>
         /// <param name="def">Custom Node definition.</param>
-        public void ResyncWithDefinition(CustomNodeDefinition def)
+        public void ResyncWithDefinition(CustomNodeDefinition def, bool removeErrorState = true)
```
mjkkirschner: you can't change a public method signature at this point, even adding a default argument is an API breaking change.

mjkkirschner: you can add a new overload and obsolete this method.

src/DynamoCore/Core/CustomNodeManager.cs
```diff
@@ -282,7 +282,18 @@ private static void RegisterCustomNodeInstanceForUpdates(Function node, CustomNo
         {
             Action defUpdatedHandler = () =>
             {
-                node.ResyncWithDefinition(workspace.CustomNodeDefinition);
+                var symbols = workspace.Nodes.OfType<Symbol>().ToList();
+                var removeErrorState = true;
+                foreach (Symbol s in symbols)
```
mjkkirschner: var if possible

src/DynamoCore/Core/CustomNodeManager.cs
```diff
@@ -282,7 +282,18 @@ private static void RegisterCustomNodeInstanceForUpdates(Function node, CustomNo
...
+                {
+                    if (!s.Parameter.nameIsValid)
+                    {
+                        node.Error("This custom node contains an input with an invalid name.");
```
mjkkirschner: this string needs to be pulled out to a resx file so it can be localized.

src/DynamoCore/Core/CustomNodeManager.cs
```diff
@@ -282,7 +282,18 @@ private static void RegisterCustomNodeInstanceForUpdates(Function node, CustomNo
...
+                        removeErrorState = false;
+                        break;
```
mjkkirschner: why break? - what if there are multiple error parameters?

src/DynamoCore/Graph/Nodes/CustomNodes/Function.cs
```diff
@@ -514,24 +520,27 @@ protected override void DeserializeCore(XmlElement nodeElement, SaveContext cont
...
             ArgumentLacing = LacingStrategy.Disabled;
         }
+        
 
         private bool TryParseInputExpression(string inputSymbol, 
                                              out IdentifierNode identifier, 
                                              out AssociativeNode defaultValue,
-                                             out string comment)
+                                             out string comment,
+                                             out List<ErrorEntry> errors)
```
mjkkirschner: it's good for api's to use a general interface if possible - IEnumerable<ErrorEntry> is more appropriate unless you really need a list.

src/DynamoCore/Core/CustomNodeManager.cs
```diff
@@ -282,9 +282,8 @@ private static void RegisterCustomNodeInstanceForUpdates(Function node, CustomNo
         {
             Action defUpdatedHandler = () =>
             {
-                var symbols = workspace.Nodes.OfType<Symbol>().ToList();
                 var removeErrorState = true;
-                foreach (Symbol s in symbols)
+                foreach (var s in workspace.Nodes.OfType<Symbol>().ToList())
```
mjkkirschner: you don't need to call ToList() here - it's an anti pattern.

src/DynamoCore/Graph/Nodes/CustomNodes/Function.cs
```diff
@@ -304,11 +304,22 @@ private void ValidateDefinition(CustomNodeDefinition def, bool removeErrorState
             }
         }
 
+        /// <summary>
+        ///     Validates passed Custom Node definition and synchronizes node with it.
+        /// </summary>
+        /// <param name="def">Custom Node definition.</param>
+        public void ResyncWithDefinition(CustomNodeDefinition def)
```
mjkkirschner: is there any reason to not obsolete this old method?

src/DynamoCore/Core/CustomNodeManager.cs
```diff
@@ -276,13 +276,29 @@ protected virtual void OnCustomNodeRemoved(Guid functionId)
...
+        }
+
+        private static bool ValidateInputSymbolNames(Function node, CustomNodeWorkspaceModel workspace)
+        {
+            foreach (var s in workspace.Nodes.OfType<Symbol>())
+            {
+                if (!s.Parameter.NameIsValid)
+                {
+                    node.Warning("This custom node contains an input with an invalid input symbol.\n" +
+                                 "Please fix the input symbol before saving the custom node.");
```
QilongTang: Since this message will be displayed to the user above node, would you put it into resource dictionary. You can find them by right click on **DynamoCore** Project-> **Properties** -> **Resources**
image

You can always give it a name and added it to the bottom of the list, once added the list will be sorted for you, and make sure you build the solution again. Then you can reference it in the code using `Resources.UnknownVersion`.

src/DynamoCore/Graph/Nodes/CustomNodes/Function.cs
```diff
@@ -460,7 +473,13 @@ public string InputSymbol
                     }
                     else
                     {
-                        Parameter = new TypedParameter(name, type, defaultValue, null, comment);
+                        Error("This input symbol is not currently valid, " +
+                              "and cannot be saved until it is fixed. " +
+                              "Valid inputs take the form:\n\n" +
+                              "name : type = defaultValue\n\n" +
+                              "The input name should be a valid variable name, without spaces. An input type and default value are optional."); //TODO: move to resx");
```
QilongTang: Same for the error message, please put it as resource string so we have localization of it

src/DynamoCore/Core/CustomNodeManager.cs
scottmitchell: @QilongTang Yes, I will before I remove the [WIP] tag.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -1372,6 +1372,15 @@ internal bool containsXmlDummyNodes()
             return this.Nodes.OfType<DummyNode>().Where(node => node.OriginalNodeContent is XmlElement).Count()> 0;
         }
 
+        /// <summary>
+        /// Returns true if the workspace contains input symbols with invalid names.
+        /// </summary>
+        /// <returns></returns>
+        internal bool containsInvalidInputSymbols()
+        {
+            return this.Nodes.OfType<Nodes.CustomNodes.Symbol>().Where(node => !node.Parameter.NameIsValid).Count() > 0;
```
mjkkirschner: you should be able to short circuit this check with any() so that you do not need to iterate all nodes in the graph to check this, just the first matching case. Consider a graph with 10,000 nodes.

mjkkirschner: hmm - I guess it's not such a big issue actually because you filter the list down to symbol nodes first. 👍

src/DynamoCore/Library/TypedParameter.cs
```diff
@@ -13,7 +13,11 @@ public class TypedParameter
         private string summary = null; // Indicating that it is not initialized.
         private readonly string defaultValueString;
 
-
+        /// <summary>
+        /// Indicates whether a valid name has been set for this parameter
+        /// </summary>
+        public bool NameIsValid { get; set; } = true;
```
mjkkirschner: can you add a comment that this property is not serialized?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1939,6 +1939,33 @@ public void AddHomeWorkspace()
...
+            if (workspace.containsInvalidInputSymbols())
+            {
+                DisplayInvalidInputSymbolWarning();
+            }
+        }
+
+        private void DisplayInvalidInputSymbolWarning()
+        {
+            string summary = "This custom node cannot be saved until invalid inputs are removed or fixed."; //TODO: move to resx
+            var description = "This custom node currently contains some invalid inputs, " +
```
mjkkirschner: seems like you can reuse these strings when you move them to the resx.

mjkkirschner:
please add a gif of the new ui when you get a chance.

aparajit-pratap:
@scottmitchell the fact that this is a regression as input nodes could have spaces earlier (versions < 2.0) and that they are broken now, I would guess that the first step would be to determine why it broke in 2.0 and whether what caused the regression could be fixed to restore the previous behavior, don't you think? @mjkkirschner , @QilongTang ?

mjkkirschner:
The current behavior was caused by serializing and deserializing a typedParameter instead of the complete user string. I think this makes maintaining the old behavior quite difficult. We also now rely on the regular input expression parser AFAIK (and now we always add a type definition to the input block).

I think the change in behavior is not necessarily a bad change though- unifying what constitutes a valid input parameter between functions and custom nodes seems like an improvement.

src/DynamoCore/Core/CustomNodeManager.cs
```diff
@@ -241,10 +241,15 @@ protected virtual void OnCustomNodeRemoved(Guid functionId)
 
             CustomNodeWorkspaceModel workspace = null;
             if (loadedWorkspaceModels.TryGetValue(id, out workspace))
+            {
                 RegisterCustomNodeInstanceForUpdates(node, workspace);
+                // Check for invalid input symbols and do an intial definition sync.
+                var removeErrorState = ValidateInputSymbolNames(node, workspace);
```
aparajit-pratap: Can you set the node state to be in an error/warning state using `node.State = ...` and later check for the `State` property on the node in `ValidateDefinition`? This way you'll not have to pass the boolean flag `removeErrorState` into functions `ResyncWithDefinition` and `ValidateDefinition` and then not have to break API's.

aparajit-pratap:
@scottmitchell this looks much cleaner now. Thanks!

test/DynamoCoreTests/CustomNodes.cs
```diff
@@ -1263,10 +1263,10 @@ public void Regress3872_XMLInputNodesDeserilization()
             Assert.IsNotNull(customNodeWs);
             var node1 = CurrentDynamoModel.CurrentWorkspace.NodeFromWorkspace("e058111a-0c58-4dbf-9293-6ab711f530bf") as Symbol;
             Assert.IsNotNull(node1);
-            Assert.AreEqual("x-start: var[]..[]", node1.Parameter.ToCommentNameString());
```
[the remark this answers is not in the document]
scottmitchell: Yeah. If the parser doesn't accept it as a variable name (as an identifier node?), its going to be invalid.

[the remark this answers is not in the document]
scottmitchell: Input names with dashes (or other illegal characters) would have caused the same bug as input names with spaces, before these changes.

src/DynamoCore/Graph/Nodes/CustomNodes/Function.cs
```diff
@@ -460,7 +465,9 @@ public string InputSymbol
                     }
                     else
                     {
-                        Parameter = new TypedParameter(name, type, defaultValue, null, comment);
+                        Error(Properties.Resources.InvalidInputSymbolErrorMessage); 
+                        Parameter = new TypedParameter("", type, defaultValue, null, comment);
+                        Parameter.NameIsValid = false;
```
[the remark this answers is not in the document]
scottmitchell: I'm not immediately sure how. I need to set that property from outside this class here: https://github.com/DynamoDS/Dynamo/pull/9387/files#diff-ff0186e5beab9841afa3a0114087ed9aR132

mjkkirschner:
LGTM

PR #9388, opened 2019-01-08
Fix over-indexing of lists and restore string indexing

aparajit-pratap:
### Purpose

JIRA: 
https://jira.autodesk.com/browse/DYN-995
https://jira.autodesk.com/browse/DYN-1056

image

image

image

image

### Declarations

### Reviewers

@mjkkirschner


src/NodeServices/Attributes.cs
```diff
@@ -342,4 +342,26 @@ public class KeepReferenceThisAttribute : Attribute
...
+    /// to do method resolution and match argument(s) to the function parameter(s). 
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method)]
+    public class AllowArrayPromotionAttribute : Attribute
+    {
+        public bool IsAllowed { get; private set; }
+
+        public AllowArrayPromotionAttribute()
+        {
+            IsAllowed = true;
```
mjkkirschner: so it looks like the default here was false previously? And now it's true? Am I interpreting that correctly? If so - does that have the potential for lots of behavior changes?

src/Libraries/DesignScriptBuiltin/Builtin.cs
```diff
@@ -29,6 +30,17 @@ public static object ValueAtIndex(IList list, int index)
                 }
                 return list[index];
             }
+
+            public static object ValueAtIndex(string stringList, int index)
+            {
+                while (index < 0)
```
mjkkirschner: what is the purpose of this loop?

src/NodeServices/Attributes.cs
aparajit-pratap: Yup or should I add an attribute called `DisallowArrayPromotion` in which case we can get rid of the extra boolean flag.

src/Libraries/DesignScriptBuiltin/Builtin.cs
aparajit-pratap: Indexing from the end of the list with negative indices.

test/Engine/ProtoTest/Associative/MicroFeatureTests.cs
```diff
@@ -1265,39 +1263,6 @@ public void TestDictionaryRegressMAGN619()
             thisTest.Verify("r", 5);
         }
 
-        [Test, Category("Failure")]
```
aparajit-pratap: This test doesn't make any sense.

src/Engine/ProtoCore/RuntimeStatus.cs
```diff
@@ -18,7 +18,7 @@ public enum WarningID
             Default,
             AccessViolation,
             AmbiguousMethodDispatch,
-            AurgumentIsNotExpected,
+            ArgumentIsNotExpected,
```
mjkkirschner: @aparajit-pratap I don't think we can change these now, they are public. 😢 maybe just add a todo?

mjkkirschner:
@aparajit-pratap looks good except I don't think we can fix the enum names now.

aparajit-pratap:
@aparajit-pratap these are changes in ProtoCore. I'm pretty sure no users use any public stuff from ProtoCore. Or are you saying this will cause binary incompatibility issues?

src/Engine/ProtoCore/Parser/AssociativeAST.cs
```diff
@@ -1663,6 +1663,7 @@ public IEnumerable<string> ReturnKeys
         public string ObsoleteMessage { get; protected set; }
         public bool IsObsolete { get { return !string.IsNullOrEmpty(ObsoleteMessage); } }
         public bool IsLacingDisabled { get; protected set; }
+        public bool AllowArrayPromotion { get; protected set; } = true;
```
mjkkirschner: @aparajit-pratap should this also be added to the imperative AST methodAttributes class - does that exist?

mjkkirschner:
@aparajit-pratap looks good to me - one question about imperative behavior vs associative?

src/Engine/ProtoCore/Parser/AssociativeAST.cs
mjkkirschner: 😆

src/Engine/ProtoCore/FFI/CLRFFIFunctionPointer.cs
```diff
@@ -5,8 +5,11 @@
 using ProtoCore.DSASM;
 using ProtoCore.Utils;
 using Autodesk.DesignScript.Runtime;
+using DesignScript.Builtin;
 using ProtoCore.Properties;
 using ProtoCore.Exceptions;
+using IndexOutOfRangeException = DesignScript.Builtin.IndexOutOfRangeException;
+using KeyNotFoundException = DesignScript.Builtin.KeyNotFoundException;
```
QilongTang: Can you make sure the usings are sorted?

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
```diff
@@ -1332,7 +1332,7 @@ void core_Dispose(ProtoCore.RuntimeCore sender)
         {
             CLRObjectMarshaler marshaller = null;
             if (!mObjectMarshlers.TryGetValue(sender, out marshaller))
-                throw new KeyNotFoundException();
+                throw new System.Collections.Generic.KeyNotFoundException();
```
QilongTang: hmm This file has `System.Collections.Generic` using already I believe

src/Libraries/DesignScriptBuiltin/Builtin.cs
```diff
@@ -16,9 +18,17 @@ public static class Get
...
             public static object ValueAtIndex(Dictionary dictionary, string key)
             {
-                return dictionary.ValueAtKey(key);
+                try
+                {
+                    return dictionary.ValueAtKey(key);
+                }
+                catch (System.Collections.Generic.KeyNotFoundException e)
+                {
+                    throw new KeyNotFoundException(e.Message);
```
QilongTang: Same here, I wonder if it is some Resharp bug..

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
mjkkirschner: this is to be explicit because we also have a key not found exception.

QilongTang: @mjkkirschner Then should we rename our own exception in a different task?

mjkkirschner: please a different task. our is 
`DesignScript.Builtin.KeyNotFoundException` 
@aparajit-pratap is aliasing them above.

PR #9390, opened 2019-01-08
do not deserialize camera data that contains NaNs

mjkkirschner:
### Purpose

if a dyn file contains a malformed camera data somehow we should not deserialize it, and instead generate default camera data using our defaults. Angelo reports this was found in a file and results in extremely slow performance as exceptions are raised every frame.

- [ ] add tests

### Declarations

### Reviewers

@QilongTang 

### FYIs
@smangarole @alfarok

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
@@ -113,6 +113,16 @@ namespace Dynamo.Wpf.ViewModels.Watch3D
                    && Math.Abs(this.FarPlaneDistance - other.FarPlaneDistance) < 0.0001;
         }
 
+        internal bool isNaN()
+        {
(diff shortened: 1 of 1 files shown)


src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -113,6 +113,16 @@ public override bool Equals(object obj)
                    && Math.Abs(this.FarPlaneDistance - other.FarPlaneDistance) < 0.0001;
         }
 
+        internal bool isNaN()
```
QilongTang: maybe `containsNaN` is clearer in this case because we are checking if any of the fields `isNan`

QilongTang:
LGTM with one comment

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
mjkkirschner: 👍

QilongTang:
Feel free to merge once the self CI passes

PR #9180, opened 2018-10-15
Upgrade IronPython to v2.7.8 removing dependency on .NET 3.0

alfarok:
### Purpose
[QNTM-5357](https://jira.autodesk.com/browse/QNTM-5357) : Upgrade IronPython to 2.7.8.1
[QNTM-3509](https://jira.autodesk.com/browse/QNTM-3509) : Bundle IronPython w/ Core

This PR upgrades IronPython to v.2.7.8.1 using NuGet and removes any .NET 3.0 dependencies as per the request from corporate Delivery and Wingman teams and allows Revit web installs to be fully silent .  Additionally, it removes the committed IronPython binaries stored in the `extern` folder and the IronPython-2.7.3.msi from the `extra` folder and adds a NuGet reference to the IronPython Standard Library for v2.7.8.1.  When Dynamo is built this package is restored and copied to the bin directory.  The standard library is also now referenced by default.

```
# NO LONGER REQUIRED
import sys
sys.path.append(r"C:\Program Files (x86)\IronPython 2.7\Lib")
```

ironpythoninfo

### Installer Changes
[Installer PR](https://git.autodesk.com/Dynamo/DynamoForRevit/pull/154)

Revit 2020 no longer uses the Dynamo installer but it is still part of our CI/CD pipeline so changes will have to be made to the installer code to skip attempting to install Python using the previously included msi since IronPython will now be built and bundled with Core.

### Testing
All tests are passing on the self-serve.  I have manually verified in Sandbox and D4R in Revit 2020 that the correct version of IronPython is loaded from the `/bin` folder if a different version happens to be installed in the GAC (specifically version 2.7.3 which we used to install).  If an exact match occurs it does not seem possible to prevent loading from the GAC but this should be a non-issue.

ironpythoninfo_testgraph

Also added a test verifying the IronPython assemblies are only getting loaded from a single location.
image

### Declarations

Check these if you believe they are true

- [x] The code base is in a better state after this PR
- [x] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [x] The level of testing this PR includes is appropriate
- [x] User facing strings, if any, are extracted into `*.resx` files
- [x] All tests pass using the self-service CI.
- [x] Snapshot of UI changes, if any.
[...]

PR #8420, opened 2017-12-30
Improve the autocomplete feature of the builtin python IDE

dimven:
### Purpose

I've been having some instabilities when using the built in IDE with large scripts and for extended periods of time and I retraced that back to the current autocomplete implementation. I've revised the current implementation so that it avoids raising as many exceptions as it did. I then tried to improve the type guessing as much as possible.

The type guessing should be relatively accurate for simple statements. We can never guarantee a 100% coverage for complicated variable definitions, because we don't have the luxury of running in the same scope as the actual script or pre-executing the entire code of the script.

Here's a before capture; notice the constant exceptions being thrown in the console:
current autocomplete

And this is with the proposed changes:
new autocomplete

### Declarations

Check these if you believe they are true

- [x] The code base is in a better state after this PR
- [x] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [x] The level of testing this PR includes is appropriate
- [x] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions), and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.

### Reviewers

@mjkkirschner 

### FYIs

This entire class is pretty independent of the rest of Dynamo, so it should work fine in both current and future versions of Dynamo.

PR #9402, opened 2019-01-10
Improve the autocomplete feature of the builtin python IDE

alfarok:
### Background

Picking up the awesome work originally completed by @dimven in #8420

- Addresses instabilities when using the built in IDE with large scripts and for extended periods of time
- Revised current autocomplete implementation to avoid several exceptions
- Improve the type guessing as much as possible

The type guessing should be relatively accurate for simple statements. It is difficult to guarantee 100% coverage for complicated variable definitions as it would require the luxury of running in the same scope as the actual script or pre-executing the entire script.

### Purpose

This PR does the following:

- addresses comments in #8420
- attempts to avoid any API breaking changes
- mark old methods `Obsolete` while making new properties/methods 
private/internal
- points python standard library references to the new [bundled Dynamo Core location](https://github.com/DynamoDS/Dynamo/pull/9180)
- comments/formatting

### Testing

- [x] all [unit tests passing](https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/172/) (regex is tested in `CodeCompletionTests.cs`)

### Examples

The examples in #8420 are all still valid.

ProtoGeometry

protogeometry_autocomplete

Python Standard Library

pythonstdlib_autocomplete

### Declarations

### Reviewers

@dimven @mjkkirschner @QilongTang 

### FYIs

@Racel


alfarok:
Looks like there are some test failures on the self-serve, looking into them now

src/Libraries/PythonNodeModelsWpf/IronPythonCompletionProvider.cs
```diff
@@ -47,6 +44,38 @@ public ScriptScope Scope
...
+        /// </summary>
+        private static string pythonLibDir
+        {
+            get
+            {
+                // Attempt to get and cache the Dynamo Core directory path
+                if (string.IsNullOrEmpty(dynamoCoreDir))
+                {
+                    // Gather executing assembly location
+                    string nodePath = Assembly.GetExecutingAssembly().Location;
```
mjkkirschner: this seems odd - I can see this returning odd results on some systems, could it be worth trying to get access to the pathManager or pathResolver which has all of these paths?

src/Libraries/PythonNodeModelsWpf/IronPythonCompletionProvider.cs
```diff
@@ -97,51 +163,84 @@ public IronPythonCompletionProvider()
...
 
+            // Determine if the Python Standard Library is available in the given context
+            var pythonStdLib = pythonLibDir;
+
+            if (pythonStdLib != null && Directory.Exists(pythonStdLib))
+            {
+                // Try to load Python Standard Library Type for autocomplete
+                try
+                {
+                    var pyLibImports = String.Format("import sys\nsys.path.append(r'{0}')\n", pythonStdLib);
```
mjkkirschner: does this still need to be done?

src/Libraries/PythonNodeModelsWpf/IronPythonCompletionProvider.cs
```diff
@@ -596,124 +979,249 @@ protected Type TryGetType(string name)
                 var val = matches[i].Groups[6].Value.Trim();
                 paramMatches.Add(name, val);
             }
+
             return paramMatches;
+        }
+
+        public List<string> findClrReferences(string code)
```
mjkkirschner: needs a summary tag

src/Libraries/PythonNodeModelsWpf/IronPythonCompletionProvider.cs
alfarok: yes it is required to load the autocomplete for the python std modules.  It is also being done for ProtoGeometry which is included in the template by default.

src/Libraries/PythonNodeModelsWpf/IronPythonCompletionProvider.cs
alfarok: this along with one other method above can actually be private, will replace them now

alfarok:
After pulling in the latest changes from master, down to 7 test failures that need to be addressed
```
1) Test Failure : DynamoPythonTests.CodeCompletionTests.CanImportLibrary
     Expected: 1
  But was:  2

at DynamoPythonTests.CodeCompletionTests.CanImportLibrary()

2) Test Failure : DynamoPythonTests.CodeCompletionTests.CanImportSystemCollectionsLibraryAndGetCompletionData
     Expected: True
  But was:  False

at DynamoPythonTests.CodeCompletionTests.CanImportSystemCollectionsLibraryAndGetCompletionData()

3) Test Failure : DynamoPythonTests.CodeCompletionTests.CanImportSystemLibraryAndGetCompletionData
     Expected: 1
  But was:  2

at DynamoPythonTests.CodeCompletionTests.CanImportSystemLibraryAndGetCompletionData()

4) Test Failure : DynamoPythonTests.CodeCompletionTests.CanMatchAllVariablesSingleLine
     Expected: 1
  But was:  0

at DynamoPythonTests.CodeCompletionTests.CanMatchAllVariablesSingleLine()

5) Test Failure : DynamoPythonTests.CodeCompletionTests.CanMatchAllVariableTypes
     Expected: 5
  But was:  0

at DynamoPythonTests.CodeCompletionTests.CanMatchAllVariableTypes()

6) Test Failure : DynamoPythonTests.CodeCompletionTests.CanMatchImportSystemAndLoadLibraryAndWithComment
     Expected: True
  But was:  False


7) Test Failure : DynamoPythonTests.CodeCompletionTests.DuplicateCallsToImportShouldBeFine
     Expected: 1
  But was:  2

at DynamoPythonTests.CodeCompletionTests.DuplicateCallsToImportShouldBeFine()
```

src/Libraries/PythonNodeModelsWpf/IronPythonCompletionProvider.cs
QilongTang: This class is initialized in ScriptEditor, you can pass in `dynamoViewModel.Model.PathManager.DynamoCoreDirectory` when initializing instead of holding on this new property here, it is a dup of what we have in `PathManager`.

src/Libraries/PythonNodeModelsWpf/IronPythonCompletionProvider.cs
```diff
@@ -47,6 +44,38 @@ public ScriptScope Scope
             set { scope = value; }
         }
 
+        /// <summary>
+        /// Cache storing a reference to the Dynamo Core directory
+        /// </summary>
+        private static string dynamoCoreDir { get; set; }
```
QilongTang: I think we can remove this property, see comments below

mjkkirschner: @alfarok I agree with @QilongTang - that seems like a good place to pass the references you need.

src/Libraries/PythonNodeModelsWpf/IronPythonCompletionProvider.cs
alfarok: ^awesome, thanks @QilongTang!

alfarok: @mjkkirschner I was thinking that's why I would need to duplicate everything from the original since it can't be called from within the new constructor but I wasn't sure

mjkkirschner: ahh, I see, it's a big constructor, yeah you can use chaining to call the other constructor with the `:this() ` syntax.

src/Libraries/PythonNodeModelsWpf/IronPythonCompletionProvider.cs
```diff
@@ -64,84 +66,166 @@ public ScriptScope Scope
...
+
+        internal static readonly string BAD_ASSIGNEMNT_ENDS = ",([{";
+
+        #endregion
+
+        /// <summary>
+        /// Class constructor
+        /// </summary>
+        [Obsolete("Use additional constructor passing the Dynamo Core Directory to enable Python Std Lib autocomplete.")]
+        public IronPythonCompletionProvider() : this("")
```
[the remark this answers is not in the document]
mjkkirschner: I'm not sure! It doesn't change the signature, so it's not immediately obvious that it is a break.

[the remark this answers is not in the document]
QilongTang: I dont think it breaks,
1. Function signature does not change
2. from your code, if it is called without a param, the behavior is still same with before

src/Libraries/PythonNodeModelsWpf/IronPythonCompletionProvider.cs
```diff
@@ -47,6 +44,11 @@ public ScriptScope Scope
             set { scope = value; }
         }
 
+        /// <summary>
+        /// Cache storing a reference to the Dynamo Core directory
```
mjkkirschner: is this comment correct? seems odd.

src/Libraries/PythonNodeModelsWpf/IronPythonCompletionProvider.cs
```diff
@@ -64,84 +66,166 @@ public ScriptScope Scope
...
                     Log(e.ToString());
-                    Log("Failed to load ProtoGeometry types for autocomplete.  Python autocomplete will not see Autodesk namespace types.");
+                    Log("Failed to load ProtoGeometry types for autocomplete. Python autocomplete will not see Autodesk namespace types.");
                 }
             }
 
+            // Determine if the Python Standard Library is available in the given context
+            if(!String.IsNullOrEmpty(dynamoCoreDir))
+            {
+                pythonLibDir = dynamoCoreDir + @"\IronPython.StdLib.2.7.8";
```
mjkkirschner: @alfarok don't do this - instead use System.Path.Combine - it will work cross platform.

test/Libraries/DynamoPythonTests/CodeCompletionTests.cs
```diff
@@ -237,7 +237,7 @@ public void CanImportLibrary()
             var completionProvider = new IronPythonCompletionProvider();
             completionProvider.UpdateImportedTypes(str);
 
-            Assert.AreEqual(1, completionProvider.ImportedTypes.Count);
+            Assert.AreEqual(2, completionProvider.ImportedTypes.Count);
```
mjkkirschner: what are the other types?

mjkkirschner:
@alfarok looks good so far, will review again when you fix the other tests. Two comments:
1. don't build up the paths manually - use the Path.Combine functions for this so this will work on other platforms when run under mono or dot net core.
2. what new imported types are present during the test - where are they coming from?

dimven:
@alfarok thanks for finishing this up, looks great so far! My only comment is on the `GetName` method because with the current implementation the following cases fail:

`a=Line.` - fails because there is no whitespace
`a =Line` - fails with an exception because the equals sign is not discarded

I recommend the following minor change:

```
string GetName(string text)
        {
            text = text.Replace("\t", "   ")
                       .Replace("\n", " ")
                       .Replace("\r", " ");
            int startIndex1 = text.LastIndexOf(' ');
            int startIndex2 = text.LastIndexOf('=');
            return text.Substring(Math.Max(startIndex1, startIndex2) + 1).Trim('.').Trim('(');
        }
```
At this point I'm stating to think it's probably better to just replace this whole method with some regex as well.

src/Libraries/PythonNodeModelsWpf/IronPythonCompletionProvider.cs
```diff
@@ -205,7 +205,7 @@ public IronPythonCompletionProvider(string dynamoCoreDir)
             // Determine if the Python Standard Library is available in the given context
             if(!String.IsNullOrEmpty(dynamoCoreDir))
             {
-                pythonLibDir = dynamoCoreDir + @"\IronPython.StdLib.2.7.8";
+                pythonLibDir = Path.Combine(dynamoCoreDir + @"\IronPython.StdLib.2.7.8");
```
mjkkirschner: @alfarok this is not right - it should be: `Path.Combine(dynamoCoreDir, "IronPython.StdLib.2.7.8");`

alfarok: @mjkkirschner nice catch, fixing now

src/Libraries/PythonNodeModelsWpf/IronPythonCompletionProvider.cs
```diff
@@ -596,124 +908,254 @@ protected Type TryGetType(string name)
...
-            foreach (var import in imports)
+            // Detect all lib references prior to attempting to import anything
+            var refs = findClrReferences(code);
+            foreach (var statement in refs)
             {
-                if (scope.ContainsVariable(import.Key))
+                var previousTries = 0;
+                badStatements.TryGetValue(statement, out previousTries);
+                // TODO - why is this 3?  Should this be a constant?
+                if (previousTries > 3)
```
alfarok: @dimven Thanks for taking a look,  I think we are pretty close!  I will take a closer look at the `GetName` method as it may be related to some of the remaining 4 test failures.  It is also possible the tests are using old methods that aren't updated with the new logic.

Do you happen to remember the significance of this `3` above?  Possibly related to the length of these known assemblies?

```C#
private static string[] knownAssemblies = {
    "mscorlib",
    "RevitAPI",
    "RevitAPIUI",
    "ProtoGeometry"
};
```

alfarok: I also don't see any references to `GetName` in the current implementation which is why I marked it `Obsolete`, but I may be missing something

dimven: Ah true. Didn't notice you split it in two methods.

alfarok:
Down to 1 failure but I am skeptical about a couple others that are simply checking the imported types count, doing a little more digging.  

The failure is `CanMatchImportSystemAndLoadLibraryAndWithComment()` and the first assertion incorrectly passes because `ImportedTypes["None"] = null;` is always present now added on line `164` which was added for `// Special case for python variables defined as null`.  

@dimven I think this is the last real blocker.  Can you elaborate on this change a bit if your recall anything on it?

test/Libraries/DynamoPythonTests/CodeCompletionTests.cs
```diff
@@ -293,6 +292,7 @@ public void CanMatchImportSystemLibraryWithComment()
             var matches = IronPythonCompletionProvider.FindBasicImportStatements(str);
 
             Assert.AreEqual(1, matches.Count);
+            Assert.IsTrue(matches.ContainsKey("System"));
```
mjkkirschner: you could use sequence equal here to be more specific with the test results

src/Libraries/PythonNodeModelsWpf/IronPythonCompletionProvider.cs
```diff
@@ -64,84 +66,166 @@ public ScriptScope Scope
...
+            badStatements = new Dictionary<string, int>();
 
-            RegexToType.Add(singleQuoteStringRegex, typeof(string));
-            RegexToType.Add(doubleQuoteStringRegex, typeof(string));
-            RegexToType.Add(doubleRegex, typeof(double));
-            RegexToType.Add(intRegex, typeof(int));
-            RegexToType.Add(arrayRegex, typeof(List));
-            RegexToType.Add(dictRegex, typeof(PythonDictionary));
+            // Special case for python variables defined as null
+            ImportedTypes["None"] = null;
```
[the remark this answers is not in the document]
alfarok: @mjkkirschner this is the other type that is added by default which is why several tests stepped by 1

alfarok:
@mjkkirschner I think this is ready for a final review, all tests should be passing.  We now handle the scenario in which an import statement ends in `.` which was causing the last failure.  Running the self-serve one last time now to confirm.

test/Libraries/DynamoPythonTests/CodeCompletionTests.cs
alfarok: This was the only location we strictly checked the count without verifying the matching key didn't change.  Should I update this for all instances?

alfarok:
All unit tests are now passing on the self-serve CI

test/Libraries/DynamoPythonTests/CodeCompletionTests.cs
```diff
@@ -434,5 +429,19 @@ public void VerifyIronPythonLoadedAssemblies()
...
+
+        private void verifyImportedTypes(Dictionary<string, Type> importedTypes)
+        {
+            var defaultImports = new string[]
+            {
+                "System",
+                "None"
+            };
+
+            foreach (var type in importedTypes)
```
mjkkirschner: this seems to reimplement sequence equals:
https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.sequenceequal?view=netframework-4.7.2 ?

any reason not to just use it?

dimven:
@alfarok would it be possible to fit in one more usability update for the 2.1 release? If so, I added a secondary generic autocomplete that pops up automatically after every 2nd character and suggest the built-in python methods and the most common "words" used in the script above:

rtcgvo1sgc

That helps me greatly reduces the chance for spelling mistakes and speeds up writing.

test/Libraries/DynamoPythonTests/CodeCompletionTests.cs
alfarok: Should be good now

alfarok:
@dimven I want to get this work merged since the original PR has been open for while and already has quite a few changes.  If you want to open a new PR with those changes we can take a look (much faster than the last PR 😁), it definitely seems like a nice to have feature.  

Unfortunately, these changes will all be going into 2.2 because the 2.1 branch has already been cut and gone through preliminary testing.  I am not 100% sure on this but I think the plan moving forward is to release updates to Dynamo Core more frequently so there should be less time between these releases.

Thanks again for all your efforts, these are some really nice enhancements and I am sure the community will be very pleased!

alfarok:
got the LGTM from @mjkkirschner.  @dimven Any chance you have a semi-long python script that doesn't touch the Revit API that you could share?  Doesn't have to do anything fancy but after speaking with Mike we thought it would be a good to add a test catching any performance degradation in the future with parsing long scripts.

dimven:
@alfarok , sorry for not getting back to you in time. Unfortunately, I didn't have anything non-revit related to test this but maybe one of the sample from the primers will be perfect for this: http://primer.dynamobim.org/13_Best-Practice/datasets/13-2/RoofDrainageSim.zip

It has a semi-long python script at the end of the graph.

dimven:
@alfarok , sorry for not getting back to you in time. Unfortunately, I didn't have anything non-revit related to test this but maybe one of the sample from the primers will be perfect for this: http://primer.dynamobim.org/13_Best-Practice/datasets/13-2/RoofDrainageSim.zip

It has a semi-long python script at the end of the graph.

PR #9408, opened 2019-01-10
DYN1198- Heterogeneous list update gives infinite loop

reddyashish:
### Purpose

This PR is to fix the Heterogeneous list crash that was reported. The infinite loop happens when more than 2 different data types of inputs are used in a list (like String, Int, Double) and that list is passed to some function. 
When calculating replication combinations for a function, there is a "for" loop in Replicator.cs(Lines 217-245) that has a complexity in exponential order. The "reducedParams" list will be containing more than 10^10 possible combinations, which I think is redundant as they are being duplicated. So removed the "for" loop from there and then calculate the value of "arrayStats" (excluding duplicates) from each of the targets in the "targets" list.  "arrayStats" will generally have all the distinct types of data that are passed into the function. 

The graph "test_hetereogenous_list.dyn" in this PR demonstrates the case which was running into infinite loop before. The infinite loop crash and the regressions that were caught last time(on Dynamo master) are fixed here. However there is one particular case that is failing, as demonstrated in "TestTryGetValuesFromDictionary09" test with the nested dictionaries. This used to fail before but we dint catch it until now. 
I have added that test with the "Category : Failure" attribute so that we can revisit this. 

### Declarations

### Reviewers

@aparajit-pratap @QilongTang @mjkkirschner


test/Engine/ProtoTest/Associative/BuiltinMethodsTest.cs
```diff
@@ -1069,6 +1069,23 @@ public void TestTryGetValuesFromDictionary08()
...
 
+        [Test]
+        [Category("Failure")]
+        // To-Do as this use case is failing, re-visit and make sure this test case passes.  
+        public void TestTryGetValuesFromDictionary09()
+        {
+            string code = @"
+a = { ""in"" : 42, ""out"" : 24 };
+b = { ""in"" : 24, ""out"" : 42 };
+c = [[a],[[b]]];
```
reddyashish: Nested Dictionaries work fine but fails when the level of nesting for list "a" i.e 1, is less than the nesting level of list "b" i.e 2. However if C is defined as c = [a,[[b]]], the output is correct.

src/Engine/ProtoCore/Lang/Replication/Replicator.cs
```diff
@@ -214,33 +214,38 @@ public static List<ReplicationInstruction> ReductionToInstructions(List<int> red
                 {
                     //This should generally be a collection, so we need to do a one phase unboxing
                     var targets = reducedParams.Select(r => r[index]).ToList();
-                    foreach (var target in targets)
+                    StackValue target = basicList[index];
```
mjkkirschner: can target be null here, should we check that first?

src/Engine/ProtoCore/Lang/Replication/Replicator.cs
```diff
@@ -214,33 +214,38 @@ public static List<ReplicationInstruction> ReductionToInstructions(List<int> red
...
 
-                        var arrayStats = ArrayUtils.GetTypeExamplesForLayer(target, runtimeCore).Values;
+                    var arrayStats = new HashSet<StackValue>();
 
-                        List<List<StackValue>> clonedList = new List<List<StackValue>>(reducedParams);
-                        reducedParams.Clear();
+                    foreach (var targetTemp in targets)
+                    {
+                        List<StackValue> temp = ArrayUtils.GetTypeExamplesForLayer(targetTemp, runtimeCore).Values.ToList();
+                        arrayStats.UnionWith(temp);
```
mjkkirschner: what is `temp` here - I dont see it in the changes? is it somewhere else near these changes?

QilongTang:
@aparajit-pratap @reddyashish Can you post the edge cases you are working on as several screen shots to help understand?

src/Engine/ProtoCore/Lang/Replication/Replicator.cs
reddyashish: Target cannot be null as we getting its value from basicList and formalParams value is passed into this function based on the input types.

reddyashish:
@QilongTang  I will create some GIFs for that and attach it here.

mjkkirschner:
@reddyashish I would also really like the 1 - 2 sentence description of what this PR does at a high level in the PR description for posterity.

reddyashish:
Different cases demonstrated below:

Working:
passing cases

Failing:
failing

aparajit-pratap:
@reddyashish thanks for adding the failing tests and for the partial fix. We seem to be so close to a solution but yet to fix all the cases. I see that you haven't added failing cases where we have heterogeneous lists of nested lists and dictionaries?

mjkkirschner:
@reddyashish thanks for the write up! 😄

reddyashish:
@aparajit-pratap I have added that failing test now.

src/Engine/ProtoCore/Lang/Replication/Replicator.cs
mjkkirschner: 👍

mjkkirschner:
looks good to me

mjkkirschner:
@Racel @aparajit-pratap @reddyashish do we have a general workaround that we can add to the release notes?

reddyashish:
My thought would be to suggest, not using more than one level of nesting for heterogenous lists that are passed to any function.

reddyashish:
My thought would be to suggest, not using more than one level of nesting for heterogenous lists that are being passed to some node.

src/Engine/ProtoCore/Utils/ArrayUtils.cs
```diff
@@ -117,25 +118,43 @@ public static ClassNode GetGreatestCommonSubclassForArray(StackValue array, Runt
             return runtimeCore.DSExecutable.classTable.ClassNodes[orderedTypes.First()];
         }
 
-        public static Dictionary<int, StackValue> GetTypeExamplesForLayer(StackValue array, RuntimeCore runtimeCore)
+        public static IEnumerable<StackValue> GetTypeExamplesForLayer(StackValue array, RuntimeCore runtimeCore)
```
mjkkirschner: this is an API break I believe.

mjkkirschner: a Dictionary is an `System.Collections.Generic.IEnumerable<System.Collections.Generic.KeyValuePair<TKey,TValue>>` not general Ienumerable of StackValue

src/Engine/ProtoCore/Utils/ArrayUtils.cs
```diff
@@ -117,25 +118,43 @@ public static ClassNode GetGreatestCommonSubclassForArray(StackValue array, Runt
...
+            var usageFreq = new Dictionary<int, List<StackValue>>();
 
             if (!array.IsArray)
             {
-                usageFreq.Add(array.metaData.type, array);
-                return usageFreq;
+                usageFreq.Add(array.metaData.type, new List<StackValue> {array});
+
+                // return flattened list of unique values
+                return usageFreq.Values.SelectMany(x => x.Select(y => y));
```
mjkkirschner: doing `return usageFreq.Values.SelectMany(x => x); won't produce the same result?

src/Engine/ProtoCore/Utils/ArrayUtils.cs
mjkkirschner: I would really appreciate a summary for this method too 😈

src/Engine/ProtoCore/Utils/ArrayUtils.cs
```diff
@@ -117,25 +118,43 @@ public static ClassNode GetGreatestCommonSubclassForArray(StackValue array, Runt
...
+                }
+                else if (!usageFreq.ContainsKey(sv.metaData.type))
+                {
+                    usageFreq.Add(sv.metaData.type, new List<StackValue> {sv});
+                }
             }
-
-            return usageFreq;
+            // return flattened list of unique values
+            return usageFreq.Values.SelectMany(x => x.Select(y => y));
```
mjkkirschner: same question about selectMany

mjkkirschner:
@aparajit-pratap looks good except for the API break and couple questions - can you now remove the failure category from the tests?

reddyashish:
Removed the failure category from the tests and started the self serve.

src/Engine/ProtoCore/Utils/ArrayUtils.cs
aparajit-pratap: @mjkkirschner the problem with `GetTypeExamplesForLayer` was that it returned with only 1 `ArrayPtr` even if there were `[ArrayPtr, ArrayPtr]` but the 2 array pointers can be different. It will always return with the first one.

So it will fail for such cases: `[[1,2,3], [{dict}]]` and it will fail for `[[dict], [[dict]]]` for the same reason, where `dict` is any dictionary.

src/Engine/ProtoCore/Utils/ArrayUtils.cs
aparajit-pratap: I guess you're right.

aparajit-pratap:
@reddyashish I think the self-serve has passed other than the excel tests (which fail all the time).

mjkkirschner:
@aparajit-pratap are you intending to cherry pick this today?

aparajit-pratap:
@mjkkirschner cherry-picked here: #9415

reddyashish:
Thanks Aparajit.

PR #9393, opened 2019-01-09
5909 report bug ux tested

radumg:
### Purpose
NOTE : replaces PR #9150 which was based on wrong branch and adds unit tests.

Fixes #5909 by pre-filling the new Github issue page with 
- an appropriate title that includes Dynamo version number : `Crash report from Dynamo 2.0.1`
- issue body filled with the crash details text, displayed in the `CrashPrompt` window bottom text box

Currently pressing `Submit to Github` takes you here : 

image

With the new behaviour, pressing `Submit to Github` takes you here :
 
image

In the above image, the `This is the error text` body would be replaced by the actual details of the crash stack trace, as per original screenshot in #5909 . 

I can't seem to be able to make Dynamo crash on purpose right now (funny how that works...), so dunno how to replicate.

#### Other notes
- sends users to `https://github.com/DynamoDS/Dynamo/issues/new` by default when no crash text is supplied (friendlier than sending to the super-busy all issues page)
- as per current implementation, you still need to be logged in to Github to be able to submit the issue

#### Further improvements
We could add the rest of the current `issue template` text to the body and wrap the `crash dump` inside a nice code block, but I'm afraid that'd make the query string way too long. Open to ideas/suggestions ?

### Declarations

Check these if you believe they are true

- [x] The code base is in a better state after this PR
- [x] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [x] The level of testing this PR includes is appropriate
- [x] User facing strings, if any, are extracted into `*.resx` files
- [x] All tests pass using the self-service CI.
- [x] Snapshot of UI changes, if any.
- [x] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions), and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.

### Reviewers
@alfarok 
@Racel 

### FYIs
@andydandy74 , @johnpierson

PR #9420, opened 2019-01-14
Revert #9393 on RC2.1.0 branch

alfarok:
### Purpose

This PR reverts #9393 as QA testing revealed the following regression.  When attempting to submit a crash report from Dynamo, while not logged in to GitHub, it is easily possible to exceeding the maximum URI length leading to `414 Request-URI Too Large` error in the browser.  Previously users were taken to `https://github.com/DynamoDS/Dynamo/issues` no matter what which does not require auth.

The team is concerned this will lead to users failing to report errors if they are not always logged in to GitHub.  We will revert this PR in RC2.1 but leave it in Master and address the issue in DYN-1194.

image

### Issue/Solution
```
According to the HTTP specification:

Various ad hoc limitations on request-line length are found in practice. It is RECOMMENDED that all HTTP senders and recipients support, at a minimum, request-line lengths of 8000 octets.

If the URL is too long, the web server fails with the 414 Request-URI Too Long HTTP status code.

The common workaround for these problems is to use POST instead of GET and store the parameters in the request body. The length limits on request bodies are typically much higher than those on URL length. For example, the limit on POST size, by default, is 2 MB on IIS 4.0 and 128 KB on IIS 5.0. The limit is configurable on Apache2 using the LimitRequestBody directive, which specifies the number of bytes from 0 (meaning unlimited) to 2147483647 (2 GB) that are allowed in a request body.
```

### Declarations

### Reviewers

@QilongTang @mjkkirschner @smangarole

### FYIs

@Racel @radumg


mjkkirschner:
LGTM (pr is busted because of eng ops job, please restart with correct params)

QilongTang:
@mjkkirschner Resource changes in it? I will let localization team know

PR #9431, opened 2019-01-16
fix to List.CartesianProduct node

aparajit-pratap:
### Purpose

Fixes regression caused by #9388 in `List.CartesianProduct` and adds missing test.

As a result of #9388 `for` loop behavior had changed. As a result of making `Get.ValueAtIndex` disallow array promotion of its list argument, for loops inside imperative language blocks no longer iterated over a single value. Previously the single value was getting promoted to an array before the `for` loop could iterate over it. This is because the for loop expression is first compiled to AST's as shown in this example: 
```
                x = 0;
                a = [10, 20, 30, 40]
                for(val in a)
                {
                    x = x + val;
                }

                Compiles down to:

                x = 0;
                a = [10, 20, 30, 40];
                val = null;
                %forloop_key = a.key;
                %forloop_expr = a;

                while( %forloop_key != null)
                {
                    val = Get.ValueAtIndex(%forloop_expr, %forloop_key); => Array promotion!!!
                    %forloop_key = %forloop_key + 1;
                    x = x + val;
                }
```
The fix is to add yet another ValueAtIndex like ZT method specifically for `for` loop compilation that does not have the `AllowArrayPromotion(false)` attribute.

### Declarations

### Reviewers

@mjkkirschner 
@QilongTang 

### FYIs

@smangarole

### DIFF
--- a/src/Libraries/DesignScriptBuiltin/Builtin.cs
+++ b/src/Libraries/DesignScriptBuiltin/Builtin.cs
@@ -48,6 +48,25 @@ namespace DesignScript
                 }
             }
 
+            public static object ValueAtIndexInForLoop(IList list, int index)
+            {
+                while (index < 0)
+                {
+                    var count = list.Count;
+                    if (count == 0) break;
+                    index += count;
+                }
+
+                try
+                {
+                    return list[index];
+                }
+                catch (ArgumentOutOfRangeException)
+                {
+                    throw new IndexOutOfRangeException(DesignScriptBuiltin.IndexOutOfRangeExceptionMessage);
+                }
+            }
+
             public static object ValueAtIndex(string stringList, int index)
             {
                 while (index < 0)
--- a/src/Libraries/DesignScriptBuiltin/Builtin.cs
+++ b/src/Libraries/DesignScriptBuiltin/Builtin.cs
@@ -53,6 +53,11 @@ namespace DesignScript
                 return ValueAtIndex(list, index);
             }
 
+            public static object ValueAtIndexInForLoop(string stringList, int index)
+            {
+                return ValueAtIndex(stringList, index);
+            }
+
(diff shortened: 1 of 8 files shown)


test/DynamoCoreTests/Nodes/ListTests.cs
```diff
@@ -945,7 +945,34 @@ public void LaceShortest_Simple()
 
 		}
 
-		[Test]
+        [Test]
+        public void LaceCartesian_Simple()
```
mjkkirschner: @aparajit-pratap can we also add a ds language level imperative test to assert the for loop behavior as well?

src/Libraries/DesignScriptBuiltin/Builtin.cs
```diff
@@ -53,6 +53,11 @@ public static object ValueAtIndexInForLoop(IList list, int index)
                 return ValueAtIndex(list, index);
             }
 
+            public static object ValueAtIndexInForLoop(string stringList, int index)
```
mjkkirschner: what code triggers this?

aparajit-pratap: A `for` loop is now compiled to an AST containing this method. In other words this method is triggered when we run `for` loops.

aparajit-pratap: More specifically it is triggered when we iterate over strings such as here:
```
r = [Imperative]
{
    s = "foo";
    for (x in "bar")
    {
         s  = s + x;
    }
    return = s;
};
```
Expected result: "foobar"

aparajit-pratap:
@mjkkirschner I'm not happy with this solution of adding more `ValueAtIndex` methods just for the `for` loop but this is the simplest solution I have so far. Are you ok with it and the new method names or can you come up with better method names?

PR #9441, opened 2019-01-18
Cherr-pick revert of heterogeneous list fix and add workflow test that highlights performance degradation

aparajit-pratap:
### Purpose

This is a cherry-pick (from RC 2.1) to revert partial fix to heterogeneous lists, which led to a performance degradation with certain cases of homogeneous nested lists. See #9429. It also contains a workflow test taken from Revit test suite that highlighted the degradation in performance. 

Reverting the partial fix causes some edge cases to be unresolved, which are marked as failing test cases in this PR. There is a follow-up bug to resolve these remaining cases (without compromising performance of course) : https://jira.autodesk.com/browse/DYN-1496

### Declarations

### FYIs

@mjkkirschner @QilongTang


QilongTang:
@aparajit-pratap For the new test added, is there other changes needed for it to be part of the performance suite?

aparajit-pratap:
@QilongTang at the moment I don't think we have a "performance test-suite" per se. A good candidate for performance tests I think could be workflow tests and system integration tests. We may need to come up with more specifically tailored to highlight perf aspects once we have some kind of perf test framework in place.

PR #9442, opened 2019-01-18
[WIP] pr for feedback

mjkkirschner:
### Purpose

Expose an API which consumers of DynamoServices can use to stop the shutdown calls to the geometry library if they are going to shut it down later or there is another assembly in process which will shut it down at some other time.

There seems to be an alternative solution which might be even simpler, just avoiding shutting down the geometry library at all, from the libG dlls, since it does not seem this is required.

Will need to think about use cases where this might change behavior. @aparajit-pratap @saintentropy 

Note I tested the use of the API in the notificationsExtension just as a substitute for an "external" client. I will remove it and write a better test if we move forward with this approach. I've verified this fixes the crash when closing revit.

### Declarations

### Reviewers

@aparajit-pratap 

### FYIs

@smangarole

### DIFF
--- a/src/Engine/ProtoCore/FFI/ExtensionAppLoader.cs
+++ b/src/Engine/ProtoCore/FFI/ExtensionAppLoader.cs
@@ -1,8 +1,10 @@
 ﻿using System;
 using System.Collections;
+using System.Diagnostics;
 using System.IO;
 using System.Reflection;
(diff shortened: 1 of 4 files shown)


aparajit-pratap:
@mjkkirschner looks good. 👍

mjkkirschner:
going to close this - replaced by change in libG

PR #7061, opened 2016-08-19
Command check file save before open

mjkkirschner:
### Purpose

http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-10495

1 - This PR adds another command to the `DynamoViewModel` - `OpenIfSaved` command which wraps a `DynamoModel` `OpenFileCommand` with a check for unsaved changes on the home workspace. If there are unsaved changes the message box requesting a save is popped up.
__________
2 - This task also requested exploring a property exposing the modal state of the Dynamo UI - I.E, are there any windows open that would disallow interactions with the `DynamoModel`.

the best I have found is the static class `ComponentDispatcher.EnterThreadModal` event which would allow us to set a bool flag somewhere on the view or view model. Unfortunately message boxes(without owners) do not fire this event so a scan over all message box and other popup / window code would need to be done and consequences of this would need to be tested.
```
 System.Windows.Interop.ComponentDispatcher.EnterThreadModal += (o,e) {modal = true;}

```
I would guess this handler would be set at view model or view construction time. Looking for thoughts and feedback on this, note this PR does not include this handler.

**For now, I will leave this property as a followup task if deemed necessary**

### Declarations

Check these if you believe they are true

- [x] The code base is in a better state after this PR
- [x] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [x] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.

### Reviewers

@ramramps 
@sharadkjaiswal 
### FYIs

@kronz @BogdanZavu

PR #9443, opened 2019-01-19
fix OpenIfSaved API for DynamoPlayer

mjkkirschner:
### Purpose

update the filePath property of the DynamoView as deserializing the view block of json files relies on it being correct.

@Racel Not sure what you want to do with this change regarding 2.1 x 2.2.
In terms of risk, it's small because this API is not called by many clients, I think only Player calls this, perhaps some extensions as well.

While this does not cause a crash, it does cause graphs to fail to open via player edit button without first closing dynamo.

### Declarations

### Reviewers

@QilongTang 

### FYIs

@Racel @smangarole

### DIFF
--- a/test/DynamoCoreWpfTests/WorkspaceSaving.cs
+++ b/test/DynamoCoreWpfTests/WorkspaceSaving.cs
@@ -447,6 +447,12 @@ namespace Dynamo.Tests
 
             //assert the request was made
             Assert.AreEqual(1,eventCount);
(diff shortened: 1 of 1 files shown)


QilongTang:
@mjkkirschner Can you update the unit tests introduced in https://github.com/DynamoDS/Dynamo/pull/7061?

QilongTang:
LGTM

PR #9447, opened 2019-01-20
Dropdown Value Xml Escaping and unescaping Exception

QilongTang:
Please Note:

### Purpose
[DYN-1195](https://jira.autodesk.com/browse/DYN-1195)
Dropdown nodes can fail to maintain previous selection in Dynamo Player.

This PR fixes an edge case that when parsing selection which contains special characters in XML doc like **&**, **<**, **>**, etc, our code failed catch the exception causing unescaping to fail silently.

For example, Revit Fill Pattern selection contains <Solid Fill> which contains invalid XML string when parsing. Internally, we escape this string to be
```
"0:&amp;lt;Solid fill&amp;gt;"
```
After unescaping, the string will went back to
```
"0:<Solid Fill>"
```

Behavior after bug fixing
playerparsingxmlstring

TODO:
 - [x] Unit Tests

### Declarations

### Reviewers

### FYIs

@BogdanZavu @dobriai

### DIFF
--- a/src/Libraries/CoreNodeModels/DropDown.cs
+++ b/src/Libraries/CoreNodeModels/DropDown.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
+using System.Security;
 using System.Xml;
 using CoreNodeModels.Properties;
 using Dynamo.Graph;
@@ -199,11 +200,21 @@ namespace CoreNodeModels
         public static int ParseSelectedIndexImpl(string index, IList<DynamoDropDownItem> items)
         {
             int selectedIndex = -1;
-
+            string name;
             var splits = index.Split(':');
             if (splits.Count() > 1)
             {
-                var name = XmlUnescape(index.Substring(index.IndexOf(':') + 1));
+                // Try unescape serialized string, used for Dynamo 1.X XML deserialization
+                // Try-catch block for DynamoPlayer UpdateValueCore call
+                try
+                {
+                    SecurityElement securityElement = SecurityElement.FromString(index.Substring(index.IndexOf(':') + 1));
+                    name = securityElement.Text;
+                }
+                catch
+                {
+                    name = index.Substring(index.IndexOf(':') + 1);
+                }
                 var item = items.FirstOrDefault(i => i.Name == name);
                 selectedIndex = item != null ?
                     items.IndexOf(item) :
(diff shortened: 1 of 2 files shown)


mjkkirschner:
can you give some more background on this? Do we xml escape strings that we save as json? Why does invalid xml break json deserialization?

QilongTang:
@mjkkirschner This is actually not part of Json deserialization, this expose a problem underneath that the undo, redo, update value commands are still using XML format. So executing these command still going through the XML code. I am not sure why we escape charters in this case, need to check with @sm6srw  why he added this guard in the first place.

mjkkirschner:
okay gotcha, it's a large refactor to remove this code, but it should be done eventually as of now we have to maintain double the serialization logic.

QilongTang:
@mjkkirschner Yes, if we removed the XML serialization/ deserialization code and finished the refactor on time, we should not need this fix at all. But it is what it is.

QilongTang:
Update to use native .Net functions instead of XML reader/writer for string with Xml special char escaping and unescaping. Unit tests added and passing.

Waiting for self CI to pass in https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/176/

QilongTang:
Ping @Dynamo

src/Libraries/CoreNodeModels/DropDown.cs
```diff
@@ -235,23 +236,7 @@ public static string SaveSelectedIndexImpl(int index, IList<DynamoDropDownItem>
             }
 
             var item = items[index];
-            return string.Format("{0}:{1}", index, XmlEscape(item.Name));
-        }
-
-        protected static string XmlEscape(string unescaped)
```
mjkkirschner: @QilongTang are these protected methods called from any of our derived classes? Might they be called from user's derived classes?

QilongTang: They are indeed called from RevitDropDown.cs although only one instance, I originally plan to remove the same calls there using same refactor. But if we feel that removing such **protected** API could cause to 3rd party dev, I can instead simply update these two functions to use the new code I found work. e.g.
```
protected static string XmlEscape(string unescaped)
{
   return SecurityElement.Escape(unescaped);
}

protected static string XmlUnescape(string escaped)
{
  return System.Web.HttpUtility.HtmlDecode(escaped);
}
```

QilongTang: @mjkkirschner @aparajit-pratap @sm6srw Let me know what you prefer, I am fine with either approach.

aparajit-pratap: @QilongTang I like the idea of keeping the methods and refactoring them as you show above. We can add a note that we must remove these in 3.0.

mjkkirschner: +1

mjkkirschner:
LGTM - do we need any updated tests in d4r @ZiyunShang @QilongTang ?

QilongTang:
@mjkkirschner With this implementation, D4R code and tests can stay as its current state. I would like to run D4R self CI though @ZiyunShang @AndyDu1985 😉

mjkkirschner:
@QilongTang do you want to wait to merge this until the tests are run on the revit side?

QilongTang:
@mjkkirschner I think the unit tests in DynamoCore are sufficient. I did not have time to make a final testing on Revit side before merging this.

QilongTang:
@mjkkirschner @aparajit-pratap @smangarole Added a gif for local testing results, merging

PR #9465, opened 2019-01-31
performance profiling of first load of trace data

mjkkirschner:
### Purpose

Element binding is very slow on first run when the size of the element binding data is greater than a few entires.

This change attempts to avoid the exponential cost of looping over all types in all assemblies to find a matching type name - instead we first try to look in assemblies that match in name only, then fallback to our slow matching strategy.

you can see in the below timeline profile that running the sample graph on first load took ~2.5 minutes and after this change runtime is ~25 seconds.

before:
screen shot 2019-01-31 at 3 44 58 pm

after:
screen shot 2019-01-31 at 3 48 29 pm

TODO:
- [x] add tests

### Declarations

### Reviewers

@aparajit-pratap 
@saintentropy 
### FYIs

@smangarole


src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -248,7 +248,21 @@ private class TraceBinder : SerializationBinder
 
             public override System.Type BindToType(string assemblyName, string typeName)
             {
+                //TODO cache
                 var assemblies = AppDomain.CurrentDomain.GetAssemblies().Where(a => !a.IsDynamic);
+                var assemblyNameObj = new AssemblyName(assemblyName);
+                //find matching assemblies by name, version is not used.
+                var matchingAssembly = assemblies.FirstOrDefault(x => x.GetName().Name == assemblyNameObj.Name);
```
aparajit-pratap: You can use `assemblyName` here directly and save the need to create `assemblyNameObj` first?

mjkkirschner: the latter - this is a way to remove the version and culture info from the string reliably.

aparajit-pratap:
@mjkkirschner I've added a test.

mjkkirschner:
@aparajit-pratap thanks! I will build on top of this test with a large element binding example of the scale which replicates the issue.

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -248,7 +248,21 @@ private class TraceBinder : SerializationBinder
...
+                if(matchingAssembly!= null)
+                {
+                   var matchingType = matchingAssembly.GetType(typeName);
+                    if(matchingType != null)
+                    {
+                        return matchingType;
+                    }
+                }
+                //if there was no matching assembly, or type, try all assemblies and all types.
```
aparajit-pratap: As discussed add a "TODO" here to revisit removing this code at a later stage. It might not be needed at all.

mjkkirschner:
@saintentropy @aparajit-pratap @QilongTang I have done the following:
* re-enabled the existing callsite tests that were marked failure after getting them to pass
* added traceBinder unit tests 
* added a perf test that asserts loading trace data for 1500 elements is accomplished in < 20 seconds. Currently this takes ~~2 seconds after this pr with my local set of assemblies.

PTAL

test/DynamoCoreTests/CallsiteTests.cs
```diff
@@ -162,5 +164,49 @@ public void Callsite_ElementBinding()
...
+        public void Callsite_ElementBinding_Timing()
+        {
+            //This graph loads trace data for 1500 "WrapperObjects" in Manual run mode.
+            var ws = Open<HomeWorkspaceModel>(TestDirectory, callsiteDir, "element_binding_large.dyn");
+            var sw = new Stopwatch();
+            sw.Start();
+
+            BeginRun();
+            sw.Stop();
+            Assert.Less(sw.Elapsed.Milliseconds, 20000);
```
aparajit-pratap: How much time would this take earlier?

mjkkirschner: I will check - with the same files I was using with similar sized data it took 2.5 minutes.

aparajit-pratap: Sorry, it took over a min when I ran it again. I might have made a mistake running it the first time. It looked a lot faster then. I think testing for <20s shouldn't be a problem on most machines if you claim that it actually takes just 2s.

mjkkirschner: The tests also do some really simple element binding and element disposal testing.

dimven:
@mjkkirschner while these changes are great and will improve the deserialization of all kinds of custom classes that people have decided to save to their dynamo graph, isn't serializing the entire wrapped object for each bound element too much? I was just wondering if a special and more light-weight method could be implemented for those cases, one that stores *only* the unique id of the element as a string? Right now every public member of the wrapped object will get serialized, when all we need is the guid. All other parameters might have changed between the time the graph was saved last and the latest version of the Revit model it was used with and that would most likely invalidate the trace data.

If that's viable, it could hopefully improve both the robustness and the performance of element binding in general.

mjkkirschner:
Hi @dimven `Right now every public member of the wrapped object will get serialized, when all we need is the guid.` this is not the case - in trace we only serialize the `serializeableID` class which is usually just an id - Directshapes serialize a bit more info. Unfortunately this uses SOAP formatter serialization which serializes a bunch of other info like required for deserialization based on specific assembly version etc. The data is also base 64 encoded which actually makes it larger...

So your initial point is definitely valid, it's just maybe not as bad initially - or bad in other ways 😄 .

You can see what's in trace by base64 decoding it, and then base 64 decoding the individual elements again.

dimven:
Thanks for shining a light of things. A single element serializes to this:

```
<SOAP-ENV:Envelope xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:SOAP-ENC="http://schemas.xmlsoap.org/soap/encoding/" xmlns:SOAP-ENV="http://schemas.xmlsoap.org/soap/envelope/" xmlns:clr="http://schemas.microsoft.com/soap/encoding/clr/1.0" SOAP-ENV:encodingStyle="http://schemas.xmlsoap.org/soap/encoding/">
<SOAP-ENV:Body>
<a1:SerializableId id="ref-1" xmlns:a1="http://schemas.microsoft.com/clr/nsassem/RevitServices.Persistence/RevitServices%2C%20Version%3D2.0.2.6833%2C%20Culture%3Dneutral%2C%20PublicKeyToken%3Dnull">
<stringID id="ref-3">2ffe7cec-9741-4601-97ab-442840b4bec8-001158a3</stringID>
<intID>1136803</intID>
</a1:SerializableId>
</SOAP-ENV:Body>
</SOAP-ENV:Envelope>
```

which is about 750 characters on average. That means that roughly for every 700 elements serialized to a dyn file you save something the size of an average novel :D

PR #9493, opened 2019-02-16
GeometryFactoryPath lookup should be more consistent with other preloader utilities.

mjkkirschner:
### Purpose

Provide a more tolerant GeometryFactoryPath lookup method that falls back to other libG/ASM version combinations that are <= to the major version requested.

* adds tests for invalid/valid inputs of the new method.
* slightly modifies existing summaries to be more explicit on existing non tolerant methods.

### Declarations

### Reviewers

@QilongTang @scottmitchell  

### FYIs

@aparajit-pratap

### DIFF
--- a/src/Tools/DynamoShapeManager/Utilities.cs
+++ b/src/Tools/DynamoShapeManager/Utilities.cs
@@ -348,7 +348,7 @@ namespace DynamoShapeManager
         /// <summary>
         /// Call this method to resolve full path to GeometryFactoryAssembly 
         /// assembly, given the root folder and the version. This method throws 
-        /// an exception if either of the folders/assembly cannot be found.
+        /// an exception if either of the folders or the exact version of the assembly cannot be found.
         /// </summary>
         /// <param name="rootFolder">Full path of the directory that contains 
         /// LibG_xxx folder, where 'xxx' represents the library version. In a 
@@ -368,7 +368,7 @@ namespace DynamoShapeManager
         /// <summary>
         /// Call this method to resolve full path to GeometryFactoryAssembly 
         /// assembly, given the root folder and the version. This method throws 
-        /// an exception if either of the folders/assembly cannot be found.
+        /// an exception if either of the folders or the exact version of the assembly cannot be found.
         /// </summary>
         /// <param name="rootFolder">Full path of the directory that contains 
         /// LibG_xxx_y_z folder, where 'xxx y z' represents the library version of asm. In a 
@@ -410,7 +410,21 @@ namespace DynamoShapeManager
 
             return assemblyPath;
         }
-
+        /// <summary>
+        /// This method will return the path to the GeometryFactory assembly location for a requested version of the geometry library.
+        /// This method is tolerant to the requested version in that it will attempt to locate an exact or lower version of the GeometryFactory assembly.
+        /// </summary>
+        /// <param name="rootFolder">Full path of the directory that contains 
+        /// LibG_xxx_y_z folder, where 'xxx y z' represents the library version of asm. In a 
+        /// typical setup this would be the same directory that contains Dynamo 
+        /// core modules. This must represent a valid directory.</param>
+        /// <param name="tolerantVersion">Version number of the targeted geometry library.
+        /// If the resulting assembly does not exist, this method will look for a lower version match.</param>
+        /// <returns>The full path to GeometryFactoryAssembly assembly.</returns>
+        /// 
+        public static string GetGeometryFactoryPathTolerant(string rootFolder, Version tolerantVersion){
+            return Path.Combine(Utilities.GetLibGPreloaderLocation(tolerantVersion, rootFolder), Utilities.GeometryFactoryAssembly);
+        }
 
         private static IEnumerable GetAsmInstallations(string rootFolder)
         {
--- a/src/Tools/DynamoShapeManager/Utilities.cs
+++ b/src/Tools/DynamoShapeManager/Utilities.cs
@@ -359,27 +359,25 @@ namespace DynamoShapeManager
         /// FileNotFoundException.</param>
         /// <returns>The full path to GeometryFactoryAssembly assembly.</returns>
         /// 
-        [Obsolete("Please use GetGeometryFactoryPathTolerant(string rootFolder, Version version).")]
+        [Obsolete("Please use GetGeometryFactoryPath2(string rootFolder, Version version).")]
         public static string GetGeometryFactoryPath(string rootFolder, LibraryVersion version)
         {
             return GetGeometryFactoryPath2(rootFolder, Preloader.MapLibGVersionEnumToFullVersion(version));
         }
 
         /// <summary>
-        /// Call this method to resolve full path to GeometryFactoryAssembly 
-        /// assembly, given the root folder and the version. This method throws 
-        /// an exception if either of the folders or the exact version of the assembly cannot be found.
+        /// This method will return the path to the GeometryFactory assembly location for a requested version of the geometry library.
+        /// This method is tolerant to the requested version in that it will attempt to locate an exact or lower version of the GeometryFactory assembly.
         /// </summary>
         /// <param name="rootFolder">Full path of the directory that contains 
         /// LibG_xxx_y_z folder, where 'xxx y z' represents the library version of asm. In a 
         /// typical setup this would be the same directory that contains Dynamo 
-        /// core modules. This must represent a valid directory.</param>
+        /// core modules. This must represent a valid directory - it cannot be null.</param>
         /// <param name="version">Version number of the targeted geometry library.
-        /// If the resulting folder does not exist, this method throws an 
-        /// FileNotFoundException.</param>
+        /// If the resulting assembly does not exist, this method will look for a lower version match.
+        /// This parameter cannot be null. </param>
         /// <returns>The full path to GeometryFactoryAssembly assembly.</returns>
         /// 
-        [Obsolete("Please use GetGeometryFactoryPathTolerant(string rootFolder, Version version).")]
         public static string GetGeometryFactoryPath2(string rootFolder, Version version)
         {
             if (string.IsNullOrEmpty(rootFolder))
@@ -393,8 +391,14 @@ namespace DynamoShapeManager
             }
             //IMPORTANT_ Going forward libg folders will be named as follows: libg_major_minor_build - in reference to ASM.
 
-            var libGFolderName = string.Format("libg_{0}_{1}_{2}", version.Major, version.Minor, version.Build);
-            var libGFolder = Path.Combine(rootFolder, libGFolderName);
+            if (version == null)
+            {
+                throw new ArgumentNullException("version");
+            }
+
+            //lookup libG with a fallback to older versions which share the major version number.
+            var libGFolder = Utilities.GetLibGPreloaderLocation(version, rootFolder);
+            
             if (!Directory.Exists(libGFolder))
             {
                 // LibG_version folder must be valid.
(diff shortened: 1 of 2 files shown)


test/DynamoCoreTests/libGPreloaderTests.cs
```diff
@@ -204,6 +204,104 @@ public void GetLibGPreloaderLocation_libGVersionFallback()
...
+            Assert.AreEqual(null, foundVersion);
+            // There is no match, so path to ASM is null.
+            Assert.AreEqual(string.Empty, foundPath);
+
+            // when passed null as a version, GetLibGPreloaderLocation will return LibG_0_0_0
+            Assert.IsTrue(DynamoShapeManager.Utilities.GetLibGPreloaderLocation(foundVersion, rootFolder).ToLower().Contains("libg_0_0_0"));
+
+
+            //when passed a null version this method should return empty string - as the geometry path was not located
+            Assert.AreEqual(string.Empty,
```
mjkkirschner: please see these cases - looking for feedback if these make sense.

src/Tools/DynamoShapeManager/Utilities.cs
```diff
@@ -410,7 +410,26 @@ public static string GetGeometryFactoryPath2(string rootFolder, Version version)
...
+        /// core modules. This must represent a valid directory - it cannot be null.</param>
+        /// <param name="tolerantVersion">Version number of the targeted geometry library.
+        /// If the resulting assembly does not exist, this method will look for a lower version match.
+        /// This parameter cannot be null. </param>
+        /// <returns>The full path to GeometryFactoryAssembly assembly.</returns>
+        /// 
+        public static string GetGeometryFactoryPathTolerant(string rootFolder, Version tolerantVersion){
+            if(tolerantVersion != null && rootFolder != null)
+            {
+                return Path.Combine(Utilities.GetLibGPreloaderLocation(tolerantVersion, rootFolder), Utilities.GeometryFactoryAssembly);
```
QilongTang: The code itself looks fine but I was expecting us to merge some code between `GetGeometryFactoryPath2` and `GetGeometryFactoryPathTolerant` because in the end either DynamoCore or our client would only be available to call one of them. But for now the code path between these two are quite different except for the fall back logic? Do you think if we also still need those folder check logic in this new function?

mjkkirschner: 1. I find these two methods both potentially useful - they do different things, and I think it's always useful to have an explicit method to avoid unknown behavior which is why I did not obsolete the other method.

2. Sandbox AFAIK does not call these methods at all.

3. The end result to the client is like you said that it appears it does a fallback - it DOES have different edge case behavior though - it returns a empty string instead of exceptions when it encounters bad input - that's kind of what I was asking about in the test cases? I'm okay with the more slow failure because this is a `tolerant` version of the method, but open to other opinions for sure.

QilongTang: Sure, I don't mean to obsolete the other one. I thought we want to include the folder check logic in the new API as well because of some legacy bug, if you think it is OK and good enough for the clients then this version should be fine to use. Do you plan to make any change in client code as an example usage of this API later when nuget is up?

aparajit-pratap: I actually don't see any reason to keep `GetGeometryFactoryPath2` after the addition of this new API and I think we should add the folder check exceptions to the new API. I don't see reasons to remove them from here. `GetGeometryFactoryPath2` can give unexpected results (in combination with other API's) like we've seen in the Revit case so I think it should be obsoleted.

src/Tools/DynamoShapeManager/Utilities.cs
```diff
@@ -410,7 +410,26 @@ public static string GetGeometryFactoryPath2(string rootFolder, Version version)
...
+        /// <param name="rootFolder">Full path of the directory that contains 
+        /// LibG_xxx_y_z folder, where 'xxx y z' represents the library version of asm. In a 
+        /// typical setup this would be the same directory that contains Dynamo 
+        /// core modules. This must represent a valid directory - it cannot be null.</param>
+        /// <param name="tolerantVersion">Version number of the targeted geometry library.
+        /// If the resulting assembly does not exist, this method will look for a lower version match.
+        /// This parameter cannot be null. </param>
+        /// <returns>The full path to GeometryFactoryAssembly assembly.</returns>
+        /// 
+        public static string GetGeometryFactoryPathTolerant(string rootFolder, Version tolerantVersion){
```
aparajit-pratap: Should not `tolerantVersion` parameter be more appropriately named `targetVersion`? The API caller passes in a targetted version that he expects to target.

test/DynamoCoreTests/libGPreloaderTests.cs
```diff
@@ -204,6 +204,104 @@ public void GetLibGPreloaderLocation_libGVersionFallback()
...
+            var libG22500path = System.IO.Directory.CreateDirectory(Path.Combine(rootFolder, "LibG_225_0_0"));
+
+            var foundVersion = DynamoShapeManager.Utilities.GetInstalledAsmVersion2(
+                versions, ref foundPath, rootFolder, (path) => { return mockedInstalledASMs; });
+
+            // The found ASM version in this case is a fallback of lowest version within same major which should be 225.2.0
+            Assert.AreEqual(targetVersion, foundVersion);
+            Assert.AreEqual("revit_2020_InstallLocation", foundPath);
+
+            // The found libG preloader version in this case is another fallback of closest version below 225.3.0
```
aparajit-pratap: you mean 225.2.0?

test/DynamoCoreTests/libGPreloaderTests.cs
```diff
@@ -204,6 +204,104 @@ public void GetLibGPreloaderLocation_libGVersionFallback()
...
+            Assert.AreEqual(libG22500path.FullName.ToLower(), DynamoShapeManager.Utilities.GetLibGPreloaderLocation(foundVersion, rootFolder).ToLower());
+
+            //assert that the geometryFactoryTolerant method returns the path of the lib225_0_0 path.
+            Assert.AreEqual(Path.Combine(libG22500path.FullName,DynamoShapeManager.Utilities.GeometryFactoryAssembly).ToLower(),
+                DynamoShapeManager.Utilities.GetGeometryFactoryPathTolerant(rootFolder, targetVersion).ToLower());
+            //assert the non tolerant version throws
+            Assert.Throws<DirectoryNotFoundException>(() => { DynamoShapeManager.Utilities.GetGeometryFactoryPath2(rootFolder, targetVersion); });
```
aparajit-pratap: The fact that `GetGeometryFactoryPath2` is expected to throw an exception in this case proves that it is probably of no use anymore.

test/DynamoCoreTests/libGPreloaderTests.cs
```diff
@@ -204,6 +204,104 @@ public void GetLibGPreloaderLocation_libGVersionFallback()
...
+                DynamoShapeManager.Utilities.GetGeometryFactoryPathTolerant(rootFolder, null));
+
+            //when passed a null root directory this method should return empty string - as a valid root directory is required.
+            Assert.AreEqual(string.Empty,
+                DynamoShapeManager.Utilities.GetGeometryFactoryPathTolerant(null, new Version(224, 24, 24)));
+
+            // when passed a non null, non matching version this method should return a path to the version that was supplied - 
+            // but it will likely fail later.
+            Assert.AreEqual(Path.Combine(libGTestPath.FullName, DynamoShapeManager.Utilities.GeometryFactoryAssembly).ToLower(),
+                 DynamoShapeManager.Utilities.GetGeometryFactoryPathTolerant(rootFolder, new Version(224,24,24)).ToLower());
```
aparajit-pratap: Why not make this throw an exception just like `GetGeometryFactoryPath2` instead? It looks like returning a non-existent location from here will just delay the user in recognizing an error.

test/DynamoCoreTests/libGPreloaderTests.cs
```diff
@@ -204,6 +204,104 @@ public void GetLibGPreloaderLocation_libGVersionFallback()
...
+                    new Version(225,0,0)
+            };
+
+            var mockedInstalledASMs = new Dictionary<string, Tuple<int, int, int, int>>()
+            {
+
+                {"someInstallWithNoMatchingASM" ,Tuple.Create<int,int,int,int>(224,0,0,0)},
+            };
+
+            var targetVersion = new Version(225, 2, 0);
```
aparajit-pratap: Where is this used?

mjkkirschner:
Thanks for the feedback @QilongTang and @aparajit-pratap - I will take another crack at this tomorrow.

test/DynamoCoreTests/libGPreloaderTests.cs
scottmitchell: Wondering the same thing

mjkkirschner:
@QilongTang @aparajit-pratap - went through this again and after it seems to avoid adding any new APIs we can just make the existing `GetGeometryFactoryPath2` method fallback, and keep all the null and directory existing checks.

Clients should not need to update their usage but they will get more robust behavior.

The tests have been updated to verify the edge cases with bad inputs and the comments have been updated for all methods that wrap this method internally.

aparajit-pratap:
@mjkkirschner excellent tests! Thanks, LGTM!

test/DynamoCoreTests/libGPreloaderTests.cs
```diff
@@ -204,6 +204,96 @@ public void GetLibGPreloaderLocation_libGVersionFallback()
...
+
+                {"someInstallWithNoMatchingASM" ,Tuple.Create<int,int,int,int>(224,0,0,0)},
+            };
+
+            // mock a folder with libASMLibVersionToVersion folders with correct names
+            var foundPath = "";
+            var rootFolder = Path.Combine(Path.GetTempPath(), "LibGTest");
+            // both versions of libG exist
+            var libG22500path = System.IO.Directory.CreateDirectory(Path.Combine(rootFolder, "LibG_225_0_0"));
+            var libGTestPath = new DirectoryInfo(Path.Combine(rootFolder, "LibG_224_24_24"));
```
scottmitchell: You end up not using this, right? Also, the "both versions of libG exist" comment in this test threw me off as I was reading through it.

src/Tools/DynamoShapeManager/Utilities.cs
```diff
@@ -392,8 +394,14 @@ public static string GetGeometryFactoryPath2(string rootFolder, Version version)
...
 
-            var libGFolderName = string.Format("libg_{0}_{1}_{2}", version.Major, version.Minor, version.Build);
-            var libGFolder = Path.Combine(rootFolder, libGFolderName);
+            if (version == null)
+            {
+                throw new ArgumentNullException("version");
+            }
+
+            //lookup libG with a fallback to older versions which share the major version number.
+            var libGFolder = Utilities.GetLibGPreloaderLocation(version, rootFolder);
```
QilongTang: With the current implementation of `GetGeometryFactoryPath2` I feel that we can remove the preloadASM() function in D4R since the default call of `GetGeometryFactoryPath` will make the fall back as well, what do you think?

mjkkirschner: yes, I think their implementation can go back to how it was previously.

test/DynamoCoreTests/libGPreloaderTests.cs
mjkkirschner: thanks @scottmitchell will fix this.

QilongTang:
The current implementation makes sense to me in a way that it is very clear to a client which API to use. LGTM

scottmitchell:
LGTM!

mjkkirschner:
@QilongTang @AndyDu1985 @ZiyunShang -  verified this fixes the issue loading ASM on on latest revit build - I confirmed this by moving back before this PR:
https://github.com/DynamoDS/DynamoRevit/pull/2342

PR #9506, opened 2019-02-20
Address failing HelixWatch3DViewModelTests in the WpfVisualizationTests

alfarok:
### Purpose

[DYN-1572](https://jira.autodesk.com/browse/DYN-1572)

This PR addresses the failing `HelixWatch3DViewModelTests` in the `WpfVisualizationTests`

These tests seems to be failing due to timing issues between opening a graph and the UI being updated.  By subscribing to `EvaluationCompleted` on the `DynamoModel` we can call `DispatcherUtil.DoEvents` immediately following to force the dispatcher to dump the entire queue.  `NUnit` has some special behavior regarding `async void` methods and how they interact with the `SynchronizationContext`, more info on this can be found [here](https://stackoverflow.com/questions/15031681/how-does-nunit-successfully-wait-for-async-void-methods-to-complete/15032244#15032244).

image

### Declarations

### Reviewers

### FYIs

@reddyashish


src/VisualizationTests/HelixWatch3DViewModelTests.cs
```diff
@@ -97,6 +97,8 @@ protected override void StartDynamo(TestSessionConfiguration testConfig)
                     ProcessMode = TaskProcessMode.Synchronous
                 });
 
+            Model.EvaluationCompleted += Model_EvaluationCompleted;
```
mjkkirschner: just to be sure, unsubscribe this handler when the test teardown.

mjkkirschner:
@alfarok the explanation in that link sounds kind of like the explanation for doing `SetTimeout(()=>{},0)` in js - does that sound right?

alfarok:
@mjkkirschner yes very similar I think, something along the lines of pausing the execution to let the rendering threads catch up

mjkkirschner:
LGTM - but are you also going to modify the test jobs to include these dlls?

alfarok:
@mjkkirschner yes we need to do that but @reddyashish is still wrapping up the remaining 3 custom node failures which are part of the `WpfVisualizationTests`

PR #7582, opened 2017-02-06
Migrate number nodes with sequence to CBNs.

ikeough:
### Purpose
While developing a migration for the number node, a behavior was noticed in the migration framework that causes a second migration to not be applied when the first migration has the `to` field set to `null`. The `to` field has never really been clear in its intent, and could potentially cause a migration to get "swallowed" by another migration. Ex. you have a migration `from: 0.0.0.1 to: 2.0.0.0` and one `from: 1.0.0.0 to: 2.0.0.0`. The second migration would never be applied because the first would set the current version to `2.0.0.0` during the loop. Similarly, without `to:` defined as in `from: 0.0.0.1`, any other migrations would not get processed because the loop exits when `to:` is null.

This PR marks the `To` field as obsolete on `NodeMigrationAttribute`, and adjusts the migration code to apply all migrations sequentially starting with the one whose version is `>=` the version of the stored workspace. All existing migrations have this intent, with migrations version ranges being sequential. So this change does not affect those migrations, and they will continue to be processed sequentially.

This PR also migrates number nodes with sequences, i.e. `0..5` to CBNs with `0..5;`. Currently, we do not allow the user to input `0..5` in a number node, but we do not disallow the number node from opening such a node. It has been tested to work on number nodes with numeric ranges like `0..5` and ranges with variables like `a..b..c`. All connectors are successfully re-established after migration.

- [x] All tests pass on the self-service CI
- [x] Changes to the API have been listed in the API Changes doc.

PTAL @mjkkirschner 
FYI @gregmarr

PR #9510, opened 2019-02-22
Remove unused code in builtin functions

aparajit-pratap:
### Purpose

Cleanup of unused code written for Builtin functions. These functions have been ported as zero touch methods in 2.0 and are no longer called as they have migration paths that reroute the calls to the equivalent zero touch methods.

### Declarations

### Reviewers

@mjkkirschner 

### FYIs

@ColinDayOrg @QilongTang


mjkkirschner:
@aparajit-pratap one of the smoke tests failed:
```
Errors and Failures:
1) Test Failure : Dynamo.Tests.DynamoSamples.Core_Strings
     Expected: True
  But was:  False

at Dynamo.Tests.DSEvaluationUnitTestBase.SelectivelyAssertValues(MirrorData data, Dictionary`2 selectedValues)
at Dynamo.Tests.DSEvaluationUnitTestBase.SelectivelyAssertValues(String varname, Dictionary`2 selectedValues)
at Dynamo.Tests.DynamoSamples.Core_Strings() in E:\Builds\Jenkins\Dynamo_master\Dynamo\test\Libraries\WorkflowTests\DynamoSamples.cs:line 504
```

can you also run self serve on this?

src/DynamoCore/BuiltInAndOperators/BuiltIn.Migrations.xml
```diff
@@ -72,6 +72,16 @@
       </attribute>
     </additionalAttributes>
   </priorNameHint>
+  <priorNameHint>
+    <oldName>RemoveNulls</oldName>
+    <newName>DSCore.List.Clean</newName>
+    <additionalAttributes>
+      <attribute>
+        <name>preserveIndices</name>
```
mjkkirschner: @aparajit-pratap what does this do? (default values?)

src/Engine/ProtoCore/DSASM/DSArray.cs
```diff
@@ -53,52 +53,6 @@ public void CollectElementsForGC(Queue<StackValue> gcQueue)
             }
         }
 
-        /// <summary>
-        /// Returns true if array contain key or not.
-        /// </summary>
-        public bool ContainsKey(StackValue key)
```
mjkkirschner: @aparajit-pratap do you think this could cause binary compatibility issues for anyone who has references to this dll - what about just obsoleting these? They are public methods...

src/Engine/ProtoCore/Lang/BuiltInMethods.cs
```diff
@@ -12,119 +12,39 @@ public class BuiltInMethods
         public enum MethodID
         {
```
mjkkirschner: kind of worried about this as well, in this case atleast the enum is still present... I'm not sure of the consequence of this.

mjkkirschner: I guess I would prefer we just obsolete each member and remove it at 3.0.

mjkkirschner:
I think unless we can guarantee via research or testing that this change does not effect any client referencing this dll - I guess it's unlikely someone is using these APIs - though it is possible. - I would prefer that we do not remove public enums or methods and instead obsolete them and file a 3.0 task to remove them.

src/DynamoCore/BuiltInAndOperators/BuiltIn.Migrations.xml
```diff
@@ -72,6 +72,16 @@
       </attribute>
     </additionalAttributes>
   </priorNameHint>
+  <priorNameHint>
+    <oldName>RemoveNulls</oldName>
+    <newName>DSCore.List.Clean</newName>
```
QilongTang: @aparajit-pratap I thought the migration framework will not do it's job after 2.0 but maybe our team dis the wrong estimation that the XML code will only used for parsing such file? As long as we keep this file's file format, does it mean we can still do node migration?

mjkkirschner: I think as @alfarok has told us a couple times :) this should work after his most recent changes to the deserialization logic - what we are missing is the framework for doing more complex migrations.

QilongTang: @mjkkirschner @alfarok What kind of more complex migrations are there? If this way is acceptable for Civil, we can still say nodes migration supported in a limited way, can we align within our team?

mjkkirschner: @QilongTang migrations like this:
https://github.com/DynamoDS/Dynamo/pull/7582

src/Engine/ProtoCore/Lang/BuiltInMethods.cs
aparajit-pratap: Ok, will make the change.

src/Engine/ProtoCore/DSASM/DSArray.cs
aparajit-pratap: Will obsolete.

src/DynamoCore/BuiltInAndOperators/BuiltIn.Migrations.xml
aparajit-pratap: It migrates the node and sets the default argument value to `false` as the default value parameter is actually set to `true`.

PR #9170, opened 2018-10-11
QNTM-5580 - Correct Typo in the Dynamo json serialization to match CoGS API

saintentropy:
### Purpose

The purpose of this PR is to correct a typo in the Dynamo serialization of the `Outputs` property as defined in the CoGS spec.  Currently the `InitialValue` property is written as `IntitialValue`.  This PR corrects that issue so Dynamo will serialize the `Outputs` json in the Dynamo file per the CoGS spec.  Note `IntitialValue` is a public property so it remains in the node but is now marked as Obsolete.

### Declarations

Check these if you believe they are true

- [x] The code base is in a better state after this PR
- [x] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [x] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [x] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [x] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions), and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.

### Reviewers

@mjkkirschner @alfarok 

### FYIs

@sm6srw @QilongTang @nonoesp

PR #9516, opened 2019-02-25
Fix typo: recieved -> received

nonoesp:
### Purpose

Rename typos where `received` was typed as `recieved`. This should be a minor change, but there's a breaking change after renaming `OnNotificationReceived` (currently `OnNotificationRecieved`).

### Declarations


mjkkirschner:
Hi @nonoesp thanks for fixing this - unfortunately we can't break a public API like this (thanks for noting it - though I think that sem var check box should *not* be checked)

I think we'll have to fix this like:
https://github.com/DynamoDS/Dynamo/pull/9170

nonoesp:
Thanks for taking a look @mjkkirschner. That makes more sense.

QilongTang:
@mjkkirschner Moving forward, how do we deal with PRs like this? Shall we mark this one as DNM and put in the Dynamo 3.0 bucket?

mjkkirschner:
Thats fine with me.

PR #9523, opened 2019-02-26
Exclude stack trace from being included in the new GitHub issues request

alfarok:
### Purpose

[DYN-1491](https://jira.autodesk.com/browse/DYN-1491)

This PR removes the stack trace data that got appended to the GitHub new issues url.  This was recently introduced in #9393 and has not yet been shipped (only lives in Master not 2.1).  There are several issues with making an authenticated request in a browser window in which we are not prepared to handle.  More details can be found in #9519 and earlier linked PRs.

It has always been the case that users who do not have a GitHub account cannot report issues.  We still gain several improvements with #9393 but automatically copying the stack trace needs to be removed until the team has the bandwidth to implement a more robust solution. 

As previously discussed a future solution could be to use a GitHub app.  The app could handle/post all crash reports tagging the user by their GitHub id but would be optional allowing all users to report issues.  Many in the AEC industry don't use or know much about GitHub.

image

### Declarations

### Reviewers

@QilongTang @mjkkirschner 

### FYIs

@radumg

### DIFF
--- a/src/DynamoCoreWpf/Utilities/CrashUtilities.cs
+++ b/src/DynamoCoreWpf/Utilities/CrashUtilities.cs
@@ -51,20 +51,29 @@ namespace Dynamo.Wpf.Utilities
         {
             return
             "# Issue Description" + Environment.NewLine +
-            "Please fill in the following information to help us reproduce the issue:" + Environment.NewLine +
+            "Please fill in the following information to help us reproduce the issue:" + Environment.NewLine + Environment.NewLine +
+
             "### What did you do?" + Environment.NewLine +
-            "(Fill in here)" + Environment.NewLine +
-            "### What did you expect to see?" + Environment.NewLine +
-            "(Fill in here)" + Environment.NewLine +
-            "### What did you see instead?" + Environment.NewLine +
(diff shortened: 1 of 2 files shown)


QilongTang:
How do we determine the issue tile? I may have asked before but forgot how we define that currently

mjkkirschner:
@alfarok do tests need to be updated?

alfarok:
@QilongTang The title is set [here](https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/Utilities/CrashUtilities.cs#L21) which attempts to append the current Dynamo version.

QilongTang:
@alfarok Thanks. If there is no unit tests to be updated like @mjkkirschner asked, LGTM

alfarok:
@mjkkirschner @QilongTang should be good to go now.  I didn't remove the stack trace parameter from the `BuildMarkdownContent` function so it can still be implemented in the future and just commented out 1 assertion that verified this content was present.

alfarok:
All tests [passing on the self-serve](https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/), merging

PR #9527, opened 2019-02-28
[WIP] DYN-1642: Callsite identifier changes to fix element binding issues with custom nodes 

aparajit-pratap:
### Purpose

The callsite ID and trace data pair serialized to a DYN file are read (if they exist for a node) and stored into TLS only if the ID that's stored in the file matches with the one generated by the compiler for the same node GUID. However if there is a custom node that makes a call to a nested element creation node, the callsite ID generated does not work well. 

This is because the ID is built from a combination of several parameters including the function scope of the creation node itself. In the case of the custom node, the custom node itself is a function in the global scope however the node that it calls into has a non-global function scope, which is the function index of the custom node in the global function table. 

Basing the callsite identifier on something like a function index in the global function table can be very flaky as it could change depending on the order in which global functions (like custom nodes) are compiled and loaded into the table. As a result of this a bug was found where on loading a graph from an older version of Dynamo into a newer version caused element binding to break. This was because the callsite ID serialized to the file did not match with that generated by the compiler.

This PR attempts to fix this situation by ignoring the function scope of the nested method call in the custom node while comparing the callsite ID's.

This change fixes one of the issues with element binding with custom nodes. The replication issue is yet to be addressed.

### Declarations

### FYIs

@QilongTang @mjkkirschner

### DIFF
--- a/src/Engine/ProtoCore/RuntimeData.cs
+++ b/src/Engine/ProtoCore/RuntimeData.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using ProtoCore.AssociativeGraph;
 using ProtoCore.DSASM;
 using ProtoCore.Utils;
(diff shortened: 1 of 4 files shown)


src/Engine/ProtoCore/RuntimeData.cs
```diff
@@ -344,10 +357,32 @@ private string GetAndRemoveTraceDataForNode(Guid nodeGuid, string callsiteID)
...
+            if (compilerGenerated == deserialized) return true;
+
+            var matches1 = Regex.Match(compilerGenerated, fullCallsiteID);
+            var matches2 = Regex.Match(deserialized, fullCallsiteID);
+
+            if (matches1.Groups.Count != matches2.Groups.Count) return false;
+
+            // If both group counts match, they should number 12 in all.
+            // We should ignore checking for the 1st, 7th, and 10th group specifically
+            // as per the Regex pattern (for fullCallsiteID) since that group includes the function scope
```
mjkkirschner: can you provide a sample here -
I'm having a hard time understanding what the 1st,7th,or 9th entry refer to?

mjkkirschner: I guess my reason for wanting to understand this better is to be able to ask more specific questions like - without these 3 groups, can we sufficiently identify these trace data entries from each other and definitely associate them with the correct nested node?

aparajit-pratap: @mjkkirschner had some time to answer you today. See the following example for a callsite identifier for a custom node:
image

In this example, the deserialized and compiler generated ID's match in every respect except for the function scope of the nested node. The function scope generated by an older version of Dynamo, which is serialized in the old DYN file is 110, while the function index of the same custom node when loaded in the latest Dynamo version (compiler generated) is 107. 

If you see the function groups, there are 12 of them broken down as shown. You can see that the 1st, 7th and 10th groups have the function scope number in them (107 in this example), which need to be ignored. This is based on the way that the Regex (funcCallsiteID) is generated.

mjkkirschner: @aparajit-pratap thanks for following up! 😄 
the team will look at finishing this one up ASAP.

PR #9528, opened 2019-02-28
DYN-1541: Issues with code block function default arguments

ColinDayOrg:
### Purpose

This PR addresses issues that were found when deleting code blocks that contain function definitions that have default arguments.

### Declarations

### Reviewers

@mjkkirschner 
@aparajit-pratap


mjkkirschner:
@ColinDayOrg should this still be marked WIP / DNM? Is it ready for review?

src/Engine/ProtoCore/DSASM/Executable.cs
```diff
@@ -141,7 +141,7 @@ public class CodeBlock
         /// <param name="procTable"></param>
         /// <param name="isBreakableBlock"></param>
         /// <param name="core"></param>
-        public CodeBlock(Guid guid, CodeBlockType type, ProtoCore.Language langId, int cbID, SymbolTable symbols, ProcedureTable procTable, bool isBreakableBlock = false, ProtoCore.Core core = null)
+        public CodeBlock(Guid guid, CodeBlockType type, ProtoCore.Language langId, int cbID, SymbolTable symbols, ProcedureTable procTable, bool isBreakableBlock, ProtoCore.Core core)
```
mjkkirschner: this is a breaking change.

ColinDayOrg: ok, you are correct, I will revert it. However, these default arguments are used in the code without being checked for being null. It looks like the code has been this way for a while, not sure what you would like to happen here (other than no changes).

test/DynamoCoreTests/CodeBlockNodeTests.cs
```diff
@@ -212,6 +212,127 @@ public void TestStatement_ArrayReference()
 #endif
         protected double tolerance = 1e-4;
 
+        // Note: This test fails, even though the individual items pass as separate tests
```
ColinDayOrg: This code is here to show an issue with the testing system, it can be removed if desired, not sure what the policy is for this.

test/DynamoCoreTests/CodeBlockNodeTests.cs
```diff
@@ -212,6 +212,87 @@ public void TestStatement_ArrayReference()
...
+            var codeBlockNode1 = CreateCodeBlockNode();
+            UpdateCodeBlockNodeContent(codeBlockNode1, "def test1(x:int = 1, y:int= 2){return = x + y;}test1();");
+            var codeBlockNode2 = CreateCodeBlockNode();
+            UpdateCodeBlockNodeContent(codeBlockNode2, "def test2(x, y = 2, z = 3){return = x + y + z;}test2(1);");
+
+            AssertPreviewValue(codeBlockNode1.GUID.ToString(), 3);
+            AssertPreviewValue(codeBlockNode2.GUID.ToString(), 6);
+        }
+
+        // Note: DYN-1684 - This test is expected to fail due to the functions being indeterminate
```
ColinDayOrg: This test is here for notating the issue of code blocks with duplicate function names. It is not really part of this issue and can be removed if desired, but I left it in to memorialize the issue for the future. Note that once this duplicate name issue is addressed this test will not be an accurate description of the functionality and will need to be changed or removed.

mjkkirschner:
@ColinDayOrg I don't think @aparajit-pratap will be able to review further, would you mind pinging another reviewer to check this out - I will also review.

src/Engine/ProtoAssociative/CodeGen.cs
```diff
@@ -4308,132 +4306,118 @@ private void EmitInlineConditionalNode(AssociativeNode node, ref ProtoCore.Type
...
+                int trueBlockId = Constants.kInvalidIndex;
+                langblockT.codeblock.Language = ProtoCore.Language.Associative;
+                core.AssocNode = bExprTrue;
+                core.InlineConditionalBodyGraphNodes.Push(new List<GraphNode>());
+                EmitDynamicLanguageBlockNode(langblockT, bExprTrue, ref inferedType, ref trueBlockId, graphNode,
+                    ProtoCore.CompilerDefinitions.SubCompilePass.None);
+                List<GraphNode> trueBodyNodes = core.InlineConditionalBodyGraphNodes.Pop();
+
+                // Append dependent nodes of the inline conditional 
+                foreach (GraphNode gnode in trueBodyNodes)
```
mjkkirschner: find this style with no braces kind of difficult to read / prone to future errors - what do you think?

test/DynamoCoreWpfTests/CodeBlockNodeTests.cs
```diff
@@ -182,5 +182,15 @@ public void TestSyntaxHighlightRuleForDigits()
             string[] expected = new string[] { "-2468.2342E+04", "34534.345345", "23423", "-98.7", "0", "10", "2", "-555" };
             Assert.IsTrue(expected.SequenceEqual(actual));
         }
+
+        [Test, RequiresSTA]
+        public void TestFunctionMultipleBlocksDefaultParametersDeleteFirst()
+        {
+            RunCommandsFromFile("DeleteCrashCommands.xml", (commandTag) =>
+            {
+                // NOTE: Nothing needs to be tested here other than that this test does not crash
```
mjkkirschner: does this need to be wrapped in an assert does not throw?

test/DynamoCoreTests/CodeBlockNodeTests.cs
mjkkirschner: what is the failure mode here?

mjkkirschner:
@ColinDayOrg there are a lot of formatting changes here - can you sum up what the spirit of this code change is?

test/DynamoCoreWpfTests/CodeBlockNodeTests.cs
ColinDayOrg: This is probably true, I will add this

test/DynamoCoreTests/CodeBlockNodeTests.cs
ColinDayOrg: The failure mode is that the second test fails even though if the two tests are run in separate test runs they both pass.

ColinDayOrg: If you look at TestFunctionDefaultParameters versus the following TestFunctionSingleDefaultParameter and TestFunctionTwoDefaultParameters you can see that the tests are the same, yet if they are concatenated into one test the test will fail. I did not really look into why, I was just notating that there is potentially an issue with the testing system there that may trip up other people in the future.

QilongTang:
Same for me, pretty hard to understand this and I am thinking about debugging myself to understand the intention better

test/DynamoCoreTests/CodeBlockNodeTests.cs
mjkkirschner: hmm - the difference between running these in a single test and and in multiple tests is that Dynamo is being shut down and new workspaces are being opened in the new tests... that could be a problem with the testing system or it could be a real bug...

mjkkirschner: when you interact manually with Dynamo you cannot reproduce this failure?

src/Engine/ProtoAssociative/CodeGen.cs
ColinDayOrg: I actually find the requirement to have the extra braces unnecessary and creates code that is more unreadable (mostly due to adding unnecessary code and code lines), but the bracing wars is a long running issue across multiple companies, I really have no dog in this race. That being said, these changes were made by Aparajit, I can change them however you would like.

ColinDayOrg:
After debugging and determining what the cause of the issue was; basically if there is a default argument in a function statement in a code block, when the code block is deleted there is a child code block (code block on the VM level, not a code block node) that gets deleted, which causes the indexing to get out of sync leading to an assertion failing which causes the VM crash. My original solution was to not delete code blocks that were created due to a default argument, but after suggesting this Aparajit said that it was probably better to not add the child code blocks in the first place. This is where most of these code changes came from, I would be happy to elaborate or have a further conversation if it would be helpful.

test/DynamoCoreTests/CodeBlockNodeTests.cs
ColinDayOrg: To memorialize our recent Zoom conversation, each test passes separately, but putting them both in the same test case fails. Put another way, I can reproduce both issues manually, but in the testing framework having both tests run together fails. I would guess at this point that it is an issue with the testing framework and not Dynamo, but this is just a guess at this point. The main thought was just to get it notated so that other people may avoid some future pain, I'm not sure how much of importance this particular issue has for the everyday user.

alfarok:
Hey @ColinDayOrg can you provide a sample of how to reproduce the crash that is now addressed?  I am looking at the associated Jira task DYN-1541 and am still a bit confused.  In the description the following sample is provided:
```
def foo(x, y = 1, z = 2)
{
    return = x + y + z;
};
foo(2); // returns null instead of 5
```
Mocking this locally (without these changes) I get a result of 4 not null or 5.  This may not be the issue that is being addressed but I am trying to better understand what is addressed in this PR vs what the task outlines.

alfarok:
@ColinDayOrg Thanks for the detailed explanation of the issue on Friday.  I was able to reproduce the VM crash locally and can verify it is no longer occurring with these changes.  I beat around on it a bit with operations such as undo/redo with no further issues.  As we discussed last week it would be good to get this merged into 2.3 if the team is okay, keeping an eye out for any strange behavior as it gets heavier usage since these changes run pretty deep.

mjkkirschner:
@ColinDayOrg after you file the followup for the test LGTM.

PR #8122, opened 2017-08-21
Python template support 2.0

radumg:
*Note : this PR replaces the previous PR #8034 as that was based on `1.3.1` branch and not on master. Re-basing was messy, so cherry-picked to this new PR instead.*

### Purpose

This PR fixes #7604 , wish for adding a Python template that can be used to populate any `Python Script` nodes when adding to workspace.

#### Current behaviour
- user adds `Python Script` node to canvas
- the node is populated with a few lines of code, but these are hard-coded in the Dynamo source code, so users are unable to change them.

#### New behaviour
- user adds `Python Script` node to canvas
- Dynamo checks if the `PythonTemplate.py` file exists at the user location root (`%appdata%/Dynamo/Core/{version}/`) and if it does, it reads the file and populates the Python script node with its contents
- if the file doesn't exits or it's empty, Dynamo falls back on the hard-coded behaviour

Behaviour when template file is missing or empty : 
default python template

Behaviour when template file is found :
python custom example dynamo2.0

This allows users and organisations to rollout the template as part of users profiles or as logon scripts.

### Implementation
- refactored `PreferenceSettings` class to include code regions, moving properties into their logical group, also fixing some comments/documentation at same time, making it clearer.
- implemented a `PythonTemplateFilePath` property in the settings class so it serialises and deserialises with user settings
- implemented required properties in the interfaces `IPreferences`, `IPathResolver` and `PathManager` class
- implemented a `static` backing store called `pythonTemplateFilePath`. This is what the property above (part of interface) `get`s and `set`s. This was necessary as static props cannot be part of the interface.
- exposed the private static property with a `public static` method called `GetPythonTemplateFilePath`, so it's accessible by nodes and anything else running in Dynamo context, such as ZT nodes.
- added a few additional lines to the hardcoded python template, providing pointers to users where to start writing code and explaining the first few lines (clr imports, etc). This is following my observations in-person and on forums that most new users struggle with those 2 things : where to put code and what those bits mean.
- replaced `\n` in hardcoded python code template with .NET standard `Environment.NewLine` line breaks, should behave more predictably when compiled on different platforms
- replaced hardcoded comments in python code with texts extracted into `.resx` variables, to add support for localisation of code comments too.
[...]

PR #9533, opened 2019-03-01
Provide a strategy for host integrator's to set the python template at startup

alfarok:
### Purpose
[DYN-1507](https://jira.autodesk.com/browse/DYN-1507)

This PR provides a strategy for a Dynamo host integrator to programmatically set the path of the Python nodes default template at startup.  This is accomplished by setting the `PythonTemplatePath` on the configuration file that will be passed as a parameter to the `DynamoModel` when instantiated.  It is required that the host integrator uses either the `DefaultStartConfiguration` struct or creates a struct that inherits from `DefaultStartConfiguration`.  For example...

```C#
var config = new DynamoModel.DefaultStartConfiguration()
            {
                PythonTemplatePath = "C:\\Users\\alfarok\\Desktop\\PythonTemplate.py"
            };

var model = DynamoModel.Start(config);
```

Users are still able to override this file if desired by modifying the `<PythonTemplateFilePath>` attribute in the `DynamoSettings.xml` file or adding a a new template file in the appropriate `AppData` location (see below).  To fallback to either host integrator path or the [hard-coded template](https://github.com/DynamoDS/Dynamo/blob/6db61fe4b6c8714f82eed8b0d5a73aa548fcfd54/src/Libraries/PythonNodeModels/PythonNode.cs#L103) provided in Dynamo Core simply remove the path from the settings xml or delete the template file from the `AppData` location.  Lastly, if a invalid path is provided the default Core template will be applied.

### Logging

(In sequential order)

#### Loaded from DynamoSettings.xml
image

#### Loaded from Host Integrator
image

#### Loaded from AppData
image

#### Loaded Default Template
image

### Overriding the default or a host integrator template as a USER
Users looking to override the template can do so in the following ways introduced in #8122:
#### Option 1 - Default AppData Location
Add a new python file named `PythonTemplate.py` to the `C:\Users\alfarok\AppData\Roaming\Dynamo\Dynamo Core\2.2` directory

#### Option 2 - Custom Location using DynamoSettings.xml
Add a custom path to the `DynamoSetting.xml` under the `PythonTemplateFilePath` tag such as
`<PythonTemplateFilePath>C:\Users\alfarok\Desktop\PythonTemplate.py</PythonTemplateFilePath>`
(Does NOT require `PythonTemplate.py` naming)

### Testing
Unit tests have been added to verify API usage for valid and invalid template paths.  Additionally, this has been manually tested by modifying the DynamoSandbox startup config to include a `PythonTemplatePath` which is set [here](https://github.com/DynamoDS/Dynamo/blob/224f5df4eb7b7b999ae3818e2e3dba3b789fa493/src/DynamoApplications/StartupUtils.cs#L183) and invoked [here](https://github.com/DynamoDS/Dynamo/blob/8afcde02f859f5dd4f261d21edb32e2b94e9450a/src/DynamoSandbox/DynamoCoreSetup.cs#L40).  Testing coverage for loading the default template as well as a user template already has coverage.

### Declarations

### Reviewers

@QilongTang @mjkkirschner

### DIFF
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -521,6 +521,7 @@ namespace Dynamo.Models
             public IEnumerable<IExtension> Extensions { get; set; }
             public TaskProcessMode ProcessMode { get; set; }
             public bool IsHeadless { get; set; }
+            public string PythonTemplatePath { get; set; }
         }
 
         /// <summary>
@@ -640,17 +641,42 @@ namespace Dynamo.Models
             // If not, check the default filepath for the Python template (userDataFolder\PythonTemplate.py).
             // If that doesn't exist either or is empty, Dynamo falls back to the hard-coded Python script template.
             // We log the behaviour to make it easy for users to troubleshoot this.
-            if (string.IsNullOrEmpty(PreferenceSettings.PythonTemplateFilePath) || !File.Exists(PreferenceSettings.PythonTemplateFilePath))
+
+            // If a custom python template path doesn't already exists in the DynamoSettings.xml
+            if(string.IsNullOrEmpty(PreferenceSettings.PythonTemplateFilePath) || !File.Exists(PreferenceSettings.PythonTemplateFilePath))
             {
-                if (string.IsNullOrEmpty(pathManager.PythonTemplateFilePath) || !File.Exists(pathManager.PythonTemplateFilePath))
-                    Logger.Log(Resources.PythonTemplateInvalid);
-                else
+                try
                 {
-                    PreferenceSettings.PythonTemplateFilePath = pathManager.PythonTemplateFilePath;
-                    Logger.Log(Resources.PythonTemplateDefaultFile + " : " + pathManager.PythonTemplateFilePath);
+                    // To supply a custom python template host integrators should supply a 'DefaultStartConfiguration' config file
+                    // or create a new struct that inherits from 'DefaultStartConfiguration' making sure to set the 'PythonTemplatePath'
+                    // while passing the config to the 'DynamoModel' constructor.
+                    var configCast = (DefaultStartConfiguration)config;
+                    var templatePath = configCast.PythonTemplatePath;
+
+                    // If a custom python template path was set in the config apply that template
+                    if (!string.IsNullOrEmpty(templatePath) && File.Exists(templatePath))
+                    {
+                        PreferenceSettings.PythonTemplateFilePath = templatePath;
+                        Logger.Log(Resources.PythonTemplateDefinedByHost + " : " + pathManager.PythonTemplateFilePath);
+                    }
+
+                    // Otherwise fallback to the default
+                    else
+                    {
+                        SetDefaultPythonTemplate();
+                    }
+                }
+                catch
+                {
+                    // Fallback to the default
+                    SetDefaultPythonTemplate();
                 }
             }
-            else Logger.Log(Resources.PythonTemplateUserFile + " : " + pathManager.PythonTemplateFilePath);
+            else
+            {
+                // A custom python template path already exists in the DynamoSettings.xml
+                Logger.Log(Resources.PythonTemplateUserFile + " : " + pathManager.PythonTemplateFilePath);
+            }
 
             pathManager.Preferences = PreferenceSettings;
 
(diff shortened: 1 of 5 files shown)


src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -640,17 +641,42 @@ protected DynamoModel(IStartConfiguration config)
...
             {
-                if (string.IsNullOrEmpty(pathManager.PythonTemplateFilePath) || !File.Exists(pathManager.PythonTemplateFilePath))
-                    Logger.Log(Resources.PythonTemplateInvalid);
-                else
+                try
+                {
+                    // To supply a custom python template host integrators should supply a 'DefaultStartConfiguration' config file
+                    // or create a new struct that inherits from 'DefaultStartConfiguration' making sure to set the 'PythonTemplatePath'
+                    // while passing the config to the 'DynamoModel' constructor.
+                    var configCast = (DefaultStartConfiguration)config;
```
alfarok: This is done because the `DynamoModel` constructor takes a `IStartConfiguration` config as its only param.  I believe modifying the interface (even adding a new property) would be an API breaking change, but feel free to correct me if I am wrong on this or there is a better strategy.

mjkkirschner: I think you should do something like:
`defaultConfig = config as DefaultStartConfiguration` and then check for null because it's possible that the concreteType is NOT a DefaultStartConfiguration ... and that needs to be handled. (ie missing the python template field you just added)

mjkkirschner: alternatively you can add a second interface and check if the object implements that interface, if it does - do the right thing and set the path.

QilongTang: 👍 Same options in https://jira.autodesk.com/browse/DYN-1644

QilongTang:
@alfarok Once you wrap this up, please also attach the screen shot of console with this change so clients know what to expect from console. Thanks

src/DynamoCore/Models/DynamoModel.cs
mjkkirschner: in general, I don't think casting like this is a great idea for the problem we are discussing as casting can throw - `as` and `is` should not.'

https://docs.microsoft.com/en-us/dotnet/api/system.invalidcastexception?view=netframework-4.7.2

alfarok: Thanks for the feedback guys.  I was not happy with this approach either but `as` cannot be used with a non-nullable value type which `DefaultStartConfiguration` is as a struct.  This is why I wrapped this logic in a try/catch returning the default template if an exception is thrown.  However, performing a check such as 
```C#
if(config is DefaultStartConfiguration)
{
    // Proceed in checking if a custom template path is available and valid
}
else
{
    // Use default template
}
```
is possible but does not guaranteed the PythonTemplatePath has been set.  Attempting to retrieve the `PythonTemplatePath` property will simply return null if it was not set and I verify this is not the case below before proceeding.  Long story short we can probably use `is` to verify the `config` is of the `DefaultStartConfiguration` type or of a type that inherits from it.  I don't think adding another struct is necessary just for this property. If the `config` is of another type we return the default template.  I believe the try/catch can also be removed with `is` as @mjkkirschner suggested since it will not throw.  What do you guys think about this approach? @QilongTang

QilongTang: Sounds good to me.

alfarok:
@QilongTang I just added one new logging statement for when a custom template is loaded via the config, however it turns out none of the old statements ever successfully logged to the console or log file.  Possibly because the `DynamoModel` is still being instantiated?  I need to debug more but at first glad I don't see any exceptions thrown stepping through the logger

QilongTang:
@alfarok Sure, could be related to the sequence of how preference setting and logger are initialized

alfarok:
@QilongTang Added screenshots for all the logging statements.  There were a couple issues with the previous logging logic which took me a little while to uncover.  It appeared the logging wasn't working because the localized strings were embedded only in the `Resources.resx` file but should have also be in the `Resources.en-US.resx`, so logging was working but printing empty strings.  I believe I enumerated all the scenarios and verified the correct statements align with the correct logic.  Will get an some unit tests added and this should be good to go.

alfarok:
Added unit tests verifying API usage for valid and invalid template paths.  I think this should be ready to go

QilongTang:
Look perfect on my side!

PR #9542, opened 2019-03-05
DYN-1476: AllowRankReductionAttribute on Zero Touch seems to be broken

scottmitchell:
### Purpose

Jira issue [DYN-1476](https://jira.autodesk.com/browse/DYN-1476). 

This PR addresses a bug where the `AllowRankReductionAttribute` does not work for Zero Touch properties. The typical use case for the `AllowRankReductionAttribute` is to reduce the rank of the return value of methods—rather than return a list of a single item, the method will return just the item. This does not currently work for properties because a property's attributes are not automatically applied to that property's getter (which is what actually get's invoked by the Zero Touch node). For example, this does not currently result in rank reduction of the `get_SomeProperty` return value:

```C#
[AllowRankReduction] // This won't have any effect
public List<object> SomeProperty => new List<object> { someObject };
```

A work around is possible by explicitly declaring the getter with `AllowRankReductionAttribute`:

```C#
[AllowRankReduction]
public List<object> SomeProperty2
{
    [AllowRankReduction] // Adding this one will make it work
    get
    {
         return new List<object> { someObject };
    }
}
```
There are a couple possible solutions here: 

**Solution 1:** Change nothing, and simply message to users that they must explicitly declare getters if they want to use this attribute. However, this behavior could be confusing to a Zero Touch developer.

**Solution 2:** Copy the attribute from the property to its getter at runtime—that could happen [here](https://github.com/DynamoDS/Dynamo/blob/e151cb9bf26f7704be681d95933378544fe24277/src/Engine/ProtoCore/FFI/CLRDLLModule.cs#L385-L397). This may be possible, but certainly isn't trivial. Some info on that here: [https://stackoverflow.com/questions/14663763/how-to-add-an-attribute-to-a-property-at-runtime](https://stackoverflow.com/questions/14663763/how-to-add-an-attribute-to-a-property-at-runtime)

**Solution 3:** Add another rank reduction check later in the function invoke process, where the property attributes are available—maybe [here](https://github.com/DynamoDS/Dynamo/blob/ec10f936824152e7dd7d6d019efdcda0d78a5264/src/Engine/ProtoCore/Lang/FFIFunctionEndPoint.cs#L154).

**Solution 4 (this PR):** Create a way to pass the property's `AllowRankReductionAttribute` down to the `FFIMemberInfo` instance created from the getter, without actually adding the attribute to the getter. This seems to be the least invasive way to add this functionality to Zero Touch properties. However, it may not be the cleanest.

### Declarations

### Reviewers

### FYIs

@QilongTang


src/Engine/ProtoCore/FFI/CLRFFIFunctionPointer.cs
```diff
@@ -64,6 +64,26 @@ protected FFIMemberInfo(MemberInfo info)
             Info = info;
         }
 
+        public void CheckForRankReductionAttribute(Dictionary<MethodInfo, Attribute[]> getterAttributes)
+        {
+            Attribute[] atts = null;
+            if (null != Info as MethodInfo)
```
aparajit-pratap:
```
var info = Info as MedthodInfo;
if(info != null)
{
    if (getterAttributes.TryGetValue(info, out atts))
    { ...
```

src/Engine/ProtoCore/FFI/CLRFFIFunctionPointer.cs
```diff
@@ -64,6 +64,26 @@ protected FFIMemberInfo(MemberInfo info)
...
+            {
+                if (getterAttributes.TryGetValue(Info as MethodInfo, out atts))
+                {
+                    var attributes = atts;
```
aparajit-pratap: Why can't we just use `atts` here?

aparajit-pratap:
@scottmitchell I think this is a clean and simple solution! I would "LGTM" this after some tests :)
Thanks!

src/Engine/ProtoCore/FFI/CLRFFIFunctionPointer.cs
```diff
@@ -64,6 +64,26 @@ protected FFIMemberInfo(MemberInfo info)
             Info = info;
         }
 
+        public void CheckForRankReductionAttribute(Dictionary<MethodInfo, Attribute[]> getterAttributes)
```
aparajit-pratap: I think we can make this `internal` instead - safer than `public`.

src/Engine/ProtoCore/FFI/CLRFFIFunctionPointer.cs
```diff
@@ -64,6 +64,26 @@ protected FFIMemberInfo(MemberInfo info)
             Info = info;
         }
 
+        internal void CheckForRankReductionAttribute(Dictionary<MethodInfo, Attribute[]> getterAttributes)
```
mjkkirschner: name is okay to me, but just considering others - copyRankReductionAttribute?

mjkkirschner:
it's looks good to me, lets add some tests.

src/Engine/ProtoCore/FFI/CLRDLLModule.cs
```diff
@@ -782,7 +782,10 @@ private void RegisterFunctionPointer(string functionName, MemberInfo method, Lis
             if (CoreUtils.IsDisposeMethod(functionName))
                 f = new DisposeFunctionPointer(Module, method, retype);
             else if (CoreUtils.IsGetter(functionName))
+            {
                 f = new GetterFunctionPointer(Module, functionName, method, retype);
+                (f as GetterFunctionPointer).ReflectionInfo.CheckForRankReductionAttribute(mGetterAttributes);
```
QilongTang: Utilizing ReflectionInfo is usually performance sensitive especially under a for loop, can you give us your estimation of the performance impact? Is it highly depending on how many getters there are?

src/Engine/ProtoCore/FFI/CLRFFIFunctionPointer.cs
QilongTang: Would you add some function comments, what you have mentioned in this PR description is good enough

QilongTang:
LGTM with a few comments and looking forward to tests

aparajit-pratap:
@scottmitchell you can add dummy test classes and methods/properties to `FFITarget` project. It's a test zero touch library we use in our FFI (Foreign Function Interface) tests.

src/Engine/ProtoCore/FFI/CLRFFIFunctionPointer.cs
scottmitchell: @QilongTang yes, I'll add some comments. @mjkkirschner yeah I like that better. Or maybe "AttachRankReductionAttribute"

scottmitchell:
Thanks @aparajit-pratap, I was going to ask about that

src/Engine/ProtoCore/FFI/CLRDLLModule.cs
```diff
@@ -784,7 +784,7 @@ private void RegisterFunctionPointer(string functionName, MemberInfo method, Lis
             else if (CoreUtils.IsGetter(functionName))
             {
                 f = new GetterFunctionPointer(Module, functionName, method, retype);
-                (f as GetterFunctionPointer).ReflectionInfo.CheckForRankReductionAttribute(mGetterAttributes);
+                (f as GetterFunctionPointer).ReflectionInfo.CopyRankReductionAttribute(mGetterAttributes);
```
aparajit-pratap: Sorry, but I think the earlier name was more appropriate considering the context in which it is used.

test/Engine/FFITarget/TestObjects.cs
```diff
@@ -457,4 +457,35 @@ public void Dispose()
         }
     }
 
+    public class TestRankReduce : IDisposable
```
aparajit-pratap: You don't need to make this class an `IDisposable`.

test/Engine/FFITarget/TestObjects.cs
```diff
@@ -457,4 +457,35 @@ public void Dispose()
...
+    {
+        private string RankReduceTestField;
+
+        [AllowRankReduction]
+        public List<string> RankReduceArrowProperty => new List<string> { RankReduceTestField };
```
aparajit-pratap: How is this syntax different from the `RankReducePropertyGetter`?

aparajit-pratap:
@scottmitchell some minor comments and then LGTM!

mjkkirschner:
why are the builds failing here?

QilongTang:
@mjkkirschner The console shows all clear 18 smoke tests pass and all good, ever see this happens?

test/Engine/FFITarget/TestObjects.cs
aparajit-pratap: It's just syntactic sugar for the getter. Testing one of these would suffice.

mjkkirschner:
all checks clear 👍

test/Engine/FFITarget/TestObjects.cs
mjkkirschner: did you intend to dispose this class?

test/Engine/FFITarget/TestObjects.cs
scottmitchell: Fair enough. I just carried this over from the test cases I used during development. I removed the arrow property and added a method test (not sure if this exists anywhere else).

test/Engine/FFITarget/TestObjects.cs
scottmitchell: I only implemented `IDisposable` because the other test objects in this file did. I didn't really have any intentions for it.

src/Engine/ProtoCore/FFI/CLRDLLModule.cs
scottmitchell: I created a test zero touch class with 100 properties with attributes and the call to `CheckForRankReductionAttribute()` took <1ms.

scottmitchell:
@aparajit-pratap @mjkkirschner @QilongTang Thanks for your alls comments. I think I've addressed everything—Let me know if I haven't. Otherwise, I'll go ahead and merge once the checks pass.

mjkkirschner:
@scottmitchell please send a cherry pick if this is labeled 2.2

Issue #2695, opened 2014-10-13
Wish: Let captured image name default to graph name


andydandy74:
This is just a minor thing, but I constantly find myself renaming the screen captures shot with Dynamo. I guess "Capture.png" is okay, but I think it would make a lot more sense if the default name would be the same as that of the graph I am taking a screenshot of.

Racel:
I also think this is an easy minor fix that we can squeeze in to our sprints. Assigning to me and UX Love for now. Will try to address soon.

johnpierson:
Tracked internally as [QNTM-5920](https://jira.autodesk.com/browse/QNTM-5920)

PR #9546, opened 2019-03-06
DYN-1175: Captured image name default to "graph name+timestamp".

reddyashish:
### Purpose

This PR addresses a minor improvement when the user tries to capture an image of the workspace or the Background 3d image snapshot. 

A user can take a snapshot in the following ways:
1) File => Export Workspace As Image
2) File => Export Background 3D Preview As Image
3) Camera Icon (Top Right) while workspace is active
4) Camera Icon while background preview is active

The snapshot name will default to the graph name along with the timestamp. 

snapshot name

### Declarations

### Reviewers

@QilongTang @mjkkirschner

### DIFF
--- a/src/DynamoUtilities/PathHelper.cs
+++ b/src/DynamoUtilities/PathHelper.cs
@@ -165,7 +165,7 @@ namespace DynamoUtilities
         /// </summary>
         /// <param name="filePath">File path</param>
         /// <returns>Returns a default name(along with the timestamp) for the workspace image</returns>
-        public static String GetSnapshotNameFromPath(String filePath)
+        public static String GetScreenCaptureNameFromPath(String filePath)
         {
             FileInfo fileInfo = new FileInfo(filePath);
             String timeStamp = string.Format("{0:yyyy-MM-dd_hh-mm-ss}", DateTime.Now);
--- a/src/DynamoUtilities/PathHelper.cs
+++ b/src/DynamoUtilities/PathHelper.cs
@@ -163,10 +163,11 @@ namespace DynamoUtilities
         /// <summary>
         /// This is a utility method for generating a default name to the snapshot image. 
         /// </summary>
-        /// <param name="fileInfo">File information</param>
+        /// <param name="filePath">File path</param>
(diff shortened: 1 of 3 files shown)


src/DynamoUtilities/PathHelper.cs
```diff
@@ -159,5 +159,12 @@ public static bool IsFileNameInValid(string fileName)
 
             return false;
         }
+
+        public static String GetSnapshotName(FileInfo fi)
```
QilongTang: Would you add comments and description for new public Utility function?

QilongTang: I would also be careful about design consistence of `pathHelper`, as you can tell all the other utility function all take a path as input. You may want to do the same and provide two APIs for your use, e.g. `GetSnapshotNameFromPath` and `getDirectoryNameFromPath`

QilongTang:
One more comment and then LGTM

src/DynamoUtilities/PathHelper.cs
reddyashish: @QilongTang I changed the API to take the filePath as input parameter for the GetSnapshotNameFromPath function. I would not need GetDirectoryNameFromPath API, as we save the snapshot image to the same location as the graph location.

src/DynamoUtilities/PathHelper.cs
```diff
@@ -159,5 +159,18 @@ public static bool IsFileNameInValid(string fileName)
 
             return false;
         }
+
+        /// <summary>
+        /// This is a utility method for generating a default name to the snapshot image. 
+        /// </summary>
+        /// <param name="filePath">File path</param>
+        /// <returns>Returns a default name(along with the timestamp) for the workspace image</returns>
+        public static String GetSnapshotNameFromPath(String filePath)
```
aparajit-pratap: How about `GetScreenCaptureNameFromPath`?

QilongTang:
I looked at the self CI results, this PR should be safe to merge. We do need to fix the self CI ASAP

johnpierson:
fixes #2695

PR #9554, opened 2019-03-08
DYN-1688

reddyashish:
### Purpose

This PR is still Work in progress.

This addresses the issue that occurs when a List with nested empty and non-empty lists is passed to a function. It fails during replication process. When determining the method resolution, the length of the arrayStats list had to be checked and the reducedParams list has to be cleared only when the arrayStats has any elements. 
https://github.com/reddyashish/Dynamo/blob/aa841e4758a34a991214e9cf660e9fd002f7495b/src/Engine/ProtoCore/Lang/Replication/Replicator.cs#L243

Capture

### Declarations

Check these if you believe they are true

- [ ] The code base is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions), and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.

### Reviewers

@aparajit-pratap @mjkkirschner @saintentropy

PR #9559, opened 2019-03-11
DYN-1688: Fix regression with replication over list inputs with nested sublists

aparajit-pratap:
### Purpose

JIRA: https://jira.autodesk.com/browse/DYN-1688
This fixes a regression from 2.0 where a list containing an empty sublist as its first element followed by another list (of the same rank) did not replicate properly over a node. 

image

Performance results before and after the fix for the same long-running test:
image

### Declarations

- [x] All tests pass using the self-service CI. - No new failures apart from those already failing on SS-CI

### Reviewers

@reddyashish 
@mjkkirschner 

### FYIs

@QilongTang @saintentropy

### DIFF
--- a/src/Engine/ProtoCore/Utils/ArrayUtils.cs
+++ b/src/Engine/ProtoCore/Utils/ArrayUtils.cs
@@ -132,8 +132,12 @@ namespace ProtoCore.Utils
             var dsArray = runtimeCore.Heap.ToHeapObject<DSArray>(array);
             foreach (var sv in dsArray.Values)
             {
+                if(IsEmpty(sv, runtimeCore)) continue;
+
                 if (!usageFreq.ContainsKey(sv.metaData.type))
+                {
                     usageFreq.Add(sv.metaData.type, sv);
+                }
             }
 
             return usageFreq;
--- a/test/DynamoCoreTests/DSEvaluationModelTest.cs
+++ b/test/DynamoCoreTests/DSEvaluationModelTest.cs
@@ -1187,6 +1187,20 @@ namespace Dynamo.Tests
             AssertPreviewValue(guidY, new object[] { null, null, null, 10.2 });
             AssertPreviewValue(guidZ, new object[] { null, null, null, 15.2 });
         }
+
+        [Test, Category("UnitTests")]
+        public void ReplicationWithEmptySubLists()
+        {
+            RunModel(@"core\dsevaluation\Replication_EmptySublist.dyn");
+            var guidCurveLength = "1b247af2b1c046fb9f8e3e27761ab5a9";
+            var guidCodeBlock = Guid.Parse("b9dec880d99347eb8a203783f54763e6");
+            AssertPreviewValue(guidCurveLength, new object[] { new object[] { }, new object[] { 6.283185 } });
+
+            var command = new Models.DynamoModel.UpdateModelValueCommand(Guid.Empty, guidCodeBlock, "Code", @"[[c],[]]");
+            CurrentDynamoModel.ExecuteCommand(command);
+            RunCurrentModel();
+            AssertPreviewValue(guidCurveLength, new object[] { new object[] { 6.283185 }, new object[] { } });
+        }
     }
 
     [Category("DSCustomNode")]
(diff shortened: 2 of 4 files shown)


src/Engine/ProtoCore/Utils/ArrayUtils.cs
```diff
@@ -132,8 +132,12 @@ public static ClassNode GetGreatestCommonSubclassForArray(StackValue array, Runt
             var dsArray = runtimeCore.Heap.ToHeapObject<DSArray>(array);
             foreach (var sv in dsArray.Values)
             {
+                if(IsEmpty(sv, runtimeCore)) continue;
```
aparajit-pratap: We don't want to add a type for an empty list. On the other hand we want to ignore it and inspect the next element of the array.

QilongTang:
Can we conclude on difference between this one and https://github.com/DynamoDS/Dynamo/pull/9554? Are they both needed?

aparajit-pratap:
> Can we conclude on difference between this one and #9554? Are they both needed?

I'm not sure about #9554. I'm not convinced it is the right fix. Since the self-serve CI is broken, I ran the language tests locally on this PR and they all pass. I will do more testing on this PR including  manual testing.

reddyashish:
Added the tests for this. Once we verify on the self serve, we can merge this. I will be closing the other PR (https://github.com/DynamoDS/Dynamo/pull/9554) related to this same issue

test/DynamoCoreTests/DSEvaluationModelTest.cs
```diff
@@ -1187,6 +1187,20 @@ public void TestHeterogenousList()
             AssertPreviewValue(guidY, new object[] { null, null, null, 10.2 });
             AssertPreviewValue(guidZ, new object[] { null, null, null, 15.2 });
         }
+
+        [Test, Category("UnitTests")]
+        public void ReplicationWithEmptySubLists()
```
aparajit-pratap: @reddyashish could you also add a language test (using DS code) in ProtoTest.dll?

aparajit-pratap:
Updated performance benchmarking results.

mjkkirschner:
So I interpret this perf benchmark to mean that the results are essentially the same?

aparajit-pratap:
> So I interpret this perf benchmark to mean that the results are essentially the same?

Yep.

test/DynamoCoreTests/DSEvaluationModelTest.cs
reddyashish: Added this.

test/Engine/ProtoTest/Associative/MethodResolution.cs
```diff
@@ -396,5 +396,24 @@ def qux2()
...
+        public void TestEmptyNestedListsForMethodResolution()
+        {
+            string code = @"import(""FFITarget.dll""); 
+pt = DummyPoint2D.ByCoordinates(0,0);
+l = [[],[pt]];
+px = DummyPoint2D.X(l);
+";
+            var mirror = thisTest.RunScriptSource(code);
+            thisTest.Verify("px", new object[] { new object[] { }, new object[] { 0 } });
+            code = @"import(""FFITarget.dll"");
```
aparajit-pratap: I don't think you need to add the import statement the second time once it's already imported.

test/Engine/ProtoTest/Associative/MethodResolution.cs
```diff
@@ -396,5 +396,24 @@ def qux2()
...
+l = [[],[pt]];
+px = DummyPoint2D.X(l);
+";
+            var mirror = thisTest.RunScriptSource(code);
+            thisTest.Verify("px", new object[] { new object[] { }, new object[] { 0 } });
+            code = @"import(""FFITarget.dll""); 
+pt = DummyPoint2D.ByCoordinates(0,0);
+l = [[pt],[]];
+px = DummyPoint2D.X(l);
+";
```
aparajit-pratap: rewrite your test like so:
```
string code = @"import(""FFITarget.dll""); 
pt = DummyPoint2D.ByCoordinates(0,0);
l = [[],[pt]];
px1 = DummyPoint2D.X(l);
pt = DummyPoint2D.ByCoordinates(0,0);
l = [[pt],[]];
px2 = DummyPoint2D.X(l);
";
```
Assert for `px1` and `px2`.

PR #7850, opened 2017-05-05
Duplicate connectors on ctrl + click on input ports

yeexinc:
### Purpose

Duplicate connectors when the user does a ctrl + click on input ports.
duplicatectrl

### Declarations

Check these if you believe they are true

- [ ] The code base is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions), and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.

### Reviewers
@Benglin 
@riteshchandawar

PR #9567, opened 2019-03-13
Duplicate connectors using `ctrl + left click` on input ports with testing

alfarok:
### Purpose

ctrl_click

This PR pulls in the work originally completed by @yeexinc in #7850 and adds several tests that leverage recorded commands.  The original work was already reviewed and all active comments were addressed.  The recorded tests verify the proper connections are made when using `ctrl + left click` in order to duplicate a connector that is already currently connected to another nodes input port.  Additionally undo/redo was verified using recorded commands but copy/paste was manually as verified as they are not currently supported as recorded commands.

### Declarations

### Reviewers

### FYIs

@Amoursol

### DIFF
--- a/src/DynamoCore/Models/DynamoModelCommands.cs
+++ b/src/DynamoCore/Models/DynamoModelCommands.cs
@@ -336,7 +336,7 @@ namespace Dynamo.Models
                 return;
             }
 
-            PortModel portModel = node.InPorts[portIndex];
+            var portModel = node.InPorts[portIndex];
             if (portModel.Connectors.Count == 0)
             {
                 // If the port doesn't have any existing connections, begin connection as per normal
--- a/src/DynamoCore/Models/DynamoModelCommands.cs
+++ b/src/DynamoCore/Models/DynamoModelCommands.cs
@@ -270,8 +270,8 @@ namespace Dynamo.Models
                 case MakeConnectionCommand.Mode.EndShiftReconnections:
                     EndShiftReconnections(nodeId, command.PortIndex, command.Type);
                     break;
-                
-                // TODO - can be removed in Dynamo 3.0
+
+                // TODO - can be removed in Dynamo 3.0 - DYN-1729
                 case MakeConnectionCommand.Mode.EndAndStartCtrlConnection:
                     BeginDuplicateConnection(nodeId, command.PortIndex, command.Type);
                     break;
--- a/src/DynamoCore/Models/DynamoModelCommands.cs
+++ b/src/DynamoCore/Models/DynamoModelCommands.cs
@@ -271,8 +271,12 @@ namespace Dynamo.Models
                     EndShiftReconnections(nodeId, command.PortIndex, command.Type);
                     break;
 
-                case MakeConnectionCommand.Mode.EndAndStartCtrlConnection:
-                    EndAndStartCtrlConnection(nodeId, command.PortIndex, command.Type);
+                case MakeConnectionCommand.Mode.BeginDuplicateConnection:
+                    BeginDuplicateConnection(nodeId, command.PortIndex, command.Type);
+                    break;
+
+                case MakeConnectionCommand.Mode.BeginCreateConnections:
+                    BeginCreateConnections(nodeId, command.PortIndex, command.Type);
                     break;
 
                 case MakeConnectionCommand.Mode.Cancel:
@@ -313,6 +317,35 @@ namespace Dynamo.Models
                 activeStartPorts = new PortModel[] { portModel };
                 firstStartPort = isInPort ? null : portModel; // Only assign firstStartPort if the port selected is an output port
             }
+        }
+
+        void BeginDuplicateConnection(Guid nodeId, int portIndex, PortType portType)
+        {
+            // If the port clicked is an output port, begin connection as per normal
+            if (portType == PortType.Output)
+            {
+                BeginConnection(nodeId, portIndex, portType);
+                return;
+            }
+
+            // Otherwise, if the port is an input port, check if the port already has a connection.
+            // If it does, duplicate the existing connection.
+            var node = CurrentWorkspace.GetModelInternal(nodeId) as NodeModel;
+            if (node == null)
+            {
+                return;
+            }
+
+            PortModel portModel = node.InPorts[portIndex];
+            if (portModel.Connectors.Count == 0)
+            {
+                // If the port doesn't have any existing connections, begin connection as per normal
+                BeginConnection(nodeId, portIndex, portType);
+                return;
+            }
+            // Duplicate the existing connection
+            activeStartPorts = new PortModel[] { portModel.Connectors[0].Start };
+            firstStartPort = portModel.Connectors[0].Start;
         }
 
         void BeginShiftReconnections(Guid nodeId, int portIndex, PortType portType)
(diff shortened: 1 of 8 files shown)


src/DynamoCore/Models/DynamoModelCommands.cs
```diff
@@ -315,6 +319,35 @@ void BeginConnection(Guid nodeId, int portIndex, PortType portType)
             }
         }
 
+        void BeginDuplicateConnection(Guid nodeId, int portIndex, PortType portType)
```
mjkkirschner: please mark this with an access modifier.

src/DynamoCore/Models/DynamoModelCommands.cs
```diff
@@ -315,6 +319,35 @@ void BeginConnection(Guid nodeId, int portIndex, PortType portType)
...
+
+            // Otherwise, if the port is an input port, check if the port already has a connection.
+            // If it does, duplicate the existing connection.
+            var node = CurrentWorkspace.GetModelInternal(nodeId) as NodeModel;
+            if (node == null)
+            {
+                return;
+            }
+
+            PortModel portModel = node.InPorts[portIndex];
```
mjkkirschner: var

src/DynamoCore/Models/DynamoModelCommands.cs
```diff
@@ -388,7 +421,7 @@ void EndShiftReconnections(Guid nodeId, int portIndex, PortType portType)
             return;
         }
 
-        void EndAndStartCtrlConnection(Guid nodeId, int portIndex, PortType portType)
+        void BeginCreateConnections(Guid nodeId, int portIndex, PortType portType)
```
mjkkirschner: access modifier

src/DynamoCore/Models/RecordableCommands.cs
```diff
@@ -1358,9 +1358,13 @@ public enum Mode
                 /// </summary>
                 EndShiftReconnections,
                 /// <summary>
-                /// End and start control connections.
+                /// Begin duplicate connection.
                 /// </summary>
-                EndAndStartCtrlConnection,
+                BeginDuplicateConnection,
```
mjkkirschner: oye... public breaking change. can you keep the old enum, obsolete it, and add the new ones at the end without changing the order?

test/DynamoCoreWpfTests/RecordedTests.cs
```diff
@@ -1155,6 +1155,62 @@ public void TestReconnectionUndo()
             Assert.AreEqual(2, workspace.Connectors.Count()); 
         }
 
+        /// <summary>
+        /// The following recorded test manually adds a connection between 2 nodes,
+        /// followed by ctrl + clicking to copy this connection to 2 additional nodes.
+        /// The final assertion will only be true if all connections were successfully established.
+        /// </summary>
+        [Test, RequiresSTA]
```
mjkkirschner: just curious, what happens if you remove the requiresSTA?

mjkkirschner:
@alfarok 
* there is a public API change here - please address that.
* please add access modifiers - and use var where possible. (makes refactoring easier)
* can we also add unit tests that run without the UI?

src/DynamoCore/Models/DynamoModelCommands.cs
alfarok: interestingly almost none of the methods `DynamoModelCommands.cs` have access modifiers, I guess because they are all expected to be `internal`?

src/DynamoCore/Models/DynamoModelCommands.cs
alfarok: Is this a required standard?  I don't understand why it's bad practice to be explicit when you know the type for readability.

mjkkirschner: this is a good guide:
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions#implicitly-typed-local-variables

I am pretty confident this type is a PortModel because I know the Inports type is always going to be a portModel list.

test/DynamoCoreWpfTests/RecordedTests.cs
alfarok: doesn't seem to matter but is used in all the previous implementations, I can remove from the tests I added

src/DynamoCore/Models/DynamoModelCommands.cs
QilongTang: `internal` seems reasonable

mjkkirschner: I think if not specific the default is actually private.

test/DynamoCoreWpfTests/RecordedTests.cs
```diff
@@ -1155,6 +1155,62 @@ public void TestReconnectionUndo()
...
+            Assert.AreEqual(7, workspace.Connectors.Count()); // 2 of which should be executed in the recording using ctrl + click
+
+            // Sum of values which is only true if connections are valid downstream
+            AssertPreviewValue("68d76b724b75499b8e96cfc106118107", 12);
+        }
+
+        /// <summary>
+        /// The following recorded test manually adds a connecting between 2 nodes,
+        /// followed by ctrl + clicking to copy this connection to 2 additional nodes.
+        /// We then call 2 undo commands and verify only our original connection exists.
```
QilongTang: Nice, @alfarok would you point below where in the recorded tests we do `ctrl`? Just curious

src/DynamoCore/Models/DynamoModelCommands.cs
QilongTang: I differs if it is interface or class or methods, see https://stackoverflow.com/questions/2521459/what-are-the-default-access-modifiers-in-c

test/DynamoCoreWpfTests/RecordedTests.cs
alfarok: @QilongTang I think in the XML it just shows up as `MakeConnectionCommand` because `BeginDuplicateConnection` is a subset `MakeConnectionCommand`.  If you play back the recording or run the test you can see the way the connections are created is the same as in the .gif in the description though.

src/DynamoCore/Models/RecordableCommands.cs
alfarok: @mjkkirschner Can I reuse the same name for the new enum?  Or do you mean use the `[Obsolete]` tag on the actual enum items?  I don't think the first is possible

test/DynamoCoreWpfTests/RecordedTests.cs
QilongTang: @alfarok OK so we do not expose the level of details of key pressed but only commands in recorded tests, that is good to know

alfarok:
@mjkkirschner @QilongTang This should be ready for a final look whenever you guys get a free moment

src/DynamoCore/Models/DynamoModelCommands.cs
```diff
@@ -52,19 +52,19 @@ protected virtual void OpenFileImpl(OpenFileCommand command)
             //ClipBoard.Clear();
         }
 
-        void RunCancelImpl(RunCancelCommand command)
+        internal void RunCancelImpl(RunCancelCommand command)
```
mjkkirschner: @alfarok - did you checkout the link @QilongTang sent - I interpret this to mean these should have been private by default?

mjkkirschner:
@alfarok looks good except for the access modifier - if it was private before (by default) lets not change it unless theres a reason to.

QilongTang:
Nicely wrapped up, LGTM

PR #9568, opened 2019-03-13
Rc2.0.3 lib crash

mjkkirschner:
@reddyashish does this look correct?

### DIFF
--- a/src/LibraryViewExtension/Handlers/NodeItemDataProvider.cs
+++ b/src/LibraryViewExtension/Handlers/NodeItemDataProvider.cs
@@ -74,9 +74,7 @@ namespace Dynamo.LibraryUI.Handlers
         protected virtual object CreateObjectForSerialization(IEnumerable<NodeSearchElement> searchEntries)
         {
             var data = new LoadedTypeData<LoadedTypeItem>();
-            data.loadedTypes = searchEntries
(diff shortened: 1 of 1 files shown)


src/LibraryViewExtension/Handlers/NodeItemDataProvider.cs
```diff
@@ -74,9 +74,9 @@ protected Stream GetNodeItemDataStream(IEnumerable<NodeSearchElement> searchEntr
         protected virtual object CreateObjectForSerialization(IEnumerable<NodeSearchElement> searchEntries)
         {
             var data = new LoadedTypeData<LoadedTypeItem>();
-            data.loadedTypes = searchEntries
-                //.Where(e => !e.ElementType.HasFlag(ElementTypes.Packaged))
```
mjkkirschner: @aparajit-pratap @reddyashish @ColinDayOrg - it just occurs to me that this commented out code is kind of peculiar, does anyone know why this was left in?

reddyashish: Sorry was waiting for others to see if they knew about this. I am not sure why that line was commented in the first place. I assumed we are not using that attribute filter anymore.  

The changes look good.

PR #9569, opened 2019-03-13
Move performance benchmarking tool into its own solution

aparajit-pratap:
### Purpose

JIRA: https://jira.autodesk.com/browse/DYN-1713
 
This PR is to fix the builds and self-serve CI by moving the performance benchmarking project outside of `Dynamo.All.sln` and into its own solution in `Dynamo\tools\Performance\DynamoPerformanceTests`. It currently has a hard dependency on the test DLL's built in the `Dynamo\bin` directory but there'll be a follow up PR to improve on that soon. 

The `xcopy` prebuild event also continues to be in place, which copies all core runtime dependencies from `Dynamo\bin` into the project's local `bin` folder.

The `BenchmarkDotNet` package has been updated to the latest available on NuGet - 0.11.4, which has a dependency on .NET standard.

### Declarations

Self-service CI builds now pass:
image

### Reviewers

@mjkkirschner 

### FYIs

@QilongTang


tools/Performance/DynamoPerformanceTests/DynamoPerformanceTests.csproj
```diff
@@ -40,122 +40,155 @@
...
     <Reference Include="System.IO.FileSystem, Version=4.0.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
-      <HintPath>..\..\..\src\packages\System.IO.FileSystem.4.3.0\lib\net46\System.IO.FileSystem.dll</HintPath>
+      <HintPath>packages\System.IO.FileSystem.4.3.0\lib\net46\System.IO.FileSystem.dll</HintPath>
       <Private>True</Private>
       <Private>True</Private>
     </Reference>
     <Reference Include="System.IO.FileSystem.Primitives, Version=4.0.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
-      <HintPath>..\..\..\src\packages\System.IO.FileSystem.Primitives.4.3.0\lib\net46\System.IO.FileSystem.Primitives.dll</HintPath>
+      <HintPath>packages\System.IO.FileSystem.Primitives.4.3.0\lib\net46\System.IO.FileSystem.Primitives.dll</HintPath>
+      <Private>True</Private>
```
mjkkirschner: this looks weird - there should not be two of these attributes

tools/Performance/DynamoPerformanceTests/DynamoPerformanceTests.csproj
```diff
@@ -40,122 +40,155 @@
...
-      <HintPath>..\..\..\src\packages\BenchmarkDotNet.0.11.3\lib\net46\BenchmarkDotNet.dll</HintPath>
+    <Reference Include="BenchmarkDotNet, Version=0.11.4.0, Culture=neutral, PublicKeyToken=aa0ca2f9092cefc4, processorArchitecture=MSIL">
+      <HintPath>packages\BenchmarkDotNet.0.11.4\lib\netstandard2.0\BenchmarkDotNet.dll</HintPath>
     </Reference>
     <Reference Include="CommandLine, Version=2.4.3.0, Culture=neutral, PublicKeyToken=de6f01bd326f8c32, processorArchitecture=MSIL">
-      <HintPath>..\..\..\src\packages\CommandLineParser.2.4.3\lib\netstandard2.0\CommandLine.dll</HintPath>
+      <HintPath>packages\CommandLineParser.2.4.3\lib\netstandard2.0\CommandLine.dll</HintPath>
+    </Reference>
+    <Reference Include="DynamoCoreTests">
+      <HintPath>..\..\..\bin\AnyCPU\Release\DynamoCoreTests.dll</HintPath>
```
mjkkirschner: can this be parameterized for the config or we should only use release anyway?

tools/Performance/DynamoPerformanceTests/DynamoPerformanceTests.csproj
```diff
@@ -164,50 +197,40 @@
...
-      <Project>{472084ed-1067-4b2c-8737-3839a6143eb2}</Project>
-      <Name>DynamoCoreTests</Name>
-    </ProjectReference>
-  </ItemGroup>
   <ItemGroup>
     <None Include="app.config" />
     <None Include="packages.config" />
   </ItemGroup>
-  <ItemGroup>
-    <Analyzer Include="..\..\..\src\packages\Microsoft.CodeAnalysis.Analyzers.2.6.1\analyzers\dotnet\cs\Microsoft.CodeAnalysis.Analyzers.dll" />
```
mjkkirschner: interesting, now these references are totally gone from the csproj file

tools/Performance/DynamoPerformanceTests/DynamoPerformanceTests.csproj
mjkkirschner: @aparajit-pratap what about using https://www.nuget.org/packages/DynamoVisualProgramming.Tests/

mjkkirschner: maybe thats weird... it makes it harder to make changes to the system - devs will need to publish to a local nuget repo... curious what you think? Are the apis we need from this dll stable?

mjkkirschner: I imagine that soon, we might need more APIs from the testing dll or core itself, for monitoring render time etc.

mjkkirschner:
@aparajit-pratap this looks good to me - a couple questions but let's prioritize getting the builds going and we can address after we know we're stable?

tools/Performance/DynamoPerformanceTests/DynamoPerformanceTests.csproj
aparajit-pratap: It's preferable to use only Release but I'll parameterize it anyway.

tools/Performance/DynamoPerformanceTests/DynamoPerformanceTests.csproj
aparajit-pratap: Oops, I manually deleted these references but it managed to build even without them. However since the NuGet restore packages added these dependencies to the packages folder, I might as well add these references back in.

tools/Performance/DynamoPerformanceTests/DynamoPerformanceTests.csproj
aparajit-pratap: Which 2 attributes? One is the reference and the other is the path.

aparajit-pratap: One is `System.IO.FileSystem` and the other is `System.IO.FileSystem.Primitives`

mjkkirschner: <Private>True</Private>
<Private>True</Private>

is there twice - am I reading this wrong?

PR #9574, opened 2019-03-14
fix flaky timing test

aparajit-pratap:
### Purpose

This attempts to fix a flaky test that times the performance: `TestMethodResolutionPerformance` in `Dynamo\test\DynamoCoreTests\HomgeneousListTests.cs`. There was a build failure due to this: https://master-15.jenkins.autodesk.com/job/DYN-DynamoCore_master/1868/consoleFull

### Declarations

### Reviewers

@saintentropy 

### FYIs

@mjkkirschner 
@QilongTang


QilongTang:
@aparajit-pratap If I understand correctly, removing multiplier means we are removing more results on the upper/lower edges thus time will sure be more stable. Do you think it is still useful this way or it is the nature of running such tests in VM?

aparajit-pratap:
I just discussed this with @saintentropy. I'm going to remove this test and add the graphs used in the test to the performance test suite for now. As discussed with @mjkkirschner we can implement a more sophisticated system later that does comparisons of running the test suite against a baseline and report them automatically with each PR.

PR #9578, opened 2019-03-15
DYN-1722

reddyashish:
### Purpose

This PR is to address a regression issue from 2.0. (https://jira.autodesk.com/browse/DYN-1722)

This also fixes most of the non-regressions from https://jira.autodesk.com/browse/DYN-1496. The only case that is failing would be the second case in the description of the above JIRA. 

Explanation:
Instead of the just returning the first replication option that matches the target function, we check for all the replications that match the target function and then choose the replication option with largest rank. In the case where both cartesian and zipped replication options are possible, we check for the first option that matches the target function and modify the replication option(in next iterations) only if 
a similar replication option, with a higher rank,  is matched with the target function. 

Replication Fix

Performance test results on around 11 graphs:

Before the fix:
before

After the fix:
after

### Declarations

### Reviewers

### DIFF
--- a/src/Engine/ProtoCore/Lang/CallSite.cs
+++ b/src/Engine/ProtoCore/Lang/CallSite.cs
@@ -826,7 +826,7 @@ namespace ProtoCore
         {
             if (oldOption.Count > 0 && newOption.Count > 0 && oldOption.Count < newOption.Count)
             {
-                if (oldOption[0].ToString().Equals(newOption[0].ToString()))
+                if (oldOption[0].Equals(newOption[0]))
                     return true;
             }
             return false;
--- a/src/Engine/ProtoCore/Lang/CallSite.cs
+++ b/src/Engine/ProtoCore/Lang/CallSite.cs
@@ -844,7 +844,7 @@ namespace ProtoCore
         {
             replicationInstructions = null;
             resolvesFeps = null;
-            Boolean flag = false;
+            Boolean matchFound = false;
 
             #region Case 1: Replication guide with exact match 
             {
@@ -874,11 +874,11 @@ namespace ProtoCore
                             //Otherwise we have a cluster of FEPs that can be used to dispatch the array
                             resolvesFeps = new List<FunctionEndPoint>(lookups);
                             replicationInstructions = replicationOption;
-                            flag = true;
+                            matchFound = true;
                         }
                     }
                 }
-                if (flag == true)
+                if (matchFound == true)
                     return;
             }
             #endregion
@@ -906,10 +906,10 @@ namespace ProtoCore
                         {
                             resolvesFeps = new List<FunctionEndPoint>() { compliantTarget };
                             replicationInstructions = replicationOption;
-                            flag = true;
+                            matchFound = true;
                         }
                     }
-                    if (flag == true)
+                    if (matchFound == true)
                         return;
                 }
             }
@@ -943,10 +943,10 @@ namespace ProtoCore
                     {
                         resolvesFeps = new List<FunctionEndPoint>() { compliantTarget };
                         replicationInstructions = replicationOption;
-                        flag = true;
+                        matchFound = true;
                     }
                 }
-                if (flag == true)
+                if (matchFound == true)
                     return;
             }
             #endregion
--- a/src/Engine/ProtoCore/Lang/CallSite.cs
+++ b/src/Engine/ProtoCore/Lang/CallSite.cs
@@ -822,7 +822,15 @@ namespace ProtoCore
             return compliantTarget;
         }
 
-        private Boolean IsCompatibleReplicationOption(List<ReplicationInstruction> oldOption, List<ReplicationInstruction> newOption)
+
+        /// <summary>
+        /// This helper function checks if the current replication option is
+        /// similar to the previous option but of a higher rank. 
+        /// Checks if the first entry is same in both the options and the current options count is more. 
+        /// </summary>
+        /// <returns>Returns true or false based on the condition described above. 
+        /// </returns>
+        private Boolean IsSimilarOptionButOfHigherRank(List<ReplicationInstruction> oldOption, List<ReplicationInstruction> newOption)
         {
             if (oldOption.Count > 0 && newOption.Count > 0 && oldOption.Count < newOption.Count)
             {
@@ -869,7 +877,7 @@ namespace ProtoCore
                     HashSet<FunctionEndPoint> lookups;
                     if (funcGroup.CanGetExactMatchStatics(context, reducedParams, stackFrame, runtimeCore, out lookups))
                     {
-                        if (replicationInstructions == null || IsCompatibleReplicationOption(replicationInstructions, replicationOption))
+                        if (replicationInstructions == null || IsSimilarOptionButOfHigherRank(replicationInstructions, replicationOption))
                         {
                             //Otherwise we have a cluster of FEPs that can be used to dispatch the array
                             resolvesFeps = new List<FunctionEndPoint>(lookups);
--- a/src/Engine/ProtoCore/Lang/CallSite.cs
+++ b/src/Engine/ProtoCore/Lang/CallSite.cs
@@ -844,7 +844,7 @@ namespace ProtoCore
         {
             replicationInstructions = null;
             resolvesFeps = null;
-            Boolean matchFound = false;
+            var matchFound = false;
 
             #region Case 1: Replication guide with exact match 
             {
@@ -878,7 +878,7 @@ namespace ProtoCore
                         }
                     }
                 }
-                if (matchFound == true)
+                if (matchFound)
                     return;
             }
             #endregion
@@ -909,7 +909,7 @@ namespace ProtoCore
                             matchFound = true;
                         }
                     }
-                    if (matchFound == true)
+                    if (matchFound)
                         return;
                 }
             }
@@ -946,7 +946,7 @@ namespace ProtoCore
                         matchFound = true;
                     }
                 }
-                if (matchFound == true)
+                if (matchFound)
                     return;
             }
             #endregion
--- a/src/Engine/ProtoCore/Lang/CallSite.cs
+++ b/src/Engine/ProtoCore/Lang/CallSite.cs
@@ -822,6 +822,16 @@ namespace ProtoCore
             return compliantTarget;
         }
 
+        private Boolean IsCompatibleReplicationOption(List<ReplicationInstruction> oldOption, List<ReplicationInstruction> newOption)
+        {
+            if (oldOption.Count > 0 && newOption.Count > 0 && oldOption.Count < newOption.Count)
+            {
+                if (oldOption[0].ToString().Equals(newOption[0].ToString()))
+                    return true;
+            }
+            return false;
+        }
+
         private void ComputeFeps(
             Context context,
             List<StackValue> arguments,
@@ -832,6 +842,10 @@ namespace ProtoCore
             out List<FunctionEndPoint> resolvesFeps,
             out List<ReplicationInstruction> replicationInstructions)
         {
+            replicationInstructions = null;
+            resolvesFeps = null;
+            Boolean flag = false;
+
             #region Case 1: Replication guide with exact match 
             {
                 FunctionEndPoint fep = GetCompleteMatchFunctionEndPoint(context, arguments, funcGroup, instructions, stackFrame, runtimeCore);
@@ -855,13 +869,17 @@ namespace ProtoCore
                     HashSet<FunctionEndPoint> lookups;
                     if (funcGroup.CanGetExactMatchStatics(context, reducedParams, stackFrame, runtimeCore, out lookups))
                     {
-                        //Otherwise we have a cluster of FEPs that can be used to dispatch the array
-                        resolvesFeps = new List<FunctionEndPoint>(lookups);
-                        replicationInstructions = replicationOption;
-                        return;
+                        if (replicationInstructions == null || IsCompatibleReplicationOption(replicationInstructions, replicationOption))
+                        {
+                            //Otherwise we have a cluster of FEPs that can be used to dispatch the array
+                            resolvesFeps = new List<FunctionEndPoint>(lookups);
+                            replicationInstructions = replicationOption;
+                            flag = true;
+                        }
                     }
                 }
-
+                if (flag == true)
+                    return;
             }
             #endregion
 
@@ -888,9 +906,11 @@ namespace ProtoCore
                         {
                             resolvesFeps = new List<FunctionEndPoint>() { compliantTarget };
                             replicationInstructions = replicationOption;
-                            return;
+                            flag = true;
                         }
                     }
+                    if (flag == true)
+                        return;
                 }
             }
 
@@ -923,9 +943,11 @@ namespace ProtoCore
                     {
                         resolvesFeps = new List<FunctionEndPoint>() { compliantTarget };
                         replicationInstructions = replicationOption;
-                        return;
+                        flag = true;
                     }
                 }
+                if (flag == true)
+                    return;
             }
             #endregion
 
--- a/src/Engine/ProtoCore/Lang/CallSite.cs
+++ b/src/Engine/ProtoCore/Lang/CallSite.cs
@@ -847,11 +847,11 @@ namespace ProtoCore
             List<ReplicationInstruction> instructions,
             StackFrame stackFrame,
             RuntimeCore runtimeCore,
-            out List<FunctionEndPoint> resolvesFeps,
+            out List<FunctionEndPoint> resolvedFeps,
             out List<ReplicationInstruction> replicationInstructions)
         {
             replicationInstructions = null;
-            resolvesFeps = null;
+            resolvedFeps = null;
             var matchFound = false;
 
             #region Case 1: Replication guide with exact match 
@@ -859,7 +859,7 @@ namespace ProtoCore
                 FunctionEndPoint fep = GetCompleteMatchFunctionEndPoint(context, arguments, funcGroup, instructions, stackFrame, runtimeCore);
                 if (fep != null)
                 {
-                    resolvesFeps = new List<FunctionEndPoint>() { fep };
+                    resolvedFeps = new List<FunctionEndPoint>() { fep };
                     replicationInstructions = instructions;
                     return;
                 }
@@ -880,7 +880,7 @@ namespace ProtoCore
                         if (replicationInstructions == null || IsSimilarOptionButOfHigherRank(replicationInstructions, replicationOption))
                         {
                             //Otherwise we have a cluster of FEPs that can be used to dispatch the array
-                            resolvesFeps = new List<FunctionEndPoint>(lookups);
+                            resolvedFeps = new List<FunctionEndPoint>(lookups);
                             replicationInstructions = replicationOption;
                             matchFound = true;
                         }
@@ -896,7 +896,7 @@ namespace ProtoCore
                 FunctionEndPoint compliantTarget = GetCompliantFEP(context, arguments, funcGroup, instructions, stackFrame, runtimeCore);
                 if (compliantTarget != null)
                 {
-                    resolvesFeps = new List<FunctionEndPoint>() { compliantTarget };
+                    resolvedFeps = new List<FunctionEndPoint>() { compliantTarget };
                     replicationInstructions = instructions;
                     return;
                 }
@@ -912,7 +912,7 @@ namespace ProtoCore
                         FunctionEndPoint compliantTarget = GetCompliantFEP(context, arguments, funcGroup, replicationOption, stackFrame, runtimeCore);
                         if (compliantTarget != null)
                         {
-                            resolvesFeps = new List<FunctionEndPoint>() { compliantTarget };
+                            resolvedFeps = new List<FunctionEndPoint>() { compliantTarget };
                             replicationInstructions = replicationOption;
                             matchFound = true;
                         }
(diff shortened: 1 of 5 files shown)


test/Engine/ProtoTest/Associative/MethodResolution.cs
```diff
@@ -396,21 +396,5 @@ def qux2()
             thisTest.Verify("z1", 22);
             thisTest.Verify("z2", 22);
         }
-
-        [Test]
-        public void TestEmptyNestedListsForMethodResolution()
```
reddyashish: Moved this test over to the Replication.cs file.

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -822,6 +822,16 @@ private void UpdateCallsiteExecutionState(Object callsiteData, RuntimeCore runti
             return compliantTarget;
         }
 
+        private Boolean IsCompatibleReplicationOption(List<ReplicationInstruction> oldOption, List<ReplicationInstruction> newOption)
+        {
+            if (oldOption.Count > 0 && newOption.Count > 0 && oldOption.Count < newOption.Count)
+            {
+                if (oldOption[0].ToString().Equals(newOption[0].ToString()))
```
mjkkirschner: what does toString() evaluate to here? What does a repliationInstruction look like when stringified?

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -832,6 +842,10 @@ private void UpdateCallsiteExecutionState(Object callsiteData, RuntimeCore runti
             out List<FunctionEndPoint> resolvesFeps,
             out List<ReplicationInstruction> replicationInstructions)
         {
+            replicationInstructions = null;
+            resolvesFeps = null;
+            Boolean flag = false;
```
mjkkirschner: what does the flag indicate? Can we name it based on that?

mjkkirschner:
@reddyashish have you had a chance to try running this change over a set of graphs before and after using the perf console app?

src/Engine/ProtoCore/Lang/CallSite.cs
reddyashish: ToString() returns the replicaion-option name and the number of indices like "Cartesian: indices:0" or "zipped:indices0,1".

mjkkirschner: do this get called repeatedly in a tight loop? I would rather we avoid a string conversion if this is performance critical code - can we just override equals and compare the two properties?

reddyashish:
Not yet, discussed this with Aparajit and we will be doing it. We would be needing a set of large graphs that would be testing cases involving function match resolution. Any ideas on if we got any?

src/Engine/ProtoCore/Lang/CallSite.cs
reddyashish: Yes that would be better. I will name it something like "MatchFound"

mjkkirschner:
I am sure if we can be specific about what graph types we want @jacobsmall and @Amoursol can provide a bunch.

src/Engine/ProtoCore/Lang/CallSite.cs
reddyashish: This wouldn't be called for every replication option. It will be called only when a replicationOption is already set in the previous trial and if it finds another match for the later replication trials. 
I can create a helper to compare those 2 properties but wouldn't be a lot different to the toStirng() method for this class. Let me know what you think.

mjkkirschner: avoiding the toString is only important I think if this is called frequently 😄 - I don't know enough about this pathway to have a gut feeling about it - leave that up to you and some profiling / perf benchmarking.

reddyashish: Yes. Will do good amount of performance testing before proceeding.

JacobSmall:
If I don’t have one ready I can likely build one (or ten) up pretty quickly. Looking for a lot of nodes or a lot of objects?

aparajit-pratap:
@reddyashish please enumerate and explain the case(s) that are yet to be fixed. You mentioned that the second example in DYN-1496 isn't fixed but you can generalize it a bit more; for example: `A heterogeneous list of sublists, where the first sublist type does not match the target function, does not replicate properly`. Example:
image

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -822,6 +822,16 @@ private void UpdateCallsiteExecutionState(Object callsiteData, RuntimeCore runti
             return compliantTarget;
         }
 
+        private Boolean IsCompatibleReplicationOption(List<ReplicationInstruction> oldOption, List<ReplicationInstruction> newOption)
+        {
+            if (oldOption.Count > 0 && newOption.Count > 0 && oldOption.Count < newOption.Count)
+            {
```
aparajit-pratap: I’m not sure I understand this change. Why are you checking equality only for the first option?

reddyashish: For those 2 cases where it was failing, as we are checking for all the replication options(to find a match), the first match is found for "cartesian:indices=0" option. Then it finds a match for "zipped:indices=0,1" and it applies zipped replication option(as this is the final option that is tested). Previously, the first option that found a match was applied directly. 
Another case is, we want to accept this new replication option, if it is of the higher rank than the previous option (new option: "cartesian:indices=0" and previous option: "cartesian:indices=0, cartesian:indices=0". The same option but checks for one extra depth level). Since all options are unique, I was checking the first element and the count for the new option to be higher than the first. Also can use, newOption.Count = oldOption.Count +1.

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -832,6 +842,10 @@ private void UpdateCallsiteExecutionState(Object callsiteData, RuntimeCore runti
             out List<FunctionEndPoint> resolvesFeps,
             out List<ReplicationInstruction> replicationInstructions)
         {
+            replicationInstructions = null;
+            resolvesFeps = null;
+            Boolean matchFound = false;
```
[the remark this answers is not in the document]
reddyashish: Yes will modify this.

src/Engine/ProtoCore/Lang/CallSite.cs
mjkkirschner: I'm definitely having a hard time understanding this one, is it possible @reddyashish you could draw a diagram or do a longer writeup of the problem here and the approach taken to fix it - I know it's kind of a tall order.

src/Engine/ProtoCore/Lang/CallSite.cs
mjkkirschner: this seems potentially scary - unless I am interpreting this wrong - 
https://github.com/DynamoDS/Dynamo/blob/ec10f936824152e7dd7d6d019efdcda0d78a5264/src/Engine/ProtoCore/Lang/Replication/ReplicationInstruction.cs#L36

it looks like if the zippedIndicies property were 1,000,000 items long, the toStrong() would be very slow.

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -888,9 +906,11 @@ private void UpdateCallsiteExecutionState(Object callsiteData, RuntimeCore runti
                         {
                             resolvesFeps = new List<FunctionEndPoint>() { compliantTarget };
                             replicationInstructions = replicationOption;
-                            return;
+                            matchFound = true;
```
mjkkirschner: @reddyashish so not returning here seems to fix this case:

this branch on the left |  1.3.4 on the right
Screen Shot 2019-03-21 at 9 33 57 PM

I guess because we walk further down through all cases of the possible replications .... does it make sense to reverse the list of replicationsTrials if we will always keep iterating on the list of `replicationInstruction(s)`?

also - I find it odd the performance benchmarks are showing the same speed.... let's try to manufacture a case which we know will be slower - for example:

Screen Shot 2019-03-21 at 9 50 09 PM

produces 25 `replicationInstruction`s to try - before your change this would have exited much earlier - it would be produce all nulls of course... but I want to see if we can get a sense of the difference in speed or bound it.

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -888,9 +906,11 @@ private void UpdateCallsiteExecutionState(Object callsiteData, RuntimeCore runti
...
                         }
                     }
+                    if (matchFound)
+                        return;
```
mjkkirschner: @reddyashish so - what about case 5 - why do we exit the other cases late, and exit case 5 as soon as we find a match?

mjkkirschner:
I think these cases really need examples.

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -832,6 +842,10 @@ private void UpdateCallsiteExecutionState(Object callsiteData, RuntimeCore runti
             out List<FunctionEndPoint> resolvesFeps,
```
mjkkirschner: should this be `resolved`? (spelling?)

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -855,13 +869,17 @@ private void UpdateCallsiteExecutionState(Object callsiteData, RuntimeCore runti
                     HashSet<FunctionEndPoint> lookups;
                     if (funcGroup.CanGetExactMatchStatics(context, reducedParams, stackFrame, runtimeCore, out lookups))
                     {
-                        //Otherwise we have a cluster of FEPs that can be used to dispatch the array
-                        resolvesFeps = new List<FunctionEndPoint>(lookups);
-                        replicationInstructions = replicationOption;
-                        return;
+                        if (replicationInstructions == null || IsCompatibleReplicationOption(replicationInstructions, replicationOption))
+                        {
+                            //Otherwise we have a cluster of FEPs that can be used to dispatch the array
```
mjkkirschner: I find this comment confusing - what is the `otherwise` referring to - theres no comment before this and we are inside an if statement... not an else statement -

src/Engine/ProtoCore/Lang/CallSite.cs
reddyashish: Yes, case4: would represent the cases where there are different ranks of array's or null elements(as this is zero and others are rank 1). From my understanding the difference between case 2 and case 4 is that in case2 it finds an exact match and in case 4 it looks for other matches(like circle object being sent to curve input). In the example you attached, it is working if you fix in case 4 because int is being sent to double. If the input list has double values, it would find the match in case 2 and we will have to check all the replication options there. 
Case 6 takes care of the cases where we have other data types like int, boolean, string(but not in any nested list)  
Example: [3, [c], [[c]], "test"].

reddyashish: I had thought of reversing the list but we do not want to do that all the time. The replication options list is sent into this function is in some order that is why the computeFeps function was designed to return the first replication option(either cartesian or zipped) as soon as the match is found. So I am keeping that the same way

src/Engine/ProtoCore/Lang/CallSite.cs
reddyashish: I did not understand this case and it adds an empty replication option to the previous list and finds the match. I was not able to find any examples related to this case.

src/Engine/ProtoCore/Lang/CallSite.cs
mjkkirschner: to follow up on this - calculating the toString() of a set of replication options seems very slow - 

I tried this
`replicationTrials.Select(x=>string.Join(",",x))`

over this:

```
x = [[[null,[1,2],[3,4]]]];
y = [[[[5,6],[7,8],[9,10]]]];
```

and passed it to `Point.ByCoordinates(double,double)`

producing this:
```
   [0]: ""
    [1]: "Cartesian, Index: 0"
    [2]: "Cartesian, Index: 0,Cartesian, Index: 0"
    [3]: "Cartesian, Index: 0,Cartesian, Index: 0,Cartesian, Index: 0"
    [4]: "Cartesian, Index: 0,Cartesian, Index: 0,Cartesian, Index: 0,Cartesian, Index: 0"
    [5]: "Cartesian, Index: 1"
    [6]: "Zipped, indecies: 0, 1 - Shortest"
    [7]: "Zipped, indecies: 0, 1 - Shortest,Cartesian, Index: 0"
    [8]: "Zipped, indecies: 0, 1 - Shortest,Cartesian, Index: 0,Cartesian, Index: 0"
    [9]: "Zipped, indecies: 0, 1 - Shortest,Cartesian, Index: 0,Cartesian, Index: 0,Cartesian, Index: 0"
    [10]: "Cartesian, Index: 1,Cartesian, Index: 1"
    [11]: "Zipped, indecies: 0, 1 - Shortest,Cartesian, Index: 1"
    [12]: "Zipped, indecies: 0, 1 - Shortest,Zipped, indecies: 0, 1 - Shortest"
    [13]: "Zipped, indecies: 0, 1 - Shortest,Zipped, indecies: 0, 1 - Shortest,Cartesian, Index: 0"
    [14]: "Zipped, indecies: 0, 1 - Shortest,Zipped, indecies: 0, 1 - Shortest,Cartesian, Index: 0,Cartesian, Index: 0"
    [15]: "Cartesian, Index: 1,Cartesian, Index: 1,Cartesian, Index: 1"
    [16]: "Zipped, indecies: 0, 1 - Shortest,Cartesian, Index: 1,Cartesian, Index: 1"
    [17]: "Zipped, indecies: 0, 1 - Shortest,Zipped, indecies: 0, 1 - Shortest,Cartesian, Index: 1"
    [18]: "Zipped, indecies: 0, 1 - Shortest,Zipped, indecies: 0, 1 - Shortest,Zipped, indecies: 0, 1 - Shortest"
    [19]: "Zipped, indecies: 0, 1 - Shortest,Zipped, indecies: 0, 1 - Shortest,Zipped, indecies: 0, 1 - Shortest,Cartesian, Index: 0"
    [20]: "Cartesian, Index: 1,Cartesian, Index: 1,Cartesian, Index: 1,Cartesian, Index: 1"
    [21]: "Zipped, indecies: 0, 1 - Shortest,Cartesian, Index: 1,Cartesian, Index: 1,Cartesian, Index: 1"
    [22]: "Zipped, indecies: 0, 1 - Shortest,Zipped, indecies: 0, 1 - Shortest,Cartesian, Index: 1,Cartesian, Index: 1"
    [23]: "Zipped, indecies: 0, 1 - Shortest,Zipped, indecies: 0, 1 - Shortest,Zipped, indecies: 0, 1 - Shortest,Cartesian, Index: 1"
    [24]: "Zipped, indecies: 0, 1 - Shortest,Zipped, indecies: 0, 1 - Shortest,Zipped, indecies: 0, 1 - Shortest,Zipped, indecies: 0, 1 - Shortest"
```

and it took visual studio approx 2 minutes ... though not sure if this was a visual studio issue, this seems bad. Let's avoid the toString and compare based on the properties.

src/Engine/ProtoCore/Lang/CallSite.cs
mjkkirschner: I will try to find when it's called. 
@saintentropy any ideas?

src/Engine/ProtoCore/Lang/CallSite.cs
reddyashish: For the example in this thread, performance results after the fix(snapshot on top) and before the fix(below snapshot):

perf-new

src/Engine/ProtoCore/Lang/CallSite.cs
reddyashish: Yes, I have modified this to use the equals method in ReplicationInstructions class. Not using ToString() anymore.

src/Engine/ProtoCore/Lang/CallSite.cs
reddyashish: Looks like it. Should I change it?

src/Engine/ProtoCore/Lang/Replication/ReplicationInstruction.cs
```diff
@@ -44,5 +45,22 @@ public override string ToString()
             }
 
         }
+
+        public Boolean Equals(ReplicationInstruction oldOption) {
```
mjkkirschner: this does not seem to actually perform a full equality check... it only returns true for zip replication - you should override Equals like this

https://docs.microsoft.com/en-us/dotnet/api/system.object.equals?view=netframework-4.7.2

use the `override keyword`

src/Engine/ProtoCore/Lang/CallSite.cs
mjkkirschner: so - I'm trying to think of cases where this will fail - 
I think naming this method or adding a summary might help - but my interpretation is that this method is used to find `List<ReplicationInstruction> ` where the first instruction matches the previous one but is of a greater depth?

What I cannot figure out or explain yet is what case this is used to avoid? Can you try to sum it up in the summary of the method or in the git here.

src/Engine/ProtoCore/Lang/CallSite.cs
mjkkirschner: yes please change it.

src/Engine/ProtoCore/Lang/Replication/ReplicationInstruction.cs
mjkkirschner: it's a struct - so @reddyashish you may want to read this:
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/how-to-define-value-equality-for-a-type

mjkkirschner:
@reddyashish thanks for updating comments and tests - is the only thing left that we discussed the performance testing of a pathologically bad case? (super nested?) or many many parameter function?

reddyashish:
Performance results for a case which has around 110 replication replication trials. 
Before the fix(snapshot on top) and after the fix(snapshot on below):

latest

src/Engine/ProtoCore/Lang/CallSite.cs
saintentropy: @mjkkirschner @reddyashish In general this whole function is still black magic to me.  I have been able to look at indivual examples and how they flow through but it is hard to make judgment on individual PR's without holistic picture.   I think we need to document examples of different data, data structures, replication settings, and functions and what the result replication instruction and function endpoint list.  Not sure if we need to do that on this PR but it is hard to validate the code changes otherwise

mjkkirschner: For a start - I've filed a followup task to cover each of the missing cases with explicit tests.

src/Engine/ProtoCore/Lang/Replication/ReplicationInstruction.cs
reddyashish: @mjkkirschner Do you want me change anything in this Equals method after our discussion on friday or is this ok?

mjkkirschner: hmm, well I guess I am a bit confused why you do not just compare the properties directly? ie Why all the use of Except?

src/Engine/ProtoCore/Lang/Replication/ReplicationInstruction.cs
```diff
@@ -44,5 +45,22 @@ public override string ToString()
...
+
+            if (this.Zipped == oldOption.Zipped)
+            {
+                if (this.ZipIndecies == null && oldOption.ZipIndecies == null)
+                    return true;
```
mjkkirschner: why is this true even if the zip algorithm is different?

src/Engine/ProtoCore/Lang/Replication/ReplicationInstruction.cs
```diff
@@ -44,5 +45,22 @@ public override string ToString()
...
+
+                if (this.ZipIndecies != null && oldOption.ZipIndecies != null)
```
mjkkirschner: I'm not getting something here - can you add a comment - but like I said above, I'm confused why we're using sequence equals or something like that?

mjkkirschner: @reddyashish explained this is an optimization.

src/Engine/ProtoCore/Lang/Replication/ReplicationInstruction.cs
reddyashish: Add the zipAlgorithm check.

src/Engine/ProtoCore/Lang/Replication/ReplicationInstruction.cs
reddyashish: Fastest way to compare elements of both the lists:
https://stackoverflow.com/questions/12795882/quickest-way-to-compare-two-list

I have added comments to the code.

src/Engine/ProtoCore/Lang/Replication/ReplicationInstruction.cs
reddyashish: Added the comments to code, on why we can use Except to compare the elements in both the lists.

mjkkirschner:
@reddyashish  - Is this read to merge?

reddyashish:
The self server looks good. 
https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/219/

The performance benchmark tests for the graphs(created by Scott) are still running, so once that is done we can merge it.

reddyashish:
Performance benchmark results for graphs in here:https://github.com/DynamoDS/Dynamo/tree/master/tools/Performance/DynamoPerformanceTests/graphs

Before these changes:
before

After these changes:
after

mjkkirschner:
@scottmitchell - we really need to figure out how to get the full name in the performance results output 😉... @reddyashish what do you make of those results?

reddyashish:
The Mean, Error, StdDev values are not that different but the run time is. That is confusing to me. 
https://user-images.githubusercontent.com/43763136/54951795-e0286b00-4f1a-11e9-9df3-7f3f79ff3f3e.PNG
Also, I ran both the tests parallely.

mjkkirschner:
if you ran the tests in parallel it makes sense they will have longer run time - it also appears that it ran them both different numbers of times in each comparison - I would look only at the benchmark times relative to each other, not the total runtime.

PR #7900, opened 2017-05-25
Selective shift + click operation

yeexinc:
### Purpose

When the user has some nodes in their selection, the shift + click operation will only be done on the selected nodes.
selectiveshift

### Declarations

Check these if you believe they are true

- [ ] The code base is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions), and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.

### Reviewers

@riteshchandawar

PR #9580, opened 2019-03-15
Address refactor bug introduced in #9567

alfarok:
### Purpose

This PR address a bug introduced in #9567 where the method `EndAndStartCtrlConnection()` in `DynamoModelCommands.cs` was renamed to `BeginCreateConnections()` and where the method `EndAndStartCtrlConnection()` in `StateMachine.cs` was renamed to `BeginCreateConnections()`.  These switch statements were updated to call the wrong methods with similar naming.  All the `RecordedTests` are now passing.

### Declarations

Check these if you believe they are true

- [x] The code base is in a better state after this PR
- [x] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [x] The level of testing this PR includes is appropriate
- [x] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [x] Snapshot of UI changes, if any.
- [x] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions), and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.

### Reviewers

@mjkkirschner @QilongTang

PR #9585, opened 2019-03-18
Selective shift + click output connections grabs connectors for selected nodes only

alfarok:
### Purpose

selective_shift_click_outport

This PR pulls in the work originally completed by @yeexinc in #7900 and adds several tests that leverage recorded commands.  The recorded tests verify the proper connectors are grabbed when a `shift+click` operation occurs.  Only selected downstream nodes that are connected to the given output port will be removed upon `shift+click`ing.  Testing coverage also verifies the graph remains in a known state after several undo/redo commands are conducted after using selective shift-click operations.  The last test performs a large series of `shift+clicks` and `ctrl+clicks` with and without selected nodes to test these commands in successive combinations verifying the graph's final state produce a specific numeric combination.  All tests pass on the self-serve CI.

### Declarations

### Reviewers

@mjkkirschner @QilongTang

### DIFF
--- a/src/DynamoCore/Models/DynamoModelCommands.cs
+++ b/src/DynamoCore/Models/DynamoModelCommands.cs
@@ -362,21 +362,31 @@ namespace Dynamo.Models
             PortModel selectedPort = node.OutPorts[portIndex];
 
             var connectorsForDeletion = new List<ModelBase>();
-            int numOfConnectors = selectedPort.Connectors.Count;
+
+            var selectedConnectors = new List<ConnectorModel>();
+            selectedConnectors = selectedPort.Connectors.Where(x => x.End.Owner.IsSelected).ToList();
+
+            // If no connectors are selected, process all of the associated nodes
+            if (selectedConnectors.Count() <= 0)
+            {
+                selectedConnectors = selectedPort.Connectors.ToList();
+            }
+
+            int numOfConnectors = selectedConnectors.Count();
             if (numOfConnectors == 0) return;
             
             activeStartPorts = new PortModel[numOfConnectors];
 
             for (int i = 0; i < numOfConnectors; i++)
             {
-                ConnectorModel connector = selectedPort.Connectors[i];
+                ConnectorModel connector = selectedConnectors[i];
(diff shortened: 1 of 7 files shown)


alfarok:
All tests passing on the self-serve.  It originally showed 3 failing tests but that was because #9580 was not pulled into this branch.  I pulled in the latest from master and verified these do pass with the fix.

QilongTang:
LGTM, thank you for all the detailed comments

src/DynamoCore/Models/DynamoModelCommands.cs
```diff
@@ -362,21 +362,31 @@ private void BeginShiftReconnections(Guid nodeId, int portIndex, PortType portTy
...
+            int numOfConnectors = selectedConnectors.Count();
             if (numOfConnectors == 0) return;
             
             activeStartPorts = new PortModel[numOfConnectors];
 
             for (int i = 0; i < numOfConnectors; i++)
             {
-                ConnectorModel connector = selectedPort.Connectors[i];
+                ConnectorModel connector = selectedConnectors[i];
                 connectorsForDeletion.Add(connector);
```
QilongTang: `connectorsForDeletion` seems a dup to me after the refactor? What do you think

alfarok: I think it's still required as removing it will break undo/redo.  Below we call `CurrentWorkspace.SaveAndDeleteModels(connectorsForDeletion)` where `SaveAndDeleteModels()` is a method in `UndoRedo.cs` which is a partial class of the `WorkspaceModel`.  Several new and old recorded tests break as well.

QilongTang: I understand before the refactor it is required, but after the refactor, it contains exactly all the connectors  in `selectedConnectors`.
Can we call code below to preserve the undo/redo behavior?
```
CurrentWorkspace.SaveAndDeleteModels(selectedConnectors);
```

QilongTang:
LGTM to me now. Thank you for addressing the comment

PR #9586, opened 2019-03-19
 license cleanup

mjkkirschner:
### Purpose

cleanup some old versioning:
* unify the license.txt file and the about rtf
* diff is here: https://draftable.com/compare/VgfqNMgYPzVn
* removed some unused dlls from extern which we are not delivering or using AFAIK.

NOTE! this PR does not address the recent inclusion of `benchmark dotNet` which is MIT licensed, I will file another task for this as we need to verify any dependencies of that and verify if it belongs in about box.

### Declarations

### Reviewers

@QilongTang 

@scottmitchell 

### FYI

@kronz  @Amoursol


QilongTang:
Found some old installer stuff related to the dll removal
```
    <ModuleSpec>
      <Name>msvcp110.dll</Name>
      <FilePath>.\libg_220\msvcp110.dll</FilePath>
      <DigitalSignature>false</DigitalSignature>
      <CopyForInstaller>true</CopyForInstaller>
      <Author>Microsoft Corporation</Author>
    </ModuleSpec>
```
I don't think it is still used but would be good if you can confirm it does not affect msi building ( I don't think it will). It would be good that we remove the harvesting part of these several dlls defined in `InstallSpec.xml`. Let me know if you have concerns

scottmitchell:
LGTM!

mjkkirschner:
@QilongTang I will try to figure out whats up with this installer spec and build the installer.sln ... there are obviously already missing files, but I will see if updating it causes any issues...

mjkkirschner:
@QilongTang @scottmitchell 

heres what I think:
1. I found that the installerSpec.sln builds a command line tool called installerSpec.exe which is used to generate the installerSpec.xml file. It also appears to be used to generate the binaries to sign text.

https://github.com/DynamoDS/Dynamo/blob/master/tools/InstallerSpec/InstallerSpec/Program.cs#L242

when we build the installer the `installerSpec.xml` file is generated as part of the installer build, and output into the harvest folder.

see the command here:
https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoInstall/DynamoInstall.wixproj#L78

here is what it runs:
https://github.com/DynamoDS/Dynamo/blob/master/tools/InstallerSpec/InstallerSpec/Program.cs#L168

you can see the `xmlfilepath` is actually an output path.

After looking at this - I don't know why this spec file is committed to the repo at all. 😢

QilongTang:
LGTM

PR #9594, opened 2019-03-25
2.2  license cleanup (#9586)

mjkkirschner:
* add missing dependencies to license and license rtf - remove things we no longer use or distribute
re order so in alphabetical order to match existing pattern
Xceed was spelled wrong

simplex noise was missing
starmath was missing
ndesk was missing from license rtf
python standard lic was missing from rtf

remove lucene.net
remove winpthreads
remove 2012 msvcp dlls

update copyright dates to 2018

* remove deps we no longer use

* update msvcp version number

* formatting

* remove old installspec.xm lets see what this breaks

cherry pick:

@QilongTang @scottmitchell PTAL


QilongTang:
LGTM

PR #9600, opened 2019-03-27
DYN-1681a: As a Dynamo user, I want to resize the Set Value dialog

scottmitchell:
### Purpose

This PR addresses the first part of [DYN-1681](https://jira.autodesk.com/browse/DYN-1681), which is to allow users to resize the Set Value dialog so that they can see more of the text they are editing.
This PR does not replicate Group text editing functionality for Notes.

EditWindow_1

### Declarations

### DIFF
--- a/src/DynamoCoreWpf/UI/Prompts/EditWindow.xaml
+++ b/src/DynamoCoreWpf/UI/Prompts/EditWindow.xaml
@@ -4,6 +4,8 @@
         xmlns:ui="clr-namespace:Dynamo.UI"
         xmlns:p="clr-namespace:Dynamo.Wpf.Properties"
         Title="{x:Static p:Resources.EditWindowTitle}" 
+        MaxHeight="{x:Static SystemParameters.PrimaryScreenHeight}"
+        MaxWidth="{x:Static SystemParameters.PrimaryScreenWidth}"
         MinHeight="130" Height="200" Width="400">
 
     <Grid>
--- a/src/DynamoCoreWpf/UI/Prompts/EditWindow.xaml
+++ b/src/DynamoCoreWpf/UI/Prompts/EditWindow.xaml
@@ -4,18 +4,24 @@
         xmlns:ui="clr-namespace:Dynamo.UI"
         xmlns:p="clr-namespace:Dynamo.Wpf.Properties"
         Title="{x:Static p:Resources.EditWindowTitle}" 
-        MinHeight="130" MaxHeight="200" Width="400">
+        MinHeight="130" Height="200" Width="400">
 
-    <StackPanel>
+    <Grid>
 
-        <StackPanel.Background>
+        <Grid.Background>
             <LinearGradientBrush  StartPoint="0.5,0" EndPoint="0.5,1">
                 <GradientStop Color="#444" Offset="0.0" />
                 <GradientStop Color="#222" Offset="1.0" />
             </LinearGradientBrush>
-        </StackPanel.Background>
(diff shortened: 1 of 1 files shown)


src/DynamoCoreWpf/UI/Prompts/EditWindow.xaml
```diff
@@ -4,18 +4,24 @@
         xmlns:ui="clr-namespace:Dynamo.UI"
         xmlns:p="clr-namespace:Dynamo.Wpf.Properties"
         Title="{x:Static p:Resources.EditWindowTitle}" 
-        MinHeight="130" MaxHeight="200" Width="400">
+        MinHeight="130" Height="200" Width="400">
```
QilongTang: @scottmitchell Do you think it would be worth to still define MaxHeight and MaxWidth so the dialog will not be infinitely big in any case?

scottmitchell: I don't see why that's necessary. It will open to 200x400, and if the user wants to get crazy resizing it after that, who are we to stop them?

mjkkirschner: it's possible they will make it very large, then switch to a lower resolution - if the dialog does not resize like a window you can get into weird states where the dialog cannot be closed.

QilongTang: @mjkkirschner @scottmitchell I would encourage you to explore code to limit dialog to screen size https://stackoverflow.com/questions/30238007/wpf-maxheight-to-screen-height or I'm fine we define a big max size you think is reasonable. 

The reason I recommend defining max size is
1. It gives us a boundary of expected behavior. You will see this often in web development that div max size defined, although the actual behavior varies between platform or display logic. But it is critical for devs to look at issue and tell instantly what should happen and what is not expected to happen.
2.  I have seen bugs before like what @mjkkirschner mentioned. It is also very common for people who use multiple monitors and switch between to get into the resolution issue.

QilongTang: I have also checked the current bahavior of set value dialog, seems there is a max width of screen size already. So worth to check out behavior after your change there on max height

radumg:
Finally ! thanks @scottmitchell !

src/DynamoCoreWpf/UI/Prompts/EditWindow.xaml
scottmitchell: @QilongTang @mjkkirschner I've defined MaxHeight and MaxWidth for the edit window

QilongTang:
LGTM

QilongTang:
@scottmitchell @mjkkirschner Forgot how we conclude to unit test this, do you recall?

scottmitchell:
@QilongTang Yes, unit test is part of the AC. Working on that now.

scottmitchell:
Merging this PR and creating a new task to test this functionality with a Selenium test.

Amoursol:
@scottmitchell A few points from testing:

- Windows ran through Parallels
- Screen resolution: 3360 x 1880

Can we please make the minimum values honour the default open values, or something more appropriate than this? This also breaks the graphical icons of Minimize/Expand/Close.

image

Having 'Set Value...' up here doesn't make sense to a typical Dynamo user. Please rename to 'Write note here'

image

This is super cool! 

Dynamo_pull9600_ResizeNote

PR #9604, opened 2019-03-27
Set Performance Test Suite graphs to Manual Runtype

scottmitchell:
### Purpose

@saintentropy noticed that some of the test suite graphs were set to Automatic, which caused both graph to both Open and Run in the `OpenModelBenchmark` rather than wait for the `RunModelBenchmark` to Run. 

### Declarations

### FYIs

@saintentropy @mjkkirschner @reddyashish @QilongTang @alfarok @ColinDayOrg


mjkkirschner:
We should just open graphs with force manual mode on if we can

QilongTang:
This LG, @mjkkirschner I can also add new APIs to the DynamoModelTestBase to open graph in force manual mode, currently such API does not exist for performance test framework to use.

test/DynamoCoreTests/DynamoModelTestBase.cs
```diff
@@ -130,10 +130,10 @@ protected void EmptyScheduler()
             }
         }
 
-        protected void OpenModel(string relativeFilePath)
+        protected void OpenModel(string relativeFilePath, bool forceManualExecutionMode = false)
```
scottmitchell: @mjkkirschner @QilongTang Can we do this? Or do we need to add a new method?

scottmitchell: @QilongTang I can remove this if you would rather just handle it as part of your task.

mjkkirschner: I think this is the way to do it.

mjkkirschner: I guess we need to consider that this might effect others who have written tests against the test base classes..... 

I guess it's safest to add a new method unless you want to test this or do some research on the API compatibility of adding default arguments to protected methods?

QilongTang: That's good, can you run self CI just to make sure?

scottmitchell: For the record, this did pass the self CI, but I'm going to remove this change for now and just merge the `RunType` changes.

PR #9607, opened 2019-03-28
Performance Test Results legibility improvements

scottmitchell:
### Purpose

- Add new `Graph` column that displays the graph name
- Remove `DynamoFilePath` column
- Rename `RunModelBenchmark` -> `Run`
- Rename `OpenModelBenchmark` -> `Open`

graphName

### Declarations

- [x] Is documented according to the [standards]
(https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)

### Reviewers
@mjkkirschner @QilongTang 

### FYIs
@saintentropy

### DIFF
--- a/tools/Performance/DynamoPerformanceTests/GraphNameColumn.cs
+++ b/tools/Performance/DynamoPerformanceTests/GraphNameColumn.cs
@@ -4,9 +4,7 @@ using System.IO;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
-using BenchmarkDotNet.Attributes;
 using BenchmarkDotNet.Columns;
-using BenchmarkDotNet.Configs;
 using BenchmarkDotNet.Reports;
 using BenchmarkDotNet.Running;
 
@@ -15,14 +13,10 @@ namespace DynamoPerformanceTests
 {
     public class GraphNameColumn : IColumn
     {
-        public string Id => nameof(GraphNameColumn);
+        public GraphNameColumn() { }
+        
         public string ColumnName => "Graph";
 
-        public GraphNameColumn()
-        {
-
-        }
-        
         public string GetValue(Summary summary, BenchmarkCase benchmarkCase)
         {
             var path = benchmarkCase.Parameters["DynamoFilePath"].ToString();
@@ -31,7 +25,8 @@ namespace DynamoPerformanceTests
         }
 
         public string GetValue(Summary summary, BenchmarkCase benchmarkCase, SummaryStyle style) => GetValue(summary, benchmarkCase);
-
+        
+        public string Id => nameof(GraphNameColumn);
         public bool IsDefault(Summary summary, BenchmarkCase benchmarkCase) => true;
         public bool IsAvailable(Summary summary) => true;
         public bool AlwaysShow => true;
--- a/tools/Performance/DynamoPerformanceTests/GraphNameColumn.cs
+++ b/tools/Performance/DynamoPerformanceTests/GraphNameColumn.cs
@@ -1,0 +1,45 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Columns;
+using BenchmarkDotNet.Configs;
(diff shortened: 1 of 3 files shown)


tools/Performance/DynamoPerformanceTests/GraphNameColumn.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using BenchmarkDotNet.Columns;
+using BenchmarkDotNet.Reports;
+using BenchmarkDotNet.Running;
+
+
+namespace DynamoPerformanceTests
+{
+    public class GraphNameColumn : IColumn
```
mjkkirschner: can you add a summary to a pubic class please.

tools/Performance/DynamoPerformanceTests/GraphNameColumn.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        public GraphNameColumn() { }
```
mjkkirschner: this is here just to be a constant?
 Summary on public members would be great.

tools/Performance/DynamoPerformanceTests/GraphNameColumn.cs
```diff
@@ -0,0 +1,40 @@
...
+        public string GetValue(Summary summary, BenchmarkCase benchmarkCase)
+        {
+            var path = benchmarkCase.Parameters["DynamoFilePath"].ToString();
+            var fi = new FileInfo(path);
+            return Path.GetFileNameWithoutExtension(fi.Name);
+        }
+
+        public string GetValue(Summary summary, BenchmarkCase benchmarkCase, SummaryStyle style) => GetValue(summary, benchmarkCase);
+        
+        public string Id => nameof(GraphNameColumn);
```
mjkkirschner: I'm okay if we don't add new doc strings for these members that come from benchmarkdotnet.

mjkkirschner:
LGTM with just a few comments around docs.
Thanks for figuring this out!

tools/Performance/DynamoPerformanceTests/PerformanceTestFramework.cs
```diff
@@ -44,7 +44,12 @@ public BenchmarkConfig(string testDir)
             {
                 Add(DefaultConfig.Instance.GetLoggers().ToArray()); // manual config has no loggers by default
                 Add(DefaultConfig.Instance.GetExporters().ToArray()); // manual config has no exporters by default
-                Add(DefaultConfig.Instance.GetColumnProviders().ToArray());
+
+                var defaultColumns = DefaultConfig.Instance.GetColumnProviders().ToList();
+                defaultColumns.RemoveAt(3); // Remove DynamoFilePath column
```
QilongTang: @scottmitchell Is this column added by default or added somewhere else in the code? I am wondering if we could skip that part elsewhere, but I could be wrong

scottmitchell: @QilongTang It automatically gets created when you create the parameter sources [here](https://github.com/DynamoDS/Dynamo/blob/master/tools/Performance/DynamoPerformanceTests/PerformanceTestFramework.cs#L79). I agree that it would be better to skip the creation of this column at all, but I'm not sure that it's possible. I'll look into it further.

scottmitchell: @aparajit-pratap Do you know if there's a way to suppress the creation of the `DynamoFilePath` column?

aparajit-pratap: Nope, I don't. Need to check.

mjkkirschner:
LGTM

tools/Performance/DynamoPerformanceTests/PerformanceTestFramework.cs
QilongTang: @scottmitchell I think we may forgot to apply the same changes here to the other debug config, I will include this in my PR

scottmitchell: @QilongTang Yeah you're right. Thanks for catching that.

PR #8755, opened 2018-04-16
QNTM-5137: VM Method resolution performance improvement

aparajit-pratap:
### Purpose

The purpose of this PR is to add an optimization to the VM related to Method Resolution.  Currently the VM determines the `FunctionEndPoint` for every specific call to the node as determined by the input parameters and replication strategy.  This ensures that varied inputs can be dispatched correctly for each specific call.  The `SelectFinalFep()` method (aka select final function endpoint) is relatively fast but is often called repeatedly with the same input parameters.  This PR adds a mechanism to validate that the input parameters for the node are of the same type (ie homogeneous) which allows the `SelectFinalFep()` to be called once and the resulting `FunctionEndPoint` to be cached.  In general most user call Dynamo nodes with homogeneous inputs which allows this optimization to improve throughput of the VM in most user scenarios.  In terms of performance optimization, the new method `AreParametersHomogeneous()` is linearly bound to the number of inputs parameters vs calls to `FunctionEndPoint` which are linearly bound to the input parameter combinations determined by replication.   Additionally, the type comparison methods utilized in `AreParametersHomogeneous()` are much faster (2 orders of magnitude) then the method resolution logic required in `SelectFinalFep()` and includes short-curcuit logic in-case input parameters are found to be dissimilar.  In my profiling I found that this optimization can improve VM throughput by 12 to 20 percent for graph topologies that have lots of homogenous lists.

[QNTM-5137](https://jira.autodesk.com/browse/QNTM-5137)

### Declarations

Check these if you believe they are true

- [X] The code base is in a better state after this PR
- [X] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [X] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions), and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.

### Reviewers

@aparajit-pratap @pboyer 

### FYIs

@ramramps @mjkkirschner

PR #9619, opened 2019-04-02
FFIFunctionEndPoint Execute() performance refactor 

saintentropy:
### Purpose

The purpose of this PR is a follow up to the enhancement in https://github.com/DynamoDS/Dynamo/pull/8755 where the VM detects list data structures with homogeneous member types and short circuits duplicative creation of redundant `FFIFunctionEndPoint`objects (FFI = Foreign Function Interface).  Based on that change, this PR refactors the `Execute()` method within `FFIFunctionEndPoint` to move its initialization logic to an `Init()` that is only called once vs every call.  The refactor also removes a dependency on using the Stack data structure to pass stack values to the eventual invoke mechanism.  In the case of FFI there is no possibility of local variables being pushed back to the Stack so this use case was a redundant duplication of the existing stack values.  Lastly this PR changes how the `CallSite.StackValue ExecWithRISlowPath()` clones replication data structures as well as stack value arrays with more efficient mechanisms.

1. Performance comparison with PR improvements (Base) and master (New) (ran model tests on standard perf test suite using fast config):
image

2. Performance comparison after PR (Base) and master (New) using default config:
image

The second comparison is confusing as it does not reflect the expected perf improvements as seen in the first even though it is expected to be more accurate given that it was run with the default config. It suggests that there isn't much of an improvement at all?

### Declarations

- [] The level of testing this PR includes is appropriate

### Reviewers

@mjkkirschner @QilongTang @reddyashish I [Aparajit] have added a commit to obsolete the older API's so as not to break API compatibility and also cleaned up the PR a bit.

### FYIs


aparajit-pratap:
2 more comparisons made:
1. Performance comparison with PR improvements except for stack refactor (Base) vs. master (New):
image


2. Performance comparison with all PR improvements (Base) vs. without Stack refactor (New):
image

It can be seen that after removing the stack refactoring improvement, the Run performance degrades as much as 20% for one of the graphs.

aparajit-pratap:
@saintentropy has agreed to give this one more look and rerun perf benchmarks on the PR and comparison with baseline to determine the improvements if any. I suggested he use the default config and run only the model tests to save time. @saintentropy you can refer to this wiki to learn how to use the latest perf benchmarking tool options: https://wiki.autodesk.com/pages/viewpage.action?spaceKey=GEN&title=Dynamo+Performance+Guidelines

PR #4674, opened 2015-06-10
Fixed wonky execution issues

Benglin:
### Purpose

This pull request is meant to fix multiple *wonky Dynamo execution* issues that have recently been reported. The crux of the issue lies with multiple execution requests have been dropped due to an incorrect condition that guards against further execution in `HomeWorkspaceModel.Run`. Take for example the following graph:

update-0

When the code block node is updated from `x` to `9`, the following calls take place in sequence (they indirectly invoke `HomeWorkspaceModel.Run` method):

1. `Dynamo.Nodes.CodeBlockNodeModel.SaveAndDeleteConnectors(...)`
2. `Dynamo.Nodes.CodeBlockNodeModel.LoadAndCreateConnectors(...)`
3. `Dynamo.Nodes.CodeBlockNodeModel.SetCodeContent(...)`

Since step `1` sets `RunSettings.RunEnabled = true`, subsequently steps `2` and `3` do not result in `UpdateGraphAsyncTask` being scheduled (even though `HomeWorkspaceModel.Run` is called, it returns prematurely). Step `3` is crucial because it ultimately generates AST node with the right value for execution, without that, downstream node executes with wrong input:

update-1

The same sequence of events take place when code block node is then updated to `5`, except the fact that downstream node no longer get any input (which turned it into a partially applied `_SingleFunctionObject`):

update-2

### Declarations

Check these if you believe they are true

- [x] The code base is in a better state after this PR
- [x] The level of testing this PR includes is appropriate

### Reviewers

Hi @sharadkjaiswal, as discussed, please have a look.

### FYIs

Hi @lukechurch, @ikeough, @Randy-Ma feel free to comment on this. If the review is completed earlier, I will merge this in for wider testing.

Hi @monikaprabhu, this pull request effectively reverts the code changes in [another pull request](https://github.com/DynamoDS/Dynamo/pull/4594). For that, I have verified the fix for `MAGN-7251` remained fixed (since selection nodes get disabled when Revit document changes), but please perform more thorough tests on relevant scenarios. Thanks!

PR #9641, opened 2019-04-10
DYN-1706: Graphs with Automatic Run Type honor RunEnabled settings

reddyashish:
### Purpose

This PR is to make sure that the graphs that are set to automatic run type honor the RunEnabled settings. The RunEnabled flag is checked in the RequestRun function as well as in the ResetEngine function. 
JIRA: https://jira.autodesk.com/browse/DYN-1706

### Declarations

### Reviewers
@QilongTang @mjkkirschner @aparajit-pratap


src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -30,7 +30,7 @@ public class HomeWorkspaceModel : WorkspaceModel
         private readonly DynamoScheduler scheduler;
         private PulseMaker pulseMaker;
         private readonly bool verboseLogging;
-        private bool graphExecuted;
+        public bool graphExecuted;
```
mjkkirschner: if you need to check this in a test, make it internal and use internalVisibleTo - there's likely other ways to do this though, like checking the node values.

mjkkirschner:
@reddyashish this is kind of worrisome:
https://github.com/DynamoDS/Dynamo/pull/4674

did you check that this issue that @Benglin was working on does not reoccur after your change?

reddyashish:
Yes, this was just a draft and let's discuss about this issue in the standup.

QilongTang:
@reddyashish @mjkkirschner I think checking the node value in this case is not easy unless I am missing pieces. I tried myself and with @reddyashish together, node value is not guaranteed to be accurate, same with `homeworkspace.evaluationCount`. I expect the following test code to pass but it did not.
``` C#
            var openPath = Path.Combine(GetTestDirectory(ExecutingDirectory), @"core\RunSettings.dyn");
            ViewModel.OpenCommand.Execute(openPath);

            var ws = Model.CurrentWorkspace as HomeWorkspaceModel;

            AssertEvaluationCount(ws, 0);
            ws.RequestRun();
            // Still in Manual mode so will not run
            AssertEvaluationCount(ws, 0);

            ws.RunSettings.RunType = RunType.Automatic;
            // Setting to automatic will triger run right away
            AssertEvaluationCount(ws, 1);

            ws.RunSettings.RunEnabled = false;
            ws.RequestRun();
            // This should not run
            AssertEvaluationCount(ws, 1);

            ws.RunSettings.RunEnabled = true;
            ws.RequestRun();
            // This should run
            AssertEvaluationCount(ws, 2);
```

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
reddyashish: Changed this back to private. Not using this boolean anymore

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -331,7 +331,7 @@ internal override void RequestRun()
         {
             base.RequestRun();
 
-            if (RunSettings.RunType != RunType.Manual)
+            if (RunSettings.RunType != RunType.Manual && RunSettings.RunEnabled)
             {
```
angelowang: will feel a little better if put runenabled check first, as it is kind of precondition before run type..

test/DynamoCoreTests/Settings.cs
```diff
@@ -84,6 +85,31 @@ public void SetPythonTemplateFromConfigWithInvalidPath()
...
+            // This should not run
+            ws.RequestRun();
+            Assert.AreEqual(0, ws.EvaluationCount);
+
+            ws.RunSettings.RunEnabled = true;
+            // This should run
+            ws.RequestRun();
+            Assert.AreEqual(1, ws.EvaluationCount);
+        }
+    }
```
angelowang: would be better if we can have a test for connecting nodes on an automatic graph won't run the graph if run is disabled.

test/DynamoCoreTests/Settings.cs
```diff
@@ -84,6 +85,31 @@ public void SetPythonTemplateFromConfigWithInvalidPath()
...
+
+            var ws = CurrentDynamoModel.CurrentWorkspace as HomeWorkspaceModel;
+            Assert.AreEqual(0, ws.EvaluationCount);
+
+            // Still in Manual mode so will not run
+            ws.RequestRun();
+            Assert.AreEqual(0, ws.EvaluationCount);
 
+            // Setting the run type to automatic
+            ws.RunSettings.RunType = RunType.Automatic;
```
QilongTang: hmm Shouldn't updating RunType trigger a new run?

reddyashish: This will not trigger the run() function when the runtype is changed in a dynamo model test as there is no handler present.

test/DynamoCoreTests/Settings.cs
reddyashish: Yes, modified the test to verify this.

QilongTang:
LGTM

PR #9666, opened 2019-04-18
Move initial setup of IsAnalyticsReportingApproved to PreferenceSettings constructor

DmytroMuravskyi:
### Purpose

Alias integration needs control over analytics tick on the first run.  To achieve this I moved the variable initialization from CheckIsFirstRun to PreferenceSettings constructor. This allows override IsAnalyticsReportingApproved  before creating DynamoCore if needed. The only change to default Dynamo behavior I see is that now IsAnalyticsReportingApproved option can now migrate from previous Dynamo version instead of always being "true" during the first run. Please inform me what is the wanted behavior.

### Declarations

- [ ] Changes to the API follow 

### Reviewers
@mjkkirschner

### DIFF
--- a/src/DynamoCore/Configuration/PreferenceSettings.cs
+++ b/src/DynamoCore/Configuration/PreferenceSettings.cs
@@ -322,6 +322,7 @@ namespace Dynamo.Configuration
             // Default Settings
             IsFirstRun = true;
             IsUsageReportingApproved = false;
+            IsAnalyticsReportingApproved = true;
             LibraryWidth = 304;
             ConsoleHeight = 0;
             ShowPreviewBubbles = true;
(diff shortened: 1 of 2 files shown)


QilongTang:
@DmytroMuravskyi 
1. Did you test if this would change the fact that every new version of Dynamo launched for user, they will see the analytics term of use first time when they launch it?
2. Can you explain the reason why you would like to propose such change? Do you mean on the first run of Alias, you always set `IsAnalyticsReportingApproved` to false?

DmytroMuravskyi:
@QilongTang I didn't change migration behavior.  Analytics term of use window would still appear during first run of every new version of Dynamo. I just removed hardcoded value set, because we need to set it as "false" by default if user has no previous preferences.

QilongTang:
@DmytroMuravskyi Thanks! I verified locally and behavior change looks good to our team. However, unit tests are missing from this PR. I will add in a different PR.

PR #9667, opened 2019-04-18
DYN-1754: As a developer I want detailed per node profiling during perf benchmarks

ColinDayOrg:
Please Note:

### Purpose

This PR uses the VMDataBridge to add extra AST nodes to the beginning and end of the list of AST nodes for each workspace node that store the start and end time of compile/run in a data structure for each node for profiling reasons.

Note that tests will be added, they are just being cleaned up.

### Declarations

### Reviewers

@mjkkirschner


src/DynamoCore/DynamoCore.csproj
```diff
@@ -342,6 +352,10 @@ limitations under the License.
       <Project>{6e0a079e-85f1-45a1-ad5b-9855e4344809}</Project>
       <Name>Units</Name>
     </ProjectReference>
+    <ProjectReference Include="..\Libraries\VMDataBridge\VMDataBridge.csproj">
```
mjkkirschner: Is this a cyclic dependency?

src/DynamoCore/Engine/CodeGeneration/AstBuilder.cs
```diff
@@ -1,125 +1,36 @@
...
-    }
-
     /// <summary>
     /// AstBuilder is a factory class to create different kinds of AST nodes.
     /// </summary>
     public class AstBuilder : LogSourceBase
     {
         private readonly IAstNodeContainer nodeContainer;
 
+        internal ProfilingSession profilingSession = null;
```
mjkkirschner: is this internal for a specific reason? (sorry have not looked further down to see if that's the case yet)

src/DynamoCore/Engine/CodeGeneration/AstBuilder.cs
```diff
@@ -309,8 +224,19 @@ private void CompileToAstNodes(NodeModel node, List<AssociativeNode> resultList,
...
+            {
+                OnAstNodeBuilding(node.GUID);
+                if (profilingSession != null)
+                {
+                    profilingSession.RegisterNode(node);
+                    resultList.Add(profilingSession.CreatePreCompilationAstNode(node, inputAstNodes));
+                }
+            }
+            else if (context == CompilationContext.PreviewGraph)
+            {
```
mjkkirschner: so what is a PreviewGraph compilation context?

src/DynamoCore/Engine/CodeGeneration/AstBuilder.cs
```diff
@@ -331,42 +257,62 @@ private void CompileToAstNodes(NodeModel node, List<AssociativeNode> resultList,
...
+                    {
                         resultList.Add(item);
+                    }
+                }
+
+                if (verboseLogging)
+                {
+                    foreach (var n in resultList)
+                    {
+                        Log(n.ToString());
```
mjkkirschner: where does this log? - just curious I know this is old code.

src/DynamoCore/Engine/CodeGeneration/AstBuilder.cs
```diff
@@ -498,8 +444,7 @@ private void CompileToAstNodes(NodeModel node, List<AssociativeNode> resultList,
         /// <param name="nodeGuid"></param>
         private void OnAstNodeBuilding(Guid nodeGuid)
         {
-            if (nodeContainer != null)
-                nodeContainer.OnCompiling(nodeGuid);
+            nodeContainer?.OnCompiling(nodeGuid);
```
mjkkirschner: hmm, maybe I am confused, did you mean to use `??` or is this just a syntax for null coalescing that also works?

mjkkirschner: ahh cool:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-conditional-operators

src/DynamoCore/Engine/CodeGeneration/AstCompilationEvents.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// <summary>
+    /// Provides AST compilation related events. Clients can provide more AST
+    /// to be included during compilation events.
+    /// </summary>
+    public static class AstCompilationEvents
+    {
+        /// <summary>
+        /// Compilation for a node has begun.
+        /// </summary>
+        public static event EventHandler<CompilationEventArgs> PreCompilation;
```
mjkkirschner: are these public static events used currently? What would an example of using them look like? Should they really be static - what if there is more than one DynamoModel instantiated?

src/DynamoCore/Engine/CodeGeneration/CompilationContext.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// </summary>
+        DeltaExecution,
+
+        /// <summary>
+        /// Compiled AST nodes used in node to code.
+        /// </summary>
+        NodeToCode,
+
+        /// <summary>
+        /// Compiled AST nodes used in previwing graph.
```
mjkkirschner: any hints at what this means?

src/DynamoCore/Engine/CodeGeneration/CompilationEventArgs .cs
```diff
@@ -0,0 +1,58 @@
...
+        /// </summary>
+        public IEnumerable<AssociativeNode> AdditionalAstNodes { get { return additionalAsts; } }
+
+        /// <summary>
+        /// Returns input ast nodes as passed to the node being compiled.
+        /// </summary>
+        public IEnumerable<AssociativeNode> InputAstNodes { get; private set; }
+
+        /// <summary>
+        /// Provides an oppotunity to the event handlers to add additional AST nodes
```
mjkkirschner: spelling(oppotunity)

src/DynamoCore/Engine/CodeGeneration/CompilationEventArgs .cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        private List<AssociativeNode> additionalAsts = new List<AssociativeNode>();
+
+        /// <summary>
+        /// Creates CompilationEventArgs
+        /// </summary>
+        /// <param name="node">The node being compiled.</param>
+        /// <param name="inputNodes">List of input ast nodes.</param>
+        /// <param name="context">Compilation context.</param>
+        public CompilationEventArgs(Guid node, IEnumerable<AssociativeNode> inputNodes, CompilationContext context)
```
mjkkirschner: I want to be conservative as we're building a new API - can this be internal to start?

src/DynamoCore/Engine/CodeGeneration/CompilationEventArgs .cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Collections.Generic;
+using ProtoCore.AST.AssociativeAST;
+
+namespace Dynamo.Engine.CodeGeneration
+{
+    /// <summary>
+    /// This class contains data about NodeModel compilation when a NodeModel is being 
+    /// compiled to a list of AST nodes.
+    /// </summary>
+    public class CompilationEventArgs : EventArgs
```
mjkkirschner: curious what you about `NodeCompilationEventArgs`

src/DynamoCore/Engine/CodeGeneration/CompilationEventArgs .cs
```diff
@@ -0,0 +1,58 @@
...
+
+        /// <summary>
+        /// Returns Guid of the node being compiled.
+        /// </summary>
+        public Guid Node { get; private set; }
+
+        /// <summary>
+        /// Returns a list of additional ast nodes as added by the event handlers.
+        /// </summary>
+        public IEnumerable<AssociativeNode> AdditionalAstNodes { get { return additionalAsts; } }
```
mjkkirschner: hmm @aparajit-pratap will these always be AssociativeNodes? could they be Imperative nodes?

mjkkirschner: is there a common base class?

src/DynamoCore/Engine/CodeGeneration/CompilationEventArgs .cs
```diff
@@ -0,0 +1,58 @@
...
+        /// Returns input ast nodes as passed to the node being compiled.
+        /// </summary>
+        public IEnumerable<AssociativeNode> InputAstNodes { get; private set; }
+
+        /// <summary>
+        /// Provides an oppotunity to the event handlers to add additional AST nodes
+        /// during the compilation.
+        /// </summary>
+        /// <param name="node">Additonal AST node to add.</param>
+        public void AddAstNode(AssociativeNode node)
```
mjkkirschner: same question about a base class for both ASTnode types.

src/DynamoCore/Engine/CodeGeneration/CompiledEventArgs.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// AST nodes.
+    /// </summary>
+    public class CompiledEventArgs : EventArgs
+    {
+        /// <summary>
+        /// Construct ASTBuiltEventArgs with NodeModel and AST nodes.
+        /// </summary>
+        /// <param name="node"></param>
+        /// <param name="astNodes"></param>
+        internal CompiledEventArgs(Guid node, IEnumerable<AssociativeNode> astNodes)
```
mjkkirschner: same question here around AST type

src/DynamoCore/Engine/Profiling/IProfilingData.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using Dynamo.Graph.Nodes;
+
+namespace Dynamo.Engine.Profiling
+{
+    /// <summary>
+    /// Returns information about time spent compiling and exectuing nodes.                                                 
+    /// </summary>
+    public interface IProfilingData
```
mjkkirschner: hmm, maybe we make this more specific or marked as `X` for experimental, my thought here is that we might want to add more information to this interface and if we release with it public we'll have to worry about API changes. Thoughts? Maybe we call it `IProfilingExecutionTimeData` so we can wrap it up as a member of a larger interface later?

src/DynamoCore/Engine/Profiling/NodeProfilingData.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System;
+using Dynamo.Core;
+using Dynamo.Graph.Nodes;
+using VMDataBridge;
+
+namespace Dynamo.Engine.Profiling
+{
+    class NodeProfilingData : NotificationObject, IDisposable
```
mjkkirschner: can you mark this private? I'd prefer it's explicit for those who are not as familiar with c# accessor rules - but not a big deal if you prefer it this way.

mjkkirschner: is this our own `NotificationObject` class or interface - what is it doing here?

src/DynamoCore/Engine/Profiling/NodeProfilingData.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        NodeModel node = null;
+
+        private DateTime? startTime = null;
+        private DateTime? endTime = null;
+
+        public NodeProfilingData(NodeModel node)
```
mjkkirschner: these last three classes using the interfaces confuse me a bit - they are internal or private and they have public methods - how are they to be used or created?

please see:

https://stackoverflow.com/questions/16500593/what-does-a-public-constructor-on-an-internal-class-mean


also can we add summary /// to the constructors?

mjkkirschner: if this was intentional until we want to open these APIs I am okay with that 👍 just pointing it out.

src/DynamoCore/Engine/Profiling/ProfilingData.cs
```diff
@@ -0,0 +1,71 @@
...
+        internal void RegisterNode(NodeModel node)
+        {
+            NodeProfilingData nodeData = null;
+            if (nodes.TryGetValue(node.GUID, out nodeData))
+            {
+                nodeData.Reset();
+                return;
+            }
+
+            nodes.Add(node.GUID, new NodeProfilingData(node));
```
mjkkirschner: should this check if the node is already present in this list?

src/DynamoCore/Engine/Profiling/ProfilingSession.cs
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        internal void UnregisterDeletedNodes(IEnumerable<NodeModel> nodes)
+        {
+            profilingData.UnregisterDeletedNodes(nodes);
+        }
+
+        internal BinaryExpressionNode CreatePreCompilationAstNode(NodeModel node, List<AssociativeNode> inputAstNodes)
+        {
+            IdentifierNode identifier = AstFactory.BuildIdentifier(node.AstIdentifierBase + "_beginCallback");
```
mjkkirschner: can we pull the text out to a variable or const?

src/DynamoCore/Engine/Profiling/ProfilingSession.cs
```diff
@@ -0,0 +1,101 @@
...
+            string id = node.GUID.ToString();
+            ExprListNode exprListNode = AstFactory.BuildExprList(inputAstNodes);
+            AssociativeNode bridgeData = DataBridge.GenerateBridgeDataAst(id, exprListNode);
+
+            return AstFactory.BuildAssignment(identifier, bridgeData);
+        }
+
+        internal BinaryExpressionNode CreatePostCompilationAstNode(NodeModel node, List<AssociativeNode> inputAstNodes)
+        {
+            IdentifierNode identifier = AstFactory.BuildIdentifier(node.AstIdentifierBase + "_endCallback");
```
mjkkirschner: same here - pull out the string.

mjkkirschner:
@ColinDayOrg so this is always enabled and gathering data? I would be very interested to know if we see any noticeable performance degradation due to the additional two function calls per node.

src/DynamoCore/Engine/CodeGeneration/AstBuilder.cs
ColinDayOrg: No, has been changed to be private now

src/DynamoCore/DynamoCore.csproj
ColinDayOrg: Both DynamoCore and VMDataBridge use ProtoCore, but VMDataBridge does not use DynamoCore. I did not really like putting this reference into DynamoCore as it is in Libraries, but using the VMDataBridge was required. Possibly VMDataBridge should be moved into core, or maybe there is another way to use it in DynamoCore that I am unaware of.

src/DynamoCore/Engine/CodeGeneration/AstBuilder.cs
ColinDayOrg: I do not know. The code from Sharad called the profiling VMDataBridge code for both, but then explicitly only did anything for DeltaCompute, so I figured it would be clearer to be explicit on the higher level.

src/DynamoCore/Engine/CodeGeneration/AstBuilder.cs
ColinDayOrg: I'm not really sure, I can find out.

src/DynamoCore/DynamoCore.csproj
ColinDayOrg: As per discussion after stand-up today this seems ok for now, leaving it as is.

src/DynamoCore/Engine/CodeGeneration/AstBuilder.cs
ColinDayOrg: After discussing today it seems that no-one knows what PreviewGraph is, possibly we need a tech-debt task to figure out.

alvpickmans:
Hi @ColinDayOrg ! Hope you don't mind me chiming in but I am very interested on this feature.

We participated on the London Dynamo Hackathon with a diagnostic toolkit for Dynamo in mind, and we based our work on the existing [RechargeEast2016](https://github.com/DynamoDS/Dynamo/tree/RechargeEast2016) branch, which I think is the same you used for this branch.

Based on our experience, I wanted to share some comments:

- `Astbuilder.cs` was a flipping hell to understand from a non-dynamo developer perpective, so thanks for tidying it up and making it a bit clearer!
- We followed the same approach as on the _RechargeEast2016_ branch by implementing a `ViewExtension` that could "tap" into the `PreExecuted` and `PostExecuted` events, registering those two extra `AssociativeNodes` using the `VMDataBridge`.
- By having the `ViewExtension` we could limit the collection of metrics to only when the extension was open, but either way didn't notice any considerable performance issue.
- We implemented a basic UI that exposed the execution time of each node as a Scatter Plot graph, plus adding the time to each `NodeView` (very rough as it was a very last minute thing).

I understand @mjkkirschner comments on making some of these classes `private` to minimize the work on the new API, but as package/viewExtension developer and even regular Dynamo user, having these metrics easily exposed is very beneficial.

I found a bit obscure and difficult to follow and implement the logic using the `DataBridge` and making sure when and how methods were registered. As a developer I would find it easier if it were some `NodePreExecution` and `NodePostExecution` events I could tap into to calculate the execution time and any other metric that could be retrieved on that interval (RAM increase for example).

Not sure how feasible it would be to implement these events or if it would even be a good idea from the Dynamo API perspective, but we are more than happy to help on anything we can, so maybe we can release the `ViewExtension` publicly (atm it still needs a bit of tidy up after the hackathon :sweat_smile:)

src/DynamoCore/Engine/CodeGeneration/AstBuilder.cs
aparajit-pratap: I looked at the preview graph functionality. It is related to node previews and therefore doesn't look like it should be removed any time soon.

src/DynamoCore/Engine/CodeGeneration/CompilingEventArgs.cs
```diff
@@ -0,0 +1,21 @@
...
+        /// Construct ASTBuildingEventArgs with NodeModel.
+        public CompilingEventArgs(Guid node)
+        {
+            Node = node;
+        }
+
+        /// <summary>
+        /// Guid of NodeModel that is being compiled to AST.
+        /// </summary>
+        public Guid Node { get; private set; }
```
mjkkirschner: this should definitely be called nodeId or something besides Node

src/DynamoCore/Engine/CodeGeneration/CompiledEventArgs.cs
mjkkirschner: please change to `Node` if that causes no issues.

src/DynamoCore/Engine/CodeGeneration/CompiledEventArgs.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        /// <summary>
+        /// Guid of node that has been built to AST nodes.
+        /// </summary>
+        public Guid Node { get; private set; }
+
+        /// <summary>
+        /// Built AST nodes.
+        /// </summary>
+        public IEnumerable<AssociativeNode> AstNodes { get; private set; }
```
mjkkirschner: please change to `Node` if that causes no issues.

src/DynamoCore/Engine/CodeGeneration/CompiledEventArgs.cs
```diff
@@ -0,0 +1,34 @@
...
+        internal CompiledEventArgs(Guid node, IEnumerable<AssociativeNode> astNodes)
+        {
+            Node = node;
+            AstNodes = astNodes;
+        }
+
+        /// <summary>
+        /// Guid of node that has been built to AST nodes.
+        /// </summary>
+        public Guid Node { get; private set; }
```
mjkkirschner: can we change this to NodeGuid?

src/DynamoCore/Engine/CodeGeneration/IAstNodeContainer.cs
```diff
@@ -0,0 +1,25 @@
...
+        /// </summary>
+        /// <param name="nodeGuid"></param>
+        void OnCompiling(Guid nodeGuid);
+
+        /// <summary>
+        /// Indicates a NodeModel has been compiled to AST nodes. 
+        /// </summary>
+        /// <param name="nodeGuid"></param>
+        /// <param name="astNodes"></param>
+        void OnCompiled(Guid nodeGuid, IEnumerable<AssociativeNode> astNodes);
```
mjkkirschner: can we use `Node` instead of `AssociativeNode` ?

src/DynamoCore/Scheduler/UpdateGraphAsyncTask.cs
```diff
@@ -63,6 +63,8 @@ internal bool Initialize(EngineController controller, WorkspaceModel workspace)
                 if (graphSyncData == null)
                     return false;
 
+                engineController.UpdateProfilingData(workspace.Nodes);
```
mjkkirschner: is there a more specific location this can be handled in - the profiling session itself or is this the cleanest?

src/DynamoCore/Engine/Profiling/NodeProfilingData.cs
ColinDayOrg: After attempting to change to private; "Elements defined in a namespace cannot be explicitly declared as private, protected, protected internal, or private protected". I am assuming that you meant making the class private in this comment.

ColinDayOrg: It seems that this is our own notification item, but is not necessary here, has been removed.

src/DynamoCore/Engine/CodeGeneration/CompilingEventArgs.cs
ColinDayOrg: I agree, obsoleted and NodeId property added.

src/DynamoCore/Engine/CodeGeneration/CompiledEventArgs.cs
ColinDayOrg: Added NodeId property (similar to other changes) and obsoleted this property with comment.

src/DynamoCore/Engine/CodeGeneration/IAstNodeContainer.cs
ColinDayOrg: I don't think so, due to this being a previously public interface. Not sure if that would break binary compatibility. An overload could be added and this obsoleted, but it could be potentially confusing.

src/DynamoCore/Scheduler/UpdateGraphAsyncTask.cs
ColinDayOrg: Moved to the profiling session, engine controller methods cleaned up and test updated to reflect the change.

src/DynamoCore/Engine/EngineController.cs
```diff
@@ -69,7 +70,7 @@ private void OnTraceReconciliationComplete(TraceReconciliationEventArgs e)
         /// </summary>
         public static CompilationServices CompilationServices;
 
-        internal ProfilingSession profilingSession;
+        internal ProfilingSession profilingSession = null;
```
aparajit-pratap: Does the `ProfilingSession` need to be owned both by `AstBuilder` and `EngineController`?

ColinDayOrg: I have moved the ownership of the profiling data into the ASTBuilder only. Latest changes have just been pushed.

aparajit-pratap:
Thanks Colin, LGTM.

mjkkirschner:
@ColinDayOrg - If I wrote an extension today how would I actually collect any timing data, is the entry point on engineController meant to be internal? Is there another entry point I am missing? Or did you just want to keep it internal until we start working on the extension?

test/DynamoCoreTests/ProfilingTest.cs
```diff
@@ -71,5 +71,28 @@ public void TestProfilingSingleNode()
             engineController.EnableProfiling(false, homeWorkspace, nodes);
             Assert.IsNull(engineController.ProfilingSession);
         }
+ 
+        [Test]
+        public void TestProfilingSingleNodePublicMethodsOnly()
```
aparajit-pratap: I feel this test is redundant; it's testing what is already being tested in the first test case you added, is it not?

ColinDayOrg: It is a little redundant, but it is testing the public data accessor as opposed to the other test that is using internal methods.

PR #9680, opened 2019-04-24
DYN-1706 Make Automatic Run Type honor RunEnabled settings

reddyashish:
### Purpose

This PR is to fix an issue that was introduced after merging: https://github.com/DynamoDS/Dynamo/pull/9641

As per the current design, the variable "RunSettings.RunEnabled" is associated with the Run button in dynamo. When successive "RequestRun()" calls are triggered, this causes an issue because when the Run() adds a task to the scheduler, the RunEnabled flag is set to false till the execution is completed. It is then be set back to true in a different thread that triggers the "OnUpdateGraphCompleted" even upon completion of the first task. During this execution, if the "RequestRun" is triggered again by the "OnNodeModified" event, the second Run() wont be called because the RunEnabled is set to false at this state. https://github.com/reddyashish/Dynamo/blob/65dff195c657f39e82813cc38c45fd3a3695c30c/src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs#L337
 
To overcome this issue, we have used another boolean to track this and trigger the "Run()" if necessary. This is a temporary fix and should be avoided by using a separate variable that checks if the Run is enabled/disabled from external applications. Once we introduce that new variable, we can make this "RunSettings.RunEnabled" property private and not expose it outside as this keeps track of internal executions. The above solution would cause an API break, so added a TODO comment to the code to change it in 3.0. I will keep a track of this. 

### Declarations

### Reviewers

@aparajit-pratap @QilongTang @mjkkirschner


src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -17,6 +10,13 @@
 using Newtonsoft.Json;
 using ProtoCore;
 using ProtoCore.Namespace;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Globalization;
+using System.Linq;
+using System.Runtime.Serialization;
+using System.Xml;
```
QilongTang: Would you set system using to be before customized one in Visual Studio settings?

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -331,9 +332,16 @@ internal override void RequestRun()
...
 
-            if (RunSettings.RunEnabled && RunSettings.RunType != RunType.Manual)
+            if (RunSettings.RunType != RunType.Manual)
             {
-                Run();
+                // TODO for Dynamo 3.0: The boolean "executingTask" that is used here is a make-do fix.
+                // We will be needing a separate variable(boolean) to check for RunEnabled flag from external applications and
+                // not confuse it with the internal flag RunEnabled which is associated with the Run button in Dynamo. 
+                // Make this RunSettings.RunEnabled private, introduce the new flag and remove the "executingTask" variable. 
+                if (RunSettings.RunEnabled || executingTask)
```
QilongTang: should it be `!executingTask`?

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -331,9 +332,16 @@ internal override void RequestRun()
         {
             base.RequestRun();
 
-            if (RunSettings.RunEnabled && RunSettings.RunType != RunType.Manual)
+            if (RunSettings.RunType != RunType.Manual)
             {
-                Run();
+                // TODO for Dynamo 3.0: The boolean "executingTask" that is used here is a make-do fix.
+                // We will be needing a separate variable(boolean) to check for RunEnabled flag from external applications and
```
aparajit-pratap: This can be worded slightly differently, maybe like this: `We will be needing a separate variable (boolean flag) to check whether run can be enabled from external applications and ...`

test/DynamoCoreWpfTests/RecordedTests.cs
```diff
@@ -1425,6 +1425,25 @@ public void TestPreviewToggleConnectionMultiOutputNode()
             });
         }
 
+        [Test, RequiresSTA]
+        public void TestAddingAndRemovingConnectors()
```
aparajit-pratap: What does this test test? I thought that testing this fix was not possible in the single threaded case?

aparajit-pratap:
@reddyashish we need to make sure that although this fixes the regression introduced from your earlier fix, this PR still fixes the original issue raised by the C3D team: https://jira.autodesk.com/browse/DYN-1706. Can you verify?

test/DynamoCoreWpfTests/RecordedTests.cs
reddyashish: This is testing changing the connectors multiple times. The test was not failing before the fix but would be better to have one more test. So kept this.

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
reddyashish: No, because this boolean tracks if the previous run is completely executed or not. In the case, where OnUpdateGraphCompleted event is not completed, we want to trigger the next runs. In the end the executing task will be set to false.

QilongTang: I see, since we are using `||`, normal case it will just be dominated by `RunSettings.RunEnabled` which is always true. But once `RunEnabled ` become false, `run()` can't be called at all

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -31,6 +31,7 @@ public class HomeWorkspaceModel : WorkspaceModel
         private PulseMaker pulseMaker;
         private readonly bool verboseLogging;
         private bool graphExecuted;
+        private bool executingTask;
```
QilongTang: Would you add some comments here even if it is private?

reddyashish:
@aparajit-pratap yes, I have tested this again with Revit locally as Civil3D already has a workaround for this presently. It fixes the original issue. We also have a test case to test that workflow.

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
reddyashish: Yes will add some comments.

QilongTang:
LGTM, and I believe this is what Civil3D team expects

PR #9690, opened 2019-04-30
Simplify build NuGet script

aparajit-pratap:
### Purpose
Simplify build NuGet script.

### Declarations

### FYIs

@QilongTang

### DIFF
--- a/tools/NuGet/BuildPackages.bat
+++ b/tools/NuGet/BuildPackages.bat
@@ -39,27 +39,13 @@ set version=%Major%.%Minor%.%Build%-beta%Revision%
 
 :: Clean files generated from the previous run
 if exist *.nupkg ( del *.nupkg )
-if exist nuspec ( rmdir /s /q nuspec )
-mkdir nuspec
-
-:: Copy .nuspec files from template folder to "nuspec" folder
-:: and replace the string "@VERSION@" with the correct value
-for %%f in (%1\*.nuspec) do (
-  for /f "tokens=* delims=¶" %%i in ( '"type %%f"') do (
-    set line=%%i
-    setlocal EnableDelayedExpansion
-    set line=!line:@VERSION@=%version%!
-    echo !line!>>nuspec\%%~nxf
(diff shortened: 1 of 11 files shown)


tools/NuGet/BuildPackages.bat
```diff
@@ -39,27 +39,13 @@ set version=%Major%.%Minor%.%Build%-beta%Revision%
...
-    setlocal EnableDelayedExpansion
-    set line=!line:@VERSION@=%version%!
-    echo !line!>>nuspec\%%~nxf
-    endlocal
-  )
-)
 
 :: Pack .nupkg files based on each .nuspec in the "nuspec" folder
-for %%f in (nuspec\*.nuspec) do (
+for %%f in (%1\*.nuspec) do (
```
mjkkirschner: hmm, I assume that not everyone on the team will be a regular .bat file author - can you add a comment for this syntax.

also... what is the main difference in behavior after this PR?

aparajit-pratap: There is no difference in behavior. Previously there was complex batch scripting logic to parse the nuspec file, search for the keyword, `@VERSION@` and then replace it with the value of the `%version%` variable obtained in the previous lines, and finally make a copy of the nuspec file with this version number in a separate folder called `nuspec`. This change is to get rid of this logic by simply parameterizing the version in the nuspec file and setting the parameter to the version number obtained in the bat file. This is done using a `-properties` option while using the `nuget pack` command. I guess the only difference is that after this change there is no need for the redundant nuspec folder. The nuspecs from the original folders can be used, namely, `template-nuget` and `template-artifactory`. These are passed as command line arguments (which is what the %1 is) while invoking the bat script.

QilongTang: Make me think worth noting this down in https://github.com/DynamoDS/Dynamo/blob/master/tools/NuGet/README.md

QilongTang:
LGTM

PR #9718, opened 2019-05-16
consume latest packageManagerClient targeted to .net45

mjkkirschner:
build succeeds locally
PTAL
@aparajit-pratap @scottmitchell


mjkkirschner:
this build fails because package has not yet been pushed to nuget - I am consuming it locally from a local nuget repo currently.

scottmitchell:
LGTM

PR #9721, opened 2019-05-18
try to fix profiling tests

mjkkirschner:
### Purpose

Make new profiling tests run locally reliably.

I added more nodes to the singleNode.dyn (now a bad name) 😉 - it now creates 15 X 15 X 15 points - so it is more measurable.

I also fixed the `AssertNonNull` mirrorData method which did not work for Collections... I've added a TODO to add the other cases... we might consider implementing them now to avoid confusing someone who tries to use it.

### Declarations

### Reviewers

@ColinDayOrg 
@aparajit-pratap


test/DynamoCoreTests/DSEvaluationUnitTestBase.cs
```diff
@@ -127,6 +127,13 @@ protected void AssertNonNull(string guid)
             Assert.IsNotNull(mirror);
 
             var mirrorData = mirror.GetData();
+            if (mirrorData.IsCollection)
+            {
+                var elements = mirrorData.GetElements();
+                Assert.IsNotNull(elements);
+                return;
+            }
+
```
aparajit-pratap: 👍

PR #9723, opened 2019-05-19
[PTAL] [WIP] DYN-1662: As a dynamo dev, I want to serialize what packages are needed in a graph

scottmitchell:
### Purpose

This PR adds `PackageDependencies` as a top level item in the .dyn file. Now, when a WorkspaceModel is serialized, it will collect locally installed package data from the package manager, and then serialize the data of those packages that are used in the graph. It currently looks like this:

```
"PackageDependencies": [
    {
      "Name": "HowickMaker",  // Zerotouch
      "Version": "0.3.0",
      "Dependents": [
        "f35f1f65418a4bf392e3176a28f2dd51"
      ]
    },
    {
      "Name": "NodeModelCharts",  // NodeModel
      "Version": "2.0.2",
      "Dependents": [
        "4c6e96b3da44456f81a52c587316cd3b",
        "d44934a410c54e72b2aeedd4b7bff526"
      ]
    },
    {
      "Name": "Clockwork for Dynamo 2.x",  // Custom Node
      "Version": "2.1.2",
      "Dependents": [
        "00d7d5e810d3443b9909980da84b8ce2"
      ]
    }
  ],
```

### TODO:

- [ ] Tests

### Declarations

### Reviewers

@mjkkirschner @QilongTang 

### FYIs

@varvaratou


mjkkirschner:
@gregmarr PTAL this PR impacts the schema.

src/DynamoPackages/PackageManagerExtension.cs
```diff
@@ -120,7 +123,33 @@ public void Startup(StartupParams startupParams)
             LoadPackages(startupParams.Preferences, startupParams.PathManager);
         }
 
-        public void Ready(ReadyParams sp) { }
+        public void Ready(ReadyParams sp)
+        {
+            sp.CurrentWorkspaceChanged += OnCurrentWorkspaceChanged;
```
mjkkirschner: these events and the ones below should be unsubscribed when those workspaces are no longer the current workspace or the workspaces will not be able to be garbage collected -- not sure about there logic there in all cases, ie what about custom node workspaces?

We should also implement cleanup in the shutdown or dispose method of this Extension, but be careful as there is an inconsistency between shutdown and dispose between viewExtension and Extensions - I can't remember which is really called on extensions 😢 - please check.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -436,6 +437,11 @@ protected virtual void OnSaving(XmlDocument obj)
             if (handler != null) handler(obj);
         }
 
+        /// <summary>
+        /// Event that is fired during the saving of the workspace to JSON.
+        /// </summary>
+        public event Func<WorkspaceModel, JObject> SavingJson;
```
mjkkirschner: there is no event for this already?

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -1040,6 +1050,26 @@ public virtual void Save(string filePath, bool isBackup = false, EngineControlle
             }
         }
 
+        internal JObject AddExtensionBlockToJSON(JObject data)
+        {
+            var extensionsBlock = new JObject();
+            if (SavingJson != null)
+            {
+                foreach (Func<WorkspaceModel, JObject> f in SavingJson.GetInvocationList())
```
mjkkirschner: this is unusual but I guess fine - I think it might be more idiomatic to call the event with the JSON object and have the extensions inject their json data into it. - Though I guess this way we retain more control over the serialized data which is nice.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -1040,6 +1050,26 @@ public virtual void Save(string filePath, bool isBackup = false, EngineControlle
...
+                {
+                    if (f.Target is IExtension)
+                    {
+                        var extensionData = f.Invoke(this);
+                        var extensionName = (f.Target as IExtension).Name;
+                        extensionsBlock.Add(new JProperty(extensionName, extensionData));
```
mjkkirschner: what if they return nothing or invalid data?

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -1021,9 +1027,13 @@ public virtual void Save(string filePath, bool isBackup = false, EngineControlle
 
                 // Stage 1: Serialize the workspace.
                 var json = this.ToJson(engine);
```
QilongTang: @mjkkirschner @scottmitchell I wonder if you guys have discussed adding a Json Serializable property `Extensions` with getter only so that ToJson() will take care of it by default. I believe 3rd party developer should not need to know special workaround `AddExtensionBlockToJSON` to be able to save the Model into the official scheme we have. Otherwise, when they make the call and for example upload the json to us, it will be missing information. My 2 cents

gregmarr:
Are the three levels of nesting here (`"Extensions": { "DynamoPackageManager": {      "InstalledPackages"`) because additional sibling items are expected to be added in the future?

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
scottmitchell: That makes sense. I'll take a look at this.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
scottmitchell: I do think it makes more sense to not give them the full json. There may still be a more idiomatic way to structure this though.

scottmitchell:
@gregmarr Yes, part of the idea here is to have a general way for extensions to serialize data to the .dyn file. In the future, this could include other internally developed extensions, and 3rd party extensions. These would all be siblings within `Extensions`. 

As far as the schema inside of `DynamoPackageManager`, I just threw `InstalledPackages` in there as a POC for now. What we really care about are the `PackageDependencies` of the .dyn (currently working on adding this). It's possible we would want to add sibling items here in the future.

gregmarr:
"part of the idea here is to have a general way for extensions to serialize data to the .dyn file. In the future, this could include other internally developed extensions, and 3rd party extensions. These would all be siblings within Extensions."

We already explicitly allow additional data under the `View` and `NodeView` sections.  That's where Dynamo is putting its additional data.

As this is data about which packages are needed to run the graph, this should be top-level data, as a sibling to the existing `Dependencies` which is for graphs that this graph depends on.
https://git.autodesk.com/aecgd/cogs/blob/master/api/service.yaml#L1538-L1545

Something like this:

      PackageDependencies:
        description: Ids of `Package`s used by this `Graph`.
        x-omitempty: true
        type: array
        items:
          $ref: "#/definitions/Package"

    Package:
      description: The information on a package available from the Package Manager.
      type: object
      required:
      - Name
      - Version
      properties:
        Name:
          description: The name of the package.
          type: string
          example: Dynamo Node Package
        Version:
          description: The version string for the package.
          type: string
          example: 1.2.3

scottmitchell:
@gregmarr thanks for the feedback. Makes sense. I’ll adopt the schema you’ve described here.

gregmarr:
Another thing that it would be good to add to a `Package` record would be the type of contents.  For example, if it's just custom nodes, or if it also contains ZT nodes.

aparajit-pratap:
@scottmitchell what was the original idea of letting extensions to serialize their data as JSON? Are there cases where the graph (nodes) could have a dependency on extensions (other than packages)?

scottmitchell:
@aparajit-pratap I'm not sure that there are cases where the graphs depend on an extension (although that's an interesting question). The idea here was just to explore the possibility of allowing extension developers to serialize data to the dyn—just a new, potentially useful api.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -513,6 +528,78 @@ public HashSet<Guid> Dependencies
...
+        {
+            get
+            {
+                // Get all packages installed locally
+                var localPackages = GetLocalPackages();
+                // Get set of assemblies used by nodes in this workspace
+                var assembliesUsed = GetAssembliesUsed();
+
+                // Create a dictionary that maps assembly names to the package they are contained in
+                var assemblyPackageDict = new Dictionary<string, IPackage>();
```
mjkkirschner: you may want to actually use assembly objects here - including assembly versions, I have a feeling that will be useful...

src/DynamoCore/Interfaces/IPackage.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        /// <summary>
+        /// Name of the package
+        /// </summary>
+        string Name { get; set; }
+
+        /// <summary>
+        /// Version of the package
+        /// </summary>
+        string VersionName { get; set; }
```
mjkkirschner: version name?

src/DynamoCore/Interfaces/IPackage.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Dynamo.Interfaces
+{
+    /// <summary>
+    /// Interface that contains basic properties of a Dynamo package
+    /// </summary>
+    public interface IPackage
```
mjkkirschner: this must exist somewhere else? Is it a case where you don't want to add a reference between projects?

src/DynamoCore/Interfaces/IPackage.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// </summary>
+        string Name { get; set; }
+
+        /// <summary>
+        /// Version of the package
+        /// </summary>
+        string VersionName { get; set; }
+
+        /// <summary>
+        /// Assemblies contained in this package that are node libraries
```
mjkkirschner: lets get more specific about this - I'm not sure if this means just zero touch assemblies, or assemblies containing NodeModels or those containing NodeModel customizations? Can we find out?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1771,6 +1789,30 @@ private void RegisterHomeWorkspace(HomeWorkspaceModel newWorkspace)
             };
         }
 
+        private HashSet<string> OnCollectingAssembliesUsed()
+        {
+            var assemblies = new HashSet<string>();
+            foreach(var node in CurrentWorkspace.Nodes)
+            {
+                IEnumerable<FunctionDescriptor> descriptors = LibraryServices.GetAllFunctionDescriptors(node.CreationName.Split('@')[0]);
```
mjkkirschner: document this

mjkkirschner: tiny nit pick, because this is a bit wide, it doesn't view well on github.

src/DynamoPackages/PackageManagerExtension.cs
```diff
@@ -23,6 +24,8 @@ public class PackageManagerExtension : IExtension, ILogSource, IExtensionSource
 
         public event Action<ILogMessage> MessageLogged;
 
+        private IWorkspaceModel currentWorkspace;
```
mjkkirschner: why store this here instead of just looking it up?

src/DynamoPackages/PackageManagerExtension.cs
```diff
@@ -149,6 +159,25 @@ private void OnMessageLogged(ILogMessage msg)
...
             }
         }
+        
+        private void OnCurrentWorkspaceChanged(IWorkspaceModel ws)
+        {
+            if (ws is WorkspaceModel)
+            {
+                if (currentWorkspace != null)
+                {
+                    (currentWorkspace as WorkspaceModel).CollectingPackageDependencies -= GetLocalPackages;
```
mjkkirschner: hmm, are you sure you want to do this when a user opens a custom node workspace for example? Is what you really want to unsubscribe when the user closes the home work space? - I am 100% not sure - just curious if you considered custom node workspaces.

src/DynamoCore/Models/DynamoModel.cs
mjkkirschner: hmm, won't this only work for ZeroTouch nodes, what about nodeModel nodes inside of packages?

src/DynamoCore/Interfaces/IPackage.cs
scottmitchell: That's what the `Package` type calls its version property. Seemed weird to me too.

mjkkirschner: 🤦‍♂

src/DynamoCore/Interfaces/IPackage.cs
scottmitchell: Yeah, we can't reference Package Manager in Core. I can't find it anywhere else.

src/DynamoCore/Interfaces/IPackage.cs
mjkkirschner: is it ever a name? Or is it always just a version string?

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
mjkkirschner: or AssemblyName objects.

mjkkirschner: https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assemblyname?view=netframework-4.8

src/DynamoPackages/PackageManagerExtension.cs
scottmitchell: Don't we want `PackageDependencies` serialization to work for both .dyn and .dyf? My intention here was that if you save while in a custom node workspace (i.e. while `currentWorkspace` is a custom node), it should save the `PackageDependencies` of the custom node.

src/DynamoCore/Interfaces/IPackage.cs
scottmitchell: As far as I can tell it's always the version

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
scottmitchell: One thing that seems tricky to me here is that we currently only have the assembly _path_ from `AssembliesUsed` (which I'm getting from Library Services). So to look up packages in the dictionary we would first need a way to get the assembly `Version` so we can make the `AssemblyName` object.

scottmitchell: We could use this? 

[https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.loadfrom?view=netframework-4.8](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.loadfrom?view=netframework-4.8)

Would that hurt performance?

src/DynamoPackages/PackageManagerExtension.cs
scottmitchell: Alternatively, we could put all `PackageDependencies`, including those of custom nodes the graph depends on, up in the .dyn.

scottmitchell: That seems weird to me though

mjkkirschner: I think your intention sounds better than the second description since we discussed top level dependencies only.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
mjkkirschner: the assemblies should already be loaded ... and if not we don't want to load them - we might want to load them into a reflection only context - also yes, loading assemblies that are not already loaded will be slow.

mjkkirschner: Instead, I think what we want to do is see if the assembly is already loaded into the current app domain.

https://docs.microsoft.com/en-us/dotnet/api/system.appdomain.getassemblies?view=netframework-4.8#System_AppDomain_GetAssemblies

this can also be slow - but faster than re loading from disk.

I think this is what we want though ⭐️ 
https://stackoverflow.com/questions/30306702/assemblyinfo-without-loading

scottmitchell: Nevermind.
https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assemblyname.getassemblyname?view=netframework-4.8

mjkkirschner: that may not work for all assemblies - only those with assembly manifests - in the case of packages - that are being used in the graph, we may be able to make the assumption that the dll is already loaded into the appdomain...

src/DynamoCore/Models/DynamoModel.cs
scottmitchell: Right, good catch. Works for ZeroTouch and NodeModel now.

src/DynamoPackages/PackageManagerExtension.cs
scottmitchell: I believe this is addressed now

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
scottmitchell: I agree with that assumption. Moving forward with this strategy.

src/DynamoPackages/Package.cs
```diff
@@ -29,7 +29,7 @@ public string Name
         }
     }
 
-    public class Package : NotificationObject, ILogSource
+    public class Package : NotificationObject, ILogSource, IPackage
```
aparajit-pratap: API break?

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -513,6 +529,77 @@ public HashSet<Guid> Dependencies
...
+                    }
+                }
+
+                // Create a list of packages that are used in this workspace
+                var packageDependencies = new HashSet<IPackage>();
+                foreach(var assemblyName in assembliesUsed)
+                {
+                    if (assemblyPackageDict.ContainsKey(assemblyName.FullName))
+                    {
+                        packageDependencies.Add(assemblyPackageDict[assemblyName.FullName]);
```
aparajit-pratap: What about custom node packages. Those won't have assemblies.

scottmitchell: Thanks for pointing this out. I just added support for this here: https://github.com/DynamoDS/Dynamo/pull/9723/commits/91b55c575f2026724582b7884917c32d6c980cad

src/DynamoPackages/Package.cs
scottmitchell: @aparajit-pratap @mjkkirschner What do you think about adding a `PackageInfo` class to avoid having to pass `Package`/`IPackage` objects around and to avoid the API break? See this commit: https://github.com/DynamoDS/Dynamo/pull/9723/commits/21181818535dc800a378c8e4af86116429348a24

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
```diff
@@ -739,6 +744,38 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
...
+        {
+            get { return false; }
+        }
+
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            IPackage p = value as IPackage;
+            writer.WriteStartObject();
+            writer.WritePropertyName("Name");
+            writer.WriteValue(p.Name);
```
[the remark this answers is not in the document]
scottmitchell: Just added

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -436,6 +438,25 @@ protected virtual void OnSaving(XmlDocument obj)
...
+        /// <summary>
+        /// Event that is fired when the workspace is collecting used assemblies
+        /// </summary>
+        internal delegate HashSet<AssemblyName> CollectingAssembliesUsedHandler();
+        internal event CollectingAssembliesUsedHandler CollectingAssembliesUsed;
+
+        /// <summary>
+        /// Event that is fired when the workspace is collecting custom node package dependencies
+        /// </summary>
+        [Obsolete("This event will ignore subscribers other than PackageManagerExtension.")]
```
scottmitchell: I'm thinking the right way to handle this is to actually to use `InternalsVisibleToAttribute`

src/DynamoCore/Extensions/PackageInfo.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Dynamo.Extensions
+{
+    /// <summary>
+    /// Class containing info about a Dynamo package. 
+    /// Used for serialization.
+    /// </summary>
+    public class PackageInfo
```
mjkkirschner: seems like an odd namespace for this class?

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
```diff
@@ -739,6 +744,41 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
         }
     }
 
+    /// <summary>
+    /// IPackageWriteConverter is used to Serialize packages to JSON.
```
mjkkirschner: comment seems incorrect.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -26,6 +29,8 @@
 using Newtonsoft.Json.Linq;
 using ProtoCore.Namespace;
 
+[assembly: InternalsVisibleTo("DynamoPackages")]
```
mjkkirschner: another way to solve this is with reflection - if you go this route - others might use this unintentionally, please add a comment here at the minimum if possible.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -513,6 +535,57 @@ public HashSet<Guid> Dependencies
...
+
+                return packageDependencies;
+            }
+        }
+
+        /// <summary>
+        /// Gathers the assemblies that are used by nodes in this workspace
+        /// </summary>
+        /// <returns></returns>
+        internal HashSet<AssemblyName> GetAssembliesUsed()
```
mjkkirschner: can you make this name and comment more explicit, it doesn't gather the full tree of dependencies `used` by the graph, it gathers the top level references of those packages that are present in the graph.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1771,6 +1789,46 @@ private void RegisterHomeWorkspace(HomeWorkspaceModel newWorkspace)
...
         }
 
+        /// <summary>
+        /// Collects assemblies used by the current workspace
+        /// </summary>
+        /// <returns></returns>
+        private HashSet<AssemblyName> OnCollectingAssembliesUsed()
+        {
+            var assemblyNames = new HashSet<AssemblyName>();
+            foreach(var node in CurrentWorkspace.Nodes)
```
mjkkirschner: this has the possibility of being quite slow... I wonder if we can turn this off during back up save? Can we look at the performance impact of this?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1771,6 +1789,46 @@ private void RegisterHomeWorkspace(HomeWorkspaceModel newWorkspace)
...
+            {
+                // Get zerotouch assemblies
+                if (node.GetType() == typeof(DSFunction))
+                {
+                    // Get FunctionDescriptors for this zerotouch node
+                    var descriptors = LibraryServices.GetAllFunctionDescriptors(node.CreationName.Split('@')[0]);
```
mjkkirschner: maybe use firstOrDefault() and please comment on why split on "@"

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1771,6 +1789,46 @@ private void RegisterHomeWorkspace(HomeWorkspaceModel newWorkspace)
...
+                    if (descriptors != null)
+                    {
+                        foreach (FunctionDescriptor fd in descriptors)
```
mjkkirschner: why even do this comparison if the descriptor is not a package member? Why not skip all this?

scottmitchell:
@gregmarr @mjkkirschner Proposing an addition to the schema change in the most recent commit (https://github.com/DynamoDS/Dynamo/pull/9723/commits/1638aeb79057cfc8e98c2c4bcdf7ec643e059d38). Each package dependency would now also serialize the IDs of dependent nodes. Could be useful for a number of things, including to determine the type of contents (zerotouch, custom node, etc) of the package as Greg suggested above. Looks like this:
```
"PackageDependencies": [
    {
      "Name": "HowickMaker",
      "Version": "0.3.0",
      "Dependents": [
        "f35f1f65-418a-4bf3-92e3-176a28f2dd51"
      ]
    },
    {
      "Name": "NodeModelCharts",
      "Version": "2.0.2",
      "Dependents": [
        "4c6e96b3-da44-456f-81a5-2c587316cd3b",
        "d44934a4-10c5-4e72-b2ae-edd4b7bff526"
      ]
    },
    {
      "Name": "Clockwork for Dynamo 2.x",
      "Version": "2.1.2",
      "Dependents": [
        "00d7d5e8-10d3-443b-9909-980da84b8ce2"
      ]
    }
  ],
```

Thoughts?

src/DynamoCore/Extensions/PackageInfo.cs
scottmitchell: Agreed. I moved it to `Dynamo.Graph.Workspaces`. Also thinking of renaming it `PackageDependency`, which might make more sense in the context of workspaces.

src/DynamoPackages/PackageManagerExtension.cs
```diff
@@ -179,6 +182,32 @@ private IEnumerable<IPackage> GetLocalPackages()
             return PackageLoader.LocalPackages;
         }
 
+        private IEnumerable<IPackage> GetCustomNodesPackagesFromGuid(IEnumerable<Guid> functionIDs)
+        {
+            // Create dictionary mapping Guids to packages
+            var guidPackageDictionary = new Dictionary<Guid, IPackage>();
+            {
```
aparajit-pratap: Additional scope does not seem to be required.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -541,28 +540,23 @@ public IEnumerable<PackageInfo> PackageDependencies
...
+                // Collect pacakge dependencies of nodes in workspace
                 var packageDependencies = new HashSet<PackageInfo>();
-                foreach(var assemblyName in assembliesUsed)
+                if (CollectingNodePackageDependencies != null)
                 {
-                    if (assemblyPackageDict.ContainsKey(assemblyName.FullName))
+                    foreach (var f in CollectingNodePackageDependencies.GetInvocationList())
                     {
-                        packageDependencies.Add(assemblyPackageDict[assemblyName.FullName]);
+                        if (f.Target is IExtension && (f.Target as IExtension).UniqueId == "FCABC211-D56B-4109-AF18-F434DFE48139")
```
aparajit-pratap: Can we get a hold of the pm extension object and query its `uniqueID` instead of hardcoding the ID here? Ideally the unique ID property should have been static.

src/DynamoCore/Models/DynamoModel.cs
scottmitchell: Good point. Moved the `IsPackageMember` check up

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
scottmitchell: Changed to `GetAssembliesReferencedByNodes`

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
scottmitchell: Comment added. I'm not sure what you mean here about solving with reflection. If you think that's a better strategy, can we discuss in person how to implement it?

src/DynamoCore/Models/DynamoModel.cs
scottmitchell: I'll look into both of these ideas

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
scottmitchell: Marking as resolved since we aren't using this approach anymore

src/DynamoCore/Graph/Workspaces/PackageInfo.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// <summary>
+        /// The full name of this package, including its name and version
+        /// Format: name_version
+        /// </summary>
+        internal string FullName { get { return Name + "_" + Version; } }
+
+        /// <summary>
+        /// Guids of nodes in the workspace that are dependent on this package
+        /// </summary>
+        internal List<Guid> Dependents
```
mjkkirschner: hmm, DependentNodes?

src/DynamoCore/Graph/Workspaces/PackageInfo.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// <summary>
+        /// Version of the package
+        /// </summary>
+        internal string Version { get; set; }
+
+        /// <summary>
+        /// The full name of this package, including its name and version
+        /// Format: name_version
+        /// </summary>
+        internal string FullName { get { return Name + "_" + Version; } }
```
mjkkirschner: what is this used for?

src/DynamoCore/Graph/Workspaces/PackageInfo.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        /// <summary>
+        /// Name of the package
+        /// </summary>
+        internal string Name { get; set; }
+
+        /// <summary>
+        /// Version of the package
+        /// </summary>
+        internal string Version { get; set; }
```
mjkkirschner: is there a reason not to use a Version object?

src/DynamoCore/Graph/Workspaces/PackageInfo.cs
mjkkirschner: can you use an auto implemented property for this or access modifiers make it tough?

src/DynamoCore/Graph/Workspaces/PackageInfo.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Dynamo.Graph.Workspaces
+{
+    /// <summary>
+    /// Class containing info about a Dynamo package. 
+    /// Used for serialization.
+    /// </summary>
+    internal class PackageInfo
```
mjkkirschner: does this same class already exist in another namespace? Just want to make sure we don't have two types with the same name.

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
```diff
@@ -739,6 +743,48 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
         }
     }
 
+    /// <summary>
+    /// PackageInfoWriteConverter is used to Serialize graph package dependency to JSON.
+    /// </summary>
+    public class PackageInfoWriteConverter : JsonConverter
```
mjkkirschner: are you imagining being able to deserialize this information will be part of another PR? Another Converter?

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
mjkkirschner: yes, let's talk about the different options we have besides this.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -513,6 +536,79 @@ public HashSet<Guid> Dependencies
...
+                        var assemblyName = assembliesUsed[nodeID];
+                        var package = CollectingNodePackageDependencies(assemblyName);
+                        if (package != null)
+                        {
+                            guidPackageDictionary[nodeID] = package;
+                        }
+                    }
+                }
+
+                // Collect pacakge dependencies for custom nodes
```
mjkkirschner: (sp)package

src/DynamoPackages/PackageManagerExtension.cs
```diff
@@ -23,6 +24,20 @@ public class PackageManagerExtension : IExtension, ILogSource, IExtensionSource
...
+        /// Dictionary mapping a custom node functionID to the package that contains it.
+        /// Used for package dependency serialization.
+        /// </summary>
+        private Dictionary<Guid, PackageInfo> CustomNodePackageDictionary;
+
+        /// <summary>
+        /// Dictionary mapping the AssemblyName.FullName of an assembly to the package that contains it.
+        /// Used for package dependency serialization.
+        /// </summary>
+        private Dictionary<string, PackageInfo> NodePackageDictionary;
```
mjkkirschner: why not use AssemblyName objects?

src/DynamoPackages/PackageManagerExtension.cs
```diff
@@ -149,6 +176,87 @@ private void OnMessageLogged(ILogMessage msg)
...
+        {
+            // Create NodePackageDictionary if it doesn't exist
+            if (NodePackageDictionary == null)
+            {
+                NodePackageDictionary = new Dictionary<string, PackageInfo>();
+            }
+            // Add new assemblies to NodePackageDictionary
+            foreach (var assembly in package.LoadedAssemblies.Select(a => AssemblyName.GetAssemblyName(a.Assembly.Location)))
+            {
+                NodePackageDictionary[assembly.FullName] = new PackageInfo(package.Name, package.VersionName);
```
mjkkirschner: so what happens if somehow theres already a mapped package here at this key - I think we should log something to console or throw - this could happen if two packages reference the same .dll and expose nodes from it... I think this is related to what @aparajit-pratap is working on. So perhaps this needs to be `lists of <PackageInfo>`

src/DynamoPackages/PackageManagerExtension.cs
```diff
@@ -149,6 +176,87 @@ private void OnMessageLogged(ILogMessage msg)
...
+            // Create CustomNodePackageDictionary if it doesn't exist
+            if (CustomNodePackageDictionary == null)
+            {
+                CustomNodePackageDictionary = new Dictionary<Guid, PackageInfo>();
+            }
+            // Add new custom nodes to CustomNodePackageDictionary
+            foreach (var cn in package.LoadedCustomNodes)
+            {
+                var pInfo = new PackageInfo(package.Name, package.VersionName);
+                CustomNodePackageDictionary[cn.FunctionId] = pInfo;
```
mjkkirschner: same issue here - what if two packages expose the same custom node... I'm not sure this is possible - BUT it's definitely possible for one package to republish another package's custom nodes - I am not sure if `LoadedCustomNodes` will have those dependent nodes included though.

src/DynamoCore/Graph/Workspaces/PackageInfo.cs
mjkkirschner: Scott likes Nodes

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
```diff
@@ -739,6 +743,48 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
...
+                    writer.WriteValue(dependent.ToString("N"));
+                }
+                writer.WriteEndArray();
+                writer.WriteEndObject();
+            }
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            throw new NotImplementedException();
```
scottmitchell: Add deserialization

src/DynamoCore/Graph/Workspaces/PackageInfo.cs
scottmitchell: We have a `PackageFileInfo` class. https://github.com/DynamoDS/Dynamo/blob/ec10f936824152e7dd7d6d019efdcda0d78a5264/src/DynamoPackages/PackageFileInfo.cs

I'm going to rename this new class `PackageDependencyInfo` for more clarity.

src/DynamoPackages/PackageManagerExtension.cs
```diff
@@ -149,6 +176,87 @@ private void OnMessageLogged(ILogMessage msg)
...
+            }
+        }
+
+        private void OnPackageRemoved(Package package)
+        {
+            var pInfo = new PackageInfo(package.Name, package.VersionName);
+            // Remove package references from NodePackageDictionary
+            foreach (var key in NodePackageDictionary.Keys)
```
aparajit-pratap: How about using `ContainsKey` or `TryGetValue` on the dictionary instead of iterating over its keys. The former methods should be faster.

aparajit-pratap:
@scottmitchell it would have been good if you could have made separate PR's for each strategy you tried out so far. That way we could have compared them. Ah well :)

scottmitchell:
@aparajit-pratap Yeah, that's a good point. Definitely something I'll consider doing next time I have a PR with iterations like this.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1771,6 +1789,44 @@ private void RegisterHomeWorkspace(HomeWorkspaceModel newWorkspace)
...
+        private Dictionary<Guid, AssemblyName> OnCollectingAssembliesReferencedByNodes()
+        {
+            var assemblyNames = new Dictionary<Guid, AssemblyName>();
+            foreach(var node in CurrentWorkspace.Nodes)
+            {
+                // Get zerotouch assemblies
+                if (node.GetType() == typeof(DSFunction))
+                {
+                    // Split CreationName to get qualified name
+                    var qualifiedName = node.CreationName.Split('@')[0];
```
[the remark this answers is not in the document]
scottmitchell: Both the `CreationName` and `FunctionSignature` look like this: `HowickMaker.hMember.ByLineVector@Autodesk.DesignScript.Geometry.Line,Autodesk.DesignScript.Geometry.Vector,string`. But you need just the part on the left of the "@" (`HowickMaker.hMember.ByLineVector`) to look up the function descriptors. Is there an easier way to do this?

src/DynamoCore/Graph/Workspaces/PackageInfo.cs
scottmitchell: Using Version object now

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
scottmitchell: Added deserialization to this converter

src/DynamoPackages/PackageManagerExtension.cs
scottmitchell: `AssemblyName.Equals()` wasn't working as expected for dictionary keys here, so I just used `AssemblyName.FullName` which contains its Name and Version.

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
```diff
@@ -56,7 +56,53 @@ internal PackageDependencyInfo(string name, Version version)
...
+        /// </summary>
+        /// <param name="obj"></param>
+        /// <returns></returns>
+        public override bool Equals(object obj)
+        {
+            if (obj == null)
+            {
+                return false;
+            }
+            if (obj.GetType() != this.GetType())
```
aparajit-pratap: `if (!(obj is PackageDependencyInfo))`

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -576,25 +576,79 @@ internal IEnumerable<PackageDependencyInfo> PackageDependencies
...
+
+                return currentPackageDependencies;
+            }
+            set
+            {
+                packageDependencies = value;
             }
         }
 
+        private IEnumerable<PackageDependencyInfo> packageDependencies;
```
aparajit-pratap: You can make this a List

src/DynamoCore/Models/DynamoModel.cs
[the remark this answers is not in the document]
aparajit-pratap: Like Mike says I have a feeling this can be very slow. Further string operations make things slower so if there is a way to avoid string manipulation we should aim for that.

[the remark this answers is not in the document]
scottmitchell: @aparajit-pratap @mjkkirschner I was able to simplify this significantly: https://github.com/DynamoDS/Dynamo/pull/9723/commits/d6d35299fe5c1293f63c9781fe3580fad4808dae

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// <summary>
+        /// The full name of this package, including its name and version
+        /// Format: name_version
+        /// </summary>
+        internal string FullName { get { return Name + "_" + Version.ToString(); } }
+
+        /// <summary>
+        /// Guids of nodes in the workspace that are dependent on this package
+        /// </summary>
+        internal HashSet<Guid> Nodes
```
mjkkirschner: this seems kind of strange to me... how do we know which workspace these nodes refer to? Can you only retrieve or generate these objects from a workspace?

src/DynamoPackages/PackageManagerExtension.cs
mjkkirschner: 👍

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -436,6 +442,23 @@ protected virtual void OnSaving(XmlDocument obj)
             if (handler != null) handler(obj);
         }
 
+
+        /// <summary>
+        /// Event that is fired when the workspace is collecting top level assemblies referenced by nodes
```
mjkkirschner: Is it true that only one workspace at a time will ever handle these events and only one extension will ever invoke them?

If those invariants needs to hold for this to make sense - then I think we should watch for those and log or throw if either of those are not true.

IE - check the innovationList and make sure it's only one etc. - At a minimum - add a comment.

mjkkirschner:
@scottmitchell this is looking solid - I am still a bit concerned about serialization performance especially during backup save because the default backup time is pretty low at 60 seconds.

I think we need to test this with a large graph - it might make sense to add serialization to the performance CI tool in the future - not sure that needs to happen here - but at the minimum we need to know the impact for a large graph with a large set of packages.

scottmitchell:
@mjkkirschner I did a couple performance tests with and without package dependency collection/serialization. Looks like it makes serialization ~10% slower:
## Model.ToJson() time in ms

| Condition | master. 12 packages. 100 nodes | master. 12 packages. 500 nodes | DYN-1622. 12 packages. 100 nodes | DYN-1622. 12 packages. 500 nodes |
|-----------|--------------------------------|--------------------------------|----------------------------------|----------------------------------|
|           | 646                            | 2614                           | 572                              | 1824                             |
|           | 569                            | 1526                           | 565                              | 2305                             |
|           | 465                            | 1591                           | 549                              | 1866                             |
|           | 464                            | 2082                           | 578                              | 1895                             |
|           | 461                            | 1548                           | 546                              | 1924                             |
|           | 469                            | 1548                           | 556                              | 1873                             |
|           | 593                            | 1838                           | 598                              | 1928                             |
|           | 462                            | 1616                           | 563                              | 2037                             |
|           | 474                            | 1615                           | 567                              | 1909                             |
|           | 534                            | 1972                           | 746                              | 1991                             |
| Average   | 513.7                          | 1795                           | 584                              | 1955.2                           |
| New/Old   |                                |                                | 113.69%                          | 108.92%                          |

src/DynamoPackages/PackageManagerExtension.cs
```diff
@@ -120,7 +144,10 @@ public void Startup(StartupParams startupParams)
             LoadPackages(startupParams.Preferences, startupParams.PathManager);
         }
 
-        public void Ready(ReadyParams sp) { }
+        public void Ready(ReadyParams sp)
+        {
+            sp.CurrentWorkspaceChanged += OnCurrentWorkspaceChanged;
```
mjkkirschner: @scottmitchell do you ever unsubscribe this handler?

src/DynamoPackages/PackageManagerExtension.cs
```diff
@@ -49,6 +64,8 @@ public string UniqueId
         public void Dispose()
```
mjkkirschner: @scottmitchell as a sanity check - it might be good to make sure this dispose method gets called during a test.

src/DynamoPackages/PackageManagerExtension.cs
scottmitchell: Nope. Just added the unsubscribe

src/DynamoPackages/PackageManagerExtension.cs
scottmitchell: Yes, it appears this does get called during serialization tests

src/DynamoCore/Graph/Workspaces/PackageInfo.cs
scottmitchell: Removed

src/DynamoPackages/PackageManagerExtension.cs
```diff
@@ -199,7 +199,7 @@ private void OnCurrentWorkspaceChanged(IWorkspaceModel ws)
 
         private PackageDependencyInfo GetNodePackageFromAssemblyName(AssemblyName assemblyName)
         {
-            if (NodePackageDictionary.ContainsKey(assemblyName.FullName))
+            if (NodePackageDictionary!= null && NodePackageDictionary.ContainsKey(assemblyName.FullName))
```
mjkkirschner: does it also get initialized to an empty dictionary somewhere? Like in the constructor?

scottmitchell: Yeah, it get's initialized when the first package is loaded

mjkkirschner:
LGTM

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
```diff
@@ -800,6 +800,10 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
                 writer.WriteEndArray();
                 writer.WriteEndObject();
             }
+            else
+            {
+                logger.LogWarning("Unnsuccessful attempt to serialize a PackageDependencyInfo object.", Logging.WarningLevel.Moderate);
```
aparajit-pratap: Does this need to be localizable?

src/DynamoPackages/PackageManagerExtension.cs
```diff
@@ -233,6 +241,14 @@ private void OnPackageLoaded(Package package)
             // Add new custom nodes to CustomNodePackageDictionary
             foreach (var cn in package.LoadedCustomNodes)
             {
+                if (CustomNodePackageDictionary.ContainsKey(cn.FunctionId))
+                {
+                    OnMessageLogged(LogMessage.Info(
+                        string.Format("{0} contains the custom node {1}, which has already been loaded " +
+                        "by another package. This may cause inconsistent results when determining which " +
+                        "package instances of this custom node are dependent on.", package.Name, cn.Name)
+                        ));
```
aparajit-pratap: Same here: do these need to be localized?

mjkkirschner: we have in the past not localized these warnings sent to the console if they are intended for us rather than customers -thats my thinking anyway and there is precedent. What do you think?

aparajit-pratap: If that's the case, I think it's okay to leave it as is.

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
QilongTang: Good point, we should escape the localization of it

PR #9665, opened 2019-04-17
DYN1805 Validate the performance tracking system by using validated improvements Part II

reddyashish:
### Purpose

This PR is a copy of the one that Craig made: https://github.com/DynamoDS/Dynamo/pull/9620

Instead of removing the TraceUtils class from from Protocore.Lang namespace, we just obsolete it for now. 

Since this trace data is checked for in the Execute function, this part of the code is executed for all the graphs. A small percentage decrease or increase in the run times for those graphs, that take less absolute run times can be ignored. 

Performance results comparison with the base line data(for model tests):
result1
result2

### Declarations

Check these if you believe they are true

- [x] The code base is in a better state after this PR
- [x] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [x] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions), and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.

### Reviewers

@mjkkirschner @aparajit-pratap @QilongTang @saintentropy

PR #9728, opened 2019-05-21
Fixing the API break that was caused by this PR (https://github.com/DynamoDS/Dynamo/pull/9665)

reddyashish:
### Purpose

This PR is to address the API break that was introduced by removing the "TraceUtils.cs" file from the ProtoCore project. 

JIRA task: https://jira.autodesk.com/browse/DYN-1898

Refer to this change: https://github.com/DynamoDS/Dynamo/commit/b859c7ba73fa4d086ab08d9bd95812a784be2ce0#r33609626

### Declarations

### Reviewers

@aparajit-pratap

### DIFF
--- a/src/Engine/ProtoCore/Lang/CallSite.cs
+++ b/src/Engine/ProtoCore/Lang/CallSite.cs
@@ -1917,6 +1917,7 @@ namespace ProtoCore
             {
                 ret = PerformReturnTypeCoerce(finalFep, runtimeCore, ret);
             }
+
             return ret;
         }
 
--- a/src/Engine/ProtoCore/Lang/CallSite.cs
+++ b/src/Engine/ProtoCore/Lang/CallSite.cs
@@ -1,11 +1,3 @@
-using DynamoServices;
-using ProtoCore.DSASM;
-using ProtoCore.Exceptions;
-using ProtoCore.Lang;
-using ProtoCore.Lang.Replication;
-using ProtoCore.Properties;
-using ProtoCore.Runtime;
-using ProtoCore.Utils;
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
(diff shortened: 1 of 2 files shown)


aparajit-pratap:
@reddyashish what about the actual file itself? Is it still checked in into the repo? Why are the build checks failing?

reddyashish:
@aparajit-pratap The actual file was never deleted and verified it is present. The issue was that it was causing ambiguity in "CallSite.cs" as TraceUtils class is present in both ProtoCore.Lang and DynamoServices namespaces. Changed it to use the one from DynamoServices namespace. I will check on the builds again once they finish.

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -16,6 +8,13 @@
 using System.Runtime.Serialization;
 using System.Runtime.Serialization.Formatters.Soap;
 using System.Text;
+using ProtoCore.DSASM;
+using ProtoCore.Exceptions;
+using ProtoCore.Lang;
```
aparajit-pratap: You can just remove this `using` as it's referencing the `TraceUtils` class. Then you won't need to prefix all usages of `TraceUtils` with `DynamoServices` namespace.

reddyashish: There are other classes that are being used(in Callsite.cs) from the ProtoCore.Lang namespace and so couldn't remove this import.

aparajit-pratap:
@reddyashish I think this is LGTM as long as you've run all tests to make sure that there are no regressions just in case. Any idea why the appveyor and travis builds are failing on the PR?

reddyashish:
Yes, I have run all the tests and there are no regressions. Not sure why the appveyor and travis builds failed. In the logs, it just shows that the build was cancelled by system administrator. Triggered them again.

reddyashish:
Merging this.

Issue #9733, opened 2019-05-23
Dynamo Double Compare Inconsistency


RyaPorter:
## Dynamo version
2.3.0.5074 (master branch)

## Operating system
Windows 10

## What did you do? 
Compared to doubles (4.701312264353, 4.70131302602723)

## What did you expect to see?
The comparison 4.701312264353 < 4.70131302602723 to evaluate as True.

## What did you see instead?
It evaluated as false:
![Dynamo_CompareError](https://user-images.githubusercontent.com/46380195/58257398-c2d01b80-7d68-11e9-96a9-f5fed28411da.png)

I'm pretty sure I'm not missing something obviously stupid. But Python confirms that something fundamental is wrong with, I'm guessing, DesignScript. (Again if I'm not missing something obviously stupid).

RyaPorter:
It seems very inconsistent. Returning different results depending on the amount of decimal points and significant figures:

![Dynamo_CompareError_the_plot_thickens](https://user-images.githubusercontent.com/46380195/58264987-bfdc2780-7d76-11e9-9736-d1412f3b256d.png)

![Dynamo_CompareError_SigFigSmall](https://user-images.githubusercontent.com/46380195/58264995-c10d5480-7d76-11e9-870b-3c2c82d104ae.png)

![Dynamo_CompareError_the_plot_thickens_again](https://user-images.githubusercontent.com/46380195/58265218-282b0900-7d77-11e9-90c6-4a185ea0a75e.png)

mjkkirschner:
this is partially resolved. Equality still remains tolerant, discussing solutions at this point.

mjkkirschner:
This should now be resolved I think as a new tolerant equals node has been added where tolerance is specified.

PR #9735, opened 2019-05-23
Fix comparison operators for double values in DS

aparajit-pratap:
### Purpose

Fixes equality and comparison operators in DesignScript for double type.
Github issue: https://github.com/DynamoDS/Dynamo/issues/9733

Main points addressed in this PR:
- DS comparison operators `<`, `>`, `<=`, `>=` use the underlying platform C# operators 
- Equality operators `==` and `!=` are left unchanged and continue to use a DS level tolerance of 1e-5 ni order to prevent breaking changes

### Declarations

### Reviewers

@mjkkirschner 

### FYIs

@DynamoDS/dynamo


src/Engine/ProtoCore/DSASM/Stack.cs
```diff
@@ -262,7 +262,7 @@ public static bool CompareStackValues(StackValue sv1, StackValue sv2, RuntimeCor
 
                     if(Double.IsInfinity(value1) && Double.IsInfinity(value2))
                         return true;
-                    return MathUtils.Equals(value1, value2);
+                    return sv1.Equals(sv2);
```
aparajit-pratap: I've updated this to use the `Equals` method override for `StackValue`.

mjkkirschner:
hmm, I'm not sure why this ever had a tolerance like this? 
I guess this better aligns to other OOTB double equality comparison in python, c# etc?

We should definitely consider though what this might break - it seems to me that it's worth it though because some of these results very confusing to users who use a > 5 digits of precision - this doesn't seem like a very uncommon use case to me regarding optimization or when dealing with parameter spaces over a curve etc.

mjkkirschner:
@jnealb @saintentropy @Dewb FYI

aparajit-pratap:
Something simple like this fails now after this change:
image

jnealb:
Jira task for this? We need a math epic/story to centralize the issues.

aparajit-pratap:
For the time being I have adjusted the tolerance to make it smaller so that floating point comparisons will be possible to higher decimal places in DS however I feel we should make it settable by the user via a UI to give more flexibility and control. Thoughts?

mjkkirschner:
@aparajit-pratap - have you considered adding an overload like: `Double.Equals(other, epsilon)`? - We could add this method and leave the current implementation as is to avoid changing current behavior - and then document the current behavior. 

- It does feel odd to have an arbitrary tolerance when considering large scale projects (kilometers) or non geometrical comparisons - fitness functions etc - I am curios what @RyaPorter was trying to do when he they ran into this issue?

ColinDayOrg:
One thing to possibly consider; one of the intentions of the Dynamo Primitives package was to be able to "grid" up a double space (in 1D, 2D, and 3D), and once the space is gridded then integers can be used instead of doubles which gets rid of the need for epsilons. Using epsilons is necessary in double comparisons, but will inevitably lead to tolerance creep. Just something to consider if it could be useful.

aparajit-pratap:
> @aparajit-pratap - have you considered adding an overload like: `Double.Equals(other, epsilon)`? - We could add this method and leave the current implementation as is to avoid changing current behavior - and then document the current behavior.
> 
@mjkkirschner how will this overload be used? How is `Double.Equals(other, epsilon)` different or advantageous over `Math.Abs(lhs - rhs) < Tolerance`? Will `epsilon` be user input or the standard `Double.Epsilon`?

aparajit-pratap:
@gregmarr do you have any inputs on what should be the best way to implement equality comparisons for floating point values in languages?

aparajit-pratap:
Hmm, I see this is an issue in Thunderstorm as well:
image

Dewb:
A useful reference for anyone who isn't already familiar with it:
http://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html
This might also be relevant:
https://possiblywrong.wordpress.com/2013/11/15/floating-point-equality-its-worse-than-you-think/

Doing string equality tests on IEEE floating point is pretty much never what the user intends. The `Equals` node in the Dynamo environment ought to have an epsilon clearly communicated to the user, whether it's configurable or not. The DS `==` operator is trickier, I think it would be reasonable for it to behave in the same way as the underlying implementation language's `==` operator does, and users should avoid using it for floating-point numbers, preferring a separate library function with an epsilon value.

But the complaint in #9733 is about the `<` operator, which *should* have stable, correct results for doubles. How does this involve a boolean comparison function that uses `==` instead of the -1/0/+1 comparison test function?

mjkkirschner:
@Dewb in this case it appears the `<` and `>` operators are checking for `!equals()` as well - I guess to implement `=< `or `>=`

Dewb:
@mjkkirschner where is that in the code?

Dewb:
Ah, I see: 
https://github.com/DynamoDS/Dynamo/blob/7b3642fd745871e858183e4aaf27bfe51312322b/src/Engine/ProtoCore/Utils/MathUtils.cs#L9

https://github.com/DynamoDS/Dynamo/blob/fa31e0eee2dd50b3a4a738d94ee96e3849a2ab8b/src/Engine/ProtoCore/Lang/BuiltInFunctionEndPoint.cs#L2601

I would remote the `Equals` call from all of these methods and implement them in terms of `Double.CompareTo(Double)` or `Int.CompareTo(Int)` as appropriate (with casting if the arguments aren't the same numeric type.) 

In the `BuiltInFunctionEndPoint.cs` case, it might also be a good idea to consider removing the flexible ascending/descending comparison flag in the constructor, as it doesn't appear to be used anywhere and complicates the implementation. 

In C-like languages, the convention is that a method named "Compare" called on `a` and `b` returns <0 if `a < b`, 0 if `a == b`, and >0 if `a > b`. Any DS VM methods named Compare probably ought to follow this convention, without flags to reverse the behavior. Similarly, any "Compare" methods that return `bool` ought to be renamed in terms of `Equals` or `IsEqual`, etc.

Dewb:
> Hmm, I see this is an issue in Thunderstorm as well:
> image

@aparajit-pratap This is the behavior that I would expect in an environment that uses IEEE floats. You'll observe the same results in C#, Javascript, Go, and Python.

aparajit-pratap:
> 
> @aparajit-pratap This is the behavior that I would expect in an environment that uses IEEE floats. You'll observe the same results in C#, Javascript, Go, and Python.

@Dewb so in that case are you suggesting that it's ok to implement the Equals method in https://github.com/DynamoDS/Dynamo/blob/351c88b58c89aadbc66309aeeb7ab08a7ac5fc34/src/Engine/ProtoCore/Utils/MathUtils.cs#L29 as `lhs.Equals(rhs)`?

Dewb:
No, any equality method that uses a tolerance value should be implemented that way (comparing the absolute value of the difference to the tolerance value.)

I'm saying that any methods implementing less than or greater than should use either the built-in operators or the built-in `CompareTo` method, and should not call built-in or custom `Equals` methods, use the `==` operator, or check a tolerance.

aparajit-pratap:
> No, any floating-point equality method that uses a tolerance value should be implemented that way (comparing the absolute value of the difference to the tolerance value.)
> 
@Dewb I got your point on removing the Equals check from the `<` and `>` operator implementations. I was specifically asking about the `==` operator implementation. 

From your comment about languages like C# I understood that for the equality check we could just use the underlying equality of the language we're implementing with (C# in this case) and we could leave it up to users to do "comparing the absolute value of the difference to the tolerance value" for floating point values in their graphs themselves. On the other hand if we wish to continue to have the tolerance check implicit in the node, the question is on what basis do we select the tolerance. The current tolerance value of `1e-5` seems too high.

Also in TS the equality operator simply uses the equality check of the language (Go): https://git.autodesk.com/aecgd/thustrm/blob/c015d689975b9e9a99711ca2f955eb82962d8c93/intrinsics/intrinsics.go#L845

aparajit-pratap:
It looks like the only difference between Double.CompareTo vs. the builtin `<` and `>` operators is the added support for 0.0d vs -0.0d and Nan numbers: https://www.tutorialspoint.com/java/lang/double_compareto.htm

aparajit-pratap:
I have left the `==` operator implementation untouched for now (at the cost of Dynamo vs. TS behavior still being different) and fixed the comparison operators as per Mike D's suggestions. This should address #9733 mostly. Please give it one more look.

src/Engine/ProtoCore/DSASM/Stack.cs
```diff
@@ -260,8 +260,6 @@ public static bool CompareStackValues(StackValue sv1, StackValue sv2, RuntimeCor
                     var value1 = sv1.DoubleValue;
                     var value2 = sv2.DoubleValue;
 
-                    if(Double.IsInfinity(value1) && Double.IsInfinity(value2))
```
mjkkirschner: was this not required?

mjkkirschner: ah I see moved to math utils

mjkkirschner:
@aparajit-pratap lets just change the name of this PR (since it address comparison operators only) and then LGTM.

mjkkirschner:
one more request - is there a test which checks the operators work for higher precision doubles (over e^-5)?

Dewb:
> It looks like the only difference between Double.CompareTo vs. the builtin `<` and `>` operators is the added support for 0.0d vs -0.0d and Nan numbers: https://www.tutorialspoint.com/java/lang/double_compareto.htm

Sorry, I didn't mean to imply that we should prefer CompareTo over operators, but that we should use CompareTo when we wanted the <0 / 0 / >0 combined comparison operation.

aparajit-pratap:
I have now removed any custom implementation for comparison operators except for `==`. It still seems odd to me that we now have a custom implementation for only `MathUtils.Equals` that uses a tolerance. So now both these statements are true, which is confusing:
```
a = 0.3;
b = 0.2;
c = a - b < 0.1;    // c is true
d = a - b == 0.1; // d is true
```
I would therefore propose that we do not have a custom implementation for `==` operator in DS either but simply use the underlying `==` available in C#. This will also make this behavior consistent with Thunderstorm. This will encourage users to do equality checks by supplying their own tolerance and write DS code like:
```
c = Math.Abs(a - b) < user_tolerance;
```
In addition we could add a new `Equals` node that takes 3 inputs - lhs (double), rhs (double), and tolerance. I think this proposal also aligns with @Dewb's thoughts, am I right?

src/Engine/ProtoCore/Lang/BuiltInFunctionEndPoint.cs
```diff
@@ -2581,24 +2581,9 @@ public StackValueComparerForDouble(bool ascending)
...
-            var v1 = sv1.IsDouble ? sv1.DoubleValue: sv1.IntegerValue;
-            var v2 = sv2.IsDouble ? sv2.DoubleValue: sv2.IntegerValue;
-
-            return MathUtils.Equals(v1, v2);
-        }
-
         public int Compare(StackValue sv1, StackValue sv2)
         {
-            if (Equals(sv1, sv2))
+            if (!sv1.IsNumeric && !sv2.IsNumeric)
```
mjkkirschner: what does this do? It looks like any items that are not numeric are evaluated as equal?

aparajit-pratap: This check was being made in the previous `Equals` method that has been removed.

src/Engine/ProtoCore/Utils/MathUtils.cs
```diff
@@ -4,30 +4,16 @@ namespace ProtoCore.Utils
 {
     public static class MathUtils
     {
-        public static double Tolerance = 1e-5; 
+        public static double Tolerance = 1e-5;
 
-        public static bool IsLessThan(double lhs, double rhs)
```
mjkkirschner: @aparajit-pratap unfortunately I think we will only be able to obsolete these, not sure we can remove them now.

aparajit-pratap: Of course. Will make the necessary changes.

mjkkirschner:
so to sum up the changes so far:
* comparison is now done using the underlying platform comparison
* == still uses a tolerance to avoid breaking changes

your proposal in the last comment sounds reasonable, but like we talked about earlier I'm not sure how gauge the numbers of graphs this would break. I'm still leaning towards adding the `Equals with tolerance method` and changing the existing behavior at a major version change.

Can we think of some way of estimating if the impact of this change will break many existing graphs?

aparajit-pratap:
@mjkkirschner yes, you have summarized these changes correctly. 
Yes, we can add a new "Equals" node with tolerance and keep the existing behavior of `==` operator in order to prevent breaking graphs until the next major release. However, note there are changes here that also could break existing behavior of `<`, `>`, etc. operators. Something like this was at least consistent before this change but is now inconsistent:
```
a = 0.3;
b = 0.2;
c = a - b < 0.1;    // false before, true now
d = a - b == 0.1; // true 
```
I would therefore recommend either deferring this whole PR for later or adding the change for equality as well, which is to simply use the platform `==`. I'll schedule a meeting for us to discuss next week.

src/Engine/ProtoCore/DSASM/Executive.cs
```diff
@@ -3902,7 +3902,7 @@ private void GE_Handler(Instruction instruction)
                 {
                     double lhs = opdata2.IsDouble ? opdata2.DoubleValue : opdata2.IntegerValue;
                     double rhs = opdata1.IsDouble ? opdata1.DoubleValue : opdata1.IntegerValue;
-                    opdata2 = StackValue.BuildBoolean(MathUtils.IsGreaterThanOrEquals(lhs, rhs));
+                    opdata2 = StackValue.BuildBoolean(lhs >= rhs);
```
gregmarr: I'm concerned that this loses the toleranced comparison.  If lhs was less than rhs but within the floating point tolerance, then this will change from true to false.

src/Engine/ProtoCore/DSASM/Executive.cs
```diff
@@ -3880,7 +3880,7 @@ private void LT_Handler(Instruction instruction)
             {
                 double value1 = opdata2.IsDouble ? opdata2.DoubleValue: opdata2.IntegerValue;
                 double value2 = opdata1.IsDouble ? opdata1.DoubleValue: opdata1.IntegerValue;
-                opdata2 = StackValue.BuildBoolean(MathUtils.IsLessThan(value1, value2));
+                opdata2 = StackValue.BuildBoolean(value1 < value2);
```
gregmarr: I'm concerned that this loses the toleranced comparison.  If value1 was less than value2 but within the floating point tolerance, then this will change from false to true.  Example: 0.1 and 0.100000001 would be false with tolerance, and true without.  If

src/Engine/ProtoCore/DSASM/Executive.cs
```diff
@@ -3929,7 +3929,7 @@ private void LE_Handler(Instruction instruction)
                 {
                     double lhs = opdata2.IsDouble ? opdata2.DoubleValue: opdata2.IntegerValue;
                     double rhs = opdata1.IsDouble ? opdata1.DoubleValue: opdata1.IntegerValue;
-                    opdata2 = StackValue.BuildBoolean(MathUtils.IsLessThanOrEquals(lhs, rhs));
+                    opdata2 = StackValue.BuildBoolean(lhs <= rhs);
```
gregmarr: as above

src/Engine/ProtoCore/DSASM/Executive.cs
gregmarr: Okay, just got to read the comments on the PR itself, and see why this is being done.

src/Engine/ProtoCore/DSASM/Executive.cs
gregmarr: Okay, just got to read the comments on the PR itself, and see why this is being done.

src/Engine/ProtoCore/DSASM/Executive.cs
gregmarr: Okay, just got to read the comments on the PR itself, and see why this is being done.

mjkkirschner:
@gregmarr we have a meeting tomorrow on this to discuss pros/cons / breaking changes - do you want to attend?

PR #9736, opened 2019-05-24
Address loading of packages with inter-dependencies 

aparajit-pratap:
### Purpose
This PR addresses the following issues (JIRA Link: https://jira.autodesk.com/browse/DYN-1863):

1. If there's a package A that depends on package B, and A does not come bundled with B, package B needs to be installed (from the PM) or loaded into Dynamo (imported in the VM) before A for A to be loaded successfully. If not, only B is loaded.
- [x] If B is already installed but not imported before A, resolve the dependencies from B that are required while importing A by searching for them in the package paths. 

- [x] If package B is not already installed there is not much that can be done unless the user installs B manually. Once B is loaded/installed, we need to restart Dynamo in order to reload A.

2. If package A depends on B and also contains B's assembly, then A is imported successfully. In this case B will not be installed (nor imported). If B is installed from the PM separately, it will fail as the VM will throw a class redefinition error.

- [x] Still making sure that B is installed properly. 

- [x] Now if B is labelled as "node library" in pkg.json manifest file it will load along with A.

3. Package load failures should now show up in notifications manager as shown:

image

screenshot2

**Performance profiling results** for loading 6 packages totaling 15,000 nodes during Dynamo startup **(savings of ~10sec out of overall time of ~66sec)**:
**Before:**
image

**After:**
image

### Declarations

### Reviewers

@mjkkirschner 

### FYIs


src/DynamoCore/Library/LibraryServices.cs
```diff
@@ -111,6 +113,35 @@ public LibraryServices(ProtoCore.Core libraryManagementCore, IPathManager pathMa
             PopulateOperators();
             PopulatePreloadLibraries();
             LibraryLoadFailed += new EventHandler<LibraryLoadFailedEventArgs>(LibraryLoadFailureHandler);
+
+            //AppDomain.CurrentDomain.AssemblyResolve += ResolveAssembly;
```
aparajit-pratap: @mjkkirschner this is the handler I was trying to register here but the event never reaches here as it encounters an exception in the other handler that is registered before this one: https://github.com/DynamoDS/Dynamo/blob/bd9ae7cb3c46ed61070daf823c81a69f6e7fa411/src/DynamoSandbox/Program.cs#L41

src/DynamoCore/Library/LibraryServices.cs
```diff
@@ -111,6 +113,34 @@ public LibraryServices(ProtoCore.Core libraryManagementCore, IPathManager pathMa
...
+            var assemblyName = new AssemblyName(args.Name).Name + ".dll";
+            var assemblyPath = assemblyName;
+            try
+            {
+                if (pathManager.ResolveLibraryPath(ref assemblyPath))
+                {
+                    return Assembly.LoadFrom(assemblyPath);
+                }
+
+                foreach (var packagesDirectory in pathManager.PackagesDirectories)
```
mjkkirschner: seems odd we need to do this - aren't the package directory /bin folders added to the assembly resolution paths?

on that note - why do we need a new assembly resolve handler?

aparajit-pratap: I don't think package directories are added to assembly resolution paths. Therefore the need to do this here.

mjkkirschner: @aparajit-pratap what about right here:
https://github.com/DynamoDS/Dynamo/blob/f6f43e2a56655181bad6195a6bd0403a8e1b1a86/src/DynamoPackages/PackageLoader.cs#L222 ?

aparajit-pratap: So it looks like the existing assembly resolvers do not use these resolution paths.

src/DynamoCore/Extensions/ExtensionLibraryLoader.cs
```diff
@@ -28,5 +31,11 @@ public void LoadNodeLibrary(Assembly library)
         {
             model.LoadNodeLibrary(library);
         }
+
+        public void LoadPackages(IEnumerable<Assembly> assemblies)
```
mjkkirschner: since this is public it should have a summary please.

src/DynamoCore/Library/ILibraryLoader.cs
```diff
@@ -11,5 +12,7 @@ public interface ILibraryLoader
         /// Loads node's library
         /// </summary>
         void LoadNodeLibrary(Assembly library);
+
+        void LoadPackages(IEnumerable<Assembly> assemblies);
```
mjkkirschner: we can't modify an interface like this. API break.

src/DynamoCore/Logging/DynamoLogger.cs
```diff
@@ -139,6 +141,8 @@ public string LogText
             }
         }
 
+        internal HashSet<NotificationMessage> Notifications { get; }
```
mjkkirschner: this definitely seems weird - so it's a cache of notifications and only useful if the notifications extension is not loaded?

src/DynamoCore/Extensions/ReadyParams.cs
```diff
@@ -63,6 +64,11 @@ public IWorkspaceModel CurrentWorkspaceModel
             }
         }
 
+        public IEnumerable<NotificationMessage> Notifications
```
mjkkirschner: seems weird that this is sent to all extensions...

mjkkirschner: can we talk about alternative solutions?

src/DynamoCore/Library/LibraryServices.cs
```diff
@@ -579,7 +612,7 @@ internal bool ImportLibrary(string library, bool isExplicitlyImportedLib = false
                     foreach (ErrorEntry error in LibraryManagementCore.BuildStatus.Errors)
                     {
                         Log(error.Message, WarningLevel.Moderate);
-                        errorMessage += error.Message + "\n";
+                        errorMessage += "\n" + error.Message;
```
mjkkirschner: lets use `enviornment.newLine` ?

src/DynamoCore/Logging/DynamoLogger.cs
aparajit-pratap: Yes.

src/DynamoCore/Extensions/ReadyParams.cs
aparajit-pratap: I'm all ears. This is one way I could think of since the Notification extension needs to be linked to the DynamoLogger some how.

src/DynamoCore/Library/ILibraryLoader.cs
aparajit-pratap: 🤦‍♂

src/DynamoCore/Library/LibraryServices.cs
```diff
@@ -579,7 +598,7 @@ internal bool ImportLibrary(string library, bool isExplicitlyImportedLib = false
                     foreach (ErrorEntry error in LibraryManagementCore.BuildStatus.Errors)
                     {
                         Log(error.Message, WarningLevel.Moderate);
-                        errorMessage += error.Message + "\n";
+                        errorMessage += Environment.NewLine + error.Message;
```
mjkkirschner: did you want the new line first?

src/DynamoCore/Library/LibraryServices.cs
```diff
@@ -1133,12 +1155,23 @@ public LibraryLoadFailedEventArgs(string libraryPath, string reason, bool throwO
 
         public class LibraryLoadedEventArgs : EventArgs
         {
+            // TODO: Remove in 3.0
+            [Obsolete]
```
mjkkirschner: can you add a message why this is obsolete?

src/DynamoCore/Library/LibraryServices.cs
```diff
@@ -1133,12 +1155,23 @@ public LibraryLoadFailedEventArgs(string libraryPath, string reason, bool throwO
...
             }
 
-            public string LibraryPath { get; private set; }
+            public LibraryLoadedEventArgs(IEnumerable<string> libraryPaths)
+            {
+                LibraryPaths = libraryPaths;
+            }
+
+            // TODO: Remove in 3.0
+            [Obsolete]
```
mjkkirschner: same here - I will definitely forget.

src/DynamoCore/Library/LibraryServices.cs
aparajit-pratap: Yes.

src/DynamoCore/Library/LibraryServices.cs
```diff
@@ -1133,12 +1155,23 @@ public LibraryLoadFailedEventArgs(string libraryPath, string reason, bool throwO
...
+            public LibraryLoadedEventArgs(IEnumerable<string> libraryPaths)
+            {
+                LibraryPaths = libraryPaths;
+            }
+
+            // TODO: Remove in 3.0
+            [Obsolete("This property is obsolete. Use property LibraryPaths that returns collection of paths instead.")]
+            public string LibraryPath { get; }
+
+            public IEnumerable<string> LibraryPaths { get; }
```
mjkkirschner: summary tag on this guy please.

src/Notifications/NotificationsViewExtension.cs
```diff
@@ -51,15 +56,19 @@ public void Loaded(ViewLoadedParams p)
         {
             viewLoadedParams = p;
             dynamoWindow = p.DynamoWindow;
+            var viewModel = dynamoWindow.DataContext as DynamoViewModel;
```
mjkkirschner: ugh, wish we could avoid this. What about adding the logger to the set of objects we pass into the extensions and make available?

aparajit-pratap: We pass stuff into view extensions via `ViewLoadedParams` or `ViewStartParams` and their base classes and I thought we agreed it isn't a good idea to add any new property to these classes so i don't see how I can make the logger available to this extension, you?

QilongTang: I feel we need to agree on a strategy since people are going to read our code and all implement similar way than us. This part is impacting my PR as well

mjkkirschner: What I said is that we should not add things to the API which are not useful to all extensions. Logger seems more generally useful - but like you said, notifications can already log messages - though it's very hard to discover.

aparajit-pratap: I think the design originally conceived for enabling logging from extensions was to inherit from/implement `ILogSource` but unfortunately `NotificationViewExtension` does not implement `ILogSource`. So for those extension that already do, giving access to the logger may be redundant.

test/Libraries/PackageManagerTests/PackageLoaderTests.cs
```diff
@@ -110,13 +110,31 @@ public void ScanPackageDirectoryReturnsNullForInvalidDirectory()
...
+            var libs = CurrentDynamoModel.LibraryServices.ImportedLibraries.ToList();
+            Assert.IsTrue(libs.Any(x => File.Exists(Path.Combine(PackagesDirectory, "AnotherPackage", "bin", "AnotherPackage.dll"))));
+            Assert.IsTrue(libs.Any(x => File.Exists(Path.Combine(PackagesDirectory, "Dependent Package", "bin", "DependentPackage.dll"))));
+            Assert.IsTrue(libs.Any(x => File.Exists(Path.Combine(PackagesDirectory, "Package", "bin", "Package.dll"))));
+
+            // Verify that interdependent packages are imported successfully
+            var entries = CurrentDynamoModel.SearchModel.SearchEntries.ToList();
+            Assert.IsTrue(entries.Any(x => x.FullName == "AnotherPackage.AnotherPackage.AnotherPackage.HelloAnotherWorld"));
+            Assert.IsTrue(entries.Any(x => x.FullName == "DependentPackage.DependentPackage.DependentPackage.HelloWorld"));
+            Assert.IsTrue(entries.Any(x => x.FullName == "Package.Package.Package.Hello"));
```
aparajit-pratap: The following code shows the methods defined in the 3 test packages. `DependentPackage` has a dependency on `Package` and `AnotherPackage` has a dependency on the first two:
image

test/DynamoCoreWpfTests/PublishPackageViewModelTests.cs
```diff
@@ -47,7 +47,7 @@ public void SetsErrorState()
             //open a dyf file and modify it
             string packagedirectory = Path.Combine(TestDirectory, "pkgs");
             var packages = Directory.EnumerateDirectories(packagedirectory);
-            var first = Path.GetFullPath(packages.First());
+            var first = Path.GetFullPath(packages.FirstOrDefault(x => Path.GetFileName(x).Equals("Custom Rounding")));
```
aparajit-pratap: These tests specifically check for the "Custom Rounding" test package that earlier happened to be the first package in the list. This order changed in this PR after I added new test packages, which is why I'm specifically searching for the "Custom Rounding" package from now on.

test/DynamoCoreTests/DynamoModelTestBase.cs
```diff
@@ -37,6 +37,7 @@ public override void Setup()
             StartDynamo(dynamoSettings);
         }
 
+        [TearDown]
```
mjkkirschner: did you find you need this for some reason? I think it's marked as teardown in the base class - but nunit / inheritance behavior always requires debugging for me to understand.

aparajit-pratap: I guess it doesn't matter. I think you're right; if it's overriding the base class method and it has the attribute in the base class, it isn't needed. No harm in having it here as well though.

aparajit-pratap: I think I'll remove it on second thoughts. One less file change in this PR :)

mjkkirschner:
LGTM!

PR #9738, opened 2019-05-24
profiling tweaks

mjkkirschner:
### Purpose

a few small issues found while using the profiling APIs

* tried to dispose Disposables when they are going out of scope
* use a unique id for the particular callback because there is the possibility someone will add a databridge callback with the id of the node only and one will be overwritten by the other - this helps decrease the risk of this happening by combining the node guid with some other string.

I have been testing this with this small extension:
https://github.com/mjkkirschner/DynamoSamples/tree/proflingExtension

Screen Shot 2019-05-24 at 2 02 54 PM

### Declarations

### Reviewers

@ColinDayOrg 

### FYIs

@smangarole

### DIFF
--- a/src/DynamoCore/Engine/EngineController.cs
+++ b/src/DynamoCore/Engine/EngineController.cs
@@ -182,6 +182,7 @@ namespace Dynamo.Engine
             }
             else
             {
+                astBuilder.ProfilingSession.Dispose();
                 astBuilder.ProfilingSession = null;
             }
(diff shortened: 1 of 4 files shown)


mjkkirschner:
@smangarole I'd like to add a test before I merge this - can we collaborate on something?

mjkkirschner:
going to merge this so @smangarole can test using it.

PR #9773, opened 2019-06-05
DYN-1662 Follow Up: Return copy of PackageDependencyInfo instead of reference

scottmitchell:
### Purpose

This PR follows up on https://github.com/DynamoDS/Dynamo/pull/9723. It fixes an issue where the IDs of deleted nodes would still be serialized inside of `PackageDependencies`. 

### Declarations

Check these if you believe they are true

- [ ] The code base is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions), and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.

### Reviewers

@mjkkirschner

PR #9784, opened 2019-06-09
DYN-1914: Add tests for package dependency collection, serialization, and deserialization

scottmitchell:
### Purpose

This PR addresses JIRA issue [DYN-1914: Add tests for package dependency collection, serialization, and deserialization](https://jira.autodesk.com/browse/DYN-1914). A new test class `PackageDependencyTests` has been added. This PR does the following:

- [x] Add test coverage for package dependency
  - [x] `ZeroTouchPackageDependencyIsCollectedAndSerialized`
  - [x] `NodeModelPackageDependencyIsCollected`
  - [x] `CustomNodePackageDependencyIsCollected`
  - [x] `PackageDependencyIsCollectedForNewWorkspace` -- fails without the [fix contained in this PR](https://github.com/DynamoDS/Dynamo/pull/9784/files#diff-fa66cb1f9f2a8513bae9e5b00ff2666cR155)
  - [x] `PackageDependenciesUpdatedAfterNodeAdded`
  - [x] `PackageDependenciesUpdatedAfterNodeRemoved` -- fails without the fix at https://github.com/DynamoDS/Dynamo/pull/9773
  - [x] `PackageDependenciesUpdatedAfterPackageAndNodeAdded`
  - [x] `PackageDependenciesPreservedWhenPackagesNotLoaded`
  - [ ] ~~`PackageDependenciesPreservedAfterPackageRemoved`~~

- [x] Address a bug found by @QilongTang where package dependency collection failed for some graphs
- [x] Override `GetHashCode` for `PackageDependencyInfo` type

### Declarations

### Reviewers

@mjkkirschner @QilongTang


src/DynamoPackages/PackageManagerExtension.cs
```diff
@@ -151,6 +151,10 @@ public void Ready(ReadyParams sp)
         {
             ReadyParams = sp;
             sp.CurrentWorkspaceChanged += OnCurrentWorkspaceChanged;
+
+            (sp.CurrentWorkspaceModel as WorkspaceModel).CollectingCustomNodePackageDependencies += GetCustomNodePackageFromID;
```
QilongTang: @scottmitchell Does this mean newly installed custom node package will be added as well? Maybe not?

scottmitchell: I think that happens here: https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoPackages/PackageManagerExtension.cs#L255-L269

mjkkirschner:
@scottmitchell what needs to be discussed on the last test?

scottmitchell:
@mjkkirschner I wanted to get some more info about when/if package removal happens within a session. When you uninstall a package, it looks like the package isn’t actually removed until you close Dynamo? If that’s true, then my original intention for the test (to remove a package within a session) doesn’t make sense. I’ll join standup tomorrow.

mjkkirschner:
That’s true for binary packages but I think custom node only packages can be removed.

mjkkirschner:
@scottmitchell does this need a review?

mjkkirschner:
@scottmitchell where is the code for your testPackage.dll?

scottmitchell:
@mjkkirschner Yes, it's ready for review. I think we should get this initial package dependency work wrapped up, and I think this gets us good test coverage to start. 

ZTTestPackage is here: https://github.com/scottmitchell/ZTTestPackage

mjkkirschner:
@scottmitchell can you guarantee that this test code will never disappear and we'll always have access to it? I guess in the worst case we can decompile the dll - I think it's a bit cumbersome to have this in another repo - let's improve this in a followup.

test/DynamoCoreTests/PackageDependencyTests.cs
```diff
@@ -0,0 +1,299 @@
...
+            Assert.AreEqual(1, package.Nodes.Count);
+        }
+
+        [Test]
+        public void PackageDependencyIsCollectedForNewWorkspace()
+        {
+            // Add one node from "Dynamo Samples" package
+            var pi = GetPackageInfo("Dynamo Samples");
+            var node = GetNodeInstance("Examples.NEWBasicExample.Create@double,double,double");
+            CurrentDynamoModel.AddNodeToCurrentWorkspace(node, true);
```
mjkkirschner: @scottmitchell how is this test different from the above test? Did you mean to add a new workspace?

mjkkirschner:
@scottmitchell please run this on the self serve - because you are adding a new package - it will likely break some other tests which count the number of packages loaded from the package folder.

scottmitchell:
@mjkkirschner everything looks good on the self serve. I didn’t put this package with the other test packages because I didn’t want it to load by default. It’s in the new “packagedDependencyTests” folder. I can replace this package with a new package in the Dynamo repo if we want. I just needed a very simple package that wasn’t already included in the default test packages. Functionality of the package does not matter for these tests.

test/DynamoCoreTests/PackageDependencyTests.cs
```diff
@@ -0,0 +1,299 @@
...
+
+            // There should now be 2 package dependencies and 3 total dependent nodes
+            packageDependencies = CurrentDynamoModel.CurrentWorkspace.PackageDependencies;
+            Assert.AreEqual(2, packageDependencies.Count);
+            foreach(var package in packageDependencies)
+            {
+                if (package.Equals(package1))
+                {
+                    // Package 1 should have two nodes
+                    Assert.AreEqual(2, package.Nodes.Count);
```
QilongTang: 👍

QilongTang:
These tests LGTM

test/DynamoCoreTests/PackageDependencyTests.cs
```diff
@@ -0,0 +1,299 @@
...
+
+        protected override void GetLibrariesToPreload(List<string> libraries)
+        {
+            libraries.Add("ProtoGeometry.dll");
+            libraries.Add("DesignScriptBuiltin.dll");
+            libraries.Add("DSCoreNodes.dll");
+            base.GetLibrariesToPreload(libraries);
+        }
+
+        private PackageLoader GetPackageLoader()
```
aparajit-pratap: This function already exists in `PackageLoaderTests`: https://github.com/DynamoDS/Dynamo/blob/55482f600fdf4bbf5cf0ae4802df41570a22b04a/test/Libraries/PackageManagerTests/PackageLoaderTests.cs#L235. You could have moved this existing method to the common base class I suppose.

PR #9786, opened 2019-06-10
DYN-1897 API to add extensions to right side bar. 

reddyashish:
### Purpose

This PR provides an internal API to add an extension to the right side extensions bar. This API can be called from inside your View Extension class. Inside the "Loaded" function, we can call this method and the window will be added to the right side bar. Tab control mechanism is implemented here. 
 For example: p.AddToExtensionsSideBar(window);

In the below GIF, I have changed the sample extension class to call this above method instead of calling the window.show() function. 

right side bar

Looking forward to any suggestions. 

### Declarations

### Reviewers

@QilongTang @mjkkirschner @aparajit-pratap @ColinDayOrg @scottmitchell


QilongTang:
Great to see progress, do you plan to make a PR to the DynamoSamples repo as well?

reddyashish:
Yes, I will make a PR for that.

mjkkirschner:
@smangarole will this break any UI automation tests?
@reddyashish If you decide to the DynamoSamples (which I don't think is part of the AC) - please create a 2.3 branch.
Also - this new API needs tests.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -68,6 +68,9 @@ public partial class DynamoView : Window, IDisposable
         private ShortcutToolbar shortcutBar;
         private bool loaded = false;
 
+        // list of items in the right side bar. 
+        private List<TabItem> _tabItems = new List<TabItem>();
```
mjkkirschner: For fields in c# I don't believe we follow this style - 
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/field

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -186,11 +189,70 @@ public DynamoView(DynamoViewModel dynamoViewModel)
...
+                if (selectedTab == null || selectedTab.Equals(tab))
+                {
+                    if (_tabItems.Count > 0) {
+                        selectedTab = _tabItems[0];
+                    }
+                }
+                tabDynamic.SelectedItem = selectedTab;
+            }
+
+            this.HideOrShowRightSideBar();
```
mjkkirschner: hmm - I would except this to be triggered when the items list changes, not manually here.

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
```diff
@@ -69,14 +69,19 @@ public Window DynamoWindow
...
-
         }
 
         public void AddMenuItem(MenuBarType type, MenuItem menuItem, int index = -1)
         {
             AddItemToMenu(type, menuItem, index);
         }
 
+        // Adds the extension window to a new tab in the right side bar.
+        public void AddToExtensionsSideBar(Window w)
```
mjkkirschner: I think the AC - propose to keep this internal for now - in addition this should not take a window but a UserControl I think, and definitely needs summary tags and param tags.

reddyashish: Yes I changed it for some testing. Will make this internal and change the parameters.

mjkkirschner:
@reddyashish I think this looks great! I wonder if we should add a dropdown hamburger 


like we do for the workspace tabs to the right.

Can we also add a larger margin between the workspace tab control and the viewExtension tabs?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
reddyashish: Modified this and added an event when the list of tab items is modified.

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
reddyashish: The API now takes in ContentControl object as the input parameter

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -186,11 +202,76 @@ public DynamoView(DynamoViewModel dynamoViewModel)
...
+        // sets the extension window as the tab content.
+        internal TabItem AddTabItem(ContentControl contentControl)
+        {
+            int count = TabItems.Count;
+
+            tabDynamic.DataContext = null;
+
+            // creates a new tab item
+            TabItem tab = new TabItem();
+            tab.Header = ((Window)contentControl).Title;
```
mjkkirschner: hmm - this is not great - you cannot expect that the contentControl is a window, what about if we think about this a bit more. 

a few questions for you to consider

* Should one extension be able to add more than on tab? - right now they can.
* Should an extension be able to decide its tab title or should it be implied from the extension name?
* Should an extension author call/pass the content directly to this method or should they implement a method which returns content and we do the rest?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -186,11 +202,76 @@ public DynamoView(DynamoViewModel dynamoViewModel)
                  }
              };
 
+            // Add an event when the collection is modified.   
+            tabItems.CollectionChanged += this.OnCollectionChanged;
```
mjkkirschner: you need to unsubscribe this when the dynamoView is disposed or closed. Checkout the dispose method of this class.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
reddyashish: I agree, just to show some title I made this line. I was thinking, setting the title of the tab with the value the user passes to the extension name would be a good start. In that case, this line would be removed and will be using the name passed by the user.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
reddyashish: yeah. I will add that.

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
```diff
@@ -69,14 +69,23 @@ public Window DynamoWindow
...
         {
             AddItemToMenu(type, menuItem, index);
         }
 
+        /// <summary>
+        /// Adds the extension UI control to a new tab in the right side bar.
+        /// </summary>
+        /// <param name="contentControl"></param>
+        /// <returns></returns>
+        internal void AddToExtensionsSideBar(ContentControl contentControl)
```
mjkkirschner: I've added some thoughts below - heres a proposal for you and @QilongTang to consider:

instead of making clients call this API and add as many tabs as they want to this tabControl - each extension implements a method where they return a `contentControl` 
We call this method sometime after we call `loaded` on viewExtensions - (I actually think calling this API currently from loaded will result in weirdness)

We call it like `AddToExtensionsSideBar(extension.GetContent(),extension));`

This has some consequences and some constraints - 
* extensions can only inject themselves once (using this API atleast)
* if they miss this chance, their tabs cannot be added again later until Dynamo is restarted.
* if the extensions do not implement this method, we should not crash.

We need to think about the balance between flexibility and maintaining a logical experience.

I'm not sure this is a much better solution - but it is more constrained.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
mjkkirschner: 👍

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -68,6 +69,21 @@ public partial class DynamoView : Window, IDisposable
         private ShortcutToolbar shortcutBar;
         private bool loaded = false;
 
+        // list of items in the right side bar. 
+        private ObservableCollection<TabItem> tabItems = new ObservableCollection<TabItem>();
+        
+        private ObservableCollection<TabItem> TabItems
```
mjkkirschner: I think you can just make this an auto implemented property
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/auto-implemented-properties

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
mjkkirschner: like you said @reddyashish  - if we don't do any of the above, we can also have the client pass the extension anyway, and still have them implement a method we can call - this lets us change the implementation later and always have access to the extension name and other events so we can keep track of what extension has created what tab/content.

IE - we could make the API simply:

`AddExtensionContentToSideBar(ViewExtension)`

QilongTang: Per our discussion, it seems we would like to 

- update it as `AddExtensionContentToSideBar(ContentControl ,ViewExtension)` so we could use extension GUID for tracking and use extension name for tab context
- give extension the right to inject anytime they want
- only create single tab for single extension

Feels like we need a wiki for this!

reddyashish: Yes, made these changes. One issue was the extension GUID was different for each instance of the same extension so couldn't use that for tracking, so used the full class name of the extension to see if it is already added to the right side bar.

reddyashish:
The changes we discussed today have been addressed. One thing pending, is to add a test case.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml
```diff
@@ -1676,6 +1677,94 @@
               Grid.Row="2"
               Grid.Column="4"
               Grid.RowSpan="2">
+            <TabControl Name="tabDynamic" ItemsSource="{Binding}">
```
QilongTang: @reddyashish Can you add some XAML comments to anything worth noticing here?

QilongTang:
Looks Solid.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml
reddyashish: Yes, I will some comments here.

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
```diff
@@ -69,14 +69,24 @@ public Window DynamoWindow
...
-
         }
 
         public void AddMenuItem(MenuBarType type, MenuItem menuItem, int index = -1)
         {
             AddItemToMenu(type, menuItem, index);
         }
 
+        /// <summary>
+        /// Adds the extension UI control to a new tab in the right side bar.
```
mjkkirschner: maybe we can call this thing ExtensionsSideBar and refer to it as such?

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
```diff
@@ -69,14 +69,24 @@ public Window DynamoWindow
...
+        /// </summary>
+        /// <param name="viewExtension"></param>
```
mjkkirschner: can you add some comments to the params?

src/DynamoCoreWpf/Properties/AssemblyInfo.cs
```diff
@@ -37,3 +37,4 @@
 [assembly: InternalsVisibleTo("CommandLineTests")]
 [assembly: InternalsVisibleTo("Watch3DNodeModelsWpf")]
 [assembly: InternalsVisibleTo("DynamoPerformanceTests")]
+[assembly: InternalsVisibleTo("SampleViewExtension")]
```
mjkkirschner: are you intending to remove this before merging?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml
```diff
@@ -1676,6 +1677,94 @@
...
+                            </Button>
+                        </DockPanel>
+                    </DataTemplate>
+                    <Style TargetType="{x:Type TabItem}">
+                        <Setter Property="MinWidth" Value="{Binding Source={x:Static configuration:Configurations.RightSideBarTabMinWidth}}" />
+                        <Setter Property="Template">
+                            <Setter.Value>
+                                <ControlTemplate TargetType="{x:Type TabItem}">
+                                    <Grid>
+                                        <Border Name="Border" Margin="0,0,0,0"  BorderThickness="0 0 1 0" BorderBrush="#FF3F4040">
```
mjkkirschner: can we pull this brush color out into the static resource dictionary so it will be easier to reuse/remap later?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -186,11 +189,94 @@ public DynamoView(DynamoViewModel dynamoViewModel)
...
+        private void OnCollectionChanged(object sender, NotifyCollectionChangedEventArgs e)
+        {
+            this.HideOrShowRightSideBar();
+        }
+
+        private Boolean IsExtensionAddedToRightSideBar(IViewExtension viewExtension)
+        {
+            foreach (TabItem tabItem in TabItems)
+            {
+                if (tabItem.Tag.Equals(viewExtension.ToString()))
```
mjkkirschner: what about if we add the fully qualified typeName as the tag? An extension author might override ToString().

mjkkirschner: then we can do typeof(ViewExtension).toString() ?

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
reddyashish: Added comments

src/DynamoCoreWpf/Views/Core/DynamoView.xaml
reddyashish: Adding this property to the configurations class.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
reddyashish: Yes, good point. Changed this to set the tag value with fully qualified type name.

src/DynamoCoreWpf/Properties/AssemblyInfo.cs
reddyashish: Yes, I will have to remove this. But how do we make sure, that this internal API is accessible from the view extension solution that the user's create?

test/DynamoCoreWpfTests/ViewExtensionTests.cs
```diff
@@ -62,6 +84,64 @@ public void Loaded(ViewLoadedParams p)
...
+
+    // A sample view extension which will be added to the extensions side bar
+    public class ExtensionsSideBarViewExtension : IViewExtension
+    {
+        public int Counter = 0;
+        public string uID = "ExtensionsSideBarDummyID";
+
+        public string UniqueId
+        {
+            set { uID = value; }
```
reddyashish: Added the setter function here only to this class because we need another instance(with a different UID) of the same extension. The extension manager checks for duplicate UID while adding extensions and it wont even call the loaded function.

src/DynamoCore/Configuration/Configurations.cs
```diff
@@ -130,6 +130,16 @@ public class Configurations
...
+
+        /// <summary>
+        /// Minimum width of tab in the extensions side bar
+        /// </summary>
+        public static readonly int ExtensionsSideBarTabMinWidth = 150;
+
+        /// <summary>
+        /// Border brush for tab items
+        /// </summary>
+        public static readonly string BorderBrushTabItems = "#FF3F4040";
```
mjkkirschner: sorry Ashish, I was not clear, I meant add it to 
https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/UI/Themes/Modern/DynamoModern.xaml

mjkkirschner: or one of these other xmal files

test/DynamoCoreWpfTests/ViewExtensionTests.cs
mjkkirschner: I don't think this will have any issues - you're only modifying your test extension right? Does this compile?

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
```diff
@@ -69,14 +69,24 @@ public Window DynamoWindow
...
 
+        /// <summary>
+        /// Adds the extension UI control element to a new tab in the extensions side bar.
+        /// </summary>
+        /// <param name="viewExtension">Instance of the view extension object that is being added to the extensions side bar.</param>
+        /// <param name="contentControl">Control UI element with a single piece of content of any type.</param>
+        /// <returns></returns>
+        internal void AddToExtensionsSideBar(IViewExtension viewExtension, ContentControl contentControl)
+        {
+            dynamoView.AddTabItem(viewExtension, contentControl);
```
mjkkirschner: hmm, I see this call returns null or the tab that gets added, did you want to raise an exception or log to the console if no tab is added?

mjkkirschner:
Looks good @reddyashish just two comments for you to consider if they worth doing.

src/DynamoCoreWpf/Properties/AssemblyInfo.cs
QilongTang: I think we can add a TODO to remove this when the API is actually public soon

reddyashish: But this will put a constraint on the user that they need to name their assembly as "SampleViewExtension" so as to access this internal API. I am wondering if there is any better solution

test/DynamoCoreWpfTests/ViewExtensionTests.cs
reddyashish: Yes, it is compiled and the test runs fine.

src/DynamoCoreWpf/Properties/AssemblyInfo.cs
QilongTang: Well, to @mjkkirschner 's point, it is not meant for 3rd party developers to use right now, let's consume it internally so I will most likely update this to PackageDependencyViewExtension in my PR. And we need to finalize the API before we make it public, otherwise we could not break it once it is made public. Not easy!

src/DynamoCore/Configuration/Configurations.cs
reddyashish: Yup, added it to the https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/UI/Themes/Modern/DynamoColorsAndBrushes.xaml.

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
reddyashish: Yes, added logging to the console.

QilongTang:
@reddyashish LGTM. Can you remind me what did we land on Unique ID for each Tab in the end, extension name or?

reddyashish:
The unique id would be the fully qualified type name.(namespace+class name of the extension)

QilongTang:
Great, looks solid

reddyashish:
Merging this. Will create a todo task to make the API public once the team finalizes it.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -186,11 +189,94 @@ public DynamoView(DynamoViewModel dynamoViewModel)
...
+                tabDynamic.DataContext = null;
+
+                // creates a new tab item
+                TabItem tab = new TabItem();
+                tab.Header = viewExtension.GetType().Name;
+                tab.Tag = viewExtension.GetType();
+                tab.HeaderTemplate = tabDynamic.FindResource("TabHeader") as DataTemplate;
+
+                // setting the extension window to the current tab content
+                tab.Content = contentControl.Content;
```
mjkkirschner: @QilongTang @reddyashish - did you guys have a chance to discuss this last week?

reddyashish: Yes, I am looking into this issue.

reddyashish: @mjkkirschner Changing this to set the contentControl object to tab.Content, worked(only for UserControl elements). This would then fail if the user is the creating a window type extension. 

I will be making this change in a new PR, but the users will have a constraint of injecting user control elements only.

PR #9787, opened 2019-06-10
Dynamo Package Dependency View Extension

QilongTang:
Please Note:

### Purpose

image

Dynamo Package Dependency View Extension using DataGrid. 

This PR:
1. Adds a view extension managed by Dynamo Team to the Dynamo.all.sln so it will always be built with Dynamo with every new commit.
2. Such view extension displays graph package dependencies when there are packages logged in dyn not installed locally (no exact package name and version match)
3. Packages are marked grey when they are installed and red when they are not
4. There is currently a dummy download button beneath the dependency table, I created so we could show people mock-ups.

image

GraphDependencyMissing

### Declarations

### Reviewers

@DynamoDS/dynamo


QilongTang:
I found adding extension project inside Dynamo repo is such a pain 👎 Localization resources seems a must-have and now getting this error which I have never encountered myself. Had no clue comparing with other view extensions and stackoverflow, will continue to look at the error

src/DynamoManipulation/DynamoManipulation.csproj
```diff
@@ -38,7 +38,7 @@
     <Reference Include="PresentationFramework" />
     <Reference Include="ProtoGeometry, Version=1.0.0.0, Culture=neutral, processorArchitecture=MSIL">
       <SpecificVersion>False</SpecificVersion>
-      <HintPath>$(DynamoExternPath)\ProtoGeometry\ProtoGeometry.dll</HintPath>
+      <HintPath>..\..\extern\ProtoGeometry\ProtoGeometry.dll</HintPath>
```
mjkkirschner: hmm weird. I think the original path was better but I guess that property was not set correctly?

src/Dynamo.All.sln
```diff
@@ -223,296 +223,594 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DynamoPackagesWPF", "Dynamo
...
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Extensions", "Extensions", "{88D45B00-E564-41DB-B57C-9509646CAA49}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "PackageDependencyViewExtension", "PackageDependencyViewExtension\PackageDependencyViewExtension.csproj", "{5E76AAB3-6302-473E-9655-081B53FB1419}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
+		Debug|x64 = Debug|x64
 		Release|Any CPU = Release|Any CPU
+		Release|x64 = Release|x64
```
mjkkirschner: why do we need these x64 configs?

src/DynamoManipulation/DynamoManipulation.csproj
QilongTang: Yup somehow other projects are all using relative path except for this one and suddenly failed for me.. @aparajit-pratap was mentioning we will move to nuget eventually so this change is good for now

src/Dynamo.All.sln
QilongTang: I had the same question but tested these are auto generated once I move all the extension projects to extension folder without even touching and project properties, visual studio magic

mjkkirschner: I am not convinced that is a good thing.

src/PackageDependencyViewExtension/PackageDependencyViewExtension.csproj
```diff
@@ -0,0 +1,103 @@
...
+    <AssemblyName>PackageDependencyViewExtension</AssemblyName>
+    <TargetFrameworkVersion>v4.7</TargetFrameworkVersion>
+    <FileAlignment>512</FileAlignment>
+    <Deterministic>true</Deterministic>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+    <DebugSymbols>true</DebugSymbols>
+    <DebugType>full</DebugType>
+    <Optimize>false</Optimize>
+    <OutputPath>..\..\bin\AnyCPU\Debug\</OutputPath>
```
mjkkirschner: @QilongTang this project should not set it's output paths like this:

see all other csprojs import the output path as it's shared:

 <ImportGroup Label="PropertySheets">
    <Import Project="$(SolutionDir)/Config/CS.props" />
  </ImportGroup>

or something like this - for example, look at the notifications csproj - it does not have output paths.

mjkkirschner: this then imports:
https://github.com/DynamoDS/Dynamo/blob/master/src/Config/CS.props#L7

src/Dynamo.All.sln
QilongTang: I dont think this is a bad thing either, for each individual project, the configuration and platform is stored in csproj, so there is no change there and those settings still dominate. The particular `GlobalSection` of `preSolution` defines the loading order, this is also auotgenerated from VS2017 not VS2019.

src/PackageDependencyViewExtension/PackageDependencyViewExtension.csproj
QilongTang: Good point, let me update

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
```diff
@@ -1,9 +1,11 @@
...
 using System.Collections.Generic;
-using System.Linq;
-using System.Reflection;
-using System.Text;
-using System.Threading.Tasks;
+using System.Runtime.CompilerServices;
+
+// Internals are visible to the Package Manager extension
+// For workspace package dependency collection
+[assembly: InternalsVisibleTo("PackageDependencyViewExtension")]
```
mjkkirschner: doesn't adding this once cover the whole assembly?

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
```diff
@@ -7,8 +13,21 @@ namespace Dynamo.PackageDependency
...
+        internal IEnumerable<Graph.Workspaces.PackageDependencyInfo> Packages;
+
+        private DynamoModel dynamoModel;
+
+        public void WorkspaceOpened(WorkspaceModel obj)
+        {
+            Packages = obj.PackageDependencies;
+        }
+
+        public PackageDependencyView(DynamoViewModel dynamoViewModel)
```
mjkkirschner: why do you need the DynamoViewModel? If you can avoid using it we can avoid exposing quite a big API surface.

src/PackageDependencyViewExtension/PackageDependencyViewExtension.cs
```diff
@@ -69,7 +70,8 @@ public void Startup(ViewStartupParams sp)
 
         public void Loaded(ViewLoadedParams p)
         {
-            var packageDependencyView = new PackageDependencyView();
+            var dynamoViewModel = p.DynamoWindow.DataContext as DynamoViewModel;
```
mjkkirschner: I know other extensions do this - but it's not ideal - we should try to avoid it or provide a clearer, smaller, more consistent API if we can - that we will be able to maintain long term.

QilongTang: Agree

mjkkirschner:
this is getting cool.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -653,6 +653,7 @@ internal List<PackageDependencyInfo> PackageDependencies
             set
             {
                 packageDependencies = value;
+                RaisePropertyChanged("PackageDependencies");
```
mjkkirschner: you can use `nameof` instead for easier refactoring.

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
```diff
@@ -47,11 +36,39 @@ internal void OnWorkspaceCleared(IWorkspaceModel obj)
...
             {
+                currentWorkspace.PropertyChanged -= OnWorkspacePropertyChanged;
                 // Clear the dependency table.
                 table.Columns.Clear();
             }
         }
 
+        private void OnWorkspacePropertyChanged(object sender, PropertyChangedEventArgs args)
+        {
+            if (args.PropertyName == "PackageDependencies")
```
mjkkirschner: you might be able to use `nameof` here as well.

aparajit-pratap:
@QilongTang you might have missed moving `Notifications` to the new `Extensions` folder. It is also an extension (view extension).

QilongTang:
@aparajit-pratap It is moved already, see my picture shot

src/PackageDependencyViewExtension/PackageDependencyView.xaml
```diff
@@ -0,0 +1,90 @@
...
+                                    </ItemsControl.ItemTemplate>
+                                </ItemsControl>
+                            </ScrollViewer>
+                        </DataTemplate>
+                    </DataGridTemplateColumn.CellTemplate>
+                </DataGridTemplateColumn>
+            </DataGrid.Columns>
+        </DataGrid>
+
+        <!--Download Button-->
```
mjkkirschner: can you hide comment this out - or add a debug compile flag around this?

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
```diff
@@ -0,0 +1,232 @@
...
+        /// </summary>
+        /// <param name="p">ViewLoadedParams</param>
+        public PackageDependencyView(ViewLoadedParams p)
+        {
+            InitializeComponent();
+            DataContext = table;
+            currentWorkspace = p.CurrentWorkspaceModel as WorkspaceModel;
+            p.CurrentWorkspaceChanged += OnWorkspaceChanged;
+            p.CurrentWorkspaceCleared += OnWorkspaceCleared;
+            currentWorkspace.PropertyChanged += OnWorkspacePropertyChanged;
```
mjkkirschner: you need to unsubscribe this event handler.

mjkkirschner: I see how this could always be the same workspace - but I think for safety's sake we should unsub this when the workspace changes and re sub it there.

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
```diff
@@ -0,0 +1,232 @@
...
+        /// Event handler for workspaceAdded event
+        /// </summary>
+        /// <param name="obj">workspace model</param>
+        internal void OnWorkspaceChanged(IWorkspaceModel obj)
+        {
+            if (obj is WorkspaceModel)
+            {
+                DependencyRegen(obj as WorkspaceModel);
+                currentWorkspace = obj as WorkspaceModel;
+                currentWorkspace.PropertyChanged += OnWorkspacePropertyChanged;
```
mjkkirschner: I think here we should unsubscribe the old workspace before subscribing the new one - even if they are usually the same.

src/PackageDependencyViewExtension/PackageDependencyViewExtension.cs
```diff
@@ -0,0 +1,101 @@
+﻿using Dynamo.Extensions;
+using Dynamo.Graph.Workspaces;
+using Dynamo.Wpf.Extensions;
+using System.Windows.Controls;
+
+namespace Dynamo.PackageDependency
+{
+    /// <summary>
+    /// The Extension framework for Dynamo allows you to extend
```
mjkkirschner: @QilongTang please add a better summary of this class.

src/PackageDependencyViewExtension/PackageDependencyViewExtension.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        public void Loaded(ViewLoadedParams viewLoadedParams)
+        {
+            DependencyView = new PackageDependencyView(viewLoadedParams);
+            viewLoadedParams.AddToExtensionsSideBar(this, DependencyView);
+
+            // TODO: When all necessary dependencies are already loaded, hide
+            // the view extension on the right. Adding a button in view menu to load
+            // manually when user needs it. May introduce inconsistent dependencies info, need testing.
+            packageDependencyMenuItem = new MenuItem { Header = "Show View Extension Sample Window" };
```
mjkkirschner: this text is incorrect right?

mjkkirschner: also please add to resx.

mjkkirschner:
@QilongTang one more UI point - the text editing cursor should not be shown over users hover over the entries in the list - this implies they can edit the field, which they cant. I would look at a textBlock instead of a text box with readonly set. Or if you prefer set the hover cursor manually.

Of course we can improve this in another PR.

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
QilongTang: yup, I removed because this is causing crashes sometimes (I think I showed you).. Let me add it back and test

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
QilongTang: Added back

QilongTang:
@mjkkirschner Please check the latest commit, addressed

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
```diff
@@ -38,6 +38,7 @@ internal void OnWorkspaceCleared(IWorkspaceModel obj)
         {
             if (obj is WorkspaceModel)
             {
+                currentWorkspace.PropertyChanged -= OnWorkspacePropertyChanged;
```
mjkkirschner: this seems like an odd place to do this -couldn't the workspace be cleared for some other reason... and if you clear the workspace but then don't add the the subscription back then you're not watching the workspace any longer...

I would have imagined this would be done in the `onCurrentWorkspaceChanged` handler - you cache the current workspace, unsubscribe the handler, then get the new current workspace and then subscribe?

Does that not work?

mjkkirschner:
@QilongTang I think then logic for workspace event subscription is not correct. I know it's not straight forward because the workspace removed is rarely triggered, and the workspaceCleared is triggered at unexpected times but even with that it mind it seems odd...

How do you want to proceed with tests, as a followup?

src/PackageDependencyViewExtension/PackageDependencyView.xaml
```diff
@@ -0,0 +1,91 @@
...
+        <Grid.RowDefinitions>
+            <RowDefinition Height="Auto"/>
+            <RowDefinition Height="Auto"/>
+        </Grid.RowDefinitions>
+        <DataGrid 
+            Name="PackageDependencyTable" 
+            HorizontalAlignment="Left"
+            VerticalAlignment="Top" 
+            Width="500"
+            Margin="5,5,0,0"
```
reddyashish: @QilongTang minor comment, this margin looks odd for some reason. There is already a white border so do you think we can get rid of this?

QilongTang: Thanks for reminding me! Removed, I noticed you have built-in margin already

QilongTang:
@mjkkirschner Yes, I would like to address tests as follow up and we should really manual test this together

mjkkirschner:
LGTM

scottmitchell:
@QilongTang This is awesome!

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
```diff
@@ -0,0 +1,238 @@
...
+                            {
+                                new ColumnData(package.Name),
+                                new ColumnData(package.Version.ToString(), 100)
+                            }
+                        });
+                        matchFound = true;
+                        continue;
+                    }
+                }
+                // TODO: Not ideal! O(N * M) complexicty, would like LoadedPackageDependencies to be a dictionary or something with constant package name search time
```
scottmitchell: Good point. I think we could improve these properties/objects a lot. What if we removed `LoadedPackageDependencies` and just added an `IsLoaded` property to the `PackageDependencyInfo` type?

scottmitchell: I think there are a few reasons that it might make sense to split `PackageDependencyInfo` into two classes: `PackageInfo` and `PackageDependency`

QilongTang: @scottmitchell We have to also not only deal with loaded but also higher version which are compatible and other cases, like lower version exist but not sure if they should satisfy.. Any case, there are some aspects to think more here

QilongTang:
@mjkkirschner @reddyashish Three UI recorded tests to be fixed but is more related to the view injection API behavior, merging

PR #9803, opened 2019-06-19
Refactor PackageDependencies property

scottmitchell:
### Purpose
This PR addresses https://jira.autodesk.com/browse/DYN-1997 and also simplifies the package dependency collection process. 
This PR removes `LoadedPackageDependencies` property and refactors `PackageDependencies` property/collection and `PackageDependencyInfo` object.

- Only one package should be listed as a dependency for a node
- By default, the original deserialized package dependency should be maintained
- API to override the original dependency with the local dependency
- Unit test
- Adds new `PackageInfo` type.
- Adds `IsLoaded` property to `PackageDependencyInfo`
- Simplifies package dependency collection process

This should fix the case found by @QilongTang where a node can have two package dependencies when multiple packages are loaded with the same custom node. Previously, both the deserialized package dependency and the collected package dependency of a single node could be contained in the `PackageDependencies` property of a workspace. Now, the *original* package dependency is maintained until it is voided using `VoidNodeDependency()`. After the original dependency has been voided, the local package dependency will be used for serialization.

### Declarations

### Reviewers

@mjkkirschner @aparajit-pratap @QilongTang @ColinDayOrg

### DIFF
--- a/src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
+++ b/src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
@@ -3,6 +3,9 @@ using System.Collections.Generic;
 
 namespace Dynamo.Graph.Workspaces
 {
+    /// <summary>
+    /// Class containing info about a package
+    /// </summary>
     internal class PackageInfo
     {
         /// <summary>
--- a/src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
+++ b/src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
@@ -3,11 +3,7 @@ using System.Collections.Generic;
 
 namespace Dynamo.Graph.Workspaces
 {
-    /// <summary>
-    /// Class containing info about a Dynamo package. 
-    /// Used for serialization.
-    /// </summary>
-    internal class PackageDependencyInfo
+    internal class PackageInfo
     {
         /// <summary>
         /// Name of the package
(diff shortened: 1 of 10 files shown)


mjkkirschner:
so this does not address the underlying issue right? (as you said above) - the point brought up by @ColinDayOrg - that likely which package is loaded depends on load order. IE if you restart Dynamo you might load packages in a different order than if install them manually.

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
```diff
@@ -3,21 +3,97 @@
 
 namespace Dynamo.Graph.Workspaces
 {
+    internal class PackageInfo
```
mjkkirschner: even though it's internal, a summary block would not hurt.

scottmitchell:
@mjkkirschner Right, this PR just allows package dependency collection to deal with cases where multiple packages, or versions of packages, are able to resolve nodes. The underlying issue will be addressed later, as the outcome of @aparajit-pratap 's spike into the issue.

mjkkirschner:
@scottmitchell this looks good to me - can you also cherry pick to 2.3 branch.

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
```diff
@@ -65,26 +65,18 @@ internal void DependencyRegen(WorkspaceModel ws)
             table.Columns.Clear();
             foreach (var package in ws.PackageDependencies)
             {
-                bool matchFound = false;
-                // Check if the target package is installed and loaded.
-                foreach (var loadedPackage in ws.LoadedPackageDependencies)
+                if (package.IsLoaded)
```
QilongTang: Much easier to refresh UI now, great. The only question is that how costly to maintain this property on each node add operation, would love to know how slow it is

QilongTang:
LGTM

aparajit-pratap:
@scottmitchell I have a DYN file with package dependency as clockwork 1.x and the same package installed. However when I load the file, I still see package dependency 1.x to be missing (in red). I tried the same thing with 2.x; existing DYN file with package dependency serialized as 2.x, and 2.x locally installed. Same issue; 2.x is still missing in view extension.

mjkkirschner:
@scottmitchell since we all missed that bug except @aparajit-pratap 🦅 - can we add a test for it?

PR #9804, opened 2019-06-19
Adding fix to show the dependency extension only when there is a missing package dependency

reddyashish:
### Purpose

This PR is to show the package dependency extension only when there is a missing package. 

1) When the user opens up a new workspace, the extension is not shown. 
2) When the user opens up a graph, that has all dependencies installed, the extension is not shown.  
3) When the user opens up a graph, that has missing dependencies, the extension is visible.

right side bar

One thing to note here is that, when the user first opens up a graph(with missing dependencies), the extension side bar is shown. Now if the user closes the graph and opens a new graph that has all the dependencies installed, the extension side bar is still shown as we do not have an API to close the extension as of now. But if the user manually closes the extension tab at some point, then the dependencies extension is shown again only when the user opens a graph that has missing dependencies. 

This fixes the failing Recorded tests for now.  I am not sure how feasible this is, looking for suggestions. 

### Declarations

### Reviewers

@QilongTang @mjkkirschner

### DIFF
--- a/src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
+++ b/src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
@@ -1,9 +1,10 @@
-﻿using Dynamo.Graph.Workspaces;
-using Dynamo.Wpf.Extensions;
+﻿using System;
 using System.Collections.ObjectModel;
 using System.ComponentModel;
 using System.Windows.Controls;
 using System.Windows.Media;
+using Dynamo.Graph.Workspaces;
+using Dynamo.Wpf.Extensions;
 
 namespace Dynamo.PackageDependency
 {
@@ -15,6 +16,25 @@ namespace Dynamo.PackageDependency
         private DependencyTable table = new DependencyTable();
 
         private WorkspaceModel currentWorkspace;
+
+        private ViewLoadedParams loadedParams;
+        private PackageDependencyViewExtension dependencyViewExtension;
+
+        private Boolean missingPackage = false;
+
+        public Boolean MissingPackage
+        {
+            get { return missingPackage; }
+            set
+            {
+                missingPackage = value;
+                if (missingPackage)
+                {
+                    loadedParams.AddToExtensionsSideBar(dependencyViewExtension, this);
+                }
+            }
+        }
+
 
         /// <summary>
         /// Event handler for workspaceAdded event
@@ -86,6 +106,7 @@ namespace Dynamo.PackageDependency
                 // TODO: Not ideal! O(N * M) complexicty, would like LoadedPackageDependencies to be a dictionary or something with constant package name search time
                 if (!matchFound)
                 {
+                    MissingPackage = true;
                     table.Columns.Add(new Column()
                     {
                         ColumnsData = new ObservableCollection<ColumnData>()
@@ -102,7 +123,7 @@ namespace Dynamo.PackageDependency
         /// Constructor
         /// </summary>
         /// <param name="p">ViewLoadedParams</param>
-        public PackageDependencyView(ViewLoadedParams p)
+        public PackageDependencyView(PackageDependencyViewExtension viewExtension,ViewLoadedParams p)
         {
             InitializeComponent();
             DataContext = table;
@@ -110,6 +131,8 @@ namespace Dynamo.PackageDependency
             p.CurrentWorkspaceChanged += OnWorkspaceChanged;
             p.CurrentWorkspaceCleared += OnWorkspaceCleared;
             currentWorkspace.PropertyChanged += OnWorkspacePropertyChanged;
+            loadedParams = p;
+            dependencyViewExtension = viewExtension;
         }
     }
 
--- a/src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
+++ b/src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
@@ -1,10 +1,10 @@
-﻿using System;
+﻿using Dynamo.Graph.Workspaces;
+using Dynamo.Wpf.Extensions;
+using System;
 using System.Collections.ObjectModel;
 using System.ComponentModel;
 using System.Windows.Controls;
 using System.Windows.Media;
-using Dynamo.Graph.Workspaces;
-using Dynamo.Wpf.Extensions;
 
 namespace Dynamo.PackageDependency
 {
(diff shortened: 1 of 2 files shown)


src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
```diff
@@ -16,6 +17,25 @@ public partial class PackageDependencyView : UserControl
 
         private WorkspaceModel currentWorkspace;
 
+        private ViewLoadedParams loadedParams;
+        private PackageDependencyViewExtension dependencyViewExtension;
+
+        private Boolean missingPackage = false;
+
+        public Boolean MissingPackage
```
mjkkirschner: no need to have a backing field - use an auto property.

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
```diff
@@ -16,6 +17,25 @@ public partial class PackageDependencyView : UserControl
...
+        {
+            get { return missingPackage; }
+            set
+            {
+                missingPackage = value;
+                if (missingPackage)
+                {
+                    loadedParams.AddToExtensionsSideBar(dependencyViewExtension, this);
```
mjkkirschner: I think adding side effects to a setter is usually frowned upon.
I might instead raise a property change notification and implement notify property changed event - BUT that said - this class is small enough at this point it's probably okay....

I think this is a good read:
https://stackoverflow.com/questions/2908415/what-getters-and-setters-should-and-shouldnt-do

it seems to me that the side effect of opening the panel is very related to the property... so I'm okay with it, but consider the alternatives.

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
```diff
@@ -86,6 +106,7 @@ internal void DependencyRegen(WorkspaceModel ws)
                 // TODO: Not ideal! O(N * M) complexicty, would like LoadedPackageDependencies to be a dictionary or something with constant package name search time
                 if (!matchFound)
                 {
+                    MissingPackage = true;
```
mjkkirschner: I think the `DependencyRegen` method needs a unit test.

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
mjkkirschner: also, please add a summary for a public property - does it need to be public?

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
```diff
@@ -1,9 +1,10 @@
-﻿using Dynamo.Graph.Workspaces;
-using Dynamo.Wpf.Extensions;
+﻿using System;
 using System.Collections.ObjectModel;
 using System.ComponentModel;
 using System.Windows.Controls;
 using System.Windows.Media;
+using Dynamo.Graph.Workspaces;
+using Dynamo.Wpf.Extensions;
```
QilongTang: Can you update your VS setting to have system usings after custom ones? I think I have updated code style so it is easier for all of us follow that?

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
```diff
@@ -16,6 +17,25 @@ public partial class PackageDependencyView : UserControl
 
         private WorkspaceModel currentWorkspace;
 
+        private ViewLoadedParams loadedParams;
+        private PackageDependencyViewExtension dependencyViewExtension;
```
QilongTang: This is the funny part I did not feel comfortable doing when I was talking about this with @mjkkirschner yesterday

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
```diff
@@ -16,6 +17,25 @@ public partial class PackageDependencyView : UserControl
...
+
+        private Boolean missingPackage = false;
```
QilongTang: hasMissingPackage

src/PackageDependencyViewExtension/PackageDependencyViewExtension.cs
```diff
@@ -78,8 +78,11 @@ public void Startup(ViewStartupParams viewLoadedParams)
 
         public void Loaded(ViewLoadedParams viewLoadedParams)
         {
-            DependencyView = new PackageDependencyView(viewLoadedParams);
-            viewLoadedParams.AddToExtensionsSideBar(this, DependencyView);
+            DependencyView = new PackageDependencyView(this, viewLoadedParams);
+
+            if (DependencyView.MissingPackage) {
+                viewLoadedParams.AddToExtensionsSideBar(this, DependencyView);
```
QilongTang: Dont think this line is needed, when the Loaded function gets called, it is almost always right after Dynamo initiation so you are always looking at a blank workspace, in that case the check will always be `false`. What do you think?

QilongTang:
This is nice, it covers the reasoning we had after demo, the only thing I was not sure is storing an extension reference in the view. Maybe we should instead of passing the extension to the inject API. pass in some class we invent like extension info. 

I wonder if the team has any strong opinion on this. FYI: @DyanmoDS/dynamo

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
QilongTang: Seem internal is enough

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
reddyashish: Yes, I wasn't sure too about storing the object here. There are not a lot of properties that are being exposed with this object.

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
reddyashish: Yes, changing it. I thought we were placing system directives first.

src/PackageDependencyViewExtension/PackageDependencyViewExtension.cs
reddyashish: yes, good point. we can move the API call here.

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
reddyashish: I will be removing the check from PackageDependencyViewExtension.cs as it is not needed. So this can be private. Will add the summary

QilongTang:
LGTM @reddyashish @mjkkirschner I'm fine if we store the reference of extension in view layer for now.

reddyashish:
Merging this and will cherry-pick it to 2.3 branch.

PR #9809, opened 2019-06-21
[WIP] do not crash when libG objects are garbage collected.

mjkkirschner:
### Purpose

This PR exposes a new API on `DynamoServices` to request an `action` be scheduled on the dynamo scheduler thread.

It also subscribes to that event and adds all requested actions to a list - when the graph raises an `EvaluationCompleted` event - the actions are scheduled to execute on the scheduler thread. 

Theres likely something better than this - but as a POC - this works.

This enables us to create 100,000 cuboids which immediately go out of scope - this throws or crashes on master within a few seconds.

### TODO
Look into native locking mechanism suggested by @aparajit-pratap - if that turns out to be successful this PR may not be needed.

### Declarations

### Reviewers

@aparajit-pratap

### FYI

@Dewb

### DIFF
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -585,11 +585,18 @@ namespace Dynamo.Models
             var thread = config.SchedulerThread ?? new DynamoSchedulerThread();
             Scheduler = new DynamoScheduler(thread, config.ProcessMode);
             Scheduler.TaskStateChanged += OnAsyncTaskStateChanged;
+
             //TODO dispose this.
+            List < Action > requestedDisposalCalls = new List<Action>();
             ExecutionEvents.RequestRunOnDynamoScheduler += (actionToSchedule) =>
             {
-                var task = new DelegateBasedAsyncTask(this.Scheduler, actionToSchedule);
-                Scheduler.ScheduleForExecution(task);
+                requestedDisposalCalls.Add(actionToSchedule);
+            };
+
+            this.EvaluationCompleted += (s,e) => {
+
+                var combined = Delegate.Combine(requestedDisposalCalls.ToArray());
+                this.Scheduler.ScheduleForExecution(new DelegateBasedAsyncTask(this.Scheduler, combined as Action));
             };
 
             geometryFactoryPath = config.GeometryFactoryPath;
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -600,6 +600,7 @@ namespace Dynamo.Models
                 {
                     this.Scheduler.ScheduleForExecution(new DelegateBasedAsyncTask(this.Scheduler, combined as Action));
                 }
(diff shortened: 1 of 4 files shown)


tools/NuGet/template-nuget/DynamoVisualProgramming.Tests.nuspec
```diff
@@ -21,8 +21,8 @@
   </metadata>
   <files>
     <!--These assemblies are referenced from the bin directory since they are not harvested for the installer-->
-    <file src="..\..\..\bin\AnyCPU\Release\DynamoCoreTests.dll" target="lib\net47" />
-    <file src="..\..\..\bin\AnyCPU\Release\SystemTestServices.dll" target="lib\net47" />
-    <file src="..\..\..\bin\AnyCPU\Release\TestServices.dll" target="lib\net47" />
+    <file src="DynamoCoreTests.dll" target="lib\net47" />
+    <file src="SystemTestServices.dll" target="lib\net47" />
+    <file src="TestServices.dll" target="lib\net47" />
```
aparajit-pratap: How do these paths get resolved with this change?

mjkkirschner: this is local to the build folder - most other nuspecs we have are like this - and I verified this works. I wanted to build a debug nuget locally and this stopped me.

mjkkirschner: hmm - I verified the other package. Let me double check this one.

mjkkirschner: @aparajit-pratap - it did work - but this may have only worked because I specified the harvest directory to pull files from as the bin folder...

not a big deal, I can revert this stuff for now.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -586,6 +586,22 @@ protected DynamoModel(IStartConfiguration config)
             Scheduler = new DynamoScheduler(thread, config.ProcessMode);
             Scheduler.TaskStateChanged += OnAsyncTaskStateChanged;
 
+            //TODO dispose this.
+            List < Action > requestedDisposalCalls = new List<Action>();
+            ExecutionEvents.RequestRunOnDynamoScheduler += (actionToSchedule) =>
+            {
+                requestedDisposalCalls.Add(actionToSchedule);
```
Dewb: This language assumes that every client that uses the RequestRunOnDynamoScheduler event to schedule an action will be doing so to request object disposal. If we're going to make that assumption, the naming should be consistent at all levels. If we're instead providing a general mechanism (which seems like a better idea, even if we restrict it to this use case initially) then that language should be consistent throughout as well.

mjkkirschner: @Dewb totally agree.
I think the general mechanism and naming are where this should end up.

mjkkirschner:
going to close this, we're attempting solution B which should not require a dynamo change AFAIK.

PR #9815, opened 2019-06-25
Prevent custom node package containing duplicate node definition from loading

aparajit-pratap:
### Purpose

JIRA: DYN-1993
- [X] Prevent custom node package containing an existing node definition from loading (conflicting package)
- [X] Display "Cannot download package" dialog with "MarkForUninstall" option when trying to install conflicting package
- [X] Display failure to load conflicting CN package notification to notification view extension during startup (when loading already downloaded package) and also during install
- [X] Test these changes with @scottmitchell's fixes in https://github.com/DynamoDS/Dynamo/pull/9803

image

TODO: 
- [ ] Turn package install button red on failure to install ZT and custom node packages
- [ ] Add failure dialog similar to failed custom node package install for ZT package install as well
- [ ] Show dialog even when opening duplicate custom node definitions

### Declarations

### FYIs

@mjkkirschner @scottmitchell @QilongTang


src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
```diff
@@ -393,7 +401,6 @@ private void ShowNodePublishInfo(ICollection<Tuple<CustomNodeInfo, CustomNodeDef
         {
             foreach (var f in funcDefs)
             {
-                var pmExtension = DynamoViewModel.Model.GetPackageManagerExtension();
```
mjkkirschner: nice catch.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
```diff
@@ -263,8 +263,9 @@ public PackageManagerSearchViewModel(PackageManagerClientViewModel client) : thi
         {
             this.PackageManagerClientViewModel = client;
             PackageManagerClientViewModel.Downloads.CollectionChanged += DownloadsOnCollectionChanged;
```
mjkkirschner: I don't think we clean these up from anywhere - probably we should do it in the PackageManagerSerachView.Unloaded framework event.

mjkkirschner:
@aparajit-pratap what is a `duplicatePackage` specifically?

aparajit-pratap:
> @aparajit-pratap what is a `duplicatePackage` specifically?

I used the term (probably misleading) to refer to a CN package that contained an already existing (duplicate) node definition. I have now tried to remove all usages of it to avoid confusion.

src/DynamoCore/Core/CustomNodeManager.cs
```diff
@@ -455,6 +456,12 @@ private void SetNodeInfo(CustomNodeInfo newInfo)
                 NodeInfos.Remove(guid);
             }
 
+            CustomNodeInfo value;
+            if (NodeInfos.TryGetValue(newInfo.FunctionId, out value))
+            {
+                throw new DuplicateCustomNodePackageLoadException(Path.GetDirectoryName(value.Path), "duplicate package");
```
aparajit-pratap: Will update the string and add it to resources.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
aparajit-pratap: Could this be added here in the `OnClosing` handler for the `PackageManagerSearchView`: https://github.com/DynamoDS/Dynamo/blob/405cac2930f0f72e5569deac241b6e4e5b3442d6/src/DynamoCoreWpf/Views/PackageManager/PackageManagerSearchView.xaml.cs#L26

mjkkirschner:
@Amoursol @kronz - can you chime in here about the wording of this error message - it's a complex problem but I think there might be some way to summarize it more easily or break up the text to make it look a little less imposing.

mjkkirschner:
heres my try:
```
{clockwork 1.x} cannot be loaded. 

Installing it will conflict with one or more node definitions that already exist in { clockwork 2.x } which is currently loaded. 

To install { clockwork 1.x }, Dynamo needs to first uninstall { clockwork 2.x }. 

Restart Dynamo to complete the uninstall, then try and download {clockwork 1.x } again.</value>

```

src/DynamoPackages/PackageLoader.cs
```diff
@@ -207,13 +207,26 @@ internal void TryLoadPackageIntoLibrary(Package package)
...
+                OnConflictingPackageLoaded(originalPackage, package);
+            }
             catch (Exception e)
             {
                 Log("Exception when attempting to load package " + package.Name + " from " + package.RootDirectory);
                 Log(e.GetType() + ": " + e.Message);
             }
         }
 
+        public event Action<Package, Package> ConflictingCustomNodePackageLoaded;
```
mjkkirschner: summary please if this needs to be public.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -207,13 +207,26 @@ internal void TryLoadPackageIntoLibrary(Package package)
                 package.Loaded = true;
                 this.PackgeLoaded?.Invoke(package);
             }
+            catch (CustomNodePackageLoadException e)
+            {
+                Package originalPackage =
```
mjkkirschner: @QilongTang what do you think - should we log this to analytics?

mjkkirschner:
@aparajit-pratap is it very difficult to add a test to assert the second package is not allowed to load?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
QilongTang: `OnClosing` seems a good spot

src/DynamoPackages/PackageLoader.cs
QilongTang: Worth! But are you thinking about public APIs for extention author to use or internal call only managed by our team?

mjkkirschner: am thinking about the regular log Analytics calls for events - I'm not sure if this would already be logged a custom exception type or we only log uncaught exceptions?

QilongTang:
LGTM

QilongTang:
Seems UI behavior is not covered in the unit test here, should we create a Jira task to cover this in the UI automation pipeline?

src/DynamoCore/Core/CustomNodeManager.cs
```diff
@@ -457,7 +457,7 @@ private void SetNodeInfo(CustomNodeInfo newInfo)
             }
 
             CustomNodeInfo info;
-            if (NodeInfos.TryGetValue(newInfo.FunctionId, out info))
+            if (!isTestMode && NodeInfos.TryGetValue(newInfo.FunctionId, out info))
```
mjkkirschner: I'm a bit confused - if you use testMode to turn off the behavior you just added how do you test it?

aparajit-pratap: I have added a separate test to test this functionality.

mjkkirschner:
@aparajit-pratap - what happened with the TODOs - were followups filed?

aparajit-pratap:
@mjkkirschner thanks for reminding. I intended to file them as separate tasks. Will do.

PR #9822, opened 2019-07-01
DYN-1999 Adding a section for users to provide feedback.

reddyashish:
### Purpose

This PR is to add a footer section to get feedback from the user regarding the package dependency view extension feature. 

JIRA: https://jira.autodesk.com/browse/DYN-1999

The LightBlue color for the section was making it hard to read the text inside, so I used another shade of blue. Let me know your thoughts on these. 

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCoreWpf/Views/Core/DynamoView.xaml
+++ b/src/DynamoCoreWpf/Views/Core/DynamoView.xaml
@@ -1711,7 +1711,7 @@
                         </Setter>
                     </Style>
                     
-                    <!--Template for each tab item in the extensions side bar-->
+                    <!--Template for tab header in the extensions side bar-->
                     <DataTemplate x:Key="TabHeader" DataType="TabItem">
                         <DockPanel HorizontalAlignment="Stretch" MinWidth="{Binding Source={x:Static configuration:Configurations.ExtensionsSideBarTabMinWidth}}">
                             <TextBlock Text="{Binding RelativeSource={RelativeSource AncestorType={x:Type TabItem}}, Path=Header}" HorizontalAlignment="Stretch"/>
--- a/src/DynamoCoreWpf/Views/Core/DynamoView.xaml
+++ b/src/DynamoCoreWpf/Views/Core/DynamoView.xaml
@@ -1677,7 +1677,7 @@
               Grid.Row="2"
               Grid.Column="4"
               Grid.RowSpan="2">
-            <TabControl Name="tabDynamic" ItemsSource="{Binding}">
+            <TabControl Name="tabDynamic" Background="#353535" BorderThickness="0" ItemsSource="{Binding}">
                 <TabControl.Resources>
                     <!--Styling for the close button in the tab-->
                     <Style TargetType="{x:Type Button}"
@@ -1862,7 +1862,7 @@
                       Grid.Row="2"
                       Grid.RowSpan="2"
                       Height="Auto"
-                      Width="3"
+                      Width="0"
                       Name="extensionSplitter"
                       HorizontalAlignment="Stretch"
                       VerticalAlignment="Stretch"
--- a/src/PackageDependencyViewExtension/PackageDependencyView.xaml
+++ b/src/PackageDependencyViewExtension/PackageDependencyView.xaml
@@ -89,7 +89,7 @@
 
         <!-- Template for the feedback banner -->
         <DockPanel Name="Feedback" Margin="0,800,0,40" Grid.Row="1">
-            <StatusBar Background="CornflowerBlue" DockPanel.Dock="Bottom"  Height="55">
+            <StatusBar Background="CornflowerBlue" DockPanel.Dock="Bottom" Height="55">
                 <!-- Preview text -->
                 <StatusBarItem>
                     <Border Height="50">
--- a/src/PackageDependencyViewExtension/PackageDependencyView.xaml
+++ b/src/PackageDependencyViewExtension/PackageDependencyView.xaml
@@ -96,7 +96,7 @@
                         <TextBlock TextWrapping="Wrap" Padding="10,0,10,0" FontSize="15" Foreground="White" Text="This Feature is in Preview!" VerticalAlignment="Center"/>
                     </Border>
                 </StatusBarItem>
-                <StatusBarItem Width="193" Padding="0,0,0,0" Margin="100,0,0,0" Height="40">
+                <StatusBarItem Width="193" Margin="100,0,0,0" Height="40">
                     <Border Name="FeedbackBorder" BorderThickness="1">
                         <Border.Style>
                             <Style>
@@ -113,7 +113,7 @@
                         <!-- Feedback button -->
                         <Button HorizontalAlignment="Right"
                             Width="150" Click="ProvideFeedback" Background="#373737" Height="30">
-                            <TextBlock TextWrapping="Wrap" FontSize="15" Padding="0,0,0,0" Foreground="White" Height="20" Text="Provide Feedback" VerticalAlignment="Center"/>
+                            <TextBlock TextWrapping="Wrap" FontSize="15" Foreground="White" Height="20" Text="Provide Feedback" VerticalAlignment="Center"/>
                             <Button.Style>
                                 <Style TargetType="{x:Type Button}">
                                     <Setter Property="Template">
--- a/src/PackageDependencyViewExtension/PackageDependencyView.xaml
+++ b/src/PackageDependencyViewExtension/PackageDependencyView.xaml
@@ -3,9 +3,18 @@
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
              xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
              xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+             xmlns:p="clr-namespace:Dynamo.Wpf.Properties;assembly=DynamoCoreWpf"
+             xmlns:ui="clr-namespace:Dynamo.UI;assembly=DynamoCoreWpf"
              mc:Ignorable="d" 
              VerticalAlignment="Top"
              HorizontalAlignment="Left">
+    <UserControl.Resources>
+        <ResourceDictionary>
+            <ResourceDictionary.MergedDictionaries>
+                <ui:SharedResourceDictionary Source="{x:Static ui:SharedDictionaryManager.DynamoColorsAndBrushesDictionaryUri}" />
+            </ResourceDictionary.MergedDictionaries>
+        </ResourceDictionary>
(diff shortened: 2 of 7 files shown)


src/PackageDependencyViewExtension/PackageDependencyView.xaml
```diff
@@ -26,7 +26,7 @@
             RowBackground="Transparent"
             FontSize="14"
             IsHitTestVisible="False"
-            Grid.Row="0">
+            Grid.Row="1">
```
QilongTang: Row number always start from 0 right? Why did we make this change?

QilongTang:
One comment then LGTM. Also can we get rid of the white board around extension view in the right panel? Looking a bit strange and inconsistent with the left one

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
```diff
@@ -38,6 +40,13 @@ private Boolean HasMissingPackage
             }
         }
 
+        /// <summary>
+        /// Re-directs to a web link to get the feedback from the user. 
+        /// </summary>
+        private void ProvideFeedback(object sender, EventArgs e)
+        {
+            System.Diagnostics.Process.Start(FeedbackLink);
```
mjkkirschner: this should be wrapped in a try

src/PackageDependencyViewExtension/PackageDependencyView.xaml
```diff
@@ -86,5 +86,51 @@
                 </TextBlock>
             </Button>
         </Grid>-->
+
+        <!-- Template for the feedback banner -->
+        <DockPanel Name="Feedback" Margin="0,800,0,40" Grid.Row="1">
```
QilongTang: @reddyashish @mjkkirschner I think the fixed margin would be problematic. We could either do it as the download button way. alway put it after the data grid or we need some new API exposed so we can watch for the Dynamo Window size change and adjust the margin based on that. Need more testing

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
QilongTang: Is this using whatever default explorer? Can you check the exception type which could be thrown from this call?

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
```diff
@@ -17,6 +17,8 @@ public partial class PackageDependencyView : UserControl
 
         private WorkspaceModel currentWorkspace;
 
+        private String FeedbackLink = "https://forum.dynamobim.com/t/call-for-feedback-on-dynamo-graph-package-dependency-display/37229";
```
mjkkirschner: @reddyashish I think this should just be:
https://github.com/DynamoDS/Dynamo/issues in general - @QilongTang what do you think? I think your forum post is auto closing in a few days?

mjkkirschner:
@Amoursol FYI

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
QilongTang: I have removed the auto closing last night. Also I think owner ship wise I would prefer forum post if you would like to set an example. I understand that team wise we prefer Dynamo issues, maybe we could message that differently

src/PackageDependencyViewExtension/PackageDependencyView.xaml
```diff
@@ -86,5 +86,51 @@
...
+            <StatusBar Background="CornflowerBlue" DockPanel.Dock="Bottom" Height="55">
```
mjkkirschner: as @ColinDayOrg says - it would be good if the colors here referenced the modern themes where possible so that in the future it's simpler for users to remap them.

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
mjkkirschner:
``` 
Exceptions
InvalidOperationException
No file name was specified in the Process component's StartInfo. -or- The UseShellExecute member of the StartInfo property is true while RedirectStandardInput, RedirectStandardOutput, or RedirectStandardError is true.

Win32Exception
There was an error in opening the associated file.

ObjectDisposedException
The process object has already been disposed.

PlatformNotSupportedException
Method not supported on operating systems without shell support such as Nano Server (.NET Core only).
```

reddyashish: Yes, added the try-catch here. Should all these exceptions be handled individually? Right now, I have a common exception message for all.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml
```diff
@@ -1677,7 +1677,7 @@
               Grid.Row="2"
               Grid.Column="4"
               Grid.RowSpan="2">
-            <TabControl Name="tabDynamic" ItemsSource="{Binding}">
+            <TabControl Name="tabDynamic" Background="#353535" BorderThickness="0" ItemsSource="{Binding}">
```
reddyashish: This would remove the white border around the extensions side bar.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml
```diff
@@ -1862,7 +1862,7 @@
                       Grid.Row="2"
                       Grid.RowSpan="2"
                       Height="Auto"
-                      Width="3"
+                      Width="0"
```
reddyashish: @QilongTang After removing the border, there was margin on the left due to this splitter element. I removed this for now, but let me know what you think with that. It was looking little odd due to the space on the left.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml
QilongTang: Thanks!

src/DynamoCoreWpf/Views/Core/DynamoView.xaml
QilongTang: Sounds good. Can you add a picture of the latest look?

src/PackageDependencyViewExtension/PackageDependencyView.xaml
QilongTang: If we have to can we all start from row 0

src/DynamoCoreWpf/Views/Core/DynamoView.xaml
reddyashish: new

QilongTang: Perfect!

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
```diff
@@ -38,6 +41,18 @@ private Boolean HasMissingPackage
...
+        /// <summary>
+        /// Re-directs to a web link to get the feedback from the user. 
+        /// </summary>
+        private void ProvideFeedback(object sender, EventArgs e)
+        {
+            try {
+                System.Diagnostics.Process.Start(FeedbackLink);
+            }
+            catch (Exception ex) {
+                LogMessage.Info("Could not re-direct to the Dynamo forum page for feedback: "+ ex.Message);
```
QilongTang: Can you add a simple dialog with the same language, also please add this to the resource dictionary and we should be able to localize it

src/PackageDependencyViewExtension/PackageDependencyView.xaml
```diff
@@ -86,5 +86,51 @@
...
+                                    <Trigger Property="Border.IsMouseOver" Value="False">
+                                        <Setter Property="Border.BorderBrush" Value="#373737" />
+                                    </Trigger>
+                                </Style.Triggers>
+                            </Style>
+                        </Border.Style>
+                        <!-- Feedback button -->
+                        <Button HorizontalAlignment="Right"
+                            Width="150" Click="ProvideFeedback" Background="#373737" Height="30">
+                            <TextBlock TextWrapping="Wrap" FontSize="15" Foreground="White" Height="20" Text="Provide Feedback" VerticalAlignment="Center"/>
```
mjkkirschner: @reddyashish - Aaron reminded me - I think this needs to come from a resx dictionary.

QilongTang: Yes, please.

reddyashish: Added this to resource file.

src/PackageDependencyViewExtension/PackageDependencyView.xaml.cs
reddyashish: Added this.

src/PackageDependencyViewExtension/PackageDependencyView.xaml
reddyashish: Yes, added this to DynamoColorsAndBrushes.xaml file.

src/PackageDependencyViewExtension/PackageDependencyView.xaml
reddyashish: @QilongTang  We need to discuss this. All the comments except this have been addressed.

QilongTang: Can we put both of them in Row 0? Currently Row 0 is empty correct from my understanding.

reddyashish: Yes, we can do that. As long as both the grid's are set to same row value, it is fine. Changed both to value 0.

src/PackageDependencyViewExtension/PackageDependencyView.xaml
reddyashish: If we are not setting the fixed margin, the feedback section will be placed based on the number of dependencies mentioned in the  dependency table. Since we want it to be placed as a footer, I have to place both the grids in the same row and set a fixed margin. We would need an API to make this responsive. Definitely needs more testing.

reddyashish:
@Amoursol Is the background color fine?

Amoursol:
@reddyashish  We should be very careful of introducing another colour to the Dynamo Palette without good reason. Can we try black text on the original blue please?

reddyashish:
After talking to Sol, since we have all the text in dynamo in White color, we wanted to keep that consistent. For the background color, we are using the same color that appears on the border of a node when it is selected in the workspace. 

Capture

reddyashish:
Addressed all the comments. If this looks good, we can merge it.

reddyashish:
Merging this. Will cherry-pick this to 2.3 also

PR #9823, opened 2019-07-01
DYN-1899: Node Execution Events

ColinDayOrg:
Please Note:

### Purpose

This PR adds the ability to fire events at the beginning and end of node execution when profiling.

### Declarations

### Reviewers

@mjkkirschner


src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -135,6 +135,38 @@ private void OnDispatchedToUI(object sender, UIDispatcherEventArgs e)
         /// </summary>
         public event Action<PortModel> PortDisconnected;
 
+        public class NodeExecutionEventArgs : EventArgs
```
mjkkirschner: all the public properties and the class itself should have some ///summary tags please.

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -135,6 +135,38 @@ private void OnDispatchedToUI(object sender, UIDispatcherEventArgs e)
...
+
+            public NodeExecutionEventArgs(NodeModel model, object data)
+            {
+                this.GUID = model.GUID;
+                this.Data = data;
+            }
+        }
+
+        /// <summary>
+        /// Event triggered before a node is executed.
```
mjkkirschner: I think this comment should indicate that these events are only fired when profiling is enabled.

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -135,6 +135,38 @@ private void OnDispatchedToUI(object sender, UIDispatcherEventArgs e)
...
+        /// </summary>
+        public event Action<NodeModel, NodeExecutionEventArgs> NodeExecutionBegin;
+
+        internal void OnNodeExecutionBegin(object data)
+        {
+            NodeExecutionBegin?.Invoke(this, new NodeExecutionEventArgs(this, data));
```
mjkkirschner: it appears this passes the nodes evaluated output to this handler? 
Can you confirm that in one of the tests?

test/DynamoCoreTests/ProfilingTest.cs
```diff
@@ -94,5 +96,58 @@ public void TestProfilingSingleNodePublicMethodsOnly()
...
+        private static void onNodeExectuionEnd(NodeModel sender, NodeModel.NodeExecutionEventArgs args)
+        {
+            // Assert that the node ending execution had the begin exectuion event fired previously
+            CollectionAssert.Contains(executingNodes, args.GUID);
+
+            nodeExecutionEndCount++;
+        }
+
+       [Test]
+        public void TestNodeExecutionEvents()
```
mjkkirschner: can we add test that asserts the value of the `data` object? Or get rid of that property if we think it should not be passed.

src/DynamoCore/Graph/Nodes/NodeModel.cs
mjkkirschner: alternatively we could rename them to be less general.

ColinDayOrg: Notes have been added to both events to indicate that they will only be fired when profiling is turned on. My thought is that the names should remain general in the case that it is decided to make the event firing non-dependent on profiling being turned on so to not break the API.

test/DynamoCoreTests/ProfilingTest.cs
ColinDayOrg: After looking into the tests for the data object it seemed to me that the data object was redundant because the NodeModel object was the important information which was already being passed in. I removed the class that contained the NodeModel GUID and the now extraneous data and just rely on the NodeModel object that was being passed through already.

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -135,6 +135,28 @@ private void OnDispatchedToUI(object sender, UIDispatcherEventArgs e)
         /// </summary>
         public event Action<PortModel> PortDisconnected;
 
+        /// <summary>
+        /// Event triggered before a node is executed.
+        /// Note: This event will only be triggered when profiling is active.
+        /// </summary>
+        public event Action<NodeModel> NodeExecutionBegin;
+
+        internal void OnNodeExecutionBegin(object data)
```
mjkkirschner: @ColinDayOrg seems these methods should not take an argument at all anymore.

mjkkirschner:
LGTM

QilongTang:
LGTM but what's up with so many merge commits..

ColinDayOrg:
I'm not sure about the number of merge commits. If there is a way to squash them please let me know. The only thing I have heard is that I should delete my fork and re-create it, but it seems to me that there should be a better way to do it.

PR #9827, opened 2019-07-02
Make packageDependencies property more extensible

mjkkirschner:
### Purpose

1. rename the `PackageDependencies` serialized property to `NodeLibraryDependencies` and add a `ReferenceType` property so this property is more extensible in the future.
2. update tests
3. [x] update packageDep viewer 
4. TODO: add tests

### Declarations

### Reviewers

@QilongTang @aparajit-pratap 

### FYIs

@scottmitchell @aparajit-pratap


mjkkirschner:
looks like build is failing here because
` public ReferenceType ReferenceType { get => ReferenceType.Package; }`

requires a newer c# compiler than the appveyor or eng ops build is using

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
```diff
@@ -73,10 +73,54 @@ public override string ToString()
         }
     }
 
+    internal enum ReferenceType{
```
QilongTang: Seems this enum is specific to `PackageDependencyInfo` object and we may need to do comparison later from other places, I suggest we move it from this name space to that class as a static enum.

mjkkirschner: it's a top level enum inside the namespace `Dynamo.Graph.Workspaces`, not nested under the `PackageDependencyInfo` class -  I can move it to interfaces if you think that makes more sense?

mjkkirschner:
@QilongTang such a wiki exists!?

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
QilongTang: Well that's what I mean.. I think the package dependency view extension later will need to access this enum and do comparison with the referenceType so it would be good for us to make it static

mjkkirschner:
Enums are inherently static - and I don’t think it needs to be nested in a class - especially because this type is referenced in the backing interface - we can make it public if needed but it’s already accessible from the extension.

QilongTang:
@mjkkirschner  https://wiki.autodesk.com/display/LOCGD/Dynamo+SW+Translation+Guidelines.

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
```diff
@@ -73,10 +73,54 @@ public override string ToString()
...
+    /// <summary>
+    /// An interface that describes a dependency a workspace can have on other code.
+    /// </summary>
+    interface INodeLibraryDependencyInfo
+    {
+        /// <summary>
+        /// The type of reference this dependency is.
+        /// </summary>
+        ReferenceType ReferenceType { get; }
+        /// <summary>
```
aparajit-pratap: Spaces between definitions will help with easier readability.

mjkkirschner:
@QilongTang I've updated the localization wiki, done some more renaming, added a simple test. PTAL

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -29,13 +29,6 @@
 using Newtonsoft.Json.Linq;
 using ProtoCore.Namespace;
 
-// Internals are visible to the Package Manager extension
-// For workspace package dependency collection
-[assembly: InternalsVisibleTo("DynamoPackages")]
```
mjkkirschner: I moved these to assemblyInfo.cs so they are easier to find all in one place. This still applies to the entire assembly even if they are only located in one file.

mjkkirschner:
please ignore the force push stuff - I want to make these easy to cherrypick so I needed to roll back before the python updates.

QilongTang:
@mjkkirschner LGTM

mjkkirschner:
@QilongTang @aparajit-pratap @reddyashish gonna merge this and cherry pick it - watch out - it might cause merge conflicts @reddyashish so you may want to pull it (potentially wait until it's tested?)

PR #9834, opened 2019-07-08
Correct package install error dialog messages

aparajit-pratap:
### Purpose

JIRA issue: https://jira.autodesk.com/browse/DYN-2026

In the case of a conflicting custom node package being installed, the "Cannot Download Package" dialog said that the already loaded package first needs to be uninstalled by Dynamo and Dynamo restarted. _Then the user would need to try installing the package again._ However, in this case the conflicting package is still **downloaded** even though it isn't **loaded** in the current Dynamo session. If the user selects "Yes" in the dialog, the already loaded package is deleted when Dynamo is restarted and he/she is not expected to reinstall the second package as it was already downloaded previously. When Dynamo is restarted the second package loads successfully after the original conflicting package is deleted.

The dialog message is updated to reflect this behavior more clearly: 
Before:
image

After:
image

In addition this fixes an issue in a similar dialog where the package version wasn't explicit in the error message:
Before:
image

After:
image

TODO: 
- [ ] Cherry-pick into RC2.3.0 branch.

### Declarations

- [ ] All tests pass using the self-service CI. [WIP]

### Reviewers

@QilongTang @reddyashish 

### FYIs

@mjkkirschner

### DIFF
--- a/src/DynamoCore/Properties/Resources.Designer.cs
+++ b/src/DynamoCore/Properties/Resources.Designer.cs
@@ -875,7 +875,7 @@ namespace Dynamo.Properties {
         ///   Looks up a localized string similar to {1} cannot be loaded. 
         ///Installing it will conflict with one or more node definitions that already exist in {0}, which is currently loaded. 
         ///To install {1}, Dynamo needs to first uninstall {0}. 
-        ///Restart Dynamo to complete the uninstall, then try and download {1} again..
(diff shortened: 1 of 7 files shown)


QilongTang:
I was wondering about that during the demo! Thank you for fixing it!

QilongTang:
LGTM

PR #9840, opened 2019-07-11
Adding back the width for the splitter between workspace and extensions side bar

reddyashish:
### Purpose

This PR is to add back the width for the Grid Splitter that is between the canvas and the extensions side bar. It was removed in this PR: https://github.com/DynamoDS/Dynamo/pull/9822.

In addition, I have decreased the margin as well as there is no extra padding. 

When the margin is set to -2:

The splitter element can be dragged easily now. 

When the margin is set to 0:

### Declarations

### Reviewers

@mjkkirschner @QilongTang

### DIFF
--- a/src/DynamoCoreWpf/Views/Core/DynamoView.xaml
+++ b/src/DynamoCoreWpf/Views/Core/DynamoView.xaml
@@ -1866,7 +1866,7 @@
                       Name="extensionSplitter"
                       HorizontalAlignment="Stretch"
                       VerticalAlignment="Stretch"
-                      Margin="-2"
+                      Margin="0"
                       Background="Transparent"
                       Cursor="/DynamoCoreWpf;component/UI/Images/resize_horizontal.cur" />
 
--- a/src/DynamoCoreWpf/Views/Core/DynamoView.xaml
+++ b/src/DynamoCoreWpf/Views/Core/DynamoView.xaml
@@ -1862,11 +1862,11 @@
                       Grid.Row="2"
                       Grid.RowSpan="2"
                       Height="Auto"
-                      Width="0"
+                      Width="3"
                       Name="extensionSplitter"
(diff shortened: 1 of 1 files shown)


src/DynamoCoreWpf/Views/Core/DynamoView.xaml
```diff
@@ -1862,11 +1862,11 @@
                       Grid.Row="2"
                       Grid.RowSpan="2"
                       Height="Auto"
-                      Width="0"
+                      Width="3"
                       Name="extensionSplitter"
                       HorizontalAlignment="Stretch"
                       VerticalAlignment="Stretch"
-                      Margin="0"
+                      Margin="-2"
```
reddyashish: What do you think about this?

QilongTang: Well this will mean user will only have 1 px to drag? I am leaning towards making it consistent with the other splitter, you can tell from line 1855

reddyashish: I think the user will still have 3px of splitter but the splitter element will be moved to the left and would overlap a little with the workspace but the color is transparent so no difference in the view.

mjkkirschner: a cool solution but googling a bit on this bring up some exceptions in wpf-  I wonder if it's best to avoid negative margins.

mjkkirschner:
please also make sure this gets cherry picked when merged.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml
reddyashish: Yes, I will remove the negative margin.

QilongTang:
LGTM

PR #9839, opened 2019-07-10
[DNM] Dyn 2038

aparajit-pratap:
### Purpose

@mjkkirschner  my 2 cents on the fix for DYN-2038. I just added 2 more checks to `CustomNodeManager::SetNodeInfo` and reverted the `IsTestMode` changes from the previous PR. This hasn't been tested extensively of course but something for your reference.

### Declarations

Check these if you believe they are true

- [x] The code base is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [x] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions), and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.

### Reviewers

(FILL ME IN) Reviewer 1  (If possible, assign the Reviewer for the PR)

(FILL ME IN, optional) Any additional notes to reviewers or testers.

### FYIs

(FILL ME IN, Optional) Names of anyone else you wish to be notified of

PR #9841, opened 2019-07-11
Package B should not overwrite CustomNodes defined in Package A

mjkkirschner:
#### Updated:

After going back and forth with this for a few days my opinion is that we should move very slowly with this change because the customNodeManager and package loading logic is quite fragile and has many cases. Some of which are not well defined today. I think before we can change these we need to define them and the expected behavior.

Having said that here is what I'm proposing and what I have found.

I propose that at this time **we only change one loading behavior** to address the initial bug we identified using the `WorkspaceDependencyViewer` that one node could belong to two packages.
#### After this PR:

1.  CustomNodes cannot be redefined by **a completely separate** package.
If you try to load two conflicting packages you see this message(added by @aparajit-pratap ):
```
 {package2} cannot be loaded. 
        ///Installing it will conflict with one or more node definitions that already exist in {package1}, which is currently loaded. 
        ///To install {package2}, Dynamo needs to first uninstall {package1}. 
        ///Restart Dynamo to complete the uninstall..
```

2.  this means they CAN be redefined by other versions of the same package. This is to support versioning publishing workflows. Until we are ready to refactor and study these cases and the inconsistency that exists today I have added a notification which is raised if an existing definition or version is being overridden by a package.

```
Attempting to load customNode EvenOdd, loaded by package EvenOdd, 
Version=1.0.0, but A previous definition named MyNewEvenOdd exists with no associated package. 
The new customNode definition has been loaded,
 but Dynamo may in an unstable state, please avoid loading multiple custom nodes with the id 3f19c484-d3f3-49ba-88c2-6c386a41f6ac. 
```

These are the only cases which are changed, and the only one that prevents loading is number 1. - I think this has low risk since most behaviors are unchanged, and provides some more information to help users understand what is being overwritten and by what.

#### These are the cases I was able to write automated tests for:

* CustomNode is loose - edited by a user, and resaved
* CustomNode is from a package and edited by a user.
* Two packages contain the same custom node definition, a node is placed from package 1, and package 2 attempts to load - conflicting package2 does not load.
* Placing a custom node from a package retains the packageInfo on the CustomNodeInfo
* 

#### These are the cases I could not write tests for, but work in Dynamo:
* Custom node with conflicting ID is in definitions folder, Package containing same ID is loaded, package overwrites customNode and logs a warning.
* The reverse of the above - loading the package first and then creating the customNode -

#### These are cases that are not well defined or have surprising behavior that I am hesitant to change.

* Publishing a new customNode to a package does not actually load the customNodes again, but it does move the custom node to a new location - the package is loaded, but contains no customNodes - this is existing behavior.
* Publishing a version does reload the custom nodes.

#### PackageInfo added to CustomNodeInfo

To handle these cases the `PackageInfo` class was made public and added to the `CusotmNodeInfo` class as a property - this property is valid if the customNode was requested to be loaded as a consequence of some package being loaded.

I found out that `IsPackageMember` was never consistently set on the CustomNodeInfo - ⚠️ I believe this is the information we absolutely must know to address case 1. (packages overwriting other packages) - To address this when `CustomNodeInfo` is set the `CustomNodeManager` requests the packageManager return the owningPackage of the customNode if it exists, this is done with a dictionary lookup, so it is fast.

Next we check the updated `info` is owned by a package, if it is, we check if the same package added the existing `customNodeInfo` - if so, we update the info. If the existing `Info` has no package - we log a notification (the one above in case 2), and still load the package.

The only case that throws, is case 1. Two different packages (tested by Name only) are trying to redefine the same `info` tested by `Guid`.

PR Also does the following:

* removes testMode added in previous PR
* simplifies message when trying to install a package that is already installed and requires restart.
* removes failure category from LibraryJSCustomNodetest.
* adds some Failing tests, with the category `TechDebt` that I hope we can address as we define the expected behavior of these other cases and eventually refactor CustomNodeManager / PackageLoader.

### Declarations

### Reviewers

### FYIs

@aparajit-pratap @scottmitchell @QilongTang


src/DynamoCore/Core/CustomNodeManager.cs
```diff
@@ -440,11 +453,12 @@ private IEnumerable<CustomNodeInfo> ScanNodeHeadersInDirectory(string dir, bool
...
-        /// Stores the path and function definition without initializing a node.  Overwrites
-        /// the existing NodeInfo if necessary
+        /// Stores the path and function definition without initializing a node.  
+        /// Overwrites the existing NodeInfo if necessary!
         /// </summary>
-        private void SetNodeInfo(CustomNodeInfo newInfo, bool isTestMode)
+        private void SetNodeInfo(CustomNodeInfo newInfo)
         {
+            //TODO what is this doing?
             var guids = NodeInfos.Where(x =>
```
QilongTang: I dont understand this filtering either..? @aparajit-pratap @reddyashish  Any idea?

src/DynamoCore/Core/CustomNodeManager.cs
```diff
@@ -398,7 +398,20 @@ public IEnumerable<CustomNodeInfo> AddUninitializedCustomNodesInPath(string path
...
                 info.IsPackageMember = isPackageMember;
 
-                SetNodeInfo(info, isTestMode);
+                SetNodeInfo(info);
+                result.Add(info);
+            }
+            return result;
+        }
+
+        public IEnumerable<CustomNodeInfo> AddUninitializedCustomNodesInPath(string path, bool isTestMode, PackageInfo packageInfo)
```
QilongTang: Would you add some comments for this new API?

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
```diff
@@ -6,7 +6,7 @@ namespace Dynamo.Graph.Workspaces
     /// <summary>
     /// Class containing info about a package
     /// </summary>
-    internal class PackageInfo
+    public class PackageInfo
```
QilongTang: Good, we still end up making it public. Would you check if there is any internal available code we need to remove?

src/DynamoCore/Core/CustomNodeManager.cs
```diff
@@ -398,7 +398,20 @@ public IEnumerable<CustomNodeInfo> AddUninitializedCustomNodesInPath(string path
...
+        {
+            var result = new List<CustomNodeInfo>();
+            foreach (var info in ScanNodeHeadersInDirectory(path, isTestMode))
```
QilongTang: Without checking the function `ScanNodeHeadersInDirectory` can you tell us a bit why testMode is needed here?

mjkkirschner: so because this function results in loading something from disk, its needed to be passed down to the xml and json deserialization methods:

https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Core/CustomNodeManager.cs#L601

eventually... it's used to fail tests, which is a good thing (instead of logging the exceptions like we do at runtime)

https://github.com/DynamoDS/Dynamo/blob/ec10f936824152e7dd7d6d019efdcda0d78a5264/src/DynamoCore/Graph/Workspaces/WorkspaceInfo.cs#L245

src/DynamoCore/Core/CustomNodeManager.cs
reddyashish: I think it is getting all the nodes from the dict that has the same path as the custom node that is being passed into this SetNodeInfo(). Not sure if this was added to prevent the package from having duplicate nodes.

src/DynamoCore/Core/CustomNodeManager.cs
mjkkirschner: will do, thanks for catching.

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
mjkkirschner: will do.

src/DynamoCore/Core/CustomNodeManager.cs
```diff
@@ -838,7 +859,7 @@ internal WorkspaceModel CreateCustomNode(string name, string category, string de
             };
             var workspace = new CustomNodeWorkspaceModel(info, nodeFactory);
 
-            RegisterCustomNodeWorkspace(workspace, isTestMode);
+            RegisterCustomNodeWorkspace(workspace);
```
aparajit-pratap: This function `CreateCustomNode` should not have the `bool isTestMode = false` argument.

mjkkirschner: thanks will remove.

src/DynamoCore/Core/CustomNodeManager.cs
aparajit-pratap: No, this was existing code. I didn't touch it. Not me 🏃

aparajit-pratap:
@mjkkirschner there are few other places where you need to make changes to account for the `isTestMode` removal - mainly in the tests. Please refer to commit: https://github.com/DynamoDS/Dynamo/pull/9839/commits/1c3893303a6e22c720da3a7ac84382dcfe325bf9

src/DynamoCore/Core/CustomNodeManager.cs
aparajit-pratap: This is the reason why it was added: https://github.com/DynamoDS/Dynamo/commit/5eeb1a431d0307568ed83295d74ff5ca037813aa

aparajit-pratap:
Aside from the comments above, looks good. Thanks @mjkkirschner

src/DynamoCore/Core/CustomNodeManager.cs
QilongTang: Thanks @aparajit-pratap, I see now. Seems without this, dyfs saving to the same location will end up causing dup nodes in library.

QilongTang:
Some comments there, with Unit tests should look good

mjkkirschner:
Unfortunately during testing I found 2 issues-
1. Customnodes often have their info reset and loose the package info and member properties - this is because they use data from the workspace only - I’m looking at solutions for this and already have one, but I’d like to use the new logic added to package loader to check if the cn belongs to a package to be more accurate.

2. More cases to test - involving the transition from custom node to package member ie when publishing a package or version of a package.

test/DynamoCoreWpfTests/WorkspaceSaving.cs
```diff
@@ -1266,6 +1268,58 @@ public void CustomNodeWorkspaceSavedToSamePlaceNotCausingDuplicatedLibraryItems(
...
+
+            //swtich back to the home workspace
+            ViewModel.Model.ExecuteCommand(new DynamoModel.SwitchTabCommand(0));
+            Assert.IsAssignableFrom(typeof(HomeWorkspaceModel), this.ViewModel.Model.CurrentWorkspace);
+
+            var nodeInstance = this.ViewModel.Model.CurrentWorkspace.Nodes.OfType<Function>().FirstOrDefault();
+            Assert.AreEqual(2, nodeInstance.OutPorts.Count());
+            Assert.AreEqual("anewoutput", nodeInstance.OutPorts[1].Name);
+
+        }
```
aparajit-pratap: Nice test 👍

test/DynamoCoreWpfTests/WorkspaceSaving.cs
```diff
@@ -1266,6 +1268,58 @@ public void CustomNodeWorkspaceSavedToSamePlaceNotCausingDuplicatedLibraryItems(
...
+        {
+
+            // load package A
+            // place custom node instance belonging to package A in homeworkspace
+            // modify custom node
+            // save custom node to same path
+            // close custom node
+            // open custom node
+            // assert instances changed their number of output ports.
+            var packageDirectory = Path.Combine(TestDirectory, "pkgs", "PackageThatWillBeModified");
```
aparajit-pratap: Does not adding this new package change the number of total packages in the test folder. Unfortunately all tests testing for the total number of packages in this folder may need updating.

test/DynamoCoreTests/CustomNodes.cs
```diff
@@ -1375,5 +1375,20 @@ public void InputNameWithSpacesShouldCauseWarningInGraph()
             // Despite the invalid input symbol, the custom node should still execute and return the correct value
             AssertPreviewValue("f91658aeafe84cbaa39d370dec283b53", 8.0);
         }
+
+        [Test]
+        public void LooseCustomNodeShouldNotHavePackageInfoOrPackageMember()
+        {
+            //load any loose custom node
+            var dynFilePath = Path.Combine(TestDirectory, @"core\CustomNodes\invalidInputName.dyn");
```
aparajit-pratap: Shouldn't this be a dyf?

mjkkirschner: yes, thanks for catching, I am still working on the tests.

mjkkirschner:
@aparajit-pratap @QilongTang - you may want to hang on before reviewing this again - I will redo my PR description as my thinking on this has changed a bit.

src/DynamoCore/Core/CustomNodeManager.cs
```diff
@@ -458,19 +480,35 @@ private IEnumerable<CustomNodeInfo> ScanNodeHeadersInDirectory(string dir, bool
...
             }
 
-            
+            // we need to check with the packageManager that this node if this node is in a package or not - 
+            // currently the package data is lost when the customNode workspace is loaded.
+            // we'll only do this check for customNode infos which don't have a package currently to verify if this
+            // is correct.
+            if(newInfo.IsPackageMember == false)
+            {
+                var owningPakckage = this.OnRequestCustomNodeOwner(newInfo.FunctionId);
```
aparajit-pratap: typo: `owningPackage`

src/DynamoPackages/Package.cs
```diff
@@ -406,8 +406,15 @@ internal void RefreshCustomNodesFromDirectory(CustomNodeManager customNodeManage
         {
             LoadedCustomNodes.Clear();
 
-            foreach (var x in customNodeManager.AddUninitializedCustomNodesInPath(CustomNodeDirectory, isTestMode))
+            var reloadedCustomNodes = (customNodeManager.AddUninitializedCustomNodesInPath(
```
aparajit-pratap: Starting parens `(` can be omitted?

aparajit-pratap:
@mjkkirschner given the new findings from your testing and the need to add these new events, I feel that adding the `PackageInfo` property to `CustomNodeInfo` has added more complexity. I'm trying to understand if it was really worth doing this and what are the advantages other than package equality is done better using `PackageInfo` objects rather than `CustomNodeInfo.Path` strings.

mjkkirschner:
@aparajit-pratap - yep we can discuss more tomorrow, I do not think the path information alone gives us enough to make the correct determination, some examples:
* if I work on a package and publish a version we can't identify this case - (granted I'm not raising any errors now for this case) - but it will not be easily possible to identify it from paths.
* I am working on a new loose custom node, then I publish it (potentially it gets reloaded from a new path) - 
* The *packageMember* flag is incorrectly set as soon as you place a node - so the check in your followup PR will not work after interacting with Dynamo, it will only work on first load. 

Alot of these cases are tricky no matter if we use paths or packageInfo - the crux being that the `PackageMember` flag was never set consistently, 

tldr;
 No matter if we use paths or packageInfo - we will have to do extra work to keep the flag indicating package membership in sync - see the latest commits using `RequestCustomNodeOwner ` - I don't think the `packageInfo` was the complicated part, it's just that we don't have good information on if a customNode belongs to a package today, and to identify the case we are trying to solve, we need to know this accurately.

src/DynamoCore/Core/CustomNodeManager.cs
```diff
@@ -518,17 +518,40 @@ private void SetNodeInfo(CustomNodeInfo newInfo)
                 var message = string.Format(Resources.MessageCustomNodePackageFailedToLoad,
                     infoPath, newInfoPath);
 
-                // if these are different packages raise an error.
-                // TODO (for now we don't raise an error for different versions of the same package)
-                if (newInfo.PackageInfo.Name != info.PackageInfo.Name)
+               
+                //only try to compare package info if both customNodeInfos have package info.
+                if(info.IsPackageMember && info.PackageInfo != null)
```
aparajit-pratap: Doesn't one being true mean the other is true - if `IsPackageMember` is true, doesn't it imply that `PackageInfo` is non-null or is this double check just to be safe?

mjkkirschner: yes, it's just a double check, the second check should always be true but I wanted to verify.

src/DynamoCore/Core/CustomNodeManager.cs
```diff
@@ -542,8 +542,10 @@ private void SetNodeInfo(CustomNodeInfo newInfo)
                     // it will be useful to know when. It might happen during a complex versioning publish workflow.
                     // This represents the case where a previous info was not from a package, but the current info
                     // has an owning package.
-                    var looseCustomNodeToPackageMessage = String.Format("Attempting to load a new definition for customNode {0}," +
-                        " loaded by package {1}, but A previous definition exists with no associated package, ", info.Name, newInfo.PackageInfo);
+                    var looseCustomNodeToPackageMessage = String.Format("Attempting to load customNode {0}," +
+                        " loaded by package {1}, but A previous definition named {2} exists with no associated package." +
+                        " The new customNode definition has been loaded, but Dynamo may in an unstable state, please avoid loading" +
```
aparajit-pratap: "Dynamo may be in an unstable state".

src/DynamoCore/Core/CustomNodeManager.cs
```diff
@@ -542,8 +542,10 @@ private void SetNodeInfo(CustomNodeInfo newInfo)
...
+                        " multiple custom nodes with the id {3}. ", newInfo.Name, newInfo.PackageInfo, info.Name, info.FunctionId);
```
aparajit-pratap: I wonder if it's useful to show the ID here. What can users do with it even if they know it?

mjkkirschner: look through their dyfs? I think this info is more for us than users.

mjkkirschner: we could change it to say ... "please avoid loading multiple custom nodes with the same id" ?

test/DynamoCoreWpfTests/PublishPackageViewModelTests.cs
```diff
@@ -116,5 +118,47 @@ public void NewPackageVersionUpload_DoesNotThrowExceptionWhenDLLIsLoadedSeveralT
...
+            newPkgVm.BuildVersion = "1";
+            newPkgVm.PublishLocallyCommand.Execute();
+
+            Assert.IsTrue(GetModel().GetPackageManagerExtension().PackageLoader.LocalPackages.Any
+                (x => x.Name == "PublishingACustomNodeSetsPackageInfoCorrectly" && x.Loaded == true && x.LoadedCustomNodes.Count ==1));
+
+
+            Assert.AreEqual(new PackageInfo("PublishingACustomNodeSetsPackageInfoCorrectly", new Version(0,0,1))
+                ,GetModel().CustomNodeManager.NodeInfos[cnworkspace.CustomNodeId].PackageInfo);
+            Assert.IsFalse(GetModel().CustomNodeManager.NodeInfos[cnworkspace.CustomNodeId].IsPackageMember);
```
aparajit-pratap: Is this not expected to be true?

test/Libraries/PackageManagerTests/PackageLoaderTests.cs
```diff
@@ -237,6 +268,150 @@ public void LoadingConflictingCustomNodePackageDoesNotGetLoaded()
...
+            Assert.IsTrue(entries.Count(x => Path.GetDirectoryName(x.Path).EndsWith(@"EvenOdd2\dyf")) == 0);
+            Assert.IsTrue(entries.Count(x => Path.GetDirectoryName(x.Path).EndsWith(@"EvenOdd\dyf") &&
+                                             x.FullName == "Test.EvenOdd") == 1);
+
+            var customNodeInstance = CurrentDynamoModel.CustomNodeManager.CreateCustomNodeInstance(Guid.Parse("3f19c484-d3f3-49ba-88c2-6c386a41f6ac"));
+            //this will reset the info.
+            this.CurrentDynamoModel.CurrentWorkspace.AddAndRegisterNode(customNodeInstance);
+
+            //load again.
+            loader.LoadPackages(new Package[] { package1, package2 });
```
aparajit-pratap: Why are you loading the packages again here? One of them is already loaded so it's not going to load again, is it? The following assertions are going to work even without this `LoadPackages` call again, won't they?

mjkkirschner: I want to try loading package2 again - after the customNodeInstance is placed - because this action resets the customNodeInfo to not include the package info - so I want to verify that loading package2 (which was not loaded the first time) - is still not loaded even after creating an instance of this custom node.

mjkkirschner: I could be more explicit and not try to load package1 again for clarity.

test/DynamoCoreWpfTests/PublishPackageViewModelTests.cs
mjkkirschner: this is currently false, because the packages does not actually get loaded. - It just gets added to the list of loaded packages... not exactly the same thing. Thats why this one is tech debt, my expectations were not what I found.

aparajit-pratap:
LGTM!

PR #9843, opened 2019-07-15
DYN-2020 State of workspace dependencies

reddyashish:
### Purpose

This PR is to add different states to the workspace dependencies. The different states are:

1) Exact version of Package Dependency is installed locally . -- Loaded
2) Different version of Package Dependency is installed locally -- IncorrectVersion

incorrect version of package dependency

3) Package Dependency is not installed locally  --  Missing

completely missing package

4) Package Dependency is not installed locally, but some or all of its dependent nodes are resolved by a different local package.   -- Warning

Missing package but resolved by different package

5) Package Dependency has been changed and is awaiting a restart of Dynamo . -- RequiresStart

The last state "RequiresStart" should be assigned when the user manually clicks on install package when it is in "Missing" or "IncorrectVersion" state. 

### Declarations

### Reviewers


mjkkirschner:
@reddyashish just wanted to make sure you are aware - it looks like there are 7 additional failures on the self serve for your PR:

```
Errors and Failures:
1) Test Failure : Dynamo.Tests.PackageDependencyTests.CustomNodePackageDependencyIsCollected
     Expected: True
  But was:  False

at Dynamo.Tests.PackageDependencyTests.CustomNodePackageDependencyIsCollected()

2) Test Failure : Dynamo.Tests.PackageDependencyTests.NodeModelPackageDependencyIsCollected
     Expected: True
  But was:  False


3) Test Failure : Dynamo.Tests.PackageDependencyTests.PackageDependenciesUpdatedAfterNodeRemoved
     Expected: True
  But was:  False

at Dynamo.Tests.PackageDependencyTests.PackageDependenciesUpdatedAfterNodeRemoved()

4) Test Failure : Dynamo.Tests.PackageDependencyTests.PackageDependenciesUpdatedAfterNodesAdded
     Expected: True
  But was:  False

at Dynamo.Tests.PackageDependencyTests.PackageDependenciesUpdatedAfterNodesAdded()

5) Test Failure : Dynamo.Tests.PackageDependencyTests.PackageDependenciesUpdatedAfterPackageAndNodeAdded
     Expected: True
  But was:  False


6) Test Failure : Dynamo.Tests.PackageDependencyTests.PackageDependencyIsCollectedForNewWorkspace
     Expected: True
  But was:  False


7) Test Failure : Dynamo.Tests.PackageDependencyTests.ZeroTouchPackageDependencyIsCollectedAndSerialized
     Expected: True
  But was:  False
```

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -28,6 +19,14 @@
...
 using Newtonsoft.Json.Linq;
 using ProtoCore.Namespace;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Xml;
```
mjkkirschner: why do you need to reference system xml?

mjkkirschner: whoops - my fault, did not see these were already there - @QilongTang - I can't remember how are these supposed to be sorted?

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -548,7 +547,29 @@ internal List<INodeLibraryDependencyInfo> NodeLibraryDependencies
                             packageDependencies[saved] = new PackageDependencyInfo(saved);
                         }
                         packageDependencies[saved].AddDependent(node.GUID);
-                        packageDependencies[saved].IsLoaded = saved.Equals(collected);
+
+                        if (collected != null)
+                        {
+                            if (saved.Name == collected.Name)
```
mjkkirschner: I think it might be helpful to label these cases

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -102,30 +102,57 @@ internal void DependencyRegen(WorkspaceModel ws)
         {
             // Clear the dependency table.
             table.Columns.Clear();
-            foreach (var package in ws.NodeLibraryDependencies)
+            foreach (PackageDependencyInfo package in ws.NodeLibraryDependencies)
```
mjkkirschner: not all objects in `NodeLibraryDependencies` will be `PackageDependencyInfo` - I think it would be good to check the type - ie - if dependency is `PackageDependencyInfo` - NodeLibraryDependencies should contain a list of `INodeLibraryDependencies`

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
QilongTang: Systems ones after project specific, scope from specific to generic

reddyashish:
@mjkkirschner checking on the failures.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
QilongTang: Agree, no need to be specific here

mjkkirschner:
@reddyashish would it be possible to add tooltips foreach state? I think we can re use these same tooltips later when we improve the UI controls, so it won't be rework.

Ie - if I hover over a missing package name, the tooltip says something like "this package is missing" or "A differnt version of this package is installed"

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
```diff
@@ -3,6 +3,18 @@
...
+    /// <summary>
+    /// Enum containing the different types of package dependency states.
+    /// </summary>
+    internal enum PackageDependencyState
+    {
+        Loaded,
+        IncorrectVersion,
+        Missing,
+        Warning,
+        RequiresRestart
```
aparajit-pratap: Where is this set and how?

reddyashish: This state is not yet assigned. This would happen when the user clicks on the "Keep local" button on the UI. Thats when it will trigger the uninstall of some package and this should be set. 

@QilongTang @mjkkirschner  Since the dependency table doesn't get refreshed until we reopen the graph or or click on the menu item, I was wondering how we are going to update the information in the extension tab. I think before handling this, it would be better to have more API's like closing the tab and refreshing the tab. What do you think?

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
reddyashish: Changed this to check for the type PackageDependencyInfo. Should the other objects be displayed in the dependency table or just those which are of PackageDependencyInfo type?

mjkkirschner: only show the PackageDep objects for now, the others don't exist yet 😉

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -548,7 +547,29 @@ internal List<INodeLibraryDependencyInfo> NodeLibraryDependencies
...
+                            {
+                                if (saved.Version == collected.Version)
+                                {
+                                    packageDependencies[saved].State = PackageDependencyState.Loaded;
```
scottmitchell: One thing that has seemed weird to me here (since even before this PR) is that package state gets set “per node” rather than “per package”. So in the course of this For loop, a package’s state gets set not once, but once per node dependent on the package. When I originally wrote this, I decided this wasn’t really an issue because all of the dependent nodes would agree on the state—it was either “loaded” or “not loaded”. Is this assumption still fair? What about a case where some (but not all) nodes in a missing package are resolved by another package?

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
QilongTang: I am not too sure about the refresh part, but Close() and SwitchTo() will be useful for sure. Refresh() will be highly depending on if the extension author implemented this method

reddyashish: Just saw this, can we call the DependencyRegen function that is in WorkspaceDependencyView.xaml.cs to generate the table again when user action is triggered.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
reddyashish: Good point Scott, we will have to discuss these cases.

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
aparajit-pratap: @reddyashish I think you mean if "Keep Local" is not chosen, in that case the state will be set to "RequiresRestart" in order to uninstall the local package and install the new one, isn't that right?

reddyashish: @aparajit-pratap , yes sorry, that's right.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
reddyashish: @mjkkirschner @scottmitchell .I have added the fix like we discussed.

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
```diff
@@ -3,6 +3,18 @@
 
 namespace Dynamo.Graph.Workspaces
 {
+    /// <summary>
+    /// Enum containing the different types of package dependency states.
+    /// </summary>
+    internal enum PackageDependencyState
+    {
+        Loaded,
```
mjkkirschner: @reddyashish - would mind adding ///summarys for each enum type to describe what it means a bit more specifically?

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
```diff
@@ -146,18 +159,23 @@ internal class PackageDependencyInfo : INodeLibraryDependencyInfo
         /// </summary>
         public bool IsLoaded { get; set; }
```
mjkkirschner: So why does this property still exist?
I don't think we're setting it correctly anymore right? 
here's 2 options:

1. obsolete it and have it return true if State == state.Loaded
2. remove it and cherry pick the removal to 2.3 before we release it.

reddyashish: Yes, I will remove this. But on 2.3 we would needing this to check if is loaded or missing as we do not have these states defined in 2.3, right?

mjkkirschner: @reddyashish you are right, lets obsolete it and set true when state == loaded.

reddyashish: Obsoleted this property and modified 2 tests that were still using this property.

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
reddyashish: Added.

reddyashish:
Addressed all the comments. PTAL.

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
```diff
@@ -118,7 +131,8 @@ interface INodeLibraryDependencyInfo
         /// <summary>
         /// Indicates whether this dependency is loaded in the current session
         /// </summary>
-        bool IsLoaded { get; set; }
+        [Obsolete("This property is obsolete", false)]
+        bool IsLoaded { get; }
```
mjkkirschner: we cannot remove the setter - we'll have to leave it - perhaps you can implement it to set the state to Loaded?

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
```diff
@@ -144,7 +158,18 @@ internal class PackageDependencyInfo : INodeLibraryDependencyInfo
         /// <summary>
         /// Indicates whether this package is loaded in the current session
         /// </summary>
-        public bool IsLoaded { get; set; }
+        [Obsolete("This property is obsolete, use PackageDependencyState property instead", false)]
+        public bool IsLoaded
+        {
+            get {
+                return this.State.Equals(PackageDependencyState.Loaded);
```
mjkkirschner: you should be able to do `==`  as an enum is an int, I would expect it is a value type.

mjkkirschner: https://stackoverflow.com/questions/17492179/c-sharp-enum-equals-vs

src/DynamoCoreWpf/Views/Core/DynamoView.xaml
```diff
@@ -1862,7 +1862,7 @@
                       Grid.Row="2"
                       Grid.RowSpan="2"
                       Height="Auto"
-                      Width="0"
+                      Width="3"
```
mjkkirschner: where is this coming from?

test/DynamoCoreTests/PackageDependencyTests.cs
```diff
@@ -323,7 +364,14 @@ public void DependencyWithTypeLoads()
             // Assert ZTTestPackage is still a package dependency
             var packageDependencies = CurrentDynamoModel.CurrentWorkspace.NodeLibraryDependencies;
             Assert.Contains(new PackageDependencyInfo("ZTTestPackage", new Version("0.0.1")), packageDependencies);
-            Assert.IsTrue(packageDependencies.First().IsLoaded == false);
```
mjkkirschner: so I totally get removing all these - BUT why not leave them so that the change you just made still has tests? Then we can remove the tests when we actually remove the property?

mjkkirschner: Test both things is what I mean.

mjkkirschner:
@reddyashish looking good so far, a few more comments.
Have you considered tests where the state changes during the session?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml
QilongTang: I thought we applied this fix already before.. Was it only added to 2.3 branch?

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
reddyashish: Yes, changed it.

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
reddyashish: Added back the setter and I am setting this when the state is modified. I have to do this because we wanted to check this IsLoaded property in the tests.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml
reddyashish: Actually, this change is in master but I did not reset my head when I was starting this PR. So it came up. https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/Views/Core/DynamoView.xaml#L1865

test/DynamoCoreTests/PackageDependencyTests.cs
reddyashish: Yes, checking both this properties now and we can remove them when removing the actual property.

reddyashish:
@mjkkirschner As of now the state doesn't change until some package is installed or uninstalled manually. We would need the expandable UI element and some action buttons to test these scenarios.

mjkkirschner:
@reddyashish well you could interact with the packageManager extension APIs right?
IE - load a new package and see if the state is changed to Loaded?

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
```diff
@@ -118,6 +131,7 @@ interface INodeLibraryDependencyInfo
         /// <summary>
         /// Indicates whether this dependency is loaded in the current session
         /// </summary>
+        [Obsolete("This property is obsolete", false)]
```
aparajit-pratap: Why can't we just remove this property? These are new, unreleased interfaces and types, right?

mjkkirschner: They are already in 2.3 branch, and cannot be removed form 2.3 - @reddyashish mentioned the conflict above.

aparajit-pratap: I'm sorry I still don't understand, maybe I'm missing something. 2.3 hasn't been released, has it? You mean daily builds have gone out?

mjkkirschner: @aparajit-pratap let me try to clarify.
1. If we remove this API from here - we need to remove it from 2.3 - thats fine, as you said 2.3 is not released.
2. We cannot remove it from 2.3 because the work that @reddyashish did in this PR to replace it is not yet present in that branch.

so - we can either obsolete it. OR cherry pick this entire PR back to 2.3.
I was hoping we could stop working on 2.3 and start testing it. Does this make sense, what do you think?

aparajit-pratap: Sorry, I was assuming this PR is going into 2.3 as well. Thanks for the clarification.

reddyashish:
Merging this and will add the test where the state changes during the session in a new PR.

PR #9854, opened 2019-07-23
DYN-1874: I want to download missing graph package dependencies easily

scottmitchell:
### Purpose

This PR addresses JIRA task [DYN-1874](https://jira.autodesk.com/browse/DYN-1874). It adds an API to the `PackageManagerClientViewModel` to initiate download and install of a package. It also adds the ability for the Dependency Viewer to download missing packages.

DepViewer2

Screen Recording 2019-07-26 at 2 59 20 PM_4

### Declarations

- [ ] ~~Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions), and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.~~ **This feature is in Preview Mode**

### Reviewers

### FYIs


src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -18,6 +20,8 @@ public partial class WorkspaceDependencyView : UserControl
 
         private WorkspaceModel currentWorkspace;
 
+        private DynamoViewModel dynamoViewModel;
```
mjkkirschner: don't do this please! let's figure out what we need to add to the extensions API to accomplish this without passing around the entire viewModel.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
```diff
@@ -441,6 +443,212 @@ public List<PackageManagerSearchElement> ListAll()
...
+
+            ExecutePackage(packageInfo.Name, version, downloadPath);
+        }
+
+        private string JoinPackageNames(IEnumerable<Package> pkgs)
+        {
+            return String.Join(", ", pkgs.Select(x => x.Name + " " + x.VersionName));
+        }
+
+        internal void ExecutePackage(string name, PackageVersion version, string downloadPath)
```
mjkkirschner: what does this name mean?

QilongTang: @scottmitchell You probably mean `DownloadPackage` here?

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
QilongTang: @mjkkirschner We need to access `PackageManagerClientViewModel` to download package here, what do you recommend? Seems we could just expose the download package part to be an API, there are a few examples there, @scottmitchell

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
scottmitchell: Most of this logic was previously in a method called `PackageOnExecuted` on `PackageManagerSearch` before I moved it to the `PackageManagerClient` in order for it to be accessed from other extensions. The name was a vestige of that. I've changed it to `ExecutePackageDownload`.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
```diff
@@ -441,6 +443,212 @@ public List<PackageManagerSearchElement> ListAll()
...
+                var result = MessageBox.Show(string.Format("{0} could not be downloaded.", packageInfo.Name),
+                "Download Error", MessageBoxButton.OKCancel, MessageBoxImage.Question);
+                return;
+            }
+
+            var version = Model.GetGregPackageVersion(header, packageInfo.Version);
+            if (version == null)
+            {
+                // TODO: move message to resources
+                var result = MessageBox.Show(string.Format("Version {0} of {1} could not be found.", packageInfo.Version.ToString(), packageInfo.Name),
```
QilongTang: Do you still intend to finish the TODO in this PR?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
```diff
@@ -441,6 +443,212 @@ public List<PackageManagerSearchElement> ListAll()
...
+                MessageBoxButton.OKCancel, MessageBoxImage.Question);
+
+            var pmExt = DynamoViewModel.Model.GetPackageManagerExtension();
+            if (result == MessageBoxResult.OK)
+            {
+                // get all of the headers
+                var headers = version.full_dependency_ids.Select(dep => dep._id).Select((id) =>
+                {
+                    PackageHeader pkgHeader;
+                    var res = pmExt.PackageManagerClient.DownloadPackageHeader(id, out pkgHeader);
```
QilongTang: Since you are already in the PackageManagerClientViewModel, I think you no longer need to access all the way from extension, `Model` here should be client right?

QilongTang:
A few comments, but looks pretty good so far and thank you for adding row details function in an early stage. Since there are lots of new style changes, have you looked at reusing color defined from Dynamo style sheets instead of using hex codes?

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
```diff
@@ -49,6 +49,11 @@ public IRenderPackageFactory RenderPackageFactory
             get { return dynamoViewModel.RenderPackageFactoryViewModel.Factory; }
         }
 
+        public PackageManagerClientViewModel PackageManagerClientViewModel
```
mjkkirschner: I don't think we should do this either 😉  - sorry for not being able to respond earlier - 
If there's no better way, than okay, but lets think more generally, are we going to do this for every view Model in Dynamo? It exposes too much of the application IMO and doesn't help people find what they need.

This seems like a very specific large surface to expose, and it exposes quite a lot of APIs ... 
What about adding just the apis you need (publish, download etc?) or event an event to request those?

mjkkirschner: another way to think about it is to follow the example set out by the rest of the extension API - everything is exposed is an interface that describes a set of small apis that we consider required or useful, then you expose the object, but only as that interface, that way you control what apis we support.

mjkkirschner: look at `IworkspaceModel` as an example, the extensions interfaces don't expose `WorkspaceModel` but give users an `IworkspaceModel` - they can cast this if they choose, but API users are making a decision to trade some safety for flexibility, and that is more clear.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -170,137 +118,81 @@ internal void DependencyRegen(WorkspaceModel ws)
...
+            packageManagerClientViewModel.InitiatePackageDownloadAndInstall(package);
+            DependencyRegen(currentWorkspace);
         }
     }
 
-    /// <summary>
-    /// Class defining data for each column
-    /// </summary>
-    public class ColumnData
+    public class NodeLibraryDependencyRow
```
mjkkirschner: did something weird happen to this file? The diff is pretty hard to read, it looks like a bunch of public classes were deleted? Or am I mis-reading the diff?

mjkkirschner: (you can't delete the public classes if they get released with another version - however short lived that version is) it's safer to just obsolete them as we don't know what binary compatibility issues this will cause without extensive testing, and it's just not worth it IMO.

QilongTang: @mjkkirschner Do you remember the last case this was an issue? This class is only used for data binding to the UI so we can't make it internal. If any class went out with a beta release that we can no longer rename, that sounds like too strict..

mjkkirschner: I'm not talking about beta releases.

If it's in 2.3 then we should consider it stable, unless @scottmitchell wants to refactor 2.3 before release to not include this class and then redo testing there.

QilongTang: Sure, I guess I am wondering if the API restriction need to be applied on extension level. I understand once it is in RC branch we should make it stable but this dll is not in our public nuget so I do not expect developers to consume it nor us maintaining the stability.. My 2 cents..

mjkkirschner: @QilongTang - that's a great great point! I think it's acceptable to make this entire .dll **not** part of the API.

👍 

I think we'll need to consider what consequences this has technically and also for build processes though.

QilongTang: @mjkkirschner I don't think we have it specified according quickly glancing at https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions or https://github.com/DynamoDS/Dynamo/wiki/API-Changes. And seems we also missed 2.x versions' API changes...

mjkkirschner: regarding the API Changes doc... I find it kind of useless work - the API diff job now computes this for us - or something like
https://www.fuget.org/packages/DynamoVisualProgramming.Core/2.2.1.5204/lib/net47/diff/2.1.0.7465/

is SOO much better than that page - it's only missing the explanations/readme portion.

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
scottmitchell: This makes sense. All we need to surface right now is "DownloadPackage". Let's discuss briefly tomorrow.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
scottmitchell: Yes, thanks for pointing this out. I have moved them now

scottmitchell:
@mjkkirschner @QilongTang I think I've addressed all the comments in this PR. In terms of the API / semantic versioning issue for the WorkspaceDependencyViewExtension--I think this will be addressed by release notes and by a PR to 2.3 marking the view extension (and/or its classes) as obsolete/beta.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -170,137 +118,112 @@ internal void DependencyRegen(WorkspaceModel ws)
...
-        /// Data in each cell
-        /// </summary>
-        public string Data
-        {
-            get; set;
-        }
+                switch (DependencyInfo.State)
+                {
+                    case PackageDependencyState.Loaded:
+                        message = string.Format("{0} {1} is installed locally.",
```
scottmitchell: @mjkkirschner @QilongTang While this view extension is in "Preview", can we leave user facing strings in the code rather than move them to resources? I imagine these messages will be changes as we add other features such as "Keep Local"

mjkkirschner: thats okay with me as long as tasks are filed and TODOs added referencing them.

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
```diff
@@ -85,7 +101,7 @@ public override string ToString()
         }
     }
 
-    internal enum ReferenceType
+    public enum ReferenceType
```
mjkkirschner: does this need to be public?

src/DynamoCoreWpf/ViewModels/PackageManager/IPackageDownloader.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// <summary>
+    /// An interface containing operations for installing Dynamo packages
+    /// </summary>
+    public interface IPackageInstaller
+    {
+        /// <summary>
+        /// Initiates download and install of a package
+        /// </summary>
+        /// <param name="package"></param>
+        /// <param name="downloadPath"></param>
```
mjkkirschner: can you add comments here? If there is nothing to say about these params leave them out of the /// comments.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -170,137 +118,112 @@ internal void DependencyRegen(WorkspaceModel ws)
...
+                string iconPath;
+
+                switch (DependencyInfo.State)
+                {
+                    case PackageDependencyState.Loaded:
+                        iconPath = "NodeLibraryDependency_Loaded.png";
+                        break;
+
+                    case PackageDependencyState.Missing:
+                        iconPath = "NodeLibraryDependency_Missing.png";
```
mjkkirschner: would it be possible to move these to a resx file? I guess we can do it as as followup - raw resources like this are difficult to manage and make patch installer issues much worse.

mjkkirschner: for example look at:
https://github.com/DynamoDS/Dynamo/blob/master/src/Libraries/GeometryUIWpf/ProtoGeometryImages.resx

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
```diff
@@ -165,7 +167,7 @@ private void ShowTermsOfUseForPublishing()
     ///     A thin wrapper on the Greg rest client for performing IO with
     ///     the Package Manager
     /// </summary>
-    public class PackageManagerClientViewModel : NotificationObject
+    public class PackageManagerClientViewModel : NotificationObject, IPackageInstaller
```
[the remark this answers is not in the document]
mjkkirschner: I don't think it is - A new method has been added, and this object now implements a new interface with that method. What API does it break?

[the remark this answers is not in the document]
aparajit-pratap: If there are derived classes they would need to implement the new interface.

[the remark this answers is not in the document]
aparajit-pratap: I guess it might be okay. I take it back. API breaking rules always confound me.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
scottmitchell: Ok, sounds like less work to just go ahead and move them to resources, so I'll do that

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
scottmitchell: Nope. Good point. Fixing this.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
scottmitchell: Moved to resx

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
scottmitchell: @mjkkirschner I spent some time trying to figure this out last night, but couldn't quite get it to work. If it's alright with you, I'd like to deal with this in a follow up UI pass.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
```diff
@@ -441,6 +443,218 @@ public List<PackageManagerSearchElement> ListAll()
...
+
+            ExecutePackageDownload(packageInfo.Name, version, downloadPath);
+        }
+
+        private string JoinPackageNames(IEnumerable<Package> pkgs)
+        {
+            return String.Join(", ", pkgs.Select(x => x.Name + " " + x.VersionName));
+        }
+
+        internal void ExecutePackageDownload(string name, PackageVersion version, string downloadPath)
```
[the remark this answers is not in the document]
scottmitchell: I agree this would be better, but I'd like to wrap up this task so I'm not going to do this right now.

scottmitchell:
@aparajit-pratap @mjkkirschner Thank you for the comments. I believe I have addressed them all--the one thing I did not fix here that will need a follow up is moving the package state icons to a resx file

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
scottmitchell: @mjkkirschner Finally got it. Turns out we have an existing utility for converting bitmaps to ImageSources: [https://github.com/DynamoDS/Dynamo/blob/ec10f936824152e7dd7d6d019efdcda0d78a5264/src/DynamoCoreWpf/Utilities/ResourceUtilities.cs](https://github.com/DynamoDS/Dynamo/blob/ec10f936824152e7dd7d6d019efdcda0d78a5264/src/DynamoCoreWpf/Utilities/ResourceUtilities.cs)

scottmitchell:
@mjkkirschner I have moved the dependency state icons to a resx, and also added a "Package Manager Terms of Use" check to the new package download API (thanks for pointing this out @kronz). This PR should be good to go now.

QilongTang:
The TOU check and image changes looks good. Thank you for removing the static images so they will not pollute binary compatibility.  Merging for testing in 2.5.

PR #9855, opened 2019-07-23
this test fails when run alone

mjkkirschner:
set the path to include core directory
## Purpose

This test requires sharpdx dlls at runtime, but the path has not been set to find them. This code was copied from the startup sequence of the helixWatch3dViewModel tests.

Screen Shot 2019-07-23 at 6 10 59 PM

### Declarations

### Reviewers

@QilongTang


QilongTang:
LGTM

PR #9858, opened 2019-07-24
Remove Debug Assertions from CallSite causing VM crash in Debug mode

reddyashish:
# Purpose

This PR is to fix the crash that was reported in this task:
 https://jira.autodesk.com/browse/DYN-2055

The debug assertions are removed and the output is correct after that. 

### Declarations

### Reviewers


src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -8,13 +15,6 @@
 using System.Runtime.Serialization;
 using System.Runtime.Serialization.Formatters.Soap;
 using System.Text;
-using ProtoCore.DSASM;
-using ProtoCore.Exceptions;
-using ProtoCore.Lang;
-using ProtoCore.Lang.Replication;
-using ProtoCore.Properties;
-using ProtoCore.Runtime;
-using ProtoCore.Utils;
```
aparajit-pratap: I think the System `using`s should be placed before ours, @QilongTang ?

mjkkirschner: I sense a code standards fight coming.

QilongTang: @mjkkirschner @aparajit-pratap I think a while ago I have made it settled, I am sorry I probably at some point told @reddyashish the wrong info... Please refer to https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards for the only source of truth.

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -174,11 +174,6 @@ public ISerializable GetLeftMostData()
 
                 if (!HasNestedData)
                     return null;
-#if DEBUG
-
-                Validity.Assert(NestedData != null, "Nested data has changed null status since last check, suspected race");
-                Validity.Assert(NestedData.Count > 0, "Empty subnested array, please file repo data with @lukechurch, relates to MAGN-4059");
```
aparajit-pratap: @reddyashish I think it makes sense to keep the `NestedData != null` assertion and just remove the second one as there is a check for `NestedData.Count` to be zero in the next line.

src/Engine/ProtoCore/Lang/CallSite.cs
reddyashish: @aparajit-pratap Changed it to place the System using's first

aparajit-pratap:
LGTM

reddyashish:
Merging this, @mjkkirschner this is not going in 2.4 right?
Task: https://jira.autodesk.com/browse/DYN-2055

PR #9884, opened 2019-08-06
Add support for loading ASM from RSA install location

aparajit-pratap:
### Purpose

JIRA: https://jira.autodesk.com/browse/DYN-2047

### Declarations

### Reviewers

@QilongTang @mjkkirschner

### DIFF
--- a/src/Tools/DynamoInstallDetective/ProductLookUp.cs
+++ b/src/Tools/DynamoInstallDetective/ProductLookUp.cs
@@ -201,8 +201,7 @@ namespace DynamoInstallDetective
         public virtual IEnumerable<string> GetProductNameList()
         {
             var key = OpenKey(REG_KEY64);
-            var list = key.GetSubKeyNames().Where(s => s.Contains(ProductLookUpName));
-            return list;
+            return key.GetSubKeyNames().Where(s => s.Contains(ProductLookUpName));
         }
 
         public virtual bool ExistsAtPath(string basePath)
--- a/src/Tools/DynamoInstallDetective/ProductLookUp.cs
+++ b/src/Tools/DynamoInstallDetective/ProductLookUp.cs
@@ -159,7 +159,7 @@ namespace DynamoInstallDetective
         public InstalledProductLookUp(string lookUpName, string fileLookup)
         {
             this.ProductLookUpName = lookUpName;
-            this.fileLocator = (path) => Directory.GetFiles(path, fileLookup).FirstOrDefault();
+            this.fileLocator = (path) => Directory.GetFiles(path, fileLookup, SearchOption.AllDirectories).FirstOrDefault();
         }
 
         public InstalledProductLookUp(string lookUpName, Func<string, string> fileLocator)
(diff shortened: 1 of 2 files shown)


src/Tools/DynamoShapeManager/Utilities.cs
```diff
@@ -14,7 +14,7 @@ public static class Utilities
         /// <summary>
         /// Key words for Products containing ASM binaries
         /// </summary>
-        private static readonly List<string> ProductsWithASM = new List<string>() { "Revit", "Civil", "FormIt" };
+        private static readonly List<string> ProductsWithASM = new List<string>() { "Revit", "Civil", "Robot Structural Analysis", "FormIt" };
```
aparajit-pratap: ASM from RSA will be used if there is no Revit or Civil3D installed on the user's machine.

src/Tools/DynamoInstallDetective/ProductLookUp.cs
```diff
@@ -159,7 +159,7 @@ static string GetDisplayName(RegistryKey key)
         public InstalledProductLookUp(string lookUpName, string fileLookup)
         {
             this.ProductLookUpName = lookUpName;
-            this.fileLocator = (path) => Directory.GetFiles(path, fileLookup).FirstOrDefault();
+            this.fileLocator = (path) => Directory.GetFiles(path, fileLookup, SearchOption.AllDirectories).FirstOrDefault();
```
QilongTang: @aparajit-pratap Did you notice any noticeable slow down with such searching option with RSA?

src/Tools/DynamoShapeManager/Utilities.cs
QilongTang: I forgot if changing the order here affect the ASM paths found?

src/Tools/DynamoInstallDetective/ProductLookUp.cs
aparajit-pratap: I didn't notice any difference as I don't suspect there is a noticeable slowdown if any after this change. In any case, I think this part of the fix might be unavoidable due to the fact that there is no guarantee that ASM will always be found in the root directory of an installation.

src/Tools/DynamoShapeManager/Utilities.cs
aparajit-pratap: Yes, it does. See: https://github.com/DynamoDS/Dynamo/blob/7a392ef25a532e0782caf57876aa43f0e4fc7ac2/src/Tools/DynamoShapeManager/Utilities.cs#L154

We first iterate over the different LibG versions and then over the supported products list and whichever matches first, comes out the winner. So it does matter if we play with the order. For example if Revit and RSA are both installed, and there is a match, ASM will be loaded from Revit and if we swap Revit with RSA, ASM will be loaded from RSA.

QilongTang: Thanks, this is refreshing to me. Keeping the order of Revit first looks safe.

src/Tools/DynamoInstallDetective/ProductLookUp.cs
QilongTang: @aparajit-pratap Thank you, sounds reasonable here.

aparajit-pratap:
Thanks @QilongTang for the review.

PR #9895, opened 2019-08-09
Keep Local Package dependency

QilongTang:
Please Note:

### Purpose

PR for keeping local package dependency.

This PR:
1. Make all resource string public instead of internal, this was causing crash for referencing resources string in xmal, I wish I know ealier!
2. Fix a few UI interaction issue in workspace dependency view extension
- [x] View extension always loaded when there is something worng, no only limit to missing case
- [x] Show both install button and keep local button for incorrect version of package loaded
- [x] Moved some resource string back from dynamo.core.wpf back to extension project so in the future we can split this one out
- [x] To finish updating package info in workspace model in memory
- [x] Expose `PackageLoader` and only its `LoadedPackages` to all extension developers via `ViewLoadedParams`
- [ ] To make sure people can unfold info, if they want to ignore

image

DependencyViewEx_KeepLocal

### Declarations

Running self CI

### Reviewers

@DynamoDS/dynamo 

### FYIs
@Amoursol


src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
```diff
@@ -175,7 +175,18 @@ internal class PackageDependencyInfo : INodeLibraryDependencyInfo, IPackageInfo
...
+        public Version Version
+        {
+            get
+            {
+                return PackageInfo.Version;
+            }
+            set
+            {
+                if(PackageInfo.Version != value)
+                    PackageInfo.Version = value;
```
QilongTang: This definitely works but unfortunately is the wrong way for sure. I would like to discuss with the team the right approach here, one way I can think of is have an internal method to update it and make the assembly internal visible to dependency view extension only. Unless we make it visible to any extension. What do you think @DynamoDS/dynamo

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
```diff
@@ -57,6 +59,14 @@ public IPackageInstaller PackageInstaller
             get { return dynamoViewModel.PackageManagerClientViewModel; }
         }
 
+        /// <summary>
+        /// A reference to package loader to query for certain packages info
+        /// </summary>
+        public IPackageLoader PackageLoader
+        {
+            get { return dynamoViewModel.Model.GetPackageManagerExtension().PackageLoader; }
```
aparajit-pratap: This property is being obtained from a specific extension, namely, the PM extension in this case, still `PackageLoader` is being made generically available to all extensions. It seems a little counter-intuitive, no?

src/WorkspaceDependencyViewExtension/Properties/Resources.resx
```diff
@@ -117,231 +117,246 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="DetailsMessageIncorrectVersion" xml:space="preserve">
+    <value>Some nodes in this graph were originally created with {0} v{1} which you do not have installed. Instead you have a different version of {0} installed, would you keep the installed version?</value>
```
aparajit-pratap: `Some nodes in this graph were originally created with {0} v{1}, which you do not have installed. Instead you have a different version of {0} installed. Do you wish to keep the installed version?`

src/WorkspaceDependencyViewExtension/Properties/Resources.resx
```diff
@@ -117,231 +117,246 @@
...
   <data name="DetailsMessageLoaded" xml:space="preserve">
-    <value>{0} {1} is installed locally.</value>
+    <value>{0} {1} is installed and loaded.</value>
   </data>
   <data name="DetailsMessageMissing" xml:space="preserve">
     <value>This graph depends on {0} {1}. Would you like to download and install this package?</value>
   </data>
   <data name="DetailsMessageWarning" xml:space="preserve">
-    <value>Some nodes in this graph were originally created with {0} {1} which you do not have installed locally. You may have a different version of {0} installed, or you may have a different package which resolves these nodes.</value>
+    <value>Some nodes in this graph were originally created with {0} v{1} which you do not have installed. You may have a different version of {0} installed, or you may have a different package which resolves these nodes.</value>
```
aparajit-pratap: `Some nodes in this graph were originally created with {0} v{1}, which you do not have installed. `

src/WorkspaceDependencyViewExtension/Properties/Resources.resx
```diff
@@ -117,231 +117,246 @@
...
+        WF4v75ZPCx2Y/4f9b+/X2K/lodKzMJSesw7Ki+XlH/ADZNMLeg3pORhK0okwlKQTYShJJ8JQkk6EoSSd
+        CENJOhGGknQiDCXpRBhK0okwlKQTYShJJ8JQkk6EoSSdCENJOhGGknQiDCXpRBhK0okwlKTzfPnunwM4
+        jDoyqebiAAAAAElFTkSuQmCC
 </value>
   </data>
+  <data name="ProvideFeedbackButton" xml:space="preserve">
+    <value>Provide Feedback</value>
+  </data>
+  <data name="ProvideFeedbackError" xml:space="preserve">
+    <value>Could not re-direct to the Dynamo forum page for feedback:</value>
```
aparajit-pratap: `redirect`

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
aparajit-pratap: I don't get the need for adding the `IPackageLoader` interface since `PackageLoader` type is accessible here.

QilongTang: image

This is so extension author will not get other members of package loader

src/WorkspaceDependencyViewExtension/Properties/Resources.resx
QilongTang: Will update

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
QilongTang: @aparajit-pratap @mjkkirschner Let me know what you think on this one, otherwise this PR is good for review with tests

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
mjkkirschner: I'm sorry @QilongTang I'm confused, why do you need to do this - why does this need a setter at all? Can't we just make a new PackageDependencyInfo object?

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
```diff
@@ -57,6 +59,14 @@ public IPackageInstaller PackageInstaller
             get { return dynamoViewModel.PackageManagerClientViewModel; }
         }
 
+        /// <summary>
+        /// A reference to package loader to query for certain packages info
+        /// </summary>
+        public IPackageLoader PackageLoader
```
mjkkirschner: It doesn't really make sense to me that this is only added to View extensions -

src/DynamoPackages/Interfaces/IPackageLoader.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Collections.Generic;
+using Dynamo.PackageManager;
+
+namespace DynamoPackages.Interfaces
+{
+    /// <summary>
+    /// This interface is for PackageLoader. It can be expanded if needed,
+    /// currently only exposed inViewLoadedParams and referenced in WorkspaceDependencyViewExtension
```
mjkkirschner: this seems like a strange comment to show to users.

src/DynamoPackages/Interfaces/IPackageLoader.cs
```diff
@@ -0,0 +1,17 @@
...
+    /// </summary>
+    public interface IPackageLoader
+    {
+        /// <summary>
+        /// List of local packages
```
mjkkirschner: to be part of an interface this needs more description, what if someone else is trying to implement this interface?

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
QilongTang: You mean?

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -145,6 +145,32 @@ private void DownloadPackage(object sender, RoutedEventArgs e)
...
+
+        /// <summary>
+        /// Handler of button which user click when choosing to keep the
+        /// installed version of package instead of the specified one.
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="e"></param>
+        private void KeepLocalPackage(object sender, RoutedEventArgs e)
+        {
+            var info = ((PackageDependencyRow)((Button)sender).DataContext).DependencyInfo;
```
mjkkirschner: all the casting makes me nervous - should we be more defensive here? or wrap in a try catch?

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
mjkkirschner: You can add it to model extensions as well by adding it to `ReadyParams`

src/DynamoPackages/Interfaces/IPackageLoader.cs
QilongTang: I guess my question would be do we want to continue use the interface this way similar to how we expose PackageManagerClientViewModel. I believe @aparajit-pratap has some concerns. I would like to have an agreement first, I can totally remove the interface and only expose a list of local packages in the LoadedParam but that can be out of the date quickly and obsolete later.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
QilongTang: hmm This private method is directly bind to the button so there should be no surprise here.. But I guess to be safe I can do it..

src/DynamoPackages/Interfaces/IPackageLoader.cs
QilongTang: See my comments below

src/DynamoPackages/Interfaces/IPackageLoader.cs
mjkkirschner: yeah, I'm not sure, @scottmitchell and I talked a lot about adding the other interface which is weirder as the idea was it would contain more UI centric functionality like install,remove - this is just a query, so I'm not sure we want to wrap them up together.


as they are now one can be used from modelExtensions as well, if we put them together we no longer have that flexibility.

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
aparajit-pratap: We can directly access `LocalPackages` in the `WorkspaceDependencyViewExtension` if that is what is needed by exposing it from `ReadyParams`:
`dynamoModel.GetPackageManagerExtension().PackageLoader.LocalPackages`?

mjkkirschner: Yep, that works too, I think the idea with adding the interface is to start grouping some of these properties together, so that we don't have these parameter classes with a bunch of disorganized properties floating around.

QilongTang: I still believe exposing package loader is more useful than local packages because packages list is pretty much useless for other extensions because they can't do anything other than list the info which are already done in `manage packages` menu item. But seems like you have a strong opinion on this @aparajit-pratap. I will remove the interface until we have a better less confusing solution.

src/DynamoPackages/Interfaces/IPackageLoader.cs
QilongTang: Ended removing the interface, marking as resolved

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
QilongTang: Good point, let me try

QilongTang: @mjkkirschner That way it is even more complex than this approach, because we will need to map the nodes property which is also read-only, and will need to switch the old info out and replace it with the new one. The whole process is not possible until we expose the setter for nodes which is not a good trade off.

QilongTang:
Ready for another round of review

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
mjkkirschner: I think you should be careful before doing this because it breaks some invariants made previously -
The same way that now updating this is useful because the nodeInfo objects have been updated - can also be harmful - because some code may have relied on this PackageInfo not changing - 

I'm worried specifically about the customNode package versioning check stuff.

mjkkirschner:
@QilongTang do all the tests pass?
I wonder if you can add a pubic updatePackageVersion method to the `PackageDependencyInfo` class which explicitly updates the packageInfo with a new object and updates all nodeInfo references?

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
```diff
@@ -57,6 +58,14 @@ public IPackageInstaller PackageInstaller
             get { return dynamoViewModel.PackageManagerClientViewModel; }
         }
 
+        /// <summary>
+        /// A reference to list of Local loaded packages
+        /// </summary>
+        public System.Collections.Generic.IEnumerable<Package> LocalPackages
```
mjkkirschner: I find this super weird - Can you avoid adding anything here for now and lets just get the ExtensionsManager and then the PackageManagerExtension to avoid polluting this class until we have a plan?

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
mjkkirschner: @QilongTang @aparajit-pratap - here is what I mean:
https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/Extensions/ViewStartupParams.cs#L13

you can access the ExtensionsManager from your extension, then get the packageManager, then get the packageLoader - without adding anything to ViewReadyParams.

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
QilongTang: @mjkkirschner I guess we need to also update the version CBN in this case.. seems whether or not create a new info object does not matter but we need to make sure all the record in memory should be aligned...

QilongTang:
@mjkkirschner All tests pass, from our conversation, it seems pretty safe for this change

test/DynamoCoreWpfTests/ViewExtensions/WorkspaceDependencyViewExtensionTests.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            // Adding the workspace dependency view extension will 
+            // not add a dup tab in the extensions side bar
+            extensionManager.Add(viewExtension);
+            Assert.AreEqual(initialNum, View.TabItems.Count);
+        }
+
+        public static void RaiseLoadedEvent(FrameworkElement element)
+        {
+            MethodInfo eventMethod = typeof(FrameworkElement).GetMethod("OnLoaded",
```
aparajit-pratap: Where is `OnLoaded` defined?

test/DynamoCoreWpfTests/ViewExtensions/WorkspaceDependencyViewExtensionTests.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            // Once choosing to keep the installed loaded version, info.Version should reflect the lower version
+            viewExtension.DependencyView.UpdateWorkspaceToUseInstalledPackage(info);
+            Assert.AreEqual("0.1.2", info.Version.ToString());
+        }
+
+        [Test]
+        public void WillNotAddDupWorkspaceDependencyViewExtensionLoadTest()
+        {
+            RaiseLoadedEvent(this.View);
```
aparajit-pratap: Could you explain what this method does?

test/DynamoCoreWpfTests/ViewExtensions/WorkspaceDependencyViewExtensionTests.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            var extensionManager = View.viewExtensionManager;
+
+            var initialNum = View.TabItems.Count;
+
+            // Adding the workspace dependency view extension will 
+            // not add a dup tab in the extensions side bar
+            extensionManager.Add(viewExtension);
```
aparajit-pratap: Why will there be a dup in this case and why is it not incremented? I guess I do not understand this test.

QilongTang: @aparajit-pratap The extension is already loaded when Dynamo is launched but the view will not show unless you open a graph, so calling the function again will not inject the extension tab because it is already there

test/DynamoCoreWpfTests/ViewExtensions/WorkspaceDependencyViewExtensionTests.cs
QilongTang: I would defer to @reddyashish in this case since I was using his viewExtensions class as template

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -145,6 +146,40 @@ private void DownloadPackage(object sender, RoutedEventArgs e)
...
+        internal void UpdateWorkspaceToUseInstalledPackage(PackageDependencyInfo info)
+        {
+            var pmExtension = dependencyViewExtension.ViewLoadedParams.ExtensionManager.Extensions.Where(x => x.Name == "DynamoPackageManager").FirstOrDefault() as PackageManagerExtension;
+            if (pmExtension != null)
+            {
+                info.Version = new Version(pmExtension.PackageLoader.LocalPackages.Where(x => x.Name == info.Name).First().VersionName);
+                info.State = PackageDependencyState.Loaded;
+                // Mark the current workspace dirty for save
+                currentWorkspace.HasUnsavedChanges = true;
+                DependencyRegen(currentWorkspace);
```
scottmitchell: @QilongTang Is the expected behavior here for the `Keep Local` change to be serialized to the dyn as well? If so, I think you will need to remove (or replace) references to the old version of the package in the current workspace's  [nodePackageDictionary](https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs#L617).

QilongTang: @scottmitchell Since we are updating the reference here directly, the package info in dictionary is updated at the same time, from my debugging we are good

scottmitchell: Oh good point!

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -145,6 +146,40 @@ private void DownloadPackage(object sender, RoutedEventArgs e)
...
+        /// <param name="sender"></param>
+        /// <param name="e"></param>
+        private void KeepLocalPackage(object sender, RoutedEventArgs e)
+        {
+            try
+            {
+                var info = ((PackageDependencyRow)((Button)sender).DataContext).DependencyInfo;
+                UpdateWorkspaceToUseInstalledPackage(info);
+            }
+            catch(Exception) {}
```
mjkkirschner: lets atleast log the exception?

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -145,6 +146,40 @@ private void DownloadPackage(object sender, RoutedEventArgs e)
...
+        }
+
+        /// <summary>
+        /// Update current workspace in memory to keep the installed package
+        /// instead of keep referencing the dependency info saved in DYN
+        /// </summary>
+        /// <param name="info">Target PackageDependencyInfo to update version</param>
+        internal void UpdateWorkspaceToUseInstalledPackage(PackageDependencyInfo info)
+        {
+            var pmExtension = dependencyViewExtension.ViewLoadedParams.ExtensionManager.Extensions.Where(x => x.Name == "DynamoPackageManager").FirstOrDefault() as PackageManagerExtension;
```
mjkkirschner: you can likely make this safer and shorter by using `OfType<PackageManagerExtension>()`

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -145,6 +146,40 @@ private void DownloadPackage(object sender, RoutedEventArgs e)
...
+            if (pmExtension != null)
+            {
+                info.Version = new Version(pmExtension.PackageLoader.LocalPackages.Where(x => x.Name == info.Name).First().VersionName);
```
mjkkirschner: can you use FirstOrDefault and then check for null?

src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
```diff
@@ -35,12 +36,14 @@ public string UniqueId
...
-        private WorkspaceDependencyView DependencyView
+        internal WorkspaceDependencyView DependencyView
         {
             get;
             set;
         }
 
+        internal ViewStartupParams ViewLoadedParams;
+
         private ReadyParams ReadyParams;
```
mjkkirschner: in general I think cacheing these is kind of worrisome, I have no idea but seems like it might cause some kind of reference path issue (memory leak)

test/DynamoCoreWpfTests/ViewExtensions/WorkspaceDependencyViewExtensionTests.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            var loadedParams = new ViewLoadedParams(View, ViewModel);
+            viewExtension.DependencyView = new WorkspaceDependencyView(viewExtension, loadedParams);
+
+            System.Version infoVersion = new System.Version("0.1.3");
+            PackageDependencyInfo info = new PackageDependencyInfo("GetHighest", infoVersion);
+
+            // Once choosing to keep the installed loaded version, info.Version should reflect the lower version
+            viewExtension.DependencyView.UpdateWorkspaceToUseInstalledPackage(info);
```
mjkkirschner: @QilongTang I think what is missing here is a test which serializes the dyn and asserts the dependencies block is as expected - can you add that before we merge?

test/DynamoCoreWpfTests/ViewExtensions/WorkspaceDependencyViewExtensionTests.cs
QilongTang: @reddyashish Is this required?

test/DynamoCoreWpfTests/ViewExtensions/WorkspaceDependencyViewExtensionTests.cs
QilongTang: Will alter this test to include file save check

test/pkgs/Dynamo Samples/extra/CustomRenderExample.dyn
```diff
@@ -0,0 +1,84 @@
...
+      "Replication": "Auto",
+      "Description": "CustomRenderExample.Create ( ): CustomRenderExample"
+    }
+  ],
+  "Connectors": [],
+  "Dependencies": [],
+  "NodeLibraryDependencies": [
+    {
+      "Name": "Dynamo Samples",
+      "Version": "2.0.1",
```
QilongTang: Notice the graph is serialized with a higher version, so opening this graph in test will display dependencyViewExtension so we can test the functionality there

mjkkirschner:
LGTM

test/DynamoCoreWpfTests/ViewExtensions/WorkspaceDependencyViewExtensionTests.cs
QilongTang: @aparajit-pratap We discussed this and found this is required when you want view extension to behave like how it is loaded normally during the test

PR #9899, opened 2019-08-11
Fix load asm

aparajit-pratap:
### Purpose

Attempted fix for JIRA: https://jira.autodesk.com/browse/DYN-2047. Fix loading of installed products that are in a non-root directory. Optimize product lookup using parallel, recursive search in all sub-directories. 

### Declarations

### Reviewers

@mjkkirschner @QilongTang


src/Tools/DynamoInstallDetective/ProductLookUp.cs
```diff
@@ -158,8 +161,16 @@ static string GetDisplayName(RegistryKey key)
...
         public InstalledProductLookUp(string lookUpName, string fileLookup)
         {
-            this.ProductLookUpName = lookUpName;
-            this.fileLocator = (path) => Directory.GetFiles(path, fileLookup).FirstOrDefault();
+            ProductLookUpName = lookUpName;
+            fileLocator = (path) =>
+            {
+                if (string.IsNullOrEmpty(corePath))
+                {
+                    corePath = Directory.GetFiles(path, fileLookup, SearchOption.AllDirectories)
```
mjkkirschner: why don't you use `enumerateFiles` which will return an Ienumerable, then when you find the first matching file you can stop the lookup instead of returning all matching files which will require the search to march all the nested subdirectories even if you already found the file.
https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.enumeratefiles?view=netframework-4.8

mjkkirschner:
I guess to actually see if this works we need to merge this and test the complete install build? Since the self serve seems like it passes?

I'm wondering if it's not less risky to just add a flag to sandbox to select the exact path -for example does this have the potential to slow down Formit Dynamo startup time since they boot CLI or Sandbox which will both do this preload step each time...

aparajit-pratap:
We will soon find out. I've triggered another self-serve with the change of `EnumerateFiles` to compare with `GetFiles`. I think we should be good if the self-serve time to run tests does not change much from what it was before the inclusion of RSA.

mjkkirschner:
@aparajit-pratap okay - but I believe the self serve machine you are running against only has one copy of Revit installed - this means the search time will be potentially less than on other machines with multiple products - in any case hopefully this solution will still prove fast for products that have their dlls in the root.

PR #9907, opened 2019-08-13
DYN-2002 Localize the node description tooltip.

reddyashish:
### Purpose

This PR is in reference to this task: https://jira.autodesk.com/browse/DYN-2002. 

In short after this fix, the description tooltip (for the nodes in the workspace) will be converted to the localized language. We will still be serializing this property but when deserializing we do not read this value from the JSON file. The value will be picked up from node's xml config and will be converted to the localized language. 

Screen Shot 2019-08-13 at 12 24 21 PM

### Declarations

### Reviewers

### DIFF
--- a/src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
+++ b/src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
@@ -376,7 +376,9 @@ namespace Dynamo.Graph.Workspaces
         }
     }
 
-    // Converter for description property. 
+    ///<Summary>
+    ///  Converter for Description property in the NodeModel class.
+    ///</Summary>
     public class DescriptionConverter : JsonConverter
     {
         public override bool CanConvert(Type objectType)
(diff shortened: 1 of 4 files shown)


mjkkirschner:
@reddyashish nice, can we also add a test for this?
The test can open a pre made .dyn file with a description set to "BADDESCRIPTION" or something like that - and then make sure the description of the loaded node is not set to that.

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
```diff
@@ -376,6 +376,31 @@ public override bool CanWrite
...
+        {
+            return (objectType == typeof(String));
+        }
+
+        /// When deserializing, we do not want to read this property from the file
+        /// so null is being returned. This is to convert the Description property
+        /// to the localized language. 
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            return null;
```
QilongTang: If possible, can we use `string.empty` instead?

reddyashish: @QilongTang If I am returning empty string instead of null value, it is overwriting the node's description to an empty string and that is being deserialized. The value from node's config is not being picked up.

reddyashish:
@mjkkirschner I have added the test. There are no failures on self-service but do you still think any of the tests have to be updated?

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
QilongTang: Thanks! Good to know!

PR #9915, opened 2019-08-16
update dev view UI when package installed.

mjkkirschner:
set package loader to update dep UI
add this outisde the view to make it clear this is intended to be a singleton view
get rid of properties that were not set

### Purpose

update dep UI when package is loaded to catch new installations.
also gets rid of ready params from extension - those are only set on model extensions...

### Declarations

### Reviewers
@QilongTang 
### FYIs

@scottmitchell

### DIFF
--- a/src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
+++ b/src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
@@ -1,12 +1,12 @@
-﻿using Dynamo.Extensions;
+﻿using System;
+using System.Linq;
+using System.Windows.Controls;
+using Dynamo.Extensions;
 using Dynamo.Graph.Workspaces;
 using Dynamo.Logging;
 using Dynamo.PackageManager;
 using Dynamo.WorkspaceDependency.Properties;
 using Dynamo.Wpf.Extensions;
-using System;
-using System.Linq;
-using System.Windows.Controls;
 
 namespace Dynamo.WorkspaceDependency
 {
@@ -86,6 +86,7 @@ namespace Dynamo.WorkspaceDependency
             DependencyView = new WorkspaceDependencyView(this, viewLoadedParams);
             // when a package is loaded update the DependencyView 
             // as we may have installed a missing package.
+
             pmExtension.PackageLoader.PackgeLoaded += (package) =>
             {
                 DependencyView.DependencyRegen(viewLoadedParams.CurrentWorkspaceModel as WorkspaceModel);
--- a/src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
+++ b/src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
@@ -1,12 +1,12 @@
-﻿using System;
-using System.Linq;
-using System.Windows.Controls;
-using Dynamo.Extensions;
+﻿using Dynamo.Extensions;
 using Dynamo.Graph.Workspaces;
 using Dynamo.Logging;
 using Dynamo.PackageManager;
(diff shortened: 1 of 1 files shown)


src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
```diff
@@ -1,12 +1,12 @@
...
-using Dynamo.Extensions;
+﻿using Dynamo.Extensions;
 using Dynamo.Graph.Workspaces;
 using Dynamo.Logging;
 using Dynamo.PackageManager;
 using Dynamo.WorkspaceDependency.Properties;
 using Dynamo.Wpf.Extensions;
+using System;
+using System.Linq;
+using System.Windows.Controls;
```
QilongTang: https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards#usings I think we agreed from the recent conversation

src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
```diff
@@ -84,6 +84,12 @@ internal void OnMessageLogged(ILogMessage msg)
         public void Loaded(ViewLoadedParams viewLoadedParams)
         {
             DependencyView = new WorkspaceDependencyView(this, viewLoadedParams);
+            // when a package is loaded update the DependencyView 
+            // as we may have installed a missing package.
+            pmExtension.PackageLoader.PackgeLoaded += (package) =>
+            {
+                DependencyView.DependencyRegen(viewLoadedParams.CurrentWorkspaceModel as WorkspaceModel);
```
QilongTang: I guess this way, if a package is downloaded using the package manager interface, the view extension here will pick up that change as well? Does that cover all the cases in Dynamo?

src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
```diff
@@ -57,16 +57,16 @@ public void Dispose()
         {
         }
 
-       
+
+        [Obsolete("This method is not implemented and will be removed.")]
         public void Ready(ReadyParams readyParams)
```
QilongTang: Good catch

QilongTang:
Looks good with one question.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
mjkkirschner: @QilongTang it doesn't cover other dependency types - I'm also not sure if it covers packages being uninstalled - do you want me to add this same handler for package removal? (if that event exists?)

mjkkirschner: it does exist ` pmExtension.PackageLoader.PackageRemoved`

mjkkirschner: a bit hesitant to get into packageRemoval now and how it relates to package's being in a pending uninstall state.
Would like to defer adding this handler until later or atleast splitting this into two PRs.

mjkkirschner:
@QilongTang flipped the usings - and one thought on the question about dynamo package cases - 
Let me know what you think or if we can merge this.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
QilongTang: Let's leave the uninstall case for now, I would like to give it some testing first

QilongTang:
@mjkkirschner I think it is good for now. :shipit:

PR #9937, opened 2019-08-26
Revert "Revert ASM Loading changes"

mjkkirschner:
Reverts DynamoDS/Dynamo#9932

@aparajit-pratap - now that we have a good build, but know this does have some associated slowdown at startup time- I'd like to get this back into the same state as master before we discuss what nexts steps ((if any)) should be to address the slowdown - 

can we merge this in to get 2.4 and master back in sync?


aparajit-pratap:
Do we want to address the slowdown in 2.4? If not, are we okay with this change going in and we accept some slowdown?

PR #9944, opened 2019-08-27
View Extension API for ViewModel Delegate Commands

scottmitchell:
### Purpose

This PR adds a new class `ViewModelCommandExecutive` and surfaces this via `ViewLoadedParams`. The new command executive gives view extensions controlled access to commands on the `DynamoViewModel` and the current `WorkspaceViewModel`.

### Declarations

### Reviewers

### FYIs


src/DynamoCoreWpf/Extensions/DynamoViewModelCommandExecutive.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// </summary>
+    public class DynamoViewModelCommandExecutive
+    {
+        private DynamoViewModel dynamoViewModel;
+
+        /// <summary>
+        /// Create a Command Executive for a DynamoViewModel
+        /// </summary>
+        /// <param name="model"></param>
+        internal DynamoViewModelCommandExecutive(DynamoViewModel model)
```
mjkkirschner: parameter name should be viewModel

src/DynamoCoreWpf/Extensions/ViewModelCommandExecutive.cs
```diff
@@ -0,0 +1,52 @@
...
+            dynamoViewModel.FitViewCommand.Execute(null);
+        }
+
+        /// <summary>
+        /// Search for an element by its ID and focus the view on it
+        /// </summary>
+        /// <param name="objectID"></param>
+        public void FindByIdCommand(string objectID)
+        {
+            dynamoViewModel.CurrentSpaceViewModel.FindByIdCommand.Execute(objectID);
```
scottmitchell: @mjkkirschner @QilongTang this is what we want instead of FitView. It focuses the view on the element, without changing the zoom. So we won't have the issue of causing users to lose a sense of where they are. 

Only problem is, it appears to be broken... 

It eventually calls this: [https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/Views/Core/WorkspaceView.xaml.cs#L816-L848](https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/Views/Core/WorkspaceView.xaml.cs#L816-L848)

As far as I can tell, this command isn't called anywhere in Dynamo. Which may explain its "only kind of working" state. It's kind of hard to explain what I mean by "only kind of working", but here it goes: currently, this command will focus the view on the element. Great. However, once the user goes to perform any other change to the view (pan, zoom in, zoom out, findByIdCommand, etc.), the view "forgets" it has been refocused on the element, and jumps back to its state before [CenterViewOnElement](https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/Views/Core/WorkspaceView.xaml.cs#L816-L848) was called. The behavior is even stranger when you use the findByIdCommand several times in a row. I'm not exactly sure what this problem is. Continuing to investigate. Considering making a fix to this problem as part of this PR.

scottmitchell: Fixed: https://github.com/DynamoDS/Dynamo/pull/9944/commits/26077520ff81b84f9ad02051a82d9f290554a2ea

It was updating the view but not the model.

scottmitchell:
@mjkkirschner So far, I've only added FitView, FindById, and ForceRunExpression commands to the new command executive. I only added what was needed in the near term for TuneUp (actually, FitView is no longer needed but I've left it in for now). It sounds like we need to have a bigger discussion about what commands we want to add here. Assuming we want to add other commands, we can either add those in a later PR, or wait to merge this PR until we've added them here. I would prefer to add them to a later PR, so that TuneUp can move forward now. What do you think?

mjkkirschner:
@scottmitchell - makes sense to me, and thank for fixing this bug, I think it has caused jumpy behavior for a long time.

Are you able to kick off a build of master and do a push to Nuget? That way you. can consume the latest beta nugets that include your new apis. - I can do this if you do not have access.

It would also be good to run the self serve to see if any tests need to be updated for this change.

scottmitchell:
@mjkkirschner I'm able to log into Jenkins and view history, but it looks like I no longer have permission to build, either self-serve or master. Is it possible to regain permissions? If not, do you mind starting the self serve for me? Will the nuget push happen automatically when the PR is merged?

mjkkirschner:
@scottmitchell no, nuget push is manual. 
I can kick off self serve for this branch, but we'll need to do another build with nuget upload after this is merged.

scottmitchell:
@mjkkirschner Self serve looks good

mjkkirschner:
LGTM

PR #9949, opened 2019-08-29
[PTAL] add a commandLine flag for specifying geometryLib path

mjkkirschner:
add flag for passing ASM path to command line args
add overload to use flag when present to sandbox, cli, cliwpf

### Purpose

Attempt to allow integrators to use DynamoSandbox without requiring potentially slow recursive file search on the users system.

If this PR is merged the consequences are:
1. ~~sandbox daily builds will not load on applications which do not have ASM in their root directory - this is the same behavior as 2.3 **(even if they are added to the list to search for)**~~
2. These applications ~~will need to use~~ can use a command line flag - `-gp` to pass the geometryPath to sandbox to have it skip searching and load that directly.

#### a thought:
~~to enable the most flexible solution - we can keep the recursive loading but move our tests and other integrators to use the command line flag (or preload option) to avoid searching - for example - we can modify our tests to stop geometry library searching and instead use this same mechanism to load the geometry library from a known a specific location.~~

This will enable us to keep the slow searching behavior for daily build users but skip it for integrators and tests... 

This PR now simply adds the commandLineFlag and keeps the behavior @aparajit-pratap added. So both can used for most flexibility depending on user type.

Screen Shot 2019-08-29 at 9 34 22 PM

### Declarations

### Reviewers

@QilongTang 
@aparajit-pratap 
### FYIs

@ColinDayOrg

### DIFF
--- a/src/DynamoApplications/StartupUtils.cs
+++ b/src/DynamoApplications/StartupUtils.cs
@@ -90,18 +90,21 @@ namespace Dynamo.Applications
                 // dll paths we'll import before running a graph 
                 var importPaths = new List<string>() ;
 
+                var asmPath = string.Empty;
+
                 bool showHelp = false;
                 var optionsSet = new OptionSet().Add("o=|O=", "OpenFilePath, Instruct Dynamo to open headless and run a dyn file at this path", o => openfilepath = o)
                 .Add("c=|C=", "CommandFilePath, Instruct Dynamo to open a commandfile and run the commands it contains at this path," +
                 "this option is only supported when run from DynamoSandbox", c => commandFilePath = c)
                 .Add("l=|L=", "Running Dynamo under a different locale setting", l => locale = l)
                 .Add("v=|V=", "Verbose, Instruct Dynamo to output all evalautions it performs to an xml file at this path", v => verbose = v)
-                .Add("x|X", "When used in combination with the 'O' flag, opens a .dyn file from the specified path and converts it to .json." + 
+                .Add("x|X", "When used in combination with the 'O' flag, opens a .dyn file from the specified path and converts it to .json." +
                 "File will have the .json extension and be located in the same directory as the original file.", x => convertFile = x != null)
                 .Add("h|H|help", "Get some help", h => showHelp = h != null)
                 .Add("g=|G=|geometry", "Geometry, Instruct Dynamo to output geometry from all evaluations to a json file at this path", g => geometryFilePath = g)
                 .Add("i=|I=|import", "Import, Instruct Dynamo to import an assembly as a node library. This argument should be a filepath to a single .dll" +
-                " - if you wish to import multiple dlls - use this flag multiple times: -i 'assembly1.dll' -i 'assembly2.dll' ", i=>importPaths.Add(i));
+                " - if you wish to import multiple dlls - use this flag multiple times: -i 'assembly1.dll' -i 'assembly2.dll' ", i => importPaths.Add(i)).
+                Add("gp|geometrypath|GeometryPath", "relative path to a directory containing ASM. If supplied ASM will not be searched for.", gp => asmPath = gp);
 
                 optionsSet.Parse(args);
 
@@ -123,7 +126,8 @@ namespace Dynamo.Applications
                     Verbose = verbose,
                     ConvertFile = convertFile,
                     GeometryFilePath = geometryFilePath,
-                    ImportedPaths = importPaths
+                    ImportedPaths = importPaths,
+                    ASMPath = asmPath,
                 };
             }
 
@@ -140,6 +144,7 @@ namespace Dynamo.Applications
             public bool ConvertFile { get; set; }
             public string GeometryFilePath { get; set; }
             public IEnumerable<String> ImportedPaths { get; set; }
+            public string ASMPath { get; set; }
         }
 
         public static void PreloadShapeManager(ref string geometryFactoryPath, ref string preloaderLocation)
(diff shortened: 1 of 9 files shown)


src/DynamoCLI/Program.cs
```diff
@@ -14,7 +14,15 @@ static internal void Main(string[] args)
             {
                 var cmdLineArgs = StartupUtils.CommandLineArguments.Parse(args);
                 var locale = StartupUtils.SetLocale(cmdLineArgs);
-                var model = StartupUtils.MakeModel(true);
+                DynamoModel model;
+                if (!String.IsNullOrEmpty(cmdLineArgs.ASMPath))
+                {
+                    model = Dynamo.Applications.StartupUtils.MakeModel(true, cmdLineArgs.ASMPath);
```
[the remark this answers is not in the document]
mjkkirschner: yes exactly

src/DynamoApplications/StartupUtils.cs
```diff
@@ -174,6 +179,55 @@ private static IUpdateManager InitializeUpdateManager()
...
+        public static DynamoModel MakeModel(bool CLImode, string geometryLibraryPath)
+        {
+            if (!Directory.Exists(geometryLibraryPath)){
+                throw new ArgumentException(nameof(geometryLibraryPath));
+            }
+            //get sandbox executing location - this is where libG will be located.
+            var exePath = Assembly.GetExecutingAssembly().Location;
+            var rootFolder = Path.GetDirectoryName(exePath);
+
+            Version libGVersion = GetVersionFromASMPath(geometryLibraryPath);
```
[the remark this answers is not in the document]
mjkkirschner: The libG lookup method already does this, I'm using the new one @QilongTang wrote.
https://github.com/DynamoDS/Dynamo/blob/1d186a5c6e35b2ce0d103b152a5ab9140cf7a52c/src/Tools/DynamoShapeManager/Utilities.cs#L251

This should do that right @QilongTang ?

src/DynamoCLI/Program.cs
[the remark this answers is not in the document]
aparajit-pratap: Can we add a todo reminding us to unify the 2 methods in 3.0?

aparajit-pratap:
Minor comment, then LGTM!

mjkkirschner:
> Does this version include the recursive product lookup or are you going to put that back in by (revert*4)ing the other PR?

this PR includes the recursive product lookup, will have to make sure it keeps it when cherry picking.

mjkkirschner:
PTAL - this is ready for another review - tested by starting sandbox from a random folder full of asm binaries and made sure asm worked and was loaded from there. See image above.

test/System/IntegrationTests/DynamoApplicationTests.cs
```diff
@@ -0,0 +1,67 @@
...
+                    new Version(224, 0, 1),
+                    new Version(223, 0, 1),
+                    new Version(225, 0, 0) };
+
+
+            //go get a valid asm path.
+            var locatedPath = string.Empty;
+            var coreDirectory = Path.GetDirectoryName(System.Reflection.Assembly.GetExecutingAssembly().Location);
+
+            DynamoShapeManager.Utilities.GetInstalledAsmVersion2(versions, ref locatedPath, coreDirectory);
```
aparajit-pratap: So for purposes of this test, this could still be doing a recursive lookup of ASM in the product installations.

mjkkirschner: yes, I just wanted the test to pass on any devs machine - BUT a better test would be to then copy the asm dlls somewhere else (like temp directory) and then load them!
what do you think?

mjkkirschner: that could be slow or potentially run into access issues on the test machines though.

mjkkirschner: thinking abit more, I'm inclined to leave this as is - for the purposes of the test like you said, it doesn't really matter that we do the recursive lookup as long as we can assert that the asm path we provided was the asm which was loaded.

src/DynamoApplications/StartupUtils.cs
```diff
@@ -174,6 +179,64 @@ private static IUpdateManager InitializeUpdateManager()
...
+        public static DynamoModel MakeModel(bool CLImode, string asmPath)
+        {
+            if (!Directory.Exists(asmPath)){
+                throw new FileNotFoundException($"{nameof(asmPath)}:{asmPath}");
+            }
+            //get sandbox executing location - this is where libG will be located.
+            var exePath = Assembly.GetExecutingAssembly().Location;
+            var rootFolder = Path.GetDirectoryName(exePath);
+
+            Version libGVersion = GetVersionFromASMPath(asmPath);
```
QilongTang: If such function could throw exception, do we want to try-catch here?

src/DynamoApplications/StartupUtils.cs
```diff
@@ -174,6 +179,64 @@ private static IUpdateManager InitializeUpdateManager()
...
+        internal static Version GetVersionFromASMPath(string asmPath, string searchPattern = "ASMAHL*.dll")
+        {
+            var ASMFilePath = Directory.GetFiles(asmPath, searchPattern, SearchOption.TopDirectoryOnly).FirstOrDefault();
+            if(ASMFilePath != null && File.Exists(ASMFilePath))
+            {
+                var asmVersion = FileVersionInfo.GetVersionInfo(ASMFilePath);
+                var libGversion = new Version(asmVersion.FileMajorPart, asmVersion.FileMinorPart, asmVersion.FileBuildPart);
+                return libGversion;
+            }
+            throw new FileNotFoundException("$Could not find geometry library binaries at : {geometryLibraryPath}");
```
QilongTang: Remind me, seems at this stage, we do not have a DynamoModel yet, nor logger. Is there a better way to catch this for user and us?

mjkkirschner:
all tests pass on self serve.

src/DynamoApplications/StartupUtils.cs
```diff
@@ -174,21 +179,85 @@ private static IUpdateManager InitializeUpdateManager()
...
+            try
+            {
+                if (!Directory.Exists(asmPath))
+                {
+                    throw new FileNotFoundException($"{nameof(asmPath)}:{asmPath}");
+                }
+                Version libGVersion = GetVersionFromASMPath(asmPath);
+
+                //get version of libG that matches the asm version that was supplied from geometryLibraryPath.
+                preloaderLocation = DynamoShapeManager.Utilities.GetLibGPreloaderLocation(libGVersion, rootFolder);
```
QilongTang: If you read the function signature of `GetLibGPreloaderLocation`, the input should be the target ASM version. I dont think we need to do the extra step here to map a libG version and pass in?

src/DynamoApplications/StartupUtils.cs
```diff
@@ -174,21 +179,85 @@ private static IUpdateManager InitializeUpdateManager()
...
+        /// <returns></returns>
+        /// <param name="searchPattern">optional - to be used for testing</param>
+        /// <returns></returns>
+        internal static Version GetVersionFromASMPath(string asmPath, string searchPattern = "ASMAHL*.dll")
+        {
+            var ASMFilePath = Directory.GetFiles(asmPath, searchPattern, SearchOption.TopDirectoryOnly).FirstOrDefault();
+            if(ASMFilePath != null && File.Exists(ASMFilePath))
+            {
+                var asmVersion = FileVersionInfo.GetVersionInfo(ASMFilePath);
+                var libGversion = new Version(asmVersion.FileMajorPart, asmVersion.FileMinorPart, asmVersion.FileBuildPart);
```
QilongTang: I mean here, see comments below

src/DynamoApplications/StartupUtils.cs
```diff
@@ -174,21 +179,85 @@ private static IUpdateManager InitializeUpdateManager()
...
 
+        /// <summary>
+        /// Extracts version of ASM dlls from a path by scanning for ASM dlls in the path.
+        /// Throws if ASM binaries cannot be found in the path.
+        /// </summary>
+        /// <param name="asmPath">path to directory containing asm dlls</param>
+        /// <returns></returns>
+        /// <param name="searchPattern">optional - to be used for testing</param>
+        /// <returns></returns>
+        internal static Version GetVersionFromASMPath(string asmPath, string searchPattern = "ASMAHL*.dll")
```
QilongTang: I prefer this function moved to `ShapeManager` if you guys prefer this function to be ASM binaries specific.. Otherwise, like what you have done in the test, this can be a generic get file version function which we use in some other places.

QilongTang:
A few comments then LGTM

src/DynamoApplications/StartupUtils.cs
mjkkirschner: we're not doing any mapping, except for getting rid of the last part of the version (revision) and returning a Version object. I can call the variable ASM version if you think that is more clear. Sounds like the right name.

src/DynamoApplications/StartupUtils.cs
mjkkirschner: yeah, let me try to move it there. I just want to keep it internal.

src/DynamoApplications/StartupUtils.cs
QilongTang: Can you remind me the reason of removing revision?

mjkkirschner: libG folders are named with major, minor, build only - revision is like the private part of a version number.

src/DynamoApplications/StartupUtils.cs
```diff
@@ -207,33 +207,34 @@ internal static Version GetVersionFromASMPath(string asmPath, string searchPatte
         /// <returns></returns>
         public static DynamoModel MakeModel(bool CLImode, string asmPath)
         {
-            if (!Directory.Exists(asmPath)){
-                throw new FileNotFoundException($"{nameof(asmPath)}:{asmPath}");
```
aparajit-pratap: Was this exception being caught somewhere before you enclosed it in a `try-catch`?

mjkkirschner: no, the application would crash.

QilongTang:
LGTM

PR #9951, opened 2019-08-29
[Cherry-pick] DYN-2101 Installing ampersand package crashes Dynamo while the graph is open

reddyashish:
### Purpose

This PR is to cherry-pick https://github.com/DynamoDS/Dynamo/pull/9957 into RC2.4.0_master branch. 

### Declarations

### Reviewers


src/DynamoPackages/PackageLoader.cs
```diff
@@ -278,11 +279,27 @@ public void LoadAll(LoadPackageParams loadPackageParams)
...
+            // Disabling the run here since new packages are being loaded. 
+            EngineController.DisableRun = true;
+
             foreach (var pkg in enumerable)
             {
                 TryLoadPackageIntoLibrary(pkg);
             }
 
+            // Setting back the DisableRun property back to false, as the package loading is completed.
+            EngineController.DisableRun = false;
```
mjkkirschner: is there ever a case where we could accidentally turn runs off and then never turn them back on?

src/DynamoPackages/PackageLoader.cs
```diff
@@ -278,11 +279,27 @@ public void LoadAll(LoadPackageParams loadPackageParams)
         public void LoadPackages(IEnumerable<Package> packages)
         {
             var enumerable = packages.ToList();
+
+            // This fix is in reference to the crash reported in task: https://jira.autodesk.com/browse/DYN-2101
+            // TODO: we will be re-evaluating this workflow, to find the best clean solution.
```
mjkkirschner: can you describe the refactor problem and file it as a jira task linking the task number here?

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -343,7 +343,11 @@ internal override void RequestRun()
                 // Make this RunSettings.RunEnabled private, introduce the new flag and remove the "executingTask" variable. 
                 if (RunSettings.RunEnabled || executingTask)
                 {
-                    Run();
+                    // Skip's the execution, if packing loading is not completed.
```
mjkkirschner: `skip the execution if runs have been disabled - currently this flag is only set by the Package Loader`

src/DynamoCore/Engine/EngineController.cs
```diff
@@ -38,6 +38,11 @@ public class EngineController : LogSourceBase, IAstNodeContainer, IDisposable
         /// </summary>
         public event AstBuiltEventHandler AstBuilt;
 
+        /// <summary>
+        /// This flag to to check if any new packages are being loaded and disable any execution run's that are triggered before the package loading is completed.
```
mjkkirschner: `This flag is used to check if any packages are currently being loaded, and to disable any executions that are triggered before the package loading is completed. See DYN-2101 for more info`

mjkkirschner:
@reddyashish can you write a simple test that asserts the flag is updated to the correct state while packages load and then set to the correct state afterwards? You can likely just use the packageLoader tests for this.

src/DynamoPackages/PackageLoader.cs
reddyashish: Added a TODO task in JIRA: https://jira.autodesk.com/browse/DYN-2120

mjkkirschner: can you add this DYN task number in the comment?

mjkkirschner:
@reddyashish - does this change make the loading of packages faster? Since we are no longer executing the graph a number of times related to the number of custom functions present in the graph?

reddyashish:
Added the test. Please take a look

reddyashish:
@mjkkirschner It would definitely reduce the number of executions when the graph is in automatic mode. If the graph is in manual mode, there would be no difference as the run's are not triggered. so I would say yes. Not sure if we can validate it using the performance benchmark.

src/DynamoPackages/PackageLoader.cs
reddyashish: I don't think so as this would be set back to false, once the package loading is completed.

aparajit-pratap: I think @mjkkirschner was alluding to a case where an exception is thrown from `TryLoadPackageIntoLibrary` and this flag-reset step is skipped in that case. I think `TryLoadPackageIntoLibrary` has a `try-catch` inside it that can catch any exception and not rethrow it in order for this flag reset to take place right after but to be safe we can have another `try-catch-finally` around the `for` loop and reset the flag in the `finally` block to ensure it always happens no matter what.

mjkkirschner: @aparajit-pratap thats exactly what I was thinking - do you think it's worth doing this:
or should it be okay since TryLoadPackageIntoLibrary will already catch and move onto the next package.

```suggestion

try {
      foreach (var pkg in enumerable)
            {
                TryLoadPackageIntoLibrary(pkg);
            }
      }
finally{
            EngineController.DisableRun = false;
           }
```

test/Libraries/PackageManagerTests/PackageLoaderTests.cs
```diff
@@ -472,6 +473,35 @@ public void GetOwnerPackageReturnsNullForInvalidFunction()
             Assert.IsNull(foundPkg);
         }
 
+        /// This test is added for this task: https://jira.autodesk.com/browse/DYN-2101. 
+        /// A followup task is added https://jira.autodesk.com/browse/DYN-2120 to refactor the approach to this solution.
+        /// This test needs to be modified in that case. 
+        [Test]
+        public void PackageLoadExceptionTest()
```
mjkkirschner: 👍

mjkkirschner:
I think as it is if we cannot find a better solution this looks good as a patch fix for the regression - @aparajit-pratap have you had a chance to consider alternatives, I am also looking but nothing lower risk has turned up yet.

src/DynamoPackages/PackageLoader.cs
aparajit-pratap: It would at least be worthwhile checking that the exception caught inside `TryLoadPackageIntoLibrary` is not re-thrown from the `Log` methods?

mjkkirschner: log doesn't seem to rethrow anything.

test/Libraries/PackageManagerTests/PackageLoaderTests.cs
mjkkirschner: can you add category as well:
`TechDebt`
https://github.com/DynamoDS/Dynamo/blob/8d4757760aa108a2aa446d42e35afd3971678369/test/DynamoCoreWpfTests/PublishPackageViewModelTests.cs#L157

reddyashish:
Merging this

PR #9953, opened 2019-08-30
Marshalling optimization

saintentropy:
Purpose is to discuss these speculative changes


saintentropy:
Not ready for primetime yet

PR #4235, opened 2015-04-14
Fix crash when loading .dyn file with custom node in run auto mode

ke-yu:
This PR fixes defect [MAGN 7073 Any dy file with any custom node crashes in run auto mode](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-7073).

This is because of inconsistent state `EngineController` of `HomeWorkspaceModel` and `DynamoModel`. When a .dyn file is loaded, 
  1. A `HomeWorkspaceModel` will be created, and its `EngineController` will be set to `DynamoModel`'s `EngineController`.
  2. `DynamoModel.RestEngine()` is called, `EngineController` is disposed and a new `EngineController` will be created. At this moment, the `EngineController` in `HomeWorkspaceModel` is invalid.
  3. `DynamoModel` will register all custom node definitions, which consequently will mark all custom node instances as dirty and trigger run for `HomeWorkspaceModel`. As `EngineController` in `HomeWorkspaceModel` is invalid, Dynamo crashes.
  4. `DynamoModel` sets `HomeWorkspaceModel`'s `EngineController` to the correct one. But it is too late.

Ideally `HomeWorkspaceModel` should use `DynamoModel`'s `EngineController` instead of keeping its own copy, but @Benglin  points out that `HomeWorkspaceModel` shouldn't keep a reference to `DynamoModel`. 

But it is not necessary to mark all custom node instances as dirty in resetting `EngineController`. The reason is, `EngineController` will be reset if a new .dyn file is loaded, in which case .dyn file will be executed eventually, or a new library is loaded, in which case it shouldn't have any change on the existing custom node instances, so here `RegisterCustomNodeDefinitionWithEngine()` only does registration work and a new function `UpdateCustomNodeDefinition()` is created for registration and updating instances, the latter is used by `CustomNodeWorkspaceModel`.

@Benglin, @ikeough  please help to review the change.

PR #9954, opened 2019-08-30
[DNM] perform library reload after superfluous package load/custom node executions.

mjkkirschner:
### Purpose

@aparajit-pratap @reddyashish -

update:
current commit uses a pre-existing lock in the engine, to block reloading the library while the engine is inside a few important functions. One of these functions is `generateGraphSyncDataForCustomNode` https://github.com/DynamoDS/Dynamo/blob/dfb938bba667e9a4ddf54df1c983f3d4a48fdc1f/src/DynamoCore/Engine/EngineController.cs#L327

which is triggered when the customNode definition is updated. - this seems to stop the crash at least. But diagramming the execution flow - doesn't really explain why it works -  as I thought the graphSync generation would occur on the same thread as the library reload. (main thread)

`getMirror` also uses this lock
https://github.com/DynamoDS/Dynamo/blob/dfb938bba667e9a4ddf54df1c983f3d4a48fdc1f/src/DynamoCore/Engine/EngineController.cs#L212
and this might be a more likely explanation, as the graph executes getMirror maybe called from the scheduler thread and a race condition is prevented. - But I have not confirmed this.

In any case the reason this code works is not easy to confirm - and it could be that getting the lock simply changes timing enough to make avoid a race condition in debug mode. 😕 .

 But -  then we need to resync the custom node definitions with the engine, though I still don't understand exactly why this is case. My assumption is that reloading the library has dumped the loaded functions and they need to be re-registered - and usually this occurs after the library load as part of the node being loaded or placed. It's still not clear though or what consequences these changes have but they might be a potential refactor for later.

#### TLDR:

I'm not confident in this change at this time.

I found these two very interesting PRs:
https://github.com/DynamoDS/Dynamo/pull/2745
https://github.com/DynamoDS/Dynamo/pull/4235

### Declarations

### Reviewers

### FYIs

### DIFF
--- a/src/DynamoCore/Engine/EngineController.cs
+++ b/src/DynamoCore/Engine/EngineController.cs
@@ -38,6 +38,8 @@ namespace Dynamo.Engine
         /// </summary>
         public event AstBuiltEventHandler AstBuilt;
 
+        public event Action dumpedLibrary;
+
         /// <summary>
         /// This event is fired when <see cref="UpdateGraphAsyncTask"/> is completed.
         /// </summary>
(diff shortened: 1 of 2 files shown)


aparajit-pratap:
@mjkkirschner this will still cause the library loaded event handler to be scheduled after the graph execution. We need to make it happen before.

mjkkirschner:
@aparajit-pratap - heres another solution.

this one works - but still the library is reloaded after the graph run - but they occur at different times because of the lock AND then all the custom nodes are resynced so they are available again in the VM.

So ideally yes we would reload the library before any node executions - this solution results in a bunch of wasted runs, lock contention (we could probably use the scheduler instead), and resyncing all the custom nodes twice....

I feel that this is way more risky to change than the small modification @reddyashish used in his pr to block the runs which cause the odd state in the first place.

PR #9957, opened 2019-09-03
DYN-2101 Installing ampersand package crashes Dynamo while the graph is open

reddyashish:
### Purpose
This PR is to fix the crash exception reported in the task: https://jira.autodesk.com/browse/DYN-2101

In short, when a new package is being loaded into Dynamo(while the graph is open in automatic mode), the VM is reinitialized and reloads all the libraries(including the core one's) into the VM. So when the node's are resolved, execution run's are triggered asynchronously. This causes a thread timing issue as the post execution run steps are called after the VM is reset and the DSExecutable object is lost during the execution run. To avoid this case, we have added a flag to check if the packages are being loaded and skip those initial run's as a final run is anyways triggered after the reloading of libraries is completed.

In order to make sure, we do not have any new regressions, we plan to merge this to master first before merging into the RC_2.4 branch. 

For testing: we have a UI automation test that @mjkkirschner added https://github.com/DynamoDS/DynamoUIAutomation/pull/155

TODO:
We plan to re-evaluate this whole workflow in the future. 
Task: https://jira.autodesk.com/browse/DYN-2120

### Declarations

### Reviewers


mjkkirschner:
LGTM

PR #9972, opened 2019-09-11
[WIP] Add better unresolved node error messages

mjkkirschner:
### Purpose

The error message when creating a dummy node does not make sense:
see this thread:
https://forum.dynamobim.com/t/standard-nodes-are-unresolved/40307/10

these are not coreNodes and yet they imply that the types are defined in dynamoCore - I think this is because the baseType is always a NodeModel or DSFunction which is defined in DynamoCore.dll

instead use the concrete type assembly, which we have, to set a more logical error message. Alternatively we could use the entire type name if that is more useful...

### Declarations

### Reviewers

### FYIs


QilongTang:
Oops, touched the wrong PR

PR #9976, opened 2019-09-13
NodeProfilingData start time should update when node is re-executed

scottmitchell:
### Purpose

Previously, the `startTime` for a `nodeProfilingData` could never be updated. When a node was re-executed, its `endTime` would be updated, but not its `startTime`. 

This caused the `executionTime` property of a `nodeProfilingData` to return the timespan from the **start of its first execution** to the **end of its most recent execution**. 

It should instead return the timespan from the **start of its most recent execution** to the **end of its most recent execution**.

An existing workaround is to re-enable profiling on all nodes between profiling graph runs. However, this forces all nodes to re-execute, making it difficult to collect profiling data on partial graph runs (delta compute). This PR makes it easy to gather profiling data on partial graph runs, and also makes it unnecessary for extension developers to re-enable profiling between runs (unless they want to re-execute all nodes). 

### Declarations

### Reviewers

@ColinDayOrg @mjkkirschner @QilongTang 

### FYI

@alvpickmans

### DIFF
--- a/src/DynamoCore/Engine/Profiling/NodeProfilingData.cs
+++ b/src/DynamoCore/Engine/Profiling/NodeProfilingData.cs
@@ -11,8 +11,6 @@ namespace Dynamo.Engine.Profiling
         private DateTime? startTime = null;
         private DateTime? endTime = null;
 
-        internal bool profilingStarted;
-
         internal NodeProfilingData(NodeModel node)
         {
             this.node = node;
@@ -32,10 +30,9 @@ namespace Dynamo.Engine.Profiling
 
         private void RecordEvaluationState(object data)
         {
-            if (!profilingStarted)
+            if (!startTime.HasValue)
             {
                 startTime = DateTime.Now;
-                profilingStarted = true;
                 node.OnNodeExecutionBegin();
                 return;
             }
--- a/src/DynamoCore/Engine/Profiling/NodeProfilingData.cs
+++ b/src/DynamoCore/Engine/Profiling/NodeProfilingData.cs
@@ -11,6 +11,8 @@ namespace Dynamo.Engine.Profiling
         private DateTime? startTime = null;
         private DateTime? endTime = null;
 
+        private bool profilingInProcess;
+
         internal NodeProfilingData(NodeModel node)
         {
             this.node = node;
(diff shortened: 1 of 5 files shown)


src/DynamoCore/Engine/Profiling/NodeProfilingData.cs
```diff
@@ -11,6 +11,8 @@ class NodeProfilingData : IDisposable
         private DateTime? startTime = null;
```
alvpickmans: being `startTime` and `endTime` nullable properties, wouldn't it be easier to use them to check if profiling has or not started? Still you'll need to reset them values to `null` before each run, but cleaner that having to deal with an extra property.

scottmitchell: @alvpickmans yeah that would work. One effect of this approach is that the `ExecutionTimeData` would only contain results for nodes executed during the most recent graph run. All other results would be null, so it would be up to the developer to cache previous results. I think I actually like this because it makes it obvious which results are "fresh".

src/DynamoCore/Engine/Profiling/ProfilingSession.cs
```diff
@@ -46,6 +46,7 @@ public IProfilingExecutionTimeData ProfilingData
         private void OnGraphPreExecution(Session.IExecutionSession session)
         {
             profilingData.StartTime = DateTime.Now;
```
mjkkirschner: should these two calls be reversed?

test/DynamoCoreTests/ProfilingTest.cs
```diff
@@ -149,5 +150,68 @@ public void TestNodeExecutionEvents()
...
+
+            // Second run / Re-execute
+            BeginRun();
+
+            // Gather second run data
+            var secondRunTimes = new List<TimeSpan>();
+            foreach (var node in nodes)
+            {
+                var executionTime = (TimeSpan)profilingData.NodeExecutionTime(node);
+                Assert.IsNotNull(executionTime);
```
mjkkirschner: I think this behavior should be documented in the public interface for the API.

src/DynamoCore/Engine/Profiling/ProfilingSession.cs
scottmitchell: Yep, makes sense

test/DynamoCoreTests/ProfilingTest.cs
scottmitchell: https://github.com/DynamoDS/Dynamo/pull/9976/commits/cd7d075450a569ed0ddc76cb21fa96e539f0a784

QilongTang:
LGTM

mjkkirschner:
@ColinDayOrg when you get a chance curious for your thoughts on this change. I feel it's a move in the right direction.

ColinDayOrg:
@mjkkirschner I totally agree that this is a move in the right direction and a good thing to do.

PR #9985, opened 2019-09-18
DYN-1454 Unresolved nodes should attempt to be resolved when a new package is loaded

reddyashish:
### Purpose

This PR is to solve the unresolved nodes issue when a new package is loaded into the Dynamo. The nodes were not being resolved until the workspace was reopened. 

JIRA: https://jira.autodesk.com/browse/DYN-1454

After this fix, when the package is downloaded, the workspace is reloaded and the dependency table is updated in the extensions side bar. But I am getting a compiler internal exception similar to the one experienced in Ampersand package crash issue. The "DSExecutable" object here is "null" for one of the "UpdateGraphAsyn" tasks that is triggered. 

https://github.com/DynamoDS/Dynamo/blob/master/src/Engine/ProtoCore/RuntimeCore.cs#L337

exception

Still work in progress on fixing the exception. 

### Declarations

### Reviewers

### DIFF
--- a/tools/Performance/DynamoPerformanceTests/graphs/GeometryDisposeLarge.dyn
+++ b/tools/Performance/DynamoPerformanceTests/graphs/GeometryDisposeLarge.dyn
@@ -1,0 +1,122 @@
+{
+  "Uuid": "07da93cd-43ea-48d2-8227-103c316374e3",
+  "IsCustomNode": false,
+  "Description": null,
+  "Name": "GeometryDisposeLarge",
+  "ElementResolver": {
+    "ResolutionMap": {}
+  },
+  "Inputs": [],
+  "Outputs": [],
+  "Nodes": [
+    {
+      "ConcreteType": "PythonNodeModels.PythonNode, PythonNodeModels",
+      "NodeType": "PythonScriptNode",
+      "Code": "# Load the Python Standard and DesignScript Libraries\r\nimport sys\r\nimport clr\r\nclr.AddReference('ProtoGeometry')\r\nfrom Autodesk.DesignScript.Geometry import *\r\n\r\n# The inputs to this node will be stored as a list in the IN variables.\r\nsize = IN[0]\r\nout = []\r\n# Place your code below this line\r\n\r\nfor i in range(size):\r\n\tCuboid.ByLengths(1,1,1);\r\n\t\r\nfor i in range(size/10):\r\n\tout.append(Cuboid.ByLengths(1,1,1));\r\n\t\r\n# Assign your output to the OUT variable.\r\nOUT = out",
+      "VariableInputPorts": true,
+      "Id": "345d0f3cf5f94b0cbf4504b3f5f83590",
+      "Inputs": [
+        {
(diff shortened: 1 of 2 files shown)


src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
```diff
@@ -20,7 +19,7 @@ namespace Dynamo.Wpf.Extensions
     public class ViewLoadedParams : ReadyParams
     {
         private readonly DynamoView dynamoView;
-        private readonly DynamoViewModel dynamoViewModel;
+        public readonly DynamoViewModel dynamoViewModel;
```
mjkkirschner: definitely don't want to do this for real. We do not want to expose the viewModel to all api users I don't think.

reddyashish: Yes for sure, I will either change it to internal or use some other way to access the model object

aparajit-pratap: ☠️

QilongTang:
Some known side effect of this change by discussing with various member on the team:
1. Unsaved changes will be lost. But that can be fixed by serializing the current homeworkspace and read from temp folder.
2.  Performance concerns on giant graph. No obvious work around here.

@reddyashish can you also share the solution you tried on deserializing the node individually? And the error from that solution?

src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
```diff
@@ -89,7 +91,13 @@ public void Loaded(ViewLoadedParams viewLoadedParams)
...
             pmExtension.PackageLoader.PackgeLoaded += (package) =>
             {
-                DependencyView.DependencyRegen(viewLoadedParams.CurrentWorkspaceModel as WorkspaceModel);
+                DynamoModel model = viewLoadedParams.dynamoViewModel.Model;
+                WorkspaceModel workspace = model.CurrentWorkspace;
+                String filePath = Path.GetFullPath(workspace.FileName);
+                model.ClearCurrentWorkspace();
+                model.OpenFileFromPath(filePath, false);
+
+              //  DependencyView.DependencyRegen(viewLoadedParams.CurrentWorkspaceModel as WorkspaceModel);
```
aparajit-pratap: So I checked; if this is triggered by the `PackageLoaded` event this code will indeed execute before all the libraries are reloaded and the VM is reset, which is not what we want.

aparajit-pratap: I imagine a step being added after this line here: https://github.com/DynamoDS/Dynamo/blob/499bd9b359fdd52bd977b2e7b5ca1113ca63e80a/src/DynamoCore/Engine/EngineController.cs#L505 )when all libraries have been reloaded after importing a new one) to deserialize/resolve the dummy nodes again. It could be a new event or something else.

PR #9993, opened 2019-09-20
Reduce Dynamo ViewModel memory leak Part I

QilongTang:
Please Note:

### Purpose

This PR is an example of things need to be done to dispose object in correct order to avoid memory leak.

- [x] PortViewModels are disposed correctly now
- [x] RunSettingsViewModel is disposed correctly now
- [x] InfoBubbleViewModels are disposed correctly now
- [x] HomeworkspaceViewModel are disposed correctly but still retained because of workspaceView leak on resource dictionary
- [x] WPF binding leak addressed. Notifications collection on notificationExtension are disposed correctly now
- [x] NodeCategoryViewModels are now disposed correctly
- [x] NodeSearchElementViewModels are now disposed correctly
- [x] DynamoViewModel no longer keeps a valid duplicate SearchViewModel now

The following is a comparison using our sample Geometry - Solids as file open candidate.

**Baseline -> 2.4.0
Memory Consumption Difference between Dynamo StartUp-> File Open Once -> File Open 10x**
image

**File Open Once**
image

**File Open 10x**
image

**After -> Release build of this branch (2.5.0)
Memory Consumption Difference between Dynamo StartUp-> File Open Once -> File Open 10x**
image

**File Open Once**
image

**File Open 10x**
image

### Declarations

### Reviewers

@DynamoDS/dynamo


src/DynamoCoreWpf/ViewModels/RunSettingsViewModel.cs
```diff
@@ -236,6 +236,14 @@ public RunSettingsViewModel(RunSettings settings, HomeWorkspaceViewModel workspa
             ToggleRunTypeEnabled(RunType.Periodic, false);
         }
 
+        /// <summary>
+        /// When switching workspace, this need to be called in HomeworkspaceViewModel dispose function
+        /// </summary>
+        internal void Dispose()
+        {
+            this.workspaceViewModel = null;
```
mjkkirschner: I think the cases where this needs to be done in c# are small in number, why do we need to do it here?

QilongTang: I believe something here is preventing HomeworkspaceViewModel to be disposed, not sure if it is related to the readonly field or simply calling dispose, will give it another try tomorrow

src/DynamoCoreWpf/ViewModels/Core/HomeWorkspaceViewModel.cs
```diff
@@ -230,6 +233,7 @@ public override void Dispose()
             hwm.EvaluationCompleted -= hwm_EvaluationCompleted;
             hwm.SetNodeDeltaState -= hwm_SetNodeDeltaState;
             RunSettingsViewModel.PropertyChanged -= RunSettingsViewModel_PropertyChanged;
+            RunSettingsViewModel.Dispose();
```
aparajit-pratap: I don't see `RunSettingsViewModel` implementing an `IDisposable` nor a `Dispose` method defined on it. Am I missing something?

src/DynamoCoreWpf/ViewModels/RunSettingsViewModel.cs
aparajit-pratap: From the retention paths you captured screenshots of it looks like the `DynamoView` is indirectly holding onto an instance of `RunSettingsViewModel` that is in turn holding onto the `HomeWorkspaceViewModel`. I think we should try and dispose off the `HomeWorkspaceViewModel` from `DynamoView` in that case, no?

QilongTang: I believe DynamoView is a singleton as well as the RunSettingsControl. So we won't be able to know when to dispose specifically from the view, but we can do that when closing the currentWorkspace. WorkspaceRemoved event will be triggered and then Dispose() on HomeWorkspaceModel will be called, that's why I added the logic there.

QilongTang:
Running Self CI, I think this PR is good for review. It does not address all the issues, but is a good step.

src/DynamoCoreWpf/ViewModels/RunSettingsViewModel.cs
```diff
@@ -81,12 +81,12 @@ public RunTypeItem(RunType runType)
     /// notifications as those notifications are raised when the value is set on the
     /// model.
     /// </summary>
-    public class RunSettingsViewModel : NotificationObject
+    public class RunSettingsViewModel : ViewModelBase
```
mjkkirschner: this is an API break.

mjkkirschner: or - could be, we need to be very careful changing the base class of a public class.

src/DynamoCoreWpf/ViewModels/Search/BrowserItemViewModel.cs
```diff
@@ -89,7 +89,7 @@ public interface ISearchEntryViewModel : INotifyPropertyChanged, IDisposable
         ElementTypes ElementType { get; }
     }
 
-    public class NodeCategoryViewModel : NotificationObject, ISearchEntryViewModel
+    public class NodeCategoryViewModel : ViewModelBase, ISearchEntryViewModel
```
mjkkirschner: same here. can this be accomplished without changing the base class by just adding more interfaces?

src/DynamoCoreWpf/ViewModels/RunSettingsViewModel.cs
QilongTang: How do we find out? ViewModelBase implements `NotificationObject` and `IDisposal` so I believe doing this will not remove any public property, that's actually why I thought it is not API breaking. Same reason below.

src/DynamoCoreWpf/ViewModels/Search/BrowserItemViewModel.cs
mjkkirschner: you can check this doc: 

and do some thought experiments or real experiments with extensions using that class before and after.

QilongTang: `public abstract class ViewModelBase : NotificationObject,IDisposable` I dont think switching out the baseclass here break API backward compatibility because it only adds two APIs
```
IsDebugBuild
```
and
```
Dispose
```

QilongTang:
Updated screenshots and could be more useful for our discussion

src/Notifications/NotificationsViewExtension.cs
```diff
@@ -11,11 +12,23 @@
...
 {
-    public class NotificationsViewExtension : IViewExtension
+    public class NotificationsViewExtension : IViewExtension, INotifyPropertyChanged
     {
         private ViewLoadedParams viewLoadedParams;
         private Action<Logging.NotificationMessage> notificationHandler;
-        public ObservableCollection<Logging.NotificationMessage> Notifications { get; private set; }
+
+        private ObservableCollection<Logging.NotificationMessage> notifications;
+        public ObservableCollection<Logging.NotificationMessage> Notifications
```
mjkkirschner: so this is kind of argg - we need to implement INotifyPropertyChanged to have .net gc this collection?

Can we add a comment saying as much so that in a year we don't delete it all when we see it is unused 😄

QilongTang: Only because the collection is binding to WPF UI, I can add comments

QilongTang:
All Unit tests passing, Ready for review

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
```diff
@@ -251,29 +251,19 @@ public bool IsCursorForced
             set { isCursorForced = value; RaisePropertyChanged("IsCursorForced"); }
         }
 
-        private CompositeCollection _workspaceElements = new CompositeCollection();
         [JsonIgnore]
-        public CompositeCollection WorkspaceElements { get { return _workspaceElements; } }
-        
-        ObservableCollection<ConnectorViewModel> _connectors = new ObservableCollection<ConnectorViewModel>();
+        public CompositeCollection WorkspaceElements { get; } = new CompositeCollection();
```
mjkkirschner: can you talk a bit about these changes?

QilongTang: These are to use auto properties

src/DynamoCoreWpf/Views/Preview/InfoBubbleView.xaml.cs
```diff
@@ -121,6 +121,12 @@ public InfoBubbleView()
         {
             InitializeComponent();
 
+            // Make sure to resubscribe to the event handlers
```
mjkkirschner: do you mean unsubscribe?

mjkkirschner:
LGTM

src/DynamoCoreWpf/Views/Preview/InfoBubbleView.xaml.cs
QilongTang: Yes, let me update

PR #9994, opened 2019-09-20
Fix crash when opening second script in French C3D

angelowang:
Please Note:

### Purpose
Fix crash when opening second script in French C3D

Steps to reproduce:
- Launch Dynamo in French OS and French Civil 3D.
- Open from `Help` menu: `Help>Samples>Core>Core_ListAtLevel.dyn`
- Without close the dyn, open from `Help` menu: `Help>Samples>Core>Core_ListLacing.dyn`.
- Exception dialog comes and C3D will crash.

The exception arises when converting `1.0` to double under `fr-FR` culture,
in `GroupFontSizeToEditorEnabledConverter.Convert()` on `values[0]`.

Questions:
- Where does this `1.0` come from?
  It is from `AnnotationView.xaml`, where `FallbackValue` of the `Visibility`
property is set to `1.0`. When we switch from first dyn to the second, the
current workspace will be removed, triggered WPF binding updates.

- Why it doesn't crash when we do the same operation in DynamoSandbox?
  WPF actually doesn't follow `CurrentCulture` or `CurrentUICulture`
when resolving the binding (https://weblog.west-wind.com/posts/2009/Jun/14/WPF-Bindings-and-CurrentCulture-Formattings.
  So by default, the `culture` passed to the `Convert` method is always `en-US`.
  So it is OK.

- How to reproduce with DynamoSandbox?
  You have to manually add several lines of code to `DynamoSandbox` entry fuction, it will crash
although no exception dialog (interestingly, it pops up IE on my side).
```
CultureInfo.DefaultThreadCurrentCulture = CultureInfo.CurrentCulture = new CultureInfo("fr-FR");
CultureInfo.DefaultThreadCurrentUICulture = CultureInfo.CurrentUICulture = new CultureInfo("fr-FR");
FrameworkElement.LanguageProperty.OverrideMetadata(typeof(FrameworkElement), new FrameworkPropertyMetadata(XmlLanguage.GetLanguage(CultureInfo.CurrentCulture.IetfLanguageTag)));
```

- Why does Civil 3D have `fr-FR` culture?
  People buying Civil 3D will get Map 3D for free. The Map 3D dll, `Autodesk.Map.IM.Loader.dll`,
  will override the Culture to be used by all WPF bindings, with the very similar code as above.
  When I rename it, the crash goes away.

- So how to fix?
  At this stage, the safest fix would be simply changing `1.0` to `1`.
  Later we should evaluate whether we really need to simply use the InvariantCulture
  in all our binding coverters. We have unit tests for those converters, so also
  need to remove/change them if decided to go that way.
  Another way, we might should catch all the exception and `Converter.ToDouble()` and use invariant culture if failed..

### Declarations

### Reviewers

Michael Kirschner
Aaran Tang

### DIFF
--- a/src/DynamoCoreWpf/Views/Core/AnnotationView.xaml
+++ b/src/DynamoCoreWpf/Views/Core/AnnotationView.xaml
@@ -256,7 +256,7 @@
                                 <Condition Value="True">
                                     <Condition.Binding>
                                         <MultiBinding Converter="{StaticResource GroupFontSizeToEditorEnabledConverter }">
-                                            <Binding Path ="DataContext.Zoom" FallbackValue="1.0" RelativeSource="{RelativeSource FindAncestor, AncestorType={x:Type views:WorkspaceView}}"></Binding>
+                                            <Binding Path ="DataContext.Zoom" FallbackValue="1" RelativeSource="{RelativeSource FindAncestor, AncestorType={x:Type views:WorkspaceView}}"></Binding>
                                             <Binding ElementName="GroupTextBlock" Path="FontSize"/>
                                             <Binding ElementName="GroupTextBlock" Path="Visibility"/>
                                         </MultiBinding>


mjkkirschner:
Thanks for the write up @angelowang - I think I understand whats happening now. This will need to end up getting merged to a different branch, and like you mentioned above I think we should add some try/catch logic, and a test for this situation. (we can try to set the localization to FR just for the test, and then set it back after).

We'll make a PR.

PR #9996, opened 2019-09-20
DYN-2093: Dynamo Sandbox VM crashes on opening graph.

ColinDayOrg:
Please Note:

### Purpose

This PR addresses an issue where the Dynamo VM crashes due to attempting to remove cached replication guides and at level items from empty lists that are returned to the DotMethod function.
Specifically, this has to do with a code block that uses List.Flatten with a null list, but I believe that this issue could be affecting users in a manner of other ways.

For the level of testing declaration below the specific issue has a test that fails without these changes and passes with the changes, but I believe that there could/should be more coverage, although I have not been able as of yet to create other test cases that show the same symptoms, although I believe that they exist.

### Declarations

### Reviewers

@mjkkirschner 
@aparajit-pratap


test/Engine/ProtoTest/MultiLangTests/NullInputTest.cs
```diff
@@ -0,0 +1,26 @@
...
+            // TODO: Determine if this is necessary for this test
+            FFITarget.DisposeCounter.Reset(0);
+        }
+
+        [Test]
+        public void DYN_2093_NullInput_Flatten()
+        {
+            var code =
+@"
+import(""DSCoreNodes.dll"");
```
mjkkirschner: ah, I see what you mean now about the import - this is a guess - but perhaps if you don't do the import, the test does not fail because the function is simply not found and that results in an error and not a crash.

ColinDayOrg: Thanks for the reply on the question. That is pretty much what I was thinking was going on, but just wanted a second opinion. The thing about it was if this is what is happening it means that not having the import caused a test to succeed where it should have failed and there was no immediate indication that that was what was going on. It makes me a little worried if there could be other cases like this now or in the future.

ColinDayOrg: Possibly we need to have some method to surface the failing import issue in tests more visibly.

mjkkirschner: well, to write a more robust test, maybe we can assert a value for a function which runs after the list.flatten call - 

somethingLike

```
import(""DSCoreNodes.dll"");
testData = [[[1]]];
DSCore.List.flatten(null);
result = DSCore.List.flatten(testData); // assert that result = [1] 
```

it's kind of unclear I guess to someone reading it later, but it should check that the execution did not crash?

mjkkirschner: @ColinDayOrg - hmm! - I'm not sure if the missing method error will throw an exception - but you could try wrapping the 
`thisTest.RunScriptSource(code);`

call in
```
 Assert.DoesNotThrow( ()=>{ thisTest.RunScriptSource(code);
}
```

ColinDayOrg: As it is now the test fails without the code changes and passes with the changes, so I don't think that there necessarily needs to be additional code for this specific test. I am trying to find other cases that can be tested to be more robust. I just wanted to point out that it appears that there can be test cases like this that could be created that would pass even though they should be failing and there was no indication about what was going on. I can add these changes you were discussing above if you think it would be better.

ColinDayOrg: I will though give your suggestions a try and see if it clarifies the situation.

aparajit-pratap:
@ColinDayOrg I see that the `List.Flatten` method is a static method and cannot be called as such:
```
l = [[1,2,3]];
l.Flatten();
```
It can therefore only be called like this: `List.Flatten(l);` and in this case the `DotMethod` will not be called. As I mentioned before the `DotMethod` is called only for dot methods, meaning instance methods. 
Further I wasn't able to reproduce the crash with a `null` list being passed into `List.Flatten(l)` or using the UI node. I think we might be looking at the wrong thing here.

mjkkirschner:
@aparajit-pratap are you saying that the `dotMethod` method is getting called when it should not be?

aparajit-pratap:
I see what's going on in Colin's reproducible example. A method that doesn't exist is being used: `List.flatten(a)`. The method `flatten` doesn't exist and should've been called using upper case. In this case (of a non-existent method being called), it seems to me that the identifier `List` is being interpreted as a member variable. Colin's fix seems to resolve this issue however I think we should address the underlying issue of not executing this code in the first place - in other words identify the non-existent method at compile time and prevent it from being executed.

ColinDayOrg:
That is sort of what I was thinking as well. The fix seems to work, but it is just working around a deeper issue that I agree should be investigated further. It feels to me that this could be the cause of other potential crashing issues that users could run into, and it would be unclear as to why the issues were occurring.

ColinDayOrg:
Note also though that it seems to me that this "fix" should be safe and address the top-level issue for now. I'm not sure what the general thought is though on this sort of thing; i.e. fix for now with a band-aid and investigate further for the deeper issue, or not fix to force the deeper fix to happen now.

mjkkirschner:
I’m not sure we have a general principle on this - usually depends on timing, complexity, risk I guess - my vote in this case would be to get to the bottom of the issue and fix that. I think this fix can be kept in addition to that - leave it up to you and Aparajit of course though.

test/Engine/ProtoTest/MultiLangTests/NullInputTest.cs
```diff
@@ -0,0 +1,26 @@
...
+namespace ProtoTest.MultiLangTests
+{
+    class NullInputTest : ProtoTestBase
+    {
+        public override void Setup()
+        {
+            base.Setup();
+
+            // TODO: Determine if this is necessary for this test
+            FFITarget.DisposeCounter.Reset(0);
```
aparajit-pratap: I don't think this is needed.

test/Engine/ProtoTest/MultiLangTests/NullInputTest.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+            var code =
+@"
+import(""DSCoreNodes.dll"");
+DSCore.List.flatten(null);
+  ";
+            thisTest.RunScriptSource(code);
+        }
+    }
+}
```
aparajit-pratap: Please also give a non-null input list to the `List.flatten` method and assert that it returns null and does not crash.

PR #10004, opened 2019-09-23
Add tolerance based equals node

aparajit-pratap:
### Purpose

JIRA: DYN-2049
- Add tolerance based `Equals` node

TODO:
~Add tests~
~Move hardcoded strings to string resources~
~Create icon for new node in Library view~

### Declarations

### Reviewers

@DynamoDS/dynamo


src/Libraries/CoreNodeModels/Equals.cs
```diff
@@ -0,0 +1,108 @@
...
+        public Equals()
+        {
+            InPorts.Add(new PortModel(PortType.Input, this, new PortData("lhs", "integer or double value")));
+            InPorts.Add(new PortModel(PortType.Input, this, new PortData("rhs", "integer or double value")));
+            InPorts.Add(new PortModel(PortType.Input, this, new PortData("tolerance", "tolerance = 0.00001")));
+            OutPorts.Add(new PortModel(PortType.Output, this, new PortData("bool", "result of equality check")));
+            RegisterAllPorts();
+        }
+
+        public override IEnumerable<AssociativeNode> BuildOutputAst(List<AssociativeNode> inputAstNodes)
```
mjkkirschner: does this deal with replication /lacing? or do you need to add some special AST for that as well?

src/Libraries/CoreNodeModels/Equals.cs
```diff
@@ -0,0 +1,108 @@
...
+    [NodeName("==")]
+    [NodeDescription("Equals with tolerance input")]
+    [NodeCategory("Core.Math")]
+    [NodeSearchTags("Equals")]
+    [InPortTypes("double ", "double", "double")]
+    [OutPortTypes("bool")]
+    [IsDesignScriptCompatible]
+    public class Equals : NodeModel
+    {
+        private static readonly string functionName = nameof(DSCore.List.Equals);
```
mjkkirschner: did you want to use this in the AST?

src/Libraries/CoreNodeModels/Equals.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        [JsonConstructor]
+        private Equals(IEnumerable<PortModel> inPorts, IEnumerable<PortModel> outPorts) : base(inPorts, outPorts)
+        {
+        }
+
+        public Equals()
+        {
+            InPorts.Add(new PortModel(PortType.Input, this, new PortData("lhs", "integer or double value")));
```
mjkkirschner: these should be localized right?

src/Libraries/CoreNodeModels/Equals.cs
```diff
@@ -0,0 +1,108 @@
...
+using Dynamo.Graph.Nodes;
+using Newtonsoft.Json;
+using ProtoCore.AST.AssociativeAST;
+using ProtoCore.DSASM;
+using ProtoCore.Utils;
+
+namespace CoreNodeModels
+{
+    [NodeName("==")]
+    [NodeDescription("Equals with tolerance input")]
```
mjkkirschner: localized resx strings please.

test/Engine/FFITarget/TestDefaultArgumentAttributes.cs
```diff
@@ -16,5 +18,10 @@ public static double ComputeCircleConflict([DefaultArgument("Point.ByCoordinates
         {
             return radius * radius * Math.PI;
         }
+
+        public static IEnumerable<Point> ComputeFlattenedPoints([DefaultArgument("FFITarget.DesignScript.Point.ByCoordinates((1..10)<1>, (2..20)<2>, 0)")]IEnumerable<IEnumerable<Point>> points)
```
mjkkirschner: is this from this PR?

src/Libraries/CoreNodeModels/Equals.cs
mjkkirschner: ah, I see your TODO statement.

src/Libraries/CoreNodeModels/Equals.cs
mjkkirschner: @aparajit-pratap 
UseLevelAndReplicationGuide()

and it seems this is missing from a bunch of nodeModel nodes .... 😢

test/Engine/FFITarget/TestDefaultArgumentAttributes.cs
aparajit-pratap: This isn't related to the PR. I just thought it would be useful to add a test with replication guides in default argument attributes. Will submit this together with the test case in a separate PR.

src/Libraries/CoreNodeModels/Equals.cs
aparajit-pratap: This is not used, I don't think. Ah, I see this is the wrong function. Will clean it up.

src/Libraries/CoreNodeModels/Equals.cs
aparajit-pratap: Yeah I think I had created a task once upon a time to add these to other nodes like the Range node.

aparajit-pratap: https://jira.autodesk.com/browse/DYN-1034

aparajit-pratap: Hmm, in the case of the Range node, the option is enabled but doesn't work due to a missing `UseLevelAndReplicationGuide` I'm guessing.

aparajit-pratap: Further investigation reveals that the Range UI node compiles down to the range expression in DS and since replication guides don't work for the range expression (filed as a separate bug), enabling the `UseLevelAndReplicationGuide` may not suffice.

src/Libraries/CoreNodeModels/Logic/If.cs
```diff
@@ -17,10 +17,15 @@ namespace CoreNodeModels.Logic
     public class If : NodeModel
     {
         [JsonConstructor]
-        private If(IEnumerable<PortModel> inPorts, IEnumerable<PortModel> outPorts) : base(inPorts, outPorts) { }
+        private If(IEnumerable<PortModel> inPorts, IEnumerable<PortModel> outPorts) : base(inPorts, outPorts)
```
aparajit-pratap: These changes basically add lacing capability to the If UI node however they are not quite related to this JIRA task.

src/Engine/ProtoCore/Utils/MathUtils.cs
```diff
@@ -4,7 +4,7 @@ namespace ProtoCore.Utils
 {
     public static class MathUtils
     {
-        public static double Tolerance = 1e-5;
```
mjkkirschner: is this an API break?
I guess the value is still accessible inside the static class?

src/Engine/ProtoCore/Utils/MathUtils.cs
```diff
@@ -30,15 +30,15 @@ public static bool IsLessThanOrEquals(double lhs, double rhs)
             return (lhs < rhs) || Equals(lhs, rhs);
         }
 
-        public static bool Equals(double lhs, double rhs)
+        public static bool Equals(double lhs, double rhs, double tolerance = Tolerance)
```
mjkkirschner: this is an API break. Sorry 😢

src/Libraries/CoreNodeModels/Equals.cs
```diff
@@ -0,0 +1,92 @@
...
+        {
+            ArgumentLacing = LacingStrategy.Auto;
+            inPorts.ElementAt(2).DefaultValue = tolerancePortDefaultValue;
+        }
+
+        public Equals()
+        {
+            ArgumentLacing = LacingStrategy.Auto;
+
+            InPorts.Add(new PortModel(PortType.Input, this, new PortData("lhs", "integer or double value")));
```
mjkkirschner: can we use x and y instead of lhs and rhs?

src/Libraries/CoreNodeModels/Logic/If.cs
```diff
@@ -58,6 +63,8 @@ public override IEnumerable<AssociativeNode> BuildOutputAst(List<AssociativeNode
             }
             else
             {
+                UseLevelAndReplicationGuide(inputAstNodes);
```
mjkkirschner: @Amoursol @Racel  FYI - I wonder why this was left off in the first place, any memory?

src/Libraries/CoreNodes/Math.cs
```diff
@@ -2,9 +2,11 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Windows.Media;
```
mjkkirschner: we can't do this. These libraries need to be cross platform.

mjkkirschner: what is using this anyway?

src/Engine/ProtoCore/Utils/MathUtils.cs
aparajit-pratap: Fixed.

src/Libraries/CoreNodeModels/Equals.cs
aparajit-pratap: Done

src/Libraries/CoreNodes/Math.cs
aparajit-pratap: Not sure. Removed.

src/Libraries/CoreNodeModels/Logic/If.cs
mjkkirschner: @aparajit-pratap I think we need to be careful adding this in this task - what if it changes the behavior of the IF node in existing graphs - in the default constructor do we need to set the lacing? Won't that come from the deserialized node?

aparajit-pratap: The default will be auto mode, which is the same if lacing is absent altogether so I don't think it will change the behavior of existing nodes. You're right about the deserialized node. Only if the lacing info is absent it should be sent to auto otherwise it should be read from the file. Will make the change.

aparajit-pratap: Hmm, it looks like this change does not affect existing nodes at all (neither xml nor json files). The `If` node if deserialized from an old file does not have the lacing option at all just like before. In a nutshell this change will affect only new nodes made from this point on.

aparajit-pratap:
@mjkkirschner removed `If` node changes. Merging this. Will create separate PR for icon. Thanks for the review!

test/Engine/FFITarget/ProtoFFITests.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Xml.Schema;
```
mjkkirschner: @aparajit-pratap is this needed?

mjkkirschner:
@aparajit-pratap looks solid! - just one last question. LGTM

PR #6724, opened 2016-06-12
Magn-9966 Crash when closing a Custom Node with a Group when in Sweden or France

sm6srw:
### Purpose

Make sure that all convert methods are using the correct culture when converting to double. There are currently three different convert methods where we don't do that. These are:

ZoomToVisibilityConverter
MenuItemCheckConverter
GroupFontSizeToEditorEnabledConverter

These convert methods are in 99% of all cases passed a double at convert time and the current culture doesn't matter but there are a few cases where a string is passed in instead and that is when we will get invalid format exceptions in regions of the world that use "," instead of "." as the decimal point.

### Declarations

Check these if you believe they are true

- [X] The code base is in a better state after this PR
- [X] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [X] The level of testing this PR includes is appropriate
- [] User facing strings, if any, are extracted into `*.resx` files
- [X] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.

### Reviewers

@ramramps 

### FYIs

@mjkkirschner @QilongTang

Issue #6261, opened 2016-03-15
Groups inside Custom node are crashing dynamo


ondejcik:
## Dynamo version

0.9.1 , 0.9.2 , 1.0.1 build 20160415T0446

## Operating system

W7 64bit

## What did you do? 

Groups inside custom node editor are crashing Dynamo when Custom node editor is closed.

Error report:

Input string was not in a correct format.

   at System.Number.ParseDouble(String value, NumberStyles options, NumberFormatInfo numfmt)
   at System.Double.Parse(String s, NumberStyles style, IFormatProvider provider)
   at Dynamo.Controls.GroupFontSizeToEditorEnabledConverter.Convert(Object[] values, Type targetType, Object parameter, CultureInfo culture)
   at System.Windows.Data.MultiBindingExpression.TransferValue()
   at System.Windows.Data.MultiBindingExpression.HandlePropertyInvalidation(DependencyObject d, DependencyPropertyChangedEventArgs args)
   at System.Windows.Data.BindingExpressionBase.OnPropertyInvalidation(DependencyObject d, DependencyPropertyChangedEventArgs args)
   at System.Windows.DependentList.InvalidateDependents(DependencyObject source, DependencyPropertyChangedEventArgs sourceArgs)
   at System.Windows.DependencyObject.NotifyPropertyChange(DependencyPropertyChangedEventArgs args)
   at System.Windows.DependencyObject.UpdateEffectiveValue(EntryIndex entryIndex, DependencyProperty dp, PropertyMetadata metadata, EffectiveValueEntry oldEntry, EffectiveValueEntry& newEntry, Boolean coerceWithDeferredReference, Boolean coerceWithCurrentValue, OperationType operationType)
   at System.Windows.TreeWalkHelper.OnInheritablePropertyChanged(DependencyObject d, InheritablePropertyChangeInfo info, Boolean visitedViaVisualTree)
   at System.Windows.DescendentsWalker`1._VisitNode(DependencyObject d, Boolean visitedViaVisualTree)
   at System.Windows.DescendentsWalker`1.WalkFrameworkElementLogicalThenVisualChildren(FrameworkElement feParent, Boolean hasLogicalChildren)
   at System.Windows.DescendentsWalker`1.IterateChildren(DependencyObject d)
   at System.Windows.DescendentsWalker`1.StartWalk(DependencyObject startNode, Boolean skipStartNode)
[...]

PR #10006, opened 2019-09-25
use invariant culture to parse inputs to converters -

mjkkirschner:
we were passing the culture before but this was usually set to en-us EVEN in localized environments
unless something set it explicitly
wrap with try catch and return some default conversion

### Purpose

This one has a bit of a history.
issue from user:
https://github.com/DynamoDS/Dynamo/issues/6261
fix:
https://github.com/DynamoDS/Dynamo/pull/6724
fix:
https://github.com/DynamoDS/Dynamo/pull/9994

When this was fixed originally by @sm6srw it appears that unlike what it implies in his PR - unintentionally the culture might have always been set to `en-us` as @angelowang points out WPF passes en-us culture to the converters EVEN if the environment is localized. (I tried both localizing dynamo and localizing windows and got en-us everytime!)

SO - this means that the original fix three years ago, was close, but in certain situations it lets the host change the culture - we don't want that.

Our strings are hardcoded in en-us format in our xaml, and I think that is okay, these converters are not really for public consumption. SO this pr does the following.

* switches to using invariant culture
* documents that the culture parameter is unusued on the convert method
* adds some more test cases (the fr-FR cases failed before this PR)
* wraps the conversions in try catch blocks which logs the exception - incase something fails or someone uses these converters from an extension or in an unanticipated way.
* adds a TODO to remove a converter I cannot find used anywhere.

I've added similar logic to two other converters that @sm6srw added culture to. It appears to me the only time strings are used on these converters is from hardcoded xaml values, or the tests.

I still need to run the self serve tests.
- [x] cherry pick to 2.4.1 

### Declarations

### Reviewers

@QilongTang 
@sm6srw 
### FYIs

@angelowang

### DIFF
--- a/src/DynamoCoreWpf/UI/Converters.cs
+++ b/src/DynamoCoreWpf/UI/Converters.cs
@@ -1414,7 +1414,7 @@ namespace Dynamo.Controls
         }
     }
 
-    //TODO remove(this is not used anywhere)
+    //TODO remove(this is not used anywhere) in Dynamo 3.0
     public class ZoomToVisibilityConverter : IValueConverter
     {
         /// <summary>
--- a/src/DynamoCoreWpf/UI/Converters.cs
+++ b/src/DynamoCoreWpf/UI/Converters.cs
@@ -2322,7 +2322,7 @@ namespace Dynamo.Controls
             //just use defaults, this will enable the text editor.
             catch(Exception e)
             {
-                Console.WriteLine($"problem attempting to parse fontsize or zoom {values[0]} {values[1]} { e.Message}");
+                Console.WriteLine($"problem attempting to parse fontsize or zoom {values[1]} {values[0]}. { e.Message}");
             }
 
             var factor = zoom*fontsize;
--- a/src/DynamoCoreWpf/UI/Converters.cs
+++ b/src/DynamoCoreWpf/UI/Converters.cs
@@ -1414,15 +1414,29 @@ namespace Dynamo.Controls
         }
     }
 
+    //TODO remove(this is not used anywhere)
     public class ZoomToVisibilityConverter : IValueConverter
     {
+        /// <summary>
+        /// Returns hidden for small zoom sizes - appears unused.
+        /// </summary>
+        /// <param name="value">zoom size</param>
+        /// <param name="targetType">unused</param>
+        /// <param name="parameter">unused</param>
+        /// <param name="culture">unused</param>
+        /// <returns></returns>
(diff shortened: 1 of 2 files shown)


src/DynamoCoreWpf/UI/Converters.cs
```diff
@@ -1414,15 +1414,29 @@ public object ConvertBack(object value, Type targetType, object parameter, Syste
         }
     }
 
+    //TODO remove(this is not used anywhere)
```
QilongTang: Can you add Dynamo 3.0? Will be easier to identify when the time comes :)

src/DynamoCoreWpf/UI/Converters.cs
```diff
@@ -2265,10 +2298,32 @@ public class GroupFontSizeToEditorEnabledConverter : IMultiValueConverter
...
+                //use invariantCulture - 
+                //fontSize should only be serialized in invariant culture
+                // and zoom should either come from fallback value or runtime value.
+                zoom = System.Convert.ToDouble(values[0], CultureInfo.InvariantCulture);
+                fontsize = System.Convert.ToDouble(values[1], CultureInfo.InvariantCulture);
+            }
+            //just use defaults, this will enable the text editor.
+            catch(Exception e)
+            {
+                Console.WriteLine($"problem attempting to parse fontsize or zoom {values[0]} {values[1]} { e.Message}");
```
QilongTang: How about 
```
$"problem attempting to parse fontsize {values[1]} or zoom {values[0]}.  { e.Message}"
```

src/DynamoCoreWpf/UI/Converters.cs
mjkkirschner:
```suggestion
    //TODO remove(this is not used anywhere) in Dynamo 3.0
```

src/DynamoCoreWpf/UI/Converters.cs
```diff
@@ -1414,15 +1414,29 @@ public object ConvertBack(object value, Type targetType, object parameter, Syste
         }
     }
 
+    //TODO remove(this is not used anywhere) in Dynamo 3.0
```
mjkkirschner: @saintentropy this is the converter I was talking about .... is this the one you were thinking of? I can't find this particular one used anywhere.

src/DynamoCoreWpf/UI/Converters.cs
mjkkirschner:
```suggestion
                Console.WriteLine($"problem attempting to parse fontsize or zoom {values[1]} {values[0]}. { e.Message}");
```

PR #10007, opened 2019-09-25
fix exception being thrown during shutdown of workspace dep viewer

mjkkirschner:
### Purpose

`LoadedParams` was never set but was being referenced during shutdown - 
shutdown calls for view extensions are wrapped in a try catch so it doesn't crash.
It's possible this causes some leaks during the test runs as the events in the shutdown handler were not unsubscribed.

### Declarations

### Reviewers

@reddyashish 
@ColinDayOrg

### DIFF
--- a/src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
+++ b/src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
@@ -83,6 +83,7 @@ namespace Dynamo.WorkspaceDependency
 
         public void Loaded(ViewLoadedParams viewLoadedParams)
         {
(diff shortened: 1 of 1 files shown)


mjkkirschner:
https://github.com/DynamoDS/Dynamo/blob/edcfbba0374c676cbefe1494f9b5fa3bb37bad44/src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs#L66

is the shutdown method which was throwing.

ColinDayOrg:
LGTM

QilongTang:
According to conversation with @mjkkirschner , I will include some code refactor in PR https://github.com/DynamoDS/Dynamo/pull/9993 to address this  since it is also related to implementing dispose function on dep view

PR #10011, opened 2019-09-26
[DYN-1454] Unresolved nodes should attempt to be resolved when a new package is loaded

reddyashish:
### Purpose

This PR is to resolve the dummy nodes once the package is downloaded. 
Task: https://jira.autodesk.com/browse/DYN-1454

Once the package is downloaded, the VM is reset and it reloads all the packages. When this is completed in here: https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Engine/EngineController.cs#L505, we attempt to resolve the dummy node's that are present in the current workspace. The ReloadDummyNodes event is triggered and the model will deserialize each of the dummy nodes. If any of them is resolved by the package that was downloaded, then that dummy node will be replaced in the workspace and all the properties that were present before will be set. If they are not resolved, we just skip them. We disable the executions temporarily when the dummy node is replaced by the new resolved node. Once all the dummy nodes are checked, a final run is triggered here: https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs#L326. 

Once all the dummy nodes are reloaded, a TriggerDependencyRegen is triggered on the Workspace Model from the Dynamo Model. This will regenerate the Dependency table and update the package dependency states on the Workspace Dependency View extension. 

Reload dummy node new

### Declarations

### Reviewers

@mjkkirschner @aparajit-pratap @QilongTang @scottmitchell


aparajit-pratap:
> Once all the dummy nodes are reloaded, a TriggerDependencyRegen is triggered on the Workspace Model from the Dynamo Model. This will regenerate the Dependency table and update the package dependency states on the Workspace Dependency View extension.
Do you need to this? Did that not work previously?

reddyashish:
The DependencyRegen was being called as soon as the package is loaded. Then the VM is reset and reloads all the libraries. We have added the Reloading of dummy nodes after the VM is reset so the dependency table was not being updated. So this was needed after we attempt to reload all the dummy nodes.

src/DynamoCore/Engine/EngineController.cs
```diff
@@ -38,6 +38,11 @@ public class EngineController : LogSourceBase, IAstNodeContainer, IDisposable
         /// </summary>
         public event AstBuiltEventHandler AstBuilt;
 
+        /// <summary>
+        /// This event is fired to reload the dummy nodes.
+        /// </summary>
+        public static event Action ReloadDummyNodes;
```
mjkkirschner: think about this when naming events 

```
Do name events with a verb or a verb phrase.

Do give event names a concept of before and after, using the present and past tense. For example, a close event that is raised before a window is closed would be called Closing and one that is raised after the window is closed would be called Closed.

Do not use Before or After prefixes or suffixes to indicate pre and post events.

Do name event handlers (delegates used as types of events) with the EventHandler suffix.

Do use two parameters named sender and e in event handler signatures.

The sender parameter should be of type Object, and the e parameter should be an instance of or inherit from EventArgs.

Do name event argument classes with the EventArgs suffix.
```

mjkkirschner: why is this static? Is there a good reason?

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -188,6 +188,19 @@ internal int CurrentPasteOffset
         /// </summary>
         private bool workspaceLoaded;
 
+        /// <summary>
+        /// This event is to trigger the DependencyRegen on the WorkspaceDependencyview extension.
+        /// </summary>
+        public static event Action TriggerDependencyRegen;
```
mjkkirschner: I think the event should not be way to signal just that we want to trigger some other method, but more about the state of the workspace - 

like

`DummyNodesResolved`

why is this static- is there a good reason?

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -188,6 +188,19 @@ internal int CurrentPasteOffset
...
+
+        /// <summary>
+        /// This method invokes the TriggerDependencyRegen event on the workspace model.
+        /// </summary>
+        public void TriggerDependencyRegenEvent()
```
mjkkirschner: Usually this would be called - 
OnEventName

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -727,6 +727,13 @@ protected DynamoModel(IStartConfiguration config)
             LibraryServices.MessageLogged += LogMessage;
             LibraryServices.LibraryLoaded += LibraryLoaded;
 
+            EngineController = new EngineController(
```
mjkkirschner: at first glance this scares me - why do we need to create a new EngineController instance???

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1715,6 +1724,81 @@ private void SetPeriodicEvaluation(WorkspaceModel ws)
             return true;
         }
 
+        // Reload's the dummy nodes and serializes the new node information once the package resolves it.
```
mjkkirschner: this comment is a bit hard to make sense of.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -116,6 +116,14 @@ internal void DependencyRegen(WorkspaceModel ws)
             PackageDependencyTable.ItemsSource = packageDependencies.Select(d => new PackageDependencyRow(d as PackageDependencyInfo));
         }
 
+        /// <summary>
+        /// Calls the DependencyRegen function when the event is triggered from the dynamo model.
+        /// </summary>
+        internal void TriggerDepenRegen()
```
mjkkirschner: whats going on here? Why does this need to live in the view?

test/DynamoCoreTests/PackageDependencyTests.cs
```diff
@@ -502,5 +503,22 @@ public void PackageDependencyStatechangeTestAfterLoadingPackage()
...
+
+            var dummyNodes = CurrentDynamoModel.CurrentWorkspace.Nodes.OfType<DummyNode>();
+            Assert.AreEqual(2, dummyNodes.Count());
+
+            // Load the HowickMaker dll and check that the dummy nodes have been resolved.
+            string packageDirectory = Path.Combine(TestDirectory, @"core\packageDependencyTests\HowickMaker");
+            LoadPackage(packageDirectory);
+
+            dummyNodes = CurrentDynamoModel.CurrentWorkspace.Nodes.OfType<DummyNode>();
+            Assert.AreEqual(0, dummyNodes.Count());
```
mjkkirschner: Can you also assert the value of some node in the graph which depends on the dummy node being resolved?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1715,6 +1724,81 @@ private void SetPeriodicEvaluation(WorkspaceModel ws)
             return true;
         }
 
+        // Reload's the dummy nodes and serializes the new node information once the package resolves it. 
+        private void ReloadDummyNodes()
+        {
+            JObject dummyNodeJSON = null;
+            WorkspaceModel currentWorkspace = this.CurrentWorkspace;
+
+            if (currentWorkspace == null || string.IsNullOrEmpty(currentWorkspace.FileName))
```
mjkkirschner: what happens if this is a customNode workspace? Have you tested that? I think we need a test for that case.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1715,6 +1724,81 @@ private void SetPeriodicEvaluation(WorkspaceModel ws)
...
+            {
+                return;
+            }
+
+            String filePath = Path.GetFullPath(currentWorkspace.FileName);
+
+            CustomNodeManager.AddUninitializedCustomNodesInPath(
```
mjkkirschner: why are you doing this here? What does this have to do with dummyNodes? This will load custom nodes??

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1715,6 +1724,81 @@ private void SetPeriodicEvaluation(WorkspaceModel ws)
...
+                                                               dummyNodeJSON.ToString(),
+                                                               LibraryServices,
+                                                               EngineController,
+                                                               Scheduler,
+                                                               NodeFactory,
+                                                               IsTestMode,
+                                                               false,
+                                                               CustomNodeManager);
+
+                    if (!(resolvedNode is DummyNode))
```
mjkkirschner: I think a comment is in order here.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1715,6 +1724,81 @@ private void SetPeriodicEvaluation(WorkspaceModel ws)
...
+                    {
+                        currentWorkspace.SetNodeViewDataOnResolvedNode(dn, resolvedNode);
+
+                        // Disable Run's temporarlity while the dummy node is removed and the resolved node is added.
```
mjkkirschner: Disable graph runs temporarily while the dummy node is replaced with the resolved version of that node.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1715,6 +1724,81 @@ private void SetPeriodicEvaluation(WorkspaceModel ws)
...
+                                                               CustomNodeManager);
+
+                    if (!(resolvedNode is DummyNode))
+                    {
+                        currentWorkspace.SetNodeViewDataOnResolvedNode(dn, resolvedNode);
+
+                        // Disable Run's temporarlity while the dummy node is removed and the resolved node is added. 
+                        EngineController.DisableRun = true;
+                        currentWorkspace.RemoveAndDisposeNode(dn);
+                        currentWorkspace.AddAndRegisterNode(resolvedNode, false);
```
mjkkirschner: does this mark the workspace modified (unsaved changes?) should we mark that false before the function returns?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1715,6 +1724,81 @@ private void SetPeriodicEvaluation(WorkspaceModel ws)
...
+
+                        // Adding back the connectors for the resolved node. 
+                        List<ConnectorModel> connectors = dn.AllConnectors.ToList();
+                        foreach (var connectorModel in connectors)
+                        {
+                            var startNode = connectorModel.Start.Owner;
+                            var endNode = connectorModel.End.Owner;
+
+                            if (startNode is DummyNode && startNode.Name == resolvedNode.Name)
```
mjkkirschner: what is this bit about?

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -124,6 +132,7 @@ public WorkspaceDependencyView(WorkspaceDependencyViewExtension viewExtension,Vi
         {
             InitializeComponent();
             currentWorkspace = p.CurrentWorkspaceModel as WorkspaceModel;
+            WorkspaceModel.TriggerDependencyRegen += TriggerDepenRegen;
```
mjkkirschner: in general unsubscribe any events that you sub to be safe unless you are sure they will not hold a reference.

test/DynamoCoreTests/PackageDependencyTests.cs
```diff
@@ -502,5 +503,22 @@ public void PackageDependencyStatechangeTestAfterLoadingPackage()
...
+        public void ResolveDummyNodesOnDownloadingPackage()
+        {
+            string path = Path.Combine(TestDirectory, @"core\packageDependencyTests\ResolveDummyNodesOnDownloadingPackage.dyn");
+            OpenModel(path);
+
+            var dummyNodes = CurrentDynamoModel.CurrentWorkspace.Nodes.OfType<DummyNode>();
+            Assert.AreEqual(2, dummyNodes.Count());
+
+            // Load the HowickMaker dll and check that the dummy nodes have been resolved.
+            string packageDirectory = Path.Combine(TestDirectory, @"core\packageDependencyTests\HowickMaker");
```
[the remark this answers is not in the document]
reddyashish: Yes, the dll is added to the test directory and I am loading it from there.

src/DynamoCore/Models/DynamoModel.cs
reddyashish: When I am adding back the connectors, if the startNode or endNode was a dummy node(and it is being replaced with the resolved node), then the new connector should be added on the resolved node. Since the graph could have many dummy nodes, I need to check the GUID before setting the startNode and endNode for the new connector. I have changed this check the GUID property.

src/DynamoCore/Models/DynamoModel.cs
reddyashish: It does mark that the workspace is modified when the node and connectors are added. The * symbol appears on the workspace.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1715,6 +1724,81 @@ private void SetPeriodicEvaluation(WorkspaceModel ws)
...
+                                                               IsTestMode,
+                                                               false,
+                                                               CustomNodeManager);
+
+                    if (!(resolvedNode is DummyNode))
+                    {
+                        currentWorkspace.SetNodeViewDataOnResolvedNode(dn, resolvedNode);
+
+                        // Disable Run's temporarlity while the dummy node is removed and the resolved node is added. 
+                        EngineController.DisableRun = true;
```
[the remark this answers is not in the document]
reddyashish: We do this because for every node and every connector that is added on the workspace, a graph run is triggered and it causes overhead. It doesn't cause any issues but we do not need it. Once all the dummy nodes are resolved, one final run is triggered in the end.

src/DynamoCore/Engine/EngineController.cs
mjkkirschner: thinking more, I think this should be named to represent the state that is happening, not one single method you are using to the event to call...

so sometihng like `VMLibrariesReset`

src/DynamoCore/Models/DynamoModel.cs
reddyashish: Yes, I forgot to remove this, we do not want to create a new EngineController object. I have removed this now.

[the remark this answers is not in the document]
reddyashish: Yes, removed this from here.

src/DynamoCore/Models/DynamoModel.cs
reddyashish: Sorry removing this as well. Not needed.

src/DynamoCore/Models/DynamoModel.cs
reddyashish: Changed the description.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
reddyashish: Renamed the event name based on the state and made it non-static.

src/DynamoCore/Engine/EngineController.cs
reddyashish: Makes sense. Changed the event name accordingly and the event is not static now.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
reddyashish: This is just a helper method to call the DependencyRegen function.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
reddyashish: Actually, this has to be static. Because when this event is attached to the current workspace in here (https://github.com/DynamoDS/Dynamo/pull/10011/files#diff-14bfe53b9085e8f1b47772149559d0a7R135), the workspace object is different. So had to keep it static for now.

test/DynamoCoreTests/PackageDependencyTests.cs
reddyashish: Yes, asserting the value of the node in the test.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1715,6 +1724,81 @@ private void SetPeriodicEvaluation(WorkspaceModel ws)
...
+                                                               EngineController,
+                                                               Scheduler,
+                                                               NodeFactory,
+                                                               IsTestMode,
+                                                               false,
+                                                               CustomNodeManager);
+
+                    if (!(resolvedNode is DummyNode))
+                    {
+                        currentWorkspace.SetNodeViewDataOnResolvedNode(dn, resolvedNode);
```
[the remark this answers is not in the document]
reddyashish: We can do that but we do not want to set these properties if the resolvedNode (that is returned from GetNodeModelForDummyNode function) is also a dummy node. Since we are already checking that in DynamoModel.cs after the nodemodel is returned, it would be one extra step if we want to do it in GetNodeModelForDummyNode().

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -1852,6 +1865,80 @@ protected void Log(string msg, WarningLevel severity)
...
+        /// </summary>
+        /// <param name="json"></param>
+        /// <param name="libraryServices"></param>
+        /// <param name="engineController"></param>
+        /// <param name="scheduler"></param>
+        /// <param name="factory"></param>
+        /// <param name="isTestMode"></param>
+        /// <param name="verboseLogging"></param>
+        /// <param name="manager"></param>
+        public static NodeModel GetNodeModelForDummyNode(string json, LibraryServices libraryServices,
```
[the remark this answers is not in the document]
reddyashish: I have moved GetNodeModelForDummyNode and SetNodeViewDataOnResolvedNode API's to the DummyNode class as it makes it easier to find them in future improvements.

src/DynamoCore/Models/DynamoModel.cs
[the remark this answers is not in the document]
reddyashish: @aparajit-pratap SetNodeViewDataOnResolvedNode() is now being called from inside GetNodeModelForDummyNode method.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
mjkkirschner: @reddyashish couldn't you just attach this handler when the workspace changes?

src/DynamoCore/Engine/EngineController.cs
```diff
@@ -38,6 +38,11 @@ public class EngineController : LogSourceBase, IAstNodeContainer, IDisposable
         /// </summary>
         public event AstBuiltEventHandler AstBuilt;
 
+        /// <summary>
+        /// This event is notify that the VMLibraries have been reset.
```
mjkkirschner: `The event notifies clients that the VMLibraries have been reset.`
@aparajit-pratap is there any other info we want to add to this comment about this state - like - it is now safe to run code?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1715,6 +1719,73 @@ private void SetPeriodicEvaluation(WorkspaceModel ws)
...
+                            ConnectorModel.Make(startNode, endNode, connectorModel.Start.Index, connectorModel.End.Index, connectorModel.GUID);
+                        }
+                        EngineController.DisableRun = false ;
+                    }
+                }
+            }
+
+            // Once all the dummy nodes are reloaded, the DummyNodesReloaded event is invoked and
+            // the Dependency table is regenerated in the WorkspaceDependencyView extension. 
+            currentWorkspace.OnDummyNodesReloaded();
```
mjkkirschner: @reddyashish I think you should set the hasUnsavedChanges = false here manually and add a test for this - @aparajit-pratap what do you think?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -727,6 +727,8 @@ protected DynamoModel(IStartConfiguration config)
             LibraryServices.MessageLogged += LogMessage;
             LibraryServices.LibraryLoaded += LibraryLoaded;
 
+            EngineController.VMLibrariesReset += ReloadDummyNodes;
```
aparajit-pratap: I think this should be done after the call to `ResetEngineInternal` as a new `EngineController` could be created there, in which case this event will not be subscribed by the new controller.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1715,6 +1719,73 @@ private void SetPeriodicEvaluation(WorkspaceModel ws)
...
+                    // If the resolved node is also a dummy node, then skip it else replace the dummy node with the resolved version of the node. 
+                    if (!(resolvedNode is DummyNode))
+                    {
+                        // Disable graph runs temporarily while the dummy node is replaced with the resolved version of that node.
+                        EngineController.DisableRun = true;
+                        currentWorkspace.RemoveAndDisposeNode(dn);
+                        currentWorkspace.AddAndRegisterNode(resolvedNode, false);
+
+                        // Adding back the connectors for the resolved node. 
+                        List<ConnectorModel> connectors = dn.AllConnectors.ToList();
```
aparajit-pratap: Can you still access the connectors of the `dn` after disposing it above? How about disposing it after this step to make the sequence of steps more logical?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1715,6 +1719,73 @@ private void SetPeriodicEvaluation(WorkspaceModel ws)
...
+        {
+            JObject dummyNodeJSON = null;
+            WorkspaceModel currentWorkspace = this.CurrentWorkspace;
+
+            if (currentWorkspace == null || string.IsNullOrEmpty(currentWorkspace.FileName))
+            {
+                return;
+            }
+
+            String filePath = Path.GetFullPath(currentWorkspace.FileName);
```
aparajit-pratap: I don't think `filePath` is used anywhere.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1715,6 +1719,73 @@ private void SetPeriodicEvaluation(WorkspaceModel ws)
...
+            // Get the dummy nodes in the current workspace. 
+            var dummyNodes = currentWorkspace.Nodes.OfType<DummyNode>();
+
+            foreach (DummyNode dn in dummyNodes)
+            {
+                dummyNodeJSON = dn.OriginalNodeContent as JObject;
+
+                if (dummyNodeJSON != null)
+                {
+                    // Deserializing the dummy node and verifying if it is resolved by the downloaded package
```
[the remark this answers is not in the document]
aparajit-pratap: Agree, it would be worth testing if the dummy nodes are also resolved upon importing a dependency (dll) using `File->Import`, not just package download.

src/DynamoCore/Engine/EngineController.cs
aparajit-pratap: Yes, we can add that the `VM is now _ready_ to run new code`.

src/DynamoCore/Models/DynamoModel.cs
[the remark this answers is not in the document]
reddyashish: Yes, tested the package import workflow and the dummy nodes are resolved as well upon package import.

src/DynamoCore/Models/DynamoModel.cs
reddyashish: Yes, I can because because the dummy node is removed from the current workspace but I have the dummy node object before the foreach loop.

src/DynamoCore/Models/DynamoModel.cs
reddyashish: Yes, removing this.

src/DynamoCore/Models/DynamoModel.cs
reddyashish: Setting hasUnsavedChanges=false and the workspace changes will be saved automatically as there are no changes to the serialized file. Will add this to the test as well.

src/DynamoCore/Models/DynamoModel.cs
reddyashish: Moved it.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -422,6 +421,7 @@ public IEnumerable<WorkspaceModel> Workspaces
         /// Returns authentication manager object for oxygen authentication.
         /// </summary>
         public AuthenticationManager AuthenticationManager { get; set; }
+        public object DependencyView { get; private set; }
```
mjkkirschner: @reddyashish what is this for?

src/DynamoCore/Graph/Nodes/DummyNode.cs
```diff
@@ -398,5 +404,76 @@ private XmlElement OriginalXmlNodeContent
...
+                resolvedNode.IsSetAsInput = dummyNode.IsSetAsInput;
+                resolvedNode.IsSetAsOutput = dummyNode.IsSetAsOutput;
+
+                // NOTE: The name needs to be set using UpdateValue to cause the view to update
+                resolvedNode.UpdateValue(new UpdateValueParams("Name", dummyNode.Name));
+                resolvedNode.UpdateValue(new UpdateValueParams("IsVisible", dummyNode.IsVisible.ToString()));
+            }
+            else
+            {
+                this.Log(string.Format("This graph has a node with id:{0} and name:{1}, but it could not be resolved",
```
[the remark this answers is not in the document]
reddyashish: Yes, moved the warning statement to the GetNodeModelForDummyNode(). The resolvedNodeModel cannot be null here so I am just checking for the dummy node

src/DynamoCore/Models/DynamoModel.cs
reddyashish: This is not needed as well and I should have removed. Thanks for catching it.

reddyashish:
I have a bunch of failures on Self CI but they are passing locally. Looking into it more.

reddyashish:
@aparajit-pratap @mjkkirschner The reason for those failures was due to the event in here: https://github.com/DynamoDS/Dynamo/pull/10011/files#diff-14bfe53b9085e8f1b47772149559d0a7R135

I have removed the invocation of DependencyRegen event for now and modified the test to use the dynamo samples and not HowickMaker. Waiting on the self serve results to confirm there are no failures. I will then remove the commented part of the code. Looking into other alternatives in the mean time.

mjkkirschner:
@reddyashish this looks like it could have been a memory leak -

reddyashish:
Self-serve has passed now. https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/424/

Please take a look and this is ready to be merged.

test/DynamoCoreTests/DummyNodeTests.cs
```diff
@@ -88,5 +97,29 @@ public void WorkspaceIsReadonly_IfXmlDummyNodePresent()
...
+            string path = Path.Combine(TestDirectory, @"core\packageDependencyTests\ResolveDummyNodesOnDownloadingPackage.dyn");
+            OpenModel(path);
+
+            var dummyNodes = CurrentDynamoModel.CurrentWorkspace.Nodes.OfType<DummyNode>();
+            Assert.AreEqual(2, dummyNodes.Count());
+
+            var output = GetPreviewValue("a1aba50a873443f2bfb88480a89e3f36");
+            Assert.IsNull(output);
+
+            // Load the HowickMaker dll and check that the dummy nodes have been resolved.
```
mjkkirschner: comment is out of date.

reddyashish: Changed it. Thanks Mike.

src/DynamoCore/Engine/EngineController.cs
```diff
@@ -38,6 +38,11 @@ public class EngineController : LogSourceBase, IAstNodeContainer, IDisposable
         /// </summary>
         public event AstBuiltEventHandler AstBuilt;
 
+        /// <summary>
+        /// The event notifies client that the VMLibraries have been reset and the VM is now ready to run the new code. 
+        /// </summary>
+        public static event Action VMLibrariesReset;
```
mjkkirschner: what do you think about making this internal? Is it useful to anyone but us?

mjkkirschner: do we want to continue supporting it?

src/DynamoCore/Graph/Nodes/DummyNode.cs
```diff
@@ -398,5 +404,75 @@ private XmlElement OriginalXmlNodeContent
                 return originalXmlElement;
             }
         }
+
+        /// <summary>
+        /// This will return the NodeModel for a dummy node.
```
mjkkirschner: this comment is a bit vague - what do you think about something like:
```
Deserializes and returns the nodeModel that is represented by the original content of this DummyNode. If this node cannot be resolved, returns a new DummyNode 
```

src/DynamoCore/Graph/Nodes/DummyNode.cs
```diff
@@ -398,5 +404,75 @@ private XmlElement OriginalXmlNodeContent
...
+        /// <param name="dummyNode"></param>
+        /// <param name="resolvedNode"></param>
+        private void SetNodeViewDataOnResolvedNode(NodeModel dummyNode, NodeModel resolvedNode)
+        {
+            if (dummyNode == null || resolvedNode == null)
+            {
+                return;
+            }
+
+            resolvedNode.X = dummyNode.X;
```
mjkkirschner: one thought I have about this is - is there a way to ensure we are transferring all view properties of the dummyNode - so we don't forget to update this if new view properties are added? Something like checking the ExtraNodeViewInfo object for all properties or something? This is not required - but something to think about.

test/DynamoCoreTests/DummyNodeTests.cs
```diff
@@ -24,6 +25,14 @@ protected override void GetLibrariesToPreload(List<string> libraries)
             base.GetLibrariesToPreload(libraries);
         }
 
+        private void LoadPackage(string packageDirectory)
```
aparajit-pratap: I think we can move this function to the base class as it is used in more than on derived class, this one and `PackageDependencyTests`.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -188,6 +188,19 @@ internal int CurrentPasteOffset
         /// </summary>
         private bool workspaceLoaded;
 
+        /// <summary>
+        /// This event is to notify that the dummy nodes have been resolved.
```
mjkkirschner: what about: `This event is raised after the workspace tries to resolve dummyNodes - for example after a new package or library is loaded.`

src/DynamoCore/Engine/EngineController.cs
reddyashish: I made this internal for now as it is being using only inside the DynamoCore package. Even after we refactor that workflow, we still might need this event.

test/DynamoCoreTests/DummyNodeTests.cs
```diff
@@ -88,5 +97,29 @@ public void WorkspaceIsReadonly_IfXmlDummyNodePresent()
             this.CurrentDynamoModel.CurrentWorkspace.RemoveAndDisposeNode(dummyNode);
             Assert.IsFalse(this.CurrentDynamoModel.CurrentWorkspace.IsReadOnly);
         }
+
+        [Test]
+        public void ResolveDummyNodesOnDownloadingPackage()
```
mjkkirschner: please add a test that does this same workflow *inside* a custom node - IE - the current workspace should be a custom node with dummy nodes in it.

mjkkirschner: Are there more complex tests you can think of that we might be missing? 
* dummy nodes connecting to dummy nodes?
* a test that asserts the view properties are transferred correctly? I don't see that in your test - ie check the name, position etc.

mjkkirschner:
@reddyashish - looks pretty solid - a few thoughts
* please fix up the comments a bit to make them more useful to other devs who are unfamiliar with this work
* I think this PR needs more tests.

test/DynamoCoreTests/DummyNodeTests.cs
reddyashish: Yes, moved it DynamoModelTestBase class.

test/DynamoCoreTests/DummyNodeTests.cs
reddyashish: Added the test for dummy nodes inside a custom node workspace

reddyashish:
@mjkkirschner merging this now and will create a followup PR for the view test.

mjkkirschner:
@reddyashish go for it.

PR #10014, opened 2019-09-27
sandbox should not crash silently

mjkkirschner:
### Purpose

Dynamo Sandbox should not crash silently.
* To avoid this snippet of stack trace and exception are shown in a message box from `system.windows` (not forms) to avoid adding any other dependencies that might fail to load.
* ASM load is now wrapped in a try catch
* okay button in message box navigates to dynamo builds wiki page with info on using sandbox.
* ~~filter out ASM locations that don't include tsplinesA - as these installs of ASM will not work with Dynamo. Discovered this as a cause of a silent crash with help from users on the forum and internal users.~~

### UPDATE:
* changed filemask to ASMH*A.dll to better match what dependencies we actually have, and also filter results that don't include tbb and tbbmalloc.dll
* raise an event when preload fails and trigger a messagebox to note the location we tried to load from - will be useful if we encounter other odd geometry library configurations in the future.
* Dynamo will now start, even if the preload fails. Geometry nodes will of course not work though.

Screen Shot 2019-09-27 at 6 08 54 PM
Screen Shot 2019-09-27 at 6 09 52 PM

### TODO:
- [x] use resx files

Screen Shot 2019-09-27 at 1 34 11 AM

### Declarations

### Reviewers

@aparajit-pratap 
@ColinDayOrg 

### FYIs

@TheBIMsider

### DIFF
--- a/src/DynamoSandbox/DynamoCoreSetup.cs
+++ b/src/DynamoSandbox/DynamoCoreSetup.cs
@@ -11,6 +11,7 @@ using Dynamo.Logging;
 using Dynamo.Models;
 using Dynamo.ViewModels;
 using Dynamo.Wpf.ViewModels.Watch3D;
+using System.Linq;
 
 namespace DynamoSandbox
 {
@@ -21,6 +22,7 @@ namespace DynamoSandbox
         private string commandFilePath;
         private Stopwatch startupTimer = Stopwatch.StartNew();
         private string ASMPath;
+        private const string sandboxWikiPage = @"https://github.com/DynamoDS/Dynamo/wiki/How-to-Utilize-Dynamo-Builds";
 
         [DllImport("msvcrt.dll")]
         public static extern int _putenv(string env);
@@ -95,9 +97,22 @@ namespace DynamoSandbox
                         // Give user a chance to save (but does not allow cancellation)
                         viewModel.Exit(allowCancel: false);
                     }
+                    else
+                    {
+                        //show a message dialog box with the exception so the user
+                        //can effectively report the issue.
+                        var shortStackTrace = String.Join(Environment.NewLine,e.StackTrace.Split(Environment.NewLine.ToCharArray()).Take(10));
+                        var result = MessageBox.Show($"could not start DynamoSandbox, unhandled exception {Environment.NewLine} {e.Message}  {Environment.NewLine} {e.InnerException?.Message} {Environment.NewLine} {shortStackTrace} {Environment.NewLine} " +
+                            $" Do you want to open the Dynamo builds wiki page ({sandboxWikiPage}) for more information?",
+                            "DynamoSandbox",
+                            MessageBoxButton.YesNo,MessageBoxImage.Error);
+                        if(result == MessageBoxResult.Yes)
+                        {
+                            System.Diagnostics.Process.Start(sandboxWikiPage);
+                        }
+                    }
                 }
-                catch
-                {
+                catch {
                 }
 
                 Debug.WriteLine(e.Message);
--- a/src/DynamoSandbox/DynamoCoreSetup.cs
+++ b/src/DynamoSandbox/DynamoCoreSetup.cs
@@ -103,11 +103,14 @@ namespace DynamoSandbox
                         //show a message dialog box with the exception so the user
                         //can effectively report the issue.
                         var shortStackTrace = String.Join(Environment.NewLine,e.StackTrace.Split(Environment.NewLine.ToCharArray()).Take(10));
-                        var result = MessageBox.Show($"could not start DynamoSandbox, unhandled exception {Environment.NewLine} {e.Message}" +
+
+                        var result = MessageBox.Show($"{Resources.SandboxCrashMessage} {Environment.NewLine} {e.Message}" +
                             $"  {Environment.NewLine} {e.InnerException?.Message} {Environment.NewLine} {shortStackTrace} {Environment.NewLine} " +
                              Environment.NewLine + string.Format(Resources.SandboxBuildsPageDialogMessage, sandboxWikiPage),
+
                             "DynamoSandbox",
                             MessageBoxButton.YesNo,MessageBoxImage.Error);
+
                         if(result == MessageBoxResult.Yes)
                         {
                             System.Diagnostics.Process.Start(sandboxWikiPage);
--- a/src/DynamoSandbox/DynamoCoreSetup.cs
+++ b/src/DynamoSandbox/DynamoCoreSetup.cs
@@ -12,6 +12,7 @@ using Dynamo.Models;
 using Dynamo.ViewModels;
 using Dynamo.Wpf.ViewModels.Watch3D;
 using System.Linq;
+using Dynamo.DynamoSandbox.Properties;
 
 namespace DynamoSandbox
 {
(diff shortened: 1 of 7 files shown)


src/Tools/DynamoShapeManager/Utilities.cs
```diff
@@ -308,10 +312,15 @@ public static void PreloadAsmFromPath(string preloaderLocation, string asmLocati
...
-            preloadMethod.Invoke(null, methodParams);
-
+            try
+            {
+                var methodParams = new object[] { asmLocation };
+                preloadMethod.Invoke(null, methodParams);
+            }
+            catch
+            {
+                Console.WriteLine("Failed to load ASM binaries");
```
aparajit-pratap: Won't the exception be silently swallowed out here? I don't say a throw?

mjkkirschner: yes, okay, after discussing a bit I see what you're saying - I think instead of Dynamo starting with no geometry - it is likely more useful to throw and message to the user where the geometry library failed to load from - this will also help us later get to the bottom of crash reports without needing to recreate the users local adsk install setup.

src/DynamoSandbox/DynamoCoreSetup.cs
```diff
@@ -95,9 +98,26 @@ public void RunApplication(Application app)
                         // Give user a chance to save (but does not allow cancellation)
                         viewModel.Exit(allowCancel: false);
                     }
+                    else
+                    {
+                        //show a message dialog box with the exception so the user
+                        //can effectively report the issue.
+                        var shortStackTrace = String.Join(Environment.NewLine,e.StackTrace.Split(Environment.NewLine.ToCharArray()).Take(10));
```
QilongTang: Since this is pointing to the wiki, we will need to update there how to file issue related to start up failure.

src/Tools/DynamoShapeManager/Utilities.cs
```diff
@@ -308,10 +312,15 @@ public static void PreloadAsmFromPath(string preloaderLocation, string asmLocati
...
-            preloadMethod.Invoke(null, methodParams);
-
+            try
+            {
+                var methodParams = new object[] { asmLocation };
+                preloadMethod.Invoke(null, methodParams);
+            }
+            catch
+            {
+                throw new FileNotFoundException($"Could not load geometry library binaries from : {asmLocation}");
```
QilongTang: Is this the only kind of cause exception? @mjkkirschner @aparajit-pratap

src/Tools/DynamoShapeManager/Utilities.cs
aparajit-pratap: I guess I'm okay with launching Dynamo without the geometry library but in addition to doing that I think it would be useful to display a message to users saying that the geometry library failed to load due to these reasons but if they wish they can still continue to use Dynamo.

aparajit-pratap:
@mjkkirschner I'm a little confused. With your first commit (without rethrowing from the try-catch around preloading ASM), how were you able to display the message box with the stacktrace around ASM preload. Wouldn't it have simply swallowed that exception and launched Dynamo?

src/Tools/DynamoShapeManager/Utilities.cs
```diff
@@ -147,6 +147,10 @@ public static Version GetInstalledAsmVersion2(IEnumerable<Version> versions, ref
                 getASMInstallsFunc = getASMInstallsFunc ?? GetAsmInstallations;
                 var installations = getASMInstallsFunc(rootFolder);
 
+                //filter install locations missing tsplinesA
+                installations = installations.Cast<KeyValuePair<string, Tuple<int,int,int,int>>>().
+                    Where(x => Directory.EnumerateFiles(x.Key, "tsplines*A.dll").Any());
```
aparajit-pratap: Does it make sense to fold this check into `getASMInstallsFunc`?

src/DynamoSandbox/DynamoCoreSetup.cs
mjkkirschner: yep, I will do that there.

src/Tools/DynamoShapeManager/Utilities.cs
mjkkirschner: this should just be `Exception` - I will change it.

mjkkirschner:
> @mjkkirschner I'm a little confused. With your first commit (without rethrowing from the try-catch around ASM preloading), how were you able to display the message box with the stacktrace around ASM preload? Wouldn't it have simply swallowed that exception and launched Dynamo?

- the first commit would have swallowed the exception and started dynamo - the message box I showed in the image was for any other unhandled exception - for example - if all the dlls are blocked.

src/Tools/DynamoShapeManager/Utilities.cs
mjkkirschner: 1. I can try to fold it in there - 
2. we're already checking for that in the findProductInstallations function when we pass `"ASMAHL*.dll"` isn't this - essentially doing the same check twice?

aparajit-pratap: There could be an "ASMAHL*A.dll" and an "ASMAHL*I.dll"?

aparajit-pratap:
> the first commit would have swallowed the exception and started dynamo - the message box I showed in the image was for any other unhandled exception - for example - if all the dlls are blocked.

So the message box showing ASM related exception was just a contrived example?

mjkkirschner:
> > the first commit would have swallowed the exception and started dynamo - the message box I showed in the image was for any other unhandled exception - for example - if all the dlls are blocked.
> 
> So the message box showing ASM related exception was just a contrived example?

yes - I threw away all the libG folders 😉

src/Tools/DynamoShapeManager/Utilities.cs
aparajit-pratap: I guess I meant to ask if it's possible to show the message box, the user clicks ok, and is able to continue to use Dynamo?

mjkkirschner: I don't think so - not without a larger refactoring - but I can look at that if you think it's worth it.

mjkkirschner: or not without multiple potential message boxes being shown.

mjkkirschner: hmm - for some context, we should not introduce a dependency on `system.windows` into the preloader - which needs to be cross platform - but the other message box raised in dynamoSandbox only is shown when it is too late to recover.

src/Tools/DynamoShapeManager/Utilities.cs
mjkkirschner: kind of, the offending folder has the `ASMA.dlls` and `tsplinesI.dll `

src/Tools/DynamoShapeManager/Utilities.cs
mjkkirschner: @aparajit-pratap PTAL - figured out a way to do this while avoiding big set of changes or new public apis. We may want to do that as part of the followup we discussed.

src/Tools/DynamoShapeManager/Utilities.cs
```diff
@@ -450,7 +447,11 @@ private static IEnumerable GetAsmInstallations(string rootFolder)
 
 
             var methodParams = new object[] { ProductsWithASM, ASMFileMask };
-            return installationsMethod.Invoke(null, methodParams) as IEnumerable;
+            var installs  = installationsMethod.Invoke(null, methodParams) as IEnumerable;
+
+            //filter install locations missing tbb and tbbmalloc.dll
+            return installs.Cast<KeyValuePair<string, Tuple<int, int, int, int>>>().
+                Where(x => Directory.EnumerateFiles(x.Key, "tbb*.dll").Count() >1);
```
aparajit-pratap: I think we can make this check more strict by specifically checking for both `tbb.dll` and `tbbmalloc.dll` just so we don't have any other surprises like `tbbX.dll` being present but neither being one or the other above.

mjkkirschner: okay, will do.

aparajit-pratap:
Thanks, LGTM!

mjkkirschner:
@smangarole brining this in, will cherry pick next.

PR #8649, opened 2018-03-14
Fix static property compilation and crash

aparajit-pratap:
### Purpose

Fixes QNTM-3279 and QNTM-3287. 

One of these is a crash when typing something like `Point.X` in a CBN and committing it. The reason being that the statement gets compiled to `Point.X()` and is passed a default argument. An assertion was made in the marshaler that the argument could only be a pointer and function pointer. For the time being I have also included the possibility of this argument being a default argument to prevent the crash. A more detailed investigation needs to be made as to the origin of this default argument. Ideally there should be no default argument generated and the function call should report a method resolution warning or something similar.

The second fix is to compile property calls using class names like `Dictionary.Count(dict);`. This wasn't supported previously as it would only work with a static getter, namely, `Dictionary.get_Count(dict);` which is an internal method not exposed to users. Also this fixes node to code operations on query nodes to appropriately display the method call without the getter:
image
image

### Declarations

Check these if you believe they are true

- [x] The code base is in a better state after this PR
- [x] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [x] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [x] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions), and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.

### Reviewers

@pboyer 

### FYIs

@kronz @jnealb

PR #10024, opened 2019-10-01
Fix VM crash due to undefined static method

aparajit-pratap:
### Purpose

DYN-2164
Refer to PR causing regression: https://github.com/DynamoDS/Dynamo/pull/8649

### Declarations

- [ ] The level of testing this PR includes is appropriate - Tests will be following once this is merged. See #9996 

### Reviewers

@ColinDayOrg 

### FYIs

@mjkkirschner

### DIFF
--- a/src/Engine/ProtoAssociative/CodeGen.cs
+++ b/src/Engine/ProtoAssociative/CodeGen.cs
@@ -1179,28 +1179,24 @@ namespace ProtoAssociative
                                 procName = Constants.kGetterPrefix + procName;
                                 procCallNode = classNode.GetFirstStaticFunctionBy(procName);
                                 isStaticCall = procCallNode != null;
-                                if (procCallNode == null)
-                                {
-                                    if (subPass == SubCompilePass.None)
-                                    {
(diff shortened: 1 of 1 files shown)


src/Engine/ProtoAssociative/CodeGen.cs
```diff
@@ -1179,23 +1179,26 @@ public ProtoCore.DSASM.ProcedureNode GetProcedureFromInstance(int classScope, Pr
                                 procName = Constants.kGetterPrefix + procName;
                                 procCallNode = classNode.GetFirstStaticFunctionBy(procName);
                                 isStaticCall = procCallNode != null;
-                            }
-                            else if (subPass == ProtoCore.CompilerDefinitions.SubCompilePass.None)
-                            {
-                                string message = String.Format(ProtoCore.Properties.Resources.kStaticMethodNotFound,
-                                                               className,
-                                                               procName);
+                                if (procCallNode == null)
```
aparajit-pratap: Added a check for `procCallNode` after computing it in line 1180. This check was missing earlier. In the case where it is null, we need to log a warning saying the method is not found.

PR #10026, opened 2019-10-02
DYN-2225 - Marshaling Performance optimization - Heap Cleanup

saintentropy:
### Purpose

The purpose of the PR is too improve the performance of Dynamo in the Marshaling process of .NET objects created my CLR Functions (Such as zero-touch or Node Model nodes).  This optimization specifically changes two aspects:

1)  Removes unused interactions with the DesignScript VM's Heap.  Currently, when .Net objects are created by CLR type functions invocations in Dynamo, the Heap is allocating an array of un-initialized stack values for each public property of the CLR object.  For example, a ProtoGeometry `Point` would create an array of 6 `StackValue` objects because a Point has 6 public properties (X, Y, Z, BoundingBox, ContextCoordinateSystem, and scaleFactor).  Later, if you also call a Getter type function on that object (Point.BoundingBox for example) and access a reference type property then  it's `StackValue` Pointer is updated on the Heap.   For getter's accessing value type properties, (Point.X for example) no changes are made to the array of `StackValues` on the Heap.  All that being said, at Runtime, Dynamo does not read the any of these values in the Heap.  No stack values need to be allocated or updated as access to CLR objects and their public properties is handled via a different mechanism.  The only pertinent reference in the Heap is it index so that the VM can properly GC expired CLR objects.  This PR removes the the unnecessary array allocation and value type updates for Getters to the HEAP for CLR objects.

2) This PR also adds a caching mechanism for repeated marshaling of the same type.  The repeated recursive lookup in the ClassTable can be short circuited when the same type is repeatedly being marshaled.

TODO:

- [ ] Update Failing Tests
- [ ] Review proposed changes. 

### Declarations

### Reviewers

### FYIs

@QilongTang


saintentropy:
image
Here is the Perf Data

src/Engine/ProtoCore/DSASM/Heap.cs
```diff
@@ -36,7 +36,11 @@ public HeapElement(int size, Heap heap)
         {
             allocated = Count = size;
             this.heap = heap;
-            values = Enumerable.Repeat(StackValue.BuildInvalid(), allocated).ToArray();
+            if (size != 0)
+            {
+                values = Enumerable.Repeat(StackValue.BuildInvalid(), allocated).ToArray();
```
mjkkirschner: if we are worrying about performance - is this worth looking at replacing with a for loop?

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
```diff
@@ -1105,20 +1108,27 @@ private StackValue CreateDSObject(object obj, ProtoCore.Runtime.Context context,
         {
             //We are here, because we want to create DS object of user defined type.
             var runtimeCore = dsi.runtime.RuntimeCore;
-            var classTable = runtimeCore.DSExecutable.classTable;
             Type objType = GetPublicType(obj.GetType());
```
mjkkirschner: this also seems recursive...?

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
```diff
@@ -1105,20 +1108,27 @@ private StackValue CreateDSObject(object obj, ProtoCore.Runtime.Context context,
...
+
+            if (mCachedObjType != objType)
             {
-                objType = objType.BaseType;
-                if (null != objType)
-                    type = classTable.IndexOf(GetTypeName(objType));
+                mCachedObjType = objType;
+
+                var classTable = runtimeCore.DSExecutable.classTable;
+                mCachedType = classTable.IndexOf(GetTypeName(objType));
```
mjkkirschner: is this implemented with a map - or does look through the entire class table?

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
```diff
@@ -1105,20 +1108,27 @@ private StackValue CreateDSObject(object obj, ProtoCore.Runtime.Context context,
...
+                
+                //Recursively get the base class type if available.
+                while (mCachedType == -1 && objType != null)
+                {
+                    objType = objType.BaseType;
+                    if (null != objType)
+                        mCachedType = classTable.IndexOf(GetTypeName(objType));
```
mjkkirschner: so... why are we doing this, why do we want to get the real base type?

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
```diff
@@ -1105,20 +1108,27 @@ private StackValue CreateDSObject(object obj, ProtoCore.Runtime.Context context,
...
+                }
             }
 
             MetaData metadata;
-            metadata.type = type;
-            StackValue retval = runtimeCore.RuntimeMemory.Heap.AllocatePointer(classTable.ClassNodes[type].Size, metadata);
+            metadata.type = mCachedType;
+            StackValue retval = runtimeCore.RuntimeMemory.Heap.AllocatePointer(0, metadata);
```
mjkkirschner: so now we allocate an empty pointer to the object? Does this pointer actually point anywhere?

src/Engine/ProtoCore/FFI/CLRFFIFunctionPointer.cs
```diff
@@ -860,27 +860,6 @@ public override object Execute(ProtoCore.Runtime.Context c, Interpreter dsi, Lis
                 return null;
             }
 
-            StackValue propValue = (StackValue)retVal;
```
mjkkirschner: so this code was updating reference types values on the heap, and we no longer do it - because it was unused?

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
saintentropy: Eventually we use this pointer StackVakue in the CLRObjectMap and DSObjectMap to retrieve or dispose the return object

src/Engine/ProtoCore/FFI/CLRFFIFunctionPointer.cs
saintentropy: As far as I can tell.  I could not find where we ever retrieved this data as a alternative to un-marshaling from the DSObjectMap

saintentropy: I should say un-marshaling and then calling the getter function via reflection in InvokeNoThow

src/Engine/ProtoCore/DSASM/Heap.cs
saintentropy: True... I can make the change

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
saintentropy: Nope.... is is trying to get an exact match first with Linq.  Next it tries to get a match by splitting the the string also with Linq.

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
saintentropy: I don't know specifically.  It is for un-marshaling the object later though.  This data is put in the the `StackValue` object pointer

saintentropy:
It's probably obvious but this also reduces memory allocation.  Sometimes significantly :-)

src/Engine/ProtoCore/FFI/CLRFFIFunctionPointer.cs
saintentropy: Oops you are correct about reference type vs value type....  I will to update the description.  We added the StackValue pointers not values to the heap

mjkkirschner:
@saintentropy is there an internal task to review this? Who is going to fix the tests etc?

saintentropy:
@mjkkirschner Should I make a new task to review this in the new year?

mjkkirschner:
> @mjkkirschner @QilongTang Should I make a new task to review this in the new year?

nope, we've got it tracked.

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
```diff
@@ -612,6 +612,9 @@ public override object UnMarshal(StackValue dsObject, ProtoCore.Runtime.Context
     class CLRObjectMarshaler : FFIObjectMarshaler
     {
         private static readonly Dictionary<Type, FFIObjectMarshaler> mPrimitiveMarshalers;
+        private static Type mCachedObjType;
```
aparajit-pratap: Minor but I would make these non-static members so that they can immediately be GC'd after the singleton goes out of scope.

aparajit-pratap:
I see the following test failures from DynamoCoreTests run locally. Recording here for lack of a better place. Will return to investigating these:
image

aparajit-pratap:
Self-serve passes.

PR #10029, opened 2019-10-04
harden assembly resolve during deserialization

mjkkirschner:
### Purpose

https://jira.autodesk.com/browse/DYN-2160

This PR hardens graph deserialization when searching for a node type/assembly. Previously assembly load could throw an exception, we would not handle it and a dummy node would not be created.

Now we catch the exception, log it to Dynamo console, and in some cases even recover and create the node correctly if the assembly can be found by the `nodeFactory` when it calls `ResolveType` later in the `nodeReader` 

The main cause of this failure is when some other code attaches an AssemblyResolve handler and then throws from inside that handler - these handlers should not throw - they should either return null or return the assembly, but we cannot control other handlers in any host so we should harden the loading code.

 

### Declarations

### Reviewers

@ColinDayOrg 
@reddyashish

### DIFF
--- a/src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
+++ b/src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
@@ -97,7 +97,15 @@ namespace Dynamo.Graph.Workspaces
             NodeModel node = null;
 
             var obj = JObject.Load(reader);
-            var type = Type.GetType(obj["$type"].Value<string>());
+            Type type = null;
+            try
+            {
+               type = Type.GetType(obj["$type"].Value<string>());
+            }
+            catch(Exception e)
+            {
+                nodeFactory?.AsLogger().Log(e);
+            }
             // If we can't find this type - try to look in our load from assemblies,
             // but only during testing - this is required during testing because some dlls are loaded
             // using Assembly.LoadFrom using the assemblyHelper - which loads dlls into loadFrom context - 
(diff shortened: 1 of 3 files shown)


test/DynamoCoreTests/DummyNodeTests.cs
```diff
@@ -91,6 +91,33 @@ public void WorkspaceIsReadonly_IfXmlDummyNodePresent()
...
+                if (i == -1)
+                {
+                    exceptionCount = exceptionCount + 1;
+                    throw new Exception("TestingTesting");
+                }
+                return null;
+            });
+            //attach our handler that will throw
+            AppDomain.CurrentDomain.AssemblyResolve += handler;
+            System.Threading.Thread.Sleep(5000);
```
mjkkirschner: this is not needed, will remove.

ColinDayOrg:
This PR LGTM (pending the checks above that have not passed yet).

mjkkirschner:
just waiting on the self serve.

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
```diff
@@ -97,7 +97,15 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
...
-            var type = Type.GetType(obj["$type"].Value<string>());
+            Type type = null;
+            try
+            {
+               type = Type.GetType(obj["$type"].Value<string>());
+            }
+            catch(Exception e)
+            {
+                nodeFactory?.AsLogger().Log(e);
+            }
```
angelowang: @mjkkirschner do we want to continue to look into the preloaded nodes for the types if the exception happens?

PR #10051, opened 2019-10-11
DYN-2149 DummyNode unresolved node messages are confusing

reddyashish:
### Purpose

This PR is to improve the warning message shown on the unresolved dummy nodes. 
Task: https://jira.autodesk.com/browse/DYN-2149

The description of the node is shown when the user hovers on the node. When the node is of type 'unresolved', a warning message is shown that specifies the function name. 

Before this change, the message was:

previous

After this change, 

Screen Shot 2019-10-11 at 2 48 16 AM

The assembly name is not stored in the object so instead we show the whole function name(along with the namespace).

### Declarations

### Reviewers
@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCore/Graph/Nodes/DummyNode.cs
+++ b/src/DynamoCore/Graph/Nodes/DummyNode.cs
@@ -45,6 +45,7 @@ namespace Dynamo.Graph.Nodes
             LegacyNodeName = "Dynamo.Graph.Nodes.DummyNode";
             LegacyFullName = LegacyNodeName;
             LegacyAssembly = string.Empty;
+            FunctionName = string.Empty;
             NodeNature = Nature.Unresolved;
             Description = GetDescription();
             ShouldDisplayPreviewCore = false;
@@ -123,6 +124,52 @@ namespace Dynamo.Graph.Nodes
             LegacyNodeName = legacyName;
             LegacyFullName = legacyName;
             Name = legacyName;
+
+            OriginalNodeContent = originalElement;
+
+            LegacyAssembly = legacyAssembly;
+            NodeNature = DummyNode.Nature.Unresolved;
+
+            Description = GetDescription();
+            ShouldDisplayPreviewCore = false;
+
+            if (originalElement != null)
+            {
+                var legacyFullName = originalElement["FunctionSignature"];
+                if (legacyFullName != null)
+                    LegacyFullName = legacyFullName.ToString();
+            }
+
+            UpdatePorts();
+        }
+
+        /// <summary>
+        /// This function creates DummyNode with specified number of ports.
+        /// </summary>
+        /// <param name="id">Id of the original node</param>
+        /// <param name="inputCount">Number of input ports</param>
+        /// <param name="outputCount">Number of output ports</param>
+        /// <param name="legacyAssembly">Assembly of the node</param>
+        /// <param name="functionName">Function name of the node</param>
+        /// <param name="originalElement">Original JSON description of the node</param>
+        public DummyNode(
+           string id,
+           int inputCount,
+           int outputCount,
+           string legacyAssembly,
+           string functionName,
+           JObject originalElement)
+        {
+            GUID = new Guid(id);
+
+            InputCount = inputCount;
+            OutputCount = outputCount;
+
+            string legacyName = "Unresolved";
+            LegacyNodeName = legacyName;
+            LegacyFullName = legacyName;
+            Name = legacyName;
+            FunctionName = functionName;
 
             OriginalNodeContent = originalElement;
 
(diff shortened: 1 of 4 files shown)


src/DynamoCore/Graph/Nodes/DummyNode.cs
```diff
@@ -142,6 +143,52 @@ public DummyNode()
...
+           string legacyAssembly,
+           string functionName,
+           JObject originalElement)
+        {
+            GUID = new Guid(id);
+
+            InputCount = inputCount;
+            OutputCount = outputCount;
+
+            string legacyName = "Unresolved";
```
QilongTang: @mjkkirschner @aparajit-pratap I cant remember, do you localize node names?

QilongTang: Just found out that we do not localize node names, so we should be fine here

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
```diff
@@ -96,11 +96,22 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
         {
             NodeModel node = null;
 
+            String typeName = null;
+            String functionName = null;
```
QilongTang: I would usually recommend to use `String.Empty` because it is less likely to cause crashes. But I saw you have wrapped string null check lots of places which is also good.

mjkkirschner:
@reddyashish - I think something we neglected to consider or discuss yesterday was DummyNodes which represent NodeModel nodes - for example, a dropdown node - here the type and assembly information can actually be gathered from the `$type` correctly - I think we need to handle both these cases, and this means we need different constructors for dummyNodes - one where we use the function name gathered from `FunctionSignature` - this will only exist for zero touch nodes - and another constructor and message where the node Type and assembly are actually known.

Have you considered that other case?

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
```diff
@@ -107,11 +108,16 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
                type = Type.GetType(obj["$type"].Value<string>());
                typeName = obj["$type"].Value<string>().Split(',').FirstOrDefault();
 
-               if (typeName.Contains("ZeroTouch"))
-               {
-                    // This assemblyName does not usually contain version information...
+                if (typeName.Contains("ZeroTouch"))
```
mjkkirschner: can we be specific about this-  it's possible anyone could make a nodeModel and name it `ZeroTouch` lets use our entire namespace and type name - including the assembly name would be fine as well for precision.

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
```diff
@@ -107,11 +108,16 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
...
+                {
+                    // If it is a zero touch node, then get the whole function name including the namespace.
                     functionName = obj["FunctionSignature"].Value<string>().Split('@').FirstOrDefault().Trim();
```
mjkkirschner: do FunctionSignatures always have the `@` token? What if it's not there, what happens?

reddyashish:
Added the typename info for the node model node. Now the message for node model node looks like this:
Screen Shot 2019-10-15 at 12 04 36 AM

@mjkkirschner If the function signature has not parameters(no '@'), it just takes the name. Have tested this case.

reddyashish:
Also have checked for the complete namespace for xero-touch nodes. The assembly name is stored as DynamoCore in the type string. So was just checking for the namespace.

src/DynamoCore/Graph/Nodes/DummyNode.cs
```diff
@@ -370,6 +466,16 @@ internal string GetDescription()
         /// </summary>
         public string LegacyAssembly { get; private set; }
 
+        /// <summary>
+        /// Type name of the node.
+        /// </summary>
+        public string TypeName { get; private set; }
```
mjkkirschner: any reason not to overload the LegacyNodeName - maybe note that this is property is only valid for ZT dummy nodes?

src/DynamoCore/Graph/Nodes/DummyNode.cs
```diff
@@ -370,6 +466,16 @@ internal string GetDescription()
...
+
+        /// <summary>
+        /// Returns the node's function name
+        /// </summary>
+        public string FunctionName { get; private set; }
```
mjkkirschner: any reason not to overload the LegacyNodeName - maybe note that this is property is only valid for ZeroTouch dummy nodes?

src/DynamoCore/Graph/Nodes/DummyNode.cs
```diff
@@ -320,29 +416,29 @@ internal string GetDescription()
...
                 {
-                    const string format = "Node of type '{0}' cannot be resolved";
-                    return string.Format(format, LegacyNodeName);
+                    const string format = "Node of type '{0}', from assembly '{1}', cannot be resolved.";
+                    return string.Format(format, TypeName, LegacyAssembly);
                 }
                 else
                 {
-                    const string format = "Node of type '{0}' ({1}) cannot be resolved";
-                    return string.Format(format, LegacyNodeName, LegacyAssembly);
```
mjkkirschner: is `LegacyNodeName` used anywhere anymore? If not maybe you want to obsolete it?

mjkkirschner:
@reddyashish it looks good, thanks for the hard work, just a few last comments.

reddyashish:
Added more comments. The LegacyNodeName and LegacyFullName are being used in some other places so I did remove them and kept it the same.

PR #9664, opened 2019-04-17
"PackageReference" Adoption

QilongTang:
Please Note:
1. Before submitting the PR, please review [How to Contribute to Dynamo](https://github.com/DynamoDS/Dynamo/blob/master/CONTRIBUTING.md)
2. Dynamo Team will meet 1x a month to review PRs found on Github (Issues will be handled separately)
3. PRs will be reviewed from oldest to newest
4. If a reviewed PR requires changes by the owner, the owner of the PR has 30 days to respond. If the PR has seen no activity by the next session, it will be either closed by the team or depending on its utility will be taken over by someone on the team
5. PRs should use either Dynamo's default PR template or [one of these other template options](https://github.com/DynamoDS/Dynamo/wiki/Choosing-a-Pull-Request-Template) in order to be considered for review.
6. PRs that do not have one of the Dynamo PR templates completely filled out with all declarations satisfied will not be reviewed by the Dynamo team.
7. PRs made to the `DynamoRevit` repo will need to be cherry-picked into all the DynamoRevit Release branches that Dynamo supports. Contributors will be responsible for cherry-picking their reviewed commits to the other branches after a `LGTM` label is added to the PR.

### Purpose

This is magic, makes the project config so much easier to read! And verified project still builds fine locally with local cache of all the dependecies. We should totally do this.

Notice there is way to migrate legacy packages.config in Visual Studio 2017 and can be opened correctly in VS2017 only according to https://docs.microsoft.com/en-us/nuget/reference/migrate-packages-config-to-package-reference#limitations.

A team decision we need to make!

image

### Declarations

Check these if you believe they are true

- [x] The code base is in a better state after this PR
- [x] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions), and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.

### Reviewers

@mjkkirschner @scottmitchell 

### FYIs

@DynamoDS/dynamo

PR #10065, opened 2019-10-16
Add system.windows.interactivity nuget reference

aparajit-pratap:
### Purpose

Add `system.windows.interactivity` reference from nuget. This is required since there is a dependency on `EventTrigger` here: https://github.com/DynamoDS/Dynamo/blob/52557ec8dbd729d658dca8df69fbe5c721c2ceae/src/DynamoCoreWpf/UI/HandlingEventTrigger.cs#L5

### Declarations

### FYIs
@DynamoDS/dynamo


mjkkirschner:
what is the difference between interactivity.eventTrigger and https://docs.microsoft.com/en-us/dotnet/api/system.windows.eventtrigger?view=netframework-4.8

mjkkirschner:
https://docs.microsoft.com/en-us/previous-versions/visualstudio/design-tools/expression-studio-4/ff726539(v%3Dexpression.40)

aparajit-pratap:
> https://docs.microsoft.com/en-us/previous-versions/visualstudio/design-tools/expression-studio-4/ff726539(v%3Dexpression.40)

Not sure. They're apparently used for the list@level UI on the input ports. Tried replacing with `System.Windows.EventTrigger` and was crashing.

QilongTang:
Thank you and couple question: 
1. Do we need to align with the version in Revit?
2. Do we want to switch to use project reference which have better support starting from VS 2019

mjkkirschner:
maybe we can figure out how to get rid of this!?

aparajit-pratap:
@QilongTang could we have you share with the team what you know about project references in VS2019?

QilongTang:
@aparajit-pratap Sorry I mean package reference. Check this out! https://github.com/DynamoDS/Dynamo/pull/9664/files. This would be essentially similar to Dynamo packages. Nuget packages referenced by different solutions will be downloaded on a central location in userfile instead of each solution folder. And sln file will be cleaner, something to consider :)

PR #10076, opened 2019-10-21
try adding more fuzz time based on test runtime

mjkkirschner:
### Purpose
This PR adds a fuzz factor the performance tests based on their mean time in the baseline data - smaller runtimes translate to larger fuzz times. This is to handle the case where very small fluctuations in actual time (ms) cause large percentage changes, and the tests fail.

### Declarations

### Reviewers

@smangarole 
@reddyashish

### DIFF
--- a/tools/Performance/DynamoPerformanceTests/ResultsComparer.cs
+++ b/tools/Performance/DynamoPerformanceTests/ResultsComparer.cs
@@ -30,8 +30,15 @@ namespace DynamoPerformanceTests
 
         internal static ComparisonResultState DefaultBenchmarkComparison(BenchmarkComparison comparisonData)
         {
-            //if the change in mean time
-            if (comparisonData.MeanDelta > 10)
+            //assumes times are in ms
+            //if the mean delta is greater than 10 + (scaled and inverted baseMean)
+            //then fail.
+            //This adds a larger fuzz factor for those tests with small run times.
+            //EX. test with runtime of 100ms will yeild fuzz factor of 10 + 1000/100 = 20%
+            //EX. test with runtime of 1000ms will yeild fuzz factor of 10 + 1000/1000 = 11%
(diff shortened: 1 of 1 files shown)


tools/Performance/DynamoPerformanceTests/ResultsComparer.cs
```diff
@@ -30,8 +30,15 @@ internal enum ComparisonResultState
...
-            if (comparisonData.MeanDelta > 10)
+            //assumes times are in ms
+            //if the mean delta is greater than 10 + (scaled and inverted baseMean)
+            //then fail.
+            //This adds a larger fuzz factor for those tests with small run times.
+            //EX. test with runtime of 100ms will yeild fuzz factor of 10 + 1000/100 = 20%
+            //EX. test with runtime of 1000ms will yeild fuzz factor of 10 + 1000/1000 = 11%
+            //This attempts to deal with test failures that have very small runtimes which are 
+            //more susceptible to uncontrollable operating system or machine workloads.
+            if (comparisonData.MeanDelta > (10 + (1000/comparisonData.BaseResult.Mean)))
```
QilongTang: Is there ways to increase times of test to reduce noise or any other ways to your point of view? I am worried this may hide issues for those tests takes less time..

mjkkirschner: not without making the graphs more complex.

reddyashish:
@mjkkirschner Looks good to me. The more the base result mean value, a smaller mean delta value will trigger a failure.

tools/Performance/DynamoPerformanceTests/ResultsComparer.cs
QilongTang: hmm, this reminds me that Revit performance farm runs the failed tests one more round. Maybe we can tune it later

PR #10080, opened 2019-10-22
Upgrade to latest Helix and simplify shader pipeline (WIP/Proposal)

Dewb:
### Purpose

This PR is a discussion point for a proposed more-aggressive strategy for Helix replacement. In addition to upgrading from Helix 2015.1 to Helix 2.9.0, it also pulls out the private Dynamo copy of the Helix shader pipeline in favor of using default shaders. 

The arguments for doing this include:
- Reducing our maintenance surface
- Taking maximum advantage of improvements in Helix
- Making it easier to implement performance enhancements and other new features in the future without requiring intimate knowledge of Dynamo's private reimplementation of the Helix pipeline.

The arguments against doing this include:
- This may be too aggressive, breaking view extensions that build their own vertex buffers
- Non-shader reimplementations of things like headlamp lighting are likely to have subtle graphical differences from the previous pipeline. These differences may be acceptable, or they might not.

There were a number of features that were provided directly in the shader that need to be reimplemented, either in the C# realm, or by re-adding minimal extra shader definitions & input buffer formats. (It may turn out that we must leverage some custom HLSL to maximize performance, but we should not continue editing the default shader.) 

- [X] Headlight
- [ ] Drawing selected geometry in blue
- [ ] Drawing frozen geometry at half alpha
- [ ] Drawing only isolated geometry in isolation mode
- [ ] Vertex colors and custom materials
- [ ] Instancing?
- [ ] What else?

### Declarations


Dewb:
FYI @LongNguyenP (can't add you as a reviewer, for some reason.)

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
```diff
@@ -36,13 +36,7 @@ private static void ShowSelectedPropertyChanged(DependencyObject obj, Dependency
...
             {
                 var geom = (GeometryModel3D)obj;
-                
-                if (geom.IsAttached)
-                {
-                    var host = geom.RenderHost;
-                    geom.Detach();
-                    geom.Attach(host);
-                }
+                // TODO DYN-973: Need new mechanism to trigger render update after selected/frozen/isolated properties change
```
Dewb: With Detach/Attach no longer possible (nor desirable, for performance reasons!) it's not clear what the most idiomatic way to force an update on these scene objects would be in our environment. In raw Helix-WPF, the idiomatic thing to do would be to use WPF dependency property binding to bind the material to the selected/frozen properties with a ValueConverter, but that doesn't look like the best option here.

src/DynamoCoreWpf/DynamoCoreWpf.csproj
```diff
@@ -86,34 +80,30 @@
...
     </Reference>
-    <Reference Include="SharpDX.Direct3D9, Version=2.6.3.0, Culture=neutral, PublicKeyToken=b4dcf0f35e5521f1, processorArchitecture=MSIL">
-      <HintPath>$(SharpDXPackageBinDir)\SharpDX.Direct3D9.dll</HintPath>
+    <Reference Include="SharpDX.Direct2D1, Version=4.2.0.0, Culture=neutral, PublicKeyToken=b4dcf0f35e5521f1, processorArchitecture=MSIL" />
+    <Reference Include="SharpDX.Direct3D11, Version=4.2.0.0, Culture=neutral, PublicKeyToken=b4dcf0f35e5521f1, processorArchitecture=MSIL">
+      <HintPath>..\packages\SharpDX.Direct3D11.4.2.0\lib\net45\SharpDX.Direct3D11.dll</HintPath>
     </Reference>
-    <Reference Include="SharpDX.DXGI, Version=2.6.3.0, Culture=neutral, PublicKeyToken=b4dcf0f35e5521f1, processorArchitecture=MSIL">
-      <HintPath>$(SharpDXPackageBinDir)\SharpDX.DXGI.dll</HintPath>
+    <Reference Include="SharpDX.Direct3D9, Version=4.2.0.0, Culture=neutral, PublicKeyToken=b4dcf0f35e5521f1, processorArchitecture=MSIL">
```
mjkkirschner: @Dewb seems there is still a compile time reference to Direct3D9 - any idea if it's needed?

src/DynamoCoreWpf/Properties/Resources.resx
```diff
@@ -1992,9 +1992,6 @@ Do you want to install the latest Dynamo update?</value>
     <value>You do not have write permission to {0}.</value>
   </data>
   <assembly alias="System.Windows.Forms" name="System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" />
-  <data name="_dynamo" type="System.Resources.ResXFileRef, System.Windows.Forms">
-    <value>..\viewmodels\watch3d\resources\_dynamo.bfx;System.Byte[], mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
```
mjkkirschner: @Dewb these are compiled shader byte code files?

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
```diff
@@ -21,12 +21,12 @@ public static class AttachedProperties
             typeof(GeometryModel3D),
             new PropertyMetadata(false, ShowSelectedPropertyChanged));
 
-        public static void SetShowSelected(UIElement element, bool value)
+        public static void SetShowSelected(DependencyObject element, bool value)
```
mjkkirschner: even though this references a less derived base type - this is still a binary break, and a missing method exception will be thrown for anything that called this previously and is not recompiled AFAIK, I don't know enough to say if it could even cause reference errors for users who reference DyanmoCoreWPF.

I think the most conservative pattern and the one we generally follow - is to avoid changes like this - obsolete the methods - and add a series of new ones - the implementations can be moved to the new methods, and the old methods can be wrappers.

If that is too much work, then extensive testing is required to see in (if) which situations this change causes runtime failures.

src/DynamoCoreWpf/ViewModels/Watch3D/DynamoEffectsManager.cs
```diff
@@ -15,74 +16,48 @@ namespace Dynamo.Wpf.ViewModels.Watch3D
     /// </summary>
     public class DynamoEffectsManager : DefaultEffectsManager
     {
-        public DynamoEffectsManager(IRenderTechniquesManager renderTechniquesManager) : base(renderTechniquesManager) { }
+        public DynamoEffectsManager() : base() {
```
mjkkirschner: another binary break - one option is to just keep this signature and ignore the use of the parameter, but I think obsoleting the method is cleaner.

mjkkirschner:
@Dewb - thanks for the hard work - I need to keep reviewing but wanted to comment on the API breaks -  I'm going to stop pointing out binary API breaks at each line as - there are many, obviously. 

I guess we can consider the work of keeping the existing APIs around (and obsoleting them) separate from the switch to this new helix API - but I think an alternative would be to keep as many of these old APIs as possible working and to introduce completely new classes which introduce the new API methods. IE Helix2ViewModel or something like that...

mjkkirschner:
in addition - it seems there are some changes to our APIs which exposed helix2015 types directly, which will simply no longer exist - like `RenderTechnique`? In that case I guess we have no choice but to wait until a major release of Dynamo (3.0) or decide to break that API?

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
Dewb: These methods are only ever called with Helix objects and the base class of those Helix objects has changed. I think these changes have to be made in a breaking fashion/major release.

Dewb:
Yes, I should have said up front that because we expose Helix types in our API and those Helix types have changed, I don’t think there’s a way to do this at all without breaking binary compatibility.

src/DynamoCoreWpf/DynamoCoreWpf.csproj
Dewb: I think this is still in the required NuGet dependency tree, will confirm.

src/DynamoCoreWpf/Properties/Resources.resx
Dewb: Correct.

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
mjkkirschner: @Dewb - doing my own little investigation, seems that Detach is still possible - we need to override `OnDetach` now and call base.Detach from there for some reason.

see:
https://github.com/helix-toolkit/helix-toolkit/pull/404

mjkkirschner: responded to this comment below, not sure why that comment about detach is not inline - BUT
in regards to freeze, select etc. Why don't we just use different materials?

src/DynamoCoreWpf/ViewModels/Watch3D/DynamoGeometryModel3D.cs
```diff
@@ -58,258 +40,5 @@ public bool RequiresPerVertexColoration
...
-
-        public override int VertexSizeInBytes
-        {
-            get
-            {
-                return DynamoMeshVertex.SizeInBytes;
-            }
-        }
-
-        protected override void OnRasterStateChanged(int depthBias)
```
mjkkirschner: this also still exists, but no longer has a depthBias parameter.

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
Dewb: The `OnDetach` methods are overridable in base classes, but the Detach methods that trigger them have moved from their original publicly accessible location. Still investigating what the intended use it here.

Yes, swapping materials would be an example of what I meant by reimplementing features like freeze/selected appearances outside the shader pipeline.

LongNguyenP:
Can you try adding me again? user names is LongNguyenP

QilongTang:
@LongNguyenP Assigned

src/DynamoCoreWpf/ViewModels/Watch3D/DynamoGeometryModel3D.cs
Dewb: It is not present at all in 2.9.

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
LongNguyenP: Here are my conclusions after some investigation

1. As we previously discussed Attach and Detach are no longer available publicly. They are now invoked internally whenever a model is being added or removed from Viewport3DX.Items
2. Hence, if you want a model to not be showed, it needed to be removed from Viewport3DX.Items
3. If a model is already been added to Viewport3DX.Items. You can still modify its properties at run-time
4. The only slight exception to 3 is the Geometry property. If you modify vertex, indices, colors you must explicitly call Geometry.UpdateVertices() or UpdateTriangles() etc... afterwards
5. I think the simplest way to highlight (or freeze) selected geometries is to modify the Color property of the model (The default shader multiplies the model's Color with each vertex color to get the final rendered color)

QilongTang:
@Dewb @LongNguyenP @mjkkirschner @aparajit-pratap @saintentropy Given that Helix upgrade is already live on master branch, closing this PR. If you found any issue, please report using Github issue or Jira

PR #10083, opened 2019-10-24
DYN-2226: Graphs with large node quantities cause long delays when setting element bound Revit nodes to update

saintentropy:
### Purpose

The purpose of this PR is to address a slow down observed in Dynamo graphs with large node quantities when Dynamo4Revit is responding to Document Change events in Revit.  The specific flaw begins when the event handler [`RevitServicesUpdater_ElementsUpdated()`](https://github.com/DynamoDS/DynamoRevit/blob/4cc183fb6eb7db13e29cf2d04192df499961a8b9/src/DynamoRevit/Models/RevitDynamoModel.cs#L871) in D4R responds to the Document Change events.  It filters for transactions originating from Dynamo then passes the associated Element Id's to [`GetNodesFromElementIds()`](https://github.com/DynamoDS/DynamoRevit/blob/4cc183fb6eb7db13e29cf2d04192df499961a8b9/src/Libraries/RevitServices/Persistence/ElementBinder.cs#L464) in the `ElementBinder`.  This method calls `GetCallsitesForNodes()` in DynamoCore to find the associated Callsites for the nodes in the Graph.   This lookup can be **extremely** slow due to a series of nested loops.  This PR address the slowdown by processing the lookup with a pre-populated dictionary mapping the node Guids to Callsite Identifiers as the existing Callsite Cache dictionary.  The runtime difference for our sample graph with 456 nodes was 23sec to ~20ms.  

### Declarations

### Reviewers

@mjkkirschner assign to whoever you want

### FYIs

@Dewb @QilongTang

### DIFF
--- a/src/Engine/ProtoCore/RuntimeData.cs
+++ b/src/Engine/ProtoCore/RuntimeData.cs
@@ -49,7 +49,7 @@ namespace ProtoCore
         /// <summary>		
         /// Map from a graph UI node to callsite identifiers. 		
         /// </summary>
-        public Dictionary<Guid, List<CallSite>> NodeToCallsiteObjectMap { get; }
+        internal Dictionary<Guid, List<CallSite>> NodeToCallsiteObjectMap { get; }
         
 #endregion
 
@@ -60,14 +60,11 @@ namespace ProtoCore
             NodeToCallsiteObjectMap = new Dictionary<Guid, List<CallSite>>();
         }
 
-      
 
         /// <summary>
         /// Retrieves an existing instance of a callsite associated with a UID
         /// It creates a new callsite if non was found
         /// </summary>
-        /// <param name="core"></param>
-        /// <param name="uid"></param>
         /// <returns></returns>
         public CallSite GetCallSite(int classScope, string methodName, Executable executable, RuntimeCore runtimeCore)
         {
--- a/src/Engine/ProtoCore/RuntimeData.cs
+++ b/src/Engine/ProtoCore/RuntimeData.cs
@@ -125,7 +125,8 @@ namespace ProtoCore
 
                 CallsiteCache[callsiteID] = csInstance;
                 CallSiteToNodeMap[csInstance.CallSiteID] = topGraphNode.guid;
-                if (NodeToCallsiteIdentifiersMap.TryGetValue(topGraphNode.guid, out var callsiteIDs))
+                List<string> callsiteIDs;
+                if (NodeToCallsiteIdentifiersMap.TryGetValue(topGraphNode.guid, out callsiteIDs))
                 {
                     callsiteIDs.Add(callsiteID);
                 }
@@ -177,14 +178,16 @@ namespace ProtoCore
                     return nodeMap;
             }
 
+            List<string> callsiteIDs;
             foreach (Guid nodeGuid in nodeGuids)
             {
-                if (NodeToCallsiteIdentifiersMap.TryGetValue(nodeGuid, out var callsiteIDs))
+                if (NodeToCallsiteIdentifiersMap.TryGetValue(nodeGuid, out callsiteIDs))
                 {
                     var matchingCallSites = new List<CallSite>();
                     foreach (var id in callsiteIDs)
                     {
-                        if (CallsiteCache.TryGetValue(id, out var callsite))
+                        CallSite callsite;
+                        if (CallsiteCache.TryGetValue(id, out callsite))
                             matchingCallSites.Add(callsite);
                     }
                     nodeMap[nodeGuid] = matchingCallSites;
--- a/src/Engine/ProtoCore/RuntimeData.cs
+++ b/src/Engine/ProtoCore/RuntimeData.cs
@@ -125,6 +125,14 @@ namespace ProtoCore
 
                 CallsiteCache[callsiteID] = csInstance;
                 CallSiteToNodeMap[csInstance.CallSiteID] = topGraphNode.guid;
+                if (NodeToCallsiteIdentifiersMap.TryGetValue(topGraphNode.guid, out var callsiteIDs))
+                {
+                    callsiteIDs.Add(callsiteID);
+                }
+                else
+                {
+                    NodeToCallsiteIdentifiersMap[topGraphNode.guid] = new List<string>() { callsiteID };
+                }
             }
 
             if (graphNode != null && !CoreUtils.IsDisposeMethod(methodName))
--- a/src/Engine/ProtoCore/RuntimeData.cs
+++ b/src/Engine/ProtoCore/RuntimeData.cs
@@ -46,13 +46,18 @@ namespace ProtoCore
         /// Map from a callsite's guid to a graph UI node. 		
         /// </summary>
         public Dictionary<Guid, Guid> CallSiteToNodeMap { get; }
- 		 
- #endregion
+        /// <summary>		
+        /// Map from a graph UI node to callsite identifiers. 		
+        /// </summary>
+        public Dictionary<Guid, List<string>> NodeToCallsiteIdentifiersMap { get; }
+        
+#endregion
 
         public RuntimeData()
         {
             CallsiteCache = new Dictionary<string, CallSite>();
             CallSiteToNodeMap = new Dictionary<Guid, Guid>();
+            NodeToCallsiteIdentifiersMap = new Dictionary<Guid, List<string>>();
         }
 
       
--- a/src/Engine/ProtoCore/RuntimeData.cs
+++ b/src/Engine/ProtoCore/RuntimeData.cs
@@ -145,8 +145,15 @@ namespace ProtoCore
             return csInstance;
         }
 
-        public Dictionary<Guid, List<CallSite>>
-        GetCallsitesForNodes(IEnumerable<Guid> nodeGuids, Executable executable)
+        /// <summary>
+        /// This API is used by host integrations such as for Revit and C3D.
+        /// It is used to gets the trace data list for all nodes binding to elements in the host.
+        /// </summary>
+        /// <param name="nodeGuids"></param>
+        /// <param name="executable"></param>
+        /// <returns></returns>
+        public Dictionary<Guid, List<CallSite>> GetCallsitesForNodes(
+            IEnumerable<Guid> nodeGuids, Executable executable)
         {
             if (nodeGuids == null)
                 throw new ArgumentNullException("nodeGuids");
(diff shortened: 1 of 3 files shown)


src/Engine/ProtoCore/RuntimeData.cs
```diff
@@ -166,20 +179,20 @@ public CallSite GetCallSite(int classScope, string methodName, Executable execut
...
-                    continue;
-
-                // Get all callsites that match the graph node ids.
-                var matchingCallSites = (from cs in CallsiteCache
-                                         from gn in matchingGraphNodes
-                                         where string.Equals(cs.Key, gn.CallsiteIdentifier)
-                                         select cs.Value);
-
-                // Append each callsite element under node element.
-                nodeMap[nodeGuid] = matchingCallSites.ToList();
```
saintentropy: I am still working on the big O notation for this but it is very bad

src/Engine/ProtoCore/RuntimeData.cs
```diff
@@ -166,20 +179,20 @@ public CallSite GetCallSite(int classScope, string methodName, Executable execut
...
+                        if (CallsiteCache.TryGetValue(id, out var callsite))
+                            matchingCallSites.Add(callsite);
+                    }
+                    nodeMap[nodeGuid] = matchingCallSites;
+                }
+                else
+                {
+                    nodeMap[nodeGuid] = new List<CallSite>();
+                }
             }
```
saintentropy: I am still working on the O notation for this but it is bad

src/Engine/ProtoCore/RuntimeData.cs
saintentropy: Not sure what the O notation for the existing loop up was but it was bad

saintentropy:
As a side note, looking at the original call in D4R [`GetNodesFromElementIds()`](https://github.com/DynamoDS/DynamoRevit/blob/4cc183fb6eb7db13e29cf2d04192df499961a8b9/src/Libraries/RevitServices/Persistence/ElementBinder.cs#L464) method, it seems like it could just use the CallsiteCache directly instead of calling `GetCallsitesForNodes()` first.  I need to check what node types it filters but may be faster just to process all the Callsites.

src/Engine/ProtoCore/RuntimeData.cs
```diff
@@ -46,13 +46,18 @@ public class RuntimeData
         /// Map from a callsite's guid to a graph UI node. 		
         /// </summary>
         public Dictionary<Guid, Guid> CallSiteToNodeMap { get; }
- 		 
- #endregion
+        /// <summary>		
+        /// Map from a graph UI node to callsite identifiers. 		
+        /// </summary>
+        public Dictionary<Guid, List<string>> NodeToCallsiteIdentifiersMap { get; }
```
QilongTang: Seems `internal` now, can we expose this later?

QilongTang:
Can you also look at why the AppVeyor build fails?

src/Engine/ProtoCore/RuntimeData.cs
saintentropy: Access modifier can be set more conservative but I would argue the other runtime dictionaries should be as well.

saintentropy: I purposefully did not use `Dictionary<Guid, List<Guid>>` as the primary lookup key for the `CallsiteCache` dictionary was via the Callsite Identifier string.  Using the `Callsite.CallsiteID` Guid would require another series of ForEach loops or the creation of another map to the `Callsite` objects which did not seem advisable.   `CallsiteCache` is the critical map here and is used in many places.  The `NodeToCallsite` is only used in the `ReconcileTraceDataAndNotify()` which we should also look at core it's inefficiency.

saintentropy:
AppVeyor fixed :-)  When can we use C# 7 features @QilongTang?

QilongTang:
@saintentropy When we move everything to Visual Studio 2019, currently only me in the team using that so I got the same pain :(

src/Engine/ProtoCore/RuntimeData.cs
mjkkirschner: so looking at this - it appears you are completely skipping the node to graphNode lookup?

mjkkirschner:
@saintentropy it is likely a good idea to run the RTF tests over this change after it is merged - perhaps @AndyDu1985 or @ZiyunShang can help there?

mjkkirschner:
I'm also curious if the elementBinding perf tests get faster?

saintentropy:
@mjkkirschner I don't think the ElementBindingLarge test is testing Revit Element binding though its more like testing the code path I updated.  ElementBindingLarge is more about the trace mechanism.  We should run it on the pre/post thread local storage changes.

src/Engine/ProtoCore/RuntimeData.cs
```diff
@@ -120,6 +125,15 @@ public CallSite GetCallSite(int classScope, string methodName, Executable execut
...
                 CallsiteCache[callsiteID] = csInstance;
                 CallSiteToNodeMap[csInstance.CallSiteID] = topGraphNode.guid;
+                List<string> callsiteIDs;
+                if (NodeToCallsiteIdentifiersMap.TryGetValue(topGraphNode.guid, out callsiteIDs))
+                {
+                    callsiteIDs.Add(callsiteID);
+                }
+                else
+                {
+                    NodeToCallsiteIdentifiersMap[topGraphNode.guid] = new List<string>() { callsiteID };
```
aparajit-pratap: Why can't we simply declare a `NodeToCallsiteMap` as:
```
Dictionary<Guid, List<CallSite>> NodeToCallsiteMap;
```
Then we can simply say:
```
NodeToCallsiteMap[topGraphNode.guid] = new List<CallSite>() { csInstance };
```

src/Engine/ProtoCore/RuntimeData.cs
```diff
@@ -164,22 +178,24 @@ public CallSite GetCallSite(int classScope, string methodName, Executable execut
...
+                        CallSite callsite;
+                        if (CallsiteCache.TryGetValue(id, out callsite))
+                            matchingCallSites.Add(callsite);
+                    }
+                    nodeMap[nodeGuid] = matchingCallSites;
+                }
+                else
+                {
+                    nodeMap[nodeGuid] = new List<CallSite>();
+                }
```
aparajit-pratap: Then if we have the `NodeToCallsiteMap`, we won't need to do any of this either; we could simply return that map; or am I missing something?

aparajit-pratap:
@saintentropy is this specific slowdown you're observing in DYN-2226 due to document-change events occurring in Revit caused by specific Revit UI interactions (Revit transactions) or caused by D4R nodes that cause the Revit document to change?

test/DynamoCoreWpfTests/RecordedTests.cs
```diff
@@ -2770,6 +2770,63 @@ public void TestCallsiteMapModifyInputConnection()
...
 
+        [Test, RequiresSTA]
+        public void TestNodeToCallsiteIdentifierMapModifyInputConnection()
+        {
+            string callsiteIdentifierFirstCall = "";
+            string callsiteIdentifierSecondCall = "";
+            Guid functionCallNodeGuid = new Guid("16e960e5-8a24-44e7-ac81-3759aaf11d25");
+
+            preloadGeometry = true;
+            RunCommandsFromFile("TestCallsiteMapModifyModifyInputConnection.xml", (commandTag) =>
```
aparajit-pratap: I think you'd also need to submit the XML file.

saintentropy:
@aparajit-pratap I haven't tested but could be due to both in the graph was in automatic mode.  Our specific case was due to a graph running in D4R

src/Engine/ProtoCore/RuntimeData.cs
saintentropy: Yes we can do that instead.... I wasn't sure the implications of having two maps that have Callsite Values.  Seems ok since the CallsiteCache is never modified other than `add`

src/Engine/ProtoCore/RuntimeData.cs
saintentropy: Yes... this function still might need to filter the map based on the incoming dynamo graph node Guids but it could be simpler.  Back in the D4R they could simply use the map

mjkkirschner:
@saintentropy I actually believe I checked with Nate that the graph was in manual mode. He should have the full stack trace from the profiling session.

saintentropy:
@mjkkirschner I mean you can also see this action cause by Revit UI interaction if D4R is open and is in automatic mode.  Our Repo case was a single run of the graph caused by D4R

src/Engine/ProtoCore/RuntimeData.cs
```diff
@@ -145,41 +159,17 @@ public CallSite GetCallSite(int classScope, string methodName, Executable execut
...
             {
-                if (executable != null)
+                if (NodeToCallsiteObjectMap.TryGetValue(nodeGuid, out callsites))
                 {
-                    var stream = executable.instrStreamList;
-                    if (stream != null && (stream.Length > 0))
-                    {
-                        var graph = stream[0].dependencyGraph;
-                        if (graph != null)
-                            graphNodes = graph.GraphList;
```
mjkkirschner: @aparajit-pratap can you give us a summary of what a `GraphNode` is in the context of the VM - and then is there any reason not to rename this to something less confusing as soon as possible?

src/Engine/ProtoCore/RuntimeData.cs
```diff
@@ -120,6 +125,15 @@ public CallSite GetCallSite(int classScope, string methodName, Executable execut
 
                 CallsiteCache[callsiteID] = csInstance;
                 CallSiteToNodeMap[csInstance.CallSiteID] = topGraphNode.guid;
+                List<CallSite> callsites;
```
mjkkirschner: @saintentropy did you notice that the params and comment for this method are completely wrong?

saintentropy: @mjkkirschner  Which method are you meaning?  `GetCallsite`

mjkkirschner: yes

src/Engine/ProtoCore/RuntimeData.cs
aparajit-pratap: 1. During codegen phase the compiler generates `graphnode`s that represent each executable statement
2. `Graphnode`s contain information about dependencies between statements​ and runtime properties instructing VM which instruction to execute next​
3. They have a property that points to start and end of instruction block​
4. Every time a graph node executes it jumps to its target label as per dependency property stored​ in it
5. The “dep” instruction is responsible for finding which graph node to execute next given what
was currently modified

PR #10096, opened 2019-10-29
Convert BigInteger to Int64 when possible from Python

aparajit-pratap:
### Purpose

https://autodesk.slack.com/archives/CED3BF6C9/p1571942488031200
image

JIRA: https://jira.autodesk.com/browse/DYN-2262

### Declarations

### Reviewers

@DynamoDS/dynamo 

### FYIs

@Dewb

### DIFF
--- a/src/DynamoUtilities/DataMarshaler.cs
+++ b/src/DynamoUtilities/DataMarshaler.cs
@@ -2,6 +2,7 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Numerics;
 
 namespace Dynamo.Utilities
 {
@@ -86,7 +87,20 @@ namespace Dynamo.Utilities
             var applicable = marshalers.Where(pair => pair.Key.IsAssignableFrom(targetType));
 
             if (!applicable.Any())
-                return obj;
+            {
+                if (!(obj is BigInteger bigInt)) return obj;
+
+                long int64;
+                try
+                {
+                    int64 = (long)bigInt;
+                }
+                catch (OverflowException)
+                {
+                    return bigInt;
+                }
+                return int64;
+            }
 
             // Find the marshaler that operates on the closest base type of the target type.
             var dispatchedMarshaler =
--- a/src/DynamoUtilities/DataMarshaler.cs
+++ b/src/DynamoUtilities/DataMarshaler.cs
@@ -26,6 +26,20 @@ namespace Dynamo.Utilities
                     // Dictionary<TKey, TValue> and IronPython.Runtime.PythonDictionary both implement IDictionary
                     return dict.Keys.Cast<object>().ToDictionary(key => Marshal(key), key => Marshal(dict[key]));
                 });
+            RegisterMarshaler(
+                (BigInteger bigInt) =>
+                {
+                    long int64;
+                    try
+                    {
+                        int64 = (long)bigInt;
+                    }
+                    catch (OverflowException)
+                    {
+                        return bigInt;
+                    }
+                    return int64;
+                });
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


src/DynamoUtilities/DataMarshaler.cs
```diff
@@ -86,7 +87,20 @@ public object Marshal(object obj)
             var applicable = marshalers.Where(pair => pair.Key.IsAssignableFrom(targetType));
 
             if (!applicable.Any())
-                return obj;
+            {
+                if (!(obj is BigInteger bigInt)) return obj;
```
aparajit-pratap: C# 7.0 syntax seems to be breaking the engops build.

Dewb:
Excellent! 

Any reason this has to be done inside the generic marshaller execution rather than by defining a custom marshaler for `BigInteger`?

Dewb:
Looks like we have custom `DataMarshaler`s for Python node inputs, but not outputs. This new code could definitely go here: https://github.com/DynamoDS/Dynamo/blob/master/src/Libraries/DSIronPython/IronPythonEvaluator.cs

The `DataMarshaler` class isn't used anywhere aside from the Python node so I would also support moving it out of `DynamoUtilities` and into the Python project, to make its purpose more clear. (Doesn't have to be part of this PR.)

aparajit-pratap:
> Looks like we have custom `DataMarshaler`s for Python node inputs, but not outputs. This new code could definitely go here: https://github.com/DynamoDS/Dynamo/blob/master/src/Libraries/DSIronPython/IronPythonEvaluator.cs

The `DataMarshaler` class is used for both input and output marshaling as seen here: https://github.com/DynamoDS/Dynamo/blob/8fb3b6829c9bcd8c01ee4ef7e880aa8185bf2cd0/src/Libraries/DSIronPython/IronPythonEvaluator.cs#L114 and here: https://github.com/DynamoDS/Dynamo/blob/8fb3b6829c9bcd8c01ee4ef7e880aa8185bf2cd0/src/Libraries/DSIronPython/IronPythonEvaluator.cs#L134, however, I agree that the marshaling for `BigInteger` can be specific to the output marshaling.

mjkkirschner:
@aparajit-pratap would you mind filing this internally?

aparajit-pratap:
Added unit test, addressed review comments about registering the `BigInteger` marshaling logic only for node output (output marshalers are added only in the `DataMarshaler` constructor, while input marshalers are added when first initializing the `IronPythonEvaluator` property). 

I haven't moved `DataMarshaler` to the Python project just yet as @mjkkirschner had some other thoughts on this. Leaving it open for discussion.

src/DynamoUtilities/DynamoUtilities.csproj
```diff
@@ -49,6 +49,7 @@
     <Reference Include="System" />
     <Reference Include="System.configuration" />
     <Reference Include="System.Core" />
+    <Reference Include="System.Numerics" />
```
QilongTang: Do we need this reference?

QilongTang:
LGTM with one comments

src/DynamoUtilities/DynamoUtilities.csproj
aparajit-pratap: Yes, `BigInteger` is defined in `System.Numerics`.

Dewb:
FYI the IronPython issue that prompted this is supposedly fixed in IronPython 2.7.9.

mjkkirschner:
hmm - so this test should have passed already - since we are using ironPython 2.7.9.

@aparajit-pratap do we now include system.numerics in our shipped binaries or this is loaded from the GAC?

It will be useful to know as we should communicate this in the release notes of 2.5

aparajit-pratap:
I reverted my change leaving the test to test locally. It didn't pass?

mjkkirschner:
@aparajit-pratap your test also includes a big integer check (last type in your list) - perhaps that gets marshalled to long now in all cases?

mjkkirschner:
here's Mike's issue:
https://github.com/IronLanguages/ironpython2/issues/681

aparajit-pratap:
> @aparajit-pratap your test also includes a big integer check (last type in your list) - perhaps that gets marshalled to long now in all cases?

This is the output I'm seeing before my change:
image

aparajit-pratap:
> @aparajit-pratap do we now include system.numerics in our shipped binaries or this is loaded from the GAC?

It is loaded from GAC and referenced from `C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.7\`

PR #10127, opened 2019-11-11
DYN-2273 - Fix crash by compiling null input to null AST node

aparajit-pratap:
### Purpose

JIRA: https://jira.autodesk.com/browse/DYN-2273

### Declarations

### Reviewers

@DynamoDS/dynamo


QilongTang:
@aparajit-pratap , my memory fading. Can you describe null node a bit?

aparajit-pratap:
@QilongTang by `nullnode`, I meant `Null` AST node, sorry for not making that clear before. We have a special AST node to represent `null`. In the case where an input AST is `null`, we would rather propagate the `null` than throw an exception either at the time of AST compilation or at runtime, which was what was happening before.

QilongTang:
@aparajit-pratap Thank you, makes sense. LGTM

PR #10129, opened 2019-11-13
[WIP] AGD-1239: Sync filename and workspace name on file open

saintentropy:
### Purpose

The purpose of this PR is to sync the filename with the DYN json `Name` property on file open operations.  Currently this occurs on initial save and saveas operations.  This can get out of sync if the DYN file is renamed in the file system.  

### Declarations

### Reviewers

@QilongTang

### FYIs

@varvaratou @mjkkirschner

### DIFF
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -1686,6 +1686,7 @@ namespace Dynamo.Models
                 CustomNodeManager);
 
             workspace.FileName = filePath;
+            workspace.Name = Path.GetFileNameWithoutExtension(filePath);
             workspace.ScaleFactor = dynamoPreferences.ScaleFactor;
 
             // NOTE: This is to handle the case of opening a JSON file that does not have a version string


mjkkirschner:
I'm worried about this change.
I think this is as designed but I am not sure.

1. for .dyfs the name should not be the same as the filename. You should be able to rename a custom node without changing its location on disk.
2. I believe there are times when the name is null or unset and we use this to determine if the graph as been saved.

mjkkirschner:
Also - conceivably we could add the ability to set the name and description (like we did in AVP) - I know we'd need to make other changes to support that as well of course. - but my point is I don't think our intention was to keep these properties synchronized.

QilongTang:
As I stated on slack, this behavior is as design. 

Enabling the sync could benefit the users who use file name to indicate certain info but could prevent future improvements getting done. An extension to update workspace name and project details could be easily introduced in the future with users' request (see https://forum.dynamobim.com/t/take-a-poll-and-ask-for-what-you-need-on-right-side-panel/42597/10?u=qilong_tang). I believe the Dynamo team had this discussion before and made this decision intentionally. I am open for discussion if there is a better way to accommodate most users' need.

PR #10132, opened 2019-11-15
Improvements and bug fixes to Helix-related stuff

LongNguyenP:
Changes:
1. Each node output port will now create only one HelixRenderPackage. This runs way faster when the output port produces a lot of geometry objects (e.g. 400000 points)
2. Fixed implementation flaws in Attach method of DynamoGeometryModel3D, DynamoLineGeometryModel3D and DynamoPointGeometryModel3D classes, where the both the base and grand-base constructor were called, causing duplicate works. This saves about 100ms in the 400000 points test-case

New:
1. A public static instance of DynamoLogger that can be accessed from anymwhere. This is useful for debugging and diagnosing, (especially in Release build)

Issues:
1. This new strategy of using only one HelixRenderPackage per output node is currently incompatible with geometries that require Custom transformation. Have not found an elegant solution around this.
2. Whoever calls Tesselate() on an IGraphicItem object will have to also immediately call the newly added FillVertexColors method so that the vertex count and vertex color counts are synchronized.

### DIFF
--- a/.idea/.idea.Dynamo.dir/.idea/$PRODUCT_WORKSPACE_FILE$
+++ b/.idea/.idea.Dynamo.dir/.idea/$PRODUCT_WORKSPACE_FILE$
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="masterDetails">
+    <states>
+      <state key="ScopeChooserConfigurable.UI">
+        <settings>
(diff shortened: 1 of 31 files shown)


QilongTang:
@saintentropy @DynamoDS/dynamo Does anyone know the state of this PR?

mjkkirschner:
I think this can be closed as @saintentropy based his latest helix pr on this one.

PR #10139, opened 2019-11-18
bump to 2.6

mjkkirschner:
bump assembly version to 2.6 and commit assemblyinfo.cs
Reviewers:

@aparajit-pratap 
@zeusongit 

@aparajit-pratap @QilongTang - should we update the time base to 2019 yet? I guess this will reset the revision number?


QilongTang:
@mjkkirschner and team, we can update the year but I prefer that if we do, we fix it for both Dynamo, Dynamo Revit and LibG and whichever other project using the same algorithm. Or we can wait until a future release like 3.0 or something since right now the version number will keep working without any problem.

Issue #10047, opened 2019-10-10
Crash report from Dynamo 2.4.0.6186


cromlyngames:
# Issue Description
Please fill in the following information to help us reproduce the issue:

### What did you do?
tried to move a yellowed out box in dynamo sandbox. Yellow error was saying it could not find a file in system

### What did you expect to see?
(Fill in here)

### What did you see instead?
(Fill in here)

### What packages or external references (if any) were used?
dynamo sandbox

### Stack Trace
```
Could not load file or assembly 'file:///C:\WINDOWS\system32\LibG.ProtoInterface.dll' or one of its dependencies. The system cannot find the file specified.

   at Autodesk.DesignScript.Geometry.HostFactory.get_Factory()
   at Autodesk.DesignScript.Geometry.Vector.XAxis()
   at Dynamo.Manipulation.MousePointManipulator.AssignInputNodes()
   at Dynamo.Manipulation.NodeManipulator.GenerateRenderPackages()
   at Dynamo.Manipulation.NodeManipulator.DrawManipulator()
   at Dynamo.Manipulation.NodeManipulator..ctor(NodeModel node, DynamoManipulationExtension manipulatorContext)
   at Dynamo.Manipulation.MousePointManipulatorFactory.Create(NodeModel node, DynamoManipulationExtension manipulatorContext)
   at System.Linq.Enumerable.WhereSelectArrayIterator`2.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Dynamo.Manipulation.ZeroTouchManipulatorFactory.Create(NodeModel node, DynamoManipulationExtension manipulatorContext)
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Dynamo.Manipulation.ManipulatorDaemon.CreateManipulator(NodeModel model, DynamoManipulationExtension manipulatorContext)
   at Dynamo.Manipulation.DynamoManipulationExtension.TryCreateManipulator(NodeModel node)
   at Dynamo.Manipulation.DynamoManipulationExtension.CreateManipulators(IEnumerable`1 nodes)
[...]

PR #10148, opened 2019-11-20
Fix for crash while drawing Point manipulator (gizmo) when LibG fails to load

aparajit-pratap:
### Purpose

https://jira.autodesk.com/browse/DYN-2294
GH Issue: https://github.com/DynamoDS/Dynamo/issues/10047

image
image

### Declarations

### Reviewers

@mjkkirschner


aparajit-pratap:
@mjkkirschner you were right; I needed to account for the case where it could be a list of nulls.

mjkkirschner:
@aparajit-pratap some first thoughts - 
1. what if it's a list of list of nulls or some jagged list of nulls and lists of nulls.
2. How often does this get called? If the input is large this function will take time `o(n)` This could be slow - can we stop the crash somewhere where we do not need to check every item in the list?

aparajit-pratap:
> @aparajit-pratap some first thoughts -
> 
> 1. what if it's a list of list of nulls or some jagged list of nulls and lists of nulls.
> 2. How often does this get called? If the input is large this function will take time `o(n)` This could be slow - can we stop the crash somewhere where we do not need to check every item in the list?

I'm checking for a single `null` and returning `true`, not checking for all null values necessarily.

mjkkirschner:
@aparajit-pratap thanks I see that now that it is recursive - so it will break out when it encounters a null - but let's consider the case where the input is 1 million items (with no nulls) -we'll need to iterate through the entire list - **how frequently does this get called?**

aparajit-pratap:
> @aparajit-pratap thanks I see that now that it is recursive - so it will break out when it encounters a null - but let's consider the case where the input is 1 million items (with no nulls) -we'll need to iterate through the entire list - **how frequently does this get called?**

Cases, where this will occur, are when a `Point.ByCoordinates` node is either using at least one default value as input or is connected to at least one number/integer slider and the node is selected.

mjkkirschner:
will this not also be called anytime the camera is moved?

aparajit-pratap:
> will this not also be called anytime the camera is moved?

I looked at this again and it seems to be called for every mouse move, which is worse. I'm trying to come up with a better solution.

src/DynamoCoreWpf/Rendering/HelixRenderPackage.cs
```diff
@@ -79,6 +79,8 @@ public HelixRenderPackage()
         /// Set the transform that is applied to all geometry in the renderPackage.
         /// </summary>
         /// <param name="transform"></param>
+        // TODO: Unused method. Adds unnecessary dependency on ProtoGeometry. Remove in 3.0
+        [Obsolete]
```
mjkkirschner: so these are not used internally, but they are quite useful for third parties as they do the hard work of converting the dynamo coordinate system into the helix coordinate system - for someone not familiar with this it's going to be hard to know to invert the axes and perform a rotation.

src/DynamoManipulation/NodeManipulator.cs
```diff
@@ -88,6 +88,28 @@ internal IRenderPackageFactory RenderPackageFactory
         }
 
         internal Point3D? CameraPosition { get; private set; }
+        
+        // This is null for each new manipulator that is initialized 
+        private bool? enabled;
```
mjkkirschner: nullable bool - scary!

src/DynamoManipulation/NodeManipulator.cs
```diff
@@ -88,6 +88,28 @@ internal IRenderPackageFactory RenderPackageFactory
...
+
+        /// <summary>
+        /// A manipulator is enabled when its corresponding node is
+        /// not frozen, its geometry preview is visible and its cached value is valid.
+        /// </summary>
+        public bool Enabled
+        {
+            get 
+            {
+                if (enabled == null)
```
mjkkirschner: I'm confused by this - so once this is cached, it will never be recomputed? Shouldn't this backing field get set to null everywhere you need a recomputation to occur? Like when the cachedvalue is modified?

aparajit-pratap: It is set to null whenever a new manipulator is instantiated.

src/DynamoCoreWpf/Rendering/HelixRenderPackage.cs
aparajit-pratap: Were they added for third parties in the first place?

QilongTang: Usually we provide an alternative for obsolete methods, in this case, can you add obsoletion message maybe just indicating that we do not intend to public this in the first place?

src/DynamoCoreWpf/Rendering/HelixRenderPackage.cs
```diff
@@ -95,12 +97,15 @@ public void SetTransform(Autodesk.DesignScript.Geometry.CoordinateSystem transfo
...
 
+        
         /// <summary>
         /// Set the transform that is applied to all geometry in the renderPackage
         /// by computing the matrix that transforms between from and to.
         /// </summary>
         /// <param name="from"></param>
         /// <param name="to"></param>
+        // TODO: Unused method. Adds unnecessary dependency on ProtoGeometry. Remove in 3.0
+        [Obsolete]
```
QilongTang: Same here

src/DynamoManipulation/NodeManipulator.cs
```diff
@@ -551,11 +573,14 @@ public RenderPackageCache BuildRenderPackage()
...
 
+        #endregion
+
         /// <summary>
         /// Checks if manipulator is enabled or not. Manipulator is enabled 
         /// only if node is not frozen or not setup as partially applied function 
         /// </summary>
         /// <returns>True if enabled and can be manipulated.</returns>
+        // TODO: Make private in 3.0.
         public bool IsEnabled()
```
QilongTang: Seems we also want to obsolete this property

src/DynamoManipulation/NodeManipulator.cs
QilongTang: Having multiple boolean in this class getting enable state is confusing.. Would you consider getting enum state instead?

aparajit-pratap: Do you mean having an `Enabled` bool is confusing with an `IsEnabled()` method?

src/DynamoCoreWpf/Rendering/HelixRenderPackage.cs
mjkkirschner: I did intend them to be public - and yes, I added them for myself acting as a third party dev writing a powerMill version of Dynamo where transforming geometry using helix was not possible and using protogeometry to do it was a non starter in terms of performance.

mjkkirschner: I guess we can move these to protogeometry to avoid the tangled dependencies?

src/DynamoManipulation/NodeManipulator.cs
mjkkirschner: sorry, I'm still confused here - does a new manipulator get created whenever the node gets executed? Or is the old manipulator potentially reused from a past execution?

src/DynamoManipulation/NodeManipulator.cs
QilongTang: Yes especially they mean different thing within the class, I do understand we can't touch the signature of `IsEnabled()` to simplify this. So I hope if we add a new property, we can return multiway states enums instead of  true/false/null. My two cents

src/DynamoManipulation/NodeManipulator.cs
aparajit-pratap: An old manipulator is reused from a past execution (i.e. even if point node is updated). A new manipulator is created each time you select a new node or reselect an existing node.

src/DynamoCoreWpf/Rendering/HelixRenderPackage.cs
mjkkirschner: or more simply, we could just create an interface for CoordinateSystems/matrices that represent the coordinate system for dynamo geometry.

src/DynamoManipulation/Gizmo.cs
```diff
@@ -163,7 +163,7 @@ internal Gizmo(NodeManipulator manipulator)
 
         private void Redraw()
         {
-            if (manipulator.Enabled)
+            if (manipulator.IsEnabled())
```
mjkkirschner: so this happens whenever the camera is moved?

aparajit-pratap: This call takes place for every camera changed event and there's another call to `IsEnabled()` in `MouseMove()`, however since the `isValidNode` field is cached, these methods will not query for the node's cached value each time.

mjkkirschner:
@aparajit-pratap LGTM

PR #10155, opened 2019-11-21
DYN-2240

reddyashish:
### Purpose

This PR is to make simplify the terminology used in the workspace dependency extension. 

Task: https://jira.autodesk.com/browse/DYN-2240

Screen Shot 2019-11-26 at 2 57 36 PM

Looking for suggestions. 

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
+++ b/src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
@@ -12,7 +12,7 @@ namespace Dynamo.Graph.Workspaces
         IncorrectVersion,  // Correct package but incorrect version. 
         Missing,           // package is completely missing.
         Warning,           // Actual package is missing but the nodes are resolved by some other package. 
-        RequiresRestart    // Restart needed in order to complete the uninstall of some package. Notice this would be only set when workspace dependency viewer is loaded.
+        RequiresRestart    // Restart needed in order to complete the uninstall of some package. Notice this would be only set when workspace references extension is loaded.
     }
 
     /// <summary>
--- a/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
+++ b/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
@@ -211,7 +211,7 @@ namespace Dynamo.Controls
 
                 // creates a new tab item
                 TabItem tab = new TabItem();
-                tab.Header = viewExtension.GetType().Name;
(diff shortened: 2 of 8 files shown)


mjkkirschner:
@reddyashish @Amoursol I don't think we should include the name `package` in this extension - right now it only supports packages of course - here is my reasoning:


1. The intention is for this extension/feature to handle more dependency types than packages - like zero touch, custom nodes, hosts themselves (like core Revit library), etc.
2. The name now includes `PackageManager` .... that seems like a confusing overload of the same name as the Dynamo package manager, as most users simply refer to it as `package manager`

what do you guys think of `Workspace Requirements`? or `Requirements Manager`? 


@reddyashish have you also looked through the tooltips and messages?

reddyashish:
@mjkkirschner Yes, I agree with you. Sol has suggested the name "Workspace Manager" but I was thinking we can make it more clear. Required Libraries looks like a good choice but I was hoping the work "Workspace" to be included as these are specifically for each workspace.

mjkkirschner:
@reddyashish -  what about `Required Workspace Libraries?` or `Referenced Workspace Libraries`?

mjkkirschner:
@kronz might have some thoughts as well.

QilongTang:
@reddyashish @mjkkirschner @Amoursol @kronz How about simply `Workspace References`, `Reference` is a widely used term already and Visual Studio use it as well for listing what current coding project is depending on in order to build and work. I would like to also limit the name as short as possible and as catchy as possible for future workshops and answer bar duty. My 2 cents

mjkkirschner:
@QilongTang I like `Workspace References` as well because it can include any reference type if we want extend it further in the future and the name still makes sense.

reddyashish:
I am good with Workspace References as well unless someone has a better suggestion. Will change it to this.

kronz:
FWIW, you might look at what similar functionality is called in our host applications. For instance, Revit has "Manage Links", Autocad has a "Data Link Manager". "Manager" is problematic with the name collision with the PM. Could also check with HIG folks if there is a standardizing language (Tom Vollaro would be a good person to start with"

reddyashish:
Changed the extension name to Workspace References and triggered a self serve run.

reddyashish:
Self-serve CI passed with no failures.

QilongTang:
LGTM

reddyashish:
Will cherry-pick this to 2.5 as well.

PR #10162, opened 2019-11-23
get ready for removal of  sys.win.int - manually hookup events

mjkkirschner:
### Purpose

One approach to removing system.windows.Interactivity - 
it's ugly and brittle but it does load - and mostly works...

when the portViewModel collection is modified we hook up some event handlers and redirect these to the commands they are intended to call on the portViewmodel... 

We need to keep track of the observableCollection of `portViewModels` and remove the handlers etc etc.

meh.

### Declarations

### Reviewers

### FYIs


aparajit-pratap:
@mjkkirschner what is the change you've made to the `sys.win.int` binary? Should it not be removed in this PR?

mjkkirschner:
Hey @aparajit-pratap - I forgot to include this in the PR descriptions, but have mentioned it a few times - it's arguable if we can remove the dll at this time completely as we still have a public type that inherits from it


`HandlingEventTrigger : System.Windows.Interactivity.EventTrigger`

it's kind of an odd case as it's in DynamoCoreWPF and is used in our views, but someone could have used or derived from this type.

aparajit-pratap:
> Hey @aparajit-pratap - I forgot to include this in the PR descriptions, but have mentioned it a few times - it's arguable if we can remove the dll at this time completely as we still have a public type that inherits from it

Oh so you've just updated the version here?

mjkkirschner:
yes updated the version - and while the `HandlingEventTrigger` type is still here, it's no longer used.

PR #10163, opened 2019-11-23
get ready for removal of sys.win.int - embed xaml.behaviors.wpf EventTrigger

mjkkirschner:
### Purpose

This was an experiment in embedding the necessary code from https://github.com/microsoft/XamlBehaviorsWpf which is MIT licensed. (this is the new open source version of system.windows.interactivity)

and wrapping that in our own namespace in DynamoCoreWPF. It works, it should be way more solid, performant, and memory stable than whatever workaround we can come up with - but it's ugly and a maintenance issue to have this code here...

 It makes me think the best course of action might be to just update the sys.win.int 3.x (align with Revit) for now - and at the next major release switch to consuming `xaml.behaviors.wpf` instead.

feedback on these two PRs is appreciated.
https://github.com/DynamoDS/Dynamo/pull/10162

### Declarations

### Reviewers

### FYIs


mjkkirschner:
we may also be able to take the existing dll, modify it with something like `mono.cecil` and re-export it with a different name, so that we don't need to keep building the code or keep it here.
I'm not sure how the exceptionStrinTable bit would work, but we can look at that in the future.

PR #10165, opened 2019-11-23
AGD-1300 Add mechanism for common package directory similar to common definition directory

saintentropy:
### Purpose

The purpose of this PR is to add a mechanism to allow a common package installation directory similar to the common custom node directly already supported in the PathManager.  The common package directly will allow for the installation of packages that are available to all users on a specific machine. This common directory will supplement the current default user package and definitions directories located in user/appdata/...  

### Declarations

### Reviewers

@mjkkirschner @QilongTang 

### FYIs

@sm6srw @Dewb

### DIFF
--- a/src/DynamoPackages/PackageLoader.cs
+++ b/src/DynamoPackages/PackageLoader.cs
@@ -91,7 +91,7 @@ namespace Dynamo.PackageManager
         {
             InitializePackageDirectories(packagesDirectories);
 
-            if (packagesDirectories == null)
+            if (packageDirectoriesToVerify == null)
                 throw new ArgumentNullException("packageDirectoriesToVerify");
 
             this.packagesDirectoriesToVerifyCertificates.AddRange(packageDirectoriesToVerify);
--- a/src/DynamoPackages/PackageLoader.cs
+++ b/src/DynamoPackages/PackageLoader.cs
@@ -75,12 +75,29 @@ namespace Dynamo.PackageManager
             get { return packagesDirectories[0]; }
         }
 
+        private readonly List<string> packagesDirectoriesToVerifyCertificates = new List<string>();
+
         public PackageLoader(string overridePackageDirectory)
             : this(new[] { overridePackageDirectory })
         {
         }
 
         public PackageLoader(IEnumerable<string> packagesDirectories)
+        {
+            InitializePackageDirectories(packagesDirectories);
+        }
+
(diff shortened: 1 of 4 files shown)


QilongTang:
@saintentropy Where would we set the default common package dir? Does this PR require follow-up PRs touching preference setting?

saintentropy:
I think that is covered.  You still configure the PathResolver the exact same way via UserDataFolder and CommonDataFolder the same way that core and integrators are doing it today.  For Dynamo we pass `C:\Users\Foo\AppData\Roaming\Dynamo\Dynamo Core` and `C:\ProgramData\Dynamo\Dynamo Core`.  In Revit we pass `C:\Users\Foo\AppData\Roaming\Dynamo\Dynamo Revit` and `C:\ProgramData\Autodesk\RVT 2021\Dynamo` her -> https://github.com/DynamoDS/DynamoRevit/blob/e3ac3ecff0032d892eb30f0c6495173f95b1f8dd/src/DynamoRevit/DynamoRevit.cs#L448.   The other thing that doesn't change is that we don't user changes these.  If a user deletes the current default package folder, we add it back in on initialization of the DynamoModel.  We don't even let you see the definitions folders we add in the common and user data folders.   This package directory works the same.  It is there even if the user deletes it

src/DynamoPackages/PackageLoader.cs
```diff
@@ -382,6 +417,34 @@ public Package ScanPackageDirectory(string directory)
...
+                                }
+
+                            }
+                        }
+                        catch (Exception e)
+                        {
+                            //catch all exception types and pass to next throw
+                        }
+
+                        throw new LibraryLoadFailedException(directory, String.Format("A package called {0} found at {1} did not have signed dll files.  Ignoring it.", discoveredPkg.Name, discoveredPkg.RootDirectory));
```
Dewb: Shouldn't this be inside the above catch block?

saintentropy: I can modify it to work that way.  I was trying to avoid two throws; one with the the try catch that covers any exception in the `Assembly.ReflectionOnlyLoadFrom` or cert creation and one I call directly because of the verify() result is `false`.  But it might be clearer code that way.

saintentropy:
Closing this for now... Created a new PR https://github.com/DynamoDS/Dynamo/pull/10208 to make more specific changes to certificate verification strategy

PR #10171, opened 2019-11-26
replace system.windows.interactivity with custom event to command attached property

mjkkirschner:
### Purpose

Here is the last approach I wanted to prototype and get feedback on for removing `system.windows.interactivity`. This essentially replicates a tiny subset of the functionality in `system.windows.interactivity` to redirect events to commands - because sometimes we need to bind multiple events and commands to each uielement I modified an example from this article:
http://www.mobilemotion.eu/?p=1822

to handle comma separated lists event and command names.

There will be some perf penalty because of the string splitting and index searching but I think it will be quite minimal.

Perf and memory need to be profiled vs the existing solution. IE what happens when hundreds or thousands of nodes are drawn, what about dispose, are the events detached correctly? etc.

### Declarations

### Reviewers

### FYIs


src/DynamoCoreWpf/UI/Themes/Modern/Ports.xaml
```diff
@@ -88,19 +87,13 @@
                 <Rectangle
                     Fill="Transparent"
                     SnapsToDevicePixels="True"
-                    IsHitTestVisible="{Binding IsHitTestVisible}">
+                    IsHitTestVisible="{Binding IsHitTestVisible}"
+                        views:EventToCommand.Events="MouseEnter,MouseLeave,MouseLeftButtonDown"
```
mjkkirschner: note that now we pass a list of event names and command names  - we can do some validation that these need to have the same number of entries at runtime if we want.

src/DynamoCoreWpf/UI/HandlingEventTrigger.cs
```diff
@@ -13,4 +19,113 @@ protected override void OnEvent(System.EventArgs eventArgs)
...
+                    var methodInfo = typeof(EventToCommand).GetMethod("OnEvent", BindingFlags.Static | BindingFlags.NonPublic);
+                    //TODO clear previous subscriptions?
+                    eventInfo.GetAddMethod().Invoke(obj,
+                        new object[] { Delegate.CreateDelegate(eventInfo.EventHandlerType, methodInfo) });
+                }
+            }
+        }
+
+        //TODO handle the events.
+        private static void OnEvent(object sender, RoutedEventArgs e)
```
mjkkirschner: this is called when the event is called.

src/DynamoCoreWpf/UI/HandlingEventTrigger.cs
```diff
@@ -13,4 +19,113 @@ protected override void OnEvent(System.EventArgs eventArgs)
...
+
+        public static void SetEvents(DependencyObject obj, string value)
+        {
+            //parse string "eventName1,eventName2";
+
+            var eventNames = parseEventNames(value);
+            obj.SetValue(EventProperty, eventNames);
+        }
+
+        public static void EventPropertyChangedCallback(DependencyObject obj, DependencyPropertyChangedEventArgs args)
```
mjkkirschner: this is called when the event property is set in the xaml.

src/DynamoCoreWpf/UI/HandlingEventTrigger.cs
```diff
@@ -13,4 +19,113 @@ protected override void OnEvent(System.EventArgs eventArgs)
...
+
+    //forked from: http://www.mobilemotion.eu/?p=1822
+
+
+    public class EventToCommand
+    {
+        public static readonly DependencyProperty CommandProperty =
+            DependencyProperty.RegisterAttached("Commands", typeof(object), typeof(EventToCommand));
+
+        public static ICommand[] GetCommands(DependencyObject obj)
```
mjkkirschner: these setters and accessors are not used when set from xaml - so unless someone else sets these dep properties from code behind it's unlikely to be used - but it's required for making an attached dep property.

aparajit-pratap:
Out of the 3 potential solutions I find this one to be the simplest and cleanest one however, I'm not sure about the pros and cons of the first solution as compared to this one.

PR #10217, opened 2019-12-13
Add ability to hide IsUsageReportingApproved and

DmytroMuravskyi:
… IsAnalyticsReportingApproved from Dynamo UI

### Purpose

Alias team needs a way to suppress the data reporting from UI, so it can be controlled by Alias.
The property controls visibility of 2 menu items and prevents report window from being shown during first run.
The HideReportOptions may  suitable in DynamoViewModel.StartConfiguration but for comparability reasons I decided to add it only as a property for now. 

Would be good if this change goes into 2.5 since this is target version for integrators. 

## Reviewers
@mjkkirschner

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
@@ -470,7 +470,7 @@ namespace Dynamo.ViewModels
             get { return BackgroundPreviewViewModel.Active; }
         }
 
-        internal bool HideReportOptions { get; set; } = false;
+        public bool HideReportOptions { get; }
 
         #endregion
 
--- a/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
@@ -470,11 +470,7 @@ namespace Dynamo.ViewModels
             get { return BackgroundPreviewViewModel.Active; }
         }
 
-        /// <summary>
-        /// If true, Analytics and Usage options are hidden from UI 
-        /// TO DO: include to StartConfiguration for Dynamo 3.0
-        /// </summary>
-        public bool HideReportOptions { get; set; } = false;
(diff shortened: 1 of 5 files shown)


src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -469,7 +469,13 @@ public bool BackgroundPreviewActive
         {
             get { return BackgroundPreviewViewModel.Active; }
         }
-        
+
+        /// <summary>
+        /// If true, Analytics and Usage options are hidden from UI 
+        /// TO DO: include to StartConfiguration for Dynamo 3.0
+        /// </summary>
+        public bool HideReportOptions { get; set; } = false;
```
QilongTang: Is this meant to be public?

DmytroMuravskyi: I need to set it from outside somehow. I would rather make it private and use StartConfiguration, because HideReportOptions is meant to be set only once, before creating the view. But I concerned about doing it in minor release. Do you have any ideas were the best place to put this boolean?

mjkkirschner: @DmytroMuravskyi as long as you do not modify the interface, and only add new members to the concrete class - I believe it should be fine API compatibility wise.

QilongTang: @mjkkirschner Agree, sorry that I did not see the view binding. With that, this has to be public

DmytroMuravskyi: @mjkkirschner @QilongTang I've put it into StartConfiguration and made it public and readonly in ViewModel. 
Similar to ShowLogin, you can skip it and default false is used.
I see the PR is already approved, please take a final look at DynamoViewModel.cs

QilongTang:
@DmytroMuravskyi Thank you for the updates, exactly what I was worried before. These looks good. Merging

PR #10222, opened 2019-12-14
Webviewcontrol lib-  option2

mjkkirschner:
### Purpose

This PR removes CEF and CEFSharp from the libraryViewExtension and replaces it with `Microsoft.Toolkit.Wpf.UI.Controls.WebView`
read more here:
https://docs.microsoft.com/en-us/windows/communitytoolkit/controls/wpf-winforms/webview
which renders with edge.

This investigation was started as a way for hosts of Dynamo that have dependencies on other versions of CEF to maintain a functioning and similar user experience.

The available apis are much simpler than CEFSharp - and much more restricted - but after much finagling I think I have almost everything working as before with slower startup time, but faster actual performance during use. The code is simpler, but currently fragile, we can work to improve its robustness over time.

❗️ 
My intention with this PR is to get feedback and testing - I think to actually deliver this component we should make it a separate extension and test it in other integrations before pulling it into core.

Description of architecture:

* This component only really allows interacting between C# and JS in two ways. - 
*window.external.notify(string)* method - which triggers an event on the c# side or - `InvokeScriptAsync(strings)` which invokes a method on the js side - only strings can be passed between the contexts.

* It was not hard to refactor out the events and method calls we were using previously to a single event handler with a dumb format (functionName, params) where params are some json encoded data.

* The biggest challenge in this integration is to deal with resource loading - unlike CEFSharp - we cannot intercept requests from the browser context and return data back - we can only serve files from specific folders - I never got this to work, and didnt' really want to serialize data on the fly just to load it back into the webcontext - instead everything(images, fonts svgs etc)  is embedded into the json and html that we load or that we serialize and pass to the library functions.

* I have refactored the resourceProviders to find the correct resources and convert them to base64 encoded strings which are then used to replace the filePaths which we passed previously. This is likely wasteful and can be optimized if we have repeated icons. We can also likely cache  repeated lookups for images we have read from disk previously. None of this is done (except for the default icon because come on 😉  )

#### interesting facts!
* this breaks API compatibility with LibrarieJS - BUT I don't think we should consider this extension an API in any sense anyway. I think we should contract what we consider our "API" and make it super clear - for example, how revit ships a super obvious `RevitAPI.dll`.
* 

Screen Shot 2019-12-13 at 8 16 49 PM

 movlib

here is what I have tested:
- [x] search
- [x] loading a package
- [x] creating custom node
- [x] placing nodes
- [x] scroll
- [x] category navigation link
- [x] filter
- [x] tooltips / hover
- [x] expanding/compressing / browse

here is what I know we need to do:
- [ ] make startup faster
- [ ] pull this out into a new extension repo
- [ ] determine how to get this signed with BRE's help and delivered to artifactory/nuget
- [ ] hide the webView when startup page is visible - this is called the airGap problem if anyone is interested:  https://github.com/dotnet/wpf/issues/152
- [ ] There are no tests for this implementation and the others were very bound to the previous implementation.
- [ ] clean up code.... 🔨 
- [ ] potentially work on getting this build going - not sure it's important as I'd like to extract it from Dynamo ASAP.

### Declarations

### Reviewers
@QilongTang @reddyashish
### FYIs

 @Dewb @aparajit-pratap @Amoursol @kronz @angelowang


Dewb:
The [WebView known issues list](https://docs.microsoft.com/en-us/windows/communitytoolkit/controls/wpf-winforms/webview-known-issues) makes me a little nervous.

It seems like it would be feasible to implement a simple resource-loading mechanism through `InvokeScriptAsync`, would it not?

PR #10230, opened 2019-12-17
[DYN-2349] Close workspace references extension tab by user action and API

reddyashish:
### Purpose

This PR is to address the bug related to the closing of the extension tab. 

Task: https://jira.autodesk.com/browse/DYN-2349

close extension

close extension new

TODO:
There is still one bug related to the check mark in the tab menu for the workspace references extension. Looking into that ----> Fixed

### Declarations

### Reviewers

### DIFF
--- a/src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
+++ b/src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
@@ -147,7 +147,6 @@ namespace Dynamo.Wpf.Extensions
             }
         }
 
-
         private void AddItemToMenu(MenuBarType type, Control itemToAdd, int index)
         {
             if (dynamoMenu == null) return;
--- a/src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
+++ b/src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
@@ -83,7 +83,6 @@ namespace Dynamo.Wpf.Extensions
             dynamoMenu = dynamoView.titleBar.ChildOfType<Menu>();
             ViewStartupParams = new ViewStartupParams(dynamoVM);
             DynamoSelection.Instance.Selection.CollectionChanged += OnSelectionCollectionChanged;
-            DynamoView.CloseExtension += this.OnExtensionTabClosedHandler;
         }
 
         public void AddMenuItem(MenuBarType type, MenuItem menuItem, int index = -1)
@@ -148,18 +147,6 @@ namespace Dynamo.Wpf.Extensions
             }
         }
 
-        /// <summary>
-        /// This event is raised when an extension tab is closed and this event 
-        /// is subscired by the Workspace dependency view extension
-        /// </summary>
-        internal event Action OnExtensionTabClosed;
-        private void OnExtensionTabClosedHandler()
-        {
-            if (OnExtensionTabClosed != null)
-            {
-                OnExtensionTabClosed();
-            }
-        }
 
         private void AddItemToMenu(MenuBarType type, Control itemToAdd, int index)
         {
--- a/src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
+++ b/src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
@@ -152,7 +152,7 @@ namespace Dynamo.Wpf.Extensions
         /// This event is raised when an extension tab is closed and this event 
         /// is subscired by the Workspace dependency view extension
         /// </summary>
-        public event Action OnExtensionTabClosed;
+        internal event Action OnExtensionTabClosed;
         private void OnExtensionTabClosedHandler()
         {
             if (OnExtensionTabClosed != null)
--- a/src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
+++ b/src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
@@ -9,7 +9,6 @@ using Dynamo.Selection;
 using Dynamo.Utilities;
 using Dynamo.ViewModels;
 using Dynamo.Visualization;
-using Dynamo.Wpf.ViewModels;
 using Dynamo.Wpf.ViewModels.Watch3D;
 
 namespace Dynamo.Wpf.Extensions
@@ -84,6 +83,7 @@ namespace Dynamo.Wpf.Extensions
             dynamoMenu = dynamoView.titleBar.ChildOfType<Menu>();
             ViewStartupParams = new ViewStartupParams(dynamoVM);
             DynamoSelection.Instance.Selection.CollectionChanged += OnSelectionCollectionChanged;
+            DynamoView.CloseExtension += this.OnExtensionTabClosedHandler;
         }
 
         public void AddMenuItem(MenuBarType type, MenuItem menuItem, int index = -1)
@@ -145,6 +145,19 @@ namespace Dynamo.Wpf.Extensions
             if (SelectionCollectionChanged != null)
             {
                 SelectionCollectionChanged(notifyCollectionChangedEventArgs);
+            }
+        }
+
+        /// <summary>
+        /// This event is raised when an extension tab is closed and this event 
+        /// is subscired by the Workspace dependency view extension
+        /// </summary>
+        public event Action OnExtensionTabClosed;
+        private void OnExtensionTabClosedHandler()
+        {
+            if (OnExtensionTabClosed != null)
+            {
+                OnExtensionTabClosed();
             }
         }
 
--- a/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
+++ b/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
@@ -69,7 +69,7 @@ namespace Dynamo.Controls
         private ShortcutToolbar shortcutBar;
         private bool loaded = false;
         // This event is raised on the dynamo view when an extension tab is closed.
-        internal static event Action CloseExtension;
+        internal static event Action<String> CloseExtension;
 
         internal ObservableCollection<TabItem> TabItems { set; get; } = new ObservableCollection<TabItem>();
 
@@ -260,7 +260,7 @@ namespace Dynamo.Controls
         {
             string tabName = (sender as Button).DataContext.ToString();
 
-            CloseExtension?.Invoke();
+            CloseExtension?.Invoke(tabName);
 
             TabItem tabitem = TabItems.OfType<TabItem>().SingleOrDefault(n => n.Header.ToString() == tabName);
             CloseTab(tabitem);
--- a/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
+++ b/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
@@ -68,6 +68,8 @@ namespace Dynamo.Controls
         internal ViewExtensionManager viewExtensionManager;
         private ShortcutToolbar shortcutBar;
         private bool loaded = false;
+        // This event is raised on the dynamo view when an extension tab is closed.
+        internal static event Action CloseExtension;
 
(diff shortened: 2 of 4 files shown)


src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -256,38 +259,37 @@ internal TabItem AddTabItem(IViewExtension viewExtension, ContentControl content
         private void CloseTab(object sender, RoutedEventArgs e)
         {
             string tabName = (sender as Button).CommandParameter.ToString();
-            CloseTab(tabName);
+
+            TabItem tabitem = TabItems.OfType<TabItem>().SingleOrDefault(n => n.Name == tabName);
+            CloseTab(tabitem);
```
mjkkirschner: you should either check for null here or in `CloseTab` but it looks like you removed the null check?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -256,38 +259,37 @@ internal TabItem AddTabItem(IViewExtension viewExtension, ContentControl content
...
         /// </summary>
-        /// <param name="tabName">tab name</param>
-        private void CloseTab(string tabName)
+        /// <param name="tabitem">tab item</param>
+        private void CloseTab(TabItem tabitem)
         {
-            TabItem tab = tabDynamic.SelectedItem as TabItem;
+            TabItem tabToBeRemoved = tabitem;
 
-            if (tab != null)
```
mjkkirschner: I think it's a good idea to keep this check?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -227,6 +229,7 @@ internal TabItem AddTabItem(IViewExtension viewExtension, ContentControl content
                 TabItem tab = new TabItem();
                 tab.Header = viewExtension.Name;
                 tab.Tag = viewExtension.GetType();
+                tab.Uid = viewExtension.UniqueId;
```
mjkkirschner: are we going to stick to name or use this?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -203,14 +203,16 @@ public DynamoView(DynamoViewModel dynamoViewModel)
...
+        /// This method will close a tab item in the right side bar based on passed extension
         /// </summary>
         /// <param name="viewExtension">Extension to be closed</param>
         /// <returns></returns>
         internal void CloseTabItem(IViewExtension viewExtension)
         {
             string tabName = viewExtension.Name;
-            CloseTab(tabName);
+
+            TabItem tabitem = TabItems.OfType<TabItem>().SingleOrDefault(n => n.Name == tabName);
```
QilongTang: Please switch to use UID here

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
QilongTang: I agree

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
QilongTang: @mjkkirschner Please use ID if we can, do we need to update AC?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -256,38 +259,37 @@ internal TabItem AddTabItem(IViewExtension viewExtension, ContentControl content
...
 
-                // clear tab control binding and bind to the new tab-list. 
-                tabDynamic.DataContext = null;
-                TabItems.Remove(tab);
-                TabItems = TabItems;
-                tabDynamic.DataContext = TabItems;
+            // clear tab control binding and bind to the new tab-list. 
+            tabDynamic.DataContext = null;
+            TabItems.Remove(tabToBeRemoved);
+            TabItems = TabItems;
```
aparajit-pratap: ??

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
reddyashish: Yes added the check back.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
reddyashish: Added it.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
reddyashish: Removed this unnecessary line. Thanks @aparajit-pratap

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
reddyashish: @QilongTang When an extension tab is closed using the "close" button, the sender object only has reference to the extension name and it does not pass the extension Uid. https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs#L258

Since we do not have a dictionary to retrieve the Uid using the name of the extension or vice-versa, this would be difficult. The extension names in the side bar are also unique in the sense that if any extension with the same name is added to the extension side bar, it would not add it. This will take care of that. https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs#L222

Even though it is definitely better to use Uid property, there is no issue using the extension names as of now.

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
```diff
@@ -148,6 +148,19 @@ private void OnSelectionCollectionChanged(object sender, NotifyCollectionChanged
             }
         }
 
+        /// <summary>
+        /// This event is raised when an extension tab is closed and this event 
+        /// is subscired by the Workspace dependency view extension
+        /// </summary>
+        public event Action OnExtensionTabClosed;
```
mjkkirschner: can we make this internal for now? 
Also we should probably put the name of the extension that is being closed here as a parameter so that other extensions can check if they should respond or not...

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -255,31 +258,35 @@ internal TabItem AddTabItem(IViewExtension viewExtension, ContentControl content
         // This method triggers the close operation on the selected tab. 
         private void CloseTab(object sender, RoutedEventArgs e)
         {
-            string tabName = (sender as Button).CommandParameter.ToString();
-            CloseTab(tabName);
+            string tabName = (sender as Button).DataContext.ToString();
```
mjkkirschner: when the button is constructed you can actually use the `tag` property to add the name of the tab instead of getting it this way - I think that makes it more explicit.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
```diff
@@ -90,6 +96,8 @@ public void Loaded(ViewLoadedParams viewLoadedParams)
                 DependencyView.DependencyRegen(viewLoadedParams.CurrentWorkspaceModel as WorkspaceModel);
             };
 
+            viewLoadedParams.OnExtensionTabClosed += OnCloseExtension;
```
mjkkirschner: I would run the self serve ASAP to see if this causes a memory leak during testing.

reddyashish: Yes, started the self-serve

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
reddyashish: Yes, I forgot to this change this to internal. Changed it now

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
mjkkirschner: hmm, maybe this is not possible - am I understanding correctly the tabitems close buttons are already present from the existing control?

reddyashish: They are present in the dynamoview.xaml and all the buttons have the same name. 
https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/Views/Core/DynamoView.xaml#L1721.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -255,31 +258,35 @@ internal TabItem AddTabItem(IViewExtension viewExtension, ContentControl content
...
+
+            CloseExtension?.Invoke();
```
mjkkirschner: I think we should pass the tabName.... can you check what type the tab's dataContext really is? is it actually the ViewExtension?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
mjkkirschner: theres probably some way to tag them with the header property of their dataContext instead of the name which was never set in the tabItem constructor.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
QilongTang: @reddyashish Cool, then we can stick to name for now and move on to GUID later. I just hope it is not going to be too many instances where people's work can't load due to other people's extension naming

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -168,6 +169,20 @@ public WorkspaceDependencyView(WorkspaceDependencyViewExtension viewExtension, V
             packageInstaller = p.PackageInstaller;
             dependencyViewExtension = viewExtension;
             DependencyRegen(currentWorkspace);
+            DynamoView.CloseExtension += this.OnExtensionTabClosedHandler;
+        }
+
+        /// <summary>
+        /// This event is raised when an extension tab is closed and this event 
+        /// is subscired by the Workspace dependency view extension
```
mjkkirschner: spelling

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
reddyashish: Added tabname as an input to the event.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
```diff
@@ -17,7 +17,8 @@ namespace Dynamo.WorkspaceDependency
     /// </summary>
     public class WorkspaceDependencyViewExtension : IViewExtension, ILogSource
     {
-        private MenuItem packageDependencyMenuItem;
+        internal MenuItem packageDependencyMenuItem;
+        private String extensionName = "Workspace References";
```
mjkkirschner: maybe set this readonly or const?

mjkkirschner:
@reddyashish LGTM just one last comment and lets see what the self serve run returns.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
reddyashish: Yes, changed it.

reddyashish:
Merging this.

PR #10233, opened 2019-12-18
add new extension LibraryViewExtension (based on Microsoft WPF.WebView)

mjkkirschner:
### Purpose

replaces https://github.com/DynamoDS/Dynamo/pull/10222

This PR differs in a few way:
* adds a new extension instead of modifying the existing one
* this one builds!
* links some static resource files from existing LibraryViewExtension but embeds them into new dll.
* duplicates most files from LibraryViewExtension - but adds a new namespace - `Dynamo.LibraryViewExtensionMSWebView` to avoid conflicts between them incase somehow these dlls are loaded at the same time by accident.

here is what I know we need to do (not all of this will be done in this PR)
- [ ] make startup faster
- [ ] hide the webView when startup page is visible - this is called the airGap problem if anyone is interested:  https://github.com/dotnet/wpf/issues/152
- [ ] There are no tests for this implementation.
- [ ] clean up code.... 🔨 
- [ ] eventually refactor out a common core of types from libraryJS and this extension to share. (longterm)
- [ ] add a new .sln for this project (TBD)
- [ ] determine how to get this signed with BRE's help or CI library and delivered to artifactory/nuget
 - [ ] this project should include a readme describing how to integrate this extension and replace the existing library.

### Declarations

Check these if you believe they are true

- [ ] The codebase is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.

### Reviewers

(FILL ME IN) Reviewer 1  (If possible, assign the Reviewer for the PR)

(FILL ME IN, optional) Any additional notes to reviewers or testers.

### FYIs

(FILL ME IN, Optional) Names of anyone else you wish to be notified of

PR #10241, opened 2019-12-24
Add new LibraryViewExtension project based on System.Windows.Controls.WebBrowser

mjkkirschner:
### Purpose

similar to: https://github.com/DynamoDS/Dynamo/pull/10233
but using WebBrower - which is not deprecated, and actually loads in complex native apps hosting their own .net runtimes. (Revit,Acad)

Uses ` <meta http-equiv="X-UA-Compatible" content="IE=edge" />` header to request latest IE 11 renderer.

LibraryJS has been compiled with `core-js` for polyfills though I think we only neeed `string.prototype.startsWith`

I'm using reflection to access Dynamo internal methods so that testing in hosts is easier for now in a few places. We can remove this later.

~~This approach still uses loading all images and content and embedding it into the html - I can't think of a nice way to implement an automatic request interceptor in js and I don't think IE11 supports service workers - I think we would have to rewrite libraryjs to make explicit requests for resources to the c# side - or something I have not thought of yet regarding the different uri schemas this component might support, none I've tried has obviously worked so I stuck with embedding.~~
 
~~A few optimizations and cacheing of data makes adding packages later fast - but the initial load is still underwhelming - about 6 seconds with a few packages installed. Profiling seems to indicate the time is spent accessing disk and using `ResourceManager.getObject` to scan for images by name within customization assemblies - this is the same as it was in vanilla libraryJS - except that now it's done all upfront instead of lazily as the library is used - which makes using the library a bit more fluid at the expense of the startup cost...~~

~~Theres likely some more optimization we could do here like cacheing the resource managers themselves - but it might be worth thinking of solutions for lazy loading - Ideally we could catch the requests from the browser and respond with images - as we could in the old implementation.~~

~~It's also possible we could load the image data on another thread to keep the UI fluid until the library is done loading.~~

* icons now work even for hosts which have added their own customizations
* the browser also is hidden when the startPage is displayed now.
* 

### Update:
* I have implemented a simple lazy loading - this loads an order of magnitude faster in my tests in both sandbox and hosts applications.
It works pretty simply, whenever the user interacts with the library (dom body click event), we get all the images from the DOM and attempt to update their `src` properties - if they already contain valid content, we do nothing - if they contain default images or no content then we send their url to the c# side which handles passes it to the `IconResourceProvider` this returns a base64 encoded version of the image, and the src is set to this updating the UI.

So the end result our first load of the library is very fast - we're still embedding a few common things like fonts, and librarie.js.min - since we immediately need to load them anyway. 

Later when a user interacts the icons are loaded as they open categories.

### instructions for testing:
This is a 7zip file renamed to txt - unzip it with 7zip, then move it to the `Dynamo/viewExtensions` folder and replace the existing `Library_ViewExtensionDefinition.xml` file.
[libraryMsWebBrowser4.txt](https://github.com/DynamoDS/Dynamo/files/4017548/libraryMsWebBrowser4.txt)

if you instead want to build this PR - the binaries and definition you want will be under `Dynamo/src/LibraryViewExtensionMSWebBrowser/bin/....` They are NOT being copied to the Dynamo/bin directory as we intend to have a separate solution / nuget package for this extension.

### Declarations

### Reviewers
@QilongTang @reddyashish


mjkkirschner:
actually we can do this:
https://blog.dotnetframework.org/2018/10/22/capture-ajax-requests-with-webbrowser-control-in-c/

override fetch or xmlHttpRequest to make all requests to an interceptor which will make requests to c# - c# side will fetch the resources and respond - this should let us get lazy loaded images.

src/Dynamo.All.sln
```diff
@@ -227,6 +227,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Extensions", "Extensions",
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "WorkspaceDependencyViewExtension", "WorkspaceDependencyViewExtension\WorkspaceDependencyViewExtension.csproj", "{5E76AAB3-6302-473E-9655-081B53FB1419}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "LibraryViewExtensionMSWebBrowser", "LibraryViewExtensionMSWebBrowser\LibraryViewExtensionMSWebBrowser.csproj", "{1A5DC90A-477E-4D4A-87BD-0BFB01F056CE}"
```
mjkkirschner: TBD if we will add a new .sln or keep it as part of this one. The binaries are already isolated from the main bin folder.

src/LibraryViewExtensionMSWebBrowser/Handlers/IconResourceProvider.cs
```diff
@@ -0,0 +1,280 @@
...
+
+        /// <summary>
+        /// Retrieves the resource for a given url as a base64 encoded string.
+        /// ie data:image/png;base64, stringxxxyyyzzz
+        /// </summary>
+        /// <param name="url">url for the requested icon</param>
+        /// <param name="extension">Returns the extension to describe the type of resource.</param>
+        /// <returns></returns>
+        public string GetResourceAsString(string url, out string extension)
+        {
```
mjkkirschner: this a new method and should be reviewed - here is where most of the work is done for looking up a url and finding the relevant icon.

src/LibraryViewExtensionMSWebBrowser/Handlers/IconResourceProvider.cs
```diff
@@ -0,0 +1,280 @@
...
+                try
+                {
+                    var icon = new IconUrl(new Uri(url));
+                    base64String = GetIconAsBase64(icon, out extension);
+
+                }
+                catch (Exception e)
+                {
+                    //look in resources for registered path and just use the stream directly
+                    var resourceDict = this.customization.Resources.ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
```
mjkkirschner: some hosts register streams directly so theres no need to look on disk, we can just return the stream.

src/LibraryViewExtensionMSWebBrowser/Handlers/LayoutSpecProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        /// <summary>
+        /// Gets the resource for the given request
+        /// </summary>
+        public override Stream GetResource(string url, out string extension)
+        {
+            extension = "json";
+            try
+            {
+                if (resourceStream == null || resourceStream.CanRead == false || resourceStream.Position == resourceStream.Length)
```
mjkkirschner: I added some extra fail safe logic here as I encountered some instances where the stream was already disposed or unreadable - I think this sometimes happens in Dynamo today actually as well. If we fail to read from the stream, then we create a new one.

src/LibraryViewExtensionMSWebBrowser/Handlers/NodeItemDataProvider.cs
```diff
@@ -0,0 +1,177 @@
...
+
+        protected Stream GetNodeItemDataStream(IEnumerable<NodeSearchElement> searchEntries, bool replaceIconURLWithData)
+        {
+            var ms = new MemoryStream();
+            var sw = new StreamWriter(ms);
+            var serializer = new JsonSerializer();
+            var stringBuilder = new StringBuilder();
+            var data = CreateObjectForSerialization(searchEntries);
+
+            if (replaceIconURLWithData)
```
mjkkirschner: we have the option of replacing the nodeItems iconUrl property with the base64Data before it gets sent to the library UI - or not (and letting) the image resolve method handle it - for searching, because the results are quite small we embed the icons as it will be fast.

src/LibraryViewExtensionMSWebBrowser/LibraryViewController.cs
```diff
@@ -0,0 +1,583 @@
...
+using Newtonsoft.Json.Linq;
+
+namespace Dynamo.LibraryViewExtensionMSWebBrowser
+{
+    /// <summary>
+    /// This object is exposed to the browser to pass data back and forth between contexts.
+    /// </summary>
+    [PermissionSet(SecurityAction.Demand, Name = "FullTrust")]
+    [ComVisibleAttribute(true)]
+    public class scriptingObject
```
mjkkirschner: this object is used to expose a few key methods to the javascript context.

src/LibraryViewExtensionMSWebBrowser/LibraryViewController.cs
```diff
@@ -0,0 +1,583 @@
...
+        public void notify(string data)
+        {
+            scriptNotifyHandler(data);
+        }
+
+        /// <summary>
+        /// Used to get access the iconResourceProvider and return a base64encoded string version of an icon.
+        /// </summary>
+        /// <param name="iconurl"></param>
+        public void getBase64StringFromPath (string iconurl)
```
mjkkirschner: called by javascript side to retrieve an image resource.

src/LibraryViewExtensionMSWebBrowser/LibraryViewController.cs
```diff
@@ -0,0 +1,583 @@
...
+        /// <param name="eventName">Event Name that gets logged to instrumentation</param>
+        /// <param name="data"> Data that gets logged to instrumentation </param>
+        public void LogEventsToInstrumentation(string eventName, string data)
+        {
+            if (eventName == "Search" || eventName == "Filter-Categories" || eventName == "Search-NodeAdded")
+            {
+                Analytics.LogPiiInfo(eventName, data);
+            }
+        }
+        internal void refreshLibraryView(WebBrowser browser)
```
mjkkirschner: called by javascript side, does a full refresh of the loadedTypes and LayoutSpec - this actually quite fast and is not called frequently. ( < 50ms in my tests with several large packages installed)

src/LibraryViewExtensionMSWebBrowser/LibraryViewCustomization.cs
```diff
@@ -0,0 +1,264 @@
...
+        /// <summary>
+        /// Serializes the layout specification to json stream and
+        /// optionally replaces icon urls with image data as base64 encoded strings
+        /// </summary>
+        /// <returns></returns>
+        private Stream ToJSONStream(LayoutSpecification spec, bool replaceIconURLWithData, IconResourceProvider iconProvider = null)
+        {
+            var ms = new MemoryStream();
+            var sw = new StreamWriter(ms);
+            if (replaceIconURLWithData && iconProvider != null)
```
mjkkirschner: option of replacing images in the layoutspec - these would be for categories  - currently this path is not used.

src/LibraryViewExtensionMSWebBrowser/LibraryViewExtensionMSWebBrowser.csproj
```diff
@@ -0,0 +1,319 @@
...
+    </EmbeddedResource>
+    <EmbeddedResource Include="..\LibraryViewExtension\web\library\resources\Tessellation.Delaunay.png">
+      <Link>web\library\resources\Tessellation.Delaunay.png</Link>
+    </EmbeddedResource>
+    <EmbeddedResource Include="..\LibraryViewExtension\web\library\resources\Tessellation.Voronoi.png">
+      <Link>web\library\resources\Tessellation.Voronoi.png</Link>
+    </EmbeddedResource>
+  </ItemGroup>
+  <ItemGroup>
+    <Content Include="Library_ViewExtensionDefinition.xml">
```
mjkkirschner: include the new manifest file.

src/LibraryViewExtensionMSWebBrowser/web/library/library.html
```diff
@@ -0,0 +1,136 @@
...
+                var src = element.src
+                if (element.orgSrc != null) {
+                    src = element.orgSrc;
+                }
+                //the icon is already set - bail.
+                if (src.startsWith("data:image")) {
+                    return;
+                }
+                //request the icon from the extension.
+                window.external.getBase64StringFromPath(src);
```
mjkkirschner: window.external is the `ObjectForScripting` we exposed before - so this lets us call c# functions.

src/LibraryViewExtensionMSWebBrowser/Handlers/IconResourceProvider.cs
```diff
@@ -0,0 +1,280 @@
...
+        /// ie data:image/png;base64, stringxxxyyyzzz
+        /// </summary>
+        /// <param name="url">url for the requested icon</param>
+        /// <param name="extension">Returns the extension to describe the type of resource.</param>
+        /// <returns></returns>
+        public string GetResourceAsString(string url, out string extension)
+        {
+            //sometimes the urls have "about:" added to them - remove this
+            //and do it before checking cache.
+            if (url.StartsWith("about:"))
```
QilongTang: hmm.. I have no idea about the reason so curious here if this can be avoided in other places?

src/LibraryViewExtensionMSWebBrowser/Handlers/LayoutSpecProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+            var stream = assembly.GetManifestResourceStream(resource);
+
+            //Get the spec from the stream
+            var spec = LayoutSpecification.FromJSONStream(stream);
+            customization.AddSections(spec.sections);
+            this.customization = customization;
+            this.customization.SpecificationUpdated += OnSpecificationUpdate;
+        }
+
+        public LayoutSpecProvider(ILibraryViewCustomization customization, IconResourceProvider iconProvider, string resource, Assembly assembly = null) :
```
QilongTang: Would you add description for this alternative constructor as well?

src/LibraryViewExtensionMSWebBrowser/Handlers/NodeItemDataProvider.cs
```diff
@@ -0,0 +1,177 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Dynamo.Search;
+using Dynamo.Search.SearchElements;
+using Dynamo.LibraryViewExtensionMSWebBrowser.Handlers;
+using Newtonsoft.Json;
+
+// TODO there are many clases in this file which are duplicates in LibraryViewExtension - 
+// these can be refactored out into a shared core.
```
QilongTang: Thanks, indeed. Will need a tech debt for this.

src/LibraryViewExtensionMSWebBrowser/Handlers/NodeItemDataProvider.cs
QilongTang: Excellent comment, would you put this in the code. I feel this explains it well

src/Dynamo.All.sln
mjkkirschner: there is one failure on the engops build job since this PR uses value tuples (c#7) I can remove these without any trouble.

src/LibraryViewExtensionMSWebBrowser/Handlers/NodeItemDataProvider.cs
mjkkirschner: will do

src/LibraryViewExtensionMSWebBrowser/Handlers/IconResourceProvider.cs
mjkkirschner: I also have no idea why this shows up sometimes, but I assume it's when the browser has trouble loading something - see:
https://en.wikipedia.org/wiki/About_URI_scheme

src/LibraryViewExtensionMSWebBrowser/EventObserver.cs
```diff
@@ -0,0 +1,167 @@
...
+        /// <param name="transform">This function will be invoked to transform the 
+        /// actual event parameter.</param>
+        public EventObserver(Action<TResult> observe, Func<TValue, TResult> transform)
+        {
+            observer = observe;
+            reducer = (x, y) => transform(y);
+            result = default(TResult);
+        }
+
+        public static TValue Identity(TValue last, TValue current)
```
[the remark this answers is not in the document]
mjkkirschner: I didn't modify this class at all.

src/LibraryViewExtensionMSWebBrowser/Handlers/IconResourceProvider.cs
```diff
@@ -0,0 +1,280 @@
...
+    /// Implements resource provider for icons
+    /// </summary>
+    class IconResourceProvider : ResourceProviderBase
+    {
+        private const string imagesSuffix = "Images";
+        private IPathManager pathManager;
+        private string defaultIconDataString;
+        private string defaultIconName;
+        private DllResourceProvider embeddedDllResourceProvider;
+        //TODO remove these at some point in future after Dynamo 2.6 release.
```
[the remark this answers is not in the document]
mjkkirschner: reflection is slow and weird, and fragile.

src/Dynamo.All.sln
mjkkirschner: fixed

src/LibraryViewExtensionMSWebBrowser/LibraryViewController.cs
```diff
@@ -0,0 +1,581 @@
...
+        #endregion
+
+        private string ReplaceUrlWithBase64Image(string html, string minifiedURL, bool magicreplace = true)
+        {
+            var ext = string.Empty;
+            // this depends on librariejs minification producing the same results - 
+            // longterm this is fragile. We should intercept these requests and handle them instead.
+            if (magicreplace)
+            {
+                minifiedURL = $"n.p+\"{minifiedURL}\"";
```
aparajit-pratap: What does this do?

src/LibraryViewExtensionMSWebBrowser/Handlers/NodeItemDataProvider.cs
mjkkirschner: filed

src/LibraryViewExtensionMSWebBrowser/LibraryViewController.cs
mjkkirschner: there are strings in the minified librarie.js file that look like this `n.p\resoruces\some\path\toAfont.font` - the n.p is some artifact of the minification - to replace these with embedded base64 versions of those resources we need to search for the whole string including the n.p part.

src/LibraryViewExtensionMSWebBrowser/LibraryViewController.cs
```diff
@@ -0,0 +1,581 @@
...
+        /// Main callback which libraryJS uses to notify the extension to perform some action.
+        /// </summary>
+        /// <param name="data"></param>
+        public void notify(string data)
+        {
+            scriptNotifyHandler(data);
+        }
+
+        /// <summary>
+        /// Used to get access the iconResourceProvider and return a base64encoded string version of an icon.
```
aparajit-pratap: `get access to ...`

src/LibraryViewExtensionMSWebBrowser/LibraryViewController.cs
```diff
@@ -0,0 +1,581 @@
...
+        private void scriptNotifyHandler(string dataFromjs)
+        {
+
+            if (dataFromjs == "requestUpdateLibrary")
+            {
+                controller.RefreshLibraryView(controller.browser);
+                
+            }
+            //a more complex action needs to be taken on the c# side.
+            else if (!string.IsNullOrEmpty(dataFromjs))
```
aparajit-pratap: You could invert the `if` condition by simply returning from the method for a null or empty string?

src/LibraryViewExtensionMSWebBrowser/ViewExtensionEntryPoint.cs
```diff
@@ -0,0 +1,100 @@
+﻿using System;
```
aparajit-pratap: How about renaming the file to `LibraryViewExtensionMSWebBrowser.cs` or simply `ViewExtension.cs`?

mjkkirschner: renamed

mjkkirschner:
@aparajit-pratap @QilongTang  comments addressed PTAL. If you want to go through this together let me know.

PR #10271, opened 2020-01-07
Event to handle the close operation of the workspace references extension

reddyashish:
# Purpose

This PR is to close the workspace references extension tab when the workspace is closed so that when the user reopens the graph with no missing dependencies, the extension is hidden. It will only be shown when the workspace has missing dependencies. 

Closing

### Declarations

### Reviewers

### DIFF
--- a/src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
+++ b/src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
@@ -33,7 +33,7 @@ namespace Dynamo.Graph.Workspaces
         private bool graphExecuted;
 
         // Event to handle closing of the extension when the workspace is closed. 
-        internal static event Action<String> CloseExtension;
+        internal static event Action WorkspaceCleared;
 
         private String workspaceReferencesExtensionName = "Workspace References";
 
@@ -425,7 +425,7 @@ namespace Dynamo.Graph.Workspaces
         /// </summary>
         public override void Clear()
         {
-            CloseExtension?.Invoke(workspaceReferencesExtensionName);
+            WorkspaceCleared?.Invoke();
             base.Clear();
             PreloadedTraceData = null;
             RunSettings.Reset();
--- a/src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
+++ b/src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
@@ -32,6 +32,12 @@ namespace Dynamo.Graph.Workspaces
         private readonly bool verboseLogging;
         private bool graphExecuted;
 
+        // Event to handle closing of the extension when the workspace is closed. 
+        internal static event Action<String> CloseExtension;
+
+        private String workspaceReferencesExtensionName = "Workspace References";
+
+
         // To check whether task is completed or not. 
         private bool executingTask;
 
@@ -419,6 +425,7 @@ namespace Dynamo.Graph.Workspaces
         /// </summary>
         public override void Clear()
         {
+            CloseExtension?.Invoke(workspaceReferencesExtensionName);
             base.Clear();
             PreloadedTraceData = null;
             RunSettings.Reset();
--- a/src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
+++ b/src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
@@ -32,11 +32,8 @@ namespace Dynamo.Graph.Workspaces
         private readonly bool verboseLogging;
         private bool graphExecuted;
 
-        // Event to handle closing of the extension when the workspace is closed. 
+        // Event to handle closing of the workspace references extension when the workspace is closed. 
         internal static event Action WorkspaceCleared;
-
-        private String workspaceReferencesExtensionName = "Workspace References";
-
 
         // To check whether task is completed or not. 
         private bool executingTask;
--- a/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
+++ b/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
@@ -197,11 +197,25 @@ namespace Dynamo.Controls
             // Add an event handler to check if the collection is modified.   
             TabItems.CollectionChanged += this.OnCollectionChanged;
 
+            HomeWorkspaceModel.CloseExtension += this.CloseExtensionTab;
+
             this.HideOrShowRightSideBar();
 
             this.dynamoViewModel.RequestPaste += OnRequestPaste;
             this.dynamoViewModel.RequestReturnFocusToView += OnRequestReturnFocusToView;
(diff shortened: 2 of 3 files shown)


reddyashish:
Started the self-serve run with these changes

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -32,6 +32,12 @@ public class HomeWorkspaceModel : WorkspaceModel
         private readonly bool verboseLogging;
         private bool graphExecuted;
 
+        // Event to handle closing of the extension when the workspace is closed. 
+        internal static event Action<String> CloseExtension;
```
mjkkirschner: this seems like a weird name for an event:
```
Do name events with a verb or a verb phrase.

Do give event names a concept of before and after, using the present and past tense. For example, a close event that is raised before a window is closed would be called Closing and one that is raised after the window is closed would be called Closed.

Do not use Before or After prefixes or suffixes to indicate pre and post events.

Do name event handlers (delegates used as types of events) with the EventHandler suffix.

Do use two parameters named sender and e in event handler signatures.

The sender parameter should be of type Object, and the e parameter should be an instance of or inherit from EventArgs.

Do name event argument classes with the EventArgs suffix.
```

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -197,13 +197,27 @@ public DynamoView(DynamoViewModel dynamoViewModel)
...
         }
 
+        /// <summary>
+        /// This method is called when the home workspace is closed. This will inturn close the workspace references extension.
+        /// </summary>
+        /// <param name="extensionTabName">Extension name</param>
+        /// <returns></returns>
+        internal void CloseExtensionTab(String extensionTabName)
+        {
+            TabItem tabitem = TabItems.OfType<TabItem>().SingleOrDefault(n => n.Header.ToString() == extensionTabName);
```
mjkkirschner: I really think this is a bad name for this control:
`TabItems` - because the DynamoView also has another set of TabItems for the workspaces...

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
reddyashish: Yes, I did not give much thought on this naming yet. I will change this accordingly, just waiting for the tests to see if any other changes are needed.

mjkkirschner:
@reddyashish - actually I have been thinking about this a bit more:

This seems weirdly specific for the `WorkspaceReferences Extension` to have code that lives in homeworkspaceModel and DynamoView...

Instead, what about just adding some code to the `WorkspaceReferences Extension` to watch for the close or clear events on the current homework space and close when that occur, I think that is a better design if this functionality is only being implemented for this extension thats where I would look for it - on the other hand if this is a general implementation for all extensions then it belongs where you put it.

I agree in general that this behavior is actually not desirable for all extensions, so my vote would be to implement this as part of the extension logic, and not as part of DynamoCore.

reddyashish:
@mjkkirschner made the necessary code changes. Running the self-serve now and will add some tests to this.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
reddyashish: Should I rename this to ExtensionTabItems so that it would be clear to differentiate?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -255,6 +243,19 @@ internal TabItem AddTabItem(IViewExtension viewExtension, ContentControl content
...
+        /// <summary>
+        /// This method will close a tab item in the right side bar based on passed extension
+        /// </summary>
+        /// <param name="viewExtension">Extension to be closed</param>
+        /// <returns></returns>
+        internal void CloseTabItem(IViewExtension viewExtension)
+        {
+            string tabName = viewExtension.Name;
+            TabItem tabitem = TabItems.OfType<TabItem>().SingleOrDefault(n => n.Header.ToString() == tabName);
+            CloseExtension?.Invoke(tabName);
```
mjkkirschner: closeTab (below) is already raising this event... why do you need to raise it twice?

reddyashish: CloseTab(object sender, RoutedEventArgs e) is triggered when the close button is clicked by the user and this event is raised to handle the Workspace references menu item.

mjkkirschner: ahh, would you mind renaming the eventHandler so the name does not conflict, would clear up some confusion here I think?

mjkkirschner: `CloseTabEventHandler`? maybe?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
QilongTang: This is renamed in https://github.com/DynamoDS/Dynamo/pull/10279

mjkkirschner:
@reddyashish I see @QilongTang renamed the method already - LGTM - lets just be sure theres no conflicts.

QilongTang:
@reddyashish Sorry for the conflicts! Once that is resolved, you should be good to merge

reddyashish:
@QilongTang @mjkkirschner No worries, thanks for the changes. I will pull in your changes.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
reddyashish: I have pulled in your changes and this PR will be merged on top of #10279. So it should be fine.

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -32,6 +32,9 @@ public class HomeWorkspaceModel : WorkspaceModel
         private readonly bool verboseLogging;
         private bool graphExecuted;
 
+        // Event to handle closing of the workspace references extension when the workspace is closed. 
+        internal static event Action WorkspaceCleared;
```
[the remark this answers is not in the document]
reddyashish: Yes, we can change it to this name. Will do it in the new PR

reddyashish:
Closing this and opening a new one to resolve the conflicts. https://github.com/DynamoDS/Dynamo/pull/10280

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
reddyashish: Resolving this as it is addressed by Aaron's PR

PR #10285, opened 2020-01-10
pass the additional to sign list to the installerspec generator

mjkkirschner:
add new files
add comments
remove files that no longer exist
update installerSpec.exe
update installerspec source

### Purpose

This PR updates the install.sln which is still run during the build to sign binaries - to pass a file which lists additional binaries to sign which might have different version numbers from dynamoCore and thus are skipped in the first pass.

It also lets us add comments to this file, and white space lines....

Additionally, if a file fails to be found which was requested in this list, the build fails.

To find the files we were missing I used SigCheck
https://docs.microsoft.com/en-us/sysinternals/downloads/sigcheck
`sigcheck -u -e -s DynamoCoreRuntime2.6xxx/ > unsigned.txt`

I've attached the list to the jira task, we may want to consider running this over every build in our binary diff job and making sure the number does not change from build to build without us knowing.

I have tested manually and the generated binariestosign.txt includes the files.

### Declarations

### Reviewers
@QilongTang 
@zeusongit 

### FYIs

@minjie1980


QilongTang:
Reviewing now

tools/InstallerSpec/InstallerSpec/Program.cs
```diff
@@ -213,22 +216,35 @@ static void Main(string[] args)
...
                                     // contains fully qualified paths instead of doing the prefix here.
                                     // 
                                     var fileNameWithoutPath = streamReader.ReadLine().Trim();
+                                    //this is a comment, ignore this line
+                                    if (fileNameWithoutPath.StartsWith("##") || string.IsNullOrWhiteSpace(fileNameWithoutPath))
+                                    {
+                                        continue;
+                                    }
                                     var fileNameWithPath = Path.Combine(binpath, fileNameWithoutPath);
                                     if (!string.IsNullOrEmpty(fileNameWithPath))
```
QilongTang: Might be too much to ask.. Can we file a follow up task to include folder pattern? I would like to add .\libG* or something like that for future bullet proof.

QilongTang:
@mjkkirschner This LGTM but I would love to add one follow up to make it bullet proof for libG updates

tools/InstallerSpec/InstallerSpec/Program.cs
mjkkirschner: I looked into that, .net does not have any glob pattern support - they instead have a nuget package you can reference: 
https://www.nuget.org/packages/Microsoft.Extensions.FileSystemGlobbing/

I didn't want to add additional deps for this now.

QilongTang: @mjkkirschner Yes, I think this PR is good in the current state. Let's do the improvement later.

Issue #9227, opened 2018-11-05
Node to code crashing Dynamo


andydandy74:
## Dynamo version

2.0.1 stable

## Operating system

Windows 10

## What did you do? 

Ran node to code

## What did you expect to see?

A code block

## What did you see instead?

Dynamo crashing

Stack trace:
```
Object reference not set to an instance of an object.

   at Dynamo.Graph.Nodes.CodeBlockNodeModel.GetTypeHintForOutput(Int32 index)
   at Dynamo.Engine.NodeToCode.NodeToCodeCompiler.GetInputOutputMap(IEnumerable`1 nodes, Dictionary`2& inputMap, Dictionary`2& outputMap, Dictionary`2& renamingMap, Dictionary`2& typeHintMap)
   at Dynamo.Engine.NodeToCode.NodeToCodeCompiler.NodeToCode(Core core, IEnumerable`1 workspaceNodes, IEnumerable`1 nodes, INamingProvider namingProvider)
   at Dynamo.Graph.Workspaces.NodesToCodeExtensions.ConvertNodesToCodeInternal(WorkspaceModel workspace, EngineController engineController, INamingProvider namingProvider)
   at Dynamo.Models.DynamoModel.ConvertNodesToCodeImpl(ConvertNodesToCodeCommand command)
   at Dynamo.Models.DynamoModel.ExecuteCommand(RecordableCommand command)
   at MS.Internal.Commands.CommandHelpers.CriticalExecuteCommandSource(ICommandSource commandSource, Boolean userInitiated)
   at System.Windows.Controls.MenuItem.InvokeClickAfterRender(Object arg)
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
```

## Steps to reproduce
* Open Core_Math.dyn from the sample files that ship with Dynamo
* Select all 
* Run node to code

johnpierson:
Curious if this is related to this [8981](https://github.com/DynamoDS/Dynamo/issues/8981)

mjkkirschner:
tracked with: https://jira.autodesk.com/browse/DYN-2289

PR #10305, opened 2020-01-19
DYN-2289: Fix crash with NodeToCode

aparajit-pratap:
### Purpose

JIRA: https://jira.autodesk.com/browse/DYN-2289
GH Issue: https://github.com/DynamoDS/Dynamo/issues/9227

This fix prevents a crash when converting nodes to code in which at least one upstream node is a CBN that has static property invocations such as:
```
p = Point.ByCoordinates(0, 0, 0);
x = Point.X(p);
```
The reason for the crash was that the property was not found in the procedure table for the class. Since property getters internally compile to `get_propertyname(object)`, a method has been added that additionally checks for a static getter that accepts one argument (namely the class object), which in this example would be searching explicitly for `get_X(p)` instead of just `X`.  

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCore/Graph/Nodes/CodeBlockNode.cs
+++ b/src/DynamoCore/Graph/Nodes/CodeBlockNode.cs
@@ -536,7 +536,7 @@ namespace Dynamo.Graph.Nodes
             if (statement == null)
                 return type;
 
-            BinaryExpressionNode expr = statement.AstNode as BinaryExpressionNode;
+            var expr = statement.AstNode as BinaryExpressionNode;
             if (expr == null || expr.Optr != Operator.assign)
                 return type;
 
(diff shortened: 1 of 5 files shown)


src/DynamoCore/Graph/Nodes/CodeBlockNode.cs
```diff
@@ -554,15 +554,21 @@ public override ProtoCore.Type GetTypeHintForOutput(int index)
...
 
                 var targetClass = core.ClassTable.ClassNodes[classIndex];
-                var func = targetClass.GetFirstMemberFunctionBy(funcNode.Function.Name);
+                var funcName = funcNode.Function.Name;
+                var func = targetClass.GetFirstMemberFunctionBy(funcName);
+                if (func == null)
+                {
+                    func = targetClass.ProcTable.GetPropertyGetterByName(funcName);
+                }
                 type = func.ReturnType;
```
aparajit-pratap: The crash occurred here before this change as `func` would be `null`.

src/Engine/ProtoCore/ProcedureTable.cs
```diff
@@ -299,6 +299,19 @@ public IEnumerable<ProcedureNode> GetFunctionsByName(string name)
...
+        /// <summary>
+        /// Returns getter function for given property name.
+        /// Ex: for X property of Point, this method returns get_X(Point p).
+        /// </summary>
+        /// <param name="name"></param>
+        /// <returns></returns>
+        public ProcedureNode GetPropertyGetterByName(string name)
+        {
+            var getterMethodName = Constants.kGetterPrefix + name;
+            return Procedures.FirstOrDefault(p => p.Name == getterMethodName &&
```
mjkkirschner: does this potentially search the entire procedure table?
Is this class actually not a table????

aparajit-pratap: This searches only the procedure table for the particular class.

mjkkirschner: funny we call it a table...

aparajit-pratap: That is the typical terminology you will find in compilers.

QilongTang:
The famous crash got fix finally! LGTM

PR #10311, opened 2020-01-21
DYN-2328 Add test for geometry labeling

reddyashish:
### Purpose

This PR is to add a test for Geometry Labeling. 
JIRA: https://jira.autodesk.com/browse/DYN-2328

The test checks that there is are no labels in the geometry when the DisplayLabels flag is set to false and asserts for the correct number of labels when the DisplayLabels flag is set to true. 

### Declarations

### Reviewers

@mjkkirschner @QilongTang @aparajit-pratap @zeusongit

### DIFF
--- a/src/VisualizationTests/HelixWatch3DViewModelTests.cs
+++ b/src/VisualizationTests/HelixWatch3DViewModelTests.cs
@@ -758,9 +758,6 @@ namespace WpfVisualizationTests
             // We set the DisplayLabels to true to view the Labels in the preview geometry.
             codeBlockNodeModel.DisplayLabels = true;
 
-            helix = ViewModel.BackgroundPreviewViewModel as HelixWatch3DViewModel;
-            Model3DDictionary = helix.Model3DDictionary;
-
             // Now the Labels are shown in the preview geometry. 
             // The code block node has 64 points, so there should be 64 labels. 
             var geometryWithLabels = (helix.Model3DDictionary[labelKey] as GeometryModel3D).Geometry as BillboardText3D;
--- a/src/VisualizationTests/HelixWatch3DViewModelTests.cs
+++ b/src/VisualizationTests/HelixWatch3DViewModelTests.cs
@@ -731,6 +731,41 @@ namespace WpfVisualizationTests
 
             Assert.True(BackgroundPreviewGeometry.HasAnyMeshVerticesOfColor(new Color4(new Color3(1.0f, 0, 1.0f))));
         }
+       
+        [Test]
+        public void Display_Geometry_Labels()
+        {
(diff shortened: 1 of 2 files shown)


mjkkirschner:
@reddyashish nice - does this actually pass 😉 ? LGTM (if it does)

mjkkirschner:
I'm also curious if had a change to look into the labels that appear when the individual preview bubble items are selected?

 Ie - when list 0 is selected - x labels should be visible, when list 1 is selected y labels should be visible.

src/VisualizationTests/HelixWatch3DViewModelTests.cs
```diff
@@ -731,6 +731,41 @@ public void Display_ByGeometryColor_HasColoredMesh()
...
+            // By default the DisplayLabels for the code block node is set to false, 
+            // so the Model3DDictionary wouldn't have the geometry object corresponding to the Labels. 
+            var geometryHasLabels = helix.Model3DDictionary.ContainsKey(labelKey); 
+            Assert.IsFalse(geometryHasLabels);
+
+            // We set the DisplayLabels to true to view the Labels in the preview geometry.
+            codeBlockNodeModel.DisplayLabels = true;
+
+            helix = ViewModel.BackgroundPreviewViewModel as HelixWatch3DViewModel;
+            Model3DDictionary = helix.Model3DDictionary;
```
aparajit-pratap: Why do we need to assign these variables again? Would they not reflect the updated values after labels are turned on?

reddyashish: @aparajit-pratap yes, you are right. The helix object is updated when the labels are turned on and we do not need these assignments. Removing it.

reddyashish:
@mjkkirschner Yes, this test passes. I am looking into labels when an individual item in the preview bubble is selected.

src/VisualizationTests/HelixWatch3DViewModelTests.cs
```diff
@@ -258,7 +258,7 @@ public void Node_Labels_Render()
             OpenVisualizationTest("Labels.dyn");
 
             // check all the nodes and connectors are loaded
-            Assert.AreEqual(2, model.CurrentWorkspace.Nodes.Count());
+            Assert.AreEqual(3, model.CurrentWorkspace.Nodes.Count());
```
reddyashish: Since I added a watch node to this Labels.dyn, I had to update the node count in this test

reddyashish:
@mjkkirschner updated the test with the case we discussed. 

This PR is ready to be merged once you guys take a look.

mjkkirschner:
nice! LGTM

reddyashish:
Merging this.

Issue #9910, opened 2019-08-14
Crash when undoing in graph with Revit nodes in Sandbox


radumg:
## Dynamo version

2.0.2

## Operating system

Windows 10

## What did you do? 

Open graph that has Revit nodes in Dynamo Sandbox.
Move group on canvas
Undo move -> crash

## What did you expect to see?

No crash.

## What did you see instead?

Crash.
Happens only on this graph, trying to move the `get room boundaries` & `combine intersecting...` groups, then undo.

![image](https://user-images.githubusercontent.com/11439624/63049541-d6ef6900-bed0-11e9-8d68-5be423ba0e53.png)

## Stack trace

```
Object reference not set to an instance of an object.

   at Dynamo.Graph.Nodes.CodeBlockNodeModel.LoadAndCreateConnectors(OrderedDictionary inportConnections, OrderedDictionary outportConnections, SaveContext context)
   at Dynamo.Graph.Nodes.CodeBlockNodeModel.DeserializeCore(XmlElement nodeElement, SaveContext context)
   at Dynamo.Graph.Workspaces.WorkspaceModel.ReloadModel(XmlElement modelData)
   at Dynamo.Core.UndoRedoRecorder.UndoActionGroup(XmlElement actionGroup)
   at Dynamo.Graph.Workspaces.WorkspaceModel.Undo()
   at Dynamo.Models.DynamoModel.ExecuteCommand(RecordableCommand command)
   at System.Windows.Input.CommandManager.TranslateInput(IInputElement targetElement, InputEventArgs inputEventArgs)
   at System.Windows.UIElement.OnKeyDownThunk(Object sender, KeyEventArgs e)
   at System.Windows.RoutedEventArgs.InvokeHandler(Delegate handler, Object target)
   at System.Windows.RoutedEventHandlerInfo.InvokeHandler(Object target, RoutedEventArgs routedEventArgs)
   at System.Windows.EventRoute.InvokeHandlersImpl(Object source, RoutedEventArgs args, Boolean reRaised)
   at System.Windows.UIElement.RaiseEventImpl(DependencyObject sender, RoutedEventArgs args)
   at System.Windows.UIElement.RaiseTrustedEvent(RoutedEventArgs args)
   at System.Windows.Input.InputManager.ProcessStagingArea()
   at System.Windows.Input.InputManager.ProcessInput(InputEventArgs input)
   at System.Windows.Input.InputProviderSite.ReportInput(InputReport inputReport)
[...]

PR #7301, opened 2016-10-26
Remove InPortData and OutPortData

ikeough:
### Purpose

This PR represents a substantial refactoring of ports and connectors.

This PR removes the `InPortData` and `OutPortData` collections on `NodeModel`. These collections pre-date the `InPorts` and `OutPorts` collections, and were previously deprecated. They acted as a way to store initialization data for ports during the construction of the node, with the assumption that at some point after they were filled, `NodeModel.RegisterAllPorts` would be called, subsequently creating/updating ports on the nodes. This indirection was not obvious or useful and led to the `InportData` and `OutportData` collections getting out of sync with the `InPorts` and `OutPorts` collections. 

In the beginning, collections in Dynamo were not observable, so we created a lot of events like `NodeModel.PortConnected` and `NodeModel.ConnectorConnected` when ports and connectors were added. As we have moved to using observable collections, it's much easier to deal with the creation and deletion of ports, connectors, nodes, etc., by simply handling the collection changed events for those collections and doing the right thing. This PR gets us part of the way towards a fully observable methodology by adjusting `NodeModel` to watch the collection changed events for the `InPorts` and `OutPorts` collections, and consolidating logic previously spread among numerous event handlers into those collection change handlers.

Additionally, this PR removes a number of methods which seemed like a good idea three years ago, but now only add confusion. For example. `PortModel.Connect` was a very thin wrapper around `PortModel.Connectors.Add(someNewConnector)`. These thin wrappers have been removed except in cases where the wrapper contained some assertion logic.

### Declarations

Check these if you believe they are true
- [x] The code base is in a better state after this PR
- [x] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [x] The level of testing this PR includes is appropriate
- [x] User facing strings, if any, are extracted into `*.resx` files
- [x] All tests pass using the self-service CI. (In progress)
- [x] Snapshot of UI changes, if any.
- [x] Follows [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) standards (_No change of public methods or types etc.._)
### Reviewers

@mjkkirschner

PR #10315, opened 2020-01-23
DYN-2286: fix for crash upon unresolved node undo

aparajit-pratap:
### Purpose

JIRA: https://jira.autodesk.com/browse/DYN-2286
GH: https://github.com/DynamoDS/Dynamo/issues/9910

New `PortModel`s are being constructed each time a dummy (unresolved) node is deserialized for undo. The index of a portmodel is obtained from the original observable collection, however, since there was no `Equals` override for a `PortModel` a default reference equals comparison would take place when querying its index from the collection. The reference equals check would fail for a new portmodel and consequently, the port model index was no longer found and would be returned as `-1`. When a code block (connected as input or output node to the dummy node) is recreated after an undo, it tries to recreate its connector to the dummy node's port model with an invalid index, which resulted in the crash.

The fix is to recreate the new ports using the original guids of the old ones in `DummyNode.Deserialize` so that we can search for the new ports using their guids (which should match with the original ones) in case the reference equality fails. In order to achieve this the `Equals` method has been added for `PortModel` to make these additional checks.

Caveat: I haven't tried to fully understand the change made by Ian in this PR (which I think should be the time since this crash started happening): https://github.com/DynamoDS/Dynamo/pull/7301. 

### Declarations

### Reviewers

@mjkkirschner 
@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCore/Graph/Nodes/DummyNode.cs
+++ b/src/DynamoCore/Graph/Nodes/DummyNode.cs
@@ -10,6 +10,7 @@ using Newtonsoft.Json.Linq;
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
 using System.Xml;
 
 namespace Dynamo.Graph.Nodes
@@ -250,11 +251,10 @@ namespace Dynamo.Graph.Nodes
             OutputCount = Int32.Parse(outputCount.Value);
             LegacyNodeName = legacyName.Value;
 
-            if (nodeElement.ChildNodes != null)
+            foreach (XmlNode childNode in nodeElement.ChildNodes)
             {
-                foreach (XmlNode childNode in nodeElement.ChildNodes)
-                    if (childNode.Name.Equals("OriginalNodeContent"))
-                        OriginalNodeContent = (XmlElement)nodeElement.FirstChild.FirstChild;
+                if (childNode.Name.Equals("OriginalNodeContent"))
+                    OriginalNodeContent = (XmlElement) nodeElement.FirstChild.FirstChild;
             }
 
             if (originalElement != null)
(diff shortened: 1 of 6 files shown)


aparajit-pratap:
Thanks @mjkkirschner for the help.

src/DynamoCore/Graph/Nodes/PortModel.cs
```diff
@@ -369,6 +369,23 @@ protected override void DeserializeCore(XmlElement nodeElement, SaveContext cont
         }
 
         #endregion
+
+        public bool Equals(PortModel other)
```
mjkkirschner: last time I tried overriding `equals` there were a bunch of unexpected failures where reference equality was expected throughout the codebase instead of equality by guid - (I can't remember what type this was on) - just a consideration.

aparajit-pratap: In this case, I'm checking for guid equality only if reference equality fails. From what I see in other nodemodel types, it looks like the equality check will either be true or false in both cases (reference equality + guid equality) since if the same port model is being reused, both conditions will be true, and if a new port model is created, still both conditions will be false, unless the new ports are created using the same guids as the old ones, which is what I'm enforcing here explicitly for dummy nodes.

mjkkirschner:
LGTM

src/DynamoCore/Graph/Nodes/DummyNode.cs
```diff
@@ -250,11 +251,10 @@ private void LoadNode(XmlNode nodeElement)
             OutputCount = Int32.Parse(outputCount.Value);
             LegacyNodeName = legacyName.Value;
 
-            if (nodeElement.ChildNodes != null)
+            foreach (XmlNode childNode in nodeElement.ChildNodes)
```
mjkkirschner: if childNodes is null - this will throw an exception - I think you should keep the if

mjkkirschner: https://stackoverflow.com/questions/3088147/why-does-net-foreach-loop-throw-nullrefexception-when-collection-is-null

aparajit-pratap: Resharper tells me that expression will always be true.

PR #10316, opened 2020-01-23
Fix the library view display issues introduced by MS WebBrowser

CoolStarLee:
### Purpose

The following library view display issues introduced by MS WebBrowser are fixed:

1. Disable the context menu on library view.
2. Zoom the library view to adapt the current DPI scale.
3. Disable the ability about zooming library view by keyboard or mouse wheel.
4. Hidden the extra close button on search bar.

Add CSS and JS code to Library.html to solve the above issues.

### Declarations

### Reviewers

@mjkkirschner 
@DynamoDS/dynamo

### DIFF
--- a/src/LibraryViewExtensionMSWebBrowser/LibraryViewController.cs
+++ b/src/LibraryViewExtensionMSWebBrowser/LibraryViewController.cs
@@ -69,7 +69,6 @@ namespace Dynamo.LibraryViewExtensionMSWebBrowser
 
         private void scriptNotifyHandler(string dataFromjs)
         {
-
             if (string.IsNullOrEmpty(dataFromjs))
             {
                 return;
@@ -231,6 +230,7 @@ namespace Dynamo.LibraryViewExtensionMSWebBrowser
                 dynamoViewModel.ImportLibraryCommand.Execute(null)
             ));
         }
+
         /// <summary>
         /// This function logs events to instrumentation if it matches a set of known events
         /// </summary>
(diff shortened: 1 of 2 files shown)


src/LibraryViewExtensionMSWebBrowser/LibraryViewController.cs
```diff
@@ -373,9 +368,14 @@ internal void AddLibraryView()
             browser.Loaded += Browser_Loaded;
             browser.SizeChanged += Browser_SizeChanged;
             LibraryViewController.SetupSearchModelEventsObserver(browser, dynamoViewModel.Model.SearchModel, this, this.customization);
-        }
 
+            browser.DpiChanged += Browser_DpiChanged;
```
mjkkirschner: I think it would be a good idea to unsubscribe this event handler in the `Dispose` method below.

src/LibraryViewExtensionMSWebBrowser/web/library/library.html
```diff
@@ -12,6 +12,9 @@
             margin: 0;
             background-color: #353535;
         }
+        input::-ms-clear {
```
mjkkirschner: can we add a comment describing this?

src/LibraryViewExtensionMSWebBrowser/web/library/library.html
```diff
@@ -27,6 +30,37 @@
         var replaceImageDelayTime = 100;
         var currentImage;
 
+        // Disable the context menu
+        document.oncontextmenu = function () {
+            return false;
```
mjkkirschner: nice 👍

src/LibraryViewExtensionMSWebBrowser/web/library/library.html
```diff
@@ -129,6 +162,37 @@
...
+
+        function adaptDPI() {
+            document.body.style.overflowX = 'hidden';
+
+            var dpiScale = getDPIScale();
+
+            var libraryHeader = document.getElementsByClassName("LibraryHeader")[0];
+            libraryHeader.style.zoom = dpiScale;
+            
+            var widthPercentage = (100.0 / dpiScale).toString() + '%';
```
mjkkirschner: this solution is perfectly fine to me - I think alternatively we could wrap this text in another div and scale it alone, and not scale the libraryHeader.

https://github.com/DynamoDS/librarie.js/blob/master/src/components/SearchBar.tsx#L256

But I actually like your solution more as we can isolate the changes from the librarie.js component.

QilongTang:
@CoolStarLee We will merge your PR now and please see follow up work from our team by @reddyashish in https://github.com/DynamoDS/Dynamo/pull/10335

PR #10343, opened 2020-01-31
Documentation browser

radumg:
### Purpose

Adds a `DocumentationBrowser` view extension to Dynamo's sidebar that can display HTML files. The main use-case enabled is showing additional information for errors (similar to `Dynamo Dictionary`) in-product, making Dynamo more accessible to novices & experienced users alike.

The `DocumentationBrowser` is invoked when clicking the `Read more` link on node info bubbles that have `warnings` or `errors`. 

The improvements also allow zero-touch developers to display their own help documentation, which hopefully will mean less googling around for users encountering errors.

#### Enriching errors with documentation is opt-in
The mechanisms added are opt-in/value-add, meaning current errors & behaviour remain unchanged. However, you can now enrich an error simply by adding a path to a documentation file  to the end of the exception message.

1-no exception

#### Local documentation files
The docs browser will happily load local (on user's computer) HTML files as documentation.
2-local documentation

#### Handle missing documentation files
When a doc file being requested is missing, we display a specific error page.

3-missing docs

#### Remote resources
When the documentation being requested is `remote` (web address or network share), we display a `warning` banner at the bottom.
Note the slowness in loading is due to the dictionary website itself.

4- remote docs

Any links that would cause the page to navigate to a new URL are opened in a new window in the OS default browser instead of allowing navigation inside the `DocumentationBrowser`. Note that we cannot limit JS-based content changes/navigation, such as for SPAs, unless they push a URL change.

#### Empty content
When the documentation content is found but it's an empty file/page, we handle it with a specific error page. This also addresses the situation when you open the `DocumentationBrowser` but haven't yet clicked on any `InfoBubble` link.

5- no content

#### Zero-touch support
The changes also allow zero-touch developers to specify their own documentation file, so the wider Dynamo community can fully and easily take advantage of this. Package authors could ship their own help `html` files within the `extra` folder and raise `Exception`s pointing to them.

6-zero touch with error

The code for that ZT node is
```csharp
public static class Throw
    {
        public static void WithText(string text)
        {
            throw new Exception(text);
        }
    }
```

#### Zero-touch backwards compatible
None of the existing ZT behaviours are broken or altered.
7-zero touch without error

### Declarations

### Reviewers

@mjkkirschner , @QilongTang , @Amoursol 

### FYIs

@Dewb 

### Dynamo Core changes
I've tried to keep changes to Core
- small & contained
- backwards-compatible as much as possible
- in line with existing patterns in the codebase

We're very open to any suggestions on whether we're hooking into the right places, causing any breakages we might have not thought of or if there's a generally better approach we could look into.

For example, we technically don't require the `Command` that's been added & exposed to make this work, but felt that it better fits in with the patterns in the codebase & making it easier for third-party developers to leverage this new capability. Happy to pull this out if you think it's not adding value, etc.

#### Enriching errors with documentation file paths
To enrich an exception with a documentation file, add a specific identifer to the `Message` property of the `Exception` being throw. The `href=` identifier is used to recognise messages with enriched documentation, an example exception message could be `Objects cannot be null href=NullReference.html`. Only this identifier is accepted and everything after it is considered to be part of the path/Uri to the documentation file.
All Dynamo exception messages are in `.resx` files, so adding/changing the `href=` identifier needs to other code changes. Autodesk localisation team has also confirmed to @mjkkirschner this will not be an issue for localisation.

#### InfoBubble is now link-aware
We've chosen to handle link-enriched exceptions at this level to avoid touching the VM or engine.

The `InfoBubbleDataPacket` class has a new `Uri` property called `Link`. In its constructor, it looks for the `href=` identifier in the incoming message text and if it finds it, it attempts to extract a `Uri` and removes the link from the message text itself. It silently ignores any parsing errors and sets `Link` to null on failure.
To support interactions with the info bubble UI, we've added an event & handler to the `InfoBubbleViewModel`. When the viewmodel of an `InfoBubble` encounters a non-null `Link`, the bubble renders and additional `Read more in the documentation` clickable link. When clicked, this link further raises events in the Dynamo `ViewModel`.

#### OpenDocumentationLink events
A `RequestOpenDocumentationLink` event & `RequestOpenDocumentationLinkHandler` handler has been added to the Dynamo `ViewModel` to handle requests from whatever calling component. If there is no registered handler, links will open with a `Process.Start` in the user's default OS browser.

#### OpenDocumentationLink commands
An `OpenDocumentationLinkCommand` command has also been added to the `DynamoViewModel` and further exposed in the new `ViewModelCommandExecutive`, with a typed input parameter instead of `object`. Chose to do this as it is accessible from extensions and other third-party code without needing a reference to the ViewModel itself.

#### ViewLoadedParams exposes new event
To allow the documentation extension to register to the relevant event, we've exposed a new `RequestOpenDocumentationLink` event on the `ViewLoadedParams` class. The implementation however is a proxy, so anyone (de)registering to this event is actually interacting with the underlying `DynamoViewModel` event. This is to avoid having `ViewLoadedParams` instances counting as handlers for the event.

### DocumentationBrowser view extension

The view extension does a very simple job 
- register itself as a handler for the `RequestOpenDocumentationLink` event
- instantiate a `ViewModel` and `View`
- forward events to the `ViewModel`
- inject the `View` into the sidebar whenever there's an event

#### The ViewModel
The ViewModel of the DocumentationBrowser does a few things 
- accepts the incoming event notification
- determines whether link is local (file on PC) or remote (on network or web address)
- tries to resolve path of file given & searches in built-in `Docs` folder too
- tries to load the content from the Uri specified
- falls back on `no content`, `error` or `missing` embedded pages when required
- ultimately loads the content of the documentation file into memory as a string, to avoid IO on view layer & keep Dynamo responsive
- raises a `LinkChanged` event when ready, so View can pick it up

#### The View
To render the documentation files, the view has an embedded web browser. 
Things to note here
- we're not using `CefSharp`, but it could be swapped since ViewModel does the heavy lifting
- we're using  the [`WebViewCompatible` control](https://docs.microsoft.com/en-us/windows/communitytoolkit/controls/wpf-winforms/webviewcompatible) which uses the `Edge` engine on modern Windows 10 and falls back on IE engine on older OS-es.
- for remote resources (web addresses or network shares), a banner is displayed at the bottom.

## Known bugs

~~We're currently tackling a bug~~. this is now fixed ✅ 

which sees the web browser control not inherit the visibility from its parent, so it remains `visible` when the start page is displayed. This is because closing the sidebar doesn't `dispose` of the control and uses some internal `magic` to hide things with a zero width. 
 
image

## Leftover tasks
Added the `DNM` tag pending review and until the below tasks are completed.

- [x] Add unit tests

Please note I've not yet included any unit tests - once @mjkkirschner signs off on the general approach (especially around `Core` changes), I'll start adding `unit tests` too.

- [x] Add the `href=` link to all documentated errors in `.resx`

Look forward to feedback on this one, been a while coming! 🙌


mjkkirschner:
@radumg whats up with these build failures:

```
error parsing solution file at E:\Builds\Jenkins\Dynamo_master\Dynamo\src\Dynamo.All.sln: Exception has been thrown by the target of an invocation.  Error parsing the nested project section in solution file. A project with the GUID "{1A5DC90A-477E-4D4A-87BD-0BFB01F056CE}" is listed as being nested under project "{88D45B00-E564-41DB-B57C-9509646CAA49}", but does not exist in the solution.  E:\Builds\Jenkins\Dynamo_master\Dynamo\src\Dynamo.All.sln
```

src/DocumentationBrowserViewExtension/DocumentationBrowserView.xaml
```diff
@@ -0,0 +1,44 @@
...
+             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
+             xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+             xmlns:p="clr-namespace:Dynamo.Wpf.Properties;assembly=DynamoCoreWpf"
+             xmlns:fa="http://schemas.fontawesome.io/icons/"
+             xmlns:w="clr-namespace:Dynamo.DocumentationBrowser.Properties"
+             xmlns:ui="clr-namespace:Dynamo.UI;assembly=DynamoCoreWpf"
+        
+            xmlns:WPF="clr-namespace:Microsoft.Toolkit.Wpf.UI.Controls;assembly=Microsoft.Toolkit.Wpf.UI.Controls.WebView"
```
mjkkirschner: can you verify you tested this in Revit? My experience is that this control does not work in Civil3d Dynamo, I'm guessing Revit is similar, but unfortunately we need something that works everywhere, even if it does work in Revit.

I have found that MSwebBrower in `system.windows.controls` works well and can be forced to display in IE11 mode which has pretty good support - it displays librariejs without issue so it should display these static html pages fine. 

In any case, I understand you don't have time to change this up which is fine, I am just wondering if you tested it in Revit or Civil at this point.

src/DocumentationBrowserViewExtension/DocumentationBrowserView.xaml.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            // if non of the above is true cancel the navigation 
+            // and redirect it to a new process that starts the browser
+            e.Cancel = true;
+            Process.Start(new ProcessStartInfo(e.Uri.AbsoluteUri));
+        }
+
+        public void NavigateToPage(Uri link)
+        {
+            if (this.viewModel.IsRemoteResource)
```
mjkkirschner: is there some way of configuring the extension to completely disallow this? I would like to be able to disable remote resource loading entirely.

src/DocumentationBrowserViewExtension/DocumentationBrowserView.xaml.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            InitializeComponent();
+            this.DataContext = viewModel;
+            this.viewModel = viewModel;
+
+            // subscribe to the link changed event on the view model
+            // so we know when to navigate to a new documentation page/document
+            viewModel.LinkChanged += NavigateToPage;
+
+            // handle brower component events & disable certain features that are not needed
```
mjkkirschner: sp(browser)

src/DocumentationBrowserViewExtension/DocumentationBrowserView.xaml.cs
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        /// <summary>
+        /// Dispose function for DocumentationBrowser
+        /// </summary>
+        public void Dispose()
+        {
+            this.viewModel.LinkChanged -= NavigateToPage;
+            this.documentationBrowser.Dispose();
+            GC.SuppressFinalize(this);
```
mjkkirschner: why do this?

src/DocumentationBrowserViewExtension/DocumentationBrowserView.xaml.cs
```diff
@@ -0,0 +1,84 @@
...
+            }
+
+            this.RemoteLinkBanner.Visibility = Visibility.Collapsed;
+            this.documentationBrowser.NavigateToString(this.viewModel.GetContent());
+        }
+
+        /// <summary>
+        /// Dispose function for DocumentationBrowser
+        /// </summary>
+        public void Dispose()
```
mjkkirschner: do you call this from somewhere? WPF controls do not get disposed automatically.

mjkkirschner: ahh, I see, it gets called from the extension dispose. I don't see a reason to suppress its garbage collection?

mjkkirschner: You would usually do that if you were implementing a finalizer and the dispose method might also be called.

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModelEventArgs.cs
```diff
@@ -126,4 +126,13 @@ public void SetIcon(ImageSource icon)
             Icon = icon;
         }
     }
+
+    public class OpenDocumentationLinkEventArgs : EventArgs
```
mjkkirschner: /// comment please

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModelEvents.cs
```diff
@@ -114,6 +115,22 @@ public virtual void OnRequestShowHideGallery(bool showGallery)
             }
         }
 
+        public event RequestOpenDocumentationLinkHandler RequestOpenDocumentationLink;
+        public virtual void OnRequestOpenDocumentationLink(OpenDocumentationLinkEventArgs e)
+        {
+            // let any registered documentation link handlers process the request
+            if (RequestOpenDocumentationLink != null) RequestOpenDocumentationLink(e);
+            // if no handlers are registered, we fall back to handling the link with the default operating system handler
+            else Process.Start(new ProcessStartInfo(e.Link.AbsoluteUri));
```
mjkkirschner: hmm, I'm not sure about this seems easy to abuse.

src/DynamoCoreWpf/ViewModels/Preview/InfoBubbleViewModel.cs
```diff
@@ -161,8 +169,10 @@ private void UpdateInfoBubbleContent(object parameter)
             InfoBubbleStyle = data.Style;
             ConnectingDirection = data.ConnectingDirection;
             UpdateContent(data.Text);
+            this.DocumentationLink = data.Link;
             TargetTopLeft = data.TopLeft;
             TargetBotRight = data.BotRight;
+            DocumentationLink = data.Link;
```
mjkkirschner: doing this twice?

src/DynamoCoreWpf/ViewModels/Preview/InfoBubbleViewModel.cs
```diff
@@ -275,27 +297,61 @@ private void GenerateContent()
...
             TopLeft = topLeft;
             BotRight = botRight;
+            Link = CheckIfMessageHasHrefLinkAndExtract(ref text);
             Text = text;
             ConnectingDirection = connectingDirection;
         }
+
+
+        //Check if has link
+        private static Uri CheckIfMessageHasHrefLinkAndExtract(ref string text)
```
mjkkirschner: why are you using a ref? So you can modify the text? ... advise considering if this is the best design, I see that even after modifying the text to remove the link the url creation can still fail... so thats a bit weird.

might make more sense to split into two functions and return a bool from the first one - ie `CheckIfMessageHasHref` and `CreateURIFromMessageText`

src/DynamoCoreWpf/Views/Preview/InfoBubbleView.xaml.cs
```diff
@@ -477,6 +492,36 @@ private TextBox GetNewTextBox(string text)
...
+                Background = Brushes.Transparent,
+                HorizontalAlignment = HorizontalAlignment.Center,
+                VerticalAlignment = VerticalAlignment.Center,
+
+                FontFamily = font as FontFamily
+            };
+
+            Hyperlink link = new Hyperlink();
+            link.NavigateUri = uri;
+            link.RequestNavigate += RequestNavigateToDocumentationLinkHandler;
```
mjkkirschner: can this be unsubscribed somewhere? I always eye handlers like this suspiciously - it's very difficult to tell if this will introduce a memory leak.

radumg:
> @radumg whats up with these build failures:
> 
> ```
> error parsing solution file at E:\Builds\Jenkins\Dynamo_master\Dynamo\src\Dynamo.All.sln: Exception has been thrown by the target of an invocation.  Error parsing the nested project section in solution file. A project with the GUID "{1A5DC90A-477E-4D4A-87BD-0BFB01F056CE}" is listed as being nested under project "{88D45B00-E564-41DB-B57C-9509646CAA49}", but does not exist in the solution.  E:\Builds\Jenkins\Dynamo_master\Dynamo\src\Dynamo.All.sln
> ```

Hey @mjkkirschner , sorry, that's my bad, picked up this PR after it was sat for a while and forgot some of its history when resolving conflicts. I'll clean it up now.

src/DocumentationBrowserViewExtension/DocumentationBrowserView.xaml.cs
radumg: you're right, not needed, this was a leftover from some copy-pasting early on (see `LibraryViewController.cs`). Sorry for this one being a bit messy, was exploring 3 different avenues at one point and combined them back into 1 a bit sloppily.

src/DynamoCoreWpf/Views/Preview/InfoBubbleView.xaml.cs
radumg: done in a2b65f28740b79974ddcb5542102bb10282a7666
debugged and confirmed the viewmodel of the `InfoBubble` gets disposed when 
- the node is deleted
- the workspace is closed/cleared

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModelEvents.cs
radumg: removed the fallback, now silently ignored if no handler is registered. also left code comment around this decision for future
90ca3425ce8e6b6e67134246ca58327900367dc3

src/DynamoCoreWpf/ViewModels/Preview/InfoBubbleViewModel.cs
radumg: the link creation being able to fail after sanitising the text was on purpose, cause you could pass in `text href=` and the text is fine but there's no link to parse. Was trying to avoid duplicate work but agree it's not cleanest design, have split them up now, with the caveat that it's slightly less efficient.

src/DocumentationBrowserViewExtension/DocumentationBrowserView.xaml.cs
```diff
@@ -0,0 +1,83 @@
...
+            this.viewModel = viewModel;
+
+            // subscribe to the link changed event on the view model
+            // so we know when to navigate to a new documentation page/document
+            viewModel.LinkChanged += NavigateToPage;
+
+            // handle browser component events & disable certain features that are not needed
+            this.documentationBrowser.AllowDrop = false;
+            this.documentationBrowser.NavigationStarting += NavigationStarting;
+            this.documentationBrowser.NavigationCompleted += NavigationCompleted;
```
QilongTang: Does the dispose function of `documentationBrowser` take care of ubsubscribing?

Dewb:
This is great!

Will right-click > `Help...` help content also show up in this dockable pane?

The "no content available or no error selected" page looks more catastrophic than it ought to be. As cute as it is, I don't think we should use the band-aid crash reporting icon outside of a crash-reporting context (and maybe not even then.) A more user-friendly way to handle this would be to show an index/landing/top tips page, and put a banner on the top with the error message. That's a lot of screen real estate to take up for a small amount of information.

radumg:
> This is great!
Thanks!

> Will right-click > `Help...` help content also show up in this dockable pane?
Not at the minute as that's [a WPF page](https://github.com/DynamoDS/Dynamo/blob/523e71b893d2d85f2457550860239ab3ca3ce2e5/src/DynamoCoreWpf/UI/Prompts/NodeHelpPrompt.xaml) but it was designed to be able to accommodate that future use too. Sadly out of scope for this particular run.

> The "no content available or no error selected" page looks more catastrophic than it ought to be. 
Agree, not particularly proud of myself with that layout 😂, will try to improve it a bit but won't go too far as the page designs are bare-bones on purpose, had agreed with @Amoursol the DYN team would do further iteration on the design (as part of other efforts I think?).

> As cute as it is, I don't think we should use the band-aid crash reporting icon outside of a crash-reporting context (and maybe not even then.) A more user-friendly way to handle this would be to show an index/landing/top tips page, and put a banner on the top with the error message. That's a lot of screen real estate to take up for a small amount of information.
`CrashyFace` (had to name it😎) is a placeholder, can replace with a NodeFace not crash-related in the absence of other icons, or with a simple `?` icon.

> "Read more..." might be more idiomatic than "Read more in the documentation."
👍 Will change.

src/DocumentationBrowserViewExtension/DocumentationBrowserView.xaml.cs
radumg: It probably doesn't, so added unsubscribing to the dispose of the view
b8646480ec7148f85ed7e4c21eec1f8094d6b41e

src/DocumentationBrowserViewExtension/DocumentationBrowserView.xaml.cs
radumg: Added a setting at extension level which defaults to false, so it ignores events that target remote resources now. In the future, when this is tied to a setting in Dynamo or maybe simply to another menu item (`Allow remote resources`), the view model is ready to handle that as per original gifs above.
9f828394a1a8688c8cef4d063f9923b9f6504748

radumg:
Made the `No content` page be a bit more like a landing page as suggested by @Dewb and as such, only error pages are now stylised as centre-aligned (vertical & text). Also unified the padding between the landing page and content pages so it breathes a bit better.

Landing page now

image

Documentation pages look like this
image

src/DocumentationBrowserViewExtension/DocumentationBrowserView.xaml
radumg: I've now switched the browser to use the `System.Windows.Controls.WebBrowser` control, @shknudsen will also be testing with `DynamoRevit` tomorrow.

radumg:
quick update on where this is @mjkkirschner 

- visibility bug is still happening, even with the replaced MS browser control
- I've noticed a bug that on first launch of an error doc the browser window is white. Can't pin it down as our events fire in the right sequence, etc, might have something to do with being added to sidebar the first time round ? clicking a `read more` link again subsequently has no issues.

mjkkirschner:
https://github.com/DynamoDS/Dynamo/blob/master/src/LibraryViewExtensionMSWebBrowser/ViewExtension.cs#L67


this works for hiding the browser in my experience.

src/DocumentationBrowserViewExtension/DocumentationBrowserView.xaml
SHKnudsen: @mjkkirschner @radumg 

test

Everything seems to be working as expected on D4R, except for the white screen bug that happens the first time you open the extension.

radumg:
> https://github.com/DynamoDS/Dynamo/blob/master/src/LibraryViewExtensionMSWebBrowser/ViewExtension.cs#L67
> this works for hiding the browser in my experience.

oh man, wish you had dug out this nugget a while back, didn't know about the airspace problem. This did the trick indeed 💯

test/DynamoCoreWpfTests/InfoBubbleTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            var docEventArgs = new OpenDocumentationLinkEventArgs(infoBubbleDataPacket.Link);
+
+            // register an extra handler for documentation events so we can confirm
+            // the Dynamo ViewModel event gets triggered from the InfoBubble viewmodel
+            RequestOpenDocumentationLinkHandler handler = (OpenDocumentationLinkEventArgs e) =>
+            {
+                Assert.IsNotNull(e);
+                Assert.AreEqual(documentationFile, e.Link.ToString());
+            };
+            this.ViewModel.RequestOpenDocumentationLink += handler;
```
mjkkirschner: please unsubscribe this after the test.

radumg:
Added tests for the `DocumentationBrowserViewExtension` itself.

image

test/DynamoCoreWpfTests/InfoBubbleTests.cs
radumg: 96e73795a3aceae6871b26f0cd828b09d5d747dc

src/DocumentationBrowserViewExtension/Docs/ArgumentNullException.html
```diff
@@ -0,0 +1,54 @@
+<!-- 
+    There are two versions of this HTML file. One that is part of the Dynamo Dictionary 
+    codebase and one that is part of the Dynamo codebase, used for displaying additional
+    error information in the Dynamo side panel.  
+    
+    These files need to kept in sync manually: if making edits to one version of the file,
+    make sure to update the other version as well.
+
+    Dynamo: https://github.com/DynamoDS/Dynamo
+    Dynamo Dictionary: https://github.com/DynamoDS/DynamoDictionary_React
```
mmisol: Hi @radumg, sorry to bother you. I was taking a look at finishing up this PR and had a question. About this comment, I checked the linked repo `DynamoDictionary_React` and could not find where the other version of this file is. Could you point me to where these can be found in that repo? Thanks

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
```diff
@@ -36,7 +36,10 @@ public class DocumentationBrowserViewExtension : IViewExtension, ILogSource
         /// </summary>
         public bool AllowRemoteResources
         {
-            get => this.allowRemoteResources;
+            get
+            {
+                return this.allowRemoteResources;
+            }
             set
```
mjkkirschner: hmm, what about making this internal?

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
```diff
@@ -52,12 +52,29 @@ public bool IsRemoteResource
...
         }
-        public bool isRemoteResource;
+        private bool isRemoteResource;
 
 
         internal Action<Uri> LinkChanged;
         private void OnLinkChanged(Uri link) => this.LinkChanged?.Invoke(link);
 
+        private bool showBrowser;
+        public bool ShowBrowser
```
mjkkirschner: for a public property would you mind adding /// comments

src/DocumentationBrowserViewExtension/Docs/ArgumentNullException.html
mjkkirschner: @mmisol looks like they might still be here and not merged to dictioanry - @Amoursol - can you remind me what that final state was supposed to be? Did we decide to only have them in Dynamo:
https://github.com/horatiubota/DynamoDictionary_React/tree/dynamo_error_messages

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
mmisol: It is used in the tests. Are you suggesting to use [InternalsVisibleTo](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute?view=netframework-4.8)?

Maybe I'm missing something, but I don't see a way to enable this in normal usage. Won't this be always disabled?

mjkkirschner: yes, I don't think we should encourage any extensions authors to enable this, potentially removing it completely.

mmisol: Ok. I'll remove the feature then.

mjkkirschner: well, open to your thoughts on it, want to proceed carefully with potential redirection to external links.

mmisol: I have no problem with it :) I actually agree it's the safest approach. There is the security aspect of it but also with embedded HTML you can always get help, even if offline. We could say we are enforcing a good practice and people can always add links to additional online resources, if needed.

src/DocumentationBrowserViewExtension/Docs/ArgumentNullException.html
Amoursol: @mjkkirschner @mmisol - Final state was to have on both Dynamo Dictionary and inside DynamoBIM. Desire was to wrap the Dictionary into Dynamo itself further down the line.

radumg: Also note we've had to ever-so-slightly tweak the `.html` docs in the DynamoCore repo because of using the IE browser, adding a header tag and some styling.

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
```diff
@@ -199,40 +208,34 @@ internal string GetContent()
...
-        {
-            var path = ResolveFilePath(link);
-            this.content = File.ReadAllText(path);
-
-            // when we have no content fall back to the no content page
-            if (string.IsNullOrWhiteSpace(this.content))
-                throw new TargetException();
-        }
-
         private string LoadContentFromResources(string name)
```
mjkkirschner: can you add a sample of what a valid resource uri looks like?

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
```diff
@@ -199,40 +208,34 @@ internal string GetContent()
...
+            if (assemblyIndex != -1)
+            {
+                var assemblyName = name.Substring(0, assemblyIndex);
+                // Ignore version and public key, in case they were specified.
+                var versionIndex = assemblyName.IndexOf(";");
+                if (versionIndex != -1)
+                {
+                    assemblyName = assemblyName.Substring(0, versionIndex);
+                }
+                assembly = Assembly.Load(assemblyName);
```
mjkkirschner: hmm, I think we should avoid adding more locations where we load assemblies...
If this is a zero touch node, the assembly should already be loaded, what about if limit this to searching the loaded assemblies in the current app domain - this will require that node authors add their documentation to the same assembly as their nodes.

@aparajit-pratap FYI.

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
```diff
@@ -245,7 +248,9 @@ private string LoadContentFromResources(string name)
             {
                 result = reader.ReadToEnd();
             }
-            return result;
+
+            // Clean up possible script tags from document
+            return Regex.Replace(result, SCRIPT_TAG_REGEX, "");
```
mjkkirschner: I'm curious if you think logging any removals to the console would be good? I can see some developers getting really confused.

mjkkirschner:
@mmisol looks good, I just have one more assembly loading constraint I would like you to consider - the other issues can be ignored if you think they are not a problem.

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
mmisol: Sure. You can also take a look at the tests for examples.

Basically you have two cases:
- No assembly specified: `href=EmbeddedInDefaultAssembly.html`
- Assembly specified: `href=MyAsssembly;EmbeddedInMyAssembly.html`

The "default assembly" is the documentation browser extension itself.

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
mmisol: You are right. I think I tried doing it that way but I found some issues with it. Will try that again.

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
mmisol: I'm not sure I understand what you mean by logging to console, but thinking about it, maybe we could replace the tag by some content instead of "". Something like `SCRIPT TAG REMOVED` maybe?

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
```diff
@@ -56,6 +56,11 @@ public void Loaded(ViewLoadedParams viewLoadedParams)
         {
             if (viewLoadedParams == null) throw new ArgumentNullException(nameof(viewLoadedParams));
 
+            this.ViewModel.MessageLogged += (msg) =>
```
mjkkirschner: can we unsubscribe this handler in the dispose method?

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
```diff
@@ -255,9 +258,17 @@ private string LoadContentFromResources(string name)
...
+            if (Regex.IsMatch(result, SCRIPT_TAG_REGEX, RegexOptions.IgnoreCase))
+            {
+                LogWarning(Resources.ScriptTagsRemovalWarning, WarningLevel.Mild);
+                result = Regex.Replace(result, SCRIPT_TAG_REGEX, "", RegexOptions.IgnoreCase);
+            }
+
+            return result;
         }
 
+        private void LogWarning(string msg, WarningLevel level) => this.MessageLogged?.Invoke(LogMessage.Warning(msg, level));
```
mjkkirschner: will this work in c# 6?

mjkkirschner:
@mmisol two last comments and lgtm.

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
mmisol: Sure. I'll do that.

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
mmisol: It works. I know because I saw this PR's build go green having `OnLinkChanged` with a very similar implementation.

mmisol:
@mjkkirschner Can you take a final look?

src/DocumentationBrowserViewExtension/README.md
```diff
@@ -0,0 +1,63 @@
...
+            }
+            return dividend / divisor;
+        }
+    }
+}
+```
+
+This a simple example of a Zero-Touch node that performs a division. Taking a look at the exception message, it ends with a special formatted URI, following `href=`. This part of the message is not actually displayed to the final user, but instead is picked up by Dynamo to trigger the display of help content. Dissecting the syntax, three parts can be distinguished:
+1. `href=`: This marks where the message ends and the link information starts.
+2. `MyZeroTouchNode;`: This provides the name of the assembly which contains the HTML document. Semicolon is used to separate the assembly from the document.
```
mjkkirschner: should we mention that this assembly should be the same as the one which contains the node or any other assembly which is already loaded?

mmisol: Done!

radumg:
🙌

PR #10354, opened 2020-02-04
[DocumentationBrowser Internal] Use native MS browser in DocumentationBrowser

radumg:


mjkkirschner:
@radumg thanks for looking into this.

PR #10363, opened 2020-02-06
Type identifier in previews

SHKnudsen:
### Purpose

This PR adds a `Value Type` label to all node previews (watch nodes and the expandable node preview).

Currently the value labels are 

- `string`
- `int`
- `double`
- `object` 

Object is generic and will display on all that is not of type string, integer or double.

The label colors all follow the DesignScript type colors as they are displayed in code blocks.

Nested `List` or `Dictionary` wont have a label.

TypeIdentifier

image

### Declarations

### Reviewers

@mjkkirschner 
@radumg 

### FYIs

@mjkkirschner will wait for your comments before doing the tests

### DIFF
--- a/src/DynamoCoreWpf/Interfaces/IWatchHandler.cs
+++ b/src/DynamoCoreWpf/Interfaces/IWatchHandler.cs
@@ -75,7 +75,7 @@ namespace Dynamo.Interfaces
                     values = dict.Values.Cast<object>();
                 }
 
-                var node = new WatchViewModel(keys.Any() ? WatchViewModel.DICTIONARY : WatchViewModel.EMPTY_DICTIONARY, tag, RequestSelectGeometry, true);
+                var node = new WatchViewModel(keys.Any() ? WatchViewModel.DICTIONARY : WatchViewModel.EMPTY_DICTIONARY, tag, RequestSelectGeometry, TypeCode.Empty, true);
 
                 foreach (var e in keys.Zip(values, (key, val) => new { key, val }))
                 {
@@ -89,7 +89,7 @@ namespace Dynamo.Interfaces
             {
                 var list = (value as IEnumerable).Cast<dynamic>().ToList();
 
-                var node = new WatchViewModel(list.Count == 0 ? WatchViewModel.EMPTY_LIST : WatchViewModel.LIST, tag, RequestSelectGeometry, true);
+                var node = new WatchViewModel(list.Count == 0 ? WatchViewModel.EMPTY_LIST : WatchViewModel.LIST, tag, RequestSelectGeometry, TypeCode.Empty, true);
                 foreach (var e in list.Select((element, idx) => new { element, idx }))
                 {
                     node.Children.Add(callback(e.element, null, runtimeCore, tag + ":" + e.idx, showRawData));
@@ -112,15 +112,15 @@ namespace Dynamo.Interfaces
                     int typeId = runtimeCore.DSExecutable.TypeSystem.GetType(stackValue);
                     stringValue = runtimeCore.DSExecutable.classTable.ClassNodes[typeId].Name;
                 }
-                return new WatchViewModel(stringValue, tag, RequestSelectGeometry);
+                return new WatchViewModel(stringValue, tag, RequestSelectGeometry, Type.GetTypeCode(value.GetType()));
             }
 
             if (value is Enum)
             {
-                return new WatchViewModel(((Enum)value).GetDescription(), tag, RequestSelectGeometry);
+                return new WatchViewModel(((Enum)value).GetDescription(), tag, RequestSelectGeometry, Type.GetTypeCode(value.GetType()));
             }
 
-            return new WatchViewModel(ToString(value), tag, RequestSelectGeometry);
+            return new WatchViewModel(ToString(value), tag, RequestSelectGeometry, Type.GetTypeCode(value.GetType()));
         }
 
         private WatchViewModel ProcessThing(SIUnit unit, ProtoCore.RuntimeCore runtimeCore, string tag, bool showRawData, WatchHandlerCallback callback)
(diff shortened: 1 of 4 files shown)


src/DynamoCoreWpf/ViewModels/Preview/WatchViewModel.cs
```diff
@@ -36,6 +36,7 @@ internal void Click()
         private bool _isOneRowContent;
         private readonly Action<string> tagGeometry;
         private bool isCollection;
+        private string _outputType;
```
mjkkirschner: this is an old style, lets not use `_`

src/DynamoCoreWpf/ViewModels/Preview/WatchViewModel.cs
```diff
@@ -197,11 +198,17 @@ public IEnumerable<int> Levels
         /// </summary>
         public bool IsTopLevel { get; set; }
 
+        public string ValueType
```
mjkkirschner: /// comments please on public properties.

src/DynamoCoreWpf/ViewModels/Preview/WatchViewModel.cs
```diff
@@ -197,11 +198,17 @@ public IEnumerable<int> Levels
...
+            set { _outputType = value; RaisePropertyChanged(nameof(ValueType)); }
+        }
+
         #endregion
 
-        public WatchViewModel(Action<string> tagGeometry) : this(null, null, tagGeometry, true) { }
+        public WatchViewModel(Action<string> tagGeometry) : this(null, null, tagGeometry, TypeCode.Empty, true) { }
 
-        public WatchViewModel(string label, string path, Action<string> tagGeometry, bool expanded = false)
+        public WatchViewModel(string label, string path, Action<string> tagGeometry, TypeCode? outputType = TypeCode.Empty, bool expanded = false)
```
mjkkirschner: this is an API break - even the addition of an optional parameter is a binary break, you will need to add a new constructor instead, you can obsolete this one and add a note to use the new one. You may find you cannot / should not make this new parameter optional or ambiguous calls will be possible.

mjkkirschner: can you explain why you choose typeCode instead of Type?

src/DynamoCoreWpf/ViewModels/Preview/WatchViewModel.cs
```diff
@@ -211,6 +218,37 @@ public WatchViewModel(string label, string path, Action<string> tagGeometry, boo
...
             isCollection = label == WatchViewModel.LIST || label == WatchViewModel.DICTIONARY;
+            ValueType = GetDisplayType(outputType);
+        }
+
+        private string GetDisplayType(TypeCode? typecode)
+        {
+            switch (typecode)
+            {
+                case TypeCode.Boolean:
+                    return "bool";
```
mjkkirschner: toString() does not suffice?

src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
```diff
@@ -313,6 +313,42 @@
...
+                                        </Style.Triggers>
+                                    </Style>
+                                </Border.Style>
+                            </Border>
+                            <TextBlock Text="{Binding ValueType}"
+                                       FontStyle="Italic"
+                                       FontSize="8"
+                                       VerticalAlignment="Center"
+                                       Width="Auto"
+                                       FontFamily="Consolas"
```
mjkkirschner: @QilongTang do you know if we have fonts stored somewhere central where they can be referenced?

src/DynamoCoreWpf/ViewModels/Preview/WatchViewModel.cs
SHKnudsen: @mjkkirschner we added a new overload to the constructor that takes an object instead of string and from the object makes all the right string conversions using the `GetStringFromObject()`. Did not obsolete the current constructor as this seemed safer and you can make the call if it should or shouldn't be obsolete.

src/DynamoCoreWpf/ViewModels/Preview/WatchViewModel.cs
SHKnudsen: could do that, but the labels would be something like:

- `String`
- `Int64`
- `Double`
- `Object`
- `Boolean`

Instead of:
- `string`
- `int`
- `double`
- `object`
- `bool`

mjkkirschner: gotcha, this is fine with me then, can you add a comment noting you wanted customized user friendly strings?

src/DynamoCoreWpf/ViewModels/Preview/WatchViewModel.cs
```diff
@@ -213,6 +237,74 @@ public WatchViewModel(string label, string path, Action<string> tagGeometry, boo
...
+
+                case TypeCode.Object:
+                    return ToString(obj);
+
+                default:
+                    return (string)obj;
+            };
+        }
+
+        private static string ToString(object obj)
```
mjkkirschner: hmm, I guess this is okay, someone looking quickly at this might think this is overriding the regular ToString

src/DynamoCoreWpf/ViewModels/Preview/WatchViewModel.cs
```diff
@@ -213,6 +237,74 @@ public WatchViewModel(string label, string path, Action<string> tagGeometry, boo
...
+        {
+            return ReferenceEquals(obj, null) ?
+                Resources.NullString
```
mjkkirschner: now this... looks confusing - please chop this up so it's more maintainable.

Also reference equals with `null`? Maybe just == `null`? What do you need to accomplish by using ReferenceEquals method - `null` is not a type so this line doesn't make much sense to me...

mjkkirschner: see:
https://stackoverflow.com/questions/17667248/is-null-in-c-sharp-value-type-or-reference-type

SHKnudsen: Actually just moved this from the `IWatchHandler` to the `WatchViewModel`  as we are now converting the object to a label string here. 

But you're right it is very confusing. Will have a look at it.

mjkkirschner: @SHKnudsen can you use git blame to find out who wrote that code originally?

src/DynamoCoreWpf/Interfaces/IWatchHandler.cs
```diff
@@ -134,17 +134,17 @@ private WatchViewModel ProcessThing(SIUnit unit, ProtoCore.RuntimeCore runtimeCo
...
         private WatchViewModel ProcessThing(System.DateTime value, ProtoCore.RuntimeCore runtimeCore, string tag, bool showRawData, WatchHandlerCallback callback)
         {
-            return new WatchViewModel(value.ToString(PreferenceSettings.DefaultDateFormat, CultureInfo.InvariantCulture), tag, RequestSelectGeometry);
+            return new WatchViewModel(value, tag, RequestSelectGeometry);
         }
 
         private WatchViewModel ProcessThing(long value, ProtoCore.RuntimeCore runtimeCore, string tag, bool showRawData, WatchHandlerCallback callback)
         {
-            return new WatchViewModel(value.ToString(CultureInfo.InvariantCulture), tag, RequestSelectGeometry);
+            return new WatchViewModel(value, tag, RequestSelectGeometry);
```
mjkkirschner: I see you moved some of these cases into the `StringFromObject` function if I am interpreting correctly, but what about `long`

SHKnudsen: Long is also handled in [`GetStringFromObject` on `WatchViewModel`]( https://github.com/SHKnudsen/Dynamo/blob/a90c38c882eb9dc4f3a4140f1365148101647afa/src/DynamoCoreWpf/ViewModels/Preview/WatchViewModel.cs#L252)

ksobon:
@SHKnudsen @mjkkirschner I am sorry for butting in like this, but you guys should performance test this particular addition. Preview bubbles when opened on large data sets are insanely process expensive. Make a list of 100,000 integers or even 1,000,000 and open that sucker up. How much longer does it take to generate these previews now that you have additional label/border objects in that UI? What if you have multiple places in a graph that have these bubbles on, and now you are creating 5,000,000 of these? Maybe consider paging the preview bubble results? A "load more" button etc. I usually turn these off since they have been a problem when it comes to performance.

mjkkirschner:
@ksobon - thanks for your comment, I have the same concern.

radumg:
Agree @ksobon , we've already identified bubbles as a performance bottleneck in a previous profiling exercise, with similar mitigation strategies suggested. Would indeed be good to A/B test this one.

SHKnudsen:
A bit confused about why no one brought this up sooner if everyone agree this is a problem..

I see your point @ksobon, did a little comparison just to see how noticeable the change is, i know this isn't a bulletproof test put might give some insight to how big the performance drop is

Comparison

I dont see a big difference and doubt any users will notice this. I get that there might be more factors in play when looking at this.

To me being able to see right in the preview what type my outputs are is valuable, even though you can already get that by using the `Object.Type`, this is so much quicker and saves you from having to find that node every time you need to check value types. This would also help new users to better understand what the standard data type used in dynamo are. 

I dont see a major problem in tagging things like `Point` or `Surface` with `object` these things are `objects` and most of the time you can see what kind of object directly in the name. Again i think as a new user this can help understand what an object actually means.

Anyways thats just my thoughts...

PR #10365, opened 2020-02-06
Renamed node suffix label

SHKnudsen:
### Purpose

This PR adds a `Original Name` section to the tooltip of a node and divides the current tooltip description into 2 additional sections `Description` and `DesignScript Syntax`. The PR also adds a `Renamed` label to the Name block of the node, if the node has been renamed by the user.

The tooltip now displays the original name of the node, this makes it easy to identify nodes which has been renamed by users.

image

To easily identify if a node has been renamed a label will appear on nodes that have been renamed
image

RenamedNode

### Declarations

### Reviewers

@mjkkirschner 
@radumg 

### FYIs

@mjkkirschner will wait for your comments before doing the tests

### DIFF
--- a/src/DynamoCore/Graph/Nodes/NodeModelExtensions.cs
+++ b/src/DynamoCore/Graph/Nodes/NodeModelExtensions.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Reflection;
 using Autodesk.DesignScript.Interfaces;
 using Dynamo.Engine;
+using Dynamo.Graph.Nodes.CustomNodes;
 using Dynamo.Graph.Nodes.ZeroTouch;
 using ProtoCore.Mirror;
 
@@ -104,6 +105,15 @@ namespace Dynamo.Graph.Nodes
                 return elNameAttrib.Name;
 
             return nodeType.FullName;
+        }
+
+        internal static string GetCustomNodeOriginalName(this NodeModel node)
+        {
+            if (node == null) 
+                return string.Empty;
+
+            var customNodeFunction = node as Function;
+            return customNodeFunction.Definition.DisplayName;
         }
 
         private static void GetGraphicItemsFromMirrorData(MirrorData mirrorData, List<IGraphicItem> graphicItems)
--- a/src/DynamoCore/Graph/Nodes/NodeModelExtensions.cs
+++ b/src/DynamoCore/Graph/Nodes/NodeModelExtensions.cs
@@ -94,6 +94,8 @@ namespace Dynamo.Graph.Nodes
         internal static string GetOriginalName(this NodeModel node)
         {
             if (node == null) return string.Empty;
+            if (node.IsCustomFunction)
+                return GetCustomNodeOriginalName(node);
 
             var function = node as DSFunctionBase;
             if (function != null)
@@ -107,11 +109,8 @@ namespace Dynamo.Graph.Nodes
             return nodeType.FullName;
         }
 
-        internal static string GetCustomNodeOriginalName(this NodeModel node)
+        private static string GetCustomNodeOriginalName(NodeModel node)
         {
-            if (node == null) 
-                return string.Empty;
-
             var customNodeFunction = node as Function;
             return customNodeFunction.Definition.DisplayName;
         }
--- a/src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
@@ -208,19 +208,9 @@ namespace Dynamo.ViewModels
 
         /// <summary>
         /// The original name of the node.
-        /// This is the name of the node as it is when first added to the canvas.
         /// </summary>
         [JsonIgnore]
-        public string OriginalName
-        {
-            get 
-            {
-                if (nodeLogic.IsCustomFunction)
-                    return nodeLogic.GetCustomNodeOriginalName();
-
-                return nodeLogic.GetOriginalName(); 
-            }
-        }
+        public string OriginalName => nodeLogic.GetOriginalName();
 
         /// <summary>
         /// If a node has been renamed.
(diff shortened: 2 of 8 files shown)


src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
```diff
@@ -197,10 +198,45 @@ public bool IsSetAsOutput
...
+        /// </summary>
+        /// <returns></returns>
+        private string OriginalNameFromCreationName()
+        {
+            string[] longName = nodeLogic.CreationName.Split('@')[0].Split('.');
+            string[] typeAndMethodName = longName.Skip(Math.Max(0, longName.Count() - 2)).ToArray();
+            return String.Join(".", typeAndMethodName);
+        }
+
+        public bool IsRenamed
```
QilongTang: Do we really need to serialize this info or we can get from a runtime comparison?

mjkkirschner: changing the schema to include additional info is a headache - if it's possible to not serialize this and determine it based on a runtime comparison this would be useful.

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
```diff
@@ -213,6 +249,24 @@ public string Description
...
+        public string NodeDescription
+        {
+            get
+            {
+                return Description.Split(new[] { "\r\n", "\r", "\n" }, StringSplitOptions.None)[0];
+            }
+        }
+
+        [JsonIgnore]
+        public string DesignScriptSyntaxDescription
```
mjkkirschner: all public properties need /// comments please.

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
```diff
@@ -197,10 +198,45 @@ public bool IsSetAsOutput
...
+                return nodeLogic.Name; 
+            }
             set { nodeLogic.Name = value; }
         }
 
+        /// <summary>
+        /// The original name of the node.
+        /// This is the name of the node as it is when first added to the canvas.
+        /// </summary>
+        public string OriginalName
```
QilongTang: Same here, this change will add attributes to Dynamo json schema, it would be best we can avoid serializing new info at this point.

SHKnudsen:
@mjkkirschner @QilongTang - made a few clean ups and fixed the serialization issue (adding `[JsonIgnore]` to the new properties is enough right, or am i missing something?)

src/DynamoCore/Graph/Nodes/NodeModelExtensions.cs
```diff
@@ -106,6 +107,15 @@ internal static string GetOriginalName(this NodeModel node)
             return nodeType.FullName;
         }
 
+        internal static string GetCustomNodeOriginalName(this NodeModel node)
+        {
+            if (node == null) 
+                return string.Empty;
+
+            var customNodeFunction = node as Function;
```
mjkkirschner: what if this node is not a customNode?

mjkkirschner: if you have thinking of tests to write - this would be a good one since you caught it manually it would be good to load a custom node which has this naming issue and check that your function returns correctly when passed that node.

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
```diff
@@ -208,19 +208,9 @@ public string Name
...
-        {
-            get 
-            {
-                if (nodeLogic.IsCustomFunction)
-                    return nodeLogic.GetCustomNodeOriginalName();
-
-                return nodeLogic.GetOriginalName(); 
-            }
-        }
+        public string OriginalName => nodeLogic.GetOriginalName();
```
mjkkirschner: it's not clear to me if our build machine support expression body members yet, best to stick with the older syntax for now.

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
radumg: +1 in favour of runtime comparison and maybe doing this with a method instead of field/prop. @SHKnudsen Could we look into that please ?

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
radumg: do we need this if we have the method ?

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
```diff
@@ -213,6 +238,30 @@ public string Description
...
 
+        /// <summary>
+        /// The node description part of the nodes description
+        /// </summary>
+        [JsonIgnore]
+        public string NodeDescription
+        {
+            get
+            {
+                return Description.Split(new[] { "\r\n", "\r", "\n" }, StringSplitOptions.None)[0];
```
radumg: a comment that details why we're doing this `Split` would be useful

src/DynamoCoreWpf/Views/Core/NodeView.xaml
```diff
@@ -262,32 +291,83 @@
...
+                                <Setter Property="Visibility"
+                                        Value="Visible" />
+                            </DataTrigger>
+                        </Style.Triggers>
+                    </Style>
+                </Border.Style>
+                <TextBlock Text="Renamed"
+                           HorizontalAlignment="Center"
+                           VerticalAlignment="Center"
+                           FontStyle="Italic"
```
radumg: Unsure if this needs to be italic, the other `tags` we're adding to preview bubble aren't iirc, so I'd keep them consistent

SHKnudsen: all tags are in italic

QilongTang:
@DynamoDS/dynamo I changed the merge base of this PR and will merge for now so that I can submit code into the target branch. Will create a new PR for this work to wrap everything

PR #7587, opened 2017-02-09
Implement selective geometry preview using shaders

ellensi:
### Purpose

This pull request is a better reimplementation of my 2016 recharge project previously addressed by #6610.

A new toggle button called _**Selective Geometry Preview**_ is added to the top-right corner of the Dynamo canvas, just to the left of the Geometry/Node toggle buttons.

When the mode is enabled, only selected nodes will have their corresponding geometries displayed in full opacity. Geometries that correspond to unselected nodes will be displayed transparent with 10% opacity for meshes and 20% opacity for lines and points.

selective

#### Improvements

NEW implementation | Old implementation
------------ | -------------
Transparency flag is passed to the Dynamo shader (more efficient) | Alpha values were recalculated on C# for each mesh triangle (inefficient)
Lines and points can be displayed transparent with 20% opacity | Lines and points were not made transparent yet
Axes and grids are visible among the transparent geometries | Axes and grids were covered behind the supposedly transparent geometries
New button design and localized tooltip string | Dummy button design

#### Limitations

* Semi-transparent `Plane` geometries when selected may cause some unselected transparent geometries to be rendered wrongly
* Changing the "Show edges" setting while under selective geometry preview mode will reset the transparency. As a workaround, user can just click at any blank space of the node canvas.
* Direct Manipulation functionality cannot work properly behind transparent geometries

### Declarations

TBA

- [ ] The code base is in a better state after this PR
- [x] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [x] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [x] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions), and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.

### Reviewers

@Benglin @Racel 

### FYIs

@riteshchandawar

PR #10375, opened 2020-02-10
DYN-2406: better handle pts/lines when selecting, freezing, and isolating.

mjkkirschner:
### Purpose

PR does the following:
* when geometry is isolated and selected, should appear normal based on previous state. This was not being done correctly for meshes - corrected this.
* when geometry is isolated and unselected geometry should appear transparent.
* when pts or lines are selected they should turn blue. - no longer are vertex colors blended with selection color - instead selection overrides the vertex colors.
* when pts or lines are deselected they should return to their original colors. (which might be default or might be vertex colors) - or might be frozen (transparent).

### updates:

* modifies field of view of camera back to master - 58.5 degrees vs 45.
* fixes background color of watch3d - grey and yellow when in a home workspace vs custom node workspace.
* starts to add image comparison tests. WIP.

To accomplish this following was implemented:
* adds a frozen material vs isolation material as these have different alpha levels.

* when pts or lines colors are generated and DynamoGeometryModel is created when aggregating renderpackages we cache the original colors from the renderpackage on the HelixViewModel - later when some attached property is modified which requires reverting colors we can call reset colors which sets the colors to the old colors cached on the viewModel.

* this lookup is done by tagging the DynamoGeometryModelPoint or Line with the id of the colors in the cache so we can do a fast lookup.

* when an attached property changes which requires setting frozen or selected colors we generate those colors and set them to the vertex colors. The point and line materials in helix do not currently support the types of blending we would need to use them like we do for meshes - and in fact - it's not clear we can use those materials either- as we don't currently support vertex coloring of meshes yet. Next task.

* when a selection change event occurs in the canvas and isolate mode is active - we set isolate attached property on all the Geometry3dModels - previously we also removed everything from the sceneItems collection and then re-added everything to kick off a render - this is no longer necessary and was very slow causing noticeable stutter on my system - removing it, the color and material updates propagate fine as far as I can tell and tests continue to pass - though hopefully we can add image comparison tests to validate this kind of thing even more robustly.

### TODO
- [x] - more tests
- [ ] at least one image comparison test.
- [x] refactor out repeated code

### Declarations

### Reviewers
@QilongTang 
@reddyashish 
### FYIs


src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -1263,11 +1278,12 @@ private void InitializeHelix()
 
             headLight = new DirectionalLight3D
             {
-                Color = System.Windows.Media.Color.FromRgb(128, 128, 128),
-                Direction = new Vector3D(0, 0, 1),
+                Color = System.Windows.Media.Color.FromArgb(255,255, 0, 0),
```
mjkkirschner: this is a mistake, I was trying to see if the headlight was rendering at all - **it is not** 😢

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
```diff
@@ -12,6 +12,85 @@ namespace Dynamo.Wpf.ViewModels.Watch3D
...
+                            meshGeom.Material = HelixWatch3DViewModel.IsolatedMaterial;
+                        }
+                        else if (GetIsFrozen(meshGeom))
+                        {
+                            meshGeom.Material = HelixWatch3DViewModel.FrozenMaterial;
+                        }
+                        else
+                        {
+                            //TODO handle vertex coloring in all these cases
+                            meshGeom.Material = HelixWatch3DViewModel.WhiteMaterial;
```
QilongTang: Remind me what state it is other than selection, isolate and frozen?

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
```diff
@@ -38,41 +117,84 @@ public static bool GetShowSelected(DependencyObject element)
 
         private static void ShowSelectedPropertyChanged(DependencyObject obj, DependencyPropertyChangedEventArgs args)
```
QilongTang: I got you now, thank you for these comments helping me to understand. The implementation here does seem breakable. Maybe we can create several function as to color mesh/point&line in different condition-> isolate, frozen, selection because it seems we always need to go through these cases again and again.

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
```diff
@@ -148,21 +270,40 @@ private static void IsFrozenPropertyChanged(DependencyObject obj, DependencyProp
...
                     else
                     {
-                        dynamoGeom3D.Material = GetIsolationMode(dynamoGeom3D) ? 
-                            HelixWatch3DViewModel.TransparentMaterial : HelixWatch3DViewModel.WhiteMaterial;
+                        OnPropertySetFalse(meshGeom3d);
                     }
 
-                    dynamoGeom3D.RequiresPerVertexColoration = true;
-                    geom = dynamoGeom3D;
+                    meshGeom3d.CullMode = SharpDX.Direct3D11.CullMode.Back;
```
QilongTang: Can you add a comment here?

src/VisualizationTests/HelixWatch3DViewModelTests.cs
```diff
@@ -1204,5 +1231,12 @@ public static bool HasAnyColorMappedMeshes(this IEnumerable<Element3D> dictionar
 
             return geoms.Any(g => ((PhongMaterial)g.Material).DiffuseMap != null);
         }
+
+        public static bool HasSpecificColorCount(this GeometryModel3D geomModel, Color4 color, int count)
```
QilongTang: Might be trivial but can you add some description to this public methods?

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -156,6 +156,8 @@ public class HelixWatch3DViewModel : DefaultWatch3DViewModel
         private readonly Color4 defaultSelectionColor = new Color4(new Color3(0, 158.0f / 255.0f, 1.0f));
         private readonly Color4 defaultMaterialColor = new Color4(new Color3(1.0f, 1.0f, 1.0f));
         private readonly Color4 defaultTransparencyColor = new Color4(1.0f, 1.0f, 1.0f, 0.5f);
+        private readonly Color4 defaultIsolatedTransparencyColor = new Color4(1.0f, 1.0f, 1.0f, 0.3f);
```
QilongTang: Are these all defined by Dynamo team somehow?

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -331,7 +337,8 @@ public LineGeometry3D Axes
 
         public static PhongMaterial SelectedMaterial { get; set; }
 
-        public static PhongMaterial TransparentMaterial { get; set; }
```
QilongTang: Forgot if we agreed that we can remove public properties in this branch?

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -1687,12 +1715,20 @@ private void AggregateRenderPackages(IEnumerable<HelixRenderPackage> packages)
...
                         AddLabelPlace(baseId, p.Positions[0], rp);
                         if (pointGeometry3D.Geometry == null)
                         {
                             pointGeometry3D.Geometry = points;
                         }
+
+                        //while the Name of the geometry is the node which created it
+                        //we tag it with the id of the graphicModel we store in the scene/viewport for fast lookup.
                         pointGeometry3D.Name = baseId;
+                        pointGeometry3D.Tag = id; // this is more specific than the base id.
```
QilongTang: Good idea, how persistent this `id` is?

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
mjkkirschner: this has to do with: https://jira.autodesk.com/browse/DYN-2439
but actually I found we have the same artifacts on master, and switching to backface culling may not be such a great idea for open surfaces... so I will remove this change.

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
mjkkirschner: this was added by @aparajit-pratap - it's not an existing member in 2.5 (I don't think)

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
mjkkirschner: it's the `nodeASTID:geomType`

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
mjkkirschner: normal state.

aparajit-pratap:
> when geometry is isolated and selected, should appear normal based on previous state. This was not being done correctly for meshes - corrected this.
I see that this is the current behavior but was wondering if it makes sense to turn it into selection color instead just like it does for frozen nodes.

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
```diff
@@ -202,27 +291,37 @@ public static bool GetIsolationMode(DependencyObject element)
         {
             return (bool)element.GetValue(IsolationModeProperty);
         }
-        
+
         private static void IsolationModePropertyChanged(DependencyObject obj, DependencyPropertyChangedEventArgs e)
         {
-            if (obj is GeometryModel3D && obj.GetType() != typeof(BillboardTextModel3D))
+            //don't isolate the grid, axes, etc
+            if (obj is GeometryModel3D && !(obj is BillboardTextModel3D) && !IsSpecialRenderPackage(obj) )
```
aparajit-pratap: 👍

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
```diff
@@ -278,6 +377,42 @@ public static bool IsSpecialRenderPackage(DependencyObject element)
...
+        
+        /// <summary>
+        /// Sets all colors on Geometry to a single color.
+        /// </summary>
+        /// <param name="geom"></param>
+        /// <param name="color"></param>
+        private static void SetAllColors(GeometryModel3D geom, Color4 color)
+        {
+            var newColorCollection = new Color4Collection(Enumerable.Repeat(color, geom.Geometry.Colors.Count));
+            geom.Geometry.Colors = newColorCollection;
```
aparajit-pratap: So we still have to figure out why this doesn't work for meshes?

mjkkirschner:
> > when geometry is isolated and selected, should appear normal based on previous state. This was not being done correctly for meshes - corrected this.
> 
> I see that this is the current behavior but was wondering if it makes sense to turn it into selection color instead just like it does for frozen nodes.

I was thinking the same thing initially but decided on keeping the current behavior as it seems useful to be able to view the colors of the selected geometry, I can already use selection as a quick way to see which geometry is which but it does not help me to quickly see vertex colored geometry.

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
aparajit-pratap: @mjkkirschner where did you get the `0.3f` from? Is it specified somewhere?

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -1468,6 +1488,14 @@ private void RemoveGeometryForUpdatedPackages(IEnumerable<IRenderPackage> packag
             }
         }
 
+        protected override void AttachedProperties_RequestResetColorsForDynamoGeometryModel(string objId)
+        {
+            if (objId!= null && this.colorCache.ContainsKey(objId)){
```
aparajit-pratap: `if(!string.IsNullOrEmpty(objId))`?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -1049,13 +1049,12 @@ private void DynamoViewModelRequestSaveImage(object sender, ImageSaveEventArgs e
 
         private void DynamoViewModelRequestSave3DImage(object sender, ImageSaveEventArgs e)
         {
-            var canvas = (DPFCanvas)BackgroundPreview.View.RenderHost;
-
+            var bitmapSource =BackgroundPreview.View.RenderBitmap();
```
aparajit-pratap: Could you explain this change?

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
```diff
@@ -435,6 +377,42 @@ public static bool IsSpecialRenderPackage(DependencyObject element)
...
+        }
+
+        /// <summary>
+        /// Sets alpha channel of existing colors to given value, optionally attempts to reset all colors
+        /// first to original colors from renderpackages.
+        /// </summary>
+        /// <param name="geom"></param>
+        /// <param name="alpha"></param>
+        /// <param name="resetColorsFirst"></param>
```
QilongTang: This is minor but can you set the description for the params as well, especially this boolean?

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
mjkkirschner: not that I could find, I just tried to match the existing look, where isolated geometry is more transparent than frozen geometry.

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
```diff
@@ -38,41 +101,74 @@ public static bool GetShowSelected(DependencyObject element)
...
+                            else
+                            {
+                                //reset the colors
+                                RequestResetColorsForDynamoGeometryModel?.Invoke(geom.Tag as string);
+                            }
+                        }
+                        //only selected.
+                        else
+                        {
+                            //TODO cache this and update after helix 2.11 is released
```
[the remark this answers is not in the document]
mjkkirschner: yep yep!

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
mjkkirschner: yes, I have just grabbed that task, I think it has to do with the shader we are using.

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
mjkkirschner: will do.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
mjkkirschner: yep, this crashed on the new helix version, so I moved to the new helix api.

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
mjkkirschner: will do.

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
```diff
@@ -12,7 +13,69 @@ namespace Dynamo.Wpf.ViewModels.Watch3D
     /// </summary>
     public static class AttachedProperties
     {
-        private const float alphaPropertyFactor = 0.5f;
+
+        // handles determining color of elementGeometry3Ds when any property is set false
+        // as the state of other properties must be checked to determine the correct color / material.
+        private static void OnPropertySetFalse(DependencyObject obj)
+        {
+            if (obj is GeometryModel3D && obj.GetType() != typeof(BillboardTextModel3D))
```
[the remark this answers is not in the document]
mjkkirschner: done, I've applied this same pattern to all of these statements in the file.

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
mjkkirschner: done.

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
aparajit-pratap: Looks like this value was originally set to 0.1f for lines and points: https://github.com/DynamoDS/Dynamo/pull/7587/files#diff-20b13dfe2f88aaf8df37e1240b79c201R30 and 0.2f .

aparajit-pratap:
Aside from another comment about the alpha value chosen, LGTM.

reddyashish:
LGTM.

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
mjkkirschner: @aparajit-pratap - thanks for finding this!

mjkkirschner:
meshes at 0.2f, pts and lines at 0.3f
Screen Shot 2020-02-20 at 2 33 10 PM


 meshes, points, and lines at .2f alpha

Screen Shot 2020-02-20 at 2 30 46 PM

if we bring pts and lines down to .1f alpha as before they are completely invisible....

@aparajit-pratap @reddyashish @QilongTang  what do you think?

QilongTang:
@mjkkirschner These are light enough I think we can stick to the new values?

mjkkirschner:
Screen Shot 2020-02-20 at 3 18 31 PM

new on the left
master on the right
using .25f for alpha for lines and points - @QilongTang  and I think this is good. Will merge after running all visualization tests locally.

mjkkirschner:
fixed latest color test additions for behavior on this branch - merging.

QilongTang:
LGTM

aparajit-pratap:
@mjkkirschner sorry, my previous comment was all jumbled up. I think it was set to 0.1f for meshes and 0.2f for points and lines earlier. Did you try these values and find them to appear to be too light?

mjkkirschner:
@aparajit-pratap .2f was too light for points/lines- but I will try .1f for meshes and compare now before I merge.

PR #10387, opened 2020-02-13
Update README.md

mjkkirschner:
@QilongTang @reddyashish

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Looking to learn or download Dynamo?  Check out [dynamobim.org](http://dynamobim
 If you're interested in developing a Node library for Dynamo, the easiest place to start is by browsing the [DynamoSamples](https://github.com/DynamoDS/DynamoSamples).  
 These samples use the [Dynamo NuGet packages](https://www.nuget.org/packages?q=DynamoVisualProgramming) which can be installed using the NuGet package manager in Visual Studio.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -14,7 +14,8 @@ Looking to learn or download Dynamo?  Check out [dynamobim.org](http://dynamobim
 If you're interested in developing a Node library for Dynamo, the easiest place to start is by browsing the [DynamoSamples](https://github.com/DynamoDS/DynamoSamples).  
 These samples use the [Dynamo NuGet packages](https://www.nuget.org/packages?q=DynamoVisualProgramming) which can be installed using the NuGet package manager in Visual Studio.
 
-[Documentation of the Dynamo API]( http://dynamods.github.io/DynamoAPI) with a searchable index of public API calls for core functionality. This will be expanded to include regular nodes and Revit functionality.
+[Documentation of the Dynamo API via Fuget.org]( https://www.fuget.org/packages/DynamoVisualProgramming.Core/2.5.0.7432) with a searchable index of public API calls for core functionality in the dynamo nuget packages. *WIP*.
```
mmisol: If we point to `https://www.fuget.org/packages/DynamoVisualProgramming.Core` instead, people are automatically shown the latest version by default. Should we remove the version from the link?

QilongTang: @mmisol It is pointing to latest beta which is actually behind in this case, I guess that is why @mjkkirschner did not put without a version :)

mjkkirschner: yep yep, what @QilongTang said.

PR #10399, opened 2020-02-18
[DYN-2341] Labels on helix-update branch cause slow down

reddyashish:
### Purpose

Fix the null pointer exception. After this fix, the labels on the workspace will render without any lags. Dynamo used to hang after being used for sometime as multiple exceptions were being thrown on the geom.Geometry.Positions object. 

JIRA: https://jira.autodesk.com/browse/DYN-2341

Before the change:
without fix
The exception is shown on the dynamo console. 

After the change:
after fix

### Declarations

### Reviewers

### DIFF
--- a/src/VisualizationTests/HelixWatch3DViewModelTests.cs
+++ b/src/VisualizationTests/HelixWatch3DViewModelTests.cs
@@ -823,6 +823,7 @@ namespace WpfVisualizationTests
         // Before the fix, this test would take around 5 mins to finish but now this test finishes in just 20 secs. 
         public void PerformanceTestOnLabelsAfterHelixUpgrade()
         {
+            System.DateTime startTime = System.DateTime.Now;
             OpenVisualizationTest("PerformanceTestOnLabelsAfterHelixUpgrade.dyn");
             var ws = ViewModel.Model.CurrentWorkspace as HomeWorkspaceModel;
 
@@ -861,6 +862,9 @@ namespace WpfVisualizationTests
                 Assert.AreEqual(1, geometry.TextInfo.Count);
                 Assert.AreEqual("["+ itemIndex  +"]", geometry.TextInfo[0].Text);
             }
+            System.DateTime endTime = System.DateTime.Now;
+            var totalExecutionTime = (endTime - startTime).TotalSeconds;
+            Assert.LessOrEqual(totalExecutionTime, 20);
         }
 
         [Test]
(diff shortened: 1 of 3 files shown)


QilongTang:
LGTM

aparajit-pratap:
@reddyashish can we verify this fix with tests pertaining to labels on this branch?

aparajit-pratap:
Any idea why the EngOps PR job is failing?

reddyashish:
@aparajit-pratap For the test "Display_Geometry_Labels", there is not much difference in the time it takes to run. The test times are not consistent and there is a difference of 0-4 secs in test time before and after the fix. 
I have added GIF's to compare the slowness on dynamo before and after this change.

reddyashish:
@mjkkirschner @aparajit-pratap Added a test to compare the performance. Before this fix, the test would take around 5 mins but now it takes 20 secs to finish.

mjkkirschner:
LGTM

src/VisualizationTests/HelixWatch3DViewModelTests.cs
```diff
@@ -816,6 +816,53 @@ public void Display_Geometry_Labels()
             Assert.AreEqual("[0,0,1]", geometry.TextInfo[0].Text);
         }
 
+        [Test]
+        // This test will select a sphere object 70 times from a list of sphere's, to display 
+        // the corresponding label for that sphere object. After the Helix update, this workflow was causing
+        // delays and would cause dynamo to hang. The fix was added in this PR: https://github.com/DynamoDS/Dynamo/pull/10399
+        // Before the fix, this test would take around 5 mins to finish but now this test finishes in just 20 secs. 
+        public void PerformanceTestOnLabelsAfterHelixUpgrade()
```
mjkkirschner: do you think it is worth timing this test and asserting it takes less than a minute or something like that?

mmisol: Also, might we pick a lower number than 70 for the test? Since the time execution reduced greatly after the improvement, we could still see a difference with maybe only 7 selections. By doing that we'd reduce the execution time of this test down to a few seconds.

reddyashish: For number of selections less than 10, there is not much difference in the test time before and after the fix. For 30 selections, we can see a time difference of around 30-40 secs. As more and more labels are rendered, number of exceptions caught increases and dynamo's performance decreases exponentially. So, I went with a bigger selection number. 
I think 30 should be fine, let me know your thoughts.

mmisol: Sounds good 👍  Thanks @reddyashish

mmisol: Nit: The comments still say 70 after the change

reddyashish: Thanks @mmisol . Updated the comments.

mmisol:
LGTM

reddyashish:
Merging this PR.

PR #10407, opened 2020-02-21
GA client not launching when terms not agreed

QilongTang:
Please Note:

### Purpose

Do not even try to launch or load GA binaries and client when terms not agreed. 

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -594,8 +594,9 @@ namespace Dynamo.Models
                 PreferenceSettings = settings;
                 PreferenceSettings.PropertyChanged += PreferenceSettings_PropertyChanged;
             }
-
-            InitializeInstrumentationLogger();
+            // If user does not agree to GA terms, do not try to launch the client at all
+            if (PreferenceSettings.IsAnalyticsReportingApproved)
+                InitializeInstrumentationLogger();
 
             if (!IsTestMode && PreferenceSettings.IsFirstRun)
             {
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -596,7 +596,9 @@ namespace Dynamo.Models
             }
             // If user does not agree to GA terms, do not try to launch the client at all
             if (PreferenceSettings.IsAnalyticsReportingApproved)
+            {
                 InitializeInstrumentationLogger();
+            }
 
             if (!IsTestMode && PreferenceSettings.IsFirstRun)
             {
--- a/test/DynamoCoreWpfTests/CoreUITests.cs
+++ b/test/DynamoCoreWpfTests/CoreUITests.cs
@@ -502,7 +502,7 @@ namespace DynamoCoreWpfTests
 
         [Test]
         [Category("DynamoUI")]
-        public void PreferenceSetting_AgreeAnalyticSharing()
+        public void PreferenceSetting_NotAgreeAnalyticsSharing()
         {
             // Test deserialization of analytics setting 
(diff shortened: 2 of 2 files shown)


src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -594,8 +594,9 @@ protected DynamoModel(IStartConfiguration config)
                 PreferenceSettings = settings;
                 PreferenceSettings.PropertyChanged += PreferenceSettings_PropertyChanged;
             }
-
-            InitializeInstrumentationLogger();
+            // If user does not agree to GA terms, do not try to launch the client at all
+            if (PreferenceSettings.IsAnalyticsReportingApproved)
```
mjkkirschner: would you mind using braces?

Also, we could write a test for this perhaps which checks that the dlls are not loaded - I will leave that up to you to decide if it's necessary to protect this behavior in the future - but here is an example I think would be good to base the test on.

https://github.com/DynamoDS/Dynamo/blob/1d49129f5fc1f0f80abf94ae9bc04833cd069064/test/System/IntegrationTests/DynamoApplicationTests.cs#L34

we would want to startup a new process and check the modules which are loaded, we can't rely on checking the loaded modules of the test because some previous test may have already loaded the modules.

QilongTang: Added braces. I did explore the unit testing side, I find Analytics.Net.Core.dll can only be detected using `AppDomain.CurrentDomain.GetAssemblies()` which requires a way to monitor the appdomain came with the new process. I will try to make it into a serparate PR. I did run self CI on this one and everything is passing. So merging for now and cherry-pick to get a build for customer testing first.

mjkkirschner:
https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.modules?redirectedfrom=MSDN&view=netframework-4.8#System_Diagnostics_Process_Modules

process.modules should return all modules (native and managed) which are loaded by a process - have you tried using it on sandbox with analytics enabled?

QilongTang:
@mjkkirschner That is indeed what I tried and I saw Analytics.Net.Core.dll is not among the list

mjkkirschner:
did you use WaitForInputIdle as in the example, if you check the modules before this there is no guarantee that all modules are loaded when the process is just started.
```
After starting the process, 
this collection is empty until the system has loaded the process. 
If the process has a main window,
 you can call WaitForInputIdle before retrieving this property
 to ensure that the collection is nonempty when you get the list.
```

QilongTang:
@mjkkirschner Here is the list for a Sandbox.exe and its modules, notice you can tell the analytics is on here because you would be able to interact with sandbox when you launch it in the way you recommend.

image
image
image

I dont think the list is equivalent than getting assemblies from AppDomain, you can tell it because even DynamoCore.dll is not in this list

mjkkirschner:
@QilongTang DynamoCore won't be loaded until Dynamo is actually loaded - IE until DynamoModel is constructed or construction starts, when you are checking this list, is the window open?

QilongTang:
@mjkkirschner Like I mentioned above, When I was checking the modules, I can already interact with Dynamo and see Analytics is enabled.

mjkkirschner:
Thanks @QilongTang  - pretty odd, let's leave it for later then, we'll have to find the magic incantation to get managed dlls to appear here...

QilongTang:
@mjkkirschner I found checking app domain worked for the current NUnit session. DynamoCore and other Dynamo dlls are all listed in `AppDomain.CurrentDomain.GetAssemblies() call`. so we may need to do something like this https://stackoverflow.com/questions/14758915/get-all-processes-with-their-corresponding-app-domains

PR #10421, opened 2020-02-27
Patch 283

saintentropy:
### Purpose

This patch address issues in [PSIRT-283](https://jira.autodesk.com/browse/PSIRT-283)

### Declarations

### Reviewers

@mjkkirschner @gregmarr @Dewb @QilongTang


src/DynamoUtilities/CertificateVerification.cs
```diff
@@ -25,37 +303,19 @@ public static bool CheckAssemblyForValidCertificate(string assemblyPath)
...
-                cert = X509Certificate.CreateFromSignedFile(assemblyPath);
+                var validCert = WinTrustInterop.WinTrust.IsTrusted(assemblyPath);
+                if (validCert)
+                {
+                    return true;
+                }
             }
             catch
             {
                 throw new Exception(String.Format(
```
mjkkirschner: would it make sense to throw a more specific exception type here and test for this type explicitly in the tests?

src/DynamoUtilities/CertificateVerification.cs
```diff
@@ -3,12 +3,290 @@
...
+
+            public static bool IsTrusted(string fileName)
+            {
+                return WinVerifyTrust(fileName) == 0;
+            }
+        }
+
+        private struct WINTRUST_FILE_INFO : IDisposable
+        {
+            public WINTRUST_FILE_INFO(string fileName, Guid subject)
```
mjkkirschner: /// comments I think would be useful

src/DynamoUtilities/CertificateVerification.cs
```diff
@@ -25,37 +303,19 @@ public static bool CheckAssemblyForValidCertificate(string assemblyPath)
...
-            }
-
-            if (cert != null)
-            {
-                var cert2 = new System.Security.Cryptography.X509Certificates.X509Certificate2(cert);
-                if (cert2.Verify())
-                {
-                    return true;
-                }
+                    "A dll file found at {0} did not have a signed certificate.", assemblyPath));
```
gregmarr: Do you need to have this `throw` here?  Could you just let it fall through to the `throw` on the next line?

src/DynamoUtilities/CertificateVerification.cs
```diff
@@ -3,12 +3,290 @@
...
+
+                fdwRevocationChecks = RevocationCheckFlags.None;
+
+                dwStateAction = StateAction.Ignore;
+
+                hWVTStateData = IntPtr.Zero;
+
+                pwszURLReference = IntPtr.Zero;
+
+                dwProvFlags = TrustProviderFlags.Safer;
```
mjkkirschner: I don't think we should use this.

src/DynamoUtilities/CertificateVerification.cs
saintentropy: The updated exceptions provide diffraction if the check failed due to exception or was actually false

src/DynamoUtilities/CertificateVerification.cs
```diff
@@ -3,12 +3,194 @@
...
+            Execute = 0,
+            Install = 1
+        }
+        #endregion
+
+        #region WinTrust structures
+        [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Unicode)]
+        class WinTrustFileInfo
+        {
+            UInt32 StructSize = (UInt32)Marshal.SizeOf(typeof(WinTrustFileInfo));
```
mjkkirschner: is this used anywhere?

src/DynamoUtilities/CertificateVerification.cs
```diff
@@ -3,12 +3,194 @@
...
+                    Marshal.FreeCoTaskMem(pszFilePath);
+                    pszFilePath = IntPtr.Zero;
+                }
+            }
+        }
+
+        [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Unicode)]
+        class WinTrustData
+        {
+            UInt32 StructSize = (UInt32)Marshal.SizeOf(typeof(WinTrustData));
```
mjkkirschner: is this used anywhere?

src/DynamoUtilities/CertificateVerification.cs
saintentropy: It is not

src/DynamoUtilities/CertificateVerification.cs
```diff
@@ -3,12 +3,194 @@
...
+                    return ret;
+                }
+                finally
+                {
+                    // free the locally-held unmanaged memory in the data structures
+                    if (winTrustFileInfo != null) winTrustFileInfo.Dispose();
+                    if (winTrustData != null) winTrustData.Dispose();
+                }
+            }
+            private WinTrust() { }
```
mjkkirschner: why is this here?

src/DynamoUtilities/CertificateVerification.cs
```diff
@@ -3,12 +3,194 @@
...
+            SubjectCertExpired = 0x800B0101,        // CERT_E_EXPIRED - Signer's certificate was expired
+            SubjectCertificateRevoked = 0x800B010C,     // CERT_E_REVOKED Subject's certificate was revoked
+            UntrustedRoot = 0x800B0109          // CERT_E_UNTRUSTEDROOT - A certification chain processed correctly but terminated in a root certificate that is not trusted by the trust provider.
+        }
+
+        public class WinTrust
+        {
+            private static readonly IntPtr INVALID_HANDLE_VALUE = new IntPtr(-1);
+            // GUID of the action to perform
+            private const string WINTRUST_ACTION_GENERIC_VERIFY_V2 = "{00AAC56B-CD44-11d0-8CC2-00C04FC295EE}";
```
mjkkirschner: should this be static?

src/DynamoUtilities/CertificateVerification.cs
saintentropy: But it fails if it is not set

src/DynamoUtilities/CertificateVerification.cs
saintentropy: Same as above

src/DynamoUtilities/CertificateVerification.cs
saintentropy: const is set at compile time vs static readonly set at runtime

src/DynamoUtilities/CertificateVerification.cs
saintentropy: This does not seem to be necessary

src/DynamoUtilities/CertificateVerification.cs
mjkkirschner: okay, I assume this is because the the class needs this field to exist when passed into native code and the layout is sequential so it needs to be first.

mjkkirschner: yes, I guess these map to `cbStruct`

test/Libraries/DynamoUtilitiesTests/CertificateVerificationTests.cs
```diff
@@ -24,5 +24,43 @@ public void CheckAssemblyForValidCertificateTest()
...
+            ); 
+        }
+
+        [Test]
+        [Category("UnitTests")]
+        public void CheckAssemblyForValidCertificateTest_TransferredCertificate()
+        {
+            var executingDirectory = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+            var directory = new DirectoryInfo(executingDirectory);
+            var testFilePath = Path.Combine(directory.Parent.Parent.Parent.FullName, "test", "pkgs_signed", "Modfied Signed Package", "bin", "LegitAssemblyTransplanted.dll");
```
Dewb: typo: `Modfied` => `Modified`

src/DynamoUtilities/CertificateVerification.cs
Dewb: The purpose is to keep other classes from being able to construct instances of this otherwise static class.

PR #10437, opened 2020-03-03
Added a couple of asserts for the properties.

manuelsaldivar525:
Description
Added a couple of asserts for the properties that have values on get that comes from a private const variables and remove the getter and setter from WorkspaceDependencyViewExtension.DependencyView because these methods do not get value.

Related GitHub issue or JIRA story
https://jira.autodesk.com/browse/DYN-2418

Checklist
[ - ] PR title reflects the changes it introduces. Consider using this template: QNTM-XXXX: Implement X Feature
[ - ] Code compiles correctly
[ - ] No linting errors
 All tests passing
[ - ] Updated README.md / documentation (if necessary)

### DIFF
--- a/src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
+++ b/src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
@@ -20,11 +20,8 @@ namespace Dynamo.WorkspaceDependency
         internal MenuItem workspaceReferencesMenuItem;
         private const String extensionName = "Workspace References";
 
-        internal WorkspaceDependencyView DependencyView
-        {
-            get;
-            set;
-        }
+        internal WorkspaceDependencyView DependencyView;
+       
 
         internal PackageManagerExtension pmExtension;
 
--- a/src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
+++ b/src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
@@ -20,8 +20,11 @@ namespace Dynamo.WorkspaceDependency
         internal MenuItem workspaceReferencesMenuItem;
         private const String extensionName = "Workspace References";
 
-        internal WorkspaceDependencyView DependencyView;
-       
+        internal WorkspaceDependencyView DependencyView
+        {
+            get;
+            set;
+        }
 
         internal PackageManagerExtension pmExtension;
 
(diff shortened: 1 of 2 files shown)


src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
```diff
@@ -20,11 +20,8 @@ public class WorkspaceDependencyViewExtension : IViewExtension, ILogSource
         internal MenuItem workspaceReferencesMenuItem;
         private const String extensionName = "Workspace References";
 
-        internal WorkspaceDependencyView DependencyView
-        {
-            get;
-            set;
-        }
+        internal WorkspaceDependencyView DependencyView;
```
mjkkirschner: I don't think it makes sense to turn this into a field - there are tests which access this property external to the class which is usually done with a property - field access across objects is discouraged -  and below the setter is called inside the loaded function:

https://github.com/DynamoDS/Dynamo/pull/10437/files#diff-5c424b646bae377139b0c20ee304d7edR92

I see the getter called inside this class as well.

The reason it's likely you do not see these properties executed during normal testing is that this is a ViewExtension which will only be loaded inside tests which actually create a full DynamoView.

mjkkirschner:
@manuelsaldivar525 when making a PR to dynamo please stick with the dynamo issues template.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
manuelsaldivar525: I already did the change mjkkirschner .

QilongTang:
Adding a test covering initialization is fine, does this help cover `viewExtension.Name`?

manuelsaldivar525:
> Adding a test covering initialization is fine, does this help cover `viewExtension.Name`?

I added them both viewExtension.Name and viewExtension.UniqueId into TestPropertiesWithCodeInIt() to cover all code including the properties which have value and yes the cover takes them into account.

QilongTang:
@manuelsaldivar525 Would you also add a screenshot covering the code coverage difference before we can merge this?

manuelsaldivar525:
> @manuelsaldivar525 Would you also add a screenshot covering the code coverage difference before we can merge this?


I added the screenshot on task https://jira.autodesk.com/browse/DYN-2418 where TestPropertiesWithCodeInIt() is reflected.

QilongTang:
@manuelsaldivar525 Thanks! LGTM and thank you for addressing @mjkkirschner 's comment

PR #10438, opened 2020-03-03
DNM for feedback - Mesh vertex colors simple

mjkkirschner:
Screen Shot 2020-03-03 at 4 49 05 PM

on the left is existing behavior in master, on the right is this branch.
This uses the new `vertexColorBlendFactor` for phong materials in helix 2.12 - it works fine for the regular display material, unfortunately for frozen and isolated geometry, where we want to use the full vertex color but modify the alpha values to a known alpha, it does not work without performing the per vertex color change to the alpha values. 

We could of course add a custom shader or modify the built in helix shader and send a PR or maintain our own helix implementation for now, but all of these add complexity, build issues, maintenance issues etc. - 

we already have the utilities to do this easily as we do it for points and lines - but I was hoping to avoid it - I will work on that in another branch to see the memory/performance impact - in the mean time I wanted to get feedback on this hack - which is simply blending between the frozen or isolated color (white with low alpha) and the vertex colors - 

you can see that the resulting color is transparent, but it's hard to tell it has any color compared to the standard white material. I have used a blend factor of .1f - which means lerp 90% from the vertex color to the original color defined by the isolatedMaterial or FrozenMaterial.

Curious what the team's thoughts are on this vs original behavior since the implementation is so simple.

I'm not loving it.

@DynamoDS/dynamo 

Will send another PR soon with the more complex implementation restoring the original behavior.

if anyone is curious I am using latest helix builds from:
https://www.myget.org/F/helix-toolkit


src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -332,12 +332,32 @@ public LineGeometry3D Axes
...
-        public static PhongMaterial IsolatedMaterial { get; set; }
+        /// <summary>
+        /// Material for frozen mesh geometry - does not support vertex colors
+        /// </summary>
+        internal static PhongMaterial FrozenMaterial { get; set; }
+        /// <summary>
+        /// Material for frozen mesh geometry - supports vertex colors 
+        /// - which means that vertex alpha must be set correctly to standard frozen alpha.
+        /// </summary>
+        internal static PhongMaterial FrozenMaterialVertColor { get; set; }
```
mjkkirschner: these additional materials are not actually used in this branch, I'm using them in the alternative implementation.

PR #10439, opened 2020-03-04
WIP Mesh vertex colors

mjkkirschner:
### Purpose

implements vertex color support for `Display.ByGeometryNode` - and arbitrary drawing of render packages with `MeshVertexColors` set to some custom array of color bytes.

This implementation is similar to the implementation on this branch for points and lines, so the logic is very similar and uses some of the functions there.

Two new materials are added - 
`IsolatedMaterialVertColor` and `FrozenMaterialVertColor` - these materials set their vertex blending to full vertex colors -and as such need their alpha channel set appropriately for frozen and isolated modes on each vertex.

The approach taken here switches to these materials when the `DynamoGeometryModel3D` objects have their `requiresPerVertexColoring` flag set true - in these cases we also walk the colors and modify their alpha values to the correct value - if the mesh contains many verts - this can be slow. We can offload some of this work by precomputing the modified colors but the work needs to be done somewhere, and then we need to cache those colors as well... currently the original colors and cached, and the alpha modified colors are computed on the fly, the same approach that was taken for points and lines.

It's not ideal to need to do this, previously this type of control was deferred to the shader. My feeling currently is to move forward with an approach like this - while weighing the performance vs maintainability implications, and doing some more testing around optimizations we can do on the c# side.

 I'd also like to see how difficult it would be to link a custom shader just for vertex coloring, leaving out the other complexities of isolate,frozen etc. The additional flag we would want is to override all alpha values of all colors. Something to look into which should be very simple in a shader.

Screen Shot 2020-03-03 at 8 08 46 PM
Screen Shot 2020-03-03 at 8 14 32 PM

NOTE:
This does NOT support the `Surface.ByGeometryColor` node as that node generates custom materials for each output and creates a bitmap texture in memory - that does not play well at all with the current implementation where materials are switched out for the different rendering effects. I think the AC for this task define that as well, but it may make sense to split it out.
TODO:
- [ ]  tests
- [ ]  surface.ByGeometryNode (diffuse texture map)
- [ ] try out minimal vertex coloring shader with alpha channel override

### Declarations


mjkkirschner:
the builds are failing because they do not have access to the helix nuget source for pre release arifacts on myget

PR #10462, opened 2020-03-12
Mesh display using shaders

mjkkirschner:
### Purpose
This PR brings Dynamo's customer shader implementation into helix 2.9 (btw 2.11 is out and we can upgrade pretty easily yay!)

I tried various other solutions but the matrix of features and drawing modes is quite large - to get them all working together without a shader is IMO - more complex, and way less performant than with one.

PR does the following:

* ~~merges in master to helix-upgrade branch (sorry for that noise) if required I can send that as a separate PR to make this diff smaller.~~

* removes our use of materials for mesh drawing - only one material is now used. The others are made internal and only kept around to refer to their properties.

* adds the `DynamoGeometryMeshCore` and `DynamoGeometryNode` classes which follow the helix customShaderDemo pattern to pick a custom shader for a GeometryModel and send data to that shader on a per model basis. Ie is this model frozen or not.

* Data is sent as a bit packed float - in the shader that gets cast to an int and decoded with a binary & - for more info on bitmasks see here: https://en.wikipedia.org/wiki/Mask_(computing)
Some flags are not currently used - these are marked as such, but I think they could be useful.

* adds a script to compile the shaders - with comments

* adds a readme about updating the shaders in the future

* adds the compiled shaders and embeds them into DynamoCoreWPF - these are loaded with the DynamoEffectsManager class as byte arrays.

* adds 3 files from helix's built in shader so that we can access common datastructures that helix passes to all shaders. I removed all other helix refs.

* adds two shader .hlsl files - one vertex shader and fragment shader - these are only used for Dynamo meshes. **Points and lines** continue to use the builtin helix shader and materials - we can consider changing these later (or next), but in my testing performance was fairly good for points/lines with the current implementation. No doubt it could be faster with a shader.

* Reverts a few things to avoid any unnecessary API breaks. I know there are still issues we cannot solve because of helix type changes.

### Declarations

### Reviewers

### FYIs


mjkkirschner:
SurfaceColors - uses a texturemap and has no shading. It's broken here and in master in a bunch of ways! Let's fix that later. (doesn't isolate or freeze).

Screen Shot 2020-03-12 at 7 29 57 PM

Regular geometry isolated.

Screen Shot 2020-03-12 at 7 30 28 PM

Regular geometry - normal mode - all flags disabled.

Screen Shot 2020-03-12 at 7 30 22 PM

vertex colored geometry - all other flags disabled.

Screen Shot 2020-03-12 at 7 30 11 PM

mjkkirschner:
supports multiple lights, here the headlight is red - one thing to note is that we have removed the headlight from the shader and have added it to the scene - in the future this might let users control the headlight color or move it etc.

Screen Shot 2020-03-12 at 5 34 07 PM

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
```diff
@@ -381,6 +320,19 @@ public static bool IsSpecialRenderPackage(DependencyObject element)
...
+        /// <summary>
+        /// Sets the property value on the DynamoMeshCore - this makes its way down to our shader and invalidates the render.
+        /// </summary>
+        /// <param name="meshGeom"></param>
+        /// <param name="args"></param>
+        private static void HandleMeshPropertyChange(DynamoGeometryModel3D meshGeom, DependencyPropertyChangedEventArgs args)
+        {
+            if ((meshGeom.SceneNode.RenderCore as DynamoGeometryMeshCore) != null)
+            {
+                (meshGeom.SceneNode.RenderCore as DynamoGeometryMeshCore).SetPropertyData(args);
```
mjkkirschner: this method directs dependency property changes on the GeometryModels and updates the meshCore class - this has the effect of updating the shader.

src/DynamoCoreWpf/ViewModels/Watch3D/DynamoGeometryModel3D.cs
```diff
@@ -13,24 +14,187 @@
...
+
+        protected override void OnUpdatePerModelStruct(RenderContext context)
+        {
+            base.OnUpdatePerModelStruct(context);
+            //store the entire state of all our flags in the X component and decode in the vertex shader.
+            //we'll pass this from the vertex shader to the fragment shader to determine color states.
+            //Params is a helix material builtin and maps to vParams in the shader.
+
+            //TODO this gets called whenever a render occurs, make sure it is fast.
+            modelStruct.Params.X = GenerateEnumFromState();
```
mjkkirschner: here is where the data is actually passed to the struct that helix will send to the shader. Params is a builtin attribute that is passed to the vertexShader as `vParams` - we're just using the X component - nothing else is using it. `vParams` is around as a general structure for sending data to the shaders.

src/DynamoCoreWpf/ViewModels/Watch3D/DynamoGeometryModel3D.cs
```diff
@@ -13,24 +14,187 @@
...
+        }
+    }
+
+
     /// <summary>
-    /// A Dynamo mesh class which supports the RenderCustom technique.
+    /// A Dynamo mesh class which supports sending data to our custom shader.
     /// </summary>
-    public class DynamoGeometryModel3D : MeshGeometryModel3D
+    public class DynamoGeometryModel3D : MaterialGeometryModel3D //TODO old base class okay? MaterialGeo?
```
mjkkirschner: I reverted this back to the old type before 2.9 update, no harm so far.

src/DynamoCoreWpf/ViewModels/Watch3D/DynamoGeometryModel3D.cs
```diff
@@ -40,5 +204,26 @@ public bool RequiresPerVertexColoration
...
+    {
+   
+        protected override RenderCore OnCreateRenderCore()
+        {
+            return new DynamoGeometryMeshCore();
+        }
+
+        protected override IRenderTechnique OnCreateRenderTechnique(IRenderHost host)
+        {
+            return host.EffectsManager[DynamoCustomShaderNames.DynamoCustomMeshShader];
```
mjkkirschner: here is how our shader is specified for DynamoMesh objects.

src/DynamoCoreWpf/ViewModels/Watch3D/shaderSource/CompileShaders.bat
```diff
@@ -0,0 +1,18 @@
...
+REM this is a comment 
+REM https://docs.microsoft.com/en-us/windows/win32/direct3dtools/dx-graphics-tools-fxc-syntax
+REM - /Vi - verbose include info
+REM - /0d - disable optimziations
+REM -/Zi Enable debugging information.
+REM /T fx_5_0 - shader model 5
+REM -/I ./../ include
+REM -  /Fo ..\Resources\_custom.bfx output binary
+REM file to compile Default.fx
+REM - RENAME THESE
```
mjkkirschner: I will remove this comment, rename is already done.

src/DynamoCoreWpf/ViewModels/Watch3D/shaderSource/CompileShaders.bat
```diff
@@ -0,0 +1,18 @@
...
+fxc /Vi /Od /Zi /T vs_5_0 /I ./helix_shader_defs/COMMON /Fo ..\compiledShaders\vsDynamoMesh vsDynamoMesh.hlsl
+fxc /Vi /Od /Zi /T ps_5_0 /I ./helix_shader_defs/COMMON /Fo ..\compiledShaders\psDynamoMesh psDynamoMesh.hlsl
```
mjkkirschner: fxc is the directX shader compiler.

aparajit-pratap:
@mjkkirschner can you merge all changes from `master` into `helix-upgrade` branch first before making this PR?

mjkkirschner:
@aparajit-pratap yes - I will do that now.

mjkkirschner:
also tested transformable interface - that is working as well:
Screen Shot 2020-03-13 at 2 22 21 PM

src/DynamoCoreWpf/ViewModels/Watch3D/DynamoGeometryModel3D.cs
aparajit-pratap: The helix `ModelStruct` type also has a `Color` property, which I guess could be used to pass color data to the shader?

src/DynamoCoreWpf/ViewModels/Watch3D/DynamoGeometryModel3D.cs
```diff
@@ -13,24 +14,187 @@
...
         public DynamoGeometryModel3D()
         {
         }
 
         protected override SceneNode OnCreateSceneNode()
         {
-            var node = base.OnCreateSceneNode();
-            node.OnSetRenderTechnique = (host) => { return host.EffectsManager["RenderCustom"]; };
+            var node = new DynamoMeshNode();
             return node;
```
aparajit-pratap: `return new DynamoMeshNode();`

aparajit-pratap: So by overriding this method, helix knows which custom core and custom shaders to use to render Dynamo mesh `GeometryModel3D` types?

src/DynamoCoreWpf/ViewModels/Watch3D/shaderSource/psDynamoMesh.hlsl
```diff
@@ -0,0 +1,127 @@
+
+#define MESH
+#include"Common.hlsl"
+#include"CommonBuffers.hlsl"
```
aparajit-pratap: These files are in a nested folder called `helix_shader_defs` - does it know how to find them?

src/DynamoCoreWpf/ViewModels/Watch3D/shaderSource/psDynamoMesh.hlsl
```diff
@@ -0,0 +1,127 @@
...
+		}
+
+		//TODO this is why lighting is not applied to textured geometry - we just return here...
+		//TODO use flat shade flag instead so users can choose to apply shading or not to any Display object.
+		return I;
+	}
+
+	// light emissive and ambient intensity
+	// this variable can be used for light accumulation
+	float4 I = vMaterialEmissive + vMaterialAmbient * vLightAmbient;
```
aparajit-pratap: Where are these defined - `vMaterialEmissive` etc.?

src/DynamoCoreWpf/ViewModels/Watch3D/shaderSource/psDynamoMesh.hlsl
```diff
@@ -0,0 +1,127 @@
...
+if(isIsolated && !isSelected)
+{
+	I.a = .1f;
+}
+
+
+if (requiresPerVertexColoration)
+{
+	//multiply the vert color by the light
+	I = I * input.c;
```
aparajit-pratap: Is the `c` the color coming from the helix `ModelStruct.Color` property?

src/DynamoCoreWpf/ViewModels/Watch3D/DynamoGeometryModel3D.cs
mjkkirschner: I think that is available to pass a single color for the entire model, I believe in the shader it turns into `vColor`. This is defined in the common buffer for all meshes:https://github.com/helix-toolkit/helix-toolkit/blob/76dc523c9bf928ef2c6f6e72246df28959cc0336/Source/Examples/WPF.SharpDX/CustomShaderDemo/Shaders/CommonBuffers.hlsl#L56 
 I don't see this used in our old implementations anywhere.

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
```diff
@@ -381,6 +320,19 @@ public static bool IsSpecialRenderPackage(DependencyObject element)
...
         #region utils
+        
+        /// <summary>
+        /// Sets the property value on the DynamoMeshCore - this makes its way down to our shader and invalidates the render.
+        /// </summary>
+        /// <param name="meshGeom"></param>
+        /// <param name="args"></param>
+        private static void HandleMeshPropertyChange(DynamoGeometryModel3D meshGeom, DependencyPropertyChangedEventArgs args)
+        {
+            if ((meshGeom.SceneNode.RenderCore as DynamoGeometryMeshCore) != null)
```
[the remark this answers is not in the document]
mjkkirschner: will look into this, thanks.

src/DynamoCoreWpf/ViewModels/Watch3D/DynamoEffectsManager.cs
```diff
@@ -2,38 +2,75 @@
...
 namespace Dynamo.Wpf.ViewModels.Watch3D
 {
+    internal static class DynamoCustomShaderNames {
+       internal static string DynamoCustomMeshShader ="DynamoCustomMeshShader";
+    }
+
+
     /// <summary>
-    /// The DynamoEffectsManager is loads Effects
+    /// The DynamoEffectsManager is loads shaders
```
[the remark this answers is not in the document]
mjkkirschner: will do, thanks.

src/DynamoCoreWpf/ViewModels/Watch3D/DynamoEffectsManager.cs
```diff
@@ -2,38 +2,75 @@
...
+    /// https://docs.microsoft.com/en-us/windows/win32/direct3dgetstarted/work-with-shaders-and-shader-resources
     /// </summary>
     public class DynamoEffectsManager : DefaultEffectsManager
     {
         public DynamoEffectsManager() : base() {
             AddDynamoTechniques();
         }
 
-        
+        internal class DynamoCustomMeshRenderVertexShaderDescription
```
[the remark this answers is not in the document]
mjkkirschner: definitely, I hate these names.

src/DynamoCoreWpf/ViewModels/Watch3D/DynamoEffectsManager.cs
```diff
@@ -2,38 +2,75 @@
...
+
+            public static ShaderDescription PixelShaderDynamoMeshDescription = new ShaderDescription(nameof(PixelShaderDynamoMeshDescription), ShaderStage.Pixel,
+          new ShaderReflector(), PSMeshDataSamplerByteCode);
+        }
+
+
         protected void AddDynamoTechniques()
         {
-            var custom = new TechniqueDescription("RenderCustom")
+            var dynamoCustomMeshTech = new TechniqueDescription(DynamoCustomShaderNames.DynamoCustomMeshShader)
```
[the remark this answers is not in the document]
mjkkirschner: should be able to.

src/DynamoCoreWpf/ViewModels/Watch3D/DynamoEffectsManager.cs
```diff
@@ -42,7 +79,7 @@ protected void AddDynamoTechniques()
                     {
                         ShaderList = new[]
                         {
-                            DefaultVSShaderDescriptions.VSMeshWireframe,
+                           DynamoCustomMeshRenderVertexShaderDescription.VertexShaderDynamoMeshDescription,
                             DefaultPSShaderDescriptions.PSMeshWireframe
```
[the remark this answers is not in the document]
mjkkirschner: thanks for pointing this out, I think we can get rid of the wireframe pass- I have no idea why it's here or in the old implementation, it appears unused.

src/DynamoCoreWpf/ViewModels/Watch3D/DynamoGeometryModel3D.cs
```diff
@@ -13,24 +14,187 @@
...
+        /// Used to determine if vertex colors should be displayed with shading.
+        /// </summary>
+        RequiresPerVertexColor = 32,
+        /// <summary>
+        /// Currently this flag is not used in the shader.
+        /// </summary>
+        FlatShade = 64
+    }
+
+    internal class DynamoGeometryMeshCore : MeshRenderCore
```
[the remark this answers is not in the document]
mjkkirschner: will do. The nodes and cores are related, my reading is that they are both wrappers at different levels of SharpDx scene nodes -  but following the sample custom shader uses both.

src/DynamoCoreWpf/ViewModels/Watch3D/DynamoGeometryModel3D.cs
mjkkirschner: only to render `DynamoGeometryModel3d` - not GeometryModel3D in general.

src/DynamoCoreWpf/ViewModels/Watch3D/shaderSource/psDynamoMesh.hlsl
mjkkirschner: yes - note the `/I ./helix_shader_defs/COMMON` flag in the compile script.

src/DynamoCoreWpf/ViewModels/Watch3D/shaderSource/psDynamoMesh.hlsl
mjkkirschner: these are defined in the common buffer for mesh and are set by the material that the mesh currently has assigned:
https://github.com/helix-toolkit/helix-toolkit/blob/76dc523c9bf928ef2c6f6e72246df28959cc0336/Source/Examples/WPF.SharpDX/CustomShaderDemo/Shaders/CommonBuffers.hlsl#L69

src/DynamoCoreWpf/ViewModels/Watch3D/shaderSource/psDynamoMesh.hlsl
mjkkirschner: I don't think so, the modelStruct is for the entire mesh, this C should be coming from the vertex as it's a vertex color.

https://github.com/helix-toolkit/helix-toolkit/blob/76dc523c9bf928ef2c6f6e72246df28959cc0336/Source/HelixToolkit.Native.ShaderBuilder/Common/DataStructs.hlsl#L50

mjkkirschner:
@reddyashish  @aparajit-pratap please take a look, I have addressed the current comments I believe.

src/DynamoCoreWpf/ViewModels/Watch3D/shaderSource/CompileShaders.bat
aparajit-pratap: What's the difference between the .fx shaders and the .hlsl shaders we have now?

src/DynamoCoreWpf/ViewModels/Watch3D/AttachedProperties.cs
```diff
@@ -16,41 +16,16 @@ public static class AttachedProperties
 
         // handles determining color of elementGeometry3Ds when any property is set false
         // as the state of other properties must be checked to determine the correct color / material.
-        private static void OnPropertySetFalse(DependencyObject obj)
+        private static void OnPoint_Line_PropertySetFalse(DependencyObject obj)
```
reddyashish: Just a minor thing but I think it is better to avoid underscore character for function names unless they are test functions. Makes it easy to differentiate. 
https://stackoverflow.com/questions/9782084/naming-conventions-in-c-sharp-underscores
https://docs.microsoft.com/en-us/dynamicsax-2012/developer/naming-conventions-underscores.

mjkkirschner: thanks @reddyashish I will change the name.

reddyashish:
@mjkkirschner Just one small comment but looks good to me.

src/DynamoCoreWpf/ViewModels/Watch3D/DynamoGeometryModel3D.cs
aparajit-pratap: Yeah, that's what I meant. Thanks.

src/DynamoCoreWpf/ViewModels/Watch3D/shaderSource/CompileShaders.bat
mjkkirschner: .fx files are effects files - a higher level directx thing that helix no longer uses.
https://gamedev.stackexchange.com/questions/129827/how-do-hlsl-files-work-together-compared-to-fx

they are kind of like wrappers for pixel and vertex shaders that tie them together.

mjkkirschner: also - 
https://github.com/Microsoft/FX11

effects and the older shader profile are deprecated...

aparajit-pratap:
@mjkkirschner thanks for addressing the review comments and answering my questions. I think this is a very clean approach. Excellent work 👍

mjkkirschner:
@reddyashish comments addressed - are we good to go?

PR #10466, opened 2020-03-13
Headlight bound to Camera LookDirection

mjkkirschner:
### Purpose

This PR fixes the binding for the headlight to the Camera's look direction and adds a test.
In this gif I have disabled the normal light, and changed the headlight to red - note that the sphere is red from every perspective as the headlight is moving.
 The headlight has not been made red in this PR - thats just for easy debugging.

headlight

#### Note this PR is based on the mesh-shader branch so it will be easier to review after that PR is merged.
https://github.com/DynamoDS/Dynamo/pull/10462

### Declarations


src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -1285,12 +1285,10 @@ private void InitializeHelix()
                 Name = HeadLightName
             };
 
-            //TODO verify this binding is working.
             headLight.SetBinding(
                 DirectionalLight3D.DirectionProperty, 
-                new Binding("LookDirection") {
-                    Source = this,
-                    Path = new PropertyPath("Camera.LookDirection")
+                new Binding(nameof(PerspectiveCamera.LookDirection)) {
```
mjkkirschner: this is the important change.

src/VisualizationTests/HelixWatch3DViewModelTests.cs
```diff
@@ -650,6 +646,23 @@ public void Watch3D_Disconnect_Reconnect_CorrectRenderings()
             Assert.AreEqual(5, ((HelixWatch3DViewModel)view.DataContext).Element3DDictionary.Count());
         }
 
+        [Test]
+        public void HelixWatch3dViewModel_HeadLight_Camera_HaveSameLookVector()
```
mjkkirschner: heres the test.

mmisol:
I took a look at it using a simple example I took from the primer. While I can definitely notice the light now, it still looks darker than in master. Here is a visual comparison:

This PR:
primer6_2020-03-13_06-17-56

master:
Capture

A little lighter would be nicer IMO but it is a different issue so for the direction part looks good to me.

mjkkirschner:
@mmisol can you share the graph or primer link?

mmisol:
Surprisingly, there is not a link to download it. It's very simple anyway. Here it goes attached as a text file. 

[primer6.txt](https://github.com/DynamoDS/Dynamo/files/4331913/primer6.txt)

mjkkirschner:
Thanks @mmisol - to me - it looks like something is wrong with Autodesk.Designscript.Geometry.Mesh object rendering in the shader - perhaps this object has normals which are facing unexpected directions - I will take a look, thanks.

mjkkirschner:
@mmisol thanks so much for catching this - it helped me find a class of geometry I missed - those single polygon thickness objects like surfaces where normals are not all aligned.

for example:
Screen Shot 2020-03-14 at 1 07 18 AM

and you mesh above...

note the mesh that is black is not colored black - it's normals are just all pointing down.

I've got these fixed now:

Screen Shot 2020-03-14 at 1 36 43 AM

It's a similar fix to what was in the original shader but should be even faster and works in more cases (builtin since directX 10)

the old shader did this:
https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/ViewModels/Watch3D/Shaders/Custom.fx#L56

now we use `SV_IsFrontFace` to flip the normal instead which doesn't break in funny ways and requires no dot product.

src/DynamoCoreWpf/ViewModels/Watch3D/shaderSource/psDynamoMesh.hlsl
```diff
@@ -0,0 +1,136 @@
...
+	// get per pixel vector to eye-position
+	float3 eye = normalize(vEyePos - input.wp.xyz);
+
+	// To support two-sided surface lighting, we flip 
+	// the normal of the surface if it's facing
+	// away from us. There are many tessellated surfaces that have
+	// normals that are not as expected.
+	// SV_IsFrontFace is a system value (GPU sets this for us) you can read about here:
+    //https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-semantics
+	 input.n = isFrontFacing ? input.n : -input.n;
```
mjkkirschner: here is the addition to the shader.

mmisol:
Looks perfect now. Thank you for fixing it @mjkkirschner

aparajit-pratap:
> note the mesh that is black is not colored black - it's normals are just all pointing down.

@mjkkirschner this means that we need to fix the mesh/geometry normals, doesn't it? Could this be related to the `surface.flipnormal` issue we have in LibG?

mjkkirschner:
@aparajit-pratap - no thats not exactly, what it means. The normals might be perfectly correct, but the user expects the object to be well lit from all sides.

In @mmisol's mesh example, the triangle winding order made all the normals face down, so the object was dark, but the lighting that was previously implemented in Dynamo always tries to flip the display normal so that the object appears like from all angles, its like faking double sided faces.

Here I'm doing the same thing, but without a dot product, it's the recommended way if you are supporting dx10 and up to display double sided surfaces in a shader without actually having double sided surfaces.

reddyashish:
@mjkkirschner Looks good to me.

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
QilongTang: Would you put comments why it is important? Not setting **Path** or?

QilongTang:
LGTM

PR #10476, opened 2020-03-19
[DYN-2540] Avoid reloading libraries while loading custom node package

aparajit-pratap:
### Purpose

Noticed unnecessary reload of libraries taking place even while loading custom node packages. This change guards against repeated reload of libraries if there are no new library (DLL) paths. 

Removed changes in #9957 as they are no longer required.

### Declarations

### FYIs

@DynamoDS/dynamo 
@reddyashish


src/DynamoCore/Engine/EngineController.cs
```diff
@@ -522,7 +522,8 @@ private void OnLibraryLoaded()
         /// </summary>
         private void LibraryLoaded(object sender, LibraryServices.LibraryLoadedEventArgs e)
         {
-            OnLibraryLoaded();
+            if(e.LibraryPaths.Any())
```
mjkkirschner: @aparajit-pratap so this will not have any paths if a package only containing custom nodes is loaded? What about a package containing nodes and zt assemblies?

Are there are tests that should be added here to check this event is not raised?

aparajit-pratap: A package containing custom nodes and ZT assemblies will have relevant paths as expected, in which case we would need to reload all libraries including the new ones in that package. 

I can try and come up with some tests.

aparajit-pratap:
@mjkkirschner @reddyashish after much thinking and testing I have decided to put back in Ashish's changes (with the `EngineController.DisableRun` flag) back in as there are regressions with mixed packages (custom nodes and ZT nodes) if I pull them out. So for now, I'm sticking with my initial improvement, which is not to reload libraries for a custom node-only package and that's it. 
Also, I didn't find a way to test if the libraries are not getting reloaded for a custom node-only package so I'm leaving this PR as is. Please let me know if this is good to merge as is.

mjkkirschner:
@aparajit-pratap so you think the existing tests guard against any regressions in loading other package types?

aparajit-pratap:
> @aparajit-pratap so you think the existing tests guard against any regressions in loading other package types?

There could be a test to load a "mixed package" type as currently, there's no test for this. Plus I'm not sure if there are tests for loading missing packages by adding new package paths in settings.  This latter workflow was broken for mixed packages in my manual testing as I was simulating the workspace dependency viewer as I didn't know of such package types on the PM other than Archi-lab. 
I tested archi-lab (which is a mixed package like you pointed out) but it loads fine in D4R, presumably because D4R is single-threaded. All-in-all it wasn't a good test.

aparajit-pratap:
I'll add the mixed package test for now and we can have a separate task for adding a workflow test for adding new package paths later?

mjkkirschner:
@aparajit-pratap sounds good to me, thanks.

aparajit-pratap:
@mjkkirschner I've added the test. Whenever you're free to take another look. Thanks!

mjkkirschner:
@aparajit-pratap looks like some failing tests on the parallel run:
https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/195/testReport/
@alfredo-pozo - 2 of these tests seem to only fail on the parallel job, any thoughts?

mjkkirschner:
It looks good, but lets make sure the test failures are not real.

aparajit-pratap:
> @aparajit-pratap can you confirm everything is passing on the regular self serve?

The old self-serve passes.

mjkkirschner:
it could be that the test requires an assembly be in the same process that is not loaded during a parallel run because a previous test did not load it.

aparajit-pratap:
So @mjkkirschner are you suggesting we don't merge this PR?

mjkkirschner:
@aparajit-pratap I'm okay merging this because it won't hold up this build, but it may break all further runs of the parallel self serve for every other PR - so I would like to look at that before or immediately after this gets merged.

aparajit-pratap:
I'm not sure about how parallel self-serve works. That's not a real use-case for Dynamo usage. There's some state that is shared between tests that are running in parallel. Package loading is very complex in Dynamo in and of itself so I'm not sure if we should support parallel use cases.

mjkkirschner:
@aparajit-pratap - theres more state shared between the regular tests - the parallel tests are actually more like a user scenario where less libraries are loaded into each session.

They are a good test to show us when tests are actually not separable. They do not run on different threads, but isolated processes.

alfredo-pozo:
@mjkkirschner Related to your question, by checking those tests, these are similar to other ones that are running with success and that are using System.Thread. The only thing that comes to my mind since we have  Parallel (C#) on Parallel (PS Workflow),  is that we could try to use a newer Powershell version since the hosts have 5.1 and the mos recent version is 7 and I remember they update some features related to For-Each Parallel

PR #10479, opened 2020-03-19
Avoid rendering Gizmo multiple times on rotate

mmisol:
### Purpose

This PR aims to fix the problem described in the title, which occurs in the `helix-upgrade` branch. 

When a `Point.ByCoordinates` node is placed on the canvas with some unfilled input this triggers the directManipulation extension to draw gizmos that let the user manipulate this point in space directly. When the user rotates the camera when the point node is still selected and the gizmos are drawn the performance is terrible, and the camera cannot be rotated smoothly.
 
Jan-27-2020 16-35-44

The issue is caused by a change of behavior in helix when upgrading from v2015 is that the `CameraChanged` event gets fired several times, once for each individual property of `Camera` that changed.

In order to avoid rendering the `Gizmo` several times, which leads to choppy camera rotation, we limit the rendering of the `Gizmo` to only be done when the position of the camera changed.

Screen-Recording-2020-03-19-at-12 51 13-PM

### Declarations

### Reviewers

### DIFF
--- a/src/DynamoManipulation/Gizmo.cs
+++ b/src/DynamoManipulation/Gizmo.cs
@@ -171,10 +171,17 @@ namespace Dynamo.Manipulation
 
         private void OnViewCameraChanged(object o, RoutedEventArgs routedEventArgs)
         {
-            cameraPosition = routedEventArgs.Source as Point3D?;
-
-            // Redraw Gizmos
-            Redraw();
+            var newCameraPosition = routedEventArgs.Source as Point3D?;
+            // A change of behavior in helix when upgrading from v2015 is that the CameraChanged event gets fired
+            // several times, once for each individual property of Camera that changed.
+            // In order to avoid rendering the Gizmo several times, which leads to choppy camera rotation, we limit the
+            // rendering of the Gizmo to only be done when the position of the camera changed.
+            if (cameraPosition != newCameraPosition)
+            {
+                cameraPosition = newCameraPosition;
+                // Redraw Gizmos
+                Redraw();
+            }
         }
 
         public abstract bool HitTest(Point source, Vector direction, out object hitObject);
--- a/src/DynamoCoreWpf/ViewModels/Watch3D/DynamoGeometryModel3D.cs
+++ b/src/DynamoCoreWpf/ViewModels/Watch3D/DynamoGeometryModel3D.cs
@@ -64,44 +64,44 @@ namespace Dynamo.Wpf.ViewModels.Watch3D
         /// <summary>
         /// Is this model Frozen.
         /// </summary>
-        public bool IsFrozenData { get => isFrozenData; internal set { SetAffectsRender(ref isFrozenData, value); } }
+        public bool IsFrozenData { get { return isFrozenData; } internal set { SetAffectsRender(ref isFrozenData, value); } }
 
         private bool isSelectedData;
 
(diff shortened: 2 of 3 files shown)


src/DynamoManipulation/Gizmo.cs
```diff
@@ -171,10 +171,17 @@ private void Redraw()
...
-            cameraPosition = routedEventArgs.Source as Point3D?;
-
-            // Redraw Gizmos
-            Redraw();
+            var newCameraPosition = routedEventArgs.Source as Point3D?;
+            // A change of behavior in helix when upgrading from v2015 is that the CameraChanged event gets fired
+            // several times, once for each individual property of Camera that changed.
+            // In order to avoid rendering the Gizmo several times, which leads to choppy camera rotation, we limit the
+            // rendering of the Gizmo to only be done when the position of the camera changed.
+            if (cameraPosition != newCameraPosition)
```
mjkkirschner: hmm 
https://docs.microsoft.com/en-us/dotnet/api/system.windows.media.media3d.point3d.op_equality?view=netframework-4.8#System_Windows_Media_Media3D_Point3D_op_Equality_System_Windows_Media_Media3D_Point3D_System_Windows_Media_Media3D_Point3D_

do you think it's worth to do an epsilon tolerant equals?

QilongTang:
Good improvement! LGTM

src/DynamoManipulation/Gizmo.cs
mjkkirschner: this is what I'm referring to:
```
Two Point3D structures are equal if the values of their X, Y , and Z properties are the same.

Because Double values can lose precision when arithmetic operations are performed on them, a comparison between two Point3D instances that are logically equal might fail.

```

mmisol: Due to the nature of the issue, there shouldn't be a need to be epsilon tolerant. Basically what's happening is:
1. Camera.LookDirection is set -> OnViewCameraChanged (ignored as Position is same)
2. Camera.Position is set -> OnViewCameraChanged (not ignored as Position is different)
3. Camera.UpDirection is set -> OnViewCameraChanged (ignored as Position is same)

QilongTang: Would the default **Equals** sufficient? https://docs.microsoft.com/en-us/dotnet/api/system.windows.media.media3d.point3d.equals?view=netframework-4.8#System_Windows_Media_Media3D_Point3D_Equals_System_Object_

mmisol: @QilongTang Yes, we are ignoring case 1 and 3 in my post above by using (X, Y, Z) equality. The `==` operator is overloaded to behave exactly as `Equals`.

mjkkirschner: cool, thanks for the explanation @mmisol

aparajit-pratap:
BTW, just curious, do you have screenshots of the dot trace profiling before and after the helix update?

mmisol:
@aparajit-pratap I did not take a screenshot, sorry

mjkkirschner:
you may be able to find the session here: `C:\Users\[username]\AppData\Local\JetBrains\dotMemory\[version]\Workspaces` - useful anyway to check because sometimes they take a bunch of hd space.

PR #10481, opened 2020-03-19
special render packages should render with materials colors only

mjkkirschner:
Screen Shot 2020-03-19 at 3 16 47 PM
add test
update shader

### Purpose

enable the `SpecialRenderPackage` flag and use it in the shader.

### Declarations


src/DynamoCoreWpf/ViewModels/Watch3D/shaderSource/psDynamoMesh.hlsl
```diff
@@ -69,6 +69,11 @@ float4 main(PSInputCustom input, bool isFrontFacing : SV_IsFrontFace) : SV_Targe
     //https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-semantics
 	 input.n = isFrontFacing ? input.n : -input.n;
 
+	 // if this is a special render package - it should render with the material colors, ambient light
+	 // and not be directionally lit.
+	 if (isSpecialRenderPackage) {
```
mmisol: In line 45 it says
```
IsSpecialRenderPackage = 8, //not used
```
Should we remove the `//not used` part now?

mjkkirschner: thanks @mmisol will do.

mjkkirschner:
Believe all comments regarding this flag are now up to date.

aparajit-pratap:
Looks prettier :)

aparajit-pratap:
Was there any visual change apart from the gizmo since the directional light, grid and grid axes are also special render packages?

mjkkirschner:
@aparajit-pratap there should not be, as 
1. the light is not a render package AFAIK - it has no geometry
2. grid and axes are lines which don't use this shader.

aparajit-pratap:
In that case, I wonder, why we are doing this: https://github.com/DynamoDS/Dynamo/blob/1464172dd3a1528deca7746a226cd5fabf0dbe13/src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs#L1226

mjkkirschner:
@aparajit-pratap yep, not sure, it doesn't seem to make sense, it's not geometry and it's not a DynamoGeometryModel3d - so this likely doesn't have much of an effect anywhere...

mjkkirschner:
@aparajit-pratap it's possible it might make a difference in the sorting algorithm - https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs#L2240

but even there, you wouldn't need to sort a light... in any case, this should only effect DynamoModelGeometry3d objects of which none of these (light,grid,axes) are AFIK.

aparajit-pratap:
LGTM

PR #10487, opened 2020-03-20
add 3d graphics image comparison tests

mjkkirschner:
### Purpose

PR does the following:
* lessens intensity of the overhead static direction light as I noticed a highlight that was bleeding out the details of some geometry, seed too bright to me.

* adds a set of pixel by pixel image comparison tests using a "snapshot" concept - using a define to update the reference images. We'll likely need to see how these workout on different machines or on the CI - but they pass locally - we may also want to increase the tolerance - ie let some % of pixels have different color values.

* here are all the reference images that are compared/generated.
Screen Shot 2020-03-20 at 7 29 26 PM

### Declarations


src/VisualizationTests/HelixImageComparisonTests.cs
```diff
@@ -0,0 +1,428 @@
...
+            var path = generateTestDataPathFromTest(MethodBase.GetCurrentMethod().Name);
+            var bitmapsource = DynamoRenderBitmap(BackgroundPreview.View, 1024, 1024);
+#if UPDATEIMAGEDATA
+            UpdateTestData(path, bitmapsource);
+#endif
+            var refimage = new BitmapImage(new Uri(path));
+            var refbitmap = BitmapFromSource(refimage);
+            var newImage = BitmapFromSource(bitmapsource);
+
+            compareImageColors(refbitmap, newImage);
```
QilongTang: Hi @mjkkirschner Given this function will always be called at the end of each test, would it be possible that we put it as part of test tear down? Just curious if we can reduce some code and ease the writing..

QilongTang:
@mjkkirschner Do we delete legacy screenshot before a new tests run or it happens automatically?

mjkkirschner:
Theres no need to delete the legacy screenshots - they are the reference images - no data is written to disk unless you uncomment the `//#define UPDATEIMAGEDATA` line - 

images are compared in memory.

if you uncomment that line, then the reference images are replaced - but this should never be done on the test machine - only locally.

Did that help?

src/VisualizationTests/HelixImageComparisonTests.cs
mjkkirschner: hmm... I'm not sure that will be clear to those who are not well aquatinted with Nunit - also different Nunit versions behave differently, so I'd be hesitant to try that - not even sure you can fail a test in a teardown but I can look.

Maybe I can just simplify the comparison a bit more into a single utility function so theres less repeated code.

src/VisualizationTests/HelixImageComparisonTests.cs
```diff
@@ -0,0 +1,428 @@
...
+        /// <param name="width">The width.</param>
+        /// <param name="height">The height.</param>
+        /// <returns>A bitmap.</returns>
+        private static BitmapSource DynamoRenderBitmap(
+             Viewport3DX view, int width, int height)
+        {
+            var w = view.RenderHost.ActualWidth;
+            var h = view.RenderHost.ActualHeight;
+            view.RenderHost.Resize(width, height);
+            var rtb = view.RenderBitmap();
```
aparajit-pratap: Is it worth doing this instead of this: https://github.com/DynamoDS/Dynamo/blob/1284ce3b59df7d7f5dcd6df89f238522ae32f199/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs#L1053 to export the background preview as an image?

src/VisualizationTests/HelixImageComparisonTests.cs
```diff
@@ -0,0 +1,428 @@
...
+        #region SpecialRenderPackages
+        [Test]
+        public void directManipulator()
+        {
+            var pointOriginNode =
+            new DSFunction(Model.LibraryServices.GetFunctionDescriptor("Point.ByCoordinates@double,double,double"));
+
+            var command = new Dynamo.Models.DynamoModel.CreateNodeCommand(pointOriginNode, 0, 0, true, false);
+            Model.ExecuteCommand(command);
+            RunCurrentModel();
```
aparajit-pratap: You could also create a dyn file with a Point node in it and add that node to the selection like you're doing here in order to trigger the creation of the manipulator.

src/VisualizationTests/HelixImageComparisonTests.cs
aparajit-pratap: I think it would be good to refactor these lines of code in a common utility method that can be called in each one of these tests.

src/VisualizationTests/HelixImageComparisonTests.cs
mjkkirschner: that method does not correctly resize the viewport - we want to resize the viewport *before* rendering so that we can accurately compare the pixels. Theres also no reason to export a 32bit image - we dont care about alpha channels here.

src/VisualizationTests/HelixImageComparisonTests.cs
mjkkirschner: do you have a preference? I just felt it was less maintenance to do this - avoid another .dyn with a bunch of repeated json in it for one point.

src/VisualizationTests/HelixImageComparisonTests.cs
mjkkirschner: definitely, will do.

src/VisualizationTests/HelixImageComparisonTests.cs
aparajit-pratap: Which is why I asked the question :)

src/VisualizationTests/HelixImageComparisonTests.cs
aparajit-pratap: I don't have a preference. I made the note just for your info as an alternative to creating the manipulator in the test.

src/VisualizationTests/HelixImageComparisonTests.cs
mjkkirschner: ahh, you mean use the new method in the save functionality? Sorry, I thought you meant the other way around.

I can try updating to helix 2.11 - then this method should be updated in both places, and we won't need to keep a custom implementation.

Let me give that a shot.

src/VisualizationTests/HelixImageComparisonTests.cs
```diff
@@ -0,0 +1,428 @@
...
+
+        #region meshes
+        [Test]
+        public void StandardMeshGeometryRender()
+        {
+            OpenVisualizationTest(@"imageComparison\spherecolors.dyn");
+            RunCurrentModel();
+            var path = generateTestDataPathFromTest(MethodBase.GetCurrentMethod().Name);
+            var bitmapsource = DynamoRenderBitmap(BackgroundPreview.View, 1024, 1024);
+#if UPDATEIMAGEDATA
```
reddyashish: Will this flag be set to false when we run the whole dynamo test fixture?

mjkkirschner: yes, this will only be set true locally, think of this like the update snapshot functionality in react snapshot testing.

src/VisualizationTests/HelixImageComparisonTests.cs
mjkkirschner: I was wrong, this fix is in helix 2.12 that is not released yet, so we'll need to do this workaround until then - for the other function, we actually don't want to use this - there are two overloads of RenderBitmap - and that one uses the user's current window resolution so it should stay as is.

src/VisualizationTests/HelixImageComparisonTests.cs
mjkkirschner: this is done, I've refactored this and made each test much smaller.

mjkkirschner:
@QilongTang @aparajit-pratap @reddyashish PTAL - addressed review comments.

aparajit-pratap:
LGTM!

QilongTang:
LGTM with the much shorter tests

PR #10493, opened 2020-03-23
[DYN-1815] Dynamo package install invalid package (sad path) error and crash

reddyashish:
### Purpose

This PR adds a fix to the package manager crash after an invalid package is downloaded. If the pkg is null, we simply do not want to add it to the LocalPackages collection. 

package manager fix

The "autodesk steel connections 2020 test 4" package has an additional parent folder steel-pkg and it looks for the pkg.json file in a different location. An error message is shown in the console after the package is downloaded notifying the user that it is an invalid package. 

Screen Shot 2020-03-24 at 8 46 35 AM

When the user tries to access the manage package window, the crash is avoided and we do not show this package as the installed package because of the invalid package structure.  
When Dynamo is restarted again, it shows a notification that this package was not loaded due to the above error. 

### Declarations

### Reviewers


QilongTang:
Would we be able to somehow unit test this?

mjkkirschner:
@reddyashish couldn't the unit test simply be adding this malformed package to the repo, and attempting to load it during a test and asserting non exception is thrown/no crash?

reddyashish:
@mjkkirschner Yes, I am checking if we can catch the exception and assert it.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
```diff
@@ -724,9 +723,16 @@ internal void DownloadAndInstall(PackageDownloadHandle packageDownloadHandle, st
                         {
                             var p = Package.FromDirectory(dynPkg.RootDirectory, DynamoViewModel.Model.Logger);
 
-                            pmExtension.PackageLoader.LoadPackages(new List<Package> {p});
-
-                            packageDownloadHandle.DownloadState = PackageDownloadHandle.State.Installed;
+                            if (p != null)
+                            {
+                                pmExtension.PackageLoader.LoadPackages(new List<Package> { p });
```
reddyashish: This fix is on the Client side and we show a message that the downloaded package is Invalid instead of showing a message regarding the null reference exception.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -310,7 +310,11 @@ public void LoadPackages(IEnumerable<Package> packages)
 
             foreach (var pkg in enumerable)
             {
-                TryLoadPackageIntoLibrary(pkg);
+                // If the pkg is null, then don't load that package into the Library.
+                if (pkg != null)
```
reddyashish: This additional check is needed to assert the tests because the testing workflow bypasses the user workflow of downloading the actual package and loading it into Dynamo after being downloaded.

test/DynamoCoreWpfTests/PackageManagerUITests.cs
```diff
@@ -233,6 +254,27 @@ public void PackageSearchDialogSearchBoxPromptTextWhenNotSyncing()
...
+            {
+                LoadPackage(packageDirectory);
+            }
+            catch (Exception e)
+            {
+                Console.WriteLine("Failed to load the package: "+ e);
+            }
+
+            var loader = GetPackageLoader();
+            Assert.AreEqual(loader.LocalPackages.Count(), 0);
```
reddyashish: Asserting that the actually packages count doesn't increase and making sure that the Manage Packages window shows up.

reddyashish:
Added the test and comments for the changes.

src/DynamoPackages/PackageLoader.cs
aparajit-pratap: In that case, why don't we simply have the check here in one place instead of having it in 2 places, one on the client and one here?

test/DynamoCoreWpfTests/PackageManagerUITests.cs
```diff
@@ -15,6 +17,25 @@ public class PackageManagerUITests : SystemTestBase
...
+        protected PackageLoader GetPackageLoader()
+        {
+            var extensions = Model.ExtensionManager.Extensions.OfType<PackageManagerExtension>();
+            if (extensions.Any())
+            {
+                return extensions.First().PackageLoader;
+            }
+
+            return null;
+        }
```
aparajit-pratap: If I remember correctly there are similar utility functions already available in other test fixtures. If so, you could try reusing them here or have them all in one place if they belong to the same class hierarchy.

src/DynamoPackages/PackageLoader.cs
reddyashish: An exception was being thrown in the PackageManagerClientViewModel.cs and null reference error message was being shown in the package search dialog. Instead we can show a more clear message to the user. I thought it would be better/cleaner to handle it on the top level as we do not want to make extra function calls. Let me know what you think about it.

test/DynamoCoreWpfTests/PackageManagerUITests.cs
reddyashish: These functionss are only defined in the DynamoModelTestBase class and this PackageManagerUITests.cs inherits only the SystemTestBase. I did not notice any UI tests using these utility functions.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
reddyashish: By the time Package.FromDirectory function is called, the package is already downloaded locally. In this method, it first looks for the pkg.json file in the root directory. If it doesn't find it, an exception is thrown. Then it tries to create a package object from the json header file. If any of this fails or if no pkg could be created from the json header file, then pkg would be null. From this, I think we can say that it is an invalid package.

reddyashish:
Yes, the self-serve passed. I have added this test package to the core folder https://github.com/DynamoDS/Dynamo/tree/master/test/core, as this test fixture has other packages in this location. 
The check for total number of packages is for the packages that are saved in here: https://github.com/DynamoDS/Dynamo/tree/master/test/pkgs. So no changes needed around that.

aparajit-pratap:
> Yes, the self-serve passed. I have added this test package to the core folder https://github.com/DynamoDS/Dynamo/tree/master/test/core, as this test fixture has other packages in this location.
> The check for total number of packages is for the packages that are saved in here: https://github.com/DynamoDS/Dynamo/tree/master/test/pkgs. So no changes needed around that.

I'm sorry, did I miss something? Why aren't you adding the test package to the same location where all the other test packages are?

reddyashish:
@aparajit-pratap We have test packages saved in https://github.com/DynamoDS/Dynamo/tree/master/test/core and https://github.com/DynamoDS/Dynamo/tree/master/test/pkgs locations. As this test is in the DynamoCoreWPFTests fixture, I just placed it in the core folder. I am not sure why we have 2 folders for the test packages but let me know if you have any preference.

reddyashish:
Did not have to update the tests that assert the total number of "pkgs" because this package is not a valid one. So, it is not being added to the package loader.

PR #10496, opened 2020-03-23
Utilizing the newly exposed endpoint in Dynamo, to get only the maintainers of a package.

zeusongit:
### Purpose

[DYN-2515](https://jira.autodesk.com/browse/DYN-2515)
- Used the newly exposed get maintainer endpoint to check author of a package for un/deprecation.
- Also [DYN-2437](https://jira.autodesk.com/browse/DYN-2437), all maintainers now have un/deprecation access for the packages they own. 

### Reviewers

@DynamoDS/dynamo


zeusongit:
test

src/DynamoPackages/PackageManagerClient.cs
```diff
@@ -99,6 +99,23 @@ internal PackageHeader GetPackageHeader(IPackageInfo packageInfo)
...
+        /// <summary>
+        /// Gets maintainers for a specific package
+        /// </summary>
+        /// <param name="packageInfo"></param>
+        /// <returns></returns>
+        internal PackageHeader GetPackageMaintainers(IPackageInfo packageInfo)
+        {
+            var header = FailFunc.TryExecute(() =>
+            {
+                var nv = new GetMaintainers("dynamo", packageInfo.Name);
```
mjkkirschner: this is the engine name? Should this be a paramter?

src/DynamoPackages/PackageManagerClient.cs
```diff
@@ -99,6 +99,23 @@ internal PackageHeader GetPackageHeader(IPackageInfo packageInfo)
...
         }
 
+        /// <summary>
+        /// Gets maintainers for a specific package
+        /// </summary>
+        /// <param name="packageInfo"></param>
+        /// <returns></returns>
+        internal PackageHeader GetPackageMaintainers(IPackageInfo packageInfo)
+        {
+            var header = FailFunc.TryExecute(() =>
```
mjkkirschner: this is a package header? Or is this a user name?

src/DynamoPackages/PackageManagerClient.cs
```diff
@@ -243,7 +260,7 @@ internal PackageManagerResult Undeprecate(string name)
         internal bool DoesCurrentUserOwnPackage(Package package,string username) 
         {
             var pkg = new PackageInfo(package.Name, new Version(package.VersionName));
-            var header = GetPackageHeader(pkg);
+            var header = GetPackageMaintainers(pkg);
```
mjkkirschner: same question about this variable name.

mjkkirschner:
whats up with the failing build here?

mjkkirschner:
I think you'll need to publish your changes to Greg.dll and update Dynamo to use that new version of Greg.dll before bringing this in.

QilongTang:
@mjkkirschner Yes we discussed the failing build at standup

src/DynamoPackages/PackageManagerClient.cs
zeusongit: part of the header, it will have all the maintainers.

src/DynamoPackages/PackageManagerClient.cs
zeusongit: I guess renaming this variable to `maintainers` will make sense

zeusongit: `maintainers` and `maintainer` are already used in the inline query below this, how about `mnt`?

mjkkirschner: sounds good.

mjkkirschner:
@QilongTang curious how would we store our dynamo nuget api key for github to access if we wanted to publish the nuget from a github action?

QilongTang:
@mjkkirschner We will need to store nuget key into the repo setting => secrect and use the nuget publish action https://github.com/marketplace/actions/publish-nuget. If you would like to automate this, can you file it? I can grab it later

test/DynamoCoreWpfTests/DynamoCoreWpfTests.csproj
```diff
@@ -45,9 +45,8 @@
     <ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>None</ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="Greg, Version=1.2.7240.28357, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\..\src\packages\Greg.1.2.7240.28357\lib\net47\Greg.dll</HintPath>
-      <Private>True</Private>
```
mjkkirschner, line 50: @zeusongit - these changes from private = true to nothing default to false -when this flag is not set to anything it is like setting CopyLocal = true - which means the dll gets copied to the bin directory, I think in most cases we only want this dll to be copied to the bin folder - not any sub folders, so it looks to me like we have let DynamoPackages.csproj copy this file and set all other references to copyLocal False - ie Private = true...

long story short - can you set the other references to CopyLocal = false and leave DynamoPackages reference to Greg to CopyLocal = true?

zeusongit: Ok updated.

mjkkirschner:
LGTM

PR #10497, opened 2020-03-23
target shader model 4 to avoid crash when user does not have GPU.

mjkkirschner:
### Purpose

compile shaders with shadermodel 4 profile targets:
see this bug report for more info:
https://github.com/helix-toolkit/helix-toolkit/issues/1328

after this change there is no more crash, but the graphics without a GPU don't appear correct- still it is an improvement over crashing.

If the user has a GPU, graphics appear correct, and image comparison tests all pass (generated with shader model 5)

should be considered after:
https://github.com/DynamoDS/Dynamo/pull/10487

### Declarations


src/DynamoCoreWpf/ViewModels/Watch3D/shaderSource/CompileShaders.bat
```diff
@@ -12,7 +12,7 @@ REM -/I ./../ include
 REM -  /Fo  output binary path
 REM file to compile Default.fx
 
-fxc /Vi /Od /Zi /T vs_5_0 /I ./helix_shader_defs/COMMON /Fo ..\compiledShaders\vsDynamoMesh vsDynamoMesh.hlsl
-fxc /Vi /Od /Zi /T ps_5_0 /I ./helix_shader_defs/COMMON /Fo ..\compiledShaders\psDynamoMesh psDynamoMesh.hlsl
+fxc /Vi /Od /Zi /T vs_4_0 /I ./helix_shader_defs/COMMON /Fo ..\compiledShaders\vsDynamoMesh vsDynamoMesh.hlsl
```
mjkkirschner: please note this is the only real change on this branch as this PR is based on the imageComparison branch.

mjkkirschner:
Screen Shot 2020-03-23 at 4 38 42 PM
Screen Shot 2020-03-20 at 8 55 28 PM

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -150,7 +150,7 @@ public class HelixWatch3DViewModel : DefaultWatch3DViewModel
         private DirectionalLight3D directionalLight;
         private DirectionalLight3D headLight;
 
-        private readonly Color4 directionalLightColor = new Color4(0.9f, 0.9f, 0.9f, 1.0f);
+        private readonly Color4 directionalLightColor = new Color4(0.7f, 0.7f, 0.7f, 1.0f);
```
aparajit-pratap: Could you explain the reason for changing this?

mjkkirschner:
@aparajit-pratap - I haven't tested all types of rendering on master (helix 2015), but I know that it did not crash.

mmisol:
@mjkkirschner The change looks good to me. Should we should keep track of the issue to further investigate the cause and try to find a better fix?

mjkkirschner:
@mmisol, yes, I'm going to file a followup, hopefully it can be fixed by helix, but it may turn out to be a WARP issue (and thus a bug report to MS) it's something incorrect in the shader - I ran it under the directx debug layer and there were no errors raised - only warnings that seemed harmless - 😞

aparajit-pratap:
Hmm, don’t seem to be able to find an “approve changes” button on the iOS app. So LGTM.

mjkkirschner:
will fix the conflicts and get this in tomorrow.

mjkkirschner:
also, just for some more info it was apparently an issue in the way helix constructed the renderer device for use by helix:
https://github.com/helix-toolkit/helix-toolkit/pull/1329

But, the additional information shared by the maintainers indicates that shader model 5 requires directx11 - if we target shader model 4 - dynamo should run on more systems including vmware fusion - which only supports DX 10.1 - my feeling is if the visuals are identical and we are not using any features of shader model 5 (compute shaders) then lets still get this in. '
My plan is to fix the conflicts now and merge.

PR #10506, opened 2020-03-26
fix shader interpolation and make image comparison tests more tolerant.

mjkkirschner:
### Purpose

This PR does the following:

* in the shader - do not pass our flags from the vertex stage to the fragment stage - the data gets interpolated and no longer can be decoded - instead just get it from the common buffer - this is faster, makes more sense, and lets us get rid of our custom input struct. yay!

* It now works on 
     * Parallels (forwarding calls to apple metal)
     * and using WARP - 
     * and on a real machine with an Nvidia Quadro 2000m.
* adds two methods for making the image tests more tolerant
    * adds a color distance metric which calculates euclidean distance in RGB space from the actual to expected color - this is a wide tolerance, the accepted distance is 128 - thats half the color space.
    * next a percentage difference is calculated, for all pixels which failed the first assertion, those pixels are counted against the total pixel count to calculate a final percent difference between images.

This let's us spot differences like this:
Screen Shot 2020-03-26 at 7 02 47 PM

but this still passes, even though there are many pixel to pixel differences, the colors are still within the same space:
Screen Shot 2020-03-26 at 7 03 21 PM

to my eye these are almost the same - but of course alpha blending in different renderers will produce slightly different results, order matters etc. 

These tests now pass on the dynamo performance machine.

### Declarations


src/DynamoCoreWpf/ViewModels/Watch3D/shaderSource/vsDynamoMesh.hlsl
```diff
@@ -58,9 +57,6 @@ PSInputCustom main(VSInput input)
     //set normal for interpolation	
     output.n = normalize(mul(input.n, (float3x3)mWorld));
 
-    //send our flags through to the frag shader
-    output.customParams.x = vParams.x;
```
aparajit-pratap: @mjkkirschner how do the flags get passed to the default `PSInput` struct? We currently have a custom struct for the vertex shader called `VSInputCustom`. Did we also get rid of that in the `helix-upgrade` branch?

aparajit-pratap: What do you mean by data interpolation? How did you figure out that that was going on?

src/VisualizationTests/HelixImageComparisonTests.cs
```diff
@@ -68,9 +70,10 @@ private static System.Drawing.Bitmap BitmapFromSource(BitmapSource bitmapsource)
             return bitmap;
         }
 
-        private string GenerateTestDataPathFromTest(string testname)
+        private string GenerateTestDataPathFromTest(string testname,bool debug = false)
```
aparajit-pratap: Could we rename this to `GenerateImagePathFromTest` unless we are generating any other "data" than just images?

src/VisualizationTests/HelixImageComparisonTests.cs
```diff
@@ -89,32 +92,72 @@ private void RenderCurrentViewAndCompare(string testName)
...
+            SaveBitMapSourceAsPNG(debugPath, bitmapsource);
+#endif
+
             compareImageColors(refbitmap, newImage);
         }
-
-        //TODO consider something like:
-        //(diferentPixelsCount / (mainImage.width* mainImage.height))*100
-        //for a percent image diff.
         private static void compareImageColors(Bitmap image1,Bitmap image2)
```
aparajit-pratap: I think it would be clearer if we can rename the parameters to `expectedImage`, `actualImage`, so it helps us identify which is which.

src/VisualizationTests/HelixImageComparisonTests.cs
```diff
@@ -89,32 +92,72 @@ private void RenderCurrentViewAndCompare(string testName)
             var refbitmap = BitmapFromSource(refimage);
             var newImage = BitmapFromSource(bitmapsource);
 
+#if SAVEDEBUGIMAGES
```
aparajit-pratap: What is this for?

src/VisualizationTests/HelixImageComparisonTests.cs
```diff
@@ -89,32 +92,72 @@ private void RenderCurrentViewAndCompare(string testName)
...
             }
+            var diff = CalculatePercentDiff(image1, differences);
+            Console.WriteLine($"% difference by out of range pixels was {(diff * 100).ToString("N" + 3)}%");
+            Assert.LessOrEqual(diff, .01);
+
+        }
+
+        /// <summary>
+        /// Euclidean distance between colors.
+        /// Does not use the alpha channel.
```
aparajit-pratap: Why is the alpha channel not used?

src/DynamoCoreWpf/ViewModels/Watch3D/shaderSource/vsDynamoMesh.hlsl
mjkkirschner: we no longer need any custom input structs - the data gets passed to each shader using a directX constant buffer, read about these here:
https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-constants

you can think of this like a uniform in opengl except per model, not shader or triangle.



So imagine we have a triangle with 3 verts - each vert has a color associated with it - to smoothly interpolate those colors for each pixel of the triangle, the renderer interpolates those values and sends those interpolated values to the pixel shader - that was being done to the flags data... it turns out that in directX apparently interpolating between the same value on each vertex does not interpolate correctly... the value should remain constant - but it obviously did not... 

in parallels(metal) it did interpolate correctly, so each pixel had the correct flags set.

What we were seeing was each pixel having a random set of flags - some frozen, some selected etc.

I figured out what was going on by noting that the data was being corrupted between the vertex stage and the pixel stage, but was fine if I set it myself using some constant number.

src/VisualizationTests/HelixImageComparisonTests.cs
mjkkirschner: yes can do.

src/VisualizationTests/HelixImageComparisonTests.cs
mjkkirschner: cool, will do.

src/VisualizationTests/HelixImageComparisonTests.cs
mjkkirschner: this will dump the new renders into the same folder where the reference images are so you can easily compare them and see what the current render looks like.

src/VisualizationTests/HelixImageComparisonTests.cs
mjkkirschner: theres no alpha channel in a full color image with no transparency. These are renders not textures or masks.

aparajit-pratap: What about when we are comparing images in isolation mode or frozen geometry?

mjkkirschner: nope, even then, the final output color is just red,green,blue, the alpha channel is still 0. These are 24bit images anyway, there isn't even an alpha channel.

src/DynamoCoreWpf/ViewModels/Watch3D/shaderSource/vsDynamoMesh.hlsl
aparajit-pratap: Wow, I would have expected interpolation only to be applied to vertex normals and colors, not for all data in the shader.

mjkkirschner: it depends on the semantic you use - check this out:
https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-struct#interpolation-modifiers-introduced-in-shader-model-4

src/VisualizationTests/HelixImageComparisonTests.cs
aparajit-pratap: So would the image comparison between an isolated geometry and non-isolated geometry return true (same image)?

aparajit-pratap: Is this the reason why @QilongTang couldn't write tests to check if isolated geometry was being rendered as expected?

mjkkirschner: no!
I'm not doing a good job explaining this :)

When we set the alpha channel to control transparency that has no real meaning in terms of color - how alpha channel changes the output of the frag shader is a decision the renderer/shader makes when it determines the final color of the rendered output.

So when we generate a bitmap the alpha channel has already been incorporated into the final RGB color - you can think of this like premultiplying or something - the color values have been modified by the alpha channel in some way to create a color blended with the other colors that were behind those pixels.

The resulting alpha channel is meaningless for the final image here because we're not creating a mask or other texture where we consider it to have some meaning - we only care about the final color.

QilongTang: @aparajit-pratap I think so, alpha value did not indicate anything for the isolated geometry case in the old `master` branch.

mjkkirschner: thats unrelated - in your test you were not looking at the color values of an image, you were looking at the color values of individual verts which were not used for computing the final color.

QilongTang: @mjkkirschner OK, let me read your comments carefully again!

mjkkirschner: I know it's a bit confusing - but you can see in the shader:

```
//if isolated - alpha should always be low - 
//overriding other alpha values (except if selected)
if(isIsolated && !isSelected)
{
	I.a = .1f;
}
```

that if the isolate mode is active we set the alpha of the returned color (returned from the shader)

Because the shader is using AlphaBlending - the renderer uses this alpha to compute a final color.

I hope that makes sense. Please let me know if not, I can find some better written docs on the subject.

aparajit-pratap: > The resulting alpha channel is meaningless for the final image here because we're not creating a mask or other texture where we consider it to have some meaning - we only care about the final color.

Ah, so it's something like homogeneous coordinates :)

aparajit-pratap: > thats unrelated - in your test you were not looking at the color values of an image, you were looking at the color values of individual verts which were not used for computing the final color.

Yup, sorry for confusing the two; my bad.

aparajit-pratap: > So when we generate a bitmap the alpha channel has already been incorporated into the final RGB color - you can think of this like premultiplying or something - the color values have been modified by the alpha channel in some way to create a color blended with the other colors that were located at the same pixels.

Ah, so it's something like homogeneous coordinates :)

QilongTang: Make sense! @mjkkirschner

PR #10509, opened 2020-03-27
disable latching which causes issues with scroll wheel

mjkkirschner:
### Purpose

Scrolling with the scroll wheel inside the node library sometimes broke in DynamoSandbox - this fixes it by using a setting flag in CEF as recommended by CefSharp project. - It is already used in Revit so thats why we could not reproduce there - when upgrading to CEF 67 or up we can remove this.

https://github.com/cefsharp/CefSharp/issues/2408

### Declarations


mjkkirschner:
@aparajit-pratap my guess for the reason this could not be reproduced in 2.0.x is the change in CEF versions between those releases and 2.1

QilongTang:
@mjkkirschner Thanks for the fix! Yeah

aparajit-pratap:
Awesome, LGTM!

PR #10512, opened 2020-03-27
Binary Expressions deleted by live runner make their first input symbols null.

mjkkirschner:
# Purpose

When a binary expression is deleted by the live runner - the binary AST is walked and all binary expressions within that subtree are assigned to null...

the change was from:
https://github.com/DynamoDS/Dynamo/pull/7282/

The behavior before this PR was that top level binary expressions were assigned to null and no AST walking was done.

I'm guessing @ke-yu was trying to support multiple assignments like:
a = 3;
b = a;
all being set to null.

but the consequence was other binary expressions like
`a && b` end up resulting in this  `a = null`

I'm opening this PR for feedback because I'm not clear on what binary expressions should actually be set to null, only assignments? Right now I am only stopping `OR` and `AND` - since that effects this particular bug, but I don't think thats the whole story.
all good.
Screen Shot 2020-03-27 at 5 41 04 PM
still good!
Screen Shot 2020-03-27 at 5 41 07 PM

### Declarations


mjkkirschner:
@aparajit-pratap please take another look, went with your third suggestion and added 2 tests. I still need to run the self serve to see if it breaks anything.

mjkkirschner:
@aparajit-pratap unfortunately, a bunch of failures in self serve, looking into it.

aparajit-pratap:
Curious, what did you think about making the UI logic nodes support replication?

src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -982,7 +983,7 @@ private List<BinaryExpressionNode> BuildNullAssignments(List<AssociativeNode> as
                 }
 
                 IdentifierNode leftNode = bNode.LeftNode as IdentifierNode;
-                if (leftNode == null || leftNode.ArrayDimensions != null)
+                if (leftNode == null || leftNode.ArrayDimensions != null  || bNode.Optr != Operator.assign)
```
aparajit-pratap, line 986: Minor, but why not have this check up here: https://github.com/DynamoDS/Dynamo/pull/10512/files#diff-2639b9fb615a6392f1215b5d9ff3c0f2R980

mjkkirschner:
> Curious, what did you think about making the UI logic nodes support replication?

I thought that could get very difficult because the ZT versions only support two inputs and the UI ones support more than 2 - so I believe we'd end up doing binary expressions anyway - would could do both though? This fix and also the change to UI nodes?

src/Engine/ProtoScript/Runners/LiveRunner.cs
mjkkirschner: you're right, that seems like a more appropriate place, where the binary node is validated, I can move it.

test/Engine/ProtoTest/LiveRunnerTests/MicroFeatureTests.cs
```diff
@@ -852,6 +852,93 @@ public void GraphILTest_DeletedNode01()
 
         }
 
+        [Test]
+        public void GraphILTest_DeletedBinaryExpresionDoesNotEffectReferences()
```
aparajit-pratap, line 856: Effectively this test is simulating the same thing that would happen in the Dynamo UI when running the graph, deleting the node, and rerunning it or is it testing something additionally?

aparajit-pratap:
> I thought that could get very difficult because the ZT versions only support two inputs and the UI ones support more than 2 - so I believe we'd end up doing binary expressions anyway - we could do both though? This fix and also the change to UI nodes?

Probably that's something to think about for later, add to the backlog maybe? For now, I think this fix looks good.

test/Engine/ProtoTest/LiveRunnerTests/MicroFeatureTests.cs
mjkkirschner: simulating the same thing as in the UI - I verified it failed before this change.

mjkkirschner:
@aparajit-pratap the self serve and parallel tests all pass, gonna merge this in.

PR #10544, opened 2020-04-06
Efficiently solve dependency metadata on install

mmisol:
### Purpose

When resolving dependencies from the workspace dependency extension,
the entire metadata of dependent packages was requested. However, only
information about the version was truly needed, and for packages with a
lot of versions this was causing latency issues.

Now only the package version of dependencies is obtained. As a result,
several code changes were needed, as part of the code path is shared
with the standard installation of packages through search, but with
different data requirements in each case. Several properties and
functions no longer used were marked as 'Obsolete'.

Added some basic unit tests for the new functions in package manager
client.

### Declarations

### Reviewers

@mjkkirschner @zeusongit 

### FYIs

@reddyashish


src/DynamoCoreWpf/Properties/Resources.en-US.resx
```diff
@@ -2233,6 +2229,10 @@ Uninstall the following packages: {0}?</value>
   <data name="DynamoViewSettingsMenuShowCodeBlockNodeLineNumber" xml:space="preserve">
     <value>Show CodeBlockNode Line Numbers</value>
   </data>
+  <data name="MessageFailedToDownloadPackageVersion" xml:space="preserve">
+    <value>Failed to download version {0} of package with id: {1}.  Please try again and report the package if you continue to have problems.</value>
+    <comment>Message box content. {0} = 1.2.3, {1} = 57d576e8f615e7725800001d</comment>
```
mjkkirschner, line 2234: are these default values?

mmisol: No, just examples of what kind of text is expected in the placeholders. I thought that's a good way of communicating that to whoever does the text translations.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
```diff
@@ -701,7 +704,7 @@ internal void DownloadAndInstall(PackageDownloadHandle packageDownloadHandle, st
                         Package dynPkg;
 
                         var pmExtension = DynamoViewModel.Model.GetPackageManagerExtension();
-                        var firstOrDefault = pmExtension.PackageLoader.LocalPackages.FirstOrDefault(pkg => pkg.Name == packageDownloadHandle.Name);
+                        var firstOrDefault = pmExtension.PackageLoader.LocalPackages.FirstOrDefault(pkg => pkg.ID == packageDownloadHandle.Id);
```
mjkkirschner, line 707: hmm, do you think this could change any existing behavior?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
```diff
@@ -721,18 +724,13 @@ internal void DownloadAndInstall(PackageDownloadHandle packageDownloadHandle, st
 
                         if (packageDownloadHandle.Extract(DynamoViewModel.Model, downloadPath, out dynPkg))
                         {
-                            var p = Package.FromDirectory(dynPkg.RootDirectory, DynamoViewModel.Model.Logger);
```
mjkkirschner, line 724: why is this no longer required?

mjkkirschner:
@mmisol looks good to me!

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
```diff
@@ -501,39 +498,39 @@ internal void ExecutePackageDownload(string name, PackageVersion version, string
...
-                var containsBinaries =
-                    allPackageVersions.Any(
-                        x => x.Item2.contents.Contains(PackageManagerClient.PackageContainsBinariesConstant) || x.Item2.contains_binaries);
-
-                var containsPythonScripts =
-                    allPackageVersions.Any(
-                        x => x.Item2.contents.Contains(PackageManagerClient.PackageContainsPythonScriptsConstant));
+                var containsBinariesOrPythonScripts = dependencyVersionHeaders.Any(x => 
+                    x.contents.Contains(PackageManagerClient.PackageContainsBinariesConstant) ||
+                    x.contains_binaries ||
```
gregmarr, line 529: We would get some minor performance gains by moving the `contains_binaries` test before the `Contains()` check.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
mmisol: I don't think so. I tested installing packages from both the search and the workspace dependency extension to be sure.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
mmisol: This was already being done inside `packageDownloadHandle.Extract`, but only for the workspace dependency extension code path. I unified both code paths and now it's always done inside `Extract`, which returns a boolean to report success.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
QilongTang: Agree

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
```diff
@@ -651,16 +642,28 @@ internal void ExecutePackageDownload(string name, PackageVersion version, string
...
+                    })
+                    .ToList()
+                    .ForEach(x => DownloadAndInstall(x, downloadPath));
             }
         }
 
         /// <summary>
         ///     Returns a newline delimited string representing the package name and version of the argument
         /// </summary>
+        [Obsolete("No longer used. Remove in 3.0.")]
```
QilongTang, line 666: @mjkkirschner I thought we made it clear somewhere that these APIs are not part of Dynamo SDK? If not, let's do that..

src/DynamoPackages/PackageManagerClient.cs
```diff
@@ -122,6 +123,7 @@ internal PackageHeader GetPackageMaintainers(IPackageInfo packageInfo)
         /// <param name="header"></param>
         /// <param name="version"></param>
         /// <returns></returns>
+        [Obsolete("No longer used. Remove in 3.0.")]
```
QilongTang, line 126: @mmisol I believe Internal APIs can be removed anytime

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
mjkkirschner: @QilongTang we only made that statement in new extension binaries like WorkspaceReferencesExtension - this is a public method in `DynamoCoreWPF` - we can't remove this method now without some of the other work on reorganizing or componentizing the API we've been discussing.

QilongTang: Thank you for clarifying. I thought this is in WorkspaceReference project. My bad

src/DynamoPackages/PackageDownloadHandle.cs
```diff
@@ -29,16 +29,21 @@ public State DownloadState
             }
         }
 
+        [Obsolete("No longer used. Remove in 3.0")]
         public Greg.Responses.PackageHeader Header { get; private set; }
         private string _name;
         public string Name { get { return Header != null ? Header.name : _name; } set { _name = value; } }
 
+        private string _id;
+        public string Id { get { return Header != null ? Header._id : _id; } set { _id = value; } }
```
mjkkirschner, line 38: for public properties please add /// summaries - also, should this really be public? Do we intend for others extensions/integrators to use this?

mmisol: This is used in `DynamoCoreWpf` so I'll add comments for it. I don't think someone might want to extend it in some way. This is used as a view model for the installed dependencies when installing a package from the search.

mjkkirschner: if it's only to be used for us - maybe we can make it internal and use the internalsVisibleTo attribute, my feeling is that we should be much more deliberate about what constitutes our public api. What do you think?

mmisol: My opinion is that using `InternalsVisibleTo` is more of a hack than a good practice. Making something public shouldn't make it "written in stone". I think we should identify exactly what we want our extension points to be instead, and only make promises for those.

PR #10550, opened 2020-04-07
Avoid duplication in package assembly list

mmisol:
### Purpose

Fixes the issue described here https://forum.dynamobim.com/t/package-contents-listed-twice-in-the-package-manager/48994. Tracked as DYN-2602.

### Declarations

### Reviewers

@aparajit-pratap @QilongTang

### FYIs

@mjkkirschner

### DIFF
--- a/src/DynamoPackages/PackageLoader.cs
+++ b/src/DynamoPackages/PackageLoader.cs
@@ -294,7 +294,7 @@ namespace Dynamo.PackageManager
         /// <param name="packages"></param>
         public void LoadPackages(IEnumerable<Package> packages)
         {
-            var enumerable = packages.ToList();
+            var packageList = packages.ToList();
 
             // This fix is in reference to the crash reported in task: https://jira.autodesk.com/browse/DYN-2101
             // TODO: https://jira.autodesk.com/browse/DYN-2120. we will be re-evaluating this workflow, to find the best clean solution.
@@ -308,7 +308,7 @@ namespace Dynamo.PackageManager
             // Disabling the run here since new packages are being loaded. 
             EngineController.DisableRun = true;
 
-            foreach (var pkg in enumerable)
+            foreach (var pkg in packageList)
             {
                 // If the pkg is null, then don't load that package into the Library.
                 if (pkg != null)
@@ -320,9 +320,11 @@ namespace Dynamo.PackageManager
             // Setting back the DisableRun property back to false, as the package loading is completed.
             EngineController.DisableRun = false;
 
-            var assemblies =
-                enumerable.SelectMany(x => x.EnumerateAssembliesInBinDirectory().Where(y => y.IsNodeLibrary));
-            OnPackagesLoaded(assemblies.Select(x => x.Assembly));
+            var assemblies = packageList
+                .SelectMany(p => p.LoadedAssemblies.Where(y => y.IsNodeLibrary))
+                .Select(a => a.Assembly)
+                .ToList();
+            OnPackagesLoaded(assemblies);
         }
 
         public void LoadCustomNodesAndPackages(LoadPackageParams loadPackageParams, CustomNodeManager customNodeManager)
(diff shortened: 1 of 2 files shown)


src/DynamoPackages/PackageLoader.cs
```diff
@@ -320,9 +320,11 @@ public void LoadPackages(IEnumerable<Package> packages)
             // Setting back the DisableRun property back to false, as the package loading is completed.
             EngineController.DisableRun = false;
 
-            var assemblies =
-                enumerable.SelectMany(x => x.EnumerateAssembliesInBinDirectory().Where(y => y.IsNodeLibrary));
-            OnPackagesLoaded(assemblies.Select(x => x.Assembly));
+            var assemblies = packageList
+                .SelectMany(p => p.LoadedAssemblies.Where(y => y.IsNodeLibrary))
+                .Select(a => a.Assembly)
+                .ToList();
```
mmisol, line 326: There are basically two things that were fixed:
1. The call to `EnumerateAssembliesInBinDirectory` was replaced by using `LoadedAssemblies`. This avoids re-adding assemblies to `LoadedAssemblies`.
2. The assemblies list is materialized. Before these changes, the call to `EnumerateAssembliesInBinDirectory` was made each time the enumerable was enumerated, which happened once per event handler for `PackageLoaded`. That caused triplication of the libraries for me on latest master.

aparajit-pratap:
@mmisol could youJIRA issue for this for tracking purposes.

src/DynamoPackages/PackageLoader.cs
aparajit-pratap: @mmisol you mentioned that `EnumerateAssembliesInBinDirectory` was being called in a couple of other places. Do those need to be replaced with `LoadedAssmeblies` as well?

aparajit-pratap:
@mmisol thanks for fixing this issue so quickly! Are there any tests you can add for this?

mjkkirschner:
should the use of this enumerateAssemblies method also be changed here:
https://github.com/DynamoDS/Dynamo/pull/10550/files#diff-e5fc16a8c62e48d288f3dd8cb3140d9cR261

src/DynamoPackages/PackageLoader.cs
mmisol: Yes, you are right. What I was seeing in the stack trace was actually this line of code being executed when the enumeration was materialized. This happened in two event handlers of `OnPackagesLoaded`, one in DynamoModel.cs(978) and another in EngineController.cs(521).

mmisol:
@mjkkirschner That code seems to have the same problem. I'll check if it's being used somewhere.

src/DynamoPackages/PackageLoader.cs
QilongTang: @mmisol So I assume whichever event handler called second time, with this fix, we will no longer re-adding assemblies to `LoadedAssemblies`. Do we want to create a test by mocking up the situation and check the end `LoadedAssemblies`?

mmisol: Yes, that was part of the issue. The call to `TryLoadPackageIntoLibrary` is the first to call `EnumerateAssembliesInBinDirectory` and after this `LoadedAssemblies` should be used to read them. 

I'll see to add a test. I have to think about what needs to be mocked, hopefully it's not too difficult.

mmisol:
Added a test. This is ready for review @aparajit-pratap @QilongTang 

@mjkkirschner About the method you mentioned, it's obsoleted and there are no usages, so I left it alone. Let me know if you would like to change ti anyway, but testing it would be difficult.

QilongTang:
@mmisol Thank you for adding the test. LGTM. Not sure if we need to include the other spot also part of this PR?

mmisol:
Re-thinking about what I wrote yesterday, unit testing the obsolete method should be easy. If you guys think it's safer to change it and unit test it rather than leaving as-is let me know and I can change it.

aparajit-pratap:
> Re-thinking about what I wrote yesterday, unit testing the obsolete method should be easy. If you guys think it's safer to change it and unit test it rather than leaving as-is let me know and I can change it.

I think this is good. LGTM.

mmisol:
Is this PR good to be merged @QilongTang @aparajit-pratap @mjkkirschner ? If so, can I get an approval? Thank you

QilongTang:
@mmisol Ship it!

PR #10557, opened 2020-04-09
Make sure extension window closes with Dynamo

mmisol:
### Purpose

View extensions that add a Window to the extension bar, run the risk of
leaving a window open when Dynamo closes, resulting in a zombie process
. The extension could avoid the problem in either of these two ways:

1. Setting the Owner property of the Window they create to the one
obtained from ViewLoadedParam.DynamoWindow.

2. Calling the Close method on the Window they created when the
Shutdown life-time method is called on the extension.

However, Dynamo should not rely on view extensions doing this. In order
to be able to still close no matter what, we assign the Owner property
explicitly in DynamoView.AddExtensionTabItem.

### Declarations

### Reviewers

@QilongTang 

### FYIs

@mjkkirschner

### DIFF
--- a/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
+++ b/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
@@ -230,6 +230,12 @@ namespace Dynamo.Controls
                 else
                 {
                     tab.Content = contentControl.Content;
+                    var extensionWindow = contentControl as Window;
+                    if (extensionWindow != null)
+                    {
+                        // Make sure the extension window closes with Dynamo
+                        extensionWindow.Owner = this;
+                    }
                 }
 
                 //Insert the tab at the end
--- a/test/DynamoCoreWpfTests/ViewExtensions/ViewExtensionTests.cs
+++ b/test/DynamoCoreWpfTests/ViewExtensions/ViewExtensionTests.cs
@@ -113,7 +113,6 @@ namespace DynamoCoreWpfTests
             };
             if (SetOwner)
             {
-                // Set the owner of the window to tuIDhe Dynamo window.
                 window.Owner = p.DynamoWindow;
             }
 
@@ -162,7 +161,6 @@ namespace DynamoCoreWpfTests
 
             var window = new Window
             {
-                // Set the owner of the window to tuIDhe Dynamo window.
                 Owner = p.DynamoWindow
             };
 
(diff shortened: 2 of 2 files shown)


test/DynamoCoreWpfTests/ViewExtensions/ViewExtensionTests.cs
```diff
@@ -87,11 +106,16 @@ public void Loaded(ViewLoadedParams p)
...
+            var window = new Window();
+            window.Closed += (sender, args) =>
             {
-                // Set the owner of the window to tuIDhe Dynamo window.
-                Owner = p.DynamoWindow
+                WindowClosed = true;
             };
+            if (SetOwner)
+            {
+                // Set the owner of the window to tuIDhe Dynamo window.
```
QilongTang, line 116: Would you fix the word `tuIDhe`, I think you mean `the`?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -230,6 +230,12 @@ internal TabItem AddExtensionTabItem(IViewExtension viewExtension, ContentContro
                 else
                 {
                     tab.Content = contentControl.Content;
+                    var extensionWindow = contentControl as Window;
+                    if (extensionWindow != null)
+                    {
+                        // Make sure the extension window closes with Dynamo
+                        extensionWindow.Owner = this;
```
QilongTang, line 237: Interesting.. I start to think this maybe why the fix did not work for me because when I tried it, I was calling close() specifically for Tuneup while your code should work for both TuneUp and NodePropertyPalette.

QilongTang: However, giving this more thoughts. I think this approach is excellent that Dynamo should handle it properly. We may still want to release a version of Tuneup which contain a close() call so that current users who are using Tuneup can be saved.

mmisol: Yes, that's why I opened https://github.com/DynamoDS/TuneUp/pull/33. We should publish a new version after we merge that one.

test/DynamoCoreWpfTests/ViewExtensions/ViewExtensionTests.cs
mmisol: That's not mine, so I'll probably remove it :)

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
QilongTang: @mmisol Gotcha

QilongTang:
LGTM. Let's fix the comments and merge

PR #10558, opened 2020-04-09
[DYN-1682]: Improve node warning, autocomplete and Node2Code behavior with class name conflicts

aparajit-pratap:
### Purpose
This attempts to fix an inconsistent behavior with multiple definition warnings being shown when making certain function calls in CBN's vs. executing the nodes without errors. 

In the example below, there are 2 `List` classes, one, the built-in type and the other from a package (Orchid). Autocomplete displays both the `List` classes, the second one prefixed with a namespace (Orchid) to distinguish it uniquely from the first. Autocomplete indicates that the first `List` type can be used as-is. Even then the node would show multiple definition warnings indicating it cannot resolve which `List` class to use. However, at runtime, this does not pose an issue and the node executes without any errors. This clearly inconsistent behavior is fixed here.

The new behavior shows no multiple definition warnings when the CBN is initially created and is consistent both with the autocomplete prompts and the successful execution.
_____________________________________________________________________________________________________
In addition to the above, few other existing bugs have been fixed in CBN autocomplete and in Node to Code. These are better explained with an example of **List**, **DSCore.List**, and **Orchid.Core.List**, where `DSCore.List` is hidden from the library and `List` is a class in the global namespace that derives from `DSCore.List`. `Orchid.Core.List` is an external class from a package.

**Some notes to keep in mind:**
1. These include fixes to a `ShortestQualifiedNameReplacer` class that basically rewrites a collection of classes, all having the same name, but with different namespaces. This component ensures that the full names of these classes are automatically shortened (to make CBN code less verbose) so that they can still be resolved uniquely. For example, in this case, `Orchid.Core.List`, can be renamed to `Orchid.List`, by dropping the `Core`, from its full namespace as DesignScript _(in its implementation of Dynamo only)_ can still recognize it uniquely, separate from the other two.
2. CBN autocomplete does not work on namespaces, only on classes as the concept of namespaces is not expected to be known by end-users. However, we cannot completely do away with showing namespaces in CBNs in cases where we need to differentiate between classes with the same classname coming from different libraries.
3. Hidden classes (with `IsVisibleInLibrary(false)` attribute) do not show in the library and for consistency, we should not make them appear in code in CBNs either. However, a hidden class can have explicitly visible methods and properties (if overridden with `IsVisibleInLibrary(true)` attribute), which will show up as nodes in the library. **Only** if such a class is inherited by a subclass, will its methods be displayed in autocomplete in CBN under the derived class. An example of such a class is `DSCore.List`, where its methods are visible as nodes in the library and they appear under `List` in CBN autocomplete. Hidden classes if used in CBN (even if not discovered by autocomplete) will still autocomplete to show its members.
4. A component called `ElementResolver` maintains a map of short names vs. fully qualified names, based on their use in CBNs. The purpose of this map is such that if a new library with a conflicting classname is imported into an existing graph, CBN code continues to work and does not break even in the presence of a new conflict that is introduced at a later time. These conflicts if left to be resolved by the compiler _could_ give new results causing existing graphs to break. In such cases, we lean on the ER to give us the exact classname that was resolved earlier.

**1. Autocomplete behavior and fixes**
- When beginning to type `L`, `i`, `s`..., autocomplete begins to filter all the classes with the name `List` and only shows `List` and `Orchid.List` as it hides the hidden `DSCore.List` class
- When autocompleting on `List`, all methods and properties on `List` and its base class (`DSCore.List`) are displayed
- When typing `DSCore.List` - (1) no autocomplete on `DSCore` namespace is expected and (2) code completion on `DSCore.List`, will display method list only on that base class (not on the derived `List` class)
- Code completion of hidden class (if typed without autocomplete assist), still displays members of hidden class

**2. Node warning fixes**
- This section essentially refers to the original bug (DYN-1682), that we planned to fix (before this task exploded to include all these other related improvements). This bug refers to the fact that a node from either `List`, `DSCore.List` or `Orchid.Core.List` must not throw a `multiple definition warning` at compile time as all of them are expected to resolve uniquely (both as nodes as well as in CBN code).

Before:
multi_defn_warning_demo

After:
multi_defn_demo

**3. Node To Code behavior and fixes**
- N2C on `List` nodes should generate `List._MethodName_` and `List._PropertyName(obj)_` in CBN
- N2C on `DSCore.List` nodes should also generate the `List._MethodName_` as above (since `DSCore.List` is hidden **and** is a base class of `List`)
- N2C on `Orchid.Core.List` node should generate CBN with the shortest unique name when compared with all other classes with the name `List`, which in this case is `Orchid.List._MethodName_`.
- What if we perform N2C on a class that is hidden, and its methods (aka nodes) are visible but it is not inherited by any other visible class? In such a **rare** case, N2C generates the fully qualified namespace (as the user might expect to see _something_ in code for a node that exists, even though its class is hidden)

Before:
n2c_before_demo

After:
n2c_after_demo

**4. Prefixing classes with full namespaces without assistance from autocomplete**
- In cases where advanced users may be aware of full namespaces of classes they wish to use in CBN, there will be no autocomplete assistance on those namespaces, only on the final classname that they type. The full namespaces will, however, be added to the `ElementResolver`. For e.g., if a user types the full name: `Orchid.Core.List`, even though `Orchid.List` would suffice, `Orchid.Core.List` would start appearing in autocomplete and in N2C code.  
- If a hidden class is typed there will be no indication of its existence in autocomplete (as it's meant to be hidden), the CBN will still execute (as it's a valid class imported into the VM), however, it will not be added to the `ElementResolver` as if it is, the hidden class will start to unexpectedly appear in CBN autocomplete and N2C behavior, which will be confusing to the end-user. For example, if `DSCore.List` is typed into a CBN, it will not be added to the `ElementResolver` so that it does not appear in any CBN autocompletion lists or N2C code generated subsequently.

_Note that there is a proposal to display the contents of the ER map in the UI, so that it is not a Blackbox and so that users can understand why in their graph certain namespaces in CBNs are working and why others are not._

### Declarations

### Reviewers

@DynamoDS/dynamo 

### FYIs

@Amoursol


test/DynamoCoreTests/NodeToCodeTest.cs
```diff
@@ -976,8 +976,7 @@ public void NonUniqueNamespaceConflict_on_node2code_doesNotCrash()
             var cbn = nodes.OfType<CodeBlockNodeModel>().FirstOrDefault();
             Assert.IsNotNull(cbn);
 
-            var error = "Multiple definitions for 'FFITarget.B.DupTargetTest' are found as FFITarget.C.B.DupTargetTest, FFITarget.B.DupTargetTest";
```
mjkkirschner, line 979: @aparajit-pratap is there still a test that asserts this error is raised in the correct situation where there actually are conflicts?

mjkkirschner:
@aparajit-pratap it looks good except for one question regarding tests.

QilongTang:
LGTM

aparajit-pratap:
There is a regression I found in node to code (for which there was no test) so I'm continuing on this to fix that.

src/DynamoCore/Engine/CodeCompletion/CodeCompletionServices.cs
```diff
@@ -191,7 +191,7 @@ private void AddTypesToCompletionData(string stringToComplete, List<CompletionDa
             var partialName = stringToComplete.ToLower();
             // Add matching Classes
             var classMirrorGroups = StaticMirror.GetAllTypes(core).
-                Where(x => !x.IsHiddenInLibrary && x.Alias.ToLower().Contains(partialName)).
+                Where(x => x.Alias.ToLower().Contains(partialName)).
```
aparajit-pratap, line 194: We should compute shortest names in `Symbol.GetShortestUniqueNames` while including hidden classes and later filter out the hidden class names so that they are not displayed to the user.

mjkkirschner: I'm pretty confused by this change. Can you tell me how it affects the following use cases.

1. Previously I wrote some code like `__Apply` or some other hidden class, like `ElementResolver.GetElement` in the revit elements namespace. What happens to my code block?

2. If I try to access those hidden types now, what happens?

aparajit-pratap: You can still access those hidden types from code block nodes. They just don't show up in the library or autocomplete. My fixes here are to ensure they also don't show up on doing node to code operations.

test/DynamoCoreTests/NodeToCodeTest.cs
aparajit-pratap: Modified an existing test to catch this scenario.

src/DynamoCore/Engine/ShortestQualifiedNameReplacer.cs
```diff
@@ -108,84 +142,83 @@ private bool TryShortenClassName(IdentifierListNode node, out AssociativeNode sh
...
 
-        private IdentifierListNode RewriteNodeWithShortName(IdentifierListNode node)
+        private AssociativeNode CreateNodeFromShortName(string className, string qualifiedName)
         {
-            // Get class name from AST
-            string qualifiedName = CoreUtils.GetIdentifierExceptMethodName(node);
+            // Get the list of conflicting namespaces that contain the same class name
+            var matchingClasses = classTable.ClassNodes.Where(x =>
+            {
+                //var isHiddenInLibrary = x.ClassAttributes?.HiddenInLibrary ?? false;
```
QilongTang, line 159: hmm are commented code still needed?

src/DynamoCore/Engine/ShortestQualifiedNameReplacer.cs
```diff
@@ -108,84 +142,83 @@ private bool TryShortenClassName(IdentifierListNode node, out AssociativeNode sh
...
+            // If shortName for fully qualified classname is not found, it could be a base class
+            // present in class hierarchy of any of the other matchingClasses, in which case
+            // set shortName to the one for the derived class.
+            var qualifiedClassNode = matchingClasses.FirstOrDefault(x => x.Name == qualifiedName);
+            var classHierarchies = matchingClasses.Where(x => x != qualifiedClassNode).
+                Select(y => classTable.GetClassHierarchy(y));
+            foreach (var hierarchy in classHierarchies)
+            {
+                // If A derives from B, which derives from C, the hierarchy for A 
+                // is listed in that order: [A, B, C], so we start searching in reverse order.
```
QilongTang, line 211: 👍

src/Engine/ProtoCore/ClassTable.cs
```diff
@@ -393,7 +393,7 @@ public class ClassTable
         //Symbol table to manage symbols with namespace
         private Namespace.SymbolTable symbolTable = new Namespace.SymbolTable();
         
-        private List<ClassNode> GetClassHierarchy(ClassNode node)
+        public List<ClassNode> GetClassHierarchy(ClassNode node)
```
QilongTang, line 396: Does `internal` work here?

src/DynamoCore/Engine/ShortestQualifiedNameReplacer.cs
aparajit-pratap: Will remove. Thanks for pointing out!

src/Engine/ProtoCore/ClassTable.cs
aparajit-pratap: No, since this method is defined in `ProtoCore` but now being used in `DynamoCore`.

mjkkirschner: I understand this is being used by dynamo core - we also can use the internalsVisibleTo attribute to expose internal members to other assemblies which we own.

We should develop this principal further, but a few points.
* ProtoCore.dll is currently shipped as a part of the DynamoCore nuget package. To me looking at this package, as an external developer I see this as part of the API.
* A public method in this assembly which is shipped as part of a nuget package, will end up needing to be maintained.

Is this method something that should be used externally by integrators? Is there a compelling reason to make it part of the API? An interesting use case? A workflow that cannot be implemented without it? 

If not, then what is the benefit?

If yes, then great, let's add documentation to it.

aparajit-pratap: Have we made a decision on going forward with reducing the API surface of DynamoCore, DynamoCoreWpf etc. by continuing to expose them as our public API or are we deciding on keeping them as-is and making them internal and having a separate API layer for public consumption (based on your RFC)? 

Whatever is the strategy, I think then we should come up with certain coding guidelines before we continue to make PR's. Going by the RFC again, it sounds like none of ProtoCore should be part of the API and we should surface just the AST's publicly. I'll make this method internal as well for now.

mjkkirschner: No, we have made no decision -very much welcome those discussion - there are many options.
Thanks for addressing this, I don't think we can stop all work before we entirely decide this - but what  I am suggesting is that we proceed conservatively until we do so. 

I think it's likely we will find that some users are *already* using ProtoCore, and other binaries we didn't consider fully as APIs.

For example, the Civil3d integration has very specific interactions with callsites and trace, I am not 100% what APIs that team uses to control and redirect trace but I guess it is in protocore.

src/DynamoCore/Engine/ShortestQualifiedNameReplacer.cs
```diff
@@ -90,7 +91,40 @@ public override AssociativeNode VisitIdentifierListNode(IdentifierListNode node)
                 RightNode = rightNode,
                 Optr = Operator.dot
             };
-            return leftNode != newLeftNode ? node : RewriteNodeWithShortName(node);
+            return node;
+        }
+
+        public override AssociativeNode VisitIdentifierNode(IdentifierNode node)
```
mjkkirschner, line 97: I know many public methods in this visiter are missing summaries - /// - does it make sense to add them? Or atleast add them to new visitors?

src/DynamoCore/Engine/ShortestQualifiedNameReplacer.cs
```diff
@@ -108,84 +142,79 @@ private bool TryShortenClassName(IdentifierListNode node, out AssociativeNode sh
...
-                    var symbolList =
-                        matchingClasses.Select(matchingClass => new ProtoCore.Namespace.Symbol(matchingClass))
-                            .ToList();
-                    var shortNames = ProtoCore.Namespace.Symbol.GetShortestUniqueNames(symbolList);
-
-                    // Get the shortest name corresponding to the fully qualified name
-                    shortName = shortNames[new ProtoCore.Namespace.Symbol(qualifiedName)];
+                    if (hierarchy[i] == qualifiedClassNode)
+                    {
+                        shortName = shortNames[new Symbol(hierarchy[0].Name)];
```
mjkkirschner, line 212: just double checking - this should always be `0`?

aparajit-pratap: Yes, we want the most derived class in that list.

src/DynamoCore/Engine/ShortestQualifiedNameReplacer.cs
aparajit-pratap: I think it is implicit in the name of the method/design pattern. If any, I think the class should have some docs if it doesn't already.

mjkkirschner: okay, we don' t need to address this now, but we should investigate silencing the xml doc warnings these methods throw during build then.

PR #10561, opened 2020-04-11
[DYN-2609] Fix crash when installing empty package

aparajit-pratap:
### Purpose

Fix for regression caused by #10544  - crash when installing an empty package like LunchBox. Changes to new package version API's on the PM (I think) return the `PackageVersion` whose contents can be `null` for an empty package. Prior to the API change, the contents would be an empty string for an empty package. Thus it was necessary to add a null check.

The UI automation test _DynamoTests.dll.DynamoTests.Tests.SmokeTestPackage_ was failing, which signaled this regression.

### Declarations

### Reviewers

@mmisol 
@mjkkirschner 

### FYIs

@alfredo-pozo

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
@@ -525,9 +525,15 @@ namespace Dynamo.ViewModels
 
                 // determine if any of the packages contain binaries or python scripts.  
                 var containsBinariesOrPythonScripts = dependencyVersionHeaders.Any(x =>
-                    x.contains_binaries ||
-                    x.contents.Contains(PackageManagerClient.PackageContainsBinariesConstant) ||
-                    x.contents.Contains(PackageManagerClient.PackageContainsPythonScriptsConstant));
+                {
+                    var are_contents_empty = string.IsNullOrEmpty(x.contents);
+                    var contains_binaries = x.contains_binaries ||
+                                            !are_contents_empty && x.contents.Contains(PackageManagerClient.PackageContainsBinariesConstant);
+                    var contains_python =
+                        !are_contents_empty && x.contents.Contains(PackageManagerClient.PackageContainsPythonScriptsConstant);
+                    return contains_binaries || contains_python;
+
+                });
 
                 // if any do, notify user and allow cancellation
                 if (containsBinariesOrPythonScripts)


src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
```diff
@@ -525,9 +525,15 @@ internal void ExecutePackageDownload(string name, PackageVersion version, string
...
-                    x.contains_binaries ||
-                    x.contents.Contains(PackageManagerClient.PackageContainsBinariesConstant) ||
-                    x.contents.Contains(PackageManagerClient.PackageContainsPythonScriptsConstant));
+                {
+                    var are_contents_empty = string.IsNullOrEmpty(x.contents);
+                    var contains_binaries = x.contains_binaries ||
+                                            !are_contents_empty && x.contents.Contains(PackageManagerClient.PackageContainsBinariesConstant);
+                    var contains_python =
+                        !are_contents_empty && x.contents.Contains(PackageManagerClient.PackageContainsPythonScriptsConstant);
+                    return contains_binaries || contains_python;
```
aparajit-pratap, line 534: `PackageManagerClient.PackageContainsPythonScriptsConstant` and `PackageManagerClient.PackageContainsBinariesConstant` are deprecated so I'm not sure if we should use these or something else.

aparajit-pratap:
@alfredo-pozo the `DynamoSelfServe` job is failing with this error:
```
Exception: hudson.AbortException: UT & IT stage: hudson.AbortException: Fail to exec CI test scripts. Please check build log for detail.
Internal error message: script returned exit code 1
```
Any clue?

alfredo-pozo:
The last execution is completed, the failures were because of Docker issues @aparajit-pratap

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
mjkkirschner: I'm not sure either, but the package header only contains one boolean field - contains binaries which is true if the package contains binaries or python...

QilongTang:
LGTM

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
gregmarr: There are 41 versions across 22 packages that are still using this "obsolete" method of marking packages.  We could theoretically update those version records in the database to have the `contains_binaries` field set to `true` and then we could remove these constants.

aparajit-pratap: Is there a `contains_python` field as well? If so, then we need to add it to the check here.

gregmarr: No, there is only a single field for both.

PR #10565, opened 2020-04-13
Python Engine Switcher for internal testing

QilongTang:
Please Note:

### Purpose

Team agreed that we are not going to introduce any setting in Debug Menu. Instead, simply use context menu options under TestMode and to hook up with the work @reddyashish and @Dewb  are working on.

image

With this setting, Python nodes can switch Engine anytime in debug mode.
The example below distinguish the case that print statement got changed in Python3 so that same statement in left (python3 node) will throw a warning instead of execute.
image

### Declarations

### Reviewers

@DynamoDS/dynamo


QilongTang:
Self serve is a pass, the failing test is a sporadic failure

src/Libraries/PythonNodeModelsWpf/PythonNode.cs
```diff
@@ -29,9 +31,32 @@ public void CustomizeView(PythonNode nodeModel, NodeView nodeView)
             dynamoViewModel = nodeView.ViewModel.DynamoViewModel;
             workspaceModel = nodeView.ViewModel.WorkspaceViewModel.Model;
 
-            var editWindowItem = new MenuItem { Header = PythonNodeModels.Properties.Resources.EditHeader, IsCheckable = false };
-            nodeView.MainContextMenu.Items.Add(editWindowItem);
-            editWindowItem.Click += delegate { EditScriptContent(); };
```
aparajit-pratap, line 34: These 3 lines were outside the `debug` flag previously. Does it not mean that they still need to be outside the `if (dynamoViewModel.IsDebugBuild)`?

src/Libraries/PythonNodeModelsWpf/PythonNode.cs
```diff
@@ -97,5 +122,25 @@ private void EditScriptContent()
                 }
             }
         }
+
+        /// <summary>
+        /// MenuItem click handler
+        /// </summary>
+        private void UpdateToPython2Engine()
+        {
+            pythonNodeModel.Engine = PythonNodeBase.DefaultPythonEngine;
```
mmisol, line 131: I don't think it's a good idea to rely on `PythonNodeBase.DefaultPythonEngine` pointing to version 2, as we should eventually make version 3 the default, and then this code would break. Is there some other constant/enum that we can use that's explicitly about Python 2.

src/Libraries/PythonNodeModelsWpf/PythonNode.cs
```diff
@@ -29,9 +31,32 @@ public void CustomizeView(PythonNode nodeModel, NodeView nodeView)
...
-            editWindowItem.Click += delegate { EditScriptContent(); };
+            // If it is a Debug build, display a python engine switcher
+            if (dynamoViewModel.IsDebugBuild)
+            {
+                var editWindowItem = new MenuItem { Header = PythonNodeModels.Properties.Resources.EditHeader, IsCheckable = false };
+                nodeView.MainContextMenu.Items.Add(editWindowItem);
+                editWindowItem.Click += delegate { EditScriptContent(); };
+                var pythonEngineVersionMenu = new MenuItem { Header = PythonNodeModels.Properties.Resources.PythonNodeContextMenuEngineSwitcher, IsCheckable = false };
+                nodeView.MainContextMenu.Items.Add(pythonEngineVersionMenu);
+                pythonEngine2Item.Click += delegate { UpdateToPython2Engine(); }; var pythonEngine3Item = new MenuItem { Header = PythonNodeModels.Properties.Resources.PythonNodeContextMenuEngineVersionThree, IsCheckable = true };
```
aparajit-pratap, line 42: Can we have a newline at the end of the `;`? 
How is `pythonEngine3Item` a `var`? Isn't it hiding the private field?

src/Libraries/PythonNodeModelsWpf/PythonNode.cs
QilongTang: Thanks, will move out

src/Libraries/PythonNodeModelsWpf/PythonNode.cs
QilongTang: Yup, should be removed, I guess missed this

src/Libraries/PythonNodeModelsWpf/PythonNode.cs
QilongTang: @mmisol If we change the value for `PythonNodeBase.DefaultPythonEngine`, we should update the code here. Let me add a TODO, this is the only new flag we added, there is no other indication. Also I think this alpha test code will be removed later.

src/Libraries/PythonNodeModelsWpf/PythonNode.cs
```diff
@@ -32,6 +34,29 @@ public void CustomizeView(PythonNode nodeModel, NodeView nodeView)
             var editWindowItem = new MenuItem { Header = PythonNodeModels.Properties.Resources.EditHeader, IsCheckable = false };
             nodeView.MainContextMenu.Items.Add(editWindowItem);
             editWindowItem.Click += delegate { EditScriptContent(); };
+            // If it is a Debug build, display a python engine switcher
+            if (dynamoViewModel.IsDebugBuild)
+            {
+                var pythonEngineVersionMenu = new MenuItem { Header = PythonNodeModels.Properties.Resources.PythonNodeContextMenuEngineSwitcher, IsCheckable = false };
```
mjkkirschner, line 40: while this is usually a good thing to do, not sure we need localization to spend time on a debug flag?

QilongTang: @mjkkirschner ha, I guess we have been doing it for 3rd party devs from a different country?

src/Libraries/PythonNodeModelsWpf/PythonNode.cs
Dewb: I agree with @mmisol, this should check for 2 directly.

QilongTang: @Dewb @mmisol Do you mean checking the actual python2 string value instead of this property? If that's the case, how about we rename it to not call it default? From the API standpoint, referencing the default still do not give me a clue which engine version it is

mmisol: I just checked out the branch to take a look at the code. Renaming the `DefaultPythonEngine` constant to `PythonNet2Engine` (or something with a 2 in it :) ) would be good to me.

QilongTang: @Dewb If you are also fine with it. I will make that change

mjkkirschner: but it's not pythonNet2 - it's ironpython2 isn't it? I could be confused...

mmisol: I'll settle for anything with a 2 in it 😂

mmisol: But to be honest, I have usually used enums, which are named exactly as the values, for things like these. Might be something to consider still.

mjkkirschner: enums are fine as well, we just cannot remove enums or reorder them without risking API issues.

mmisol: Order should not be a problem if they were serialized as strings, but I agree that you lose the flexibility of maybe having other python engines loaded dynamically.

Dewb: I agree enums would be superior here. What we really would like is a type-checked enum that can be dynamically populated by what assemblies you have loaded (so the enum only contains CPython3 if DSCPython.dll is loaded and DSIronPython.dll is not.) 

This sort of fancy, module-spanning dynamic enum has been built in C++ for other products. I'm not sure there's an easy technical solution in C#, but it seems possible. I think this is a good area to investigate further in the tasks to make the engines optional; it doesn't need to gate merging this feature. 

It might be a good idea to make the property non-public so we preserve the ability to change the API before the Python 3 features are finalized. The property will have to be serialized to a string in .dyn files regardless, so a move to enums won't have to break graphs.

Dewb: Ah, I see the source of the confusion now. The `DefaultPythonEngine` static property was not intended to be "the name of Python 2 engine" it was intended to mean "which of the possible values of this string-faked enumeration should be assigned to nodes that do not have a value of the Engine property at deserialization time." The idea was it would be set once here, and referenced wherever we needed to, and in the future when we make the default 3, we can change it in one place.

Ideally, the NodeModel code wouldn't know about the details of either of these engines. We could create a `RegisteredPythonEngines` data structure and each engine DLL could call a `RegisterPythonEngine` method during initialization, providing a serialization name and an entry point. 

We should also separate the concept of "the default value of this property on new nodes" from "the value we attach for graphs from 2.6 and earlier." We will want to change the default engine from 2 to 3 in the future, but pre-2.6 graphs should always be set to 2.

mjkkirschner: these are important point's Mike - thanks for clarifying them.

QilongTang: hmmm I think contractor's work especially the UI part and editor switch will need the whole series of the values somewhere. Do you have a preference between Enums with display name or set of strings? I think we need them available anyway.

QilongTang:
Hi @DynamoDS/dynamo and @Dewb I have added Enums with display value instead of static string from our discussion to remove the whole confusion around version checking. Currently it is as 
```
    /// <summary>
    /// Enum of possible values of python engine versions
    /// </summary>
    public enum PythonEngineVersions
    {
        [Display(Name = "IronPython2")]
        IronPython2 = 0,

        [Display(Name = "CPython3")]
        CPython3 = 1,
    }
```
Maybe for the enum values we can move to use the actual version number so they are still identical plus we reserve more information? What do you think. I am supportive of getting version info loading time as a future improvement as well. Notice after the change, the graph schema is still consistent with before.

Dewb:
I don't think we need to maintain both a string and an enum version of the property for compatibility. The enum alone should be sufficient.

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -25,12 +39,12 @@ protected PythonNodeBase(IEnumerable<PortModel> inPorts, IEnumerable<PortModel>
...
-        public static readonly string DefaultPythonEngine = "IronPython2";
-        public static readonly string PythonNet3Engine = "CPython3";
+        internal PythonEngineVersions DefaultPythonEngine = PythonEngineVersions.IronPython2;
 
-        private string engine = DefaultPythonEngine;
+        private PythonEngineVersions engine = PythonEngineVersions.IronPython2;
 
-        public string Engine
+        [JsonIgnore]
+        public PythonEngineVersions EngineEnum
```
mmisol, line 47: For enums that are not flags (i.e. can be combined using `|`) the name should be singular. In this case it would be `PythonEngineVersion`.

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -43,6 +57,21 @@ public string Engine
...
+        /// </summary>
+        public string Engine
+        {
+            get
+            {
+                return engine.GetType()?
+                    .GetMember(engine.ToString())?
+                    .First()?
+                    .GetCustomAttribute<DisplayAttribute>()?
+                    .Name;
```
mmisol, line 71: `ToString` should actually work and it is much simpler. We could get rid of the attribute.

QilongTang: @mmisol Thanks!

QilongTang:
Addressed some comments, please let me know. Due to the impact of this work to what @SHKnudsen  is doing, I would like to wrap up and sync with the UI work potentially building on top of it. Plus this should give @SHKnudsen an easier way to test the engine switch. I would like to address the good points raised by @Dewb here in the follow up Jira task.

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -25,24 +39,33 @@ protected PythonNodeBase(IEnumerable<PortModel> inPorts, IEnumerable<PortModel>
...
-                    RaisePropertyChanged("Engine");
+                    RaisePropertyChanged("EngineEnum");
                 }
             }
         }
 
+        /// <summary>
+        /// Return the display name of python engine enum.
+        /// </summary>
+        public string Engine
```
aparajit-pratap, line 61: Where is this intended to be used?

QilongTang: Serialized in DYN

mmisol:
@QilongTang How about having a single property called `Engine`, that is an enum, but serializes as a string? I think that should be the simplest approach and it should be possible if we use the `[JsonConverter(typeof(StringEnumConverter))]` attribute on top of the property.
https://stackoverflow.com/questions/2441290/javascriptserializer-json-serialization-of-enum-as-string

Also, is there a reason to keep the `Display` attributes now?

QilongTang:
@mmisol Thank you for pointing out. Addressed your comments and cleaned up changes

src/Libraries/PythonNodeModels/PythonNode.cs
QilongTang: Removed

src/Libraries/PythonNodeModelsWpf/PythonNode.cs
```diff
@@ -97,5 +122,26 @@ private void EditScriptContent()
                 }
             }
         }
+
+        /// <summary>
+        /// MenuItem click handler
+        /// </summary>
+        private void UpdateToPython2Engine()
+        {
+            // If PythonNodeBase.DefaultPythonEngine is updated to python 3, we should update this piece of code
```
mmisol, line 131: We no longer need this comment.

QilongTang:
@mmisol Removed. Merging

PR #10569, opened 2020-04-14
Handle memory outage on render (cherry pick)

mmisol:
### Purpose

This brings the fix from master to the helix-upgrade branch. The only
difference is that the test now mocks the AggregateRenderPackage
function directly, because the function mocked in master no longer
exists. The visibility was changed to protected virtual in order to be
able to mock it.

Related PR: #10535 

### Declarations

### Reviewers

@mjkkirschner

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
@@ -1647,7 +1647,7 @@ namespace Dynamo.Wpf.ViewModels.Watch3D
         /// attaches them to the visual scene.
         /// </summary>
         /// <param name="packages">An <see cref="IEnumerable"/> of <see cref="HelixRenderPackage"/>.</param>
-        protected virtual void AggregateRenderPackages(IEnumerable<HelixRenderPackage> packages)
+        internal virtual void AggregateRenderPackages(IEnumerable<HelixRenderPackage> packages)
         {
             IEnumerable<string> customNodeIdents = null;
             if (InCustomNode())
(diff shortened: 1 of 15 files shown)


QilongTang:
@mmisol can you remind me if `Moq` is a new dependency?

mmisol:
@QilongTang We were using `Moq` in other tests projects but not in this one, so it's new here but not in our test suite.

QilongTang:
@mmisol  Thanks! In general Cherry-pick PRs can be merged by your own. So feel free. The EngOps build usually fails because of wrong branch params sent to CI.

mmisol:
Thanks @QilongTang . Maybe I'll wait for @mjkkirschner to take a quick look just to make sure he is ok with the visibility change to `AggregateRenderPackage`, as that part differs from the changes done in `master`.

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -1635,7 +1647,7 @@ private void ToggleTreeViewItemHighlighting(string path, bool isSelected)
         /// attaches them to the visual scene.
         /// </summary>
         /// <param name="packages">An <see cref="IEnumerable"/> of <see cref="HelixRenderPackage"/>.</param>
-        private void AggregateRenderPackages(IEnumerable<HelixRenderPackage> packages)
+        protected virtual void AggregateRenderPackages(IEnumerable<HelixRenderPackage> packages)
```
mjkkirschner, line 1650: This has the effect of exposing this method to the API in any derived class, as this is currently a public method - (In the current API) Of course we'd like to do a refactor and change our API policies as we have been discussing but for now I'm a bit hesitant.

If this method was made internal could it be mocked?

If it has to become protected then we can use an interface as a the argument type and check the type inside the method.

mmisol: > If this method was made internal could it be mocked?

Apparently this could be done https://stackoverflow.com/a/2823554. Let me give that a try.

mmisol: @mjkkirschner That worked! Would you take one last look?

PR #10577, opened 2020-04-17
Place VM reset in Runtime Core mutex context

mmisol:
### Purpose

Enlarged the mutually exclusive context in ResetVMAndResyncGraph to
also include ReInitializeLiveRunner. This is important as the
reinitialization of the runner resets the runtime core, making the
Design Script executable null, which will cause any running executions
to crash Dynamo, even respecting the mutually exclusive context.

### Declarations

### Reviewers

@aparajit-pratap @mjkkirschner @reddyashish 

### FYIs

@QilongTang

### DIFF
--- a/src/Engine/ProtoScript/Runners/LiveRunner.cs
+++ b/src/Engine/ProtoScript/Runners/LiveRunner.cs
@@ -1532,27 +1532,30 @@ namespace ProtoScript.Runners
         /// <param name="syncData"></param>
         public void ResetVMAndResyncGraph(IEnumerable<string> libraries)
         {
-            // Reset VM
-            ReInitializeLiveRunner();
-
-            if (!libraries.Any())
+            lock (mutexObject)
             {
-                return;
+                // Reset VM. This needs to be in mutex context as it turns DSExecutable null.
+                ReInitializeLiveRunner();
+
+                if (!libraries.Any())
+                {
+                    return;
+                }
+
+                // generate import node for each library in input list
+                List<AssociativeNode> importNodes = new List<AssociativeNode>();
+                foreach (string lib in libraries)
+                {
+                    ProtoCore.AST.AssociativeAST.ImportNode importNode = new ProtoCore.AST.AssociativeAST.ImportNode();
+                    importNode.ModuleName = lib;
+
+                    importNodes.Add(importNode);
+                }
+                ProtoCore.CodeGenDS codeGen = new ProtoCore.CodeGenDS(importNodes);
+                string code = codeGen.GenerateCode();
+
+                SynchronizeInternal(code);
             }
-
-            // generate import node for each library in input list
-            List<AssociativeNode> importNodes = new List<AssociativeNode>();
-            foreach (string lib in libraries)
-            {
-                ProtoCore.AST.AssociativeAST.ImportNode importNode = new ProtoCore.AST.AssociativeAST.ImportNode();
-                importNode.ModuleName = lib;
-
-                importNodes.Add(importNode);
-            }
-            ProtoCore.CodeGenDS codeGen = new ProtoCore.CodeGenDS(importNodes);
-            string code = codeGen.GenerateCode();
-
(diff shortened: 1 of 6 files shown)


src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -1532,27 +1532,30 @@ public void ReInitializeLiveRunner()
...
+                    importNode.ModuleName = lib;
 
-            UpdateCmdLineInterpreter(code);
+                    importNodes.Add(importNode);
+                }
+                ProtoCore.CodeGenDS codeGen = new ProtoCore.CodeGenDS(importNodes);
+                string code = codeGen.GenerateCode();
+
+                SynchronizeInternal(code);
+            }
```
mmisol, line 1558: If you hide whitespace changes the change is much easier to see.

mmisol:
I did manual tests using the `PACOTES` package and this worked without issues 10/10 times.

src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -1532,27 +1532,30 @@ public void ReInitializeLiveRunner()
...
-            ProtoCore.CodeGenDS codeGen = new ProtoCore.CodeGenDS(importNodes);
-            string code = codeGen.GenerateCode();
+                // generate import node for each library in input list
+                List<AssociativeNode> importNodes = new List<AssociativeNode>();
+                foreach (string lib in libraries)
+                {
+                    ProtoCore.AST.AssociativeAST.ImportNode importNode = new ProtoCore.AST.AssociativeAST.ImportNode();
+                    importNode.ModuleName = lib;
 
-            UpdateCmdLineInterpreter(code);
```
aparajit-pratap, line 1555: @mmisol does this mean we can get rid of this method? I see it is only used in tests apart from here and we could update the tests to use `SynchronizeInternal` directly.

aparajit-pratap:
@mmisol do you also think that the changes in [this PR ](https://github.com/DynamoDS/Dynamo/pull/9957) are still needed now that we are enlarging the scope of the mutex?

src/Engine/ProtoScript/Runners/LiveRunner.cs
mmisol: Since it's public I don't know if we can just do that, but we can definitely mark it as obsolete. I'll do that now.

mmisol:
@aparajit-pratap Did you mean #9957 ? Maybe we don't, but I would be safer if @reddyashish could take a look at it, as he was involved with it.

aparajit-pratap:
> @aparajit-pratap Did you mean #9957 ? Maybe we don't, but I think it would be safer if @reddyashish could take a look at it, as he was involved with it and knows about that particular problem he ran into with more detail.

Yes, sorry, I forgot to paste the link. Yes, I was referring to #9957.

reddyashish:
@mmisol yes those changes should't be needed as that was a temporary fix at that time. Can you try reverting those changes and checking that the graph mentioned in that task works fine after downloading the Ampersand package. We saw this happening with this package for the first time. 

Original regression: https://jira.autodesk.com/browse/DYN-2101

You could also try running the tests in that PR after removing that temporary fix of disabling the runs.

mmisol:
Thanks for the input @reddyashish . I adapted the existing test so that we can keep checking that loading the dependency with the graph open does not cause any exceptions. Also, manually testing worked 10/10 times.

aparajit-pratap:
@mmisol could you look into why Dynamo.PackageManager.Tests.PackageLoaderTests.MixedPackageLoadTest is failing in the `DynamoSelfServe` job: https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/315/

mmisol:
@aparajit-pratap That test started failing after I removed the `DisableEngine` flag. I'll restore since it was not part of the original fix and it seems there is more to it than meets the eye.

PR #10591, opened 2020-04-23
Register custom node before package load reset

mmisol:
### Purpose

This fixes a problem where existing custom nodes in the home workspace
became unresolved after a package that contained binaries was loaded.

The cause of the problem was that the compiled function was not
available at the time of the VM reset. Now the data is registered on
package load, by queueing it in pendingCustomNodeSyncData. This results
in a CompileCustomNodeAsyncTask being scheduled before the update of
the home workspace graph takes place.

This also removes the DisableRun flag, ~hopefully~ without breaking any
tests.

### Declarations

### Reviewers

@reddyashish @aparajit-pratap @mjkkirschner 

### FYIs

@QilongTang


test/Libraries/PackageManagerTests/PackageLoaderTests.cs
```diff
@@ -545,7 +533,25 @@ public void MixedPackageLoadTest()
...
+            // Load a package which contains binaries, when the graph is open in the workspace. 
+            var loader = GetPackageLoader();
+            var pkg = loader.ScanPackageDirectory(Path.Combine(PackagesDirectory, "Mixed Package"));
+            loader.LoadPackages(new List<Package> { pkg });
 
+            // Custom node in workspace should not have any warnings on them.
+            var functionNodes = CurrentDynamoModel.CurrentWorkspace.Nodes.OfType<Function>();
+            Assert.AreEqual(1, functionNodes.Count());
+            var functionNode = functionNodes.First();
+            Assert.AreEqual(ElementState.Active, functionNode.State);
```
mjkkirschner, line 554: can we also assert this custom node returns the correct value?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1467,11 +1469,18 @@ protected void ResetEngineInternal()
...
             EngineController.TraceReconcliationComplete += EngineController_TraceReconcliationComplete;
+            EngineController.RequestCustomNodeRegistration += EngineController_RequestCustomNodeRegistration;
 
             foreach (var def in CustomNodeManager.LoadedDefinitions)
                 RegisterCustomNodeDefinitionWithEngine(def);
         }
 
+        private void EngineController_RequestCustomNodeRegistration(object sender, EventArgs e)
+        {
+            foreach (var def in CustomNodeManager.LoadedDefinitions)
```
mjkkirschner, line 1480: I'm curious about the timing of this - as I think @reddyashish mentioned - what if the package to be loaded contains custom nodes, would those exist in the `LoadedDefinitions` list before this call is made?

src/DynamoCore/Engine/EngineController.cs
```diff
@@ -518,8 +513,18 @@ private void OnLibraryLoaded()
         /// </summary>
         private void LibraryLoaded(object sender, LibraryServices.LibraryLoadedEventArgs e)
         {
-            if(e.LibraryPaths.Any())
+            if (e.LibraryPaths.Any())
```
mjkkirschner, line 516: so libraryPaths will include paths to custom nodes or only binaries?

mmisol: This is used for binaries only

src/DynamoCore/Models/DynamoModel.cs
mmisol: That seems with investigating. Taking a look at the interaction during the `PackgeLoaded` event, fired on the first stage of `LoadPackages`, there are some steps related to custom nodes that I did not fully capture. I'll take a closer look and see if I can find where this happens and add it to the internal documentation.

test/Libraries/PackageManagerTests/PackageLoaderTests.cs
mmisol: To be honest I'm not exactly sure what the example is doing, as I took it straight from @aparajit-pratap repro. I can probably check that it returns some value at least.

QilongTang: This example load the dyf first then followed by the dyn home workspace, at this point, we should have a working custom node and a dummy node. Then we load the package and make sure custom node still function correctly. I would encourage before and after the package load, we assert custom node in Active State and give correct value.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1039,6 +1039,7 @@ private void OnAsyncTaskStateChanged(DynamoScheduler sender, TaskStateChangedEve
         public void Dispose()
         {
             EngineController.TraceReconcliationComplete -= EngineController_TraceReconcliationComplete;
+            EngineController.RequestCustomNodeRegistration -= EngineController_RequestCustomNodeRegistration;
```
QilongTang, line 1042: Thank you for make sure we unsubscribe. Things like this take much more time to identify later and do memory analysis

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1467,11 +1469,18 @@ protected void ResetEngineInternal()
 
             EngineController.MessageLogged += LogMessage;
             EngineController.TraceReconcliationComplete += EngineController_TraceReconcliationComplete;
+            EngineController.RequestCustomNodeRegistration += EngineController_RequestCustomNodeRegistration;
```
aparajit-pratap, line 1472: I wonder if we should just move all these subscriptions to the constructor for `EngineController`?

src/DynamoCore/Engine/EngineController.cs
```diff
@@ -518,8 +513,18 @@ private void OnLibraryLoaded()
...
+            {
+                // Need to make compiled custom nodes available before resetting the VM
+                OnRequestCustomNodeRegistration();
                 OnLibraryLoaded();
+            }
+        }
+
+        internal event EventHandler RequestCustomNodeRegistration;
```
aparajit-pratap, line 524: Can this be a normal delegate as it doesn't require any arguments that are required by the `EventHandler` type?

src/DynamoCore/Engine/EngineController.cs
```diff
@@ -518,8 +513,18 @@ private void OnLibraryLoaded()
         /// </summary>
         private void LibraryLoaded(object sender, LibraryServices.LibraryLoadedEventArgs e)
         {
-            if(e.LibraryPaths.Any())
+            if (e.LibraryPaths.Any())
+            {
+                // Need to make compiled custom nodes available before resetting the VM
+                OnRequestCustomNodeRegistration();
                 OnLibraryLoaded();
```
aparajit-pratap, line 520: So now it looks like the custom node definitions could be scheduled to be added to the VM before it is reset in `OnLibraryLoaded`, which could again wipe out all of the definitions potentially. Wouldn't it be safer to call `OnRequestCustomNodeRegistration` after `OnLibraryLoaded`?

src/DynamoCore/Engine/EngineController.cs
mmisol: You are right, but on the other hand events can support multiple handlers if needed. If you feel strongly about this I can still change it, probably multiple handlers will never be needed.

src/DynamoCore/Engine/EngineController.cs
mmisol: This is where the method names might be misleading. I named this `OnRequestCustomNodeRegistration` mainly because it calls  `RegisterCustomNodeDefinitionWithEngine` downstream. However, that function does not register anything with the `RunCore`, that's the misleading part. What this does is fill a queue that gets processed afterwards, when calling `OnLibraryLoaded`.

src/DynamoCore/Engine/EngineController.cs
aparajit-pratap: It's minor, I'm not really particular about it, just a thought.

src/DynamoCore/Engine/EngineController.cs
aparajit-pratap: I'm not sure I understand. The `pendingCustomNodeSyncData` queue is processed in `EngineController.ProcessPendingCustomNodeSyncData`, which is called in `HomeWorkspaceModel.Run`. `Run` is not called from `OnLibraryLoaded`. My question is, what is the harm in calling `OnRequestCustomNodeRegistration` after the VM is reset in `OnLibraryLoaded`?

mmisol: > Run is not called from OnLibraryLoaded

You are right, I had overlooked this. Looking at the sequence of calls, the most appropriate place for this code should be in `HomeWorkspaceModel.LibraryLoaded`, which is the event handler triggering the run. That way we avoid doing extra work if no home workspace is open. Do you agree?

aparajit-pratap: Please see my comments below.

aparajit-pratap: Please see my comments below.

mmisol:
Hi @aparajit-pratap . In my comment https://github.com/DynamoDS/Dynamo/pull/10591#discussion_r414584253 I try to explain why the code was moved. Let me know if the point did not make sense and let's discuss, as the  suggested refactor does not take it into account.

aparajit-pratap:
> Hi @aparajit-pratap . In my comment [#10591 (comment)](https://github.com/DynamoDS/Dynamo/pull/10591#discussion_r414584253) I try to explain why the code was moved. Let me know if the point did not make sense and let's discuss, as the suggested refactor does not take it into account.

Yes, I did see your comment but I still propose the refactor in preference to your  [#10591 (comment)].

mmisol:
Hi @aparajit-pratap .

After our discussion I went ahead and reviewed the calls being done on package load regarding custom nodes. It turns out the handler that performs the queuing of the custom node is not registered for a custom node that is not used. That would mean going with the refactor you suggested could be a good idea, as it simplifies the code.

However, after doing that, the custom node test introduced here started to fail. The custom node appears with the warning both before and after installing the package.

Since this is a suggested refactor, but it doesn't quite work, can we leave it as a separate task?

mmisol:
Regarding the refactor and the results I see, I guess we'd also need to pass the `EngineController_RequestCustomNodeRegistration` event handler to the constructor of `EngineController` and also assign it before calling `OnLibraryLoaded`. I tried that too and it fixes the issue with the custom node. However, the node from the `QRImage` package disappears from the design after loading the package, so there seems to be something else going on.

Again, I think a refactor would be nice too, but we should probably take a look at this in a more global context rather than this PR.

aparajit-pratap:
@mmisol thanks for trying out the suggested changes. AFAIK it shouldn't have caused any failures as it is essentially doing the same thing as your PR. The sequence of calls is exactly the same except that the code is in different methods. Anyway, I don't wish to hold you up. I'm okay with these changes as-is. Thanks a lot for working on this.

mmisol:
No problem @aparajit-pratap ! You are more than welcome.

Can I ask for your approval if you feel confident about this PR? Thanks!

PR #10548, opened 2020-04-07
[DYN-2322] Add PythonNet 3.7 CPython as a selectable engine.

reddyashish:
# Purpose

This PR is a cloned from the original PR(https://github.com/DynamoDS/Dynamo/pull/10472) by @Dewb. 

The auto-completion part for the IronPython is fixed. The IronPythonCompletionData and IronPythonCompletionProvider classes have been moved from the PythonNodeModelsWPF project to the DSIronPython project. We can extend this similarly with the new DSCPython. 

There is only one failing test left that is related to function marshaling: FirstClassFunctions

### Declarations

Check these if you believe they are true

- [ ] The codebase is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.

### Reviewers

@DynamoDS/dynamo

PR #10596, opened 2020-04-24
Fixing some of the issues for PythonNet3.7 on DSCPython3 branch. 

reddyashish:
### Purpose

This PR is a followup PR to the base PR(https://github.com/DynamoDS/Dynamo/pull/10548).

1) Removed the failing FirstClassFunctions as that is not being supported.
2) The test VerifyIronPythonLoadedAssemblies was failing on self-serve because the IronPython.dll was also being copied into the "nodes" folder and the parent bin folder. This was causing an issue when the test was trying to load the dll from multiple locations. After this change, the IronPython.dll is just copied to the parent bin folder and the test passes. 

### Declarations

### Reviewers

@mjkkirschner @Dewb @QilongTang

### DIFF
--- a/src/Libraries/DSCPython/CPythonEvaluator.cs
+++ b/src/Libraries/DSCPython/CPythonEvaluator.cs
@@ -40,15 +40,6 @@ namespace DSCPython
             IList bindingNames,
             [ArbitraryDimensionArrayImport] IList bindingValues)
         {
-            // TODO - it would be nice if users could modify a preference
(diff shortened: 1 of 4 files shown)


QilongTang:
@reddyashish I remember we have a few comments code need to be cleaned up?

reddyashish:
Yes, looking into it to see if we would need that

QilongTang:
LGTM

PR #10600, opened 2020-04-24
fix display of documentation browser with display scaling.

mjkkirschner:
### Purpose
https://jira.autodesk.com/browse/DYN-2644
Text size of the documentation browser content was too large, and would not react with DPI scaling modes correctly.

I implemented a solution using css zoom - very similar to the solution used by the civil team to address dpi scaling in the LibraryMSWebBrowser extension.

While addressing this I found 3 other issues that are now fixed.
* some error links had a `.` at the end and would lead to a 404. - these are fixed.
* many images were not embedded - I've embedded these as base64 encoded pngs following the pattern already started in this extension.
* The images had a fixed width, which was larger than the displayable area inside the extension, as a result there was always an annoying horizontal scroll bar - I have added a style for all images with a max width slightly smaller than the max size of the sidebar, so theres no horizontal scroll unless the extension is shrunk down.

I've tested a few error pages on 100% and 200% scaling.

The DPI change and global image style are injected into the page after we remove other script tags.

The function sets the zoom, and then adjusts the width of the body to be slightly less than 100% - I found that 100% width would add a scroll bar - I believe this is because of extra padding the other inline styles add.

The DPI change function is also triggered when the DPIchanged event of the browser is raised- so the browser responds correctly if the user changes DPI scaling in the middle of a Dynamo session.

images:
Screen Shot 2020-04-23 at 5 23 50 PM
Screen Shot 2020-04-23 at 5 22 59 PM

### Declarations

FYI @Amoursol


src/Engine/ProtoCore/Properties/Resources.en-US.resx
```diff
@@ -885,7 +885,7 @@
   </data>
   <data name="ArgumentNullException" xml:space="preserve">
     <value>Value cannot be null.
-Parameter name: {0}. href=ArgumentNullException.html.</value>
+Parameter name: {0}. href=ArgumentNullException.html</value>
```
mjkkirschner, line 888: this was preventing these pages from ever loading.

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
```diff
@@ -19,7 +19,41 @@ public class DocumentationBrowserViewModel : NotificationObject, IDisposable
...
+            document.body.style.width = widthPercentage;
+        }
+        adaptDPI() </script>";
+
+        private const string globalimgStyle = @"<style> img{
+        min-width:300px;
+        max-width:450px;
+        width:100%;
+                }
+        </style>";
```
mmisol, line 56: Just a note about this. Defining the style like this, without CSS class, will affect any `` tag, including those that are included by library developers in their html help files. That means in practice we are putting restrictions to their pages, which may be a good thing taking into account the scrollbar issue, but could be limiting if they wanted to show a really big image.

src/Engine/ProtoCore/Properties/Resources.en-US.resx
mmisol: Good catch!

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
mjkkirschner: that's a good point, I guess for max flexibility I can just add this style to each of our pages that uses images. Not too much work, what do you think?

mmisol: @mjkkirschner That could work as well

mjkkirschner: done.

mjkkirschner:
@QilongTang yes, it's very very similar.

mjkkirschner:
updated the failing test and added a new one - waiting for test results.

src/DocumentationBrowserViewExtension/Docs/Deprecated.html
```diff
@@ -19,7 +19,7 @@
 <style>
 body {
     color: white;
-    font-size: 20px;
+    font-size: 14px;
```
aparajit-pratap, line 22: Have you changed the default over here?

mjkkirschner: yes, I changed the default font size on all pages.

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
```diff
@@ -19,7 +19,34 @@ public class DocumentationBrowserViewModel : NotificationObject, IDisposable
...
+            if (window.screen.deviceXDPI != undefined)
+            {
+                dpi = window.screen.deviceXDPI;
+            }
+            else
+            {
+                var tmpNode = document.createElement('DIV');
+                tmpNode.style.cssText = 'width:1in;height:1in;position:absolute;left:0px;top:0px;z-index:99;visibility:hidden';
+                document.body.appendChild(tmpNode);
+                dpi = parseInt(tmpNode.offsetWidth);
```
aparajit-pratap, line 34: So what is this doing?

mjkkirschner: it's changing the zoom style based on the dpi - if this was not done, then the text would be extremely small when using a high dpi screen with display scaling over 100%.

mjkkirschner:
@QilongTang @alfredo-pozo 
Screen Shot 2020-04-27 at 2 45 48 PM

this PR passes in 2 hours on the serial self serve - leads me to believe this is a machine issue like @alfredo-pozo surmised.

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
aparajit-pratap: This looks like it's just returning a `dpi` and the style is being set in `adaptDPI` based on the `dpi` value. What I was trying to understand is how this `tmpNode.offsetWidth`, whatever it is, is related to the `dpi`. Just curious.

mjkkirschner: ah, sorry, this was implemented by lepus team - my understanding from them was that this was a standard polyfill to get the DPI of an HTML element. Using the value on the WPF side yielded incorrect results sometimes.

mmisol: It's actually a clever way of measuring the DPI. It's creating a `div` element with fixed size of 1 inch x 1 inch. The element is absolutely positioned so that it's not affected by anything else in the page, but it's visibility is hidden so we cannot see it either. The `offsetWidth` property gives the actual size in pixels of the element. Since the CSS size was specified as `1in` that's the amount of pixels that equate to 1 inch, which is the DPI.

aparajit-pratap: Clever indeed!

mjkkirschner:
I reran the PR validation job and it passed - merging. Will send 2.6 cherry pick after.

PR #10605, opened 2020-04-27
Handle runtime table gaps on code block deletion

mmisol:
### Purpose

When the runtime table are built there is an implicit assumption that
the code block ids are consecutive. However, that is not always the
case, as the deletion of a procedure causes the deletion of its child
code blocks, which may generate gaps in the id numbering.

In order to make the code resiliant to these gaps, the runtime tables
are sized based on the largest code block id, rather than in the amount
of code blocks.

### Declarations

### Reviewers

### FYIs

@QilongTang

### DIFF
--- a/src/Engine/ProtoCore/Core.cs
+++ b/src/Engine/ProtoCore/Core.cs
@@ -770,7 +770,13 @@ namespace ProtoCore
         private int GetRuntimeTableSize()
         {
             // Due to the way this list is constructed, the largest id is the one of the last block.
-            return CompleteCodeBlockList.Last().codeBlockId + 1;
+            var lastBlock = CompleteCodeBlockList.Last();
+            // If the last block has children, then its last child has the largest id.
+            if (lastBlock.children.Count > 0)
+            {
+                lastBlock = lastBlock.children.Last();
+            }
+            return lastBlock.codeBlockId + 1;
         }
 
         public string GenerateTempVar()
--- a/src/Engine/ProtoCore/Core.cs
+++ b/src/Engine/ProtoCore/Core.cs
@@ -771,6 +771,8 @@ namespace ProtoCore
         {
             // Due to the way this list is constructed, the largest id is the one of the last block.
             var lastBlock = CompleteCodeBlockList.LastOrDefault();
+            // If there are no code blocks yet, then the required size for tables is 0.
+            // This happens when the first code block is being created and its id is being generated.
             if (lastBlock == null)
             {
                 return 0;
--- a/src/Engine/ProtoCore/Core.cs
+++ b/src/Engine/ProtoCore/Core.cs
@@ -367,21 +367,6 @@ namespace ProtoCore
             }
         }
 
-        //private void RemoveCodeBlockRecursive(CodeBlock root, int cbID)
-        //{
-        //    for (int i = root.children.Count - 1; i >= 0; i--)
-        //    {
-        //        if (root.children[i].codeBlockId == cbID)
-        //        {
-        //            root.children.RemoveAt(i);
-        //        }
-        //        else
-        //        {
-        //            RemoveCodeBlockRecursive(root.children[i], cbID);
-        //        }
-        //    }
-        //}
-
         /// <summary>
         /// Reset the VM state for delta execution.
         /// </summary>
--- a/src/Engine/ProtoCore/Core.cs
+++ b/src/Engine/ProtoCore/Core.cs
@@ -763,13 +763,13 @@ namespace ProtoCore
         }
 
         /// <summary>
-        /// Gets the size of the runtime table. Note: since blocks are stored consecutively
-        /// but may have gaps due to procedures being delete, this is based on largest id
-        /// rather than amount of blocks.
+        /// Gets the size to be used for runtime tables of symbols, procedures and instruction streams.
+        /// Note: since blocks are stored consecutively but may have gaps due to procedures being deleted,
+        /// this is based on largest id rather than amount of blocks.
         /// </summary>
         private int GetRuntimeTableSize()
         {
-            // Due to the way this list is constructed, the largest id is the one of the las block.
+            // Due to the way this list is constructed, the largest id is the one of the last block.
             return CompleteCodeBlockList.Last().codeBlockId + 1;
         }
 
--- a/src/Engine/ProtoCore/Core.cs
+++ b/src/Engine/ProtoCore/Core.cs
@@ -367,6 +367,21 @@ namespace ProtoCore
             }
         }
 
+        //private void RemoveCodeBlockRecursive(CodeBlock root, int cbID)
+        //{
+        //    for (int i = root.children.Count - 1; i >= 0; i--)
+        //    {
+        //        if (root.children[i].codeBlockId == cbID)
+        //        {
+        //            root.children.RemoveAt(i);
+        //        }
+        //        else
+        //        {
+        //            RemoveCodeBlockRecursive(root.children[i], cbID);
+        //        }
+        //    }
+        //}
+
         /// <summary>
         /// Reset the VM state for delta execution.
         /// </summary>
@@ -727,7 +742,7 @@ namespace ProtoCore
             // The TypeSystem is a record of all primitive and compiler generated types
             DSExecutable.TypeSystem = TypeSystem;
 
-            RuntimeTableIndex = CompleteCodeBlockList.Count;
+            RuntimeTableIndex = GetRuntimeTableSize();
 
 
             // Build the runtime symbols
@@ -760,6 +775,17 @@ namespace ProtoCore
             DSExecutable.Configurations = Configurations;
             DSExecutable.CodeToLocation = CodeToLocation;
             DSExecutable.CurrentDSFileName = CurrentDSFileName;           
+        }
+
+        /// <summary>
+        /// Gets the size of the runtime table. Note: since blocks are stored consecutively
+        /// but may have gaps due to procedures being delete, this is based on largest id
+        /// rather than amount of blocks.
+        /// </summary>
+        private int GetRuntimeTableSize()
+        {
+            // Due to the way this list is constructed, the largest id is the one of the las block.
+            return CompleteCodeBlockList.Last().codeBlockId + 1;
         }
 
         public string GenerateTempVar()
--- a/src/Engine/ProtoCore/Core.cs
+++ b/src/Engine/ProtoCore/Core.cs
@@ -767,10 +767,15 @@ namespace ProtoCore
         /// Note: since blocks are stored consecutively but may have gaps due to procedures being deleted,
         /// this is based on largest id rather than amount of blocks.
         /// </summary>
-        private int GetRuntimeTableSize()
+        internal int GetRuntimeTableSize()
         {
             // Due to the way this list is constructed, the largest id is the one of the last block.
-            var lastBlock = CompleteCodeBlockList.Last();
+            var lastBlock = CompleteCodeBlockList.LastOrDefault();
+            if (lastBlock == null)
+            {
+                return 0;
+            }
+
             // If the last block has children, then its last child has the largest id.
             if (lastBlock.children.Count > 0)
             {
--- a/src/Engine/ProtoCore/DSASM/Executable.cs
+++ b/src/Engine/ProtoCore/DSASM/Executable.cs
@@ -155,10 +155,10 @@ namespace ProtoCore.DSASM
             procedureTable = procTable;
 
             isBreakable = isBreakableBlock;
+            codeBlockId = core.GetRuntimeTableSize();
             core.CompleteCodeBlockList.Add(this);
-            this.codeBlockId = core.CompleteCodeBlockList.Count - 1;
 
-            symbols.RuntimeIndex = this.codeBlockId;
+            symbols.RuntimeIndex = codeBlockId;
 
             if (core.ProcNode != null)
             {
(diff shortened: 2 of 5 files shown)


mmisol:
Making this a WIP since I still need to add a test, but please take a look.

aparajit-pratap:
@mmisol so is the reason for the crash the deletion of a DS function? Could you add some details to explain the reason for the crash?

src/Engine/ProtoCore/Core.cs
```diff
@@ -762,6 +762,17 @@ public void GenerateExecutable()
...
         }
 
+        /// <summary>
+        /// Gets the size of the runtime table. Note: since blocks are stored consecutively
+        /// but may have gaps due to procedures being delete, this is based on largest id
+        /// rather than amount of blocks.
+        /// </summary>
+        private int GetRuntimeTableSize()
+        {
+            // Due to the way this list is constructed, the largest id is the one of the las block.
```
mjkkirschner, line 772: (last)sp

src/Engine/ProtoCore/Core.cs
```diff
@@ -762,6 +762,17 @@ public void GenerateExecutable()
             DSExecutable.CurrentDSFileName = CurrentDSFileName;           
         }
 
+        /// <summary>
+        /// Gets the size of the runtime table. Note: since blocks are stored consecutively
+        /// but may have gaps due to procedures being delete, this is based on largest id
+        /// rather than amount of blocks.
+        /// </summary>
+        private int GetRuntimeTableSize()
```
mjkkirschner, line 770: can we be more descriptive ? Runtime table of what? Codeblocks? functions?

QilongTang:
Is this PR still WIP? I guess for unit tests?

mmisol:
@aparajit-pratap Yes, the relevant related code is in `SetFunctionInactive` after the comment that says `// Remove codeblock defined in procNode from CodeBlockList and CompleteCodeBlockList`. That's where the gap is created.

aparajit-pratap:
What about line https://github.com/DynamoDS/Dynamo/blob/c00c80bee60f0d8d497af87f7774bcab463c23e6/src/Engine/ProtoCore/Core.cs#L383
Is `RuntimeTableIndex` being set correctly here?

mmisol:
@aparajit-pratap I would say so. As you probably know, we only have exactly 1 top-level scope, so in practice this is setting the `RuntimeTableIndex` to 1.

mmisol:
@QilongTang I added the test and removed the `WIP` tag just now. Please take a look

src/Engine/ProtoCore/Core.cs
```diff
@@ -770,7 +770,13 @@ public void GenerateExecutable()
...
         {
             // Due to the way this list is constructed, the largest id is the one of the last block.
-            return CompleteCodeBlockList.Last().codeBlockId + 1;
+            var lastBlock = CompleteCodeBlockList.Last();
+            // If the last block has children, then its last child has the largest id.
+            if (lastBlock.children.Count > 0)
+            {
+                lastBlock = lastBlock.children.Last();
+            }
+            return lastBlock.codeBlockId + 1;
```
aparajit-pratap, line 779: If the assertion that the last codeblock always has the largest id is true, then this function needs to be recursive.

mmisol:
@aparajit-pratap Thanks for the find! I was indeed able to reproduce it, but only using Josh's design. The problem arises when the last code block contains children. In that case the last block id should come from the last child. I pushed a fix for this and it no longer crashes. Please give it a try.

By the way, I do agree that the problem deserves more investigation, but I also think that if we can prevent the crash with this, it provides short term value.

src/Engine/ProtoCore/Core.cs
mmisol: I have wondered about that too, but if you take a look at the part of the code I mention here https://github.com/DynamoDS/Dynamo/pull/10605#issuecomment-620108435, you'll see that only that the code doesn't do recursion, it also looks only at the first level of children. Maybe these code blocks can contain only one level in practice?

mmisol:
After further testing I was able to make it fail again, also using Josh's design by way of undo/redo several times.

I think there is a deeper problem going on with undo/redo. Taking a look at the `CodeBlockList` I see there are two code blocks with the same id, located in different parts of the tree. I'd guess  these ids are meant to be unique. Here is a screen capture showing this:

Screen Shot 2020-04-28 at 12 20 49 PM

mmisol:
My last push fixed that issue. I still have one more crash to tackle:
```
An item with the same key has already been added.

   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at ProtoScript.Runners.ChangeSetComputer.GetDeltaAstListAdded(IEnumerable`1 addedSubTrees) in C:\Users\t_misom\dev\github\Dynamo\src\Engine\ProtoScript\Runners\LiveRunner.cs:line 493
   at ProtoScript.Runners.ChangeSetComputer.GetDeltaASTList(GraphSyncData syncData) in C:\Users\t_misom\dev\github\Dynamo\src\Engine\ProtoScript\Runners\LiveRunner.cs:line 753
   at ProtoScript.Runners.LiveRunner.SynchronizeInternal(GraphSyncData syncData) in C:\Users\t_misom\dev\github\Dynamo\src\Engine\ProtoScript\Runners\LiveRunner.cs:line 1470
   at ProtoScript.Runners.LiveRunner.UpdateGraph(GraphSyncData syncData) in C:\Users\t_misom\dev\github\Dynamo\src\Engine\ProtoScript\Runners\LiveRunner.cs:line 1240
   at Dynamo.Engine.LiveRunnerServices.UpdateGraph(GraphSyncData graphData, Boolean verboseLogging) in C:\Users\t_misom\dev\github\Dynamo\src\DynamoCore\Engine\LiveRunnerServices.cs:line 97
   at Dynamo.Engine.EngineController.UpdateGraphImmediate(GraphSyncData graphSyncData) in C:\Users\t_misom\dev\github\Dynamo\src\DynamoCore\Engine\EngineController.cs:line 462
   at Dynamo.Scheduler.UpdateGraphAsyncTask.HandleTaskExecutionCore() in C:\Users\t_misom\dev\github\Dynamo\src\DynamoCore\Scheduler\UpdateGraphAsyncTask.cs:line 106
   at Dynamo.Scheduler.AsyncTask.Execute() in C:\Users\t_misom\dev\github\Dynamo\src\DynamoCore\Scheduler\AsyncTask.cs:line 189
```

mmisol:
I managed to see what caused that last issue, but I don't think it's related to the imperative block undo/redo. There is a discrepancy between the incoming `syncData` and the `currentSubTreeList` in the `changeSetComputer`. An AST node that already exists is trying to be added, which results in the above error. The node is a `DSCore.List.UniqueItems@var[]..[]`. The code that is coming in `syncData` is actually the same that is already in the `changeSetComputer`:
```
{var_291a044825354a55a49807d44bdb1494 = __CreateFunctionObject(DSCore.List.UniqueItems, 1, [], [null], true);}
```

Do you think we should tackle this a separate issue?

mmisol:
Adding to my previous comment, I was able to reproduce that issue with a totally unrelated graph.
Do you know if we have seen that before @mjkkirschner ?
Screen Shot 2020-04-28 at 4 21 32 PM

src/Engine/ProtoCore/Core.cs
aparajit-pratap: The `BfsBuildSequenceTable` is recursive.

src/Engine/ProtoCore/Core.cs
```diff
@@ -762,6 +762,28 @@ public void GenerateExecutable()
...
+        /// this is based on largest id rather than amount of blocks.
+        /// </summary>
+        internal int GetRuntimeTableSize()
+        {
+            // Due to the way this list is constructed, the largest id is the one of the last block.
+            var lastBlock = CompleteCodeBlockList.LastOrDefault();
+            if (lastBlock == null)
+            {
+                return 0;
+            }
```
aparajit-pratap, line 777: Could you explain this change?

src/Engine/ProtoCore/Core.cs
mmisol: @aparajit-pratap Yes, but it's using `CodeBlockList` which is actually a tree. Both this method and the one I pointed to in my previous comment use `CompleteCodeBlockList` which contains the same nodes but is flattened, to one level of children it would seem.

src/Engine/ProtoCore/Core.cs
mmisol: Sure. I'll add a comment in code.

src/Engine/ProtoCore/Core.cs
aparajit-pratap: There should be a codeblock for each language block (such as `[Imperative {...}]`, `if-else`, `while`, `for` etc., and function definition) so there can be any number of nesting or children of child codeblocks and so on.

src/Engine/ProtoCore/Core.cs
aparajit-pratap: I meant for reviewers to understand the change.

src/Engine/ProtoCore/Core.cs
aparajit-pratap: Ok, I thought so too that `CompleteCodeBlockList` was flattened. If that is confirmed, then I guess checking for one level is fine but I'm not quite sure why it is the last one that is always the largest.

mmisol: Could you provide an example where this is the case? It would definitely be nice to have a repro.

src/Engine/ProtoCore/Core.cs
mmisol: Sorry, I thought if it was added to the code it would be useful for future reference too. I'll copy it here as well:
```
            // If there are no code blocks yet, then the required size for tables is 0.
            // This happens when the first code block is being created and its id is being generated.
```

src/Engine/ProtoCore/Core.cs
mmisol: > Ok, I thought so too that CompleteCodeBlockList was flattened. If that is confirmed, then I guess checking for one level is fine but I'm not quite sure why it is the last one that is always the largest.

I think I know why. If you take a look at the constructor of `CodeBlock`, the created code block is added immediately to the `CompleteCodeBlockList` using `Add`: `core.CompleteCodeBlockList.Add(this);`

aparajit-pratap:
@mmisol I think we can file the last issue you saw with the delta computer separately but could you describe what steps you followed to produce the last crash?

mmisol:
@aparajit-pratap The last crash was the `changeSetComputer` one. The previous to that was also reproduced by undoing and redoing with Josh's graph about three times, but without going too fast or you hit the `changeSetComputer` one.

src/Engine/ProtoCore/Core.cs
aparajit-pratap: Then how can it be that the children can have the largest id? The largest should always be at the top level, i.e. `CompleteCodeBlockList`?

mmisol: I don't know if I understood what you meant. What I was saying is that the `List.Add` method adds to the end of the list, and since we always get a bigger id each time, by construction the biggest id should be at the end of the flattened list. Does that make sense?

aparajit-pratap: Yes, I understood that. If we're adding to the end of the `CompleteCodeBlockList`, we simply need to check the size of that list or the `codeBlockId` of the last item in `CompleteCodeBlockList`. Why are you checking for the last item in the `CompleteCodeBlockList`?

mmisol: > we simply need to check the size of that list or the codeBlockId of the last item in CompleteCodeBlockList

If we go back to the purpose of this PR, those two do not always match, so it's not the same to go with one or the other. Just wanted to make that clear.

> Why are you checking for children of the last item in the CompleteCodeBlockList?

By way of undo/redo in Josh's graph I was able to see a crash where the last code block contained children with larger ids than the id of its parent or the amount of elements in the list. It does seem strange that the children themselves were not part of the list, so it could be a byproduct of the undo/redo bug.

aparajit-pratap: I'm not so sure I understand the changes, to be honest. If you feel confident you are welcome to get this approved by someone else on the team. I'll do some investigation on my own to try to convince myself of your changes :)

src/Engine/ProtoCore/Core.cs
```diff
@@ -762,6 +762,30 @@ public void GenerateExecutable()
...
+            // This happens when the first code block is being created and its id is being generated.
+            if (lastBlock == null)
+            {
+                return 0;
+            }
+
+            // If the last block has children, then its last child has the largest id.
+            if (lastBlock.children.Count > 0)
+            {
+                lastBlock = lastBlock.children.Last();
```
QilongTang, line 784: I guess I had similar question here why we are checking children of last block but other than this, everything else looks fine to me..

mmisol: I logged a ticket for the undo/redo bug. Maybe we can revisit this after we have fixed that? I feel they may be related.

I could also place a comment here mentioning that.

QilongTang: Thanks, would be good if you do that

PR #10607, opened 2020-04-27
Python ux improvements

SHKnudsen:
### Note
This is meant as a working PR, that will progress into the final changes, so we have a central place to discuss the approach we want to take on these changes.

### Purpose

This PR contains the UX work being done around the new Python3 engine.

### Engine Selector in Python editor
The engine selector dropdown contains two engines the `IronPython2` and the `CPython3`
EngineSelector

EngineLabel

### Python Engine Label
Added a label on the node which display which python engine the node is using, this also expands the whitespace on the python node.
EngineLabel

 ### IronPython Notifications
ATM new notifications are logged everytime a Python node set to use the `IronPython2` engine is added. Notifications are not aggregated and they are not removed until the user actively dismiss the notification. 
IronPythonNotifiaction

IronPythonNotifications

### IronPython Graph warning
The graph warning checks the .dyn file if there are any python nodes with the `Engine` property set to `IronPython2` and displays a dialog if there is.

OpenGraphDialog

OpenGraphDialog

### DocumentaionBrowser
The graph warning dialog has a `More Information` button that will open a more detailed explanation of the warning in the `DocumentationBrowser`. The description outline why the move to Python 3 is being made and what the user needs to be aware of.
DocumentationBrowser

### Declarations

### Reviewers

@mjkkirschner 
@Dewb 
@QilongTang 
@Amoursol 

### FYIs

Except for the code comments it might be good to get inputs on text for both the dialog and the notification


src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -1205,6 +1206,19 @@ public void SetNameFromNodeNameAttribute()
...
 
+        /// <summary>
+        /// Logs a user defined messages to the Dynamo Logger, this notifiaction will be dispalyed
+        /// in the Notification Extension.
+        /// </summary>
+        /// <param name="dynamoModel"></param>
+        /// <param name="shortMessage"></param>
+        /// <param name="detailedMessage"></param>
+        /// <param name="title"></param>
+        public void Notification(DynamoModel dynamoModel, string shortMessage, string detailedMessage, string title = "Notification")
```
QilongTang, line 1217: Maybe `Internal` for now and make it a verb like `LogNotification`. Our team usually keep the new APIs internal and need to discuss when to expose them.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1693,6 +1693,7 @@ private void SetPeriodicEvaluation(WorkspaceModel ws)
           bool forceManualExecutionMode,
           out WorkspaceModel workspace)
         {
+            CheckForIronPythonDependencies(fileContents);
```
QilongTang, line 1696: So this will happen before Dynamo tries to open the file.. 
1. About the behavior. This may be less ideal than after the graph has been opened so people can see the graph, then the dialog.. I guess we did not make it clear @mjkkirschner @Amoursol .
2. Also curious about the performance impact on this as well since we can do the check in the `FromJson` function below as well to reduce one time of Json deserialization

mjkkirschner: I don't think this needs to be done here - As @QilongTang said theres a few reasons.

1. would be nice if users could see the graph when the warning displays, instead of the start page.

2. The entire file is being parsed by json.net a second time, I think you can just let the graph get opened, and then check the deserialized nodes later using the engine property, instead of checking the JSON. It's better to rely on the runtime apis, even internal ones I think than to use the serialized json via string comparison.

mjkkirschner: on second thought - I guess one reason you went with this approach was to avoid needing to reference PythonNodeModel from DynamoCore - and this is a good thing. We do not want to have a compile time or runtime dependency between them.

I would ask that you consider alternatives to the JSON parsing with the constraint mentioned above and see if you can think of another approach. If possible I would also avoid reflection if you need to scan everynode as that will be slow.

⚠️ 
Here's another idea...
Why don't you just have the PythonNodeModel constructor raise this notification itself, theres no reason to need to scan anything.

radumg: @mjkkirschner yep, @SHKnudsen was trying to avoiding that circular dependency.

If we're ok with raising this `after` the workspace has loaded, then we could look to move all of this into an extension, should give us the required event hooks and can safely reference `PythonNodeModels` project too, let us know if there's any objections to going that route instead ?

src/Libraries/PythonNodeModels/Properties/Resources.en-US.resx
```diff
@@ -166,4 +166,10 @@
   <data name="PythonStringPortDataScriptToolTip" xml:space="preserve">
     <value>Python script to run.</value>
   </data>
+  <data name="IronPythonNotificationDetailedMessage" xml:space="preserve">
+    <value>This graph currently contains python nodes that are using the old IronPython Engine which will be deprecated in later versions.Consider updating these nodes to use the new CPython 3.7 Engine.</value>
```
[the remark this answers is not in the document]
radumg: if we move to an extension, then that'd be the only place this is needed in the end.

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -18,6 +18,29 @@ namespace PythonNodeModels
 {
     public abstract class PythonNodeBase : VariableInputNode
     {
+        public static readonly string CPythonEngine = "CPython3";
+        public static readonly string DefaultPythonEngine = "IronPython2";
+        public static List<string> PythonEngines { get { return new List<string> { DefaultPythonEngine, CPythonEngine }; } }
```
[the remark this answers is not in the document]
radumg: if we're not using an Enum in Core, i'd suggest making the changes required there too.

src/DynamoCore/Graph/Nodes/NodeModel.cs
radumg: think we can avoid adding this altogether.

SHKnudsen: why would we avoid it? as discussed on slack we are adding an API to the nodemodel to log notifications?

src/DynamoCore/Models/DynamoModel.cs
mjkkirschner: @radumg @SHKnudsen @QilongTang  - hmm - heres some thoughts.
* we know we're eventually going to want to create a migration assistant, it seems likely it will need some application level access - what if we start this now by creating a new extension like @radumg suggests.
* we can use that extension to unify both the warning dialog and the warning notifications on the python nodes.
* The extension can reference python without adding forcing a reference on python to any of our core extensions or binaries.
* We need to design that extension in such a way that it is robust and does not crash, even if the python or IronPython binaries are completely missing, as this may be the case in the future.

still thinking through this for now.

mjkkirschner: just to be clear - I was not imagining this new extension to have any UI at this time - but only act as a centralized location for python migration analysis, raising of warnings and notifications, etc.

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -18,6 +18,29 @@ namespace PythonNodeModels
 {
     public abstract class PythonNodeBase : VariableInputNode
     {
+        public static readonly string CPythonEngine = "CPython3";
```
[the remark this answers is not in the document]
QilongTang: @SHKnudsen and @radumg Please hold on to that idea. Our team have quickly turn around to use enum for this. I will try to bring the changes to master soon for your reference, but feel free to preview in https://github.com/DynamoDS/Dynamo/pull/10565

src/Libraries/PythonNodeModels/PythonNode.cs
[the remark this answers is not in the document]
QilongTang: See my comments above

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -18,6 +18,29 @@ namespace PythonNodeModels
...
+        public static readonly string DefaultPythonEngine = "IronPython2";
+        public static List<string> PythonEngines { get { return new List<string> { DefaultPythonEngine, CPythonEngine }; } }
+
+        private string engine = DefaultPythonEngine;
+
+        /// <summary>
+        /// Evaluation engine used to execute the Python script. 
+        /// Currently supported values: "IronPython2" (the default)
```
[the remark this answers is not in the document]
QilongTang: CPython will be the new default for Revit 2022 or some release later this year. So we will stage the change of default engine

src/DynamoCore/Models/DynamoModel.cs
SHKnudsen: @mjkkirschner @QilongTang 

added the extension to handle the node and graph warnings, along with a few other small changes

src/PythonMigrationViewExtension/PythonMigrationViewExtension.csproj
```diff
@@ -0,0 +1,106 @@
...
+      <Project>{51BB6014-43F7-4F31-B8D3-E3C37EBEDAF4}</Project>
+      <Name>DynamoCoreWpf</Name>
+      <Private>False</Private>
+    </ProjectReference>
+    <ProjectReference Include="..\DynamoCore\DynamoCore.csproj">
+      <Project>{7858FA8C-475F-4B8E-B468-1F8200778CF8}</Project>
+      <Name>DynamoCore</Name>
+      <Private>False</Private>
+    </ProjectReference>
+    <ProjectReference Include="..\Libraries\PythonNodeModels\PythonNodeModels.csproj">
```
mjkkirschner, line 77: @SHKnudsen @radumg @QilongTang I would like to be careful about this - consider a future where PythonNodeModel might be completely missing from Dynamo - I'm okay if this extension does not work - I just want to make sure it does not crash Dynamo. (ie, use try catches if we need to).

mjkkirschner: note my original point above 
```
We need to design that extension in such a way that it is robust and does not crash, even if the python or IronPython binaries are completely missing, as this may be the case in the future. The best way to do this ins unclear... try catch, reflection, bundling python itself into this extension...
```

QilongTang: @mjkkirschner I think the dependency here will be refactored later in this quarter when we work on running Dynamo without Python, I would vote for making this part of Dynamo so we can assess this problem together as a team. But depending on node specific knowledge etc is actually a good point, currently I dont see a way around it, I think we may want to just merge python nodes and this extension to be in a single project and move them out from Dynamo code base later by looking at it. But I dont think we need to to it as part of this PR. What do you think? @mjkkirschner

mjkkirschner: @QilongTang @SHKnudsen @aparajit-pratap I'm fine with that approach - we can handle it inside the team.

src/Libraries/PythonNodeModelsWpf/Controls/EngineLabel.xaml
```diff
@@ -0,0 +1,35 @@
...
+        <Grid.RowDefinitions>
+            <RowDefinition Height="15" />
+            <RowDefinition Height="*" />
+            <RowDefinition Height="*" />
+        </Grid.RowDefinitions>
+
+        <!--<TextBlock Foreground="{StaticResource headerForegroundInactive}"
+                    Grid.Row="1"
+                   Text="Python Engine:"
+                   FontSize="8" />-->
```
QilongTang, line 29: According to slack conversation, seems we can remove this part for clarity. Thank you for providing this option!

src/Libraries/PythonNodeModelsWpf/Controls/EngineLabel.xaml
```diff
@@ -0,0 +1,35 @@
...
+        <TextBlock Foreground="{StaticResource headerForegroundInactive}"
+                    Grid.Row="1"
```
QilongTang, line 31: @SHKnudsen Can you explain why we are only using one of thr rows defined?

src/Libraries/PythonNodeModelsWpf/PythonNode.cs
```diff
@@ -39,6 +40,10 @@ public void CustomizeView(PythonNode nodeModel, NodeView nodeView)
             nodeView.MouseDown += view_MouseDown;
             nodeModel.DeletionStarted += NodeModel_DeletionStarted;
             nodeModel.Disposed += NodeModel_Disposed;
+
+            nodeView.PresentationGrid.Visibility = Visibility.Visible;
+            nodeView.PresentationGrid.DataContext = this.pythonNodeModel;
+            nodeView.PresentationGrid.Children.Add(new EngineLabel());
```
QilongTang, line 46: Cool. You may want to merge the latest debug modes change in and put this block under debug mode

src/PythonMigrationViewExtension/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+//      Major Version
+//      Minor Version
+//      Build Number
+//      Revision
+//
+// You can specify all the values or you can default the Build and Revision Numbers
+// by using the '*' as shown below:
+// [assembly: AssemblyVersion("1.0.*")]
+[assembly: AssemblyVersion("1.0.0.0")]
+[assembly: AssemblyFileVersion("1.0.0.0")]
```
QilongTang, line 36: Hmm Let's not do it this way, there are other view extension example in the repo setup so that the binary version is always same with DynamoCore, we would generally avoid defining a version in each project.

src/PythonMigrationViewExtension/PythonMigrationViewExtension.csproj
```diff
@@ -0,0 +1,106 @@
...
+    <Reference Include="System.Xaml" />
+    <Reference Include="System.Xml.Linq" />
+    <Reference Include="System.Data.DataSetExtensions" />
+    <Reference Include="Microsoft.CSharp" />
+    <Reference Include="System.Data" />
+    <Reference Include="System.Net.Http" />
+    <Reference Include="System.Xml" />
+    <Reference Include="WindowsBase" />
+  </ItemGroup>
+  <ItemGroup>
```
QilongTang, line 49: Try linking to the assemblySharedInfo.cs under root.
```
<Compile Include="..\AssemblySharedInfoGenerator\AssemblySharedInfo.cs">
      <Link>Properties\AssemblySharedInfo.cs</Link>
    </Compile>
```

src/Libraries/PythonNodeModelsWpf/Controls/EngineLabel.xaml
SHKnudsen: one of them should be removed as we have now removed the `Python Engine:` text. I added an extra row to give the node some extra white space

src/Libraries/PythonNodeModelsWpf/PythonNode.cs
SHKnudsen: @QilongTang Do we only want the `Engine Label` to be displayed in debug mode?

src/Libraries/PythonNodeModelsWpf/Controls/EngineLabel.xaml
QilongTang: @SHKnudsen Thanks!

src/Libraries/PythonNodeModelsWpf/PythonNode.cs
QilongTang: @SHKnudsen The original scope is that the labelas well as combo box on editor, we make them display only under the debug mode. After some thoughts, I think the lable maybe fine but can we at least put the switcher under debug mode, you may need to add a converter for visibility of it. Other than that, this PR looks good.

src/Libraries/PythonNodeModelsWpf/PythonNode.cs
```diff
@@ -62,6 +63,10 @@ public void CustomizeView(PythonNode nodeModel, NodeView nodeView)
             nodeView.MouseDown += View_MouseDown;
             nodeModel.DeletionStarted += NodeModel_DeletionStarted;
             nodeModel.Disposed += NodeModel_Disposed;
+
+            nodeView.PresentationGrid.Visibility = Visibility.Visible;
+            nodeView.PresentationGrid.DataContext = this.pythonNodeModel;
```
mjkkirschner, line 68: I think usually the datacontext is the viewModel - I'm not sure it matters - but I think thats the pattern in most of the repo - does that sufficiently complicate things?

src/PythonMigrationViewExtension/IronPythonInfoDialog.xaml.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Windows.Media;
+using System.Windows.Media.Imaging;
+using System.Windows.Navigation;
+using System.Windows.Shapes;
+using Dynamo.UI;
+
+namespace Dynamo.PythonMigration
+{
+    /// <summary>
+    /// Interaction logic for UserControl1.xaml
```
mjkkirschner, line 20: comment

src/PythonMigrationViewExtension/PythonMigrationViewExtension.cs
```diff
@@ -0,0 +1,162 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows.Media;
+
+namespace Dynamo.PythonMigration
+{
+    public class PythonMigrationViewExtension : IViewExtension
```
mjkkirschner, line 19: if you make this internal does that cause problems?

src/PythonMigrationViewExtension/PythonMigrationViewExtension.cs
```diff
@@ -0,0 +1,162 @@
...
+        /// </summary>
+        public string UniqueId { get { return EXTENSION_GUID; } }
+
+        /// <summary>
+        /// Extension Name
+        /// </summary>
+        public string Name { get { return EXTENSION_NAME; } }
+
+        public void Shutdown()
+        {
```
mjkkirschner, line 44: theres a problem with the current implementation for viewExtensions - they don't have dispose called on them directly, but they do have Shutdown called - can you add a dispose() call here.

mjkkirschner:
failing test
```
MESSAGE:
  Expected: equivalent to < "Documentation Browser", "DynamoManipulationExtension", "LibraryUI", "NotificationsExtension", "Workspace References", "Sample View Extension", "PackageManagerViewExtension" >
  But was:  < "Documentation Browser", "DynamoManipulationExtension", "LibraryUI", "NotificationsExtension", "Sample View Extension", "PackageManagerViewExtension", "Python Migration", "Workspace References" >

+++++++++++++++++++
STACK TRACE:
   at DynamoCoreWpfTests.PackageManagerExtensionLoadingTests.PackageManagerLoadsAndAddsViewExtension() in c:\WorkspaceDynamo\test\DynamoCoreWpfTests\PackageManagerExtensionLoadingTests.cs:line 61
```

src/Libraries/PythonNodeModelsWpf/PythonNode.cs
SHKnudsen: I see, this was only so i could bind to the Engine property on the node model. I changed it so im initializing the `EngineLabel` with the Node model so i can reference that directly from the xaml.

- https://github.com/DynamoDS/Dynamo/pull/10607/files#diff-dc0bc4a072798ad7f020dc869621e9c2R68 

- https://github.com/DynamoDS/Dynamo/pull/10607/files#diff-ccc4c3c093f111d95e8c19e41880fe75R28)

src/PythonMigrationViewExtension/PythonMigrationViewExtension.cs
```diff
@@ -21,6 +20,8 @@ internal class PythonMigrationViewExtension : IViewExtension
         private NotificationMessage IronPythonNotification { get; set; }
         internal WorkspaceModel CurrentWorkspace { get; set; }
         internal GraphPythonDependencies PythonDependencies { get; set; }
+        internal Dispatcher Dispatcher { get; set; }
```
radumg, line 23: should probably make both of these `private` and use methods to delegate stuff to them if needed from outside this extension.

src/Libraries/PythonNodeModelsWpf/PythonStringNode.cs
```diff
@@ -0,0 +1,78 @@
...
+            pythonStringNodeModel = nodeModel;
+            pythonStringNodeView = nodeView;
+
+            // If it is a Debug build, display a python engine switcher
+            if (Dynamo.Configuration.DebugModes.IsEnabled("Python3DebugMode"))
+            {
+                var pythonEngineVersionMenu = new MenuItem { Header = PythonNodeModels.Properties.Resources.PythonNodeContextMenuEngineSwitcher, IsCheckable = false };
+                nodeView.MainContextMenu.Items.Add(pythonEngineVersionMenu);
+                pythonEngine2Item.Click += delegate { UpdateToPython2Engine(); };
+                pythonEngine3Item.Click += delegate { UpdateToPython3Engine(); };
```
QilongTang, line 37: @SHKnudsen You may want to implement something similar to https://github.com/DynamoDS/Dynamo/blob/master/src/Libraries/PythonNodeModelsWpf/PythonNode.cs#L64 to unsubscribe these

src/PythonMigrationViewExtension/PythonMigrationViewExtension.csproj
```diff
@@ -0,0 +1,141 @@
...
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{10AF430D-0D3A-49CE-A63D-848912959745}</ProjectGuid>
+    <OutputType>Library</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>Dynamo.PythonMigration</RootNamespace>
+    <AssemblyName>PythonMigrationViewExtension</AssemblyName>
+    <TargetFrameworkVersion>v4.7</TargetFrameworkVersion>
```
QilongTang, line 13: Oops.. Sorry we just updated all projects to .Net48, can you grab master into your fork and update this project as well?

QilongTang:
@SHKnudsen I am merging this in and checking the builds. I think the missing two files should be there on master branch so not sure what's wrong with this PR.. Will tackle updating the .Net Framework on my side

PR #10621, opened 2020-05-01
Validate ASM installations before loading

mmisol:
### Purpose

The check for the specific assemblies tbb.dll and tbbmalloc.dll is
generalized to a full file list validation of detected ASM locations.
This way, Dynamo is guarded against any incomplete/unusual ASM binary
folders that other applications might include.

The lists of files for each version were taken from LibG. They cannot
be reused from LibG without involving major changes in the preloader,
so the lists should be kept in sync as new major release of ASM occur.

### Declarations

### Reviewers

@DynamoDS/dynamo


mmisol:
Creating this as a WIP to get early feedback. Will add unit tests before removing it.

mjkkirschner:
@mmisol I like it - but I want to make sure you saw the comment from @aparajit-pratap suggesting a more strict registry search.

mmisol:
I saw it. I think this approach is safer because we explicitly declare the files we need. Reducing the  registry search results to filter out known bad ASM installations gets trickier to update, as we don't know for sure all released products containing ASM. On the other hand, updating the list of ASM binaries for each release of ASM should be easy to update.

aparajit-pratap:
@mmisol what are the csv files for?

mmisol:
They are used in the tests. Wanted to provide an input as similar to `Directory.EnumerateFiles` as possible.

aparajit-pratap:
I thought about this some more and have more questions. I think when we need to support an Inventor integration tomorrow, we may need to do more work in addition to this in order to support loading ASM from Inventor. **If, say, a user has Inventor installed on their machine (and no other product), will ASM still be loaded correctly?** I think even in that case, Dynamo will try and load ASM from the Revit folder that comes with the Inventor install (as shown by Martin) and fail. Plus there will be no fallback option as Revit or any other product supporting ASM will not be installed on the user's system.

mmisol:
Link to self-serve job: https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/568/

mmisol:
@aparajit-pratap Trying to answer your question.
This PR does not go in the direction of "adding support for Inventor's installation of ASM". I don't know if that's planned to be honest, but might be something to discuss with the team.

QilongTang:
@mmisol Adding Inventor is not planned until our team got PM confirmation, so let's not include it in the current task scope.

aparajit-pratap:
@mmisol @QilongTang my question was not intended at Inventor specifically. I just took it as an example and the reason I took the example of Inventor is it is on the cards for an integration. Yes, this task was not about loading ASM from a specific product but we should keep in mind such integrations and the future possibilities of supporting them in our architecture/solution before we realize we need to revisit this time and again.

mjkkirschner:
@aparajit-pratap - but if we need to add product X - we just add X to the list of products to load from - then it's install location will be present in the list of ASM installs, and we'll find it later.

QilongTang:
@aparajit-pratap hmm Are you proposing a config based list instead of hard coded strings? If that is the case or you have some other ideas around this, let's discuss. I had thought about it before but that approach did not seem safe for me but we can revisit this discussion.

mmisol:
@aparajit-pratap When I said this PR does not go in the direction of "adding support for another product's installation of ASM" I didn't mean it goes against that direction either. If you think about what this PR is doing carefully, this is just a safeguard. Of course, products will need to provide a valid ASM installation for us to pick it up, but that's another story.

PR #10622, opened 2020-05-01
SQ null reference bug fix

zeusongit:
### Purpose

Addressing the issue described here: https://jira.autodesk.com/browse/DYN-2606
SonarQube null reference bug fix

### Declarations

### Reviewers
@DynamoDS/dynamo

### DIFF
--- a/src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
+++ b/src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
@@ -133,8 +133,8 @@ namespace Dynamo.DocumentationBrowser
         {
             if (e == null)
                 NavigateToNoContentPage();
-
-            this.IsRemoteResource = e.IsRemoteResource;
+            else
+                this.IsRemoteResource = e.IsRemoteResource;
 
             // Ignore requests to remote resources
             if (!this.IsRemoteResource)


src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
```diff
@@ -133,8 +133,8 @@ public void HandleOpenDocumentationLinkEvent(OpenDocumentationLinkEventArgs e)
         {
             if (e == null)
                 NavigateToNoContentPage();
-
-            this.IsRemoteResource = e.IsRemoteResource;
+            else
+                this.IsRemoteResource = e.IsRemoteResource;
```
mmisol, line 137: I'm not sure if `e` can actually be null here, but if it was, we probably should just `NavigateToNoContentPage();` and return afterwards. We already navigated at this point, and calling `HandleLocalResource` probably does not work with a null `e`.

zeusongit: Yup, `e` would not be null as we will `NavigateToNoContentPage();` before reaching that line. But it is shown as a bug in the sonarqube dashboard as a bug in the new code.
This is just one way to fix that, I guess sq does not identify non-return statements as navigation/redirection.

mmisol: Oh, I see. The how about doing:
```
if (e == null)
{
    NavigateToNoContentPage();
    return;
}
```
I think that way the code resembles the actual behaviour a little more.

PR #10624, opened 2020-05-01
[WIP][FEEDBACK] Reduce test time by substantially reducing number of serialization tests.

mjkkirschner:
### Purpose

https://jira.autodesk.com/browse/DYN-2666

While investigating the test job timeouts, I found that the serialization tests were a major portion of the testing time, and quite variable, as they require disk access 6 times each.
`(open file, save file, open file, save file, save DS file, save results file)`

In this PR I have simply limited the total number of files returned in the test generation function `FindWorkspaces` each returned workspace acts as a test case, but this function is accessed from multiple test classes for a total of approximately 5000 tests.

`SerializationTests`
`WPFJsonSerializationTests`
`SerializationNonGUIDtests`
`SerializationCultureTests`

After this change, the number becomes 1000~ tests total.

**The time savings is quite large.
The complete install job completes in 1hour 30 minutes compared to 2:30 - 3:30 average.
The parallel job running on each PR completes in 11 minutes when run on the perf machine - compared to 30 minutes.**

I am not sure how a reduction in the number of tests like this will effect coverage - the job is currently in progress. I will add that info when I have it. I am hopeful many of these tests are covering the same functionality... serialization.

Here are a few thoughts:

These serialization tests were used during the transition from xml to json files - that transition has completed, it's *definitely* still valuable to run the json tests over a subset of files (especially .xml files) in the repo - but it does not make sense to run these tests over the ever increasing collection of .json based files we already have! It will continue to grow and IMO does not provide much as a test besides wasting time. 

These tests also used to produce output files that could be used to compare with previous executions - this is quite useful for other implementations of DesignScript, but has been turned off for some time.

If we need this functionality for an ever increasing set of files we should run it as a separate job which produces this suite of executions and not tie to our tests which should be fast!

here are some other proposals to balance testing coverage vs speed.
1. disregard all files inside `FindWorkspaces` which are not xml files.
2. create a handpicked list of files instead of finding the first `n` files in the repo.
3. disable the SerializationTests - but keep the JsonSerialization tests - these duplicate much of the testing in the SerializationTests - as they are a super set.

Screen Shot 2020-05-01 at 5 04 53 PM
Screen Shot 2020-05-01 at 3 34 42 PM

### Declarations

### DIFF
--- a/src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
+++ b/src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
 using Dynamo.Configuration;
@@ -1076,9 +1077,9 @@ namespace Dynamo.Graph.Workspaces
             Guid deterministicGuid;
             if (!Guid.TryParse(obj.Value<string>(), out deterministicGuid))
             {
-                Console.WriteLine("the id was not a guid, converting to a guid");
+                Debug.WriteLine("the id was not a guid, converting to a guid");
                 deterministicGuid = GuidUtility.Create(GuidUtility.UrlNamespace, obj.Value<string>());
-                Console.WriteLine(obj + " becomes " + deterministicGuid);
+                Debug.WriteLine(obj + " becomes " + deterministicGuid);
             }
             return deterministicGuid;
         }
@@ -1191,7 +1192,7 @@ namespace Dynamo.Graph.Workspaces
             if (!Guid.TryParse(reference, out id))
             {
                 // If this is not a guid, it won't be in the resolver.
-                Console.WriteLine("not a guid");
+                Debug.WriteLine("not a guid");
                 return null;
             }
             object model;
@@ -1213,7 +1214,7 @@ namespace Dynamo.Graph.Workspaces
             if (!Guid.TryParse(reference, out id))
             {
                 // If this is not a guid, it won't be in the resolver.
-                Console.WriteLine("not a guid");
+                Debug.WriteLine("not a guid");
                 return null;
             }
             object model;
--- a/test/DynamoCoreWpfTests/SerializationTests.cs
+++ b/test/DynamoCoreWpfTests/SerializationTests.cs
@@ -27,6 +27,8 @@ namespace DynamoCoreWpfTests
 {
     internal class SerializationTests : DynamoViewModelUnitTest
     {
+        private const int MAXNUM_SERIALIZATIONTESTS_TOEXECUTE = 300;
+
         [Test]
         [Category("UnitTests")]
         public void TestBasicAttributes()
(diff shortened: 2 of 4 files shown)


mmisol:
Great find @mjkkirschner ! Thinking about possible ways to deal with this, we could also move these tests to another job that runs on a fixed schedule, maybe daily.

mjkkirschner:
@smangarole I'm not sure - but from looking at the latest sq  coverage activity at 7:38pm it appears for all these tests not run - we only lost .3% coverage... that sounds like a good tradeoff for the speedup - and I bet we can make it back by targeting those lines specifically or with specific graphs more efficiently. 

I'm not 100%  though on if the job you ran made it into the report or where I can go to see that independently?  Is that possible?
https://sq.autodesk.com/project/activity?graph=coverage&id=DynamoCA&selected_date=2020-05-01T23%3A38%3A21%2B0000

gregmarr:
I think we could definitely do this as a daily scheduled job.

aparajit-pratap:
> These tests also used to produce output files that could be used to compare with previous executions - this is quite useful for other implementations of DesignScript, but has been turned off for some time.

@mjkkirschner what do you mean by this? Are the serialization tests comparing results from XML and JSON files? If so, what is tested with new JSON files that do not have XML versions?

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
```diff
@@ -1076,9 +1077,9 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
             Guid deterministicGuid;
             if (!Guid.TryParse(obj.Value<string>(), out deterministicGuid))
             {
-                Console.WriteLine("the id was not a guid, converting to a guid");
+                Debug.WriteLine("the id was not a guid, converting to a guid");
```
QilongTang, line 1080: @mjkkirschner Can you remind me the consequence of this change, if I remember correctly this will reduce the console log size we see in Jenkins correct?

test/DynamoCoreWpfTests/SerializationTests.cs
```diff
@@ -1006,7 +1009,7 @@ public object[] FindWorkspaces()
             var di = new DirectoryInfo(TestDirectory);
             var fis = new string[] { "*.dyn", "*.dyf" }
             .SelectMany(i => di.GetFiles(i, SearchOption.AllDirectories));
-            return fis.Select(fi => fi.FullName).ToArray();
+            return fis.Select(fi => fi.FullName).Take(MAXNUM_SERIALIZATIONTESTS_TOEXECUTE).ToArray();
```
QilongTang, line 1012: 👍 I am supportive

gregmarr:
@aparajit-pratap I'll point you to internal resources on this question.

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
mjkkirschner: yes, we should not get logs filled with random guids - these were useful when we were creating this functionality, but not so much anymore.

QilongTang:
LGTM

PR #10631, opened 2020-05-05
Tests for CPython3 Engine as well as for IronPython.

reddyashish:
### Purpose

This PR is in reference to the task: https://jira.autodesk.com/browse/DYN-2190
This adds more tests to the DynamoPythonTests solution and executes the current tests with both the IronPython and CPython3 Engine. 

There is one failing test related to the Integer marshaling. 
**Update on the test:**
Integers of type int64 and lower are now being marshaled but BigIntegers are failing as PythonNet doesn't have a class like the PyInt or PyLong. This will need some more work and will file a task for it. 

### Declarations

### Reviewers

@DynamoDS/dynamo


test/Libraries/DynamoPythonTests/PythonEditTests.cs
```diff
@@ -226,7 +284,17 @@ public void ReturnPythonDictionary_AsDynamoDictionary()
 
             var guid = "490a8d54d0fa4782ae18c81f6eef8306";
 
-            AssertPreviewValue(guid, new Dictionary<string, int> { { "abc", 123 }, { "def", 345 } });
+            var nodeModel = ViewModel.Model.CurrentWorkspace.NodeFromWorkspace(guid);
+            var pynode = nodeModel as PythonNode;
+
+            foreach (var pythonEngine in pythonEngineVersionsList)
+            {
+                UpdateEnginePropertyForPythonNode(pynode, pythonEngine);
```
aparajit-pratap, line 292: Is the workspace in an unsaved state at this stage due to changes to engine property?

test/Libraries/DynamoPythonTests/PythonEditTests.cs
```diff
@@ -237,9 +305,76 @@ public void InputDynamoDictionary_AsPythonDictionary()
...
+            var nodeModel = ViewModel.Model.CurrentWorkspace.NodeFromWorkspace(pythonGUID);
+            var pynode = nodeModel as PythonNode;
+
+            foreach (var pythonEngine in pythonEngineVersionsList)
+            {
+                UpdateEnginePropertyForPythonNode(pynode, pythonEngine);
+
+                ViewModel.HomeSpace.Run();
+
+                var nodeValue = GetPreviewValue(pythonGUID);
```
aparajit-pratap, line 369: What node is this that shows the engine version in its preview?

reddyashish: This is just a python node that has OUT=sys.version statement executed in python.

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -25,6 +27,8 @@ public enum PythonEngineVersion
 
     public abstract class PythonNodeBase : VariableInputNode
     {
+         public static IEnumerable<PythonEngineVersion> pythonEngineVersionsList = Enum.GetValues(typeof(PythonEngineVersion)).Cast<PythonEngineVersion>();
```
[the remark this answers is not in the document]
reddyashish: Changed it.

test/Libraries/DynamoPythonTests/PythonEditTests.cs
reddyashish: The workspace doesn't have any changes as we are just modifying the Engine property on the node.

src/Libraries/PythonNodeModelsWpf/PythonNodeModelsWpf.csproj
```diff
@@ -100,6 +100,7 @@
 	<ProjectReference Include="..\DSIronPython\DSIronPython.csproj">
       <Project>{9eef4f42-6b3b-4358-9a8a-c2701539a822}</Project>
       <Name>DSIronPython</Name>
+      <Private>False</Private>
```
reddyashish, line 103: This will not copy the DSIronPython.dll into the nodes folder as we have that binary in the debug folder. Before this, it was being copied into 2 locations.

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -33,6 +37,7 @@ public abstract class PythonNodeBase : VariableInputNode
         protected PythonNodeBase(IEnumerable<PortModel> inPorts, IEnumerable<PortModel> outPorts) : base(inPorts, outPorts)
         {
             ArgumentLacing = LacingStrategy.Disabled;
+            pythonEngineVersionsList = Enum.GetValues(typeof(PythonEngineVersion)).Cast<PythonEngineVersion>();
```
mmisol, line 40: Please notice we are re-initializing a static field each time a Python node is created.

Also, adding a field to our code just to be used in the tests is not something we should do if we can avoid it. In this case it seems easy to avoid, we just need to a helper method in our test class.

test/Libraries/DynamoPythonTests/PythonEditTests.cs
```diff
@@ -42,6 +68,37 @@ public void PythonScriptEdit_WorkspaceChangesReflected()
...
+
+            var fileContents = File.ReadAllText(path);
+            if (string.IsNullOrWhiteSpace(fileContents))
+                return;
+
+            JObject dynObj = JObject.Parse(fileContents);
+            var pythonTokens = dynObj["Nodes"].Where(t => t.Value<string>("NodeType") == "PythonScriptNode").Select(t => t);
+            Assert.IsNotNull(pythonTokens);
+            Assert.IsTrue(pythonTokens.Any(t => t.Value<string>("Engine") == PythonEngineVersion.IronPython2.ToString()));
+        }
```
mmisol, line 100: How about if you changed the value to `CPython3` before saving? This would serve to check that the property can serialize both values properly.

test/Libraries/DynamoPythonTests/PythonEditTests.cs
```diff
@@ -251,28 +385,42 @@ public void ReturnIronPythonDictionary_AsDynamoDictionary()
...
 
-            ViewModel.ExecuteCommand(command);
+                ViewModel.HomeSpace.Run();
+
+                AssertPreviewValue(guid, new Dictionary<string, int> { { "abc", 123 }, { "def", 10 } });
+            }
         }
 
         [Test]
+        // This is failing for CPython3 Engine.
```
mmisol, line 402: Should we mark this as failing and track as a follow-up?

src/Libraries/PythonNodeModels/PythonNode.cs
reddyashish: Initially, I was using a static list in the tests to get the values. But the tests were failing because it couldn't find the PythonNodeModels.dll (the enum is defined in here) at runtime. May be the static field was the problem. Getting the list of values inside a function worked. So changed it back. Thanks @mmisol.

test/Libraries/DynamoPythonTests/PythonEditTests.cs
reddyashish: Marked it as failure.

test/Libraries/DynamoPythonTests/PythonEditTests.cs
mjkkirschner: I think that means it does have unsaved changes- the engine property is a serializable value of the node right?

test/Libraries/DynamoPythonTests/PythonEditTests.cs
reddyashish: This test was added by Aaron to check serialization and deserialization of the Engine property on global level. There is another test PythonNodeEnginePropertyChangeTest that is added to assert the python node values for both the engines.

test/Libraries/DynamoPythonTests/PythonEditTests.cs
reddyashish: But this flag ViewModel.Model.CurrentWorkspace.HasUnsavedChanges is false.

mjkkirschner: that sounds like a bug!?

test/Libraries/DynamoPythonTests/PythonEditTests.cs
mmisol: I saw that test but it's not quite the same thing. My point regarding this suggestion is that it could catch a bug on serialization where the `Engine` of a python node would always be set to `IronPython2`. Not saying that's a bug we have or that it is likely to ever appear, but I just thought it would make the test more interesting to change the engine before serializing.

test/Libraries/DynamoPythonTests/PythonEditTests.cs
reddyashish: The python node object is also marked as unmodified. Should it be marked as modified if the workspace is expected to have any changes?

reddyashish: I am look more into this.

mjkkirschner: it makes sense to me, that if the engine is changed the node should be marked as modified as this should re-execute the graph, right?

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -49,7 +51,8 @@ public PythonEngineVersion Engine
                 if (engine != value)
                 {
                     engine = value;
-                    RaisePropertyChanged("Engine");
+                    RaisePropertyChanged(nameof(Engine));
+                    OnNodeModified();
```
reddyashish, line 55: @mjkkirschner @aparajit-pratap this will mark the node as modified when the engine property is modified. Also, the workspace will have unsaved changes.

test/Libraries/DynamoPythonTests/PythonEditTests.cs
reddyashish: Added the comment above to address this.

test/Libraries/DynamoPythonTests/PythonEditTests.cs
```diff
@@ -195,10 +266,10 @@ public void VerifyPythonLoadsFromCore()
             // reference to specific testing nodes in test graph
             string[] testingNodeGUIDS = new string[]
             {
-                "845d532fdf874d939f2ed66509413ea6",
-                "cb037a9debd54ce79a4007b6ea11de25",
-                "a9bb1b12fbbd4aa19299f0d30c9f99b2",
-                "b6bd3049034f488a9bed0373f05fd021"
+                "845d532f-df87-4d93-9f2e-d66509413ea6",
```
reddyashish, line 269: @QilongTang This test was not asserting the truth values as it doesn't match the GUID pattern. Had to insert the '-' character.

test/Libraries/DynamoPythonTests/PythonEditTests.cs
```diff
@@ -251,28 +400,75 @@ public void ReturnIronPythonDictionary_AsDynamoDictionary()
...
+                 
+                ViewModel.HomeSpace.Run();
+
+                AssertPreviewValue(guid,
+                         new[] { "System.Int64", "System.Double", "System.Int64", "System.Int64", "System.Numerics.BigInteger" });
+            }
+        }
+
+        [Test]
+        public void TestWorkspaceWithMultiplePythonEngines()
```
reddyashish, line 444: This test will verify that workspace has unsaved changes when the PythonEngine property is modified on a node. Also checks that python nodes with different Python Engines in a single workspace works.

test/Libraries/DynamoPythonTests/PythonEditTests.cs
QilongTang: @reddyashish Thanks for fixing! Make sense

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -82,14 +85,16 @@ protected override string GetInputTooltip(int index)
             var vals = additionalBindings.Select(x => x.Item2).ToList();
             vals.Add(AstFactory.BuildExprList(inputAstNodes));
 
-            Func<string, IList, IList, object> pythonEvaluatorMethod;
```
QilongTang, line 85: This will create conflict with the DSPython3 branch, just FYI

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -153,6 +153,10 @@ public static DataMarshaler OutputMarshaler
                                 }
                                 return dict;
                             }
+                            else if(PyLong.IsLongType(pyObj))
+                            {
+                                return PyLong.AsLong(pyObj).ToInt64();
+                            }
```
aparajit-pratap, line 159: @reddyashish is this the fix for the int to double param method resolution issue?

mjkkirschner: curious as well as I thought that was filed as a seperate bug.

mjkkirschner: also ... isn't a `Long` the same as `Int64`?

reddyashish: Not for double. This is the case for int64 bit integers. Before this only int32 were working. 
I will be filing the issue for BigInteger marshaling.

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -43,39 +43,55 @@ public static class CPythonEvaluator
...
-                PythonEngine.Initialize();
-                using (Py.GIL())
+
+                if (!PythonEngine.IsInitialized)
                 {
-                    using (PyScope scope = Py.CreateScope())
-                    {
-                        int amt = Math.Min(bindingNames.Count, bindingValues.Count);
+                    PythonEngine.Initialize();
+                    PythonEngine.BeginAllowThreads();
```
mjkkirschner, line 50: @reddyashish - you will need to tell me what is happening here 😉 - did one of your new tests introduce a deadlock? 
@Dewb FYI

mjkkirschner: Just saw your comment, in what situations (which test jobs) are multiple test fixtures creating multiple python engines in the same process?

mmisol: I think the explanation is here https://github.com/pythonnet/pythonnet/wiki/Threading

> The Python interpreter only actually runs single-threaded. You can create additional threads, but only one thread at a time has the Global Interpreter Lock (GIL), and only that thread can run. 

So that would mean only one thread should call PythonNet at a time. Not sure what test was failing though.

reddyashish:
Team, the Nunit tests were running into a deadlock situation for DynamoPythonTests.dll. This was happening specifically when the function CPythonEvalutor.EvaluatePythonScript was called from 2 different test fixtures in a single test run. They were passing separately and also when all the tests are in a single test fixture. 
To avoid this, we have to call PythonEngine.BeginAllowThreads() and use thread locks during python evaluation. We also will not be initializing the PythonEngine if it is already initialized from the previous tests.

src/Libraries/DSCPython/CPythonEvaluator.cs
reddyashish: Sorry guys, was waiting for the self-serve. Tested it locally. I wrote the explanation before.

reddyashish: It isn't happening with a single test but when the PythonEngine is initialized from 2 different test fixtures.

reddyashish:
self-serve link: https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/603/

reddyashish:
Merging this.

PR #10633, opened 2020-05-06
Fix non-array item search for dot operation

mmisol:
### Purpose

When an array is passed to the dot operation, it needs to get an item
to determine the actual class being processed. This was done in
ArrayUtils.GetFirstNonArrayStackValue, but the function only checked
the first item of the array and its descendants. This caused the dot
operation to failed when passed an array that contained an empty array
as its first item.

In order to fix the problem, the function was changed to check for
non-array items in the entire array. The function should stop as soon
as the first non-array item is found.

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/Engine/ProtoCore/Utils/ArrayUtils.cs
+++ b/src/Engine/ProtoCore/Utils/ArrayUtils.cs
@@ -302,25 +302,49 @@ namespace ProtoCore.Utils
                 return false;
             }
 
+            var svFound = GetFirstNonArrayStackValueRecursive(svArray, runtimeCore);
+            if (svFound.HasValue)
+            {
+                sv = svFound.Value.ShallowClone();
+                return true;
+            }
+
+            return false;
+        }
+
+        /// <summary>
+        /// Recursively searches a stack value that is not an array in the given array.
+        /// </summary>
+        /// <param name="svArray">Stack value representing an array</param>
+        /// <param name="runtimeCore">Runtime core</param>
+        /// <returns>The first stack value found that is not an array. Null if none is found.</returns>
+        private static StackValue? GetFirstNonArrayStackValueRecursive(StackValue svArray, RuntimeCore runtimeCore)
+        {
+            RuntimeMemory rmem = runtimeCore.RuntimeMemory;
             var array = rmem.Heap.ToHeapObject<DSArray>(svArray);
             if (!array.Values.Any())
             {
-                return false;
+                return null;
             }
 
-            while (array.GetValueFromIndex(0, runtimeCore).IsArray)
+            foreach (var svItem in array.Values)
             {
-                array = rmem.Heap.ToHeapObject<DSArray>(array.GetValueFromIndex(0, runtimeCore));
-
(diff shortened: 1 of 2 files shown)


mjkkirschner:
@mmisol can you design a pathologically bad test case for this change?

mjkkirschner:
I think we'll need to be careful here regarding performance - I remember that @reddyashish and @aparajit-pratap implemented a few solutions that worked but had surprising performance issues with actual user data.

mmisol:
@mjkkirschner Do you have any suggestions? Just wanted to note that the function is not slower than what it was in scenarios where it actually worked.

mjkkirschner:
@reddyashish might remember better than me, or @Amoursol may have a better sense for a general limit to nesting in common user workflows.

What happens if the array only contains an empty array 1000 levels deep as the last item? Will it iterate through the entire array?

mmisol:
> What happens if the array only contains an empty array 1000 levels deep as the last item? Will it iterate through the entire array?

It would. Should we not use recursion to contemplate that level of nesting?

reddyashish:
I am not aware about the general limit on nesting in user workflows but it makes sense for this function to be recursive. We would need the input type for the zeroth nesting level and that would set the replication criteria. 

Lets run the performance tests with these changes and see if that shows any difference?

mmisol:
@mjkkirschner Would you say these changes are good to go? Let me know if not. Thanks!

QilongTang:
@mmisol Can you talk about the average scale of problem size we are addressing here? I do agree this code is performance sensitive but I also would like to get a sense of how many entries we are iterating usually.

mmisol:
@QilongTang I think I can give you a theoretical idea of the order of this algorithm. Being `n` the number of nodes and `p` the number of levels of the input, you could say best case scenario is `o(p)` and worst case scenario is `o(n)`.

To give you an average I would have to base on an actual dataset, which I don't think we have. However, worst case scenarios are triggered by multi-level and extremely sparse inputs, which my intuition tells me might be uncommon. Therefore I estimate the average would continue being `o(p)`

mjkkirschner:
let's merge it and pay attention to the performance tests.

mmisol:
@mjkkirschner Can you please approve?

Amoursol:
@mjkkirschner @mmisol Did you still need clarity on what kind of nesting users will do with DesignScript?

mmisol:
@Amoursol In the end we settled for running the performance tests after merging this.

I guess the question Mike raised was about how deep was the deepest nested `List` you have seen in practice in Dynamo. For instance, have you ever seen a `List` of `List` of `List` ... of `List` with ... representing a repetition of around 1000 levels? I hope not :)

Amoursol:
The deepest List I've seen in practise is around 4-5 levels - I've seen some random things beyond that on the on the Forum but we probably don't need to worry about it. In reality managing lists that deep is pretty difficult for users. 

Typically it's List / List of Lists / List of List of Lists.

PR #10638, opened 2020-05-11
Add ADP opt-in ui under DebugMode

pinzart:
### Purpose

Added UI for ADP (replaces the Instrumentation consent UI)
Still fully under a debug mode.
Exposed the ADP opt-in/opt-out API in Preferences as a non-serializable property,
Google and ADP are handled separately.

### Declarations

### Reviewers

### FYIs


pinzart:
ADP opt-in/opt-out values are common to all ADSK products (not per individual product). This means that Dynamo will not have exclusive control of ADP opt-in/opt-out. 
This complicates how we can expose ADP optin/optout at API level. GoogleAnalytics is fully under Dynamo's control...and is exposed in Preferences (also serialized)...This might not make sense for ADP.
Maybe we can wrap both Google and ADP opt-in/opt-out under a Common Dynamo Analytics Toggle. This would ensure that ADP is under Dynamo full control. So something like this:
[X] Enable Analytics Reporting 
     [X] Enable Google Analytics
     [X] Enable ADP Analytics

@QilongTang @mjkkirschner thoughts?

mjkkirschner:
This needs to be review carefully, I see you undid some recent changes by @QilongTang which stopped instrumentation and analytics from being started/loaded at all.

Was that needed?

src/DynamoCore/Logging/DynamoAnalyticsClient.cs
```diff
@@ -38,11 +37,8 @@ public void Start(DynamoModel model)
             if(service.GetTrackerFactory(GATrackerFactory.Name) == null)
                 service.Register(new GATrackerFactory(ANALYTICS_PROPERTY));
 
-            if (true == Configuration.DebugModes.IsEnabled("ADPAnalyticsTracker"))
```
QilongTang, line 41: Any reason we are bringing this out of debug modes on master?

pinzart: Yes. With this PR I plan to control the optin/optout values from the UI.
We can register the ADP tracker here.even when the debugmode is not enabled. It is a step forward to fully enabling ADP.
If the adp debugmode is off, no analytics will pass through because of the new filtering mechanism (line 191).

src/DynamoCore/Logging/DynamoAnalyticsClient.cs
```diff
@@ -198,11 +209,18 @@ public DynamoAnalyticsClient(DynamoModel dynamoModel)
         /// </summary>
         public void Start()
         {
-            if (preferences!= null && preferences.IsAnalyticsReportingApproved)
+            if (!AreAnyAnalyticsEnabled()) return;
+
+            //If not ReportingAnalytics, then set the idle time as infinite so idle state is not recorded.
+            Service.StartUp(product, new UserInfo(Session.UserId), preferences.IsAnalyticsReportingApproved ? TimeSpan.FromMinutes(30) : TimeSpan.MaxValue);
```
QilongTang, line 215: `TimeSpan.MaxValue` here is equivalent to not recording? Did we verify that?

src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -66,6 +66,15 @@ public class PreferenceSettings : NotificationObject, IPreferences, IRenderPreci
         /// Indicates whether analytics reporting is approved or not.
         /// </summary>
         public bool IsAnalyticsReportingApproved { get; set; }
+
+        /// <summary>
+        /// Indicates whether ADP analytics reporting is approved or not.
+        /// </summary>
+        [XmlIgnore]
```
QilongTang, line 73: Oh thanks, I totally forgot the preference setting is in xml format!

pinzart:
@mjkkirschner Regarding
`I see you undid some recent changes by @QilongTang which stopped instrumentation and analytics from being started/loaded at all if disabled. These were put in to investigate http errors in specific proxy environments.`
THat check did not make a lot of sense to me...unless it was as you said...for some edge case ...
I added the check back (with adp in the condition)

src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -66,6 +66,15 @@ public class PreferenceSettings : NotificationObject, IPreferences, IRenderPreci
...
+        public bool IsADPAnalyticsReportingApproved { 
+            get { return Logging.Analytics.ReportingADPAnalytics; }
+            set { Logging.Analytics.ReportingADPAnalytics = value; } 
+        }
```
pinzart, line 77: Exposed the ADP option in Preferences.
Since the optin value is not specific to Dynamo...(but a global option, common for all adks products) ..I added XmlIgnore so that we do not serialize the value.

src/DynamoCore/Logging/DynamoAnalyticsClient.cs
```diff
@@ -174,11 +180,16 @@ public DynamoAnalyticsClient(DynamoModel dynamoModel)
...
+            adpAnalyticsUI = new ADPAnalyticsUI();
+
             if (Session == null) Session = new DynamoAnalyticsSession();
 
             //Setup Analytics service, StabilityCookie, Heartbeat and UsageLog.
             Session.Start(dynamoModel);
 
+            Service.Instance.AddTrackerFactoryFilter(GATrackerFactory.Name, () => true == ReportingAnalytics);
+            Service.Instance.AddTrackerFactoryFilter(ADPTrackerFactory.Name, () => true == ReportingADPAnalytics);
+
```
pinzart, line 192: These filters will ensure that if ADP is **off** then no events will be passed to ADP. Same for google, if GoogleAnalytics are turned off...then no events will be processed by the GoogleTracker.

src/DynamoCore/Logging/DynamoAnalyticsClient.cs
```diff
@@ -198,12 +209,21 @@ public DynamoAnalyticsClient(DynamoModel dynamoModel)
         /// </summary>
         public void Start()
         {
-            if (preferences!= null && preferences.IsAnalyticsReportingApproved)
+            if (preferences != null && 
+                (preferences.IsAnalyticsReportingApproved || preferences.IsADPAnalyticsReportingApproved))
             {
                 //If not ReportingAnalytics, then set the idle time as infinite so idle state is not recorded.
-                Service.StartUp(product, new UserInfo(Session.UserId), preferences.IsAnalyticsReportingApproved ? TimeSpan.FromMinutes(30) : TimeSpan.MaxValue);
```
pinzart, line 204: I removed the backup value `TimeSpan.MaxValue` since it seems it was never hit before...

pinzart:
I did not refactor the code to remove the Instrumentation option (for logging PII). That can be done in a separate PR

pinzart:
In this PR, ADP analytics are still fully under a debug mode

pinzart:
image

mjkkirschner:
@QilongTang - I'm having trouble remembering the option added by Alias team, but we should make sure that this option to completely disable the UI for analytics (and also disable all analytics) continues to work.

pinzart:
@mjkkirschner Regarding #10217
I just tested that..and it works as expected...Just hides the Agreement Menu Item and the First Run prompt window.

src/DynamoCore/Logging/DynamoAnalyticsClient.cs
QilongTang: @pinzart Yes, now I remember because we made the change that if GA is not agreed, we do not initialize the client at all. It seems this change will undo that previous change though. e.g. if the ADP is agreed on, the service will be started regardless

pinzart: @QilongTang 
`
 ADP is agreed on, the service will be started regardless
`
That is correct. My intention was to make ADP and Google independent of each other. So if a user opts into ADP (but not Google) ...then Analytics service will be started (but only ADPTracker will process events). Similar if Google is opted in but ADP is not.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -594,8 +594,9 @@ protected DynamoModel(IStartConfiguration config)
                 PreferenceSettings = settings;
                 PreferenceSettings.PropertyChanged += PreferenceSettings_PropertyChanged;
             }
-            // If user does not agree to GA terms, do not try to launch the client at all
-            if (PreferenceSettings.IsAnalyticsReportingApproved)
+            // If user does not agree to GA or ADP terms, do not try to launch the client at all
+            if (PreferenceSettings.IsAnalyticsReportingApproved ||
+                PreferenceSettings.IsADPAnalyticsReportingApproved)
```
QilongTang, line 599: This is GA instrumentation here, will be removed later and should only be tied to GA agreement. We may want to revert the change here

src/DynamoCore/Logging/DynamoAnalyticsClient.cs
QilongTang: Thanks. That mechanism implementation is exellent. There are some other aspects, let's talk offline

mjkkirschner:
@pinzart @QilongTang - I think the host name is unneeded - as confusing as it is the AppVersion contains the process name.
https://github.com/DynamoDS/Dynamo/blob/31cb44e1f1dc2d77f6a4bc2f32a5d4b5e6706279/src/DynamoCore/Models/DynamoModel.cs#L331

QilongTang:
@mjkkirschner hmm process name for DynamoForAdavanceSteel and DynamoForCivil3D is same, both `Acad`. This could be a chance for us to fix that. What do you think?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -594,8 +595,27 @@ protected DynamoModel(IStartConfiguration config)
...
+                }
+            } catch(Exception)
+            {}
+            
+            if (areAnalyticsDisabled)
+            {
+                // If user skipped analytics from assembly configuration, do not try to launch the client at all 
+                // Skip call to instrumentation logger initialization.
+            }
+            else
```
pinzart, line 618: @QilongTang @mjkkirschner 
Turns out that we need Analytics binaries to be loaded to figure out the state of ADP opt-in/opt-out.

I am proposing we use a configuration setting to disable analytics when needed.
Opinions?

src/DynamoCore/Models/DynamoModel.cs
pinzart: changed code to use configuration setting.

src/DynamoCore/Logging/DynamoAnalyticsClient.cs
```diff
@@ -213,7 +211,7 @@ public DynamoAnalyticsClient(DynamoModel dynamoModel)
                 buildId = $"{version.Major}.{version.Minor}.{version.Build}"; // BuildId has the following format major.minor.build, ex: 2.5.1
                 releaseId = $"{version.Major}.{version.Minor}.0"; // ReleaseId has the following format: major.minor.0; ex: 2.5.0
             }
-            product = new ProductInfo() { Id = "DYN", Name = hostName, VersionString = appversion, AppVersion = appversion, BuildId = buildId, ReleaseId = releaseId };
+            product = new ProductInfo() { Id = "DYN", Name = "Dynamo", VersionString = appversion, AppVersion = appversion, BuildId = buildId, ReleaseId = releaseId };
```
QilongTang: hmm Why did we revert the change here?

src/DynamoCore/Models/DynamoModel.cs
QilongTang: I wouldn't do it this way, it seems this info need to be serialized, I would rather maintaining another property under preference setting which is returning `PreferenceSettings.IsAnalyticsReportingApproved|| PreferenceSettings.IsADPAnalyticsReportingApproved`. This way it can be inherited between Dynamo versions. What do you think?

src/DynamoCore/Logging/DynamoAnalyticsClient.cs
pinzart: I can switch back to hostName. I was waiting for your go ahead.

src/DynamoCore/Logging/DynamoAnalyticsClient.cs
```diff
@@ -335,6 +379,11 @@ public void LogPiiInfo(string tag, string data)
 
         public void Dispose()
         {
+            // If the Analytics Client was initialized, shut it down.
+            // Otherwise skip this step because it would cause an exception.
+            if (Service.IsInitialized)
+                Service.ShutDown();
```
pinzart, line 385: Service.ShutDown() will clean all factories and all tracker related data

src/DynamoCore/Logging/DynamoAnalyticsClient.cs
```diff
@@ -30,20 +23,6 @@ public DynamoAnalyticsSession()
 
         public void Start(DynamoModel model)
         {
-            //Whether enabled or not, we still record the startup.
```
pinzart, line 33: Moved this tracker registration/cleanup code in the Client instance. (to have access to client's data)

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -594,8 +595,27 @@ protected DynamoModel(IStartConfiguration config)
                 PreferenceSettings = settings;
                 PreferenceSettings.PropertyChanged += PreferenceSettings_PropertyChanged;
             }
-            // If user does not agree to GA terms, do not try to launch the client at all
-            if (PreferenceSettings.IsAnalyticsReportingApproved)
+
+            bool areAnalyticsDisabled = false;
```
QilongTang, line 599: @pinzart can we rename this to `areAnalyticsDisabledFromConfig`?

PR #10639, opened 2020-05-11
Support heterogeneous arrays in repl FEP matching

mmisol:
### Purpose

The no longer failing test pictures this scenario better than words.
Before this change, we would not be able to find a function endpoint
that matched the arguments, even with replication and loose type check.

This was caused by a heuristic done at the Replicator that I have named
'array sampling'. Basically, when getting type examples to produce the
reduced parameters for replication, this optimization would sample only
the first non-empty array it found in the given parent array layer.

In the failing test, the first non-empty array led to reduced params of
types string and int for the first argument of Dictionary.ValueAtKey.
As the function was expecting a dictionary, and no cast is possible
from those types, the FEP was not resolved.

In order to solve this problem, an alternative replication parameter
production strategy that does not sample arrays was introduced. By
applying this in the failing test case, we are now also inspecting the
nested array that contains the dictionary, which leads to resolving the
FEP as compliant.

For performance reasons, this strategy is used only in the context of
the most permissive FEP matching, which is case 6, as the amount of
reduced parameters to consider for replication could increase greatly.

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/Engine/ProtoCore/Lang/CallSite.cs
+++ b/src/Engine/ProtoCore/Lang/CallSite.cs
@@ -923,10 +923,7 @@ namespace ProtoCore
             #region Case 5: Replication and replication guide with type conversion and array promotion
             {
                 //Add as a first attempt a no-replication, but allowing up-promoting
-                if (!replicationTrials.Any(rt => rt.Count == 0))
-                {
-                    replicationTrials.Insert(0, new List<ReplicationInstruction>());
-                }
+                replicationTrials.Add(new List<ReplicationInstruction>());
 
                 foreach (List<ReplicationInstruction> replicationOption in replicationTrials)
                 {
--- a/src/Engine/ProtoCore/Lang/CallSite.cs
+++ b/src/Engine/ProtoCore/Lang/CallSite.cs
@@ -923,7 +923,10 @@ namespace ProtoCore
             #region Case 5: Replication and replication guide with type conversion and array promotion
             {
                 //Add as a first attempt a no-replication, but allowing up-promoting
-                replicationTrials.Add(new List<ReplicationInstruction>());
+                if (!replicationTrials.Any(rt => rt.Count == 0))
+                {
+                    replicationTrials.Insert(0, new List<ReplicationInstruction>());
+                }
 
                 foreach (List<ReplicationInstruction> replicationOption in replicationTrials)
                 {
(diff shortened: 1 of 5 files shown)


src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -923,7 +923,10 @@ private static bool IsSimilarOptionButOfHigherRank(List<ReplicationInstruction>
             #region Case 5: Replication and replication guide with type conversion and array promotion
             {
                 //Add as a first attempt a no-replication, but allowing up-promoting
-                replicationTrials.Add(new List<ReplicationInstruction>());
+                if (!replicationTrials.Any(rt => rt.Count == 0))
+                {
+                    replicationTrials.Insert(0, new List<ReplicationInstruction>());
+                }
```
mmisol, line 929: This part is actually unrelated. I just noticed that we were adding an empty replication trial even if it was already there. Also, we were adding it to the end of the list, which contradicted the comment, so I changed that as well.

src/Engine/ProtoCore/Utils/ArrayUtils.cs
```diff
@@ -142,6 +142,44 @@ public static ClassNode GetGreatestCommonSubclassForArray(StackValue array, Runt
...
         }
 
+        /// <summary>
+        /// Similar to GetTypeExamplesForLayer but it returns all arrays. Its purpose is to support
+        /// inspecting heterogeneous arrays in replication scenarios.
+        /// </summary>
+        /// <param name="array"></param>
+        /// <param name="runtimeCore"></param>
+        /// <returns></returns>
+        internal static List<StackValue> GetTypeExamplesForLayerWithoutArraySampling(StackValue array, RuntimeCore runtimeCore)
```
aparajit-pratap, line 152: So it looks like this method is not ignoring empty arrays while the other method with array sampling, is?

mmisol: Thinking about it, I could add it here too. It would not make a big difference for this one, as we are keeping all arrays, but getting the empty ones out of the way sooner may reduce the reduce params list size (no pun intended).

reddyashish:
Looks good to me. Just curious, did you check for any regressions on self-serve?

mmisol:
@reddyashish Yes, I checked https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/504/ and there are no failures.

PR #10651, opened 2020-05-13
[DYN-2680] Write logs to std out instead of creating new files for each test.

reddyashish:
### Purpose

This PR is in reference to the task: https://jira.autodesk.com/browse/DYN-2680. 

In the current design, a log file is created everytime dynamo is started and closed. So during a test run on the VM, around 5000-10000 logs are created. Over time, this is causing performance issues on the VM. We are not using these logs anyways so it is better to avoid it. 

After this change, during tests, we write the logs only to std out instead of file or to dynamo console. 

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCore/Logging/DynamoLogger.cs
+++ b/src/DynamoCore/Logging/DynamoLogger.cs
@@ -157,7 +157,20 @@ namespace Dynamo.Logging
         /// </summary>
         /// <param name="debugSettings">Debug settings</param>
         /// <param name="logDirectory">Directory path where log file will be written</param>
-        public DynamoLogger(DebugSettings debugSettings, string logDirectory, Boolean IsTestMode = false)
+        [Obsolete("This will be removed in 3.0, please use DynamoLogger(debugSettings, logDirectory, isTestMode) instead.")]
+        public DynamoLogger(DebugSettings debugSettings, string logDirectory) : this(debugSettings, logDirectory, false)
+        {
+            
+        }
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="DynamoLogger"/> class
+        /// with specified debug settings and directory where to write logs
+        /// </summary>
+        /// <param name="debugSettings">Debug settings</param>
+        /// <param name="logDirectory">Directory path where log file will be written</param>
+        /// <param name="isTestMode">Test mode is true or false.</param>
+        public DynamoLogger(DebugSettings debugSettings, string logDirectory, Boolean isTestMode)
         {
             lock (guardMutex)
             {
@@ -169,7 +182,7 @@ namespace Dynamo.Logging
 
                 notifications = new List<NotificationMessage>();
 
-                testMode = IsTestMode;
+                testMode = isTestMode;
 
                 if (!testMode)
                 {
--- a/src/DynamoCore/Logging/DynamoLogger.cs
+++ b/src/DynamoCore/Logging/DynamoLogger.cs
@@ -5,7 +5,6 @@ using System.Text;
 using Dynamo.Configuration;
 using Dynamo.Core;
 using Dynamo.Exceptions;
-using Dynamo.Interfaces;
 
 namespace Dynamo.Logging
 {
@@ -75,6 +74,7 @@ namespace Dynamo.Logging
         private string _warning;
         private WarningLevel _warningLevel;
         private bool _isDisposed;
+        private bool testMode;
 
         private TextWriter FileWriter { get; set; }
         private StringBuilder ConsoleWriter { get; set; }
@@ -157,7 +157,7 @@ namespace Dynamo.Logging
         /// </summary>
         /// <param name="debugSettings">Debug settings</param>
         /// <param name="logDirectory">Directory path where log file will be written</param>
-        public DynamoLogger(DebugSettings debugSettings, string logDirectory)
+        public DynamoLogger(DebugSettings debugSettings, string logDirectory, Boolean IsTestMode = false)
         {
             lock (guardMutex)
             {
@@ -169,7 +169,12 @@ namespace Dynamo.Logging
 
                 notifications = new List<NotificationMessage>();
 
-                StartLogging(logDirectory);
+                testMode = IsTestMode;
+
+                if (!testMode)
+                {
+                    StartLoggingToConsoleAndFile(logDirectory);
+                }
             }
         }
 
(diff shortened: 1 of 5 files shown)


src/DynamoCore/Logging/DynamoLogger.cs
```diff
@@ -157,7 +157,7 @@ public IEnumerable<NotificationMessage> StartupNotifications
         /// </summary>
         /// <param name="debugSettings">Debug settings</param>
         /// <param name="logDirectory">Directory path where log file will be written</param>
-        public DynamoLogger(DebugSettings debugSettings, string logDirectory)
+        public DynamoLogger(DebugSettings debugSettings, string logDirectory, Boolean IsTestMode = false)
```
mmisol, line 160: Should we used chained constructors instead? Optional parameters require a client recompile, so technically this is a breaking change.

mmisol: NIT: The parameter name should start with lower case: `isTestMode`

src/DynamoCore/Logging/DynamoLogger.cs
```diff
@@ -197,6 +202,13 @@ private void Log(string message, LogLevel level, bool reportModification)
                 if (debugSettings.VerboseLogging)
                     Analytics.LogPiiInfo("LogMessage-" + level.ToString(), message);
 
+                // In test mode, write the logs only to std out. 
+                if (testMode)
+                {
+                    Console.WriteLine(string.Format("{0} : {1}", DateTime.UtcNow.ToString("u"), message));
+                    return;
+                }
```
mmisol, line 210: Why not just leverage on the existing code for the console logging by setting the `level` variable to `LogLevel.Console` if `testMode` is true?

src/DynamoCore/Logging/DynamoLogger.cs
mjkkirschner: I finally found an official version of this:
https://github.com/dotnet/runtime/blob/master/docs/coding-guidelines/breaking-change-rules.md

but It does not seem as comprehensive as:

please review it whenever making changes to public apis.
relevant section:
`Adding a parameter with a default value.`
though it requires careful reading.

reddyashish: Yes will be changing this to an overloaded constructor. I tried to look if adding a optional parameter is a good design or not for public API's but that link is really comprehensive

src/DynamoCore/Logging/DynamoLogger.cs
reddyashish: The LogLevel.Console is configured to write to both the dynamo console and to the log file. So could't use it.

mmisol: You are indeed right. The fact that it does both things having that name is confusing :)

reddyashish:
Fixed a couple of existing tests. These tests verify that the logs are written to dynamo console and to a file. When we changed the behaviour to write the logs to std out in test mode, they were failing. Fixed them accordingly.

test/DynamoCoreTests/Logging/DynamoLoggerTest.cs
```diff
@@ -45,27 +63,9 @@ public void test_dynamologger_LogEventArgs()
...
         {
-            //Arrange
-            //This has been done in the base class but we need to do it again because the pathManager is private
-            var config = CreateStartConfiguration(dynamoSettings);
-            pathManager = new PathManager(new PathManagerParams
-            {
-                CorePath = config.DynamoCorePath,
-                HostPath = config.DynamoHostPath,
-                PathResolver = config.PathResolver
-            });
```
mmisol, line 56: Just to confirm, these params were not needed, right?

test/DynamoCoreTests/Migration/WorkspaceMigrationsTests.cs
```diff
@@ -15,21 +16,31 @@ class WorkspaceMigrationsTests : DynamoModelTestBase
...
+
+                StreamWriter sw = new StreamWriter(Console.OpenStandardOutput());
+                Console.SetOut(sw);
+            }
 
-            //Assert
-            Assert.IsTrue(File.Exists(currentLoggerPath));//Check that the log was created successfully
-            Assert.IsTrue(CurrentDynamoModel.Logger.LogText.Contains("migration from 0.5.3.x to 0.6.1.x"));//Checks that the log text contains the migration info
+            //Checks that the console output text contains the migration info
+            Assert.IsTrue(consoleOutput.Contains("migration from 0.5.3.x to 0.6.1.x"));
```
mmisol, line 43: Maybe for this test we actually want to use the non-test version of the logger? I'm thinking that having a persistent log after a migration is done is desired and should be tested to exist.

test/DynamoCoreTests/Logging/DynamoLoggerTest.cs
reddyashish: yeah, these were null values.

test/DynamoCoreTests/Migration/WorkspaceMigrationsTests.cs
reddyashish: The thing is, the dynamo logger is created from the dynamo model when the dynamo model is initialized from the tests. It is a read only property and we create a instance of logger in non test mode here and add it to the model

test/DynamoCoreTests/Migration/WorkspaceMigrationsTests.cs
```diff
@@ -15,21 +16,31 @@ class WorkspaceMigrationsTests : DynamoModelTestBase
...
+                Console.SetOut(stringWriter);
+
+                //Act
+                //Create a instance of WorkspaceMigration and call the function that writes to the log
+                var workMigration = new WorkspaceMigrations(CurrentDynamoModel);
+                workMigration.Migrate_0_5_3_to_0_6_0(xmlDoc);
+
+                consoleOutput = stringWriter.ToString();
+
+                StreamWriter sw = new StreamWriter(Console.OpenStandardOutput());
```
mjkkirschner, line 38: should you call dispose on StreamWriter? I believe it implements IDisposable!

mjkkirschner: sorry, looked too quickly, didn't see this was one test, not the logger!

reddyashish: We can avoid creating the StreamWriter object. Saving the std out beforehand and setting it back.

PR #10656, opened 2020-05-14
Fix transpose bug and handle executive revert

mmisol:
### Purpose

Under circumstances related to multiple node deletion in the graph
provided, the current executive of the runtime core can present a
mismatch on the value of the fepRun field. This later caused a crash
when the VM, assuming that it was calling a function endpoint, tried
to get a procedure node at an invalid index.

The cause of the this anomaly was an incorrect error handling at
JILFunctionEndPoint.Execute, where an error in a downstream call would
cause the revert of the replaced executive never to happen.

Also, the function that caused the improperly handled error is fixed.
Now the built-in transpose function should behave correctly when
provided values that are a mix of empty arrays and scalar values.

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/Engine/ProtoCore/Lang/JILFunctionEndPoint.cs
+++ b/src/Engine/ProtoCore/Lang/JILFunctionEndPoint.cs
@@ -45,108 +45,114 @@ namespace ProtoCore.Lang
                 runtimeCore.CurrentExecutive.CurrentDSASMExec = interpreter.runtime;
             }
 
-            // Assert for the block type
-            activation.globs = runtimeCore.DSExecutable.runtimeSymbols[runtimeCore.RunningBlock].GetGlobalSize();
-
-            //
-            // Comment Jun:
-            // Storing execution states is relevant only if the current scope is a function,
-            // as this mechanism is used to keep track of maintining execution states of recursive calls
-            // This mechanism should also be ignored if the function call is non-recursive as it does not need to maintains state in that case
-            int execStateSize = procedureNode.GraphNodeList.Count;
-            stackFrame.ExecutionStateSize = execStateSize;
-            for (int n = execStateSize - 1; n >= 0; --n)
+            try
             {
-                AssociativeGraph.GraphNode gnode = procedureNode.GraphNodeList[n];
-                interpreter.Push(StackValue.BuildBoolean(gnode.isDirty));
-            }
+                // Assert for the block type
+                activation.globs = runtimeCore.DSExecutable.runtimeSymbols[runtimeCore.RunningBlock].GetGlobalSize();
 
-            // Push Params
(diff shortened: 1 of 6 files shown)


aparajit-pratap:
@mmisol can you explain the context of this fix. Which bug are you trying to fix?

mmisol:
Hi @aparajit-pratap . First of all, just wanted to make clear this is not yet ready for review. I just created the PR to get the tests run against this branch.

This is related to internal ticket 2674. I'll be happy to share details in standup.

src/Engine/ProtoCore/Lang/JILFunctionEndPoint.cs
```diff
@@ -45,108 +45,114 @@ public override StackValue Execute(ProtoCore.Runtime.Context c, List<StackValue>
...
-            //
-            // Comment Jun:
-            // Storing execution states is relevant only if the current scope is a function,
-            // as this mechanism is used to keep track of maintining execution states of recursive calls
-            // This mechanism should also be ignored if the function call is non-recursive as it does not need to maintains state in that case
-            int execStateSize = procedureNode.GraphNodeList.Count;
-            stackFrame.ExecutionStateSize = execStateSize;
-            for (int n = execStateSize - 1; n >= 0; --n)
+            try
             {
```
mjkkirschner, line 49: ahh, this PR is so much easier to review with white space changes turned off 👍

src/Engine/ProtoCore/Lang/JILFunctionEndPoint.cs
```diff
@@ -45,108 +45,114 @@ public override StackValue Execute(ProtoCore.Runtime.Context c, List<StackValue>
...
-            {
-                svRet = interpreter.Run(runtimeCore.RunningBlock, activation.pc, Language.NotSpecified, runtimeCore.Breakpoints);
-                runtimeCore.RunningBlock = origRunningBlock;
-            }
-
-            if (runtimeCore.CurrentExecutive != null)
+            finally
             {
-                runtimeCore.CurrentExecutive.CurrentDSASMExec = oldDSASMExec;
+                if (runtimeCore.CurrentExecutive != null)
```
mjkkirschner, line 151: okay, I think I get it -
when transpose thew an exception (or rather any function would throw an exception - I'm curious if transpose was special in some way? or the way it failed was special?)

then this switch back to the executive which I guess represents the original calling scope is never set back and chaos ensues.

Does this sound about right?

mmisol: You got it @mjkkirschner ! I guess this would have happened with any exception during the lapse the executive is replaced. Now with try/finally the executive should be switched back in any case.

PR #10671, opened 2020-05-18
[SPIKE] View extension to load Python nodes optionally

aparajit-pratap:
### Purpose

### Declarations

### Reviewers

@DynamoDS/dynamo 

### FYIs

@Dewb


src/DynamoApplications/PathResolvers.cs
```diff
@@ -29,7 +29,7 @@ public SandboxPathResolver(string preloaderLocation)
                 "DesignScriptBuiltin.dll",
                 "DSCoreNodes.dll",
                 "DSOffice.dll",
-                "DSIronPython.dll",
+                //"DSIronPython.dll",
```
aparajit-pratap, line 32: This assembly is now loaded explicitly as part of the python view extension.

mjkkirschner:
@aparajit-pratap for the package loading portion of this - I assume one workaround is to add a dummy NodeModel type to the WPF binary, and mark it as a nodeBinary - then those types might be present when Dynamo searches for matching customizations later?

I know thats not ideal, just curious if it would work or theres a more complex reason than the types simply not being loaded that package loading doesn't support customizations in separate assemblies at the moment.

aparajit-pratap:
> @aparajit-pratap for the package loading portion of this - I assume one workaround is to add a dummy NodeModel type to the WPF binary, and mark it as a nodeBinary - then those types might be present when Dynamo searches for matching customizations later?
> 
> I know thats not ideal, just curious if it would work or theres a more complex reason than the types simply not being loaded at customization time - that package loading doesn't support customizations in separate assemblies at the moment.

@mjkkirschner yes, I think the workaround of adding a nodemodel node to the WPF binary and marking it as a node library will work. The reason is quite simply, not complex and it lies here: https://github.com/DynamoDS/Dynamo/blob/d148cfe655e96baf92dd6d85ae45eafe5c60f098/src/DynamoCore/Models/DynamoModel.cs#L1294. This condition checks if there are no NodeModel subtypes in the assembly and then imports/loads them as a ZT library. In the case of the separate WPF binary, this check succeeds (i.e. if the WPF binary doesn't contain a nodemodel derived type), and it tries to import it as a ZT node, because of which not only do its customizations not get added, but worse still, some of the types not intended for import get imported into the VM.

QilongTang:
Merging to the active branch as baseline and start with converting view extension to extension

aparajit-pratap:
@QilongTang curious how you changed the base branch on the same PR. Please enlighten me :)

QilongTang:
@aparajit-pratap When you click Edit on the top right corner, you should be able to see dropdown of the merge base, you can set other branches other than master.

aparajit-pratap:
> @aparajit-pratap When you click Edit on the top right corner, you should be able to see dropdown of the merge base, you can set other branches other than master. But you need to do that for an open PR to be able to see the dropdown.

Ah, okay. I see that now. Thanks.

PR #10672, opened 2020-05-18
Reduce memory leak in tests

mjkkirschner:
### Purpose

Before these changes running just DynamoCoreWPFTests.dll consumed 6GB of ram on my machine.
After these changes memory use is approximately 1GB and forcing a garbage collection brings it down to 300~ megs.

There are likely a bunch of extraneous changes here and also - these changes are just WIP - they need to be cleaned up.

But here are some notes on what I found:

* NodeViews leak, completely, every single customized node view was not being GC'd - these hold all sorts of things and some of them even have references to DynamoView/DynamoModel so then EVERYTHING leaks as a consequence.
* DynamoView.Dispose was never called - as a consequence:

- [x] ViewExtensions never had their Dispose method called, only their shutdown method ... I removed the dispose code because some extensions already called this from shutdown and theres some caught exceptions being thrown - this is a TODO  

* many retention paths were holding onto the underlying canvases in Helix.

* WorkspaceView and DynamoView left a bunch of handlers dangling.

- [ ] Static references under some circumstances do not get GC'd as they live in a global object the CLR builds, these might not get cleaned up during tests so I've set them null during cleanup. Need to check if these are required.

- [x]  I noticed some very strange behavior with readonly fields not being cleaned up - research seems split on if these can be a problem, need to check if this can be reverted.

- [x] Because of the design of `ViewLoadedParams` passed to ViewExtensions it's difficult to cleanup the handlers that are created there, I ended up storing a reference to a single params and using it for all view extensions that are loaded at Dynamo startup - I think this is okay. - though likely we could solve this more elegantly - it also means that view extensions loaded after startup, still leak, but thats an edge case.

- [x] InforBubble (error view) leaks - made some improvments.

- [x] removed Suppress GC on Defaultwatch3dViewModel - need to check if this was required - I don't see why we would want this though - we don't implement a finalizer.

* removed a bunch of anonymous event handlers which caused leaks, while these are great and make code more succinct and readable, I think we should disallow their use unless the author memory profiles them or has a convincing argument why they will not introduce a leak... usually they are fine if they do not reference any external objects, IE an event on the same object on which the handler is defined.

* `CustomNodeWorkspaceModels` leak - these were never cleaned up because we only clean the workspaces in the DynamoModel.Workspaces collection - but custom node workspaces are not in that collection unless they are actually open in a tab. When the customNodeManager is cleaned up now, these are uninitialized by the CNM which cleans them up.

- [x]  Added additional dispose steps to HelixWatch3dViewModel as it did not call the base class dispose, and it did not clear all collections - not sure if this is required, but calling the base class is.

### Declarations


src/DynamoCore/Extensions/ReadyParams.cs
```diff
@@ -10,10 +10,10 @@ namespace Dynamo.Extensions
     /// Application-level handles provided to an extension when 
     /// Dynamo has started and is ready for interaction
     /// </summary>
-    public class ReadyParams
+    public class ReadyParams : IDisposable
     {
-        private readonly DynamoModel dynamoModel;
-        private readonly StartupParams startupParams;
+        private  DynamoModel dynamoModel;
```
mjkkirschner, line 15: readonly weirdness - need to investigate.

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -67,7 +67,7 @@ public int ScaleFactorLog
         private Point transformOrigin;
         private bool showStartPage = false;
         
-        private List<DefaultWatch3DViewModel> watch3DViewModels = new List<DefaultWatch3DViewModel>();
+        private ObservableCollection<DefaultWatch3DViewModel> watch3DViewModels = new ObservableCollection<DefaultWatch3DViewModel>();
```
mjkkirschner, line 70: binding to a List<T> is a common WPF leak - should only bind to observables.

src/DynamoCoreWpf/ViewModels/Watch3D/DefaultWatch3DViewModel.cs
```diff
@@ -55,13 +55,13 @@ public Watch3DViewModelStartupParams(DynamoModel model)
     /// </summary>
     public class DefaultWatch3DViewModel : NotificationObject, IWatch3DViewModel, IDisposable
     {
-        protected readonly NodeModel watchModel;
```
mjkkirschner, line 58: readonly again.

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -624,7 +624,10 @@ public override void AddGeometryForRenderPackages(RenderPackageCache packages, b
 
         protected override void OnShutdown()
         {
+            viewModel = null;
+            EffectsManager.Dispose();
```
mjkkirschner, line 628: besides setting this null we should dispose it.

mjkkirschner: it was not clear to me if Shutdown and Dispose were both called in every test.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -73,7 +74,7 @@ public partial class DynamoView : Window, IDisposable
         // called on the view model and the process is not cancelled
         private bool isPSSCalledOnViewModelNoCancel = false;
         private readonly DispatcherTimer _workspaceResizeTimer = new DispatcherTimer { Interval = new TimeSpan(0, 0, 0, 0, 500), IsEnabled = false };
-
+        private ReadyParams sharedExtensionParams;
```
mjkkirschner, line 77: adding a shared viewLoadedParams object for all viewExtensions so we can call dispose on it later.

test/DynamoCoreWpfTests/PackageManagerUITests.cs
```diff
@@ -269,7 +269,8 @@ public void PackageManagerCrashTestOnDownloadingInvalidPackage()
             }
 
             var loader = GetPackageLoader();
-            Assert.AreEqual(loader.LocalPackages.Count(), 0);
+            var packageFound = loader.LocalPackages.Any(x => x.Name == "Autodesk Steel Connections 2020");
```
mjkkirschner, line 272: not sure whats up with this - maybe something got reverted on master.

src/Libraries/Watch3DNodeModels/Watch3D.cs
```diff
@@ -104,7 +104,7 @@ public class Watch3D : NodeModel
         // to be applied when the view model is constructed.
         internal XmlNode initialCameraData;
 
-        internal event Action<XmlNode> Deserialized;
+        public event Action<XmlNode> Deserialized;
```
aparajit-pratap, line 107: Why make public?

test/DynamoCoreWpfTests/PackageManagerUITests.cs
aparajit-pratap: I think this package is not meant to be installed as it is in a nested folder. @reddyashish would know.

reddyashish: Yes, this package shouldn't be loaded as it is an invalid one. I think if you have have any other package with similar name in the packages folder, this fails?. do we want to change the name of this package?

src/DynamoCoreWpf/Controls/IncanvasSearchControl.xaml.cs
```diff
@@ -44,6 +44,7 @@ private void InCanvasSearchControl_Unloaded(object sender, RoutedEventArgs e)
             {
                 Application.Current.Deactivated -= currentApplicationDeactivated;
             }
+            MembersListBox.ItemsSource = null;
```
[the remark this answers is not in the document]
mjkkirschner: yes, this can help remove a binding to a non Observable collection - for example this MembersListBox element is bound to the Ienumerable `FilteredResults` - bindings like this can leak.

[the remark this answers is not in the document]
mjkkirschner: please see the collection leak section:
https://blog.jetbrains.com/dotnet/2014/09/04/fighting-common-wpf-memory-leaks-with-dotmemory/

src/DynamoCoreWpf/ViewModels/Watch3D/DefaultWatch3DViewModel.cs
```diff
@@ -55,13 +55,13 @@ public Watch3DViewModelStartupParams(DynamoModel model)
...
+        protected  NodeModel watchModel;
 
-        protected readonly IDynamoModel dynamoModel;
-        protected readonly IScheduler scheduler;
-        protected readonly IPreferences preferences;
-        protected readonly ILogger logger;
-        protected readonly IEngineControllerManager engineManager;
```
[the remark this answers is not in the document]
mjkkirschner: 😉 one reason I really want to try that dotmemory unit test package.

src/Libraries/Watch3DNodeModels/Watch3D.cs
mjkkirschner: to be able to unsubscribe from it later, but we should be able to use internal, I just need to use the internalsVisible attr

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -1053,8 +1053,6 @@ public virtual void Dispose()
             var handler = Disposed;
             if (handler != null) handler();
             Disposed = null;
-
-            WorkspaceEvents.WorkspaceAdded -= computeUpstreamNodesWhenWorkspaceAdded;
```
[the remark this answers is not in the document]
mjkkirschner: this was done twice, it's also called on line `1051` above.

test/DynamoCoreWpfTests/PackageManagerUITests.cs
mjkkirschner: ahh, I see what happened, this is a fix I had sitting around for a while - instead of checking if no packages are loaded I changed this test so it checks that *this* package is not loaded. I wanted to make this change because I had other packages installed locally so the test failed.

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
```diff
@@ -648,6 +648,7 @@ public override void Dispose()
                 p.Dispose();
             }
             ErrorBubble.Dispose();
+            ErrorBubble = null;
```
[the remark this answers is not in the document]
mjkkirschner: reverted this - does not seem required in a smaller test suite - I will check again with all of DynamoCoreWPF though before marking this question resolved.

src/DynamoCoreWpf/ViewModels/Watch3D/DefaultWatch3DViewModel.cs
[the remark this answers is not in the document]
mjkkirschner: reverted these - ran all DynamoCoreWPF - seems minimal.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -2038,6 +2053,9 @@ private void Window_PreviewMouseLeftButtonUp(object sender, MouseButtonEventArgs
 
         public void Dispose()
         {
+            //TODO we should be careful with this, viewExtensions also have their shutdown method called
+            //when the dynamoWindow closes so it's possible there are some double dispose issues for ours or
+            //external extensions. Looking for feedback here!
```
mmisol, line 2058: For the extensions I have seen we usually call `Dispose` from `Shutdown`. I'm not sure there is any documentation established around that pattern though.

You might want to check in which cases, `DynamoView.Dispose` was being called directly, and maybe use another `DisposeAllButExtensions` in the context of `WindowClosed`. In any case this is not so bad as any error will be catched.

test/DynamoCoreWpfTests/CoreUITests.cs
```diff
@@ -572,6 +572,7 @@ public void PreferenceSetting()
                 var isTestMode = DynamoModel.IsTestMode;
 
                 // First time run, check if dynamo did set it back to false after running
+                //TODO this is a very fragile test, it fails under debug here as the view does not load fast enough.
```
mmisol, line 575: Please make a JIRA ticket instead.

test/DynamoCoreWpfTests/NodeViewCustomizationTests.cs
```diff
@@ -58,7 +58,8 @@ public void Watch3DHasViewer()
             }
 
             Open(@"UI\CoreUINodes.dyn");
-
+            //TODO why did this pass before?
```
mmisol, line 61: Anything pending to investigate here?

src/DynamoCore/Extensions/ExtensionCommandExecutive.cs
```diff
@@ -11,6 +11,7 @@ internal class ExtensionCommandExecutive : ICommandExecutive
         public ExtensionCommandExecutive(DynamoModel model)
         {
             dynamoModel = model;
+            //TODO cleanup.
```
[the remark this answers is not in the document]
mjkkirschner: thanks, missed this one, need to investigate.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
mjkkirschner: DynamoView.Dispose was never called!

test/DynamoCoreWpfTests/CoreUITests.cs
mjkkirschner: will do.

src/DynamoCore/Extensions/ReadyParams.cs
```diff
@@ -10,9 +10,9 @@ namespace Dynamo.Extensions
     /// Application-level handles provided to an extension when 
     /// Dynamo has started and is ready for interaction
     /// </summary>
-    public class ReadyParams
+    public class ReadyParams : IDisposable
     {
-        private readonly DynamoModel dynamoModel;
+        private  DynamoModel dynamoModel;
```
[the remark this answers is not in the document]
mjkkirschner: thanks @mmisol this is one I missed - I want to revert this change and test the memory consumption again.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
mmisol: Could integrators be using it? If not, maybe we can remove the part where we `Dispose` extensions, just to keep things the way they were before.

src/DynamoCore/Extensions/ReadyParams.cs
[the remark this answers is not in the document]
mjkkirschner: I don't know why - but for sample size of 30 tests - keeping this readonly introduces an extra 150 megs of memory which cannot be garbage collected.
I'm going to remove the readonly modifier here.

The other readonly modifiers I removed made no difference so I put them back, but I will keep this change.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
mjkkirschner: that was a possible solution, the safest thing to do for now is to leave out the ViewExtension dispose calls and properly document the relationship between shutdown and dispose.

src/DynamoCore/Extensions/ReadyParams.cs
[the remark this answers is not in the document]
mjkkirschner: gah, just ran this 3 times more and now it's fine with readonly, I will chalk it up to a fluke I guess - removing this for now. :suspect:

src/DynamoCore/Extensions/ExtensionCommandExecutive.cs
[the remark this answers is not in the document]
mjkkirschner: looks fine from my profiling, will remove the TODO.

test/DynamoCoreWpfTests/NodeViewCustomizationTests.cs
mjkkirschner: nope, this seems fine, must have temporarily broken something as I was in progress.

mjkkirschner:
https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service_net47/1/consoleText - only failing tests are the analytics ones that Tibi has fixed in his latest prs.


@aparajit-pratap  @mmisol  what do you think we should do for the `ViewExtension` disposal issue - the previous state was that `Dispose` was never called on `ViewExtensions` - now `Dispose` is called when the `DynamoView window` is closed - the calls are in addition to the `Shutdown` calls already made right beforehand.

otherwise this should be complete - believe I addressed all comments.

src/DynamoManipulation/DynamoManipulationExtension.cs
```diff
@@ -238,7 +238,7 @@ private static bool InspectInputsForNode(NodeModel node)
 
         public void Shutdown()
         {
-            
+            Dispose();   
         }
```
aparajit-pratap, line 242: If Dispose is called after Shutdown in `DynamoView.WindowClosed`, what's the need  call it here?

mjkkirschner: @aparajit-pratap - this depends on if we leave the ViewExtension disposal calls in or not - @mmisol was suggesting maybe leaving them out to keep the behavior as it was before this PR - if we do do, I can leave this.

If we keep the disposal calls after shutdown, I can remove this.

src/DynamoCoreWpf/Controls/IncanvasSearchControl.xaml.cs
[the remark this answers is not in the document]
mjkkirschner: yes, we can, I'll file a task.

mjkkirschner:
@aparajit-pratap I'll take a look at moving the disposal into the ViewExtensionsManager if you prefer that approach.

mjkkirschner:
@aparajit-pratap fixed, and also, this was a good catch it freed up even more memory in my test suite!

PR #10674, opened 2020-05-19
[DYN-2655]: Fix crash for badly formatted xml docs on ZT node

aparajit-pratap:
### Purpose

DYN-2655
Take the following example of XML documentation on a zero touch method. Strictly speaking it isn't a badly formatted XML as there is no XML exception that is thrown, however it does result in a different unhandled exception from our own code that parses the XML and converts it into node tooltips. Notice that here the `<returns>` tag following the `<params>` tag was missing that led to the exception and consequently the crash.
```
/// <summary> 
/// Create a T-Spline Surface from the string in T-Spline Mesh format. 
/// </summary> 
/// <param name="content">String representation of T-Spline Mesh file</param> 
/// <param name="inSmoothMode">Show T-Spline Surface in box or smooth visualization</param> 
/// Newly loaded T-Spline surface in list 
/// <search>tspline,import,serialize</search> 
[AllowRankReduction] 
public static TSplineSurface[] DeserializeFromTSM(string content, bool inSmoothMode = false) {...}
```

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCore/Library/XmlDocumentationExtensions.cs
+++ b/src/DynamoCore/Library/XmlDocumentationExtensions.cs
@@ -370,7 +370,7 @@ namespace Dynamo.Engine
                                 currentDocNode.Summary = reader.Value.CleanUpDocString();
                                 break;
                             case XmlTagType.Parameter: 
-                                // If <returns> tag is missing around text after <params> tag, the text is added as a new parameter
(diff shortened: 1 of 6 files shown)


QilongTang:
LGTM

PR #10685, opened 2020-05-20
AGD-1437  Can not preload a package with an apostrophe in its the folder path.

saintentropy:
### Purpose

This PR is to resolve an issue found with localized French samples for GD where the included packages would not load into Dynamo.  We found while these packages will load during runtime (ie if the user adds a folder to the "Manage Node and Package Paths" dialog) the package will not load upon a restart of Dynamo.  This PR adds a fix for the formatted string path which includes an apostrophe used to preload packages during start up.

### Declarations

### Reviewers

@QilongTang @mrahmaniasl 

### FYIs

@mjkkirschner


src/Engine/ProtoCore/Utils/FileUtils.cs
```diff
@@ -23,6 +23,9 @@ public static string GetDSFullPathName(string fileName, Options options = null)
             var trimChars = new[] {'\n','\t','\r',' '};
             fileName = fileName.Trim(trimChars);
 
+            // Fix file paths which include an apostrophe
+            fileName = fileName.Replace("\\'", "'");
```
QilongTang, line 27: If I understand it correctly this is replacing `\'` with `'`, but how would the test case reflect this fix? Maybe I missed something?

test/DynamoCoreTests/DSLibraryTest.cs
```diff
@@ -121,6 +121,32 @@ public void TestLoadDllFileSuccess()
...
+            Assert.IsTrue(LibraryLoaded);
+        }
+
+        [Test]
+        [Category("UnitTests")]
+        public void TestLoadDllFileWithApostropheInPathSuccess()
+        {
+            LibraryLoaded = false;
+
+            string libraryPath = Path.Combine(TestDirectory, "pkgs", "Pack'age", "bin", @"Package.dll");
```
saintentropy, line 143: @QilongTang This test fails with out the formatting fix

QilongTang: @saintentropy maybe I misunderstand the issue, how would `fileName.Replace("\\'", "'")` help this case?

QilongTang:
@saintentropy or @mrahmaniasl Can you also make it clear if this needs to go into Revit 2021.1 or can be included in the later point release?

test/DynamoCoreTests/DSLibraryTest.cs
saintentropy: Sorry now I get your question @QilongTang.  The path here gets modified here

https://github.com/DynamoDS/Dynamo/blob/master/src/Engine/ProtoCore/Utils/CompilerUtils.cs#L258 in `TryLoadAssemblyIntoCore`

The return value of `ToLiteral` is modified from _Pack'age_ to _Pack\\'age_ when it is parsed here https://github.com/DynamoDS/Dynamo/blob/master/src/Engine/ProtoCore/Utils/CompilerUtils.cs#L242

saintentropy: image

saintentropy: Above is one example of the of the call to `ToLiteral`

saintentropy:
@QilongTang Up to you regarding release.  This was originally filed by localization but is not a blocker for GD

QilongTang:
LGTM

Issue #10281, opened 2020-01-10
Crash report from Dynamo 2.5.0.7186


minjie1980:
# Issue Description
Please fill in the following information to help us reproduce the issue:

### What did you do?
I opened Dynamo 2.5 in Revit application, after running several Dynamo Revit sample scripts, and I would like to create a new Dynamo file, the crash just happened. Please notice there is no open dynamo files.

### What did you expect to see?
No crash when creating a new dynamo file.

### What did you see instead?
(Fill in here)

### What packages or external references (if any) were used?
No package has been downloaded.

### Stack Trace
```
(From the Dynamo crash window select 'Details' -> 'Copy' and paste here)
```
Starting a transaction from an external application running outside of API context is not allowed.

   at Autodesk.Revit.DB.Transaction.Start()
   at RevitServices.Transactions.TransactionWrapper.StartTransaction(Document document)
   at RevitServices.Transactions.TransactionManager.EnsureInTransaction(Document document)
   at RevitServices.Persistence.DocumentManager.DeleteElement(ElementUUID element)
   at ProtoFFI.CLRObjectMarshaler.core_Dispose(RuntimeCore sender)
   at ProtoCore.RuntimeCore.DisposeDelegate.Invoke(RuntimeCore sender)
   at ProtoCore.RuntimeCore.OnDispose()
   at ProtoScript.Runners.LiveRunner.Dispose(Boolean disposing)
   at Dynamo.Engine.LiveRunnerServices.Dispose()
   at Dynamo.Engine.EngineController.Dispose()
   at Dynamo.Models.DynamoModel.ResetEngineInternal()
   at Dynamo.Models.DynamoModel.ResetEngine(Boolean markNodesAsDirty)
   at Dynamo.ViewModels.DynamoViewModel.MakeNewHomeWorkspace(Object parameter)
   at MS.Internal.Commands.CommandHelpers.CriticalExecuteCommandSource(ICommandSource commandSource, Boolean userInitiated)
   at System.Windows.Controls.Primitives.ButtonBase.OnClick()
   at System.Windows.Controls.Button.OnClick()
   at System.Windows.Controls.Primitives.ButtonBase.OnMouseLeftButtonUp(MouseButtonEventArgs e)
   at System.Windows.RoutedEventArgs.InvokeHandler(Delegate handler, Object target)
   at System.Windows.RoutedEventHandlerInfo.InvokeHandler(Object target, RoutedEventArgs routedEventArgs)
[...]

PR #10687, opened 2020-05-20
[DYN-2206] Reset engine on scheduler thread upon creation of new workspace

aparajit-pratap:
## Purpose

Brings back the changes proposed in #10113, the difference being that now we are being careful to reset the engine on the scheduler thread so that in the context of D4R, the engine reset and consequently any Revit element cleanup takes place as a transaction on the Revit idle thread thus preventing the crash in #10281.

As mentioned in #10113, this also fixes https://jira.autodesk.com/browse/DYN-1021.

### Declarations

This fix has been tested with the Revit-Dynamo samples in Revit2020 and the crash is no longer reproducible. To verify I have tried doing the reset outside of the scheduler thread and have been able to reproduce the crash with the same steps.

### Reviewers
@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
@@ -23,6 +23,7 @@ using Dynamo.Graph.Workspaces;
 using Dynamo.Interfaces;
 using Dynamo.Models;
 using Dynamo.PackageManager;
+using Dynamo.Scheduler;
 using Dynamo.Selection;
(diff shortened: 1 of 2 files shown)


mjkkirschner:
@aparajit-pratap this looks good, is there anyway to test this in core?

aparajit-pratap:
> @aparajit-pratap this looks good, is there anyway to test this in core?

@mjkkirschner I've added a test to verify that function definitions are cleared on creating a new workspace.

reddyashish:
Looks good to me.

PR #10688, opened 2020-05-20
Remove adp debug mode

pinzart:
# Purpose

Changes:
* Removed ADP debug mode.
* Added HostName to the analytics product info.
* Adjusted a test to cover the new AddFactoryFilter functionality
* Removed my previously added test that was checking that Google.dll was not being loaded in case of opt out (reason - other tests were loading the Google dll if the test ran as part of the whole suite ..so not way to check validate the test)
* Obsoleted the Instrumentation Optin API and removed the Instrumentation doc (but I did not remove the Instrumentation related tests ... or the other related APIs - like LogPII methods)

### Declarations

### Reviewers

### FYIs


src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -60,10 +60,11 @@ public class PreferenceSettings : NotificationObject, IPreferences, IRenderPreci
...
         /// Indicates whether usage reporting is approved or not.
         /// </summary>
+        [Obsolete("Property will be deprecated in Dynamo 3.0")]
         public bool IsUsageReportingApproved { get; set; }
 
         /// <summary>
-        /// Indicates whether analytics reporting is approved or not.
+        /// Indicates whether Google analytics reporting is approved or not.
         /// </summary>
         public bool IsAnalyticsReportingApproved { get; set; }
```
pinzart, line 69: At the moment the `IsAnalyticsReportingApproved ` property name does not match it's purpose.
I can do the following:
1. Obsolete it and add a new property IsGoogleAnalyticsReportingApproved.
2. Leave it as it is (name will not math descriptions/purpose)
3. Leave it as it was and add a new property IsGoogleAnalyticsReportingApproved - this way we can re-purpose it later on to represent the Parent Analytics Toggle (if we think we might do that)

QilongTang: @pinzart Can we leave it as it is for now given the impact on migrating preference setting? I believe renaming this property, we will need to cover much edge cases and make sure people who use 1.3, 2.1 etc (which is a majority of our customer base) can migrate successfully.. I think we can do either 2 or 3. Since this particular PR is going to make it into 2.6.1, let's stick with 2 and once 2.6.1 is shipped, we can tackle option 3 with a new PR and only in newer Dynamo version

pinzart: Ok..sounds good to me.

src/DynamoCore/Logging/DynamoAnalyticsClient.cs
```diff
@@ -183,7 +184,7 @@ public DynamoAnalyticsClient(DynamoModel dynamoModel)
                 buildId = $"{version.Major}.{version.Minor}.{version.Build}"; // BuildId has the following format major.minor.build, ex: 2.5.1
                 releaseId = $"{version.Major}.{version.Minor}.0"; // ReleaseId has the following format: major.minor.0; ex: 2.5.0
             }
-            product = new ProductInfo() { Id = "DYN", Name = "Dynamo", VersionString = appversion, AppVersion = appversion, BuildId = buildId, ReleaseId = releaseId };
+            product = new ProductInfo() { Id = "DYN", Name = hostName, VersionString = appversion, AppVersion = appversion, BuildId = buildId, ReleaseId = releaseId };
```
QilongTang, line 187: Thanks!

src/DynamoCore/Logging/DynamoAnalyticsClient.cs
```diff
@@ -233,16 +232,8 @@ public void Start()
...
-
-                if (Configuration.DebugModes.IsEnabled("ADPAnalyticsTracker"))
-                {
-                    TrackPreferenceInternal("ReportingADPAnalytics", "", ReportingADPAnalytics ? 1 : 0);
-                }
-                else
-                {
-                    TrackPreferenceInternal("ReportingUsage", "", ReportingUsage ? 1 : 0);
-                }
+                TrackPreferenceInternal("ReportingGoogleAnalytics", "", ReportingAnalytics ? 1 : 0);
```
QilongTang, line 235: I think this will be inconsistent with what we are tracking right now.. Making it hard to calculate certain the GA events.. But the impact seems small enough

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -605,7 +605,8 @@ protected DynamoModel(IStartConfiguration config)
                 if (assemblyConfig != null)
                 {
                     var disableAnalyticsValue = assemblyConfig.AppSettings.Settings["DisableAnalytics"];
-                    bool.TryParse(disableAnalyticsValue.Value, out areAnalyticsDisabledFromConfig);
+                    if (disableAnalyticsValue != null)
+                        bool.TryParse(disableAnalyticsValue.Value, out areAnalyticsDisabledFromConfig);
```
QilongTang, line 609: Thank you for fixing this

src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -60,10 +60,11 @@ public class PreferenceSettings : NotificationObject, IPreferences, IRenderPreci
         /// <summary>
         /// Indicates whether usage reporting is approved or not.
         /// </summary>
+        [Obsolete("Property will be deprecated in Dynamo 3.0")]
         public bool IsUsageReportingApproved { get; set; }
```
QilongTang, line 64: Can you also make it always false? I think the changes should be applied here?

src/DynamoCore/Logging/DynamoAnalyticsClient.cs
pinzart: Are you suggesting to revert the event name "ReportingGoogleAnalytics" back to "ReportingUsage" ?

QilongTang: @pinzart Just want to you to be aware in case you are not. We are tracking events under preference setting change, currently only include the `ReportingAnalytics` event and `ReportingUsage`. If we change the event name here, this will not be treated as same event as before so we will not be able to compare the data from same time last year etc. So unless it is a must do, I don't think we need to touch this only for code clarity, what do you think?

pinzart: I understand. Since we are obsoleting the Instrumentation logic (ReportingUsage) then we probably do not need to track that.
So Isuggest:
1. Remove the ReportUsage event (already done in PR).
2. Keep "ReportingAnalytics " event for compatibility reasons.
3. Add ADP and Google preferences so that they can individually be tracked
So basically :
`
TrackPreferenceInternal("ReportingAnalytics", "", ReportingAnalytics ? 1 : 0); //Either adp or google
TrackPreferenceInternal("ReportingADPAnalytics", "", ReportingADPAnalytics ? 1 : 0);//Adp preference
TrackPreferenceInternal("ReportingGoogleAnalytics", "", ReportingGoogleAnalytics ? 1 : 0);//Google preference
`
Or should we just completely revert to the old way (just reportingUsage + reportingAnalytics)?

QilongTang: @pinzart By looking at all the history, given we are only tracking minimal set, I am fine to proceed with the current change and add more events tracking for preference setting later

QilongTang:
@pinzart Thank you for addressing the comments. LGTM

mjkkirschner:
@pinzart 

> Removed my previously added test that was checking that Google.dll was not being loaded in case of opt out (reason - other tests were loading the Google dll if the test ran as part of the whole suite ..so not way to check validate the test)

if you want to write a test asserting this you can start a new DynamoSandbox process and check the modules which it loads - we do this for some ASM loading tests.

PR #10689, opened 2020-05-20
Bug Fix: Points that can be directly manipulated should be in Red

zeusongit:
### Purpose

JIRA : [DYN-2703](https://jira.autodesk.com/browse/DYN-2703)

When navigating to preview node the directly manipulatable points were not turning red. This fixes that problem by calling the `UpdateColors()` method on point geometry.
Also updated image comparison test for DirectManipulator.

Bug screenshot:
redb

Now:
Screen Shot 2020-05-20 at 6 20 53 PM

### Declarations

### Reviewers

@DynamoDS/dynamo


mjkkirschner:
@zeusongit can you check the direct manipulation image comparison test to see if it's correct already, and if not correct it? Even if the test is not sensitive enough to catch this failure we should at least have the correct image for reference. IMO.

zeusongit:
> @zeusongit can you check the direct manipulation image comparison test to see if it's correct already, and if not correct it? Even if the test is not sensitive enough to catch this failure we should at least have the correct image for reference. IMO.

The image seems to be wrong, the point is selected, and therefore it is blue in color. Will update the test too.

src/VisualizationTests/HelixImageComparisonTests.cs
```diff
@@ -321,6 +321,7 @@ public void DirectManipulator()
 
             var command = new Dynamo.Models.DynamoModel.CreateNodeCommand(pointOriginNode, 0, 0, true, false);
             Model.ExecuteCommand(command);
+            View.BackgroundPreview.ViewModel.NavigationKeyIsDown = true;
```
QilongTang, line 324: Excellent!

PR #10694, opened 2020-05-21
[DYN-2681] Handle Replication for an input list that has only empty lists inside it. 

reddyashish:
### Purpose

This PR fixes the issue mentioned in the task: https://jira.autodesk.com/browse/DYN-2681

### Issue:
A nested list, that has only empty lists inside it, was failing to replicate. 
Example: [[],[]].

Screen Shot 2020-05-21 at 11 58 55 AM  Screen Shot 2020-05-21 at 12 15 57 PM

The custom node's output was null before this fix but we now we get the desired output. 

### Solution:
For this input, the computed replication trials will be
1) **Empty replication instruction list**: This is the default one and no FEP is found as expected.
2) **[Cartesian:0]** :  This will check for all the input parameter's type at array-rank level 1, so the reduced params list will have only one ArrayPointer object in it as both inputs are empty nested lists. This will find a matching FEP and replication instruction is set to this option. This is expected.  
3) **[Cartesian:0, Cartesian:0]**: Then we check for the input parameter's type at array-rank level 2. The reduced params list will be empty here as both the lists inside the main input list are empty. With the reduced params list being empty, we check for the matching FEP here: https://github.com/DynamoDS/Dynamo/blob/ac05e1cf4613ca656f895827d2dbdaacb7b25763/src/Engine/ProtoCore/Lang/FunctionGroup.cs#L84

Previously, if the reduced params list is empty, we would not check for any matching FEP and return true by default. This will overwrite the replication instruction to this new option. This should not be happening as we couldn't find a matching FEP with this option. To avoid it, we check for the reduced params list in the CanGetExactMatchStatics() beforehand and return false if it is empty. 

@aparajit-pratap I reverted the fix where you would skip adding it to the reduced params list if it is empty. As we saw in the regression case where the custom node function will just return the input list as it is,  we still want to process the empty list for the first time and skip it for following empty lists. This is already handled.

While investigating this bug, I found some more but they are not regressions. They were just present from the beginning but the output doesn't look correct. Still investigating these. 

Screen Shot 2020-05-21 at 1 42 21 PM

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/Engine/ProtoCore/Lang/FunctionGroup.cs
+++ b/src/Engine/ProtoCore/Lang/FunctionGroup.cs
@@ -81,6 +81,10 @@ namespace ProtoCore
             out HashSet<FunctionEndPoint> lookup)
         {
             lookup = new HashSet<FunctionEndPoint>();
+
+            if (reducedFormalParams.Count == 0)
+                return false;
+
             foreach (List<StackValue> formalParamSet in reducedFormalParams)
             {
                 List<FunctionEndPoint> feps = GetExactTypeMatches(context, formalParamSet, new List<ReplicationInstruction>(), stackFrame, runtimeCore);
--- a/src/Engine/ProtoCore/Utils/ArrayUtils.cs
+++ b/src/Engine/ProtoCore/Utils/ArrayUtils.cs
@@ -131,8 +131,6 @@ namespace ProtoCore.Utils
             var dsArray = runtimeCore.Heap.ToHeapObject<DSArray>(array);
             foreach (var sv in dsArray.Values)
             {
-                if(IsEmpty(sv, runtimeCore)) continue;
-
                 if (!usageFreq.ContainsKey(sv.metaData.type))
                 {
                     usageFreq.Add(sv.metaData.type, sv);
--- a/test/Engine/ProtoTest/TD/Associative/Replication.cs
+++ b/test/Engine/ProtoTest/TD/Associative/Replication.cs
@@ -4969,6 +4969,32 @@ px2 = DummyPoint2D.X(l2);
             thisTest.Verify("px2", new object[] { new object[] { 0 }, new object[] { } });
         }
 
+        [Test]
+        public void TestReplicationWithNestedEmptyLists()
+        {
+            string code =
+@"
+def foo( x:var[] )
+{
+    return x;
+}
+
+a = [[],[]];
+b = [[1],[]];
+c = [[],[1]];
+d = [null,[]];
+out1 = foo(a);
+out2 = foo(b);
+out3 = foo(c);
+out4 = foo(d);
+";
+            var mirror = thisTest.RunScriptSource(code);
+            thisTest.Verify("out1", new object[] { new object[] { }, new object[] { } });
+            thisTest.Verify("out2", new object[] { new object[] { 1 }, new object[] { } });
+            thisTest.Verify("out3", new object[] { new object[] { }, new object[] { 1 } });
+            thisTest.Verify("out4", new object[] { null, new object[] { } });
+        }
+
         // This tests the case 4 block in the computeFeps method (CallSite.cs)
         [Test]
         public void TestReplicationWithNullElementInNestedLists()


src/Engine/ProtoCore/Utils/ArrayUtils.cs
```diff
@@ -131,8 +131,6 @@ public static ClassNode GetGreatestCommonSubclassForArray(StackValue array, Runt
             var dsArray = runtimeCore.Heap.ToHeapObject<DSArray>(array);
             foreach (var sv in dsArray.Values)
             {
-                if(IsEmpty(sv, runtimeCore)) continue;
```
mjkkirschner, line 134: are you able to document what the behavior of this function should be?
`GetTypeExamplesForLayer` ? and add it as a /// comment?

Does this change - sufficiently change that description of the behavior if it had already existed?

reddyashish: Yeah I will add the comments for the behaviour of the function. I do not see any description already present. 
From my understanding: If an empty list is passed into a function that takes in an object, we can just skip the empty list completely as it won't needed. But if we are passing the empty list into a function that takes a list, then we cannot skip it because it can match a FEP with empty array-pointer as reduced params and we cannot just skip it then.  @aparajit-pratap should know more

mmisol: @reddyashish While the tests may still pass after this, I think this could have changed the behaviour of function endpoint matching for some cases. To give more context, that line was added by @aparajit-pratap in this PR #9559 , and it seems he was addressing performance improvements. I'm not sure if we still have those tests around, but it would be nice to run them to make sure we did not regress in that.

reddyashish: I think the fix that was added in the PR-9559 was mainly to address the failing regression from 2.0. This regression happened after some behavioral change to nested lists. Also, the performance results in that PR show that they were almost same before and after the fix. So i doubt if it was about performance improvement. Either way, I agree it is good to run them again. Will be doing that.

src/Engine/ProtoCore/Utils/ArrayUtils.cs
```diff
@@ -117,18 +117,25 @@ public static ClassNode GetGreatestCommonSubclassForArray(StackValue array, Runt
             return runtimeCore.DSExecutable.classTable.ClassNodes[orderedTypes.First()];
         }
 
-        public static Dictionary<int, StackValue> GetTypeExamplesForLayer(StackValue array, RuntimeCore runtimeCore)
+        /// <summary>
+        /// This method returns the distinct reduced params for all the elements inside the
```
mmisol, line 121: I think we should be clear about what `distinct` means here. In this case it is different types, so this method gives you one `example` per type from type from the first level of the array.

src/Engine/ProtoCore/Utils/ArrayUtils.cs
```diff
@@ -117,18 +117,25 @@ public static ClassNode GetGreatestCommonSubclassForArray(StackValue array, Runt
...
+        /// currentReducedParam, if it is an array. If it is not an array, it just returns the currentReducedParam.
+        /// </summary>
+        /// <param name="currentReducedParam"></param>
+        /// <param name="runtimeCore"></param>
+        /// <returns> A dictionary where the value is the current ReducedParam and the key is its metaData type</returns>
+        public static Dictionary<int, StackValue> GetTypeExamplesForLayer(StackValue currentReducedParam, RuntimeCore runtimeCore)
```
mmisol, line 127: I would avoid calling these reduced params. Even though this method is called while computing them, the method itself is independent from that.

src/Engine/ProtoCore/Lang/FunctionGroup.cs
```diff
@@ -81,6 +81,10 @@ public void CopyPublic(List<FunctionEndPoint> rhs)
             out HashSet<FunctionEndPoint> lookup)
         {
             lookup = new HashSet<FunctionEndPoint>();
+
```
aparajit-pratap, line 84: Could you add some comments to the `returns` tag as discussed?

reddyashish:
@mmisol @aparajit-pratap Addressed the comments. Please take a look and let me know if I missed anything.

reddyashish:
Merging this.

mjkkirschner:
Hey guys, do custom nodes have their own path through replication? IE why are we just now finding these other cases that @reddyashish shows here? Do they not fail with ZT nodes?

reddyashish:
I think the replication path is same for custom nodes or zero touch nodes. The difference in this case is that the target function here is taking an array pointer as input parameter. Previously the bugs found were around the target function that takes a Pointer object as input parameter. So, for the first case, we still have issues around lists with different array-ranks. These have different replication paths. It goes into the 4th case. https://github.com/DynamoDS/Dynamo/blob/master/src/Engine/ProtoCore/Lang/CallSite.cs#L898
and my guess is we would need to refactor the method here to check for other values as well. https://github.com/DynamoDS/Dynamo/blob/master/src/Engine/ProtoCore/Lang/Replication/Replicator.cs#L301. Also some other functions around this method should be changed accordingly.

PR #10696, opened 2020-05-22
Obsolete public classes referencing Helix in Core

mmisol:
### Purpose

This is done in anticipation of a refactor that will move these code
elements out of DynamoCoreWPF. This would allow to establish a boundary
between what's considered Dynamo's core and our rendering library,
which would in turn simplify upgrading Helix safely in the future.

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Watch3D/DefaultWatch3DViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Watch3D/DefaultWatch3DViewModel.cs
@@ -305,7 +305,7 @@ namespace Dynamo.Wpf.ViewModels.Watch3D
         /// Handle requests to reset colors on geometry graphics objects given an id.
         /// </summary>
         /// <param name="objId">id of the object</param>
-        [Obsolete("This will be moved to a new project in a future version of Dynamo")]
+        [Obsolete("Do not use! This will be moved to a new project in a future version of Dynamo.")]
         protected virtual void AttachedProperties_RequestResetColorsForDynamoGeometryModel(string objId)
         {
             //override in derived classes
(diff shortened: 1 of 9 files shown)


mmisol:
Build worked except for the regression we have in master https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/633

src/DynamoCoreWpf/ViewModels/Watch3D/DefaultWatch3DViewModel.cs
```diff
@@ -305,6 +305,7 @@ private void RegisterEventHandlers()
         /// Handle requests to reset colors on geometry graphics objects given an id.
         /// </summary>
         /// <param name="objId">id of the object</param>
+        [Obsolete("This will be moved to a new project in a future version of Dynamo")]
         protected virtual void AttachedProperties_RequestResetColorsForDynamoGeometryModel(string objId)
```
aparajit-pratap, line 309: Why does this need to be moved? It is empty and the overridden method in `HelixWatch3DViewModel` will be moved anyway.

mmisol: Actually, it will not be moved, but it will cease to exist. I copied that message too much :) Good catch!

mmisol: No it's ok. This will be moved. The plan is to, instead of implementing a handler at the `DefaultWatch3DViewModel` level, to provide a virtual method were extending implementations can register their own handlers. This decouples the base class from an event it doesn't handle and provides more flexibility to inheritors.

mmisol: @aparajit-pratap I'd advice you to take a look at the internal wiki documenting the future refactor as well, as the reasons for these obsoletions are better explained there.

mjkkirschner:
What do you think about adding to this message a friendly - `Do Not Use!`

mmisol:
Sure. I'll make that change

mjkkirschner:
LGTM

mjkkirschner:
@mmisol any issue with actually linking your wiki page here? (or its contents) I think it will be useful to refer back to it easily when looking at this commit.

mmisol:
@mjkkirschner That's a good question. In general I do not link any internal resources in public GitHub. I think it kind of goes against the spirit of an open source project.

Including the link in the Jira ticket would be fine, because that's internal as well. Is it ok if I do that instead?

PR #10701, opened 2020-05-25
Fix black flicker when resizing window

mmisol:
### Purpose

The problem is fixed by setting Background the same as BackgroundColor:
https://github.com/helix-toolkit/helix-toolkit/issues/891

ezgif com-video-to-gif(1)

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCoreWpf/Views/Preview/Watch3DView.xaml
+++ b/src/DynamoCoreWpf/Views/Preview/Watch3DView.xaml
@@ -95,6 +95,7 @@
             IsChangeFieldOfViewEnabled="False" 
             SpinReleaseTime="0" 
             EffectsManager="{Binding EffectsManager}"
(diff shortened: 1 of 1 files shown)


QilongTang:
@mmisol Can you move the left library extension broad for team to compare with the video in Jira?

mmisol:
@QilongTang I updated the gif to show that. Please don't mind the white buttons in the bottom part, they are related to something else I'm working on.

zeusongit:
@mmisol The fix looks good, but on a side note did you check if there is some way we can scale the workspace grid while resizing, or make it similar to the previous versions? Don't know if this is possible

mmisol:
@zeusongit I hadn't noticed that difference, but it might be something to take a look at as well as we finish up the Helix upgrade work. Let's discuss tomorrow in standup

mjkkirschner:
LGTM

PR #10705, opened 2020-05-26
Dyn 2562 coverage engine

RobertGlobant20:
# Purpose

Increase the code coverage adding test cases for the DynamoCore/Engine folder.
I created several unit tests methods using nUnit for the next classes:

- EngineController.
- LinkedListOfList
- LiveRunnerServices
- ShortestQualifiedNameReplacer.
- SyncDataManager.

### Declarations

### Reviewers

@QilongTang 

### FYIs

@alfredo-pozo 

@aparajit-pratap


RobertGlobant20:
This a screenshot about the code coverage using Resharper DotCover for the next files:
EngineController.cs
LinkedListOfList.cs
LiveRunnerServices.cs
ShortestQualifiedNameReplacer.cs
SyncDataManager.cs

I will create another pull request about the Engine folder at the end of this week or next monday.

image

RobertGlobant20:
This a screenshot after running all the test methods in DynamoCoreTests locally in my computer with VS2019, I got 2240 tests passed and the usual 67 failed.
Also the DYN-DevCI_Self_Service job in master-15 ran successfully for my branch.

image


Thanks

test/DynamoCoreTests/Engine/EngineControllerTest.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        /// <summary>
+        /// This test method will execute the void ReconcileTraceDataAndNotify() method from the EngineController class
+        /// </summary>
+        [Test]
+        [Category("UnitTests")]
+        public void EngineControllerReconcileTraceDataAndNotify()
+        {
+            //Arrange
+            CurrentDynamoModel.EngineController.AstBuilt += EngineController_AstBuilt;
```
mjkkirschner, line 110: please unsubscribe this handler

test/DynamoCoreTests/Engine/EngineControllerTest.cs
```diff
@@ -0,0 +1,157 @@
...
+
+            //Act
+            var codeBlockNodeOne = CreateCodeBlockNode();
+            CurrentDynamoModel.AddNodeToCurrentWorkspace(codeBlockNodeOne, true);
+            CurrentDynamoModel.EngineController.AstBuilt -= EngineController_AstBuilt;
```
mjkkirschner, line 115: 👍

test/DynamoCoreTests/Engine/LiveRunnerServicesTest.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            //Arrange
+            var openPath = Path.Combine(TestDirectory, @"core\DetailedPreviewMargin_Test.dyn");
+
+            RunModel(openPath);
+            //With this flag set to true we reach the section for logging.
+            CurrentDynamoModel.EngineController.VerboseLogging = true;
+
+            //We cannot set the VerboseLogging flag before the Model is created then in order to execute the UpdateGraph()
+            //we need to call the OpenFileCommand againt with the VerboseLogging already set to true
```
mjkkirschner, line 51: spelling (again)

RobertGlobant20: @mjkkirschner 
The spelling was fixed as a part or the commit: https://github.com/DynamoDS/Dynamo/pull/10705/commits/b6e00f7d2dbe2c067039bb11bcb164b3bea856c8

Please let me know if you have more comments.
Thanks

test/DynamoCoreTests/Engine/EngineControllerTest.cs
```diff
@@ -0,0 +1,157 @@
+﻿using Dynamo.Engine;
+using Dynamo.Graph.Nodes;
+using NUnit.Framework;
+using ProtoCore.AST.AssociativeAST;
+using ProtoScript.Runners;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
```
QilongTang, line 10: Please check Dynamo coding standards, I forgot to tell you about earlier that our using sorting is not the VS Studio default ones.

QilongTang: @RobertGlobant20 https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards

RobertGlobant20: @QilongTang 
Sorry about that, now that I'm using VS2019 I forgot to set up the coding standards configuration for usings.
Now it's fixed in all files as a part of the commit:  https://github.com/DynamoDS/Dynamo/pull/10705/commits/945250885e4d5e03d0f2ff09aead6509a4fa9209

Please review and let me know if you have any other comment.

test/DynamoCoreTests/Engine/EngineControllerTest.cs
```diff
@@ -0,0 +1,157 @@
...
+            graphSyncDataQueue.Enqueue(graphSyncdata);
+
+            //Act
+            //Using reflection we need to set up the Queue field (graphSyncDataQueue) because is private and all the times is empty (all the elements are deleted)
+            FieldInfo field = CurrentDynamoModel.EngineController.GetType().GetField("graphSyncDataQueue", BindingFlags.Instance | BindingFlags.NonPublic);
+            field.SetValue(CurrentDynamoModel.EngineController, graphSyncDataQueue);
+
+            //Assert
+            //The GenerateGraphSyncDataForCustomNodeTest method will raise an exception due that the graphSyncDataQueue has one element
+            Assert.Throws<InvalidOperationException>(() => CurrentDynamoModel.CustomNodeManager.CreateCustomNode("someNode", "someCategory", ""));
```
aparajit-pratap, line 99: This test looks like we are artificially creating a situation for an exception to occur, not creating a real scenario that could cause it to occur. Is it therefore really useful to have such tests just for the sake of increasing code coverage?

test/DynamoCoreTests/Engine/LinkedListOfListTest.cs
```diff
@@ -0,0 +1,45 @@
...
+            nodes.AddItem(Guid.NewGuid(), "test1");
+            nodes.AddItem(Guid.NewGuid(), "test2");
+            nodes.AddItem(Guid.NewGuid(), "test3");
+
+            //Act
+            var nodesKeys = nodes.GetKeys();
+
+            //We need to use reflection since the GetEnumerator methods are not used (no references) and are private to the LinkedListOfList class
+            MethodInfo dynMethod = nodes.GetType().GetMethod("System.Collections.IEnumerable.GetEnumerator", BindingFlags.NonPublic | BindingFlags.Instance);
+            var enumerator1 = (IEnumerator)dynMethod.Invoke(nodes, null);
```
aparajit-pratap, line 33: Why are we testing private methods? It would be good to test the public/internal methods that are making use of the private ones.

test/DynamoCoreTests/Engine/EngineControllerTest.cs
QilongTang: @aparajit-pratap Good point, any case we know a crash will be thrown for sure from an existing dyn?

test/DynamoCoreTests/Engine/LinkedListOfListTest.cs
RobertGlobant20: @aparajit-pratap 
When checking sonarqube, coverage section for the file Engine/LinkedListOfList.cs you will see that both methods (GetEnumerator) doesn't have coverage. I agree with you about using public methods that calls the private methods (in most cases i use them) but if you check in Visual Studio you will see that the references to that function won't call this method, for example this is one reference:

    IEnumerator IEnumerable.GetEnumerator()
    {
                return GetEnumerator();//This doesn't call the one in LinkedListOfList.cs the the coverage will be the same
     }

Then i didn't find any method that call the LinkedListOfList.GetEnumerator() and because is private using reflection was the only way that i found to do it (another one could be creating a child class and call the method).

Thanks

test/DynamoCoreTests/Engine/EngineControllerTest.cs
RobertGlobant20: @aparajit-pratap 
This is the code section from the GenerateGraphSyncDataForCustomNode( ) method that is reporting without coverage.

              // Any graph updates through the scheduler no longer store their 
                // GraphSyncData in 'graphSyncDataQueue' (any such entry will be 
                // withdrawn from the queue and get associated with an AsyncTask.
                // This check is to ensure that such case does not exist.
                // 
                if (graphSyncDataQueue.Count > 0)
                {
                    throw new InvalidOperationException(
                        "'graphSyncDataQueue' is not empty");
                }
By reading the comments I understood that probably is related to a deprecated code that now always use scheduler for graph updates, then I didn't find any way to execute that code, the graphSyncDataQueue.Count is always 0, then that's why I decided using reflection to populate the Queue in order to reach that section. If you know any other ways (or you know a specific .dyn file) that enable to execute that code please let me know and I can do the changes.

Thanks

test/DynamoCoreTests/Engine/SyncDataManagerTest.cs
```diff
@@ -0,0 +1,32 @@
...
+            RunModel(openPath);
+            var nodesList = CurrentDynamoModel.CurrentWorkspace.Nodes;
+
+            //Act
+            //Internally the PreviewGraphSyncData calls the SyncDataManager.Clone() method
+            CurrentDynamoModel.EngineController.PreviewGraphSyncData(nodesList, true);
+
+            //Assert
+            //Verify that at least we have 1 node
+            Assert.Greater(nodesList.Count(), 0);
```
aparajit-pratap, line 28: It doesn't look like `nodesList` is being modified inside `PreviewGraphSyncData`, so why test it here? It looks like you should test something that has changed as expected after the call to `PreviewGraphSyncData`.

test/DynamoCoreTests/Engine/LinkedListOfListTest.cs
alfredo-pozo: I think that another option could be to remove that code, if it's private and not referenced, I think is unnecessary code that we can get rid off.

Probably another task could be to clean those classes when we find a case like that @QilongTang

aparajit-pratap: I don't think that we can get rid of these methods either as they are required by the interface. In that case, I'm okay with keeping this test.

test/DynamoCoreTests/Engine/EngineControllerTest.cs
aparajit-pratap: Not that I know of. I guess we can mark this code to be removed soon; as you said it sounds like it's deprecated. If it's going to be deprecated soon, my vote would be to not add this test.

test/DynamoCoreTests/Engine/SyncDataManagerTest.cs
RobertGlobant20: @aparajit-pratap 
After the ideas that we analyzed by Slack about how to test the Clone() method I will consider this test method for the next sprint pull request, then I will proceed to remove this test case from this pull request.

test/DynamoCoreTests/Engine/EngineControllerTest.cs
RobertGlobant20: @aparajit-pratap 
Ok, I will discard this test case from this pull request.

Thanks

RobertGlobant20: @aparajit-pratap 
This test method was removed from the pull request in the commit : https://github.com/DynamoDS/Dynamo/pull/10705/commits/4d07fccfc603beb0ef3d0140a6307d3bad04df9d

Please review
cc: @QilongTang

test/DynamoCoreTests/Engine/SyncDataManagerTest.cs
RobertGlobant20: @aparajit-pratap 
This test method (the whole file) was removed from the pull request in the commit : https://github.com/DynamoDS/Dynamo/pull/10705/commits/4d07fccfc603beb0ef3d0140a6307d3bad04df9d

Please review
cc: @QilongTang

QilongTang:
Comments addressed. Merging

PR #10707, opened 2020-05-27
Make Helix render more backwards compatible

mmisol:
### Purpose

The fixed directional light coming from top is eliminated. This light
behaved completely different with the previous shader. Basically, this
was being moved with the view so there was no light coming from above,
but this was actually coming from the camera, which is what the head
light now does.

Also, some color calculations are changed in the shader so that the
final color outcome looks how it used to.

These screen capture were done after this change:

Aaron's Road Simulation_v2_2020-05-27_02-50-55

Column_Builder_Multi_2020-05-27_02-52-08

HelixTest_Exports_2020-05-27_02-50-42

simple_sphere_2020-05-27_04-25-08

Special thanks to @mjkkirschner for the help with spotting the shader differences :) 

### Declarations

### Reviewers

@DynamoDS/dynamo


mmisol:
Fixed! Thanks for pointing that out @mjkkirschner

PR #10712, opened 2020-05-28
Update new tests

mjkkirschner:
just in case, cleanup event handlers in tests.

### DIFF
--- a/test/DynamoCoreWpfTests/ViewExtensions/PythonMigrationViewExtensionTests.cs
+++ b/test/DynamoCoreWpfTests/ViewExtensions/PythonMigrationViewExtensionTests.cs
@@ -33,6 +33,8 @@ namespace DynamoCoreWpfTests
                 Preferences = new PreferenceSettings() { CustomPackageFolders = new List<string>() { this.PackagesDirectory } }
             };
         }
+        private List<string> raisedEvents = new List<string>();
+
 
         /// <summary>
         /// This test is created to check if the extension displays a dialog to the user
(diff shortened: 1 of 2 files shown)


mjkkirschner:
@mmisol yes, I noticed that as well @byorda-glb was there supposed to be some assertion in that handler?

byorda-glb:
> @mmisol yes, I noticed that as well @byorda-glb was there supposed to be some assertion in that handler?

Not in the handlers themselves, the assertions are in the tests.

test/DynamoCoreWpfTests/ViewExtensions/PythonMigrationViewExtensionTests.cs
```diff
@@ -43,10 +33,6 @@ public void WillDisplayDialogWhenOpeningGraphWithIronPythonNodes()
         {
             DebugModes.LoadDebugModesStatusFromConfig(Path.Combine(GetTestDirectory(ExecutingDirectory), "DynamoCoreWpfTests", "python3DebugMode.config"));
             DynamoModel.IsTestMode = false;
-            // Arrange
-            RaiseLoadedEvent(this.View);
-            var extensionManager = View.viewExtensionManager;
-            extensionManager.Add(viewExtension);
```
QilongTang, line 49: I guess this should be here anyway because by default Dynamo loads all viewExtension with xml file there

test/DynamoCoreWpfTests/ViewExtensions/PythonMigrationViewExtensionTests.cs
```diff
@@ -43,10 +33,6 @@ public void WillDisplayDialogWhenOpeningGraphWithIronPythonNodes()
         {
             DebugModes.LoadDebugModesStatusFromConfig(Path.Combine(GetTestDirectory(ExecutingDirectory), "DynamoCoreWpfTests", "python3DebugMode.config"));
             DynamoModel.IsTestMode = false;
-            // Arrange
-            RaiseLoadedEvent(this.View);
```
QilongTang, line 47: This one I thought needed by some reason @reddyashish mentioned long time ago.. is the assumption no longer valid somehow?

test/DynamoCoreWpfTests/ViewExtensions/PythonMigrationViewExtensionTests.cs
mjkkirschner: yes, this was attempting to reload and causing issues.

test/DynamoCoreWpfTests/ViewExtensions/PythonMigrationViewExtensionTests.cs
mjkkirschner: it's only required if we need whatever the Loaded event does to be re triggered - for this test it's not required.

PR #10727, opened 2020-06-02
update performance tool to run with .net 4.8

mjkkirschner:
### Purpose

PerformanceTests do not run on .net 4.8 -
* The assembly bind redirects were doing very funny things to run the job on .net 4.7. This now runs with .net 4.8 is installed on my machine. The tests now run locally.

* package.config reference location is corrected.

I had nuget regenerate the assembly bind redirects and this is the much smaller app.config that it generated.

I have not yet actually tested this on the CI machine, will merge this first and address any failures after.

One thing I noticed is the test log will be much longer now since the logs are redirected to this log.

### Declarations


mjkkirschner:
actually remembered I can run this:
https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoPerformance-CI/job/master/419/console

QilongTang:
@mjkkirschner Did you run both the Core Perf tests and View ones?

tools/Performance/DynamoPerformanceTests/packages.config
```diff
@@ -1,61 +1,61 @@
...
+  <package id="System.Threading.Tasks.Parallel" version="4.3.0" targetFramework="net48" />
+  <package id="System.Threading.Thread" version="4.3.0" targetFramework="net48" />
+  <package id="System.ValueTuple" version="4.5.0" targetFramework="net48" />
+  <package id="System.Xml.ReaderWriter" version="4.3.0" targetFramework="net48" />
+  <package id="System.Xml.XDocument" version="4.3.0" targetFramework="net48" />
+  <package id="System.Xml.XmlDocument" version="4.3.0" targetFramework="net48" />
+  <package id="System.Xml.XmlSerializer" version="4.3.0" targetFramework="net48" />
+  <package id="System.Xml.XPath" version="4.3.0" targetFramework="net48" />
+  <package id="System.Xml.XPath.XDocument" version="4.3.0" targetFramework="net48" />
+  <package id="System.Xml.XPath.XmlDocument" version="4.3.0" targetFramework="net48" />
```
mmisol, line 60: Do we need to have all of these `System` packages? I don't think we have anything like that in our `Dynamo.All` solution after the upgrade.

mjkkirschner:
@QilongTang I did not run the view perf tests as those are not part of the test framework on the CI. Maybe now that we have reduced our memory use significantly they will actually run without monotonically increasing in time.

tools/Performance/DynamoPerformanceTests/packages.config
mjkkirschner: I think these are just direct dependencies of the version of benchmarkdotnet that we use - they are not actually copied to the bin folder of the perf executable, it seems they are just referenced from the GAC at runtime.

mmisol: Oh I see. The package targets .NET Standard 2.0 only.

I see dependencies for all .NET framework specific classes it uses, but also their dependencies recursively. I guess this is expected.

Feels like we should probably consider multi-targeting for our own NuGet packages in the future, as a bloated packages.config is not nice.

PR #10728, opened 2020-06-02
[DYN-2754] Fix the in-canvas search break.

reddyashish:
### Purpose

The in-canvas node search was broken when we were trying to reduce the memory leaks in the tests. I have reverted the particular change that broke the search functionality. 

Running the self-serve to make sure the total test time is fine after this change.  

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCoreWpf/Controls/IncanvasSearchControl.xaml.cs
+++ b/src/DynamoCoreWpf/Controls/IncanvasSearchControl.xaml.cs
@@ -34,7 +34,7 @@ namespace Dynamo.UI.Controls
             {
                 Application.Current.Deactivated += currentApplicationDeactivated;
             }
-            Unloaded += InCanvasSearchControl_Unloaded; ;
+            Unloaded += InCanvasSearchControl_Unloaded;
 
         }
 
@@ -44,7 +44,6 @@ namespace Dynamo.UI.Controls
             {
                 Application.Current.Deactivated -= currentApplicationDeactivated;
             }
-            MembersListBox.ItemsSource = null;
         }
 
         private void currentApplicationDeactivated(object sender, EventArgs e)
--- a/src/DynamoCoreWpf/Controls/IncanvasSearchControl.xaml
+++ b/src/DynamoCoreWpf/Controls/IncanvasSearchControl.xaml
@@ -121,7 +121,7 @@
                      Background="Transparent"
                      BorderBrush="Transparent"
                      ItemContainerStyle="{StaticResource SearchMemberStyle}"
-                     ItemsSource="{Binding FilteredSearchResults, NotifyOnTargetUpdated=True}"
+                     ItemsSource="{Binding FilteredResults, NotifyOnTargetUpdated=True}"
                      TargetUpdated="OnMembersListBoxUpdated"
                      PreviewMouseWheel="OnMembersListBoxMouseWheel">
                 <ListBox.ItemTemplate>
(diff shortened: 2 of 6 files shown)


src/DynamoCoreWpf/Controls/IncanvasSearchControl.xaml.cs
```diff
@@ -34,7 +34,7 @@ public InCanvasSearchControl()
             {
                 Application.Current.Deactivated += currentApplicationDeactivated;
             }
-            Unloaded += InCanvasSearchControl_Unloaded; ;
+            Unloaded += InCanvasSearchControl_Unloaded;
```
QilongTang, line 37: 👍

QilongTang: Wow build will pass in this case?

mmisol: Yes, but I bet it creates a code smell :)

mjkkirschner:
@reddyashish so the problem here is that `MembersListBox` itemsource is bound to `FilteredResults` - which is an `Ienumerable` but because it's used as an itemsource with a binding it should be an observableCollection - unfortunately we can't change its type as it's public.

what we can do is to create a new member, obsolete the old one, and bind to that instead.

see this issue:
https://blog.jetbrains.com/dotnet/2014/09/04/fighting-common-wpf-memory-leaks-with-dotmemory/

scroll down to `CollectionBindingLeak`

reddyashish:
I have added a new observable collection and binded the list box to this object. In that case, we can obsolete the FilteredResults list in 3.0. 
I am running the self-serve to make sure there are no failures.

reddyashish:
1) I had to revert the ObservableCollection binding to the ListBox because the below event handler, related to TargetUpdated on ListBox, was not getting triggered for some reason. 
https://github.com/DynamoDS/Dynamo/blob/65b703d359401ca9646c65a96250c7085fd3d351/src/DynamoCoreWpf/Controls/IncanvasSearchControl.xaml.cs#L125
Investigated various possibilites but still the results were not getting populated on the UI side when using the ObservableCollection. 
2) Tried setting MembersListBox.ItemsSource = null in the dispose method of IncanvasSearchControl control and call this dispose method from DynamoView class when dynamo is closed. But the DynamoView doesn't have a reference to the IncanvasSearchControl and so it was not possible. 
3) With the unit test, I couldn't reproduce a failing test case for this particular workflow. This is because the SearchViewModel correctly returns the list of the nodes based on the query but the ListBox data was not being updated on the UI side after the source was set to null. So we might need a UI automation test to verify the data on the ListBox control. 

I verified that the memory isn't rising while running the DynamoCoreWPF tests after removing this null assignment. The maximum memory it reached was 1.3GB. Discussed with @mjkkirschner about this and will be monitoring the total test time on the self-serve after this PR is merged.
Let me know if you guys have any thoughts.

mmisol:
I think adding a test to detect this issue in the future might be a good idea. My reasoning is that in the future someone might want to stop leaks and make this change again.

Adding a test was not part of the original scope of the task, so maybe it can be tackled as a separate issue. What do you think @mjkkirschner ?

mjkkirschner:
yes, agree let's merge this, watch the tests and please @reddyashish file a task to figure out how to write a test that would detect this defect in the future.

PR #10731, opened 2020-06-03
Enable Python engine selection by default

mmisol:
### Purpose

This makes the Python3EngineDebugMode enabled by default, no need to
have a config file for it. Also, creates a new mode named
Python2ObsoleteMode which controls whether using IronPython2 engine
should be notified as a warning.

Also fixed a bug where notifications were shown when Python nodes were
added, no matter the state of the flags.

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCore/Configuration/DebugModes.cs
+++ b/src/DynamoCore/Configuration/DebugModes.cs
@@ -34,7 +34,7 @@ namespace Dynamo.Configuration
             public bool IsEnabled;
         }
 
-        private static void AddDebugMode(string name, string description, bool isEnabled = false)
+        internal static void AddDebugMode(string name, string description, bool isEnabled = false)
         {
             debugModes[name] = new DebugMode()
             {
@@ -43,6 +43,7 @@ namespace Dynamo.Configuration
                 IsEnabled = isEnabled
             };
         }
+
         private static void RegisterDebugModes()
         {
             // Register app wide new debug modes here.
--- a/src/DynamoCore/Configuration/DebugModes.cs
+++ b/src/DynamoCore/Configuration/DebugModes.cs
@@ -12,7 +12,6 @@ namespace Dynamo.Configuration
     internal static class DebugModes
     {
         static private readonly Dictionary<string, DebugMode> debugModes = new Dictionary<string, DebugMode>();
-        private static bool debugModesEnabled;
 
         /// <summary>
         /// Represents an instance of a debug mode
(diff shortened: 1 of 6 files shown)


src/DynamoCore/Configuration/DebugModes.cs
```diff
@@ -35,32 +34,30 @@ public class DebugMode
...
-                IsEnabled = false
+                IsEnabled = isEnabled
             };
         }
         private static void RegisterDebugModes()
         {
             // Register app wide new debug modes here.
-            AddDebugMode("Python3DebugMode", "Enable/disable Python3 Engine.");
+            AddDebugMode("Python3DebugMode", "Enable/disable Python3 Engine.", true);
+            AddDebugMode("Python2ObsoleteMode", "Enable/disable warnings regarding Python 2 obsoletion.");
```
mjkkirschner, line 50: I think this makes this very clear, good idea.

src/DynamoCore/Configuration/DebugModes.cs
```diff
@@ -35,32 +34,30 @@ public class DebugMode
...
         }
 
         internal static void LoadDebugModesStatusFromConfig(string configPath)
```
mjkkirschner, line 53: is this method only used for testing?

mjkkirschner:
https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/684/testReport/

reflection 💀

mmisol:
I missed those. Will take a look.

src/DynamoCore/Configuration/DebugModes.cs
mmisol: It is also used in normal initialization, for a file named `debug.config`. That's how Dynamo picks it up normally.

src/DynamoCore/Configuration/DebugModes.cs
```diff
@@ -35,32 +34,30 @@ public class DebugMode
...
                 Description = description,
-                IsEnabled = false
+                IsEnabled = isEnabled
             };
         }
         private static void RegisterDebugModes()
         {
             // Register app wide new debug modes here.
-            AddDebugMode("Python3DebugMode", "Enable/disable Python3 Engine.");
+            AddDebugMode("Python3DebugMode", "Enable/disable Python3 Engine.", true);
```
QilongTang, line 49: Seems after this change, even without a debug.config, debug modes can be turned on purely with the API, is that the intention here?

mmisol: Yes. I thought it would be better not to require any file for enabling the Python engine selection, and have it turned on by default.

QilongTang: - Do users' disable debug mode setting still get respected?
- This way we will have two source of truth.. Which we can discuss within the team, I understand it maybe annoying to setup debug.config locally but in general I would support keep only one source, and use as least debug modes as possible.
- I feel this also opens up the door for extensions to touch debug modes potentially

Please give some consideration.

mmisol: Hi Aaron. In this case we are only allowing to set the default to something other than `false`. Other than that I didn't make any changes.

Of course a debug mode can still be affected by a loaded XML, that's no different from before. The fact that it was `false` by default didn't mean it didn't exist. We always had "two sources of truth" if you may: a default and a config override.

I don't see how an extension could change these settings, without resorting to reflection. In any case, I only changed the visibility of one function to internal, so the door is not any more open than before.

QilongTang: hmm but previously we are not allowing to set the default to enabled without a config correct? So extensions will not be able to do that either.

QilongTang: Cool. Thanks for the discussion. I think it is safe to proceed!

PR #10733, opened 2020-06-04
More Info on Python Editor

QilongTang:
Please Note:

### Purpose
Changes:
- Fixed the space issue in the engine selection drop down. Notice this will also make the run mode dropdown and host dependency dropdown UI change a couple pix
- Updated layout of controls in Python Editor
- Added Icons and tooltips for controls in Python Editor
- Added a button like control in Python3 learning material for user to report an issue
- Added minHeight and minWidth for Python Editor as a request by @Amoursol 

Final version:
image

image

Fina workflow gif (does not include the space fix):
PythonEditorMoreInfoButton_v4

### Declarations

### Reviewers

@DynamoDS/dynamo 

### FYIs


src/Libraries/PythonNodeModels/Properties/Resources.resx
```diff
@@ -176,4 +176,7 @@
   <data name="PythonNodeContextMenuEngineVersionTwo" xml:space="preserve">
     <value>IronPython2</value>
   </data>
+  <data name="PythonScriptEditorMoreInfoButton" xml:space="preserve">
+    <value>More Information</value>
+  </data>
```
mmisol, line 181: Given this is temporary, maybe we should keep `Save Changes` as is and make this `More info` instead.

Also, I find it worrying that we need to change texts to make up for space. In other languages these texts could be longer and I hope that is not breaking thee layout. Do we need to make additional changes to provide the flexibility?

src/Libraries/PythonNodeModelsWpf/PythonNodeModelsWpf.csproj
```diff
@@ -97,6 +97,10 @@
       <Name>DynamoServices</Name>
       <Private>False</Private>
     </ProjectReference>
+    <ProjectReference Include="..\..\PythonMigrationViewExtension\PythonMigrationViewExtension.csproj">
+      <Project>{10AF430D-0D3A-49CE-A63D-848912959745}</Project>
+      <Name>PythonMigrationViewExtension</Name>
+    </ProjectReference>
```
mmisol, line 103: I agree this is not nice. We should be able to tunnel an event through the model that can reach the extension.

src/PythonMigrationViewExtension/Docs/PythonMigrationWarning.html
```diff
@@ -207,18 +207,18 @@ <h3>How to make changes to your Python nodes for Python 3 </h3>
...
-        <li>
-            <a href="http://python3porting.com/">Supporting Python 3: An in-depth guide" guide</a>
-        </li>
-    </ul>
+        <ul>
+            <li>
+                <a href="https://docs.python.org/3/howto/pyporting.html">
+                    The official "Porting Python 2 Code to Python
+                    3"
+                    guide
```
mmisol, line 215: This was already
like
this
but we
should make
it one line
I guess.
:)

src/Libraries/PythonNodeModels/Properties/Resources.resx
QilongTang: This worries me as well. I changed this because for other buttons we are using `Revert` `Run`, etc so I would like to be consistent, if we can move to use icons with tooltips, that would be even better IMO. This python editor did not get any UX love for so many years.. I would love to get our UX to give it some review :)

`More info` is fine, I did this version mainly wanted to align with the migration dialog, they are doing pretty much the samething becaue

src/Libraries/PythonNodeModelsWpf/PythonNodeModelsWpf.csproj
QilongTang: yeah, I do not like this at all but failed to find a supporting workaround. Maybe event based would be better

src/PythonMigrationViewExtension/Docs/PythonMigrationWarning.html
QilongTang: Let me check the whole file, I agree

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml
```diff
@@ -48,6 +59,15 @@
...
             <StackPanel Orientation="Horizontal" HorizontalAlignment="Right" Grid.Column="1">
+                <Button Content="{x:Static p:Resources.PythonScriptEditorMoreInfoButton}" 
+                        Style="{DynamicResource ResourceKey=STextButton}" 
+                        Name="MoreInfoButton" 
+                        Click="OnMoreInfoClicked" 
+                        HorizontalAlignment="Right" 
+                        VerticalAlignment="Bottom" 
+                        Margin="2" 
+                        Grid.Column="1"
+                        Visibility="{Binding nodeModel.Engine, Converter={StaticResource PythonEngineVersionToVisibilityConverter}}"/>
```
QilongTang, line 70: @mmisol also any suggestion on debugging this? I found my converter not triggered at all

mmisol: Oh. I think the problem is not with your button, but that the engine combo is not firing a change. I think you had to do `UpdateSourceTrigger=PropertyChanged`.

QilongTang:
@mmisol Feel free to take another look

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
```diff
@@ -167,8 +172,13 @@ private void OnRunClicked(object sender, RoutedEventArgs e)
...
         }
 
-        #endregion
 
+        #endregion
 
+        private void OnMoreInfoClicked(object sender, RoutedEventArgs e)
+        {
+            var viewEx = (WpfUtilities.FindUpVisualTree<DynamoView>(nodeView) as DynamoView).viewExtensionManager.ViewExtensions.FirstOrDefault(x => x.UniqueId == "1f8146d0-58b1-4b3c-82b7-34a3fab5ac5d");
+            (viewEx as PythonMigrationViewExtension).OpenPythonMigrationWarningDocumentation();
```
mmisol, line 181: It would be nice if we could avoid the reference to the extension, finding it by id, and then directly calling a method on it. Can we do something similar to what the migration extension is doing?

`LoadedParams.ViewModelCommandExecutive.OpenDocumentationLinkCommand(link);`

QilongTang: @mmisol Even we try to get `ViewModelCommandExecutive` the actual hyper link to open will be only accessible from view extension still. Then we still need a reference to `PythonMigrationViewExtension`. That's why I left this code unchanged for now.

I also think through the event based way, but that will require `PythonMigrationViewExtension` subscribe to every python node in new workspace plus every new python new added to the workspace. That still does not seem ideal to me.

One thing @mjkkirschner remind me is that in previous development, we have implemented something like a broadcast mechanism for a center model to broadcast action message to all extensions and only the extension meeting criteria will react to that. That design may fit our case here, but I feel it is out of the scope.

On the other hand, I still hold my original opinion that we may want to move the PythonEditor, Python3 Engine + PythonMigrationViewExtension into a single view extension and make it optional from Dynamo. That seems to be the cleanest way I can see without introducing new mechanism.

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
```diff
@@ -177,8 +177,7 @@ private void OnRunClicked(object sender, RoutedEventArgs e)
 
         private void OnMoreInfoClicked(object sender, RoutedEventArgs e)
         {
-            var viewEx = (WpfUtilities.FindUpVisualTree<DynamoView>(nodeView) as DynamoView).viewExtensionManager.ViewExtensions.FirstOrDefault(x => x.UniqueId == "1f8146d0-58b1-4b3c-82b7-34a3fab5ac5d");
-            (viewEx as PythonMigrationViewExtension).OpenPythonMigrationWarningDocumentation();
+            dynamoViewModel.OpenDocumentationLinkCommand.Execute(new OpenDocumentationLinkEventArgs(PythonMigrationViewExtension.Python3HelpLink));
```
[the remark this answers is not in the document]
QilongTang: @mjkkirschner That way it will require the extension instance or LoadedParams but here we can call API directly?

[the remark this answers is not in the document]
mjkkirschner: oh @QilongTang you're totally right! Sorry, I thought this was in an extension for some reason... 😕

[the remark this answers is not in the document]
QilongTang: @mjkkirschner No worries. That's the pain point actually :)

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
mmisol: The way the documentation browser is designed, it can load content from any loaded assembly. You can take a look at the readme here https://github.com/DynamoDS/Dynamo/tree/master/src/DocumentationBrowserViewExtension

QilongTang:
@mmisol PTAL

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
```diff
@@ -29,6 +29,7 @@ public partial class ScriptEditorWindow : ModelessChildWindow
         public PythonNode nodeModel { get; set; }
         private bool nodeWasModified = false;
         private string originalScript;
+        internal NodeView nodeView;
```
mmisol, line 32: I don't see this field being used. Can we remove it?

QilongTang: Sure removed

PR #6240, opened 2016-03-10
Fix MAGN-9568 Crash in preview control

ke-yu:
### Purpose

Fix [MAGN-9568 Crash in preview control](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-9568)

There are two issues, both related to outdated mirror data that used in preview control to query the node value from the DesignScript virtual machine.

Updating a graph will update mirror data in all executed UI nodes. Ideally, the following preview operation will get the latest mirror data to display value. But because updating graph (`UpdateGraphAsyncTask`), querying mirror data (`QueryMirrorDataAsyncTask`) and displaying preview value (wrapped in `DelegateBasedAsyncTask`) are all asynchronous operations, and their creation is not necessarily in scheduler thread, their execution order is not guaranteed and the outdated mirror data will be used and cause crash.

**The first issue** is `DelegateBasedAsyncTask` is scheduled before `QueryMirrorDataAsyncTask`.

In this case, `UpdateGraphAsyncTask` has been executed and node's value *has been updated*. Now a `DelegateBasedAsyncTask` is created and put in scheduler thread (because of previewing operation on the UI thread). It will be executed immediately because `QueryMirrorDataAsyncTask`has not been scheduled yet:

```
Scheduler task queue
--------------------

         +----------if complete, schedule query task -----+
         |                                                |
         |                                                v
+----------------------+------------------------+--------------------------+
| UpdateGraphAsyncTask | DelegateBasedAsyncTask | QueryMirrorDataAsyncTask |
+----------------------+------------------------+--------------------------+
                                ^
                                |
                                +----- Scheduled from UI thread

```
`NodeModel.CachedValue` has not been updated yet, but the value of node has been changed, so `NodeModel.CachedValue` becomes invalid. 

There are three ways to fix this case:
  1. Set `NodeModel.CachedValue` to some invalid state. In `DelegateBasedAsyncTask` we will check if mirror data is invalid or not, and display "..." in the preview control accordingly. Mirror data will finally be updated in `QueryMirrorDataAsyncTask`. But introducing more states will make the code be harder to maintain.
  2. As `MirrorData` is a snapshot of value, we could add the other kind of mirror data which always reflects current value. But `MirrorData` is widely used now, making this change will incur too many other changes. Besides, always querying value from the engine would be a potential performance issue.
[...]

PR #10736, opened 2020-06-04
Deleting assignment statements after function definition in CBN now updates output ports

aparajit-pratap:
### Purpose

1. Fixes DYN-2778 - deleting assignment statements after function definition in CBN now updates output ports
2. Cached value when set to null now raises property changed event which updates node preview (see updated gif)
3. Remove `QueryMirrorDataAsyncTask` as all usages were removed in https://github.com/DynamoDS/Dynamo/pull/6240

func_def_cbn_fixed

### Declarations

### Reviewers
@DynamoDS/dynamo


src/DynamoCore/Graph/Nodes/CodeBlockNode.cs
```diff
@@ -827,18 +827,18 @@ internal void SetErrorStatePortData(List<string> inputPortNames, List<int> outpu
...
             // cannot be cleaned up properly
             
             // Clear out all the output port models
-            OutPorts.RemoveAll((p) => { return true; });
+            OutPorts.RemoveAll((p) => true);
+            
+            var allDefs = CodeBlockUtils.GetDefinitionLineIndexMap(codeStatements);
+
+            if (allDefs.Any() == false)
+                return;
```
QilongTang, line 841: ah So always clean the out put ports and re-add them back.  Is there similar pattern we need to worry about input ports or they are already following this?

aparajit-pratap: It's already being done for input ports.

src/DynamoCore/Graph/Nodes/CodeBlockNode.cs
```diff
@@ -827,18 +827,18 @@ internal void SetErrorStatePortData(List<string> inputPortNames, List<int> outpu
...
-                return;
-
             // This extension method is used instead because 
             // observableCollection has very odd behavior when cleared - 
             // there is no way to reference the cleared items and so they 
             // cannot be cleaned up properly
             
             // Clear out all the output port models
-            OutPorts.RemoveAll((p) => { return true; });
+            OutPorts.RemoveAll((p) => true);
```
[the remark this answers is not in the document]
aparajit-pratap: This is our own custom extension we have written for `ObservableCollection`s

aparajit-pratap:
I find it strange that the node preview for the CBN containing the function definition continues to show the previous value (see gif). I'm trying to fix that as well.

src/DynamoCore/Graph/Nodes/CodeBlockNode.cs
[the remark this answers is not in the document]
mmisol: I took a look at it but I don't see it doing anything special. Calling `Clear` should be more efficient, but shouldn't make much difference, so I'll leave it up to you.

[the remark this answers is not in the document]
mjkkirschner: clear does not raise the necessary events.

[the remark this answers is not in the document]
mmisol: I guess you are right https://stackoverflow.com/questions/2480714/does-the-clear-method-on-a-collection-release-the-event-subscriptions . That sucks

aparajit-pratap:
> I find it strange that the node preview for the CBN containing the function definition continues to show the previous value (see gif). I'm trying to fix that as well.

Also fixed (see updated gif)

src/DynamoCore/Scheduler/QueryMirrorDataAsyncTask.cs
```diff
@@ -1,65 +0,0 @@
...
-    struct QueryMirrorDataParams
-    {
-        internal IScheduler Scheduler { get; set; }
-        internal EngineController EngineController { get; set; }
-        internal string VariableName { get; set; }
-    }
-
-    class QueryMirrorDataAsyncTask : AsyncTask
-    {
-        #region Class Data Members and Properties
```
mjkkirschner, line 16: so was anything constructing this task type?

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -2344,26 +2344,14 @@ protected virtual ExecutionHints GetExecutionHintsCore()
         ///
         internal void RequestValueUpdate(EngineController engine)
         {
-            // A NodeModel should have its cachedMirrorData reset when it is
-            // requested to update its value. When the QueryMirrorDataAsyncTask
-            // returns, it will update cachedMirrorData with the latest value.
-            //
-            lock (cachedValueMutex)
```
mjkkirschner, line 2351: can the mutex be removed? What else locks on it? Why remove it?

mjkkirschner: is there behavior that depended on cachedValue being set null?

test/DynamoCoreTests/CodeBlockNodeTests.cs
```diff
@@ -339,6 +339,43 @@ public void TestVarRedefinitionInFunctionDef()
...
+                                           "}");
+           Assert.IsNotNull(cbn);
+
+           Assert.AreEqual(1, cbn.CodeStatements.Count());
+
+           Assert.AreEqual(0, cbn.InPorts.Count);
+           Assert.AreEqual(0, cbn.OutPorts.Count);
+           Assert.AreEqual(0, cbn.AllConnectors.Count());
+        }
+
```
mmisol, line 378: Can we add a test for a case in which we have the same initial state but after the edit we also have the assignment `a = [ 1, 2 ];`?

src/DynamoCore/Graph/Nodes/NodeModel.cs
aparajit-pratap: `CachedValue` is still being set to `null` in certain cases, one such case being this one where code is deleted from the CBN, leaving only the function definition in place. This is now happening in line: 2354 below. Earlier only the field was being set to `null` and not the property because of which the property changed event wasn't being raised. As a result, the preview of the node didn't update to `null` upon updating the CBN. 
As `CachedValue` is being set to a non-null value outside of the lock anyway, I thought it would be safe to set it to `null` in cases where the mirror is `null`. Also the `lock` was in place specifically for `QueryMirrorDataAsyncTask` as per the comments, which has now been removed.

aparajit-pratap: I can check if the mutex is being used elsewhere and if it can be removed altogether.

aparajit-pratap: Ok, I just checked the `cachedValueMutex` is being used implicitly in the `CachedValue` getter and setter itself so it is safe to remove from here. No more to do here.

test/DynamoCoreTests/CodeBlockNodeTests.cs
aparajit-pratap: @mmisol done.

mmisol: Nice. Thanks!

test/DynamoCoreWpfTests/RecordedTests.cs
```diff
@@ -3708,7 +3708,7 @@ public void Defect_MAGN_3212()
 
                     //Check the CBN for input/output ports
                     Assert.AreNotEqual(ElementState.Error, cbn.State);
-                    Assert.AreEqual(1, cbn.OutPorts.Count);
+                    Assert.AreEqual(0, cbn.OutPorts.Count);
```
mmisol, line 3711: Was this test doing an assertion that was incorrect and is now fixed by the change here?

aparajit-pratap: Correct. Now that we remove the output ports for a function def, the count was updated to zero for this particular test.

PR #10737, opened 2020-06-04
Update license

mmisol:
### Purpose

Added Python.NET, removed unused libraries, and updated some links.

### Declarations

### Reviewers

@mjkkirschner @QilongTang

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,6 +1,8 @@
+Apache license 2.0
+
 ## Dynamo License
 
-Copyright 2013-2019 Autodesk
+Copyright 2013-2020 Autodesk
 
 Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at
 
@@ -12,15 +14,12 @@ Unless required by applicable law or agreed to in writing, software distributed 
 ## Third Party Licenses ##
 
 ### Avalon Edit
-http://www.codeproject.com/Articles/42490/Using-AvalonEdit-WPF-Text-Editor
-http://opensource.org/licenses/lgpl-3.0.html
+http://avalonedit.net/
+https://opensource.org/licenses/lgpl-2.1.php
+  (LGPL-2.1 license)
 
 ### CEFSharp
 https://github.com/cefsharp/CefSharp/blob/master/LICENSE
-
-### DirectX
-https://github.com/DynamoDS/Dynamo/tree/master/tools/install/Extra/DirectX/License Agreements/DirectX SDK EULA.txt
-https://github.com/DynamoDS/Dynamo/tree/master/tools/install/Extra/DirectX/License Agreements/directx redist.txt
 
 ### CSharpAnalytics
 https://github.com/AttackPattern/CSharpAnalytics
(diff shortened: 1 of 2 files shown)


LICENSE.txt
```diff
@@ -41,13 +44,13 @@ https://www.gnu.org/software/gettext/
 https://www.gnu.org/software/gettext/manual/html_node/GNU-LGPL.html#GNU-LGPL
 
 ### Helix Toolkit
-https://helixtoolkit.codeplex.com/
-https://helixtoolkit.codeplex.com/license 
+https://github.com/helix-toolkit/helix-toolkit
```
QilongTang, line 47: Wonderful

doc/distrib/License.rtf
```diff
@@ -1,526 +1,119 @@
...
+\par
+\cf5\b\f0 StarMath\par
+{\cf0\b0\f2{\field{\*\fldinst{HYPERLINK https://github.com/DesignEngrLab/StarMath/blob/master/LICENSE }}{\fldrslt{https://github.com/DesignEngrLab/StarMath/blob/master/LICENSE\ul0\cf0}}}}\cf0\b0\f2\fs22\par
+\f3\par
+\cf1\b\f0 Xceed Extended WPF Toolkit\f1\par
+{{\field{\*\fldinst{HYPERLINK "https://opensource.org/licenses/ms-pl.html"}}{\fldrslt{\ul\cf4\cf0\b0\f2 Microsoft Public License}}}}\cf0\b0\f3\fs22\par
+{\f2{\field{\*\fldinst{HYPERLINK https://github.com/xceedsoftware/wpftoolkit/blob/master/license.md }}{\fldrslt{https://github.com/xceedsoftware/wpftoolkit/blob/master/license.md\ul0\cf0}}}}\f3\fs22\par
+
+\pard\nowidctlpar\sb240\par
+}
```
QilongTang, line 118: For some reason seems there are way more content in the previous version? Is this IDE related? @mjkkirschner  should know more here

mjkkirschner: yeah, the same RTF view can be created with different elements of course - we should really test this in the application though as .net renders rtf kinda funny.

mmisol: I noticed that too. I initially edited the file in WordPad and opened it in Dynamo -> About. Things looked good except for a few links I added that were invisible! To fix that I had to play with the "RTF tags" until they decided to appear.

LICENSE.txt
```diff
@@ -12,16 +14,13 @@ Unless required by applicable law or agreed to in writing, software distributed
 ## Third Party Licenses ##
 
 ### Avalon Edit
-http://www.codeproject.com/Articles/42490/Using-AvalonEdit-WPF-Text-Editor
-http://opensource.org/licenses/lgpl-3.0.html
+http://avalonedit.net/
+https://opensource.org/licenses/lgpl-2.1.php
+  (LGPL-2.1 license)
```
mjkkirschner, line 19: it looks like AvalonEdit is actually MIT:
https://github.com/icsharpcode/AvalonEdit/blob/master/LICENSE

LICENSE.txt
```diff
@@ -123,25 +110,28 @@ http://msdn.microsoft.com/en-us/library/gg405489(PandP.40).aspx
 
 ### Python Standard Library
 https://www.python.org/
-https://www.python.org/download/releases/2.7/license/
```
mjkkirschner, line 126: we are still shipping the 2.7 std lib and so I think this should remain.

LICENSE.txt
mmisol: Lates version is, but the one we are using is LGPL-2.1. See https://www.nuget.org/packages/AvalonEdit/4.3.1.9430

LICENSE.txt
mmisol: Actually, the link I replaced this with includes the license of every Python version up to this day. It's an updated version of the old link, which included the license of every Python version up to 2.7.

LICENSE.txt
mjkkirschner: hmm, I wonder if it is worth it to check which version Revit ships these days and align if they have moved up to a more recent one.

mmisol: That's a good question. I guess currently we rely on them to test, but if it works we don't know if we could have upgraded. In that case we would be sticking with an old version for no reason.

Can we maybe run the D4R tests and see if causes problems? Or is it more complex than that?

mjkkirschner:
Let me just dig up their manifest for the version they use - testing d4r unfortunately won’t catch the conflict

PR #10747, opened 2020-06-08
Add version miss check

zeusongit:
[DYN-2779](https://jira.autodesk.com/browse/DYN-2779)

Added null check for failed version install due to a cache miss and display a dialog box.
Added resource string.

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCoreWpf/Properties/Resources.Designer.cs
+++ b/src/DynamoCoreWpf/Properties/Resources.Designer.cs
@@ -3142,6 +3142,15 @@ namespace Dynamo.Wpf.Properties {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to Failed to download package version. Please try again and report the package if you continue to have problems..
+        /// </summary>
+        public static string MessageFailedToDownloadPackageVersion {
+            get {
+                return ResourceManager.GetString("MessageFailedToDownloadPackageVersion", resourceCulture);
+            }
+        }
+        
+        /// <summary>
         ///   Looks up a localized string similar to No node could be found with that Id..
         /// </summary>
         public static string MessageFailedToFindNodeById {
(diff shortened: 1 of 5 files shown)


QilongTang:
@zeusongit Would you add a gif or screenshot of the change?

mmisol:
I wonder if we should actually do this. The changes done in package manager should prevent this inconsistency from happening, right? Also, this potentially affects every version of Dynamo from 2.5 backwards to whenever Package Manager began to exist probably. Even if we back port it, Dynamo is integrated into products that probably won't be updated anymore.

mjkkirschner:
my vote would be to do this, theres any number of reasons the result could be null... - can you add to the message *where* to report the package or the issue?

Also, it seems there are failing tests and a failing build?

zeusongit:
How about this?
> Failed to download package version. Please try again and report the package on the Dynamo Github repository if you continue to have problems.

zeusongit:
Looked into the build failure, seems to be an issue with loading the assembly:

> Build Failed:
> Error parsing solution file at E:\Builds\Dynamo_RC2.5.0_master\Dynamo\src\Dynamo.All.sln: Exception has been thrown by the target of an invocation.  The project file could not be loaded. Could not load file or assembly 'Microsoft.Build.Framework, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.  

And those 3 test failing seems to be unrelated to the PR but they point towards a missing signed certificate.

> DynamoUtilitiesTests.CertificateVerificationTests.CheckAssemblyForValidCertificateTest
> MESSAGE:
> System.Exception : A dll file found at C:\Jenkins\workspace\Dynamo\DynamoSelfServe\pullRequestValidation\Dynamo\test\pkgs_signed\Signed Package\bin\SignedPackage.dll did not have a signed certificate.
> 

> Dynamo.Tests.ExcelTests.ReadChangeFilename
> MESSAGE:
>   Expected: True
>   But was:  False

> Dynamo.PackageManager.Tests.PackageLoaderTests.ScanPackageDirectoryWithCheckingCertificatesEnabledWillLoadPackageWithValidCertificate
> MESSAGE:
>   Expected: not null
>   But was:  null

Certificate verification at https://github.com/DynamoDS/Dynamo/blob/88eff5749e13b7ad1891d850603147a66dcd2a74/src/DynamoUtilities/CertificateVerification.cs#L55 is returning false.

PR #10748, opened 2020-06-08
Enforce split between View and Model with Python projects to fix Mono build.

mjkkirschner:
### Purpose

This PR does the following:
1. adds a github action for mono builds on linux - this will run on each pr - yay.
https://github.com/DynamoDS/Dynamo/runs/751846262?check_suite_focus=true - it doesn't run any tests yet, but it does check that DynamoCLI.exe was produced.
2. undoes some changes that were made in bringing in DSCpython project - namely:
    * image resources were moved to DSIronPython
    * AvalonEdit was made a reference of DSIronPython (can't do this as it requires WPF)
    * `IronPythonCompleteionProvider` and `IronPythonCompletionData` were brought into DSIronPython (can't do this, as they reference AvalonEdit.

3. To achieve what I believe was the intention of this reorganization, and to get us ready to support multiple completion providers in the future I have created a new class and set of interfaces.
These are:
`SharedCompletionProvider`
`IExternalCompletionProviderCore`
`IExternalCodeCompletionData `
`ExternalCodeCompletionType ` (enum)

`IronPythonCompletionprovderCore`
`IronPythonCompletionData2`

The idea is that the `SharedCompletionProvider` is a wrapper around a `CompletionProviderCore` - that implements `IExternalCompletionProviderCore` - when the `SharedCompletionProvider` is constructed it searches the appDomain for types which implement this interface, instantiates them and checks if they match the engine type that the pythonNode is currently set to - if a match is found this core is used to provide completions and methods are redirected to that core.

I have effectively done the same refactor to the existing `IronPythonCompletionProvider` to keep it around - because it was a public type I do not want to move or remove it - to keep it working it works similarly to the `SharedCompletionProvider` and points to an `IronPythonCompletionProviderCore` - which exists in `DSIronPython` so this completionCore can operate with no view and builds on mono.

The existing `IronPythonCompletionProvider` type is basically an empty wrapper pointing to `IronPythonCompletionProviderCore`.

This type also implements `ILegacyPythonCompletionCore` - an internal interface that exists for no other reason than for us to have a shared interface between the View and Model half of the Python implementation. We can remove this interface, and the legacy completion class in 3.0 - but it should be noted again that all use of this type are already removed from Dynamo as of this PR. (I have no decided if they should be removed from the tests yet).

Reflection was used for access to the static methods, as c# 7 cannot define static interface members. Maybe theres a cleaner solution than this, but again, this class is not used anywhere in our code so I felt it was okay to keep it working in this manner for clients who somehow might depend on this type (likely no one).

Screen Shot 2020-06-08 at 7 27 18 PM

TODO - 
- [x] mono build succeeds.
- [x] update tests that used legacy provider to point to new provider - or keep them as they actually just pass their calls down to new provider?...
- [x] add tests for shared provider logic.
- [x] finish todos in code
- [ ] test with CPython3 engine (since there is no provider yet for that engine)

### Declarations


src/Libraries/PythonNodeModelsWpf/IronPythonCompletionProvider.cs
```diff
@@ -0,0 +1,502 @@
...
+        {
+            return (this.providerImplementation as ILegacyPythonCompletionCore).LookupMember(name);
+        }
+
+        /// <summary>
+        /// Generates completion data for the specified text. The text should be everything before
+        /// the dot character that triggered the completion. The text can contain the command line prompt
+        /// '>>>' as this will be ignored.
+        /// </summary>
+        [Obsolete("Please use additional GetDescription method as this one will be removed in Dynamo 3.0.")]
```
aparajit-pratap, line 349: Why tag specific methods from this class as obsolete when the entire class is going to be obsoleted?

mjkkirschner: I did not add this, this obsoletion message was already here, but since I am moving this file back to this project everything appears like an "addition"

mjkkirschner: I can remove these specific obsolete attributes if you think thats more clear and just keep the class level one?

mjkkirschner:
an update, looks like we have more work to do here, likely out of scope for this PR:
while the mono build works, actually trying to use the CLI on ubuntu yields:
```
2020-06-09 01:40:08Z : Python template set to default.
39
2020-06-09 01:40:08Z : Could not create an instance of Dynamo.PackageManager.PackageManagerExtension
40
2020-06-09 01:40:08Z : Invalid Image
41
2020-06-09 01:40:08Z :   at (wrapper managed-to-native) System.Reflection.Assembly.LoadFrom(string,bool,System.Threading.StackCrawlMark&)
42
  at System.Reflection.Assembly.LoadFrom (System.String assemblyFile) [0x00002] in <f2b3ab7dfff746f594d2ef5b16ec3c90>:0 
43
  at Dynamo.Extensions.ExtensionLoader.Load (Dynamo.Extensions.ExtensionDefinition extension) [0x0001a] in <02bd5a0d16c64c4f970ccfe9a394f591>:0 
44
2020-06-09 01:40:09Z : Dynamo -- Build 2.7.0.8624
45
2020-06-09 01:40:09Z : Duplicate migration type registered for SunPathDirection
```

I'm guessing this could be some x64 x86 thing, but not sure.

src/NodeServices/Interfaces.cs
```diff
@@ -188,4 +191,108 @@ internal interface IExtensionApplication
         /// <param name="baseDirectory"></param>
         void PreloadAsmLibraries(string baseDirectory);
     }
+
+    //TODO should this live in DynamoServices?
+    //make these internal somehow?
```
mmisol, line 196: We need this to be public, right? Otherwise only we could add completion providers.

src/NodeServices/Interfaces.cs
```diff
@@ -188,4 +191,108 @@ internal interface IExtensionApplication
...
+        /// <param name="item">Everything after the stub</param>
+        /// <param name="isInstance">Whether it's an instance or not</param>
+        string GetDescription(string stub, string item, bool isInstance);
+
+        bool MatchingEngine(string engineName);
+
+        //TODO think about this, more general?
+        void ImportStdLibrary(string dynamoCorePath);
+
+        void Log(string message);
```
mmisol, line 232: We might want to clean this up.
- `Log` does not feel like it should belong here. If needed, it should probably be in another interface.
- `ImportStdLibrary` is not generic enough for a name. Would `Initialize` be a good name?
- How is `MatchingEngine` used? Shouldn't we ask the engine its name, rather than passing one and receiving a `bool`?
- It's not clear to me what `GetDescription` does by looking at its signature.

src/NodeServices/Interfaces.cs
```diff
@@ -188,4 +191,108 @@ internal interface IExtensionApplication
...
+    public enum ExternalCodeCompletionType
+    {
+        NAMESPACE,
+        METHOD,
+        FIELD,
+        CLASS,
+        PROPERTY,
+        ENUM,
+        OTHER
+    };
```
mmisol, line 275: NIT: In .NET SNAKE_CASE is not the usual for naming enum values. Also, I don't think we are using that convention either. Please change to PascalCase.
Also, the usual convention is to have the "unknown" value (in this case `OTHER`) be 0, so it should go first in the list.

src/NodeServices/Interfaces.cs
```diff
@@ -188,4 +191,108 @@ internal interface IExtensionApplication
...
+        string Text { get; }
+        //
+        // Summary:
+        //     The displayed content. This can be the same as 'Text', or a WPF UIElement if
+        //     you want to display rich content.
+        object Content { get; }
+        //
+        // Summary:
+        //     Gets the description.
+        object Description { get; }
```
mmisol, line 258: Can this also be a WPF UIElement?

src/Dynamo.Mono.sln
```diff
@@ -139,14 +139,14 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DynamoApplications", "Dynam
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "PythonNodeModels", "Libraries\PythonNodeModels\PythonNodeModels.csproj", "{8872CA17-C10D-43B9-8393-5C5A57065EB0}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ShapewaysClient", "ShapewaysClient\ShapewaysClient.csproj", "{756CEBA5-4E20-4403-8448-C3C47957975B}"
-EndProject
```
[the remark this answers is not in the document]
mjkkirschner: it's gone already.

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
```diff
@@ -101,6 +101,8 @@ private void OnTextAreaTextEntered(object sender, TextCompositionEventArgs e)
                 if (e.Text == ".")
                 {
                     var subString = editText.Text.Substring(0, editText.CaretOffset);
+                    //TODO this should return some mono compatible format
+                    //we'll then convert it to PythonCompletionData
```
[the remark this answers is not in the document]
mjkkirschner: nope, this is all set, I will remove the TODO.

src/NodeServices/Interfaces.cs
mjkkirschner: well... yes, but are we happy with this API enough to make it public and keep it around forever? I guess I'd rather switch to it public in the future instead - thoughts?

src/NodeServices/Interfaces.cs
mjkkirschner: `MatchingEngine` is used to check if this provider can work for a given engine. It needs to be able to be called using reflection. Any thoughts for a different signature?

```
if(CompletionProvider.MatchingEngine(engine.name){
//cool lets use this provider
return CompletionProvider
}
```

src/NodeServices/Interfaces.cs
mjkkirschner: The names are used to match the old interface... I could write a converter I guess with a dictionary map instead.

src/NodeServices/Interfaces.cs
mjkkirschner: I think in our case it's always text, but the `ICompletionData` interface from Avalon has this typed as an object, and I was following that pattern, I'll see if we can make it a string.

mjkkirschner:
@mmisol thanks for the review comments! Before I start addressing them I want to see if I can get tests in a better state, right now the completion tests are failing, probably with a null reference somewhere.

mjkkirschner:
@mmisol tests should be good now, waiting for another test run to confirm, but I will start working on addressing your comments now.

src/Libraries/DSIronPython/IronPythonCodeCompletionProviderCore.cs
```diff
@@ -530,683 +797,321 @@ public object LookupMember(string name)
...
-            text = text.Replace("\r", " ");
-            int startIndex = text.LastIndexOf(' ');
-            return text.Substring(startIndex + 1).Trim('.').Trim('(');
-        }
-        #endregion
     }
-}
\ No newline at end of file
+
+}
```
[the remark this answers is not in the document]
mjkkirschner: I will transfer over the summaries - one thing to keep in mind is that I deliberately made this type internal, the members are "public" only to satisfy c#'s interface checking, but they are really internal.

src/NodeServices/Interfaces.cs
mjkkirschner: I think this is resolved.

* Log is removed from the interface and implemented using an existing DynamoCore interface
* ImportStdLibrary is renamed Initialize 
* MatchingEngine I have not changed - open to suggestions.
* GetDescription is a core method needed for the implementation of completion - I'm not yet ready to dive deep and change that - so I'd like to leave it as is, but I did add the summary that I think makes it more clear what it does.

src/NodeServices/Interfaces.cs
mjkkirschner: it compiles with this changed to string - I will do some testing while I am adding additional tests for the new classes.

src/NodeServices/Interfaces.cs
mjkkirschner: I've made these all internal for now...

mjkkirschner:
@mmisol I've added a few tests here to test the new `sharedCompletionProvider` class, and left the existing tests since the new providerCore is just a transplant of that existing class so the coverage should not change (AFAIK).

all tests seem to pass locally, waiting for the result here.

please take another look.

src/Libraries/PythonNodeModelsWpf/IronPythonCompletionProvider.cs
```diff
@@ -0,0 +1,502 @@
...
+        /// <param name="code">The code from which to get the assignments</param>
+        /// <returns>A dictionary matching the name of the variable to a tuple of typeName, character at which the assignment was found, and the CLR type</returns>
+        public Dictionary<string, Tuple<string, int, Type>> FindAllVariables(string code)
+        {
+            return (providerImplementation as ILegacyPythonCompletionCore).FindAllVariables(code);
+        }
+
+        #endregion
+    }
+}
```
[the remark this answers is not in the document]
mjkkirschner: I think I got rid of them - please let me know if you catch any stragglers.

src/Libraries/PythonNodeModelsWpf/IronPythonCompletionProvider.cs
```diff
@@ -0,0 +1,502 @@
...
+        internal static readonly Regex MATCH_FROM_IMPORT_STATEMENTS = new Regex(@"from\s+?([\w.]+)\s+?import\s+?([\w, *]+)", RegexOptions.Compiled | RegexOptions.Multiline);
+        internal static readonly Regex MATCH_VARIABLE_ASSIGNMENTS = new Regex(@"^[ \t]*?(\w+(\s*?,\s*?\w+)*)\s*?=\s*(.+)", RegexOptions.Compiled | RegexOptions.Multiline);
+
+        internal static readonly Regex STRING_VARIABLE = new Regex("[\"']([^\"']*)[\"']", RegexOptions.Compiled);
+        internal static readonly Regex DOUBLE_VARIABLE = new Regex("^-?\\d+\\.\\d+", RegexOptions.Compiled);
+        internal static readonly Regex INT_VARIABLE = new Regex("^-?\\d+", RegexOptions.Compiled);
+        internal static readonly Regex LIST_VARIABLE = new Regex("\\[.*\\]", RegexOptions.Compiled);
+        internal static readonly Regex DICT_VARIABLE = new Regex("{.*}", RegexOptions.Compiled);
+
+        internal static readonly string BAD_ASSIGNEMNT_ENDS = ",([{";
```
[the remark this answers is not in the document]
mjkkirschner: gone

src/Libraries/PythonNodeModelsWpf/IronPythonCompletionProvider.cs
```diff
@@ -0,0 +1,378 @@
...
+            set { (providerImplementation as ILegacyPythonCompletionCore).Scope = value; }
+        }
+
+        /// <summary>
+        /// Already discovered variable types
+        /// </summary>
+        public Dictionary<string, Type> VariableTypes
+        {
+            get => providerImplementation.VariableTypes;
+            set => providerImplementation.VariableTypes = value;
```
mmisol, line 65: Interesting. I guess we should be allowed to use this now that our build machines have been upgraded

mjkkirschner:
@mmisol I'll do one more scan for TODOs to make sure they are not mine.

PR #10756, opened 2020-06-11
Do not copy local interop excel, move calc

mmisol:
### Purpose

Avoid adding Interop excel to our output as it is not expected. Also
make ncalc appear only inside of the 'nodes' folder by copying local
only from CoreNodeModels.

### Declarations

### Reviewers

@mjkkirschner @QilongTang

### DIFF
--- a/src/Libraries/CoreNodeModels/CoreNodeModels.csproj
+++ b/src/Libraries/CoreNodeModels/CoreNodeModels.csproj
@@ -39,6 +39,7 @@
   <ItemGroup>
     <Reference Include="NCalc, Version=1.3.8.0, Culture=neutral, PublicKeyToken=973cde3f1cafed03, processorArchitecture=MSIL">
       <HintPath>..\..\packages\ncalc.1.3.8\lib\NCalc.dll</HintPath>
+      <Private>False</Private>
     </Reference>
     <Reference Include="Newtonsoft.Json, Version=8.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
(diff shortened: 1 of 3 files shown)


src/Libraries/CoreNodeModels/CoreNodeModels.csproj
```diff
@@ -39,6 +39,7 @@
   <ItemGroup>
     <Reference Include="NCalc, Version=1.3.8.0, Culture=neutral, PublicKeyToken=973cde3f1cafed03, processorArchitecture=MSIL">
       <HintPath>..\..\packages\ncalc.1.3.8\lib\NCalc.dll</HintPath>
+      <Private>False</Private>
```
mjkkirschner, line 42: this should be true actually.

previous was:
```
<ItemGroup>
    <Reference Include="NCalc">
      <HintPath>..\..\..\extern\NCalc\NCalc.dll</HintPath>
      <Private>True</Private>
    <Reference Include="NCalc, Version=1.3.8.0, Culture=neutral, PublicKeyToken=973cde3f1cafed03, processorArchitecture=MSIL">
      <HintPath>..\..\packages\ncalc.1.3.8\lib\NCalc.dll</HintPath>
```

we need to deliver this binary somewhere, just not in two places.

mmisol: Added it. Is it because the dll is expected to be in one NuGet package but not in another one?

mjkkirschner: this should end up in the /nodes folder, not in the root directory.

PR #10761, opened 2020-06-12
Support embedded docs in satellites

mmisol:
### Purpose

Adds support for localized documentation content. Localized content
should use the 'EmbeddedResource' build action, just like invariant
content. The only consideration is that the file name should follow the
same convention used for resource files, i.e. add the culture name
before the file extension.

Example:
- DivisionByZero.html <- Invariant culture
- DivisionByZero.en-us.html <- Specific en-us culture
- DivisionByZero.es.html <- Neutral es culture

### Declarations

### Reviewers

@QilongTang @mjkkirschner


mmisol:
Opening this as a WIP to get feedback on the approach. The good thing is we can keep using embedded resources. The bad thing is we kinda need to do the satellite assembly finding ourselves. AFAIK the framework does not provide this for embedded resources.

mjkkirschner:
@mmisol - not to do with this PR but I think I have to mess with the mono build action, I used the `push` event and I guess this does not run on the PR being opened - let me fix that, you might need to do an empty or nonesensical push to get this check to run for now (or an admin can merge)

mmisol:
Don't worry about that. If this PR were to go in I would still need to add tests to it. For now I will just leave it open for feedback.

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
```diff
@@ -276,7 +277,30 @@ private string LoadContentFromResources(string name)
...
+            if (resourceAssembly == null)
+            {
+                if (culture.IsNeutralCulture)
+                {
+                    var specificCulture = CultureInfo.CreateSpecificCulture(culture.Name);
+                    resourceAssembly = assembly.GetSatelliteAssembly(specificCulture);
+                }
+                else
+                {
+                    while (resourceAssembly == null && culture.Parent != CultureInfo.InvariantCulture)
```
mjkkirschner, line 292: hmm, why is this a while loop? Will assembly or culture every be changed by the while loop?

mmisol: You found a bug. That's why I said it needed testing 😂 

In theory we shouldn't need a while loop. There are only two levels of cultures in .NET, beyond invariant which we don't really care about here. Maybe I can just remove the loop to avoid complexity.

mmisol: @mjkkirschner More importantly, what do you think of this approach? My feeling is that it's the natural extension to what we had, and not too complex on our side. I believe the localization team should find this approach easy to use.

QilongTang: I propose we RC2.7.0 now and test this in master ASAP 👍

mmisol: Thanks @QilongTang ! I'll start working on the tests now

mmisol:
This should be ready for review now. Let me know what you think @mjkkirschner @QilongTang

mjkkirschner:
@mmisol looks like some failing tests?

mjkkirschner:
I think this PR addresses part of the problem - we may need to work with localization to come to a solution - because of our design of only surfacing embedded files - this means that the localization team needs to submit files to Dynamo before the build, not after, usually they provide localized .xml files after the build - so we'll need to see if submitting a PR with these localized html files is an acceptable workflow...

Theres also another problem regarding extensions being localized with satellite assemblies in general, ie using the standard workflow described above. (similar to the rest of DynamoCore/DynamoCoreWPF) - IE I'm just not sure if this works for extensions?

I do think that's out of scope for this PR though, and maybe there is actually a different spike on it?

mmisol:
I saw a couple of package loader tests failing. Probably due to the fact that I added a package and they are probably not expecting it.

Did something happen to the log? I think it grew a lot since last time I checked.

About the globalization team, we should probably ask them. What you say about submitting the files before the build is what I would do with normal resx resources as well, but they seem to do things differently. In the approach we are using here, each html file acts exactly in the same role as a resx file: they use the same extensions and end up in satellite assemblies as well. It shouldn't be a special case to them but more of the same.

mjkkirschner:
to avoid filling the machines with dynamo logs, @reddyashish redirected the dynamo log output to the std out when test mode is active.

mjkkirschner:
I've pinged the Globalization team in the task to consider if this approach will work for them.

mjkkirschner:
@mmisol are you planning to move those resources around in this PR or in a followup?

mmisol:
I'll do that here, so they can also get that change. Should be a few minutes.

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
```diff
@@ -16,8 +16,8 @@ public class DocumentationBrowserViewModel : NotificationObject, IDisposable
         #region Constants
 
         private const string HTML_TEMPLATE_IDENTIFIER = "%TEMPLATE%";
-        private const string BUILT_IN_CONTENT_INTERNAL_ERROR_FILENAME = nameof(Resources.InternalError) + ".html";
-        private const string BUILT_IN_CONTENT_FILE_NOT_FOUND_FILENAME = nameof(Resources.FileNotFound) + ".html";
+        private const string BUILT_IN_CONTENT_INTERNAL_ERROR_FILENAME = "InternalError.html";
+        private const string BUILT_IN_CONTENT_FILE_NOT_FOUND_FILENAME = "FileNotFound.html";
```
mjkkirschner, line 20: is there another one of these somewhere for NoContent?

mjkkirschner: oh right below this...

mmisol: The fact those files were included as resources was unnecessary. They were also embedded resources, so that was redundant. Too many hands on this feature was probably the cause :)

PR #10762, opened 2020-06-12
[DYN-2767] Use Python logo in the PythonUpdate dialog.

reddyashish:
### Purpose

We will be using the Python logo in the Python update dialog instead of using an image that is a bit confusing for users. 

Regarding the text size in the dialog, it is consistent with the design in other dialogs that are currently being used. If it is going to be a small change to fix it, do we still want to file a different task for it?

Screen Shot 2020-06-12 at 11 03 56 AM

### Declarations

### Reviewers

@DynamoDS/dynamo


QilongTang:
@reddyashish Looks much better! Looks like we are not using a dialog template here, were you able to adjust the font for the dialog?

reddyashish:
We have using an API to create the dialog and the font is not set through that. I am looking into it.

reddyashish:
For both the IronPython dialog and the general task dialog, the summaryText font-size is set to 13 and the descriptionText font-size is set to 16. These are the only 2 dialog templates present. https://github.com/DynamoDS/Dynamo/blob/e3d60fbe3a9aa009308fa2233bdd0091eb30caf1/src/PythonMigrationViewExtension/IronPythonInfoDialog.xaml#L45
https://github.com/DynamoDS/Dynamo/blob/e3d60fbe3a9aa009308fa2233bdd0091eb30caf1/src/PythonMigrationViewExtension/IronPythonInfoDialog.xaml#L45

This is how it looks with summaryText size: 15 and descriptionText size: 12. 
Screen Shot 2020-06-15 at 10 00 05 AM
 
Let me know what you guys think.

mjkkirschner:
LGTM

reddyashish:
This is how the crash dialog would look after the new changes:
Screen Shot 2020-06-15 at 1 06 47 PM

PR #10763, opened 2020-06-12
python code smells

mjkkirschner:
### Purpose

working through some codesmells from the sonar scan - 
theres one particular bug it identified that I think deserves some consideration.
@reddyashish @mmisol  - 

`prev_code` was never being set so python was being setup over and over again... does this change look good is should we move the thread call out of this if statement?

### Declarations

### DIFF
--- a/src/Libraries/DSCPython/CPythonEvaluator.cs
+++ b/src/Libraries/DSCPython/CPythonEvaluator.cs
@@ -49,6 +49,8 @@ namespace DSCPython
                     PythonEngine.Initialize();
                     PythonEngine.BeginAllowThreads();
                 }
+                
+                prev_code = code;
 
                 IntPtr gs = PythonEngine.AcquireLock();
                 try
@@ -77,14 +79,14 @@ namespace DSCPython
                             catch (Exception e)
                             {
                                 OnEvaluationEnd(false, scope, code, bindingValues);
-                                throw e;
+                                throw;
                             }
                         }
                     }
                 }
                 catch (PythonException pe)
                 {
-                    throw pe;
+                    throw;
                 }
                 finally
                 {
--- a/src/Libraries/DSCPython/CPythonEvaluator.cs
+++ b/src/Libraries/DSCPython/CPythonEvaluator.cs
@@ -23,8 +23,6 @@ namespace DSCPython
     [IsVisibleInDynamoLibrary(false)]
     public static class CPythonEvaluator
     {
-        /// <summary> stores a copy of the previously executed code</summary>
-        private static string prev_code { get; set; }
 
         /// <summary>
         ///     Executes a Python script with custom variable names. Script may be a string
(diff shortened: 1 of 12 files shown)


src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -49,6 +49,8 @@ public static class CPythonEvaluator
                     PythonEngine.Initialize();
                     PythonEngine.BeginAllowThreads();
                 }
+                
+                prev_code = code;
```
mmisol, line 53: Wouldn't this make the check in line 43 be useless?

mmisol, line 53: So I guess check in line 43 was not really working and now it is. Probably not what we want to do, given that we are now getting null as a result if the code didn't change. I'd say we get rid of the check unless @reddyashish says otherwise.

src/Libraries/DSIronPython/IronPythonCodeCompletionProviderCore.cs
```diff
@@ -219,7 +217,6 @@ private Type TryGetTypeFromFullName(string name)
             for (var i = 0; i < matches.Count; i++)
             {
                 var wholeLine = matches[i].Groups[0].Value.TrimEnd('\r', '\n');
-                var libName = matches[i].Groups[3].Value.Trim();
```
mmisol, line 222: Is it ok that this is not needed? Who would know?

src/Libraries/DSCPython/CPythonEvaluator.cs
mjkkirschner: We do a similar check in the ironPythonEvaluator upon which this code is based: 
https://github.com/DynamoDS/Dynamo/blob/master/src/Libraries/DSIronPython/IronPythonEvaluator.cs#L104

I think it is avoid re-executing and importing libraries unnecessarily, I'm not sure what it means it this context, but it looks like the python std lib is reloaded somehow if we don't make this check?

src/Libraries/DSIronPython/IronPythonCodeCompletionProviderCore.cs
mjkkirschner: the local var seems to be copy paste error from another similar method.

src/Libraries/DSCPython/CPythonEvaluator.cs
mmisol: But right now it doesn't seem correct. If you take a look at CPython, it's returning null when the code did not change. For IronPython that is not the case. I don't think the optimization as it is now in CPython is safe in all cases.

mjkkirschner: ahh, I see, let me see if theres a smarter place to put this check.

mjkkirschner: @mmisol I think you are right - I got rid of it, heres the setupPython code:
https://github.com/henon/Python.Included/blob/master/Python.Included/Installer.cs#L37

we're not calling `force` so this code should only end up loading python once.

mjkkirschner:
@reddyashish please take a look at this when you can.

reddyashish:
Looks good to me.

PR #10769, opened 2020-06-15
fix failing meshtoolkit tests

mjkkirschner:
fix failing meshtoolkit tests - now that Dynamo trunk version is 2.8 - we need to force mesh toolkit to be loaded since 2.7 folder is not loaded by default.

To avoid breaking these tests at each update, just set 2.7 as a custom package path. Verified they pass locally.


src/DynamoCoreWpf/ViewModels/Watch3D/shaderSource/vsDynamoPointLine.hlsl
```diff
@@ -13,6 +13,8 @@ GSInputPS main(VSInputPS input)
 
 	//set position into clip space	
 	output.p = mul(output.p, mWorld);
+	//geometryShader uses this field, so it must be set.
+	output.wp = output.p;
```
mjkkirschner, line 17: this is unrelated, I will remove it and send a separate PR.

PR #10777, opened 2020-06-17
[DYN-2786] PythonMigrationExtension should find CustomNodes that use IronPython2.x

reddyashish:
### Purpose

This PR will add the functionality to check for any IronPython dependencies in the CN's that are in the workspace. It recursively checks all the custom nodes in the workspace and marks each custom node with one of the 3 following values. 
 1) NoDependency.
2) NestedDependency.
3) DirectDependency. 

This data is stored in a dictionary and the value from here will be used if the CN is already processed. This is to avoid any performance issues. 

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/PythonMigrationViewExtension/GraphPythonDependencies.cs
+++ b/src/PythonMigrationViewExtension/GraphPythonDependencies.cs
@@ -44,15 +44,15 @@ namespace Dynamo.PythonMigration
             var customNodeManager = ViewLoaded.StartupParams.CustomNodeManager;
             var customNodes = workspace.Nodes.OfType<Function>();
 
-            return CustomNodeContainsIronPythonDependency(customNodes, customNodeManager);           
+            return CustomNodesContainIronPythonDependency(customNodes, customNodeManager);           
         }
 
         // This function returns true, if any of the custom nodes in the input list has an IronPython dependency. 
         // It traverses all CN's in the given list of customNodes and marks them as true in CustomNodePythonDependency, 
         // if the prarent custom node or any of its child custom nodes contain an IronPython dependency.
-        internal static bool CustomNodeContainsIronPythonDependency(IEnumerable<Function> customNodes, ICustomNodeManager customNodeManager)
+        internal static bool CustomNodesContainIronPythonDependency(IEnumerable<Function> customNodes, ICustomNodeManager customNodeManager)
         {
-            var currentCNContainsPythonDependency = false;
+            var containIronPythonDependency = false;
 
             foreach (var customNode in customNodes)
             {
@@ -64,7 +64,7 @@ namespace Dynamo.PythonMigration
                 {
                     if (dependency == CNPythonDependency.DirectDependency || dependency == CNPythonDependency.NestedDependency)
                     {
-                        currentCNContainsPythonDependency = true;
+                        containIronPythonDependency = true;
                     }
                     continue;
                 }
@@ -74,7 +74,7 @@ namespace Dynamo.PythonMigration
                 if (hasPythonNodesInCustomNodeWorkspace)
                 {
                     CustomNodePythonDependency.Add(customNodeWS.CustomNodeId, CNPythonDependency.DirectDependency);
-                    currentCNContainsPythonDependency = true;
+                    containIronPythonDependency = true;
                 }
                 else
                 {
@@ -86,19 +86,19 @@ namespace Dynamo.PythonMigration
 
                 if (nestedCustomNodes.Any())
                 {
-                    hasPythonNodesInCustomNodeWorkspace = CustomNodeContainsIronPythonDependency(nestedCustomNodes, customNodeManager);
+                    hasPythonNodesInCustomNodeWorkspace = CustomNodesContainIronPythonDependency(nestedCustomNodes, customNodeManager);
 
                     // If a custom node contains an IronPython dependency in its sub-tree,
                     // update its corresponding value to 'NestedDependency' in CustomNodePythonDependency.
                     if (hasPythonNodesInCustomNodeWorkspace)
                     {
                         CustomNodePythonDependency[customNodeWS.CustomNodeId] = CNPythonDependency.NestedDependency;
-                        currentCNContainsPythonDependency = true;
+                        containIronPythonDependency = true;
                     }
                 }
             }
 
-            return currentCNContainsPythonDependency;
+            return containIronPythonDependency;
         }
 
         internal bool ContainsCPythonDependencies()
--- a/src/PythonMigrationViewExtension/GraphPythonDependencies.cs
+++ b/src/PythonMigrationViewExtension/GraphPythonDependencies.cs
@@ -13,9 +13,10 @@ namespace Dynamo.PythonMigration
     {
         private ViewLoadedParams ViewLoaded { get; set; }
 
-        private static Dictionary<Guid, CNDependency> CustomNodePythonDependency = new Dictionary<Guid, CNDependency>();
+        // A dictionary to mark Custom Nodes if they have a IronPython dependency or not. 
+        private static Dictionary<Guid, CNPythonDependency> CustomNodePythonDependency = new Dictionary<Guid, CNPythonDependency>();
 
-        enum CNDependency
+        private enum CNPythonDependency
         {
             NoDependency,
             SubtreeDependency,
@@ -59,9 +60,9 @@ namespace Dynamo.PythonMigration
 
                 // If a custom node workspace is already checked for IronPython dependencies, 
                 // check the CustomNodePythonDependency dictionary instead of processing it again. 
-                if (CustomNodePythonDependency.TryGetValue(customNodeWS.CustomNodeId, out CNDependency dependency))
+                if (CustomNodePythonDependency.TryGetValue(customNodeWS.CustomNodeId, out CNPythonDependency dependency))
                 {
-                    if (dependency == CNDependency.DirectDependency || dependency == CNDependency.SubtreeDependency)
+                    if (dependency == CNPythonDependency.DirectDependency || dependency == CNPythonDependency.SubtreeDependency)
                     {
                         currentCNContainsPythonDependency = true;
                     }
@@ -72,12 +73,12 @@ namespace Dynamo.PythonMigration
 
                 if (hasPythonNodesInCustomNodeWorkspace)
                 {
-                    CustomNodePythonDependency.Add(customNodeWS.CustomNodeId, CNDependency.DirectDependency);
+                    CustomNodePythonDependency.Add(customNodeWS.CustomNodeId, CNPythonDependency.DirectDependency);
                     currentCNContainsPythonDependency = true;
                 }
                 else
                 {
-                    CustomNodePythonDependency.Add(customNodeWS.CustomNodeId, CNDependency.NoDependency);
+                    CustomNodePythonDependency.Add(customNodeWS.CustomNodeId, CNPythonDependency.NoDependency);
                 }
 
                 // Recursively check for IronPython dependencies in the nested custom nodes.
@@ -92,7 +93,7 @@ namespace Dynamo.PythonMigration
                     CustomNodePythonDependency.TryGetValue(customNodeWS.CustomNodeId, out dependency);
                     if (hasPythonNodesInCustomNodeWorkspace)
                     {
-                        CustomNodePythonDependency[customNodeWS.CustomNodeId] = CNDependency.SubtreeDependency;
+                        CustomNodePythonDependency[customNodeWS.CustomNodeId] = CNPythonDependency.SubtreeDependency;
                     }
 
                     if (hasPythonNodesInCustomNodeWorkspace)
--- a/src/PythonMigrationViewExtension/GraphPythonDependencies.cs
+++ b/src/PythonMigrationViewExtension/GraphPythonDependencies.cs
@@ -69,7 +69,7 @@ namespace Dynamo.PythonMigration
                 // Recursively check for IronPython dependencies in the nested custom nodes.
                 var nestedCustomNodes = customNodeWS.Nodes.OfType<Function>();
 
-                if (nestedCustomNodes.Count() > 0)
+                if (nestedCustomNodes.Any())
                 {
                     hasPythonNodesInCustomNodeWorkspace = CustomNodeContainsIronPythonDependency(nestedCustomNodes, customNodeManager);
 
--- a/src/PythonMigrationViewExtension/GraphPythonDependencies.cs
+++ b/src/PythonMigrationViewExtension/GraphPythonDependencies.cs
@@ -13,7 +13,14 @@ namespace Dynamo.PythonMigration
     {
         private ViewLoadedParams ViewLoaded { get; set; }
 
-        private static Dictionary<Guid, bool> CustomNodePythonDependency = new Dictionary<Guid, bool>();
+        private static Dictionary<Guid, CNDependency> CustomNodePythonDependency = new Dictionary<Guid, CNDependency>();
+
+        enum CNDependency
+        {
+            NoDependency,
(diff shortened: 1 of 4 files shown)


test/DynamoCoreWpfTests/ViewExtensions/PythonMigrationViewExtensionTests.cs
```diff
@@ -264,5 +261,29 @@ public void CanOpenDocumentationBrowserWhenMoreInformationIsClicked()
...
 
+        /// <summary>
+        /// This test checks that the IronPython dialog is shown to the user,
+        /// when the workspace has custom nodes that contain a python node in it. 
+        /// </summary>
+        [Test]
+        public void WillDisplayDialogWhenCustomNodeInsideWorkspaceHasIronPythonNode()
+        {
+            DebugModes.LoadDebugModesStatusFromConfig(Path.Combine(GetTestDirectory(ExecutingDirectory), "DynamoCoreWpfTests", "python2ObsoleteMode.config"));
+            DynamoModel.IsTestMode = false;
```
mjkkirschner, line 272: why do we need to set test mode false?

mjkkirschner:
I wonder what kind of performance impact this has to graphs that contain lots of custom nodes.

test/DynamoCoreWpfTests/ViewExtensions/PythonMigrationViewExtensionTests.cs
reddyashish: In the test mode true, we don't show the dialogs. https://github.com/DynamoDS/Dynamo/blob/master/src/PythonMigrationViewExtension/PythonMigrationViewExtension.cs#L139

reddyashish:
I was thinking after the user gets this information for the first time, they can control it with the MenuItem. If the user doesn't doesn't want to see these dialogs anymore, we shouldn't be checking for IronPython dependencies anymore. 
On the other hand, in the case where it is enabled, the complexity should be linear if there is one nesting level of CN's. If there are many levels (with no python nodes), it would be polynomial. Do you think it would be good to profile it?

mjkkirschner:
@reddyashish I think profiling it would be a great idea using a few different graphs (ie normal case, and pathologically bad case) (many levels of nesting)

mjkkirschner:
we can also think about cacheing this data - for example, once we figure out if a custom node and its sub tree contains ironPython... we probably don't need to recheck that subtree of nodes right?

reddyashish:
For the first example, if any custom node or its subtree has a python node, we return true immediately, so we wouldn't check for dependencies in its subtree or in other CN's in the workspace. 
But yes, it would make a difference, if that custom node is placed multiple times. We can probably store the CN's that we checked already and not re-check them.

mjkkirschner:
also consider the case where there are many custom nodes and none of them contain python, all custom node instances will be searched over and over - I think it's likely we should implement cacheing here, but let's wait to see what the difference in time is for a bad case.

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
```diff
@@ -18,10 +21,49 @@ internal GraphPythonDependencies(ViewLoadedParams viewLoadedParams)
...
+            {
+                customNodeManager.TryGetFunctionWorkspace(customNode.FunctionSignature, false, out customNodeWS);
+
+                if (customNodeWS.Nodes.Any(n => IsIronPythonNode(n)))
+                {
+                    return true;
+                }
+
+                // Recursively check for IronPython dependencies in the nested custom nodes. 
+                var nestedCumtomNodes = customNodeWS.Nodes.OfType<Function>();
```
aparajit-pratap, line 59: typo: `nestedCustomNodes`

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
```diff
@@ -18,10 +21,49 @@ internal GraphPythonDependencies(ViewLoadedParams viewLoadedParams)
...
+            // Check if any of the custom nodes has IronPython dependencies in it. 
+            var customNodeManager = ViewLoaded.StartupParams.CustomNodeManager;
+            var customNodes = workspace.Nodes.OfType<Function>();
+
+            if (CustomNodesContainIronPythonDependencies(customNodes, customNodeManager))
+            {
+                return true;
+            }
+
+            return false;
```
aparajit-pratap, line 42: You can directly return `CustomNodesContainIronPythonDependencies(customNodes, customNodeManager)` here.

aparajit-pratap:
I like the caching solution if I understand it correctly that you guys are talking about implementing it for the custom node definition itself as it wouldn't make sense for the custom node instance.

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
reddyashish: Actually, if it returns false we don't want to return it because we needs to continue checking for other custom nodes till it finds the first one.

aparajit-pratap: The above single liner is effectively doing the same thing as the `if-else`.

reddyashish: Ah, sorry, I thought you were referring to the CustomNodesContainIronPythonDependencies() call in the recursive function. Changed it accordingly.

reddyashish:
@mjkkirschner @aparajit-pratap I did some testing with these changes.
1) When the workspace has less than 100 CN's, the workspace opens up immediately with the dialog.
2) When the workspace has around 250 CN's, it takes 1-3 secs to open up the workspace and the dialog. 
3) When the workspace has around 500 CN's, it takes around 3-5 secs. It is almost the same time that the current master takes to open the workspace. 

Current master:
Current master

After this PR change:
After this PR change

I don't think there is a noticeable difference after these changes. Let me know what you think about implementing a caching solution.

reddyashish:
Added a caching solution to not check the custom node workspace, that is already processed. We store the custom node workspace id's in the dictionary object when it is processed for the first time and avoid re-checking of the Python dependencies. Note that this would not check for the whole custom node subtree, but just the last CN workspace that has Python nodes.

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
```diff
@@ -46,14 +47,32 @@ internal bool CustomNodesContainIronPythonDependencies(IEnumerable<Function> cus
...
-                if (customNodeWS.Nodes.Any(n => IsIronPythonNode(n)))
+                // If a custom node workspace is already checked for Python dependencies, 
+                // check the dictionary object instead of processing it again. 
+                if (CustomNodePythonDependency.ContainsKey(customNodeWS.CustomNodeId))
+                {
+                    if (CustomNodePythonDependency[customNodeWS.CustomNodeId])
+                    {
+                        return true;
+                    }
+                    else
```
aparajit-pratap, line 58: you can directly `continue` here.

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
```diff
@@ -46,14 +47,32 @@ internal bool CustomNodesContainIronPythonDependencies(IEnumerable<Function> cus
             {
                 customNodeManager.TryGetFunctionWorkspace(customNode.FunctionSignature, false, out customNodeWS);
 
-                if (customNodeWS.Nodes.Any(n => IsIronPythonNode(n)))
+                // If a custom node workspace is already checked for Python dependencies, 
+                // check the dictionary object instead of processing it again. 
+                if (CustomNodePythonDependency.ContainsKey(customNodeWS.CustomNodeId))
```
aparajit-pratap, line 52: You can use `TryGetValue` on the dictionary. You can check for the key and get the value for that key in one step.

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
```diff
@@ -46,14 +47,32 @@ internal bool CustomNodesContainIronPythonDependencies(IEnumerable<Function> cus
...
+                if (pythonNodesInCustomNodeWorkspace)
                 {
                     return true;
                 }
 
                 // Recursively check for IronPython dependencies in the nested custom nodes. 
                 var nestedCustomNodes = customNodeWS.Nodes.OfType<Function>();
-                if (CustomNodesContainIronPythonDependencies(nestedCustomNodes, customNodeManager))
+
+                if (nestedCustomNodes.Count() > 0 && CustomNodesContainIronPythonDependencies(nestedCustomNodes, customNodeManager))
```
aparajit-pratap, line 75: use `Any()` instead of `Count()`.

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
aparajit-pratap: How do you keep the dictionary up-to-date if a custom node WS is edited - say, one or more python nodes are deleted and the WS no longer has any python node.

mjkkirschner: I was also wondering this - but - since these messages are only raised once per session anyway maybe this is a non issue, next time dynamo is opened the data will be up to date.

reddyashish:
I noticed one issue which can be described by the following example. If the workspace has CustomNodeA in the current workspace. CustomNodeA has CustomNodeB and CustomNodeC. CustomNodeC has Python nodes in it but CustomNodeB doesn't. Now when you open this graph, the dialog will be shown and the caching dictionary at this point, will have these values: [CustomNodeA, false], [CustomNodeB, false], [CustomNodeC, true] (since we are just marking the actual CN workspace that has Python nodes). Now when you open CustomNodeA.dyf, it would check the dictionary because we already processed that CN and since its value is false, the CustomNodeContainsIronPythonDependency() returns false. 
To avoid these cases, I have modified this function to mark all the CN's in the subtree if any of its child CN's has Python dependency. By doing this, we avoid these recursive calls when nested CN's are opened from the home workspace. 

Let me know what you guys think about it.

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
reddyashish: Yeah thats a good point, but like Mike said, it wouldn't be a problem if the python node is deleted from CN workspace because the dialog would be shown already when it was opened for the first time. On the other hand, when you add a new Python node to a CN workspace, save it and open that CN workspace again, the dialog would be shown for the first time because it checks the current workspace (that is opened) for Python dependencies here https://github.com/DynamoDS/Dynamo/pull/10777/files#diff-f13147450a5b0038bab03d9f6493aff8R32

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
```diff
@@ -18,10 +23,72 @@ internal GraphPythonDependencies(ViewLoadedParams viewLoadedParams)
...
+            }
+
+            // Check if any of the custom nodes has IronPython dependencies in it. 
+            var customNodeManager = ViewLoaded.StartupParams.CustomNodeManager;
+            var customNodes = workspace.Nodes.OfType<Function>();
+
+            return CustomNodeContainsIronPythonDependency(customNodes, customNodeManager);           
+        }
+
+        internal bool CustomNodeContainsIronPythonDependency(IEnumerable<Function> customNodes, ICustomNodeManager customNodeManager)
```
mjkkirschner, line 42: I think it would be good to write /// summary for this function.

mjkkirschner: to help clarify my understanding - would a good summary be: 

```
This function traverses all custom nodes and their subtrees and marks them as containing ironPython dependencies if any of their child nodes contain IronPython.

```

mjkkirschner: in addition to a summary, I think this function should have a unit test written if possible.

reddyashish: That would be a good summary but with condition that it traverses until the first CN with an IronPython dependency is found. Also had to make the method static, so as to call it from the tests. Added a unit test for this function as well.

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
```diff
@@ -18,10 +23,75 @@ internal GraphPythonDependencies(ViewLoadedParams viewLoadedParams)
...
+                    }
+                    continue;
+                }
+
+                var hasPythonNodesInCustomNodeWorkspace = customNodeWS.Nodes.Any(n => IsIronPythonNode(n));
+                CustomNodePythonDependency.Add(customNodeWS.CustomNodeId, hasPythonNodesInCustomNodeWorkspace);
+
+                if (hasPythonNodesInCustomNodeWorkspace)
+                {
+                    return true;
```
aparajit-pratap, line 69: I think you should add it to the dictionary here: 
`CustomNodePythonDependency.Add(customNodeWS.CustomNodeId, true);`

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
```diff
@@ -18,10 +23,75 @@ internal GraphPythonDependencies(ViewLoadedParams viewLoadedParams)
...
+                }
+
+                // Recursively check for IronPython dependencies in the nested custom nodes.
+                var nestedCustomNodes = customNodeWS.Nodes.OfType<Function>();
+
+                if (nestedCustomNodes.Any())
+                {
+                    hasPythonNodesInCustomNodeWorkspace = CustomNodeContainsIronPythonDependency(nestedCustomNodes, customNodeManager);
+
```
aparajit-pratap, line 78: Then simply add to the dictionary here:
`CustomNodePythonDependency.Add(customNodeWS.CustomNodeId, hasPythonNodesInCustomNodeWorkspace);`
Then you won't need any of the logic below.

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
reddyashish: @aparajit-pratap I am not sure about this. Can we discuss this after the standup?

reddyashish:
Modified the CustomNodeContainsIronPythonDependency() to traverse all the CN's(and its subtree) in the workspace and not return after the first Python dependency is found. This change was made to address the next task: https://jira.autodesk.com/browse/DYN-2797.

reddyashish:
@mjkkirschner @aparajit-pratap Using an enum to differentiate between CN's that have a direct dependency or a subtree dependency. So we can use this dictionary object to find all the CN's that have a direct dependency and show them in the dialog. I think I addressed all the comments now, Please take a look and let me know If I missed anything.

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
```diff
@@ -13,7 +13,14 @@ public class GraphPythonDependencies
     {
         private ViewLoadedParams ViewLoaded { get; set; }
 
-        private static Dictionary<Guid, bool> CustomNodePythonDependency = new Dictionary<Guid, bool>();
+        private static Dictionary<Guid, CNDependency> CustomNodePythonDependency = new Dictionary<Guid, CNDependency>();
+
+        enum CNDependency
+        {
+            NoDependency,
+            SubtreeDependency,
```
aparajit-pratap, line 22: I think renaming this to NestedDependency will be clearer.

aparajit-pratap:
@reddyashish for the next task can we get the custom node names from the guids in this dictionary?

reddyashish:
Yes, I think it can be done as I am storing the GUID from FunctionSignature attribute and not the custom node instance id. Then, we can get the CustomNode info using the following API. https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Core/CustomNodeManager.cs#L935.

PR #10779, opened 2020-06-17
[DYN-2834] Fix the issue where package manager extension object was being null 

reddyashish:
#Purpose

This PR fixes the regression introduced in 2.7. 

When the package is being installed from the workspace references, the SetPackageState() looses the reference to the pmExtension object and it is null. So it is throwing a null pointer exception and as a result of this the nodes are not being resolved. 
When you open the Search for a Package window, you can see this message.
Screen Shot 2020-06-17 at 5 59 17 PM

The package is actually downloaded but failed to load in Dynamo. 

So, reverting this part of the code to as it was before. 

### Declarations

### Reviewers

@QilongTang @mjkkirschner


mjkkirschner:
I'm confused how this change introduced the regression?
Look's like it came from a code smell PR?

mjkkirschner:
ahhh @reddyashish here's the issue:

https://github.com/DynamoDS/Dynamo/blob/0b8f36629c516ec318b8acb2a1966f12429b416f/src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs#L188

this class has another field with the same name as that local variable - thats why this compiles at all - just change add the pmExtension as a parameter, change the name, or set the field, no need to revert the change.

QilongTang:
@mjkkirschner @reddyashish Shall we just change this line to set the field instead of defining a local one? https://github.com/DynamoDS/Dynamo/pull/10779/files#diff-87708e28021df7b325448575f92c8e69R712

mjkkirschner:
nice!
@reddyashish do you think you can write a test for this that would have failed?

QilongTang:
@mjkkirschner Do you still prefer a test written before merging?

reddyashish:
@QilongTang Discussed with Mike, about adding the test in a different PR because there are some issues with writing the test for this workflow. I am creating a task to followup it up. So merging this to unblock 2.7

PR #10781, opened 2020-06-18
Changing engine version in python editor should run current code in editor.

mjkkirschner:
### Purpose

When the python engine was changed a run is triggered, but it used the NodeModel.Script code, not the latest code which was entered into the text editor.

A handler is attached to the dropdown selection changed event - and on this event the code on the model is updated.

A test is added.

@QilongTang pointed out that this approach was causing multiple executions to be kicked off as setting the Engine property modified the node, and so did calling `UpdateScript` from the script editor window.

To avoid this, I've removed the nodeModified call from the `Engine` property setter - so it is unified with the `Script` property, to make these nodes run again `NodeModified` must be called explicitly after these operations, this lets us better control when they run.

I've also added explicit `NodeModified` calls into the handlers for the `PythonFromString` node when a user right clicks that node and selects the engine.

The test is updated to check how many evaluations have occurred.

### Declarations


test/DynamoCoreWpfTests/PythonNodeCustomizationTests.cs
```diff
@@ -82,6 +72,69 @@ public void CanChangeEngineFromScriptEditorDropDown()
...
+
+            //modify code in editor
+            Assert.AreEqual("ok",(nodeModel as PythonNode).Script);
+            SetTextEditorText(scriptWindow, "OUT = 100");
+            //modify engine
+            engineSelectorComboBox.SelectedItem = engineChange;
+
+            //assert model code is updated.
+            Assert.AreEqual("OUT = 100", (nodeModel as PythonNode).Script);
+
```
mmisol, line 104: Can you also assert that the value of the node is 100? Also, a test name like `ChangingDropdownEngineSavesCodeBeforeRunning` would better express what we want to achieve here.

mjkkirschner:
@mmisol - review comments addressed.

test/DynamoCoreWpfTests/PythonNodeCustomizationTests.cs
```diff
@@ -101,6 +107,9 @@ public void ChangingDropdownEngineSavesCode()
 
             //assert model code is updated.
             Assert.AreEqual("OUT = 100", (nodeModel as PythonNode).Script);
+            DispatcherUtil.DoEvents();
+            Run();
```
mmisol, line 111: Do you need to call run manually? I'm confused now

mjkkirschner: This graph is in manual mode.
https://github.com/DynamoDS/Dynamo/blob/master/test/core/python/python.dyn#L55

mmisol: Oh. Should we use that uses auto then? With manual mode you wouldn't be able to reproduce this issue in the first place.

mjkkirschner: sure, I'll find a python graph in auto.

mmisol: Ok. Then feel free to ignore my previous comment or go all the way and use auto mode. Sorry about the misunderstanding.

mjkkirschner: ok, instead of finding a python graph with auto, I just set the runtype to auto right after opening it, and I removed the manual run call.

QilongTang:
@mjkkirschner Although I understand the intension is to make sure Code is updated and trigger a run. If I remember correctly both Engine property or Script property triggers run in Python node? If that is the case, are we introducing two runs now instead of one, can you verify?

mjkkirschner:
@QilongTang I think you're right - kind of.
modifying the `Script` property does not trigger a nodeModified call - thats only done when the text is updated from the script editor using a command and manual node modified call.

The `engine` DOES always trigger a nodeModified call.

Let me try to unify this to avoid extra runs.

mjkkirschner:
@QilongTang @mmisol please take another look, did a bit of a refactor based on @QilongTang 's comment about multiple runs.

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -42,7 +42,6 @@ public PythonEngineVersion Engine
                 {
                     engine = value;
                     RaisePropertyChanged(nameof(Engine));
-                    OnNodeModified();
```
mmisol, line 45: I'm afraid this may have unforeseen consequences. Should we check all places where we are setting this property and make sure we are doing what we want to do (trigger or not).

src/Libraries/PythonNodeModelsWpf/PythonStringNode.cs
```diff
@@ -72,6 +73,7 @@ private void UpdateToPython2Engine(object sender, EventArgs e)
         private void UpdateToPython3Engine(object sender, EventArgs e)
         {
             pythonStringNodeModel.Engine = PythonEngineVersion.CPython3;
+            pythonStringNodeModel.OnNodeModified();
```
mmisol, line 76: I see this is being done for `PythonStringNode` but I think `PythonNode` has similar handlers. Please check.

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
```diff
@@ -174,5 +174,16 @@ private void OnMoreInfoClicked(object sender, RoutedEventArgs e)
         {
             dynamoViewModel.OpenDocumentationLinkCommand.Execute(new OpenDocumentationLinkEventArgs(new Uri(PythonNodeModels.Properties.Resources.PythonMigrationWarningUriString, UriKind.Relative)));
         }
+
+        private void OnEngineChanged(object sender, System.Windows.Controls.SelectionChangedEventArgs e)
+        {
+            //removedItems will be empty during the first binding
+            //as the window is constructed, we don't want to execute the node just
+            //as a consequence of opening the editor.
+            if (e.RemovedItems.Count > 0)
```
mmisol, line 183: So this is always 1 because the selection changing implies "removing" the previous item from the selection, right?

src/Libraries/PythonNodeModels/PythonNode.cs
mjkkirschner: Engine is a relatively new property only having been introduced since the python 3 changes were brought in - *I think*.

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
mjkkirschner: it's always 1 after the first time the binding is setup when the window is created.

Dewb:
Given all of the above, is it good for the user if we trigger a run on engine change? What's the motivation and what would we lose by not triggering a run immediately?

mjkkirschner:
@mmisol @QilongTang I've done the following:

1. added the nodeModified calls to the `pythonScript` node context menu handlers as @mmisol pointed out, I missed those, and I added a view test.
2. modified the utility methods for setting the engine in the python edit tests to also call NodeModified so that the code actually runs and perform a valid check of old result vs new result. (it might be wise to check that the number of executions has changed in these tests).

mjkkirschner:
Hi @Dewb,  we discussed this previously, and a bit more today.

I think the consensus is that for the user experience, changing the engine should mark the node as modified, in automatic mode this means the newest code is executed immediately, and in manual mode, next time the run button is pressed.

There are 2 reasons we discussed-

* consistency:  when changing most properties that have an effect on execution results via the UI a node is re-executed or marked modified - for example, a `string` or `number` node's UI input.
 Because `engine` can bet set from the context menu ,and the script editor, we thought this was an even more compelling case for the modifying the engine UI marking the node as modified. For example, `lacing` or `frozen` properties changed via the context menu mark the node as modified as they may change execution result.

*  avoiding confusion during python migration testing: As a user is experimenting with python3, we want to avoid confusing results where the node appears to execute code successfully, but in fact has simply not executed at all. The reason we are investigating this issue to begin with is to avoid the existing bug where a run is kicked off, but old code was executed.

mjkkirschner:
@mmisol @QilongTang self serve passed, seems we'll have to be careful with the PR check until we restore the perf machine.

test/Libraries/DynamoPythonTests/PythonEditTests.cs
```diff
@@ -11,6 +11,7 @@
 using PythonNodeModels;
 using DynCmd = Dynamo.Models.DynamoModel;
 using Newtonsoft.Json.Linq;
+using Dynamo.Graph.Workspaces;
```
mmisol, line 14: This is going to add a smell if you don't sort, I think

src/Libraries/PythonNodeModelsWpf/PythonNode.cs
```diff
@@ -157,6 +157,7 @@ private void EditScriptContent(object sender, EventArgs e)
         private void UpdateToPython2Engine(object sender, EventArgs e)
         {
             pythonNodeModel.Engine = PythonEngineVersion.IronPython2;
+            pythonNodeModel.OnNodeModified();
```
aparajit-pratap, line 160: Isn't this equivalent to calling `OnNodeModified` in the setter of the `Engine` property?

aparajit-pratap: That way 4 calls to `OnNodeModified` are reduced to one.

mjkkirschner: yes, please see the conversation above, by having the engine property setter always call `OnNodeModified()` there is no way to avoid multiple runs which are triggered by setting the code content, since these also trigger executions - and it is difficult to ensure the engine run is actually executing with valid code (most updated).

This implementation has a bit more overhead but aligns `script` and `engine` setters, they now both require calling `NodeModified` and let's us control when to execute more precisely.

QilongTang: @aparajit-pratap This is purposefully written to avoid multiple runs of the same python code.

PR #10782, opened 2020-06-18
Add trace back for pythonnet errors

mmisol:
### Purpose

When errors occur in Python.NET, a PythonException is thrown. This
class, unfortunately, does not put the trace back information in the
message. It is available in the stack trace, but showing that did not
seem a good idea. Instead, reflection is used to access the private
field that holds the trace back. The trace back, if available, is
appended to the message of a new exception that wraps the original.

### Declarations

### Reviewers

@DynamoDS/dynamo


src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -38,55 +40,78 @@ public static class CPythonEvaluator
...
+        private static string GetTraceBack(Exception e)
+        {
+            var pythonExc = e as PythonException;
+            if (!(e is PythonException))
+            {
+                return null;
+            }
+
+            // Return the value of the trace back field (private)
+            return typeof(PythonException).GetField("_tb", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(pythonExc) as string;
```
mjkkirschner, line 114: should this be wrapped in its own try catch?

mmisol: I thought about it, but decided not to do that. Whenever that happens it would be because we have updated Python.NET and that fields no longer exists. Should we try to hide that under the rug or let it show up so that we know we need to update this code?

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -1,10 +1,12 @@
...
+using DynamoUtilities;
+using Python.Runtime;
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
-using Autodesk.DesignScript.Runtime;
-using Dynamo.Utilities;
-using Python.Runtime;
+using System.Reflection;
```
QilongTang, line 9: @marimano Can you check your our coding standards and adjust your VS setting? https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards

src/Libraries/DSCPython/CPythonEvaluator.cs
mjkkirschner: we should definitely not crash....
I can think of a few options
* in the python node send a message asking the user to report
* send an event to analytics with an error which we can look for
* open something like a VM crash dialog - dynamo doesn't actually crash, but theres a message for the user to post the given log to github.

src/DynamoUtilities/DynamoException.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace DynamoUtilities
+{
+    /// <summary>
+    /// Represents an error raised by Dynamo
+    /// </summary>
+    [Serializable]
+    public class DynamoException : Exception
```
QilongTang, line 13: Do we want to be more specific here? I am not sure if this is going to be more widely used, seems that was what you were referring to in the earlier PRs, if that is the case, I would vote for making this one particularly about Python, and we create other DynamoExceptions going forward. But I am actually not sure if @DynamoDS/dynamo team agreed where we are heading. Let's discuss this within team some time

src/DynamoUtilities/DynamoException.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
```
QilongTang, line 5: I think these are just default ones, please remove the unused ones

src/DynamoUtilities/DynamoException.cs
mmisol: The creation of this class was something I saw could bring up a discussion. The problem is this:
- `PythonException` is a Python.NET class. We should not throw these exceptions.
- `Exception` is considered a smell by SQ.
- `ApplicationException` should not be used according to MS.
- We don't have a base exception, that I know of, that says `this error comes from Dynamo`.

This comes from a recommendation I have followed in other projects. You have a base exception like this and if you need specific ones you inherit from this one.

QilongTang: Understood. I am supportive but I would like some consensus from team. If we agree, in this case I would be in favor of an inherit `DynamoPythonException` or something similar.

src/DynamoUtilities/DynamoException.cs
mmisol: Does SQ generate a smell for keeping the default using? That's even worse than FxCop :)

QilongTang: @mmisol They do :)

src/Libraries/DSCPython/CPythonEvaluator.cs
mmisol: I don't think an exception at this point would make the application crash. I said show up not blow up :)
Let me give it try and I'll attach a screenshot of how it would eventually look like.

mmisol: Well, it does not crash, but the node just gives a generic `Object reference not set to an instance of an object.` warning. I guess I'll throw a `DynamoException` with a special message to contact us :)

src/DynamoUtilities/DynamoException.cs
aparajit-pratap: I vote with Aaron about making it specific to Python.

mmisol: I vote against, but I think we should probably not have this discussion here. Let's keep this PR open until we can meet to discuss.

aparajit-pratap: Are you proposing changing all existing Dynamo exceptions and make them inherit from this one? That doesn't make sense to me.

mjkkirschner:
@mmisol I wanted to say, thank you for the sleuthing here, I think this on the whole a great improvement for users with minimal maintenance cost for us, awesome!

src/DynamoUtilities/DynamoException.cs
mmisol: I'm against the proposal of subclassing the base exception because we don't need to handle these exceptions in any particular way. The base type should suffice.

mjkkirschner:
@mmisol sorry I caused a merge conflict for you!

mmisol:
No problem @mjkkirschner . I think I got the merge ok, but you can take a look just in case.

PR #10792, opened 2020-06-22
Add support for alternate resource names

mmisol:
### Purpose

The alternate name supported is the same as the original resource name,
but adds the culture name before the extension.

This is needed to support the workflow the localization team is using
for translations. The tool they are using produces resources with names
that do not match the name of the original embedded resource.

Example: German translation of NoContent.html will be searched in de-DE
satellite assembly by the names of:
1. NoContent.html
2. NoContent.de-DE.html

Unfortunately, I couldn't add a unit test for this, because it requires to use the tool the localization team is using to generate a test satellite assembly. I did manual test with a fictitious translation of German they did and was able to see the localized content after these changes. Here is a screenshot:
Screen Shot 2020-06-22 at 1 02 44 PM
 

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
+++ b/src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
@@ -345,7 +345,7 @@ namespace Dynamo.DocumentationBrowser
         /// </summary>
         /// <param name="name">Resource name</param>
         /// <returns>Resource name with the culture name appended before the extension</returns>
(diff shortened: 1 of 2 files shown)


mjkkirschner:
@mmisol it looks good - what about just unit testing
`GetResourceNameWithCultureName` function?

mmisol:
Tests passed here https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/676/

PR #10793, opened 2020-06-22
Cherrypicking https://github.com/DynamoDS/Dynamo/pull/10779 into master.

reddyashish:
### Purpose

This PR is to cherrypick the commit that was merged to 2.7. It would fix a regression that was caused by installing the package from workspace dependency viewer. 

### Declarations

### Reviewers

@mjkkirschner @QilongTang


test/DynamoPackagesUITests/DynamoPackagesUITests.cs
```diff
@@ -1,189 +0,0 @@
...
-using System.Reflection;
-using Moq;
-using Dynamo.DynamoPackagesUI.ViewModels;
-using Newtonsoft.Json;
-using Dynamo.Wpf.Extensions;
-
-namespace DynamoPackagesUITests
-{
-    [TestFixture]
-    public class DynamoPackagesUITests : SystemTestBase
```
mmisol, line 24: I saw these tests were removed in the original PR, stating they referred to the "old PackageManager". Can you extend on that?

reddyashish: I am working on the test and will add it in a different PR. There are some issues related to GregClient while installing package from tests. 
The deleted tests are not being built at all and they were using a different PM client which we moved away from.

PR #10796, opened 2020-06-23
Graph coverage

byorda-glb:
### Purpose

Increase the coverage of the NodeModel class

### Declarations

### Reviewers

Aaron Tang
Michael Kirschner

### FYIs

Alfredo Pozo


test/DynamoCoreTests/Graph/Nodes/NodeModelTests.cs
```diff
@@ -0,0 +1,70 @@
...
+            var inPorts = new ObservableCollection<PortModel>() { input };
+            testNodeModel.InPorts = inPorts;
+
+            Assert.Contains(input, testNodeModel.InPorts);
+
+            var output = new PortModel(PortType.Output, testNodeModel, new PortData("output A", "This is output A."));
+            var outPorts = new ObservableCollection<PortModel>() { output };
+            testNodeModel.OutPorts = outPorts;
+
+            Assert.Contains(output, testNodeModel.OutPorts);
```
aparajit-pratap, line 47: This doesn't look like a very useful test.
All you're doing is adding something to a collection and then checking if the collection contains that thing :)

byorda-glb: What I'm trying to do here is to cover the setter for the InPorts and OutPorts properties. 

Since these are both collections, naturally every manipulation of them in the project is either adding or removing elements, but never replacing the entire collection by a new one like I'm doing there.

How do you think we could write a more useful test for this scenario?

aparajit-pratap: It looks like there is no place in the code base where these properties are being assigned directly, which is why the setter isn't covered. Nor do I think does the `RaisePropertyChangedEvent` do anything. I think this is a candidate for dead code removal?

byorda-glb: That sounds like a better approach. Should I go ahead and remove the setters for these two properties together with the test?

aparajit-pratap: I don't think you can remove them right away as they are public setters.

byorda-glb: Should I remove that test?

aparajit-pratap: I think you can apply the `IsObsolete` custom attribute to the setter with a message saying it will be deprecated in Dynamo 3.0. Yes, I also feel the test is not necessary.

byorda-glb: Alright! I'll add the attribute and remove the test.

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -369,6 +369,7 @@ public string Name
         ///     Collection of PortModels representing all Input ports.
         /// </summary>
         [JsonProperty("Inputs")]
+        [Obsolete("Property will be deprecated in Dynamo 3.0")]
```
aparajit-pratap, line 372: I don't think we can apply this to the setter specifically, which is why I suggested using our custom `IsObsolete` attribute only on the setter. I don't think we want to obsolete the getter.

PR #10797, opened 2020-06-23
python should evaluate current code.

mjkkirschner:
### Purpose

please see:
https://github.com/DynamoDS/Dynamo/pull/10781

this PR adds a few dispatcherUtils calls which are likely not 100% required, but more importantly it avoids an **access violation** that the CPythonEvaluator throws when passing null code to pythonNet.

### Declarations


src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -38,55 +38,59 @@ public static class CPythonEvaluator
             IList bindingNames,
             [ArbitraryDimensionArrayImport] IList bindingValues)
         {
-           
-                Python.Included.Installer.SetupPython().Wait();
+            if (code == null)
```
mjkkirschner, line 41: @mmisol this is the change.

mjkkirschner:
passing self serve:
https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/678/

QilongTang:
Looks good!

PR #10799, opened 2020-06-23
Use Python.NET and Python.Included as externs

mmisol:
### Purpose

This will allow us to use custom builds of Python.NET for our CPython
engine. Projects now reference the binaries in 'extern' folder rather
than the published NuGets.

### Declarations

### Reviewers

@DynamoDS/dynamo


src/Libraries/PythonNodeModels/PythonNodeModels.csproj
```diff
@@ -39,13 +39,13 @@
       <HintPath>..\..\packages\Newtonsoft.Json.8.0.3\lib\net45\Newtonsoft.Json.dll</HintPath>
       <Private>False</Private>
     </Reference>
-    <Reference Include="Python.Included, Version=3.7.3.4, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\Python.Included.3.7.3.4\lib\netstandard2.0\Python.Included.dll</HintPath>
-      <Private>False</Private>
```
mjkkirschner, line 44: please retain the copyLocal / private elements set to False where they were previously or we'll get multiple copies of these dlls potentially pulled into various output folders.

mmisol:
I had to retry the SelfServe job so I guess it got confused. The build passed and it is https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/805/

mmisol:
https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/682/ also passed

mmisol:
@mjkkirschner Can I ask you to take another look? The issues you pointed out should have been addressed. Thanks

test/Libraries/DynamoPythonTests/PythonEvalTests.cs
```diff
@@ -165,7 +165,7 @@ public void CPythonEngineIncludesTraceBack()
             catch (Exception exc)
             {
                 // Trace back is not available for this call, but we still get a reasonable message back
-                Assert.AreEqual(@"SyntaxError : ('invalid syntax', ('<string>', 3, 13, ""print 'hello'\n""))", exc.Message);
+                Assert.AreEqual(@"SyntaxError : ('invalid syntax', ('<string>', 3, 7, ""print 'hello'\n""))", exc.Message);
```
mjkkirschner, line 168: why did this change?

mmisol: That's a good question. That's the type of error where the traceback is built-in. My guess is that was a result of the change in the version of Python, from 3.7 to 3.8, but I couldn't find it in the release notes. The column number changed to point to the beginning of the section where the invalid syntax was detected, so technically this seems more correct.

mmisol: Maybe this https://bugs.python.org/issue38535?

PR #10819, opened 2020-06-29
copy missing files from 2.6.1 to their respective folders after build.

mjkkirschner:
### Purpose

This PR deals with the binaries which were removed from the 2.7 distribution relative to 2.6.1. It adds empty files renamed to to `file.dll` for each file and after DynamoCore's build copies them to the correct folder.

I went with empty files to avoid potentially loading types twice from different locations, which can cause very strange runtime failures in .net.

I've done some limited testing, and will do some more using the modules tab in VS -  but @smangarole we should run some manual checks to make sure nothing strange occurs. Particularly IronPython, Excel, and 3dGraphics.

### Declarations

- [ ] cherry pick to 2.7

### DIFF
--- a/extern/legacy_remove_me/readme.md
+++ b/extern/legacy_remove_me/readme.md
@@ -1,0 +1,21 @@
+### What is this folder?
+
+This folder contains binaries that we want to remove from dynamo, but at the current time cannot - without causing issues for Dynamo hosts.
+
+These files are actually empty .txt files renamed .dll to avoid including any duplicate types which may be accidentally loaded.
+
+* \bin\Assimp32.dll, 
+* \bin\Assimp64.dll,
+* \bin\AssimpNet.dll
+* \bin\Office.dll
+* \bin\SharpDX.Direct3D11.Effects.dll
+* \bin\SharpDX.Toolkit.Compiler.dll
+* \bin\SharpDX.Toolkit.dll
+* \bin\SharpDX.Toolkit.Graphics.dll
+* \bin\stdole.dll
+* \bin\en-US\DynamoPython.resources.dll
+* \bin\nodes\DSIronPython.dll
+
+TODO 
+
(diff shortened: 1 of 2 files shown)


extern/legacy_remove_me/readme.md
```diff
@@ -0,0 +1,17 @@
...
+* \bin\Assimp32.dll, 
+* \bin\Assimp64.dll,
+* \bin\AssimpNet.dll
+* \bin\Office.dll
+* \bin\SharpDX.Direct3D11.Effects.dll
+* \bin\SharpDX.Toolkit.Compiler.dll
+* \bin\SharpDX.Toolkit.dll
+* \bin\SharpDX.Toolkit.Graphics.dll
+* \bin\stdole.dll
+* \bin\en-US\DynamoPython.resources.dll
```
QilongTang, line 16: Ah.. Totally forgot about localization resources, hopefully empty binary works on their pipeline as well

mjkkirschner: yeah, I am going to test this some more, worst case is to add an empty resx to the build... easy, but I want to try keeping everything consistent for now if possible.

mmisol: What is/was DynamoPython? Does the resource no longer exist but the main assembly still exists?

mjkkirschner: yeah, the project remains but is emtpy, I am not sure what it was... we'll get rid of it as soon as we can.

mjkkirschner:
@mmisol sounds good, will remove.

extern/legacy_remove_me/readme.md
mmisol: Ok. Feel free to leave it if it's there. Or if you would like to delete it and replace it with one of these fake ones, you can also do that. Up to you.

PR #10827, opened 2020-06-30
Python.NET binaries update (PR3)

mmisol:
### Purpose

Updates Python.NET binaries to PR3

### Declarations

### Reviewers

@DynamoDS/dynamo


mjkkirschner:
Did you want to include the tests you mentioned in a followup?

mmisol:
Yes. This actually includes changes from 2 different PRs, so it didn't make much sense to add that too. You could say this is the automatic LibG binaries update, only that it's manual and for Python.NET.

QilongTang:
@mmisol Curious if this is a release build? Just sanity checking

mmisol:
@QilongTang Yes, but thanks for checking. The debug binary weights around 200kb, which is around 30kb more than the release build, so that's a possible way to tell.

PR #10830, opened 2020-06-30
Add support for BigInteger in CPython engine

mmisol:
### Purpose

Support is added for encoding decoding BigInteger values from Python
int that have a size that exceeds what can fit in a long.

Also a BigInteger encoder/decoder is added to the custom encoders of
Python.NET in order to support encoding and decoding in the context of
function calls.

### Declarations

### Reviewers

@DynamoDS/dynamo


mjkkirschner:
@mmisol 😢  
```
(CoreCompile target) -> 
  Encoders\BigIntegerEncoder.cs(23,25): error CS8107: Feature 'default literal' is not available in C# 7.0. Please use language version 7.1 or greater. [C:\projects\dynamo\src\Libraries\DSCPython\DSCPython.csproj]
    2057 Warning(s)
    1 Error(s)
Time Elapsed 00:01:17.13

```

maybe time to retire this build and use our github action instead?

mmisol:
Yeah. I saw that appveyor thing probably needs an update, but it's not even ours! Probably it is a good time to retire it. I read we should just remove the web hook and that's it.

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -101,6 +106,17 @@ public static class CPythonEvaluator
             }
         }
 
+        /// <summary>
+        /// Registers custom encoders and decoders with the Python.NET runtime
+        /// </summary>
+        private static void InitializeEncoders()
+        {
+            var encoders = new IPyObjectEncoder[] { new BigIntegerEncoder() };
```
mjkkirschner, line 114: @aparajit-pratap these are the encoders I mentioned before.

src/Libraries/DSCPython/Encoders/BigIntegerEncoder.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using System.Numerics;
+using Python.Runtime;
+
+namespace DSCPython.Encoders
+{
+    internal class BigIntegerEncoder : IPyObjectEncoder, IPyObjectDecoder
```
mjkkirschner, line 7: thank you for making this internal!
🎉

test/Libraries/DynamoPythonTests/PythonEvalTestsWithLibraries.cs
```diff
@@ -0,0 +1,65 @@
...
+using NUnit.Framework;
+using static DSIronPythonTests.IronPythonTests;
+
+namespace DynamoPythonTests
+{
+    public class PythonEvalTestsWithLibraries : DynamoModelTestBase
+    {
+        protected override void GetLibrariesToPreload(List<string> libraries)
+        {
+            libraries.Add("ProtoGeometry.dll");
```
mjkkirschner, line 19: why do we need protogeo for this test?

test/Libraries/DynamoPythonTests/IronPythonTests.csproj
```diff
@@ -117,6 +119,10 @@
       <Name>PythonNodeModels</Name>
       <Private>False</Private>
     </ProjectReference>
+    <ProjectReference Include="..\..\..\src\Tools\DynamoShapeManager\DynamoShapeManager.csproj">
+      <Project>{263fa9c1-f81e-4a8e-95e0-8cdae20f177b}</Project>
+      <Name>DynamoShapeManager</Name>
```
mjkkirschner, line 124: just curious about the requirement for these dependencies.

test/Libraries/DynamoPythonTests/PythonEvalTestsWithLibraries.cs
mmisol: We don't actually. I just figured this may come in handy for the tests we would add.

aparajit-pratap:
Why are we spending a whole bunch of time with BigIntegers. They just happen to be supported by IronPython but not a big use case for zero touch nodes or even python nodes for that matter.

test/Libraries/DynamoPythonTests/IronPythonTests.csproj
mmisol: So, when you inherit from `DynamoModelTestBase` and want to customize `CreateStartConfiguration` you pass some parameters that will complain because you don't have these.

mjkkirschner: ahh, okay, generally I would say avoid loading libraries that we don't need during the test as these types will be imported into the VM over and over again for each test which can be kind of slow. So if we write 1000 ironPython tests we'll load all the geometry types into the VM foreach test.

src/Libraries/DSCPython/CPythonEvaluator.cs
aparajit-pratap: Was the biginteger encoder/decoder added by you? I may have missed it in your PR's in Python.NET repo.

mmisol: `BigIntegerEncoder` is part of this submission. The API to add them already existed. One thing I did add in Python.NET is the function `PyLong.ToBigInteger`, which is used by the decoder.

src/Libraries/DSCPython/Encoders/BigIntegerEncoder.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            if (!PyLong.IsLongType(pyObj))
+            {
+                value = default;
+                return false;
+            }
+
+            using (var pyLong = PyLong.AsLong(pyObj))
+            {
+                value = (T)(object)pyLong.ToBigInteger();
```
mjkkirschner, line 29: staring at this I'm assuming that PythonNet will only ever call this method when T is `bigInteger` ?

mmisol: Yes. I think there is something strange with the definition of the interface because it's not a generic one but one of its methods is generic. I didn't know how to handle that exactly but I took this from an example they have in a test.

mmisol:
Hi @mjkkirschner @aparajit-pratap . If there is nothing else missing here, can I ask for your approval? Thanks

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -177,30 +193,44 @@ public static DataMarshaler OutputMarshaler
...
                             else if (PyLong.IsLongType(pyObj))
                             {
-                                return PyLong.AsLong(pyObj).ToInt64();
+                                using (var pyLong = PyLong.AsLong(pyObj))
+                                {
+                                    try
+                                    {
+                                        return pyLong.ToInt64();
+                                    }
+                                    catch (PythonException exc) when (exc.Message.StartsWith("OverflowError"))
```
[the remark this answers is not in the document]
mmisol: These are called exception filters. They allow to catch an exception type conditionally to the fact they also match the expression given in `when`. They were introduced in C# 6 https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-6#exception-filters

PR #10834, opened 2020-07-01
IronPython2 node in Dynamo Home Workspace should not raise a python notification

reddyashish:
### Purpose

The node added event is subscribed to this PythonMigrationViewExtension class and it raises a notification for the first time. After that the Home WS id is stored in NotificationTracker dictionary and will not raise it again. 

I am not sure in what case, do we raise a notification when a Python script node is added to the workspace. I couldn't find any other scenarios where we were doing it apart from the bug that was filed. Let me know if I am missing anything here. 

We also have 2 tests WillLogNotificationWhenAddingAnIronPythonNode and WillOnlyLogNotificationWhenAddingAnIronPythonNodeOnce present in our code which are similar to the bug we filed. Should these be modified to check that no notification is raised?

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/PythonMigrationViewExtension/PythonMigrationViewExtension.cs
+++ b/src/PythonMigrationViewExtension/PythonMigrationViewExtension.cs
@@ -102,14 +102,12 @@ namespace Dynamo.PythonMigration
         private void SubscribeToDynamoEvents()
         {
             LoadedParams.CurrentWorkspaceChanged += OnCurrentWorkspaceChanged;
-            DynamoViewModel.CurrentSpaceViewModel.Model.NodeAdded += OnNodeAdded;
             DynamoViewModel.Model.Logger.NotificationLogged += OnNotificationLogged;
         }
 
         private void UnsubscribeFromDynamoEvents()
         {
             LoadedParams.CurrentWorkspaceChanged -= OnCurrentWorkspaceChanged;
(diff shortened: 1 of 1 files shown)


mjkkirschner:
@reddyashish ah - thanks for looking into this - I guess the idea was to notify users who are building graphs that they should try moving up to python3 given the EOL of Python2, and not reserve this warning only for users *opening* existing graphs.

It does make sense to dissuade creation of new content with the older python engine I guess.

I think what caused it to seem like a bug was the fact that it's only added for that first python node.

I'm leaning on the side of leaving it, but not sure how we could improve it to be more useful.

@Amoursol any thoughts on this?

QilongTang:
@mjkkirschner @Amoursol Can we revisit this? I can see good and bad for either way we touch it. I think the current way might be user friendly but actually inconsistent with other IDE-ish env. I think a lot of IDEs will shoot for accuracy and the ability of navigation to the the place causing warnings. If we do the same but do not provide a way for users to easily fix them, that may be also less ideal.. I vote for leave it for now and gather user feedback early

Amoursol:
I concur with @QilongTang - lets leave it for now, gather feedback and see how users expect it to behave via daily builds and get it amended before 2.8.

PR #10836, opened 2020-07-01
Add PackagePaths as Parameter of ExecutionSession

mjkkirschner:
### Purpose

Try to address https://github.com/DynamoDS/DynamoWishlist/issues/71
* add parameter key for packagePaths
* add parameter
* add test
* add python test asserting we can reference and use this parameter.

Screen Shot 2020-07-01 at 1 20 38 PM

### Declarations

### DIFF
--- a/src/NodeServices/ExecutionSession.cs
+++ b/src/NodeServices/ExecutionSession.cs
@@ -69,5 +69,11 @@ namespace Dynamo.Session
         /// The duration of an execution covered by an <see cref="IExecutionSession"/>
         /// </summary>
         public static readonly string LastExecutionDuration = "LastExecutionDuration";
+
+        /// <summary>
+        /// The current collection of packagepaths that Dynamo is loading packages from.
+        /// The Return value is IEnumerable
+        /// </summary>
+        public static readonly string PackagePaths = nameof(PackagePaths);
     }
 }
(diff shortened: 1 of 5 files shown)


mjkkirschner:
all tests pass on self serve, but parallel has a concurrency issue with file load - will take a look.

mjkkirschner:
I ran the parallel self serve again and all passed - so I think this is good to review now:
https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/837

src/NodeServices/ExecutionSession.cs
```diff
@@ -69,5 +69,11 @@ public class ParameterKeys
         /// The duration of an execution covered by an <see cref="IExecutionSession"/>
```
aparajit-pratap, line 69: I think we should make this class static as well.

mjkkirschner: I agree, but I think that would sadly break the API. Do you prefer I obsolete this class, create a static version and use that one?

mjkkirschner: or just risk the change?

aparajit-pratap: Can we add a TODO to add it later?

mjkkirschner: yep, will do.

mjkkirschner: added and task created.

PR #10840, opened 2020-07-02
Add tests for conversions in Python method calls

mmisol:
### Purpose

Adds tests for the following conversion scenarios:
- Python list => .NET IList (CPython fails)
- .NET array => Python list
- Python tuple => .NET array
- Python tuple => .NET IList (CPython fails)
- Python range => .NET array
- Python range => .NET IList (CPython fails)
- Python dict => .NET IDictionary (CPython fails)
- .NET IDictionary => Python dict (CPython unkonwn)

Also commented out some tests that fail in both engines but may be
considered to fix in CPython:
- .NET IList => Python list (call a Python list method on an IList?)
- Python array => .NET IList (builtin library, may be uncommon) 
- Python dict => DS Dictionary (may deserve special attention)

Other types considered but not added as tests:
- classes (can't convert them)
- dictionary view objects (uncommon)
- bytes, bytearray, memoryview (uncommon)
- set, frozenset (uncommon)
- complex (uncommon)

Other scenarios that might be interesting:
- Mutability of collections (IronPython supports this for a IList)

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/test/Libraries/DynamoPythonTests/PythonEvalTestsWithLibraries.cs
+++ b/test/Libraries/DynamoPythonTests/PythonEvalTestsWithLibraries.cs
@@ -12,6 +12,7 @@ namespace DynamoPythonTests
         protected override void GetLibrariesToPreload(List<string> libraries)
         {
             libraries.Add("FFITarget.dll");
+            libraries.Add("DSCoreNodes.dll");
         }
 
         public IEnumerable<PythonEvaluatorDelegate> Evaluators = new List<PythonEvaluatorDelegate> {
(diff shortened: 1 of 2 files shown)


aparajit-pratap:
Are these supported in IronPython:
> * Python tuple => .NET array
> * Python tuple => .NET IList (CPython fails)

mmisol:
@aparajit-pratap Yes. Where I wrote `CPython fails` it should be interpreted as that  an also `IronPython works`. Also, where I didn't write anything is for cases where they both work.

test/Libraries/DynamoPythonTests/PythonEvalTestsWithLibraries.cs
```diff
@@ -43,5 +44,164 @@ import clr
...
+            string code = @"
+import sys
+import clr
+clr.AddReference('FFITarget')
+clr.AddReference('DesignScriptBuiltin')
+from DesignScript.Builtin import Dictionary
+from FFITarget import DummyCollection
+
+d = {'one': 1, 'two': 2, 'three': 3}
+# Python dict => DS Dictionary - Does not work in either engine
```
aparajit-pratap, line 185: I'm not sure if this is intended. It would be good to have but if it were to be supported, we would most likely need to make custom changes in pythonnet for interop with DS that would not be accepted by the main repo for back-porting and we would need to continue to maintain our own custom fork.

aparajit-pratap: We didn't do it for IronPython either.

test/Libraries/DynamoPythonTests/PythonEvalTestsWithLibraries.cs
```diff
@@ -43,5 +44,164 @@ import clr
...
+import clr
+clr.AddReference('FFITarget')
+clr.AddReference('DSCoreNodes')
+from FFITarget import DummyCollection
+from DSCore import List
+
+r = range(0, 10, 2)
+# Python range => .NET array - Works in both
+a = DummyCollection.MakeArray(r)
+# Python range => .NET IList - Does not work in CPython
```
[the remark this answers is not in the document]
mmisol: Not problem at all. These are very good questions @aparajit-pratap .

> Are Python ranges mutable? 

What happens when a Python primitive gets passed to a .NET function call, is that the primitive gets converted to a .NET object. This means that in practice they are not mutable because what we get is not even in the Python world.

> Do they result in lists or arrays?

This will depend on the declared type in the function call. Support for arrays was already built into Python.NET. For lists, a few minor changes were needed, but most of the logic can be reused.

> Could there by differences in IronPython vs. CPython in the way that tuples, ranges, etc. behave in terms of mutability?

I don't think I have tested that. That's a very good point. I'll take a look at IronPython and see if I'm able to mutate a list for instance.

test/Libraries/DynamoPythonTests/PythonEvalTestsWithLibraries.cs
mmisol: Thanks for the feedback. I thought of a way to do this without referencing DS Dictionary in Python.NET. What we would do is a two step conversion:
1. Convert it to a .NET Dictionary, using primitives from Python.NET
2. Convert the .NET Dictionary to a DS Dictionary, as part of a custom encoder in Dynamo

test/Libraries/DynamoPythonTests/PythonEvalTestsWithLibraries.cs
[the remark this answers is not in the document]
mmisol: I just checked and mutability can be achieved in IronPython for a List. You need to use a Zero Touch node to test it though, because List methods always make a copy.

aparajit-pratap:
@mmisol thanks for answering my questions and for the detailed enumeration of all possible conversions and limitations.

mmisol:
I'll go ahead and merge this. I'll link the task to the PR so that we can take a look whenever we see which of these scenarios we would like to address.

PR #10841, opened 2020-07-03
Improve the truncated tooltip when the node description is too long

ZiyunShang:
### Purpose

Improve the truncated tooltip
https://jira.autodesk.com/browse/REVIT-164503 

### Declarations

### Reviewers

@QilongTang @mjkkirschner 

### FYIs

Before improvement:
image

After improvement:
#### NodeWithNoIcon
image

#### NodeWithIcon
image

### DIFF
--- a/src/DynamoCoreWpf/Controls/TooltipWindow.xaml
+++ b/src/DynamoCoreWpf/Controls/TooltipWindow.xaml
@@ -25,7 +25,7 @@
             Padding="10">
         <Grid Width="325">
             <Grid.RowDefinitions>
-                <RowDefinition Height="96"></RowDefinition>
+                <RowDefinition Height="Auto"></RowDefinition>
                 <RowDefinition Height="Auto"></RowDefinition>
                 <RowDefinition Height="Auto"></RowDefinition>
                 <RowDefinition Height="Auto"></RowDefinition>


mjkkirschner:
@ZiyunShang I think this is a good improvement thanks! - I can't recall, but if I had to guess the reason we didn't do this originally, was that it creates an inconsistent experience when hovering over a bunch of library items.

Ideally we would find way to enable the working scroll interaction - maybe we could just slide the panel over by 2px or whatever that spacing is so that the tooltip does not close.

This also does not solve the issue where there is a very long tooltip on a low resolution screen.

@Amoursol @QilongTang  any thoughts on this one?

QilongTang:
I am supportive of this change as long as we keep the width. If you can make the hover over tooltip also display when mouse off library entry, that might be ideal, but I am not sure how much code that would be

Amoursol:
Only comment here to your point @mjkkirschner is that it's difficult to get into those hover states sometimes in Dynamo. If the 2px change helps this then awesome. My suggestion is that we want to retain both _consistency_ (Huge for our users understanding) and _clarity_ - happy with any approach that ticks these boxes. These changes seem like a huge boost for clarity!

QilongTang:
@ZiyunShang  Discussed with @mjkkirschner and @Amoursol , will merge to unblock user for now. We will create a testing task in our backlog to test this with a few packages.

PR #10845, opened 2020-07-03
Prevent SO for unknown type in CPython

mmisol:
### Purpose

When unmarhsalling output data from a Python script using the CPython
engine, a StackOverlowException would occur for certain uncommon types
like Python set. The unmarshaller is trying to recursively convert the
output by calling AsManagedObject, but in cases where Python.NET does
not have a specialized representation, like it is the case for set, it
just returns a different object pointing to the same handle, thus
causing the unmarshaller to call himself again with basically the same
underlying Python object.

The problem is fixed by explicitly checking if the returned object by
AsManagedObject has the same handle as the original one, in which case
an exception is thrown with a proper message.

### Declarations

### Reviewers

@DynamoDS/dynamo 

### FYIs

@Amoursol This should better handle the bug you found, by providing a warning instead of a crash.

### DIFF
--- a/src/Libraries/DSCPython/CPythonEvaluator.cs
+++ b/src/Libraries/DSCPython/CPythonEvaluator.cs
@@ -249,10 +249,8 @@ namespace DSCPython
                                     }
                                 }
                             }
-                            else
-                            {
-                                return outputMarshaler.Marshal(unmarshalled);
-                            }
+
+                            return outputMarshaler.Marshal(unmarshalled);
                         });
                 }
                 return outputMarshaler;
--- a/src/Libraries/DSCPython/CPythonEvaluator.cs
+++ b/src/Libraries/DSCPython/CPythonEvaluator.cs
@@ -251,7 +251,7 @@ namespace DSCPython
                             }
                             else
                             {
-                                return unmarshalled;
+                                return outputMarshaler.Marshal(unmarshalled);
                             }
                         });
                 }
--- a/src/Libraries/DSCPython/CPythonEvaluator.cs
+++ b/src/Libraries/DSCPython/CPythonEvaluator.cs
@@ -233,7 +233,26 @@ namespace DSCPython
                                 }
                             }
 
-                            return outputMarshaler.Marshal(pyObj.AsManagedObject(typeof(object)));
+                            var unmarshalled = pyObj.AsManagedObject(typeof(object));
+                            if (unmarshalled is PyObject)
+                            {
+                                using (unmarshalled as PyObject)
+                                {
+                                    if (unmarshalled.Equals(pyObj))
+                                    {
+                                        // Object can't be unmarshalled. Prevent a stack overflow.
+                                        throw new InvalidOperationException(Properties.Resources.FailedToUnmarshalOutput);
+                                    }
+                                    else
+                                    {
+                                        return outputMarshaler.Marshal(unmarshalled);
+                                    }
+                                }
+                            }
+                            else
+                            {
+                                return unmarshalled;
+                            }
                         });
                 }
                 return outputMarshaler;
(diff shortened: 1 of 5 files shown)


QilongTang:
Looks like there are a bunch of falures?

mmisol:
@QilongTang I took a look but only found one test failure related to my changes, which I just fixed. No idea what the others are.

mmisol:
I just run those locally and they work just fine. Maybe something broke on the build machine?

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -233,7 +233,26 @@ public static DataMarshaler OutputMarshaler
                                 }
                             }
 
-                            return outputMarshaler.Marshal(pyObj.AsManagedObject(typeof(object)));
+                            var unmarshalled = pyObj.AsManagedObject(typeof(object));
```
aparajit-pratap, line 236: Can we just have the using here:
`using (var unmarshalled = pyObj.AsManagedObject(typeof(object)))`
and remove the one in line 239?

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -233,7 +233,26 @@ public static DataMarshaler OutputMarshaler
...
+                            if (unmarshalled is PyObject)
+                            {
+                                using (unmarshalled as PyObject)
+                                {
+                                    if (unmarshalled.Equals(pyObj))
+                                    {
+                                        // Object can't be unmarshalled. Prevent a stack overflow.
+                                        throw new InvalidOperationException(Properties.Resources.FailedToUnmarshalOutput);
+                                    }
+                                    else
```
aparajit-pratap, line 246: I don't think we need this `else` or the `else` in line 252? We can simply have the `return` statement in line 254 outside of the `else`.

src/Libraries/DSCPython/CPythonEvaluator.cs
mmisol: No, we can't. The problem is that `AsManagedObject` returns `object` so that code wouldn't compile. We want to dispose it if what we get back is a `PyObject`, but that's not always the case.

src/Libraries/DSCPython/CPythonEvaluator.cs
mmisol: Sorry, I think I misunderstood. I removed the one in line in 252, but not this one.

Conversion in line 248 happens inside the `using` statement context. The one on line 254 happens afterwards. After we dispose the variable, we shouldn't use it anymore.

test/Libraries/DynamoPythonTests/PythonEvalTests.cs
```diff
@@ -168,5 +168,32 @@ public void CPythonEngineIncludesTraceBack()
...
+        public void UnsupportedOutputShouldFailGracefullyInCPython()
+        {
+            var code = @"
+import weakref
+
+class myobj:
+  pass
+
+o = myobj()
+wr = weakref.ref(o)
```
aparajit-pratap, line 182: Could you explain what you're doing here in this test? Are you returning a weak reference to an object and was that resulting in a stack overflow exception earlier?

aparajit-pratap: Your PR description mentions it happens with a set?

mmisol: Yes, you are right. That deserves an explanation.

I used `weakref` in the test because I thought it was a type we would not likely add support for, unlike `set` which is more commonly used and we might decide to support.

I think the PR description is OK as it says `in cases where Python.NET does not have a specialized representation, like it is the case for set` so `set` is one of those types, but `weakref` is one too.

PR #10850, opened 2020-07-06
T1.1 Python 2 Code Migration

SHKnudsen:
### Purpose

This PR is the first step in adding a `Migration Assistant` that helps users automatically upgrade their Python code from Python 2 to Python 3. The PR focuses on adding the functionality to Migrate Python 2 code to Python 3 compatible code, using Pythons 2to3 library. It also adds a new VisualDifferenceViewer to show the changes being made by the migration.

This is the biggest of the PRs concerning the Python 3 migration work, the following PRs will be smaller.

image

### Declarations

### Reviewers

@mjkkirschner 
@QilongTang 
@radumg 

### FYIs

The `VisualDifferenceViewer` are using the `DiffPlex` library. I know the decision on whether or not we want to include this dependency is still pending, but the rest of the code should be fine to review.

The button icons on the `VisualDifferenceViewer` are also pending final confirmation, we can easily revert back to text if this turns out to be an issue.


QilongTang:
Thank you @SHKnudsen . This is exciting. @mjkkirschner will log a Jira task for our team to review your work soon

mjkkirschner:
https://jira.autodesk.com/browse/DYN-2888

test/DynamoCoreTests/PythonMigrationAssistantTests.cs
```diff
@@ -0,0 +1,54 @@
...
+            var expectedPython3CodeAfterMigration = @"for x in range(100): continue";
+
+            // Act
+            var migratedScript = ScriptMigrator.MigrateCode(originalPython2Code);
+            var migratedScript2 = ScriptMigrator.MigrateCode(migratedScript);
+
+            // Assert
+            Assert.AreEqual(expectedPython3CodeAfterMigration, migratedScript);
+            Assert.AreEqual(expectedPython3CodeAfterMigration, migratedScript2);
+        }
```
mmisol, line 52: By the name I understand the intention of this test, but I don't think it tests anything different than the previous test. We already know ScriptMigrator is calling a script and that it is stateless, so calling 2to3 many times does not seem a useful scenario.

test/DynamoCoreTests/PythonMigrationAssistantTests.cs
```diff
@@ -0,0 +1,54 @@
...
+    }
```
mmisol, line 53: I don't see any tests regarding special cases of `migrate_2to3.py` like the `bad input` retry or the use of a `fixer`. If unit tests just can't be added, it would be nice to at least count with better code documentation about them. For instance, where would we place a fixer if we wanted to use one?

src/PythonMigrationViewExtension/Controls/VisualDifferenceViewer.xaml
```diff
@@ -0,0 +1,107 @@
...
+                        Value="1" />
+                <Setter Property="VerticalAlignment"
+                        Value="Bottom" />
+            </Style>
+
+        </Grid.Resources>
+
+        <diffplex:DiffViewer x:Name="DiffView"
+                             OldTextHeader="Old"
+                             NewTextHeader="New"
```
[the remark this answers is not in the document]
Amoursol: Agree with @mmisol - please use `Before` and `After`

src/AssemblySharedInfoGenerator/AssemblySharedInfo.cs
```diff
@@ -64,4 +64,4 @@
 // You can specify all the values or you can default the Build and Revision Numbers 
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyFileVersion("2.8.0.1665")]
+[assembly: AssemblyFileVersion("2.8.0.1823")]
```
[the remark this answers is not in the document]
SHKnudsen: yea sorry about that, usually dont commit these. Let me know if it causes any issues

src/DynamoCoreWpf/UI/Themes/Modern/DynamoModern.xaml
```diff
@@ -758,6 +758,64 @@
...
+                                    Value="#373737" />
+                        </Trigger>
+
+                    </ControlTemplate.Triggers>
+
+                </ControlTemplate>
+            </Setter.Value>
+        </Setter>
+    </Style>
+
```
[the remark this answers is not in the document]
SHKnudsen: i have a new PR which fixes this and some additional things on the script editor. Wanted to keep this PR small so didn't included it in this, should have been more clear about this in the description.

src/PythonMigrationViewExtension/Controls/VisualDifferenceViewer.xaml
```diff
@@ -0,0 +1,107 @@
...
+                         Height="44">
+                    <Button.Resources>
+                        <fa:ImageAwesome x:Key="Shape"
+                                         Icon="Close"
+                                         Foreground="#bbbbbb"
+                                         Height="17"
+                                         Margin="13,0,13,0" />
+                    </Button.Resources>
+                </Button>
+            </StackPanel>
```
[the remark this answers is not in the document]
SHKnudsen: haha, no not in this PR. Next PR adds functionality for this, again was trying to make the PRs as small as possible.

src/PythonMigrationViewExtension/Controls/VisualDifferenceViewer.xaml
```diff
@@ -0,0 +1,107 @@
...
+
+        <diffplex:DiffViewer x:Name="DiffView"
+                             OldTextHeader="Old"
+                             NewTextHeader="New"
+                             FontFamily="Consolas"
+                             FontSize="10pt"
+                             Background="#272822"
+                             Foreground="#F8F8F2"
+                             HeaderForeground="#808080">
+        </diffplex:DiffViewer>
```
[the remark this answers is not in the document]
SHKnudsen: sorry @mmisol not sure i understand about the minimum resolution part?

It should only fit the code in the editor once pr view, if you change the view to it will refit the code to the window. Using text wrap is not an easy fix as the `DiffViewer` uses ScrollViewers for the content, i guess wrapping the text in scrollviewer isnt really possible and might not make sense. Let me know if you guys want to look into wrapping the text instead of using scrollviewers.

When making the window smaller the intent was that horizontal scrollbars should appear, i guess that this is the question of whether we want to wrap the text or have scrollbars.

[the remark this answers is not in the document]
Amoursol: Preference would be to match the existing Dynamo experience with wrapping - it's just a matter of effort. I agree @mmisol that if it's hard to do then we don't, but @SHKnudsen please have a look and see if it's an easy win.

src/PythonMigrationViewExtension/packages.config
```diff
@@ -0,0 +1,7 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="DiffPlex" version="1.6.3" targetFramework="net48" />
+  <package id="DiffPlex.Wpf" version="1.1.1" targetFramework="net48" />
```
mmisol, line 4: One thing I forgot to mention is that new external libraries that we add references to should be mentioned in the license files. We have [LICENSE.txt](https://github.com/DynamoDS/Dynamo/blob/master/LICENSE.txt) at the root and [License.rtf](https://github.com/DynamoDS/Dynamo/blob/master/doc/distrib/License.rtf) under doc/distrib. Sorry in advance about the RTF editing, but we are using it because it displays nicely in the About window.

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
```diff
@@ -167,6 +167,13 @@ private void OnRunClicked(object sender, RoutedEventArgs e)
             }
         }
 
+        private void OnMigrationAssistantClicked(object sender, RoutedEventArgs e)
+        {
+            if (nodeModel == null)
+                throw new NullReferenceException(nameof(nodeModel));
+
+            nodeModel.RequestCodeMigration(e);
+        }
```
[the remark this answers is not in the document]
QilongTang: Maybe refer to line 163 in this file

src/PythonMigrationViewExtension/packages.config
QilongTang: @mmisol Do we need to update both and keep them in sync or?

mmisol: @QilongTang Yes, we should add entries for the libraries in both.

test/DynamoCoreTests/PythonMigrationAssistantTests.cs
SHKnudsen: @mmisol regarding the `fixers` as it is now im using the default fixers that ships with `lib2to3`, so adding to them might be complicated this way. We could ship our own `fixers` with Dynamo in some central location so that adding new fixers would just be a case of adding them to that location. Not sure what your thoughts are on that?

I think that the default fixers should handle most cases.

mmisol: Thanks for the feedback @SHKnudsen . I don't think we have a story for it, but at some point we discussed the possibility to address some `Iron Python => Native Python` compatibility issues using a custom fixer. My feeling is that we probably don't require to have that ability as part of this PR, and could do something like what you suggest later if required. What do you think about this @Amoursol @mjkkirschner ?

Amoursol: I'll defer to @mjkkirschner on this one, but I agree @mmisol that we probably don't require it as part of this PR.

test/DynamoCoreTests/PythonMigrationAssistantTests.cs
radumg: chiming in with a bit of context @mmisol for this test.

Because the `2to3` migration is an external (py) component that we don't really control, we wanted to make sure the user's experience of migrating a script is `idempotent`. While not useful in itself, it's more to address the edge cases where users might apply the migration multiple times to the same code, either 
- by accident
- because script came from someone else and they're unsure if it's Python 2 or 3

It also confirms that the external migration script doesn't miss anything if only run once.

I'd suggest we keep this test for the reasons above, but up to you folks :)

mmisol: @radumg Since you mentioned idempotency let's get mathematical :) The second test shows that `x = f(x)` when `x` doesn't have Python 2 in it. The third tests shows that `y = f(f(y))` with a `y` with the same characteristics as `x` in the previous test. The second test is already showing idempotency, therefore my argument to delete this third test.

test/DynamoCoreTests/PythonMigrationAssistantTests.cs
mmisol: Brought this to standup the other day and I didn't get any remarks from Mike. Unless he comments otherwise, let's assume we don't need this for now, for the sake of unblocking this PR.

src/PythonMigrationViewExtension/packages.config
mmisol: @SHKnudsen Please address this.

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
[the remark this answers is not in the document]
mmisol: Please address this bug before merging @SHKnudsen

mjkkirschner:
Thanks for the reminder Martin - as it stands I would appreciate info on loading fixers if that’s something that was considered already, but if not I don’t want to hold up this pr for comments on a feature we may not use.

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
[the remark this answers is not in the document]
SHKnudsen: @mmisol added the `UpdateScript` before requesting the code migration as @QilongTang suggested, which seems to fix this.

src/PythonMigrationViewExtension/packages.config
SHKnudsen: @mmisol added these dependencies to both Licences files

SHKnudsen:
@mmisol addressed the last changes, sorry for the delay! Let me know if there is anything else.

mmisol:
Thanks @SHKnudsen . I see the bug is fixed. Could I ask you to place diffplex in the license.rtf file so that it appears in alphabetical order in the about window?
Screen Shot 2020-07-20 at 5 04 14 PM

SHKnudsen:
@mmisol order of the licence files are updated

mmisol:
I couldn't reproduce the test failure locally. I have triggered a manual test run on a different server just in case this is just a flaky failure. If that works I'm merging this.

mmisol:
Also got an error here https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/709/, although it seems unrelated (PackageDependencyTests). I think we can probably merge but I'll confirm with @QilongTang first.

test/DynamoCoreTests/PythonMigrationAssistantTests.cs
radumg: > @radumg let's get mathematical :)

haha, always love me some math🤓 

> @radumg The second test is already showing idempotency, therefore my argument to delete this third test.

You're totally right, hadn't looked at second test when I made the comment, thought we only had the normal migration test. Let's delete this one then since we all agree ? 😁

PR #10853, opened 2020-07-06
Annotation fix

Astul-Betizagasti:
### Purpose

Makes changes on how the annotation's area is calculated.

### Declarations

### Reviewers

Aaron Tang (@QilongTang )
Michael Kirschner (@mjkkirschner )

### FYIs

Alfredo Pozo (@alfredo-pozo )


QilongTang:
@Astul-Betizagasti Can you remind me the Jira task you created?

src/DynamoCore/Graph/Annotations/AnnotationModel.cs
```diff
@@ -358,20 +325,7 @@ internal void UpdateBoundaryFromSelection()
...
         /// <returns> the width and height of the last model </returns>
-        private Tuple<Double,Double> CalculateWidthAndHeight()
-        {           
-            var xgroup = Nodes.OrderBy(x => x.X).ToList();
-            var ygroup = Nodes.OrderBy(x => x.Y).ToList();
-            double yheight = ygroup.Last().Height;
 
-            //If the last model is Node, then increase the height so that 
-            //node border does not overlap with the group
-            if (ygroup.Last() is NodeModel)
```
mmisol, line 369: This part does not seem to be taken into account in the new solution, or maybe I'm missing it. Could `Nodes` contain something that is not a `NodeModel` (not a node)? What would it be?

Still maybe it should be fine to sum `ExtendYHeight` to give some extra space.

src/DynamoCore/Graph/Annotations/AnnotationModel.cs
```diff
@@ -305,43 +300,15 @@ internal void UpdateBoundaryFromSelection()
...
-                {
-                    if (!region.Contains(nodesList.Rect))
-                    {
-                        overlap = nodesList;
-                        if (overlap.Rect.Top < this.X ||
-                                    overlap.Rect.Bottom > region.Bottom) //Overlap in height - increase the region height
-                        {
-                            if (overlap.Rect.Bottom - region.Bottom > 0)
-                            {
-                                this.Height += overlap.Rect.Bottom - region.Bottom + ExtendSize + ExtendYHeight;
```
mmisol, line 329: On the other hand here, the code here doesn't check if the `nodesList` (which should be a single node) is a `NodeModel` or not, so maybe that check was not even needed.

Astul-Betizagasti:
> @Astul-Betizagasti Can you remind me the Jira task you created?

@QilongTang , I didn't create a jira task for this, the one I created a few days ago was for the problem after right-clicking on the annotation. But I can create one if needed

src/DynamoCore/Graph/Annotations/AnnotationModel.cs
Astul-Betizagasti: > I left some comments, but I think the solution looks good to the best of my knowledge. Are the tests going to be added in a different PR?

Yes, the tests will be added on a different PR, I wanted to resolve this first to know if i have to add any new tests

src/DynamoCore/Graph/Annotations/AnnotationModel.cs
mjkkirschner: yeah... so `Nodes` is a terrible name for this member as it is really a list of `ModelBase` - it is the connotation of `Nodes` and `Notes` (little grey boxes containing only text).

I think this code is trying to give differing offsets for various children types and removing this code as @mmisol said is doing away with that check.

Can you please make sure to test these changes with `Notes` included in the group? (annotation)

mjkkirschner: see:
https://github.com/DynamoDS/Dynamo/blob/bca65e58c4be124a3b21f1a648909ca585273425/src/DynamoCore/Graph/Annotations/AnnotationModel.cs#L139

Astul-Betizagasti: image
note
I just tried it and it works fine with notes too, the only difference being that if the lowest "node" is in fact a note, then the margin will be 5 pixels bigger (thats the hard-coded value of ExtendYHeight). The difference is barely noticeable but it could be fixed by checking if the node is a NodeModel before adding ExtendYHeight to the annotation height.

Astul-Betizagasti:
@QilongTang I just created a jira task for this, its DYN-2894

src/DynamoCore/Graph/Annotations/AnnotationModel.cs
mmisol: I would be fine with approving this as is, but maybe you can fix this if it is not much work @Astul-Betizagasti . What do you think @mjkkirschner  @QilongTang ?

QilongTang: Agree @Astul-Betizagasti Please fix and we will cover the testing part as a team

Astul-Betizagasti:
@mjkkirschner @mmisol @QilongTang I included the changes you mentioned

QilongTang:
@Astul-Betizagasti Looks good, can you double check the regressions? Looks like there are some image comparison diff?

QilongTang:
@Astul-Betizagasti I have seen the failing regressions in other PRs so I assume those are not caused by this PR. Merging

Issue #24, opened 2012-09-12
nodes and examples to show how to implement a basic optimization algorithm 


tatlin:
current SR example shows a brute force approach for one parameter, show how to generalize, add more params to domain, switch to different optimization strategies.

PR #10854, opened 2020-07-07
[DYN-2755] Preview disabled nodes can be renamed with double click

zeusongit:
### Purpose

[DYN-2755](https://jira.autodesk.com/browse/DYN-2755)
This adds the ability to the preview disabled nodes to interpret the double click event and open rename dialog box.
UI Automation Test PR: [#24](https://git.autodesk.com/Dynamo/DynamoUIAutomation/pull/24)

Before:
preview-before

After:
preview-after2

### Declarations

### Reviewers

@DynamoDS/dynamo


src/DynamoCoreWpf/Views/Core/NodeView.xaml.cs
```diff
@@ -376,6 +376,15 @@ private void OnPreviewMouseLeftButtonDown(object sender, MouseButtonEventArgs e)
...
             BringToFront();
+            if (e.ClickCount == 2)
+            {
+                Debug.WriteLine("Name double clicked!");
+                if (ViewModel != null && ViewModel.RenameCommand.CanExecute(null))
+                {
+                    ViewModel.RenameCommand.Execute(null);
+                }
+                e.Handled = true;
+            }
```
QilongTang, line 387: The part is not clear to me is that for nodes in other state, how are we handling the double click? On a different route?

mmisol: Should Python script nodes with preview disabled also be editable? This code only accounts for rename, but the code editor of those nodes can be opened as well by double clicking. It actually depends on the part of the node where you double click.

zeusongit: Just confirmed with Sol, nodes should/can be renamed in any state.

mmisol: But isn't this change overriding the behaviour of the handler, by handling the double click as a rename and marking the event as handled? Can you check that you can still edit a Python node by double clicking after your change?

QilongTang: Yup, @zeusongit if you can confirm the other cases, I am fine with this change

zeusongit: Can confirm.
python preview

PR #10857, opened 2020-07-07
Python Evaluation Helper

QilongTang:
Please Note:

### Purpose

Creating a new Singleton class Python Evaluation Helper so that it can be accessed multiple places to query if a particular Python engine is loaded. The class is internal for now.

When the selected engine is not loaded, Python node will display a node error currently showing 
image

FYI: @aparajit-pratap I believe the error state can be also used to look for missing Python Engine, please consider.

@mmisol has expressed before that the error message is not straight forward in this case. @mjkkirschner @Amoursol Can you suggest a better one? Seems we do not even need the warning part which is generic for Python node..

TODO:
- [x] Unit tests

### Declarations

### Reviewers

@DynamoDS/dynamo 

### FYIs

### DIFF
--- a/src/Dynamo.All.sln
+++ b/src/Dynamo.All.sln
@@ -83,7 +83,7 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DynamoCoreTests", "..\test\
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DynamoCoreWpfTests", "..\test\DynamoCoreWpfTests\DynamoCoreWpfTests.csproj", "{7DD8077A-201E-4C56-96C5-3C901A51BDF3}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "IronPythonTests", "..\test\Libraries\DynamoPythonTests\IronPythonTests.csproj", "{773988FE-EDF6-45CB-A63F-482955EB3553}"
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DynamoPythonTests", "..\test\Libraries\DynamoPythonTests\DynamoPythonTests.csproj", "{773988FE-EDF6-45CB-A63F-482955EB3553}"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DSIronPython", "Libraries\DSIronPython\DSIronPython.csproj", "{9EEF4F42-6B3B-4358-9A8A-C2701539A822}"
 	ProjectSection(ProjectDependencies) = postProject
--- a/src/Libraries/PythonNodeModels/Properties/Resources.en-US.resx
+++ b/src/Libraries/PythonNodeModels/Properties/Resources.en-US.resx
@@ -196,4 +196,7 @@
   <data name="PythonNodeContextMenuLearnMoreButton" xml:space="preserve">
     <value>Learn more about Python</value>
   </data>
+  <data name="PythonNodeErrorUnloadedEngineMsg" xml:space="preserve">
+    <value>Unloaded Python engine: </value>
+  </data>
 </root>
--- a/src/Libraries/PythonNodeModels/Properties/Resources.en-US.resx
+++ b/src/Libraries/PythonNodeModels/Properties/Resources.en-US.resx
@@ -197,6 +197,6 @@
     <value>Learn more about Python</value>
   </data>
   <data name="PythonNodeErrorUnloadedEngineMsg" xml:space="preserve">
-    <value>Unloaded Python engine: </value>
+    <value>Missing Python engine: </value>
   </data>
 </root>
--- a/src/Libraries/PythonNodeModels/PythonEngineSelector .cs
+++ b/src/Libraries/PythonNodeModels/PythonEngineSelector .cs
@@ -17,20 +17,20 @@ namespace PythonNodeModels
     /// Singleton class that other class can access and use for query loaded Python Engine info.
     /// TODO: Make the Singleton class public when Dynamic loading is also ready.
     /// </summary>
-    internal sealed class PythonEvaluationHelper
+    internal sealed class PythonEngineSelector 
     {
         /// <summary>
-        /// Use Lazy<![CDATA[PythonEvaluationHelper]]> to make sure the Singleton class is only initialized once
+        /// Use Lazy<![CDATA[PythonEngineSelector ]]> to make sure the Singleton class is only initialized once
         /// </summary>
-        internal static readonly Lazy<PythonEvaluationHelper>
+        internal static readonly Lazy<PythonEngineSelector >
             lazy =
-            new Lazy<PythonEvaluationHelper>
-            (() => new PythonEvaluationHelper());
+            new Lazy<PythonEngineSelector >
+            (() => new PythonEngineSelector ());
 
         /// <summary>
         /// The actual instance stored in the Singleton class
         /// </summary>
-        internal static PythonEvaluationHelper Instance { get { return lazy.Value; } }
+        internal static PythonEngineSelector  Instance { get { return lazy.Value; } }
 
         // TODO: The following fields might be removed after dynamic loading applied
         internal static bool IsIronPythonEnabled = false;
@@ -44,14 +44,14 @@ namespace PythonNodeModels
         /// <summary>
         /// Singleton class initialization logic which will be run in a lazy way the first time Dynamo try to evaluate a Python node
         /// </summary>
-        private PythonEvaluationHelper()
+        private PythonEngineSelector ()
         {
             var assems = AppDomain.CurrentDomain.GetAssemblies();
             // Currently we are using try-catch to validate loaded assembly and evaluation method exist
             // but we can optimize by checking all loaded types against evaluators interface later
             try
             {
-                var IronPythonAssem = assems.First(x => x.FullName.Contains("DSIronPython"));
+                var IronPythonAssem = assems.First(x => x.GetName().Name.Equals("DSIronPython"));
                 if (IronPythonAssem != null &&
                     IronPythonAssem.GetType("DSIronPython.IronPythonEvaluator").GetMethod(IronPythonEvaluationMethod) != null)
                 {
@@ -61,7 +61,7 @@ namespace PythonNodeModels
             catch{ }
             try
             {
-                var CPythonAssem = assems.First(x => x.FullName.Contains("DSCPython"));
+                var CPythonAssem = assems.First(x => x.GetName().Name.Equals("DSCPython"));
                 if (CPythonAssem != null &&
                     CPythonAssem.GetType("DSCPython.CPythonEvaluator").GetMethod(CPythonEvaluationMethod) != null)
                 {
@@ -80,16 +80,16 @@ namespace PythonNodeModels
         internal void GetEvaluatorInfo(PythonEngineVersion engine, out string evaluatorClass, out string evaluationMethod)
         {
             // Provide evaluator info when the selected engine is loaded
-            if (engine == PythonEngineVersion.IronPython2 && PythonEvaluationHelper.IsIronPythonEnabled)
+            if (engine == PythonEngineVersion.IronPython2 && PythonEngineSelector .IsIronPythonEnabled)
             {
-                evaluatorClass = PythonEvaluationHelper.IronPythonEvaluatorClass;
-                evaluationMethod = PythonEvaluationHelper.IronPythonEvaluationMethod;
+                evaluatorClass = PythonEngineSelector .IronPythonEvaluatorClass;
+                evaluationMethod = PythonEngineSelector .IronPythonEvaluationMethod;
                 return;
             }
-            if (engine == PythonEngineVersion.CPython3 && PythonEvaluationHelper.IsCPythonEnabled)
+            if (engine == PythonEngineVersion.CPython3 && PythonEngineSelector .IsCPythonEnabled)
             {
-                evaluatorClass = PythonEvaluationHelper.CPythonEvaluatorClass;
-                evaluationMethod = PythonEvaluationHelper.CPythonEvaluationMethod;
+                evaluatorClass = PythonEngineSelector .CPythonEvaluatorClass;
+                evaluationMethod = PythonEngineSelector .CPythonEvaluationMethod;
                 return;
             }
 
--- a/src/Libraries/PythonNodeModels/PythonEngineSelector .cs
+++ b/src/Libraries/PythonNodeModels/PythonEngineSelector .cs
@@ -20,7 +20,7 @@ namespace PythonNodeModels
     internal sealed class PythonEngineSelector
     {
         /// <summary>
-        /// Use Lazy<![CDATA[PythonEngineSelector ]]> to make sure the Singleton class is only initialized once
+        /// Use Lazy<<![CDATA[PythonEngineSelector ]]>> to make sure the Singleton class is only initialized once
         /// </summary>
         internal static readonly Lazy<PythonEngineSelector>
             lazy =
@@ -51,7 +51,7 @@ namespace PythonNodeModels
             // but we can optimize by checking all loaded types against evaluators interface later
             try
             {
-                var IronPythonAssem = assems.First(x => x.GetName().Name.Equals("DSIronPython"));
+                var IronPythonAssem = assems.FirstOrDefault(x => x.GetName().Name.Equals("DSIronPython"));
                 if (IronPythonAssem != null &&
                     IronPythonAssem.GetType("DSIronPython.IronPythonEvaluator").GetMethod(IronPythonEvaluationMethod) != null)
                 {
@@ -63,7 +63,7 @@ namespace PythonNodeModels
             }
             try
             {
-                var CPythonAssem = assems.First(x => x.GetName().Name.Equals("DSCPython"));
+                var CPythonAssem = assems.FirstOrDefault(x => x.GetName().Name.Equals("DSCPython"));
                 if (CPythonAssem != null &&
                     CPythonAssem.GetType("DSCPython.CPythonEvaluator").GetMethod(CPythonEvaluationMethod) != null)
                 {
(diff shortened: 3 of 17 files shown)


Amoursol:
Hello @QilongTang - do you mean there is some code inside the Python node that will query if the entire Dynamo space contains Python2 or Python3 engines loaded? If so, is this the correct image? This appears to be the message to point users towards upgrading their Packages 😊

QilongTang:
@Amoursol Yes, the code here will make that helper available so not only Dynamo can query but also extension can also query that if a particular engine is enabled or not. But I would like to run the current error message with you to see if it looks good to you or we want to show something different, let me know and refer to the screen shot above :)

src/Libraries/PythonNodeModels/PythonEvaluatorsHelper.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// Singleton class initialization logic which will be run in a lazy way the first time Dynamo try to evaluate a Python node
+        /// </summary>
+        private PythonEvaluatorsHelper()
+        {
+            var assems = AppDomain.CurrentDomain.GetAssemblies();
+            // Currently we are using try-catch to validate loaded assembly and evaluation method exist
+            // but we can optimize by checking all loaded types against evaluators interface later
+            try
+            {
+                var IronPythonAssem = assems.First(x => x.FullName.Contains("DSIronPython"));
```
mjkkirschner, line 54: while this is dynamic in the sense that it checks if the assemblies are present without needing to reference them - it is not dynamic in all aspects. For example, we need to know the names of the assemblies to check, and only check for 2 engine versions. It would be nice if this handled the case of new engines we have not written yet - ie IronPython 3.

Is that coming in another PR or you don't think it's useful enough?

QilongTang: Yes, @mjkkirschner Please refer to the TODO and comments.. We have a dedicate Jira task about Dynamic loading, this task is to make sure API is available so I left comments at places where we could apply dynamic loading later

mjkkirschner: got it thanks.

src/Libraries/PythonNodeModels/PythonEvaluationHelper.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// Singleton class initialization logic which will be run in a lazy way the first time Dynamo try to evaluate a Python node
+        /// </summary>
+        private PythonEvaluationHelper()
+        {
+            var assems = AppDomain.CurrentDomain.GetAssemblies();
+            // Currently we are using try-catch to validate loaded assembly and evaluation method exist
+            // but we can optimize by checking all loaded types against evaluators interface later
+            try
+            {
+                var IronPythonAssem = assems.First(x => x.FullName.Contains("DSIronPython"));
```
aparajit-pratap, line 54: I was thinking to check if the IronPython extension is loaded as a way to tell if DSIronPython is loaded or not. My concern with this approach is that there are ways that DSIronPython can be loaded in the appdomain but this does not guarantee that it is imported into the VM. The only way it will be imported into the VM right now, is in the presence of the extension.

aparajit-pratap: Also, it being imported in the VM implies that it will also be loaded in the appdomain.

QilongTang: hmm I thought about it briefly but I ended up this way because even extension is there, if DSIronPython is not loaded then IronPython is still not enabled. I think either of the approach here is ideal but should be replaced with the Dynamic loading logic based on interface type check later. What do you think?

Amoursol:
@QilongTang - Message looks pretty fine to me bar one thing: should it be "unknown" or "unloaded" ?

src/Libraries/PythonNodeModels/PythonEvaluationHelper.cs
aparajit-pratap: @QilongTang you got a point too.

test/Libraries/DynamoPythonTests/PythonEvalHelperTests.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    [TestFixture]
+    class PythonEvalHelperTests
+    {
+        [Test]
+        public void TestHelperInitialState()
+        {
+            Assert.AreEqual(false, PythonEvaluationHelper.lazy.IsValueCreated);
+            Assert.AreEqual(false, PythonEvaluationHelper.IsCPythonEnabled);
+            Assert.AreEqual(false, PythonEvaluationHelper.IsIronPythonEnabled);
```
QilongTang, line 14: @DynamoDS/dynamo Any idea that I kept getting this error..
image

reddyashish: I have seen this error when I was trying to directly assign the PythonVersion Enum values to a field in PythonEdit tests class. I couldn't find the actual reason why it was happening but when I did the same from inside a helper function, it worked. 
https://github.com/DynamoDS/Dynamo/blob/master/test/Libraries/DynamoPythonTests/PythonEditTests.cs#L33
I am curious to know what would be the difference though. Will have to look more into it.
In your case, PythonEvaluationHelper is the issue I think.

src/Libraries/PythonNodeModels/PythonEvaluationHelper.cs
```diff
@@ -0,0 +1,100 @@
...
+    {
+        IronPython2,
+        CPython3
+    }
+
+    /// <summary>
+    /// Singleton class that other class can access and use for query loaded Python Engine info.
+    /// TODO: Make the Singleton class public when Dynamic loading is also ready.
+    /// </summary>
+    internal sealed class PythonEvaluationHelper
```
aparajit-pratap, line 20: How about naming this `PythonEngineSelector` or something like that?

test/Libraries/DynamoPythonTests/PythonEvalHelperTests.cs
aparajit-pratap: What if you make the enum part of the `PythonEvaluationHelper` class, does that work?

mjkkirschner: are your tests actually preloading the libraries?

src/Libraries/PythonNodeModels/PythonEvaluationHelper.cs
```diff
@@ -0,0 +1,100 @@
...
+                if (IronPythonAssem != null &&
+                    IronPythonAssem.GetType("DSIronPython.IronPythonEvaluator").GetMethod(IronPythonEvaluationMethod) != null)
+                {
+                    IsIronPythonEnabled = true;
+                }
+            }
+            catch{ }
+            try
+            {
+                var CPythonAssem = assems.First(x => x.FullName.Contains("DSCPython"));
```
mmisol, line 64: To be accurate here, please use `x.GetName().Name` and compare for equality.

QilongTang: Will do

src/Libraries/PythonNodeModels/Properties/Resources.en-US.resx
```diff
@@ -196,4 +196,7 @@
   <data name="PythonNodeContextMenuLearnMoreButton" xml:space="preserve">
     <value>Learn more about Python</value>
   </data>
+  <data name="PythonNodeErrorUnloadedEngineMsg" xml:space="preserve">
+    <value>Unloaded Python engine: </value>
```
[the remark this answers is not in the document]
QilongTang: Will do

src/Libraries/PythonNodeModels/PythonEvaluationHelper.cs
aparajit-pratap: I just used the following code from an extension to determine whether the package was loaded or not:
```
var installedPackages = pmExtension.PackageLoader.LocalPackages;
var isPackageLoaded = installedPackages.Any(x =>
                x.Name == "IronPythonPackageName" && x.VersionName == "packageVersion");
```
where `pmExtension` is a reference to the package manager extension object.

src/Libraries/PythonNodeModels/PythonEvaluationHelper.cs
QilongTang: Will do

src/Libraries/PythonNodeModels/PythonEngineSelector .cs
```diff
@@ -0,0 +1,100 @@
...
+    {
+        IronPython2,
+        CPython3
+    }
+
+    /// <summary>
+    /// Singleton class that other class can access and use for query loaded Python Engine info.
+    /// TODO: Make the Singleton class public when Dynamic loading is also ready.
+    /// </summary>
+    internal sealed class PythonEngineSelector
```
QilongTang, line 20: @aparajit-pratap It is not renamed

src/Libraries/PythonNodeModels/PythonEngineSelector .cs
```diff
@@ -0,0 +1,100 @@
...
+        /// Singleton class initialization logic which will be run in a lazy way the first time Dynamo try to evaluate a Python node
+        /// </summary>
+        private PythonEngineSelector ()
+        {
+            var assems = AppDomain.CurrentDomain.GetAssemblies();
+            // Currently we are using try-catch to validate loaded assembly and evaluation method exist
+            // but we can optimize by checking all loaded types against evaluators interface later
+            try
+            {
+                var IronPythonAssem = assems.First(x => x.GetName().Name.Equals("DSIronPython"));
```
QilongTang, line 54: @mjkkirschner Addressed the comment here

src/Dynamo.All.sln
```diff
@@ -83,7 +83,7 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DynamoCoreTests", "..\test\
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DynamoCoreWpfTests", "..\test\DynamoCoreWpfTests\DynamoCoreWpfTests.csproj", "{7DD8077A-201E-4C56-96C5-3C901A51BDF3}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "IronPythonTests", "..\test\Libraries\DynamoPythonTests\IronPythonTests.csproj", "{773988FE-EDF6-45CB-A63F-482955EB3553}"
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DynamoPythonTests", "..\test\Libraries\DynamoPythonTests\DynamoPythonTests.csproj", "{773988FE-EDF6-45CB-A63F-482955EB3553}"
```
QilongTang, line 86: This is intentional and do not change assembly name because it was already named `DynamoPythonTests.dll`

test/DynamoCoreTests/DynamoCoreTests.csproj
```diff
@@ -268,11 +268,6 @@
       <Name>DynamoUtilities</Name>
       <Private>False</Private>
     </ProjectReference>
-    <ProjectReference Include="..\..\src\Libraries\PythonNodeModelsWpf\PythonNodeModelsWpf.csproj">
```
QilongTang, line 271: This is removed intentionally because I find DynamoCore tests should not depend on any wpf lib

src/Libraries/PythonNodeModels/PythonEvaluationHelper.cs
QilongTang: @aparajit-pratap I think that should serve your need. Once this PR is merged, you should be able to call `PythonEngineSelector.Instance.IsIronPythonEnabled`  from extension to also do the check.

aparajit-pratap: @QilongTang I actually prefer to do it this way as I think it's cleaner and was proposing the same solution to you.

QilongTang: @aparajit-pratap hmm understood, the reason I am doing it this way is because
1. We may not have access to package manager in a lot of the places where we need to call the API
2. We need to validate the evaluation method exist so that Dynamo do not get tricked by some dll just with the same name
3. These logic will be replaced with Dynamic loading later, so in the future we will be checking against in memory types implementing a particular interface but not just the method exist. That would be the most accurate in my opinion.
4. The IronPython extension only exist for Dynamo 2.7 and beyond, checking extension exist may not apply to lower version, although this depends on if there will be extension author implementing similar APIs.

To conclude, I believe the API will be improved in the future for a more generic and dynamic way of determining the loaded Python Engine.

src/Libraries/PythonNodeModels/PythonEngineSelector .cs
```diff
@@ -0,0 +1,100 @@
...
+                if (IronPythonAssem != null &&
+                    IronPythonAssem.GetType("DSIronPython.IronPythonEvaluator").GetMethod(IronPythonEvaluationMethod) != null)
+                {
+                    IsIronPythonEnabled = true;
+                }
+            }
+            catch{ }
```
mjkkirschner, line 61: I think this is going to raise a code small

src/Libraries/PythonNodeModels/PythonEngineSelector .cs
```diff
@@ -0,0 +1,100 @@
...
+            try
+            {
+                var CPythonAssem = assems.First(x => x.GetName().Name.Equals("DSCPython"));
+                if (CPythonAssem != null &&
+                    CPythonAssem.GetType("DSCPython.CPythonEvaluator").GetMethod(CPythonEvaluationMethod) != null)
+                {
+                    IsCPythonEnabled = true;
+                }
+            }
+            catch { }
```
mjkkirschner, line 71: same here

test/Libraries/DynamoPythonTests/DynamoPythonTests.csproj
```diff
@@ -44,30 +44,31 @@
     </Reference>
     <Reference Include="IronPython, Version=2.7.9.0, Culture=neutral, PublicKeyToken=7f709c5b713576e1, processorArchitecture=MSIL">
       <HintPath>..\..\..\src\packages\IronPython.2.7.9\lib\net45\IronPython.dll</HintPath>
-      <Private>True</Private>
+      <Private>False</Private>
```
mjkkirschner, line 47: I don't think these changes are great....
It means the tests will run - sure, but now we have multiple projects copies assemblies to the same output path - and it will be hard to figure out where they are coming from later.

mjkkirschner: whoops, I'm thinking of this backwards right?!

QilongTang: @mjkkirschner hmm I dont think these need to be true. I also have seen these properties are in consistent for some other test projects

mjkkirschner: yeah, I had it backwards.

src/Libraries/PythonNodeModels/PythonEngineSelector .cs
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /// <summary>
+    /// Singleton class that other class can access and use for query loaded Python Engine info.
+    /// TODO: Make the Singleton class public when Dynamic loading is also ready.
+    /// </summary>
+    internal sealed class PythonEngineSelector
+    {
+        /// <summary>
+        /// Use Lazy<![CDATA[PythonEngineSelector ]]> to make sure the Singleton class is only initialized once
```
mmisol, line 23: Did you forget the `<` and `>`? Also it might be easier to encode the individual characters like `Lazy&lt;PythonEngineSelector&gt;`

src/Libraries/PythonNodeModels/PythonEngineSelector .cs
mmisol: To avoid an unnecessary exception, even if it is catched, please use `FirstOrDefault` instead

test/Libraries/DynamoPythonTests/PythonEngineSelectorTests.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        /// <summary>
+        /// This test will cover the initial state of the Singleton
+        /// </summary>
+        [Test]
+        public void TestEngineSelectorInitial_State()
+        {
+            Assert.AreEqual(false, PythonEngineSelector .lazy.IsValueCreated);
+            Assert.AreEqual(false, PythonEngineSelector .IsCPythonEnabled);
+            Assert.AreEqual(false, PythonEngineSelector .IsIronPythonEnabled);
```
mmisol, line 27: Strange place to put a space. Was it intended?

QilongTang: ah these comes with the class renaming, will fix

QilongTang:
@mmisol Addressed. Let me know

src/Libraries/PythonNodeModels/PythonEngineSelector .cs
```diff
@@ -20,7 +20,7 @@ public enum PythonEngineVersion
     internal sealed class PythonEngineSelector
     {
         /// <summary>
-        /// Use Lazy<![CDATA[PythonEngineSelector ]]> to make sure the Singleton class is only initialized once
+        /// Use Lazy<<![CDATA[PythonEngineSelector ]]>> to make sure the Singleton class is only initialized once
```
mmisol, line 23: I think this way it's malformed XML, so it could create problems. But I notice now Visual Studio does not handle CDATA, so we should encode the characters individually like `Lazy&lt;PythonEngineSelector&gt;`. That one worked fine for me, here is a screenshot:
Screen Shot 2020-07-09 at 5 50 22 PM

QilongTang: ah @mmisol Thanks, will use that instead

QilongTang:
@mmisol Thank you for the review. Merging now

PR #10858, opened 2020-07-07
harden discovery of codeCompletionProvider.

mjkkirschner:
### Purpose

When search the appdomain for the CompletionProvider interfaces, if this linq statement encountered an exception it would bail out and fail to find the loaded completion providers. Now it does continues.

I could not figure out quick way to test this, if others believe this is required for this scenario I appreciate ideas.

### Declarations

### DIFF
--- a/src/Libraries/PythonNodeModelsWpf/SharedCompletionProvider.cs
+++ b/src/Libraries/PythonNodeModelsWpf/SharedCompletionProvider.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Linq;
+using System.Reflection;
 using Autodesk.DesignScript.Interfaces;
 using Dynamo.Logging;
 using ICSharpCode.AvalonEdit.CodeCompletion;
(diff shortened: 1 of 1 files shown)


src/Libraries/PythonNodeModelsWpf/SharedCompletionProvider.cs
```diff
@@ -31,15 +32,22 @@ internal SharedCompletionProvider(PythonEngineVersion version ,string dynamoCore
...
-                    .Where(p => completionType.IsAssignableFrom(p) && !p.IsInterface);
+                    .SelectMany(s => {
+                        try {
+                            return s.GetTypes();
+                        }
+                        catch (ReflectionTypeLoadException)
+                        {
+                            return new Type[0];
+                        }
+                    }).Where(p => completionType.IsAssignableFrom(p) && !p.IsInterface);
```
mmisol, line 50: One thing we could add is an explicit call to `ToList`. I see `loadedCodeCompletionTypes` is used afterwards a couple of times.

mjkkirschner: sure will do - I tested it manually.

PR #10860, opened 2020-07-07
[DYN-2870] Increase the code coverage: DynamoCore/Graph Folder

Astul-Betizagasti:
### Purpose

Remove unused code in the Workspaces sub folder

### Declarations

### Reviewers

Aaron Tang (@QilongTang )
Michael Kirschner (@mjkkirschner )

### FYIs

Aldredo Pozo (@alfredo-pozo )


QilongTang:
@DynamoDS/dynamo any comments? I think this is safe to remove since we are no longer serializing the whole workspace using XML. We still preserve the node serialization logic using XML for undo/redo which is un touched.

mjkkirschner:
unfortunately ... removing a protected member from a public method probably constitutes an API change? Maybe not a break... - If some client inherited from this class, and used that method, it will no call the base class right?

mjkkirschner:
always good to read this again:
https://github.com/dotnet/runtime/blob/master/docs/coding-guidelines/breaking-change-rules.md

I don't see protected members called out specifically...

QilongTang:
How about this
```
❌ DISALLOWED: Reducing the visibility of a member

This includes reducing the visibility of a protected member when there are accessible (public or protected) constructors and the type is not sealed. If this is not the case, reducing the visibility of a protected member is allowed.

Increasing the visibility of a member is allowed.
```
From https://docs.microsoft.com/en-us/dotnet/core/compatibility/

mjkkirschner:
well... does removing it count as lowering its visibility  😉 ?
I think the API will still be callable, an exception will not be thrown at compile or runtime - but behavior will potentially change. For example, I could imagine `Dynamo Player` uses this API.

QilongTang:
@mjkkirschner So obsoletion then?

mjkkirschner:
that would be the safe bet.

QilongTang:
@Astul-Betizagasti Can you obsolete these methods then, their should be plenty example in the code base about obsolete messaging. Sorry about the back and forth.

Astul-Betizagasti:
Ok, no problem I'll change it. Should I add a message on the Obsolete attribute or there is no need?

QilongTang:
@Astul-Betizagasti Yes
```
[Obsolete("Method will be deprecated in Dynamo 3.0.")]
```
I dont think there is a replacement on workspace model.. We can append use `ToJson()` or `Save()` but they are not the same in all ways.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -1499,6 +1499,7 @@ internal bool containsInvalidInputSymbols()
             return this.Nodes.OfType<Nodes.CustomNodes.Symbol>().Any(node => !node.Parameter.NameIsValid);
         }
 
+        [Obsolete("Method will be deprecated in Dynamo 3.0.")]
         private void SerializeElementResolver(XmlDocument xmlDoc)
```
aparajit-pratap, line 1503: Why can't we remove this right away?

QilongTang: @Astul-Betizagasti For private functions we should be able to remove any time. unlike protected and public ones

PR #10862, opened 2020-07-07
[DYN-2797] Provide custom node info in the IronPython2 migration dialog. 

reddyashish:
### Purpose

This PR implements UI functionality, to provide more information on the custom nodes in the workspace that have a direct dependency on IronPython2. This info is shown categorically in the Python migration dialog so that the users will have a better understanding of what they need to update. 

The list is shown in a tree view to differentiate between CN's that are from packages and the ones that are user defined. 
If the CN is user defined, then users can double click on the CN item and switch to its workspace to update the custom node to use CPython3 engine. If the CN is from a specific package, we show the package name and would expect the package authors to update the packages to use CPython3. The items in the list can be copy/paste-able. 

Based on the mockup provided in the task (https://jira.autodesk.com/browse/DYN-2797), I have created a first look of how the dialog would look like with all this information. 

iron python dialog 2

As of now, the custom nodes from a single package are not grouped together and are instead shown as separate list items. To achieve this, we might need another dictionary which would contain all the list of CN's per package id. Also binding wpf dictionary data with multiple values seems little complicated. 

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/Libraries/DSCPython/DSCPython.csproj
+++ b/src/Libraries/DSCPython/DSCPython.csproj
@@ -25,6 +25,7 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <Prefer32Bit>false</Prefer32Bit>
+    <LangVersion>7.1</LangVersion>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <DebugType>pdbonly</DebugType>
@@ -34,6 +35,7 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <Prefer32Bit>false</Prefer32Bit>
+    <LangVersion>7.1</LangVersion>
   </PropertyGroup>
   <ItemGroup>
     <!-- TODO Remove this dummy reference in Dynamo 3.0 or a version that is delivered with newer version of Global lunch products -->
--- a/src/Libraries/DSCPython/DSCPython.csproj
+++ b/src/Libraries/DSCPython/DSCPython.csproj
@@ -25,7 +25,6 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <Prefer32Bit>false</Prefer32Bit>
-    <LangVersion>7.1</LangVersion>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <DebugType>pdbonly</DebugType>
@@ -35,7 +34,6 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <Prefer32Bit>false</Prefer32Bit>
-    <LangVersion>7.1</LangVersion>
   </PropertyGroup>
   <ItemGroup>
     <!-- TODO Remove this dummy reference in Dynamo 3.0 or a version that is delivered with newer version of Global lunch products -->
--- a/src/PythonMigrationViewExtension/GraphPythonDependencies.cs
+++ b/src/PythonMigrationViewExtension/GraphPythonDependencies.cs
@@ -14,9 +14,9 @@ namespace Dynamo.PythonMigration
         private ViewLoadedParams ViewLoaded { get; set; }
 
         // A dictionary to mark Custom Nodes if they have a IronPython dependency or not. 
-        internal static Dictionary<Guid, CNPythonDependency> CustomNodePythonDependency = new Dictionary<Guid, CNPythonDependency>();
+        internal static Dictionary<Guid, CNPythonDependencyType> CustomNodePythonDependencyMap = new Dictionary<Guid, CNPythonDependencyType>();
 
-        internal enum CNPythonDependency
+        internal enum CNPythonDependencyType
         {
             NoDependency,
             NestedDependency,
@@ -28,7 +28,7 @@ namespace Dynamo.PythonMigration
             this.ViewLoaded = viewLoadedParams;
         }
 
-        internal bool ContainsIronPythonDependency()
+        internal bool ContainsIronPythonDependencyInCurrentWS()
         {
             var containsIronPythonDependency = false;
             var workspace = ViewLoaded.CurrentWorkspaceModel;
@@ -66,9 +66,9 @@ namespace Dynamo.PythonMigration
 
                 // If a custom node workspace is already checked for IronPython dependencies, 
                 // check the CustomNodePythonDependency dictionary instead of processing it again. 
-                if (CustomNodePythonDependency.TryGetValue(customNodeWS.CustomNodeId, out CNPythonDependency dependency))
+                if (CustomNodePythonDependencyMap.TryGetValue(customNodeWS.CustomNodeId, out CNPythonDependencyType dependency))
                 {
-                    if (dependency == CNPythonDependency.DirectDependency || dependency == CNPythonDependency.NestedDependency)
+                    if (dependency == CNPythonDependencyType.DirectDependency || dependency == CNPythonDependencyType.NestedDependency)
                     {
                         containIronPythonDependency = true;
                     }
@@ -79,12 +79,12 @@ namespace Dynamo.PythonMigration
 
                 if (hasPythonNodesInCustomNodeWorkspace)
                 {
-                    CustomNodePythonDependency.Add(customNodeWS.CustomNodeId, CNPythonDependency.DirectDependency);
+                    CustomNodePythonDependencyMap.Add(customNodeWS.CustomNodeId, CNPythonDependencyType.DirectDependency);
                     containIronPythonDependency = true;
                 }
                 else
                 {
-                    CustomNodePythonDependency.Add(customNodeWS.CustomNodeId, CNPythonDependency.NoDependency);
+                    CustomNodePythonDependencyMap.Add(customNodeWS.CustomNodeId, CNPythonDependencyType.NoDependency);
                 }
 
                 // Recursively check for IronPython dependencies in the nested custom nodes.
@@ -96,9 +96,9 @@ namespace Dynamo.PythonMigration
 
                     // If a custom node contains an IronPython dependency in its sub-tree,
                     // update its corresponding value to 'NestedDependency' in CustomNodePythonDependency.
-                    if (hasPythonNodesInCustomNodeWorkspace && CustomNodePythonDependency[customNodeWS.CustomNodeId] != CNPythonDependency.DirectDependency)
+                    if (hasPythonNodesInCustomNodeWorkspace && CustomNodePythonDependencyMap[customNodeWS.CustomNodeId] != CNPythonDependencyType.DirectDependency)
                     {
-                        CustomNodePythonDependency[customNodeWS.CustomNodeId] = CNPythonDependency.NestedDependency;
+                        CustomNodePythonDependencyMap[customNodeWS.CustomNodeId] = CNPythonDependencyType.NestedDependency;
                         containIronPythonDependency = true;
                     }
                 }
--- a/src/PythonMigrationViewExtension/GraphPythonDependencies.cs
+++ b/src/PythonMigrationViewExtension/GraphPythonDependencies.cs
@@ -14,9 +14,9 @@ namespace Dynamo.PythonMigration
         private ViewLoadedParams ViewLoaded { get; set; }
 
         // A dictionary to mark Custom Nodes if they have a IronPython dependency or not. 
-        private static Dictionary<Guid, CNPythonDependency> CustomNodePythonDependency = new Dictionary<Guid, CNPythonDependency>();
+        internal static Dictionary<Guid, CNPythonDependency> CustomNodePythonDependency = new Dictionary<Guid, CNPythonDependency>();
 
-        private enum CNPythonDependency
+        internal enum CNPythonDependency
         {
             NoDependency,
             NestedDependency,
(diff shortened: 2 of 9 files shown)


Amoursol:
@reddyashish looking really good! It's super clean and will be a huge addition that helps users understand the change. 

A couple of thoughts:
- Do we have a way for them to notify the package author? I'm not sure if due to GDPR we are publishing any personal details.
- If you select a node in the **Loose** section, does it zoom to fit on that node in the graph?

src/Libraries/DSCPython/DSCPython.csproj
```diff
@@ -25,6 +25,7 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <Prefer32Bit>false</Prefer32Bit>
+    <LangVersion>7.1</LangVersion>
```
reddyashish, line 28: I had to update the lang version here to 7.1 to get the DSCPython project build. I was getting this error: https://github.com/DynamoDS/Dynamo/pull/10830#issuecomment-652030087. @mmisol How did you get around it?

reddyashish:
@Amoursol I am not aware of any way, we can let users notify the package author directly from Dynamo. So I just added a tooltip for now. @mjkkirschner any thoughts?
The custom nodes shown here are the nodes that has a python script inside its workspace directly. So it can be that the CN shown here is not present directly in the workspace the user has opened but could be inside some nested custom nodes. It would get complicated if we want to zoom to fit into the selected CN.

Amoursol:
@reddyashish - Just to confirm, there are two sections: One = **Packaged** and two = **Loose**. Is the Packaged one inside Custom Nodes, and the Loose one just inside the graph itself? If so - can we zoom to fit on the Loose nodes in graph, the same way we do from **TuneUp**?

reddyashish:
Packaged refers to the Custom nodes that are defined in packages and Loose refers to the CN's that are user defined and are stored in %AppData%\Roaming\Dynamo\Dynamo Core\2.8\definitions. So Loose CN's can be used a different custom node.

src/Libraries/DSCPython/DSCPython.csproj
reddyashish: I was using 2017 VS. Will move to 2019 and remove this change.

Amoursol:
@reddyashish Ah, I see! Yes, in that case users will not understand 😄  Can we please change the loose name to **Definitions**? This matches the %appdata% name.

reddyashish:
@Amoursol  Changed the name to Definitions. This is how it is now:
Screen Shot 2020-07-09 at 11 00 47 AM

Amoursol:
@reddyashish LGTM

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
```diff
@@ -14,9 +14,9 @@ public class GraphPythonDependencies
         private ViewLoadedParams ViewLoaded { get; set; }
 
         // A dictionary to mark Custom Nodes if they have a IronPython dependency or not. 
-        private static Dictionary<Guid, CNPythonDependency> CustomNodePythonDependency = new Dictionary<Guid, CNPythonDependency>();
+        internal static Dictionary<Guid, CNPythonDependency> CustomNodePythonDependency = new Dictionary<Guid, CNPythonDependency>();
```
mjkkirschner, line 17: is this internal for testing or so it can be called from some other assembly?

mjkkirschner: because this an internal field I guess it should be camelCase.

mjkkirschner, line 17: to me this sounds like a singular item, maybe add Map to the name?

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
```diff
@@ -28,23 +28,29 @@ internal GraphPythonDependencies(ViewLoadedParams viewLoadedParams)
             this.ViewLoaded = viewLoadedParams;
         }
 
-        internal bool ContainsIronPythonDependencies()
+        internal bool ContainsIronPythonDependency()
```
mjkkirschner, line 31: hmm, I think a more appropriate name would be `CurrentWorkspaceContainsIronPythonDependency` - what do you think?

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
```diff
@@ -14,9 +14,9 @@ public class GraphPythonDependencies
         private ViewLoadedParams ViewLoaded { get; set; }
 
         // A dictionary to mark Custom Nodes if they have a IronPython dependency or not. 
-        private static Dictionary<Guid, CNPythonDependency> CustomNodePythonDependency = new Dictionary<Guid, CNPythonDependency>();
+        internal static Dictionary<Guid, CNPythonDependency> CustomNodePythonDependency = new Dictionary<Guid, CNPythonDependency>();
 
-        private enum CNPythonDependency
+        internal enum CNPythonDependency
```
mjkkirschner, line 19: CNPythonDependencyType?

src/PythonMigrationViewExtension/IronPythonInfoDialog.xaml
```diff
@@ -51,12 +50,57 @@
...
+                            </ListBox.CommandBindings>
+                            <ListBox.ItemTemplate>
+                                <DataTemplate>
+                                    <Grid Margin="0,2" ToolTip="{x:Static w:Resources.PackagedCustomNodesTooltip}">
+                                        <Grid.ColumnDefinitions>
+                                            <ColumnDefinition Width="100" />
+                                            <ColumnDefinition Width="*" />
+                                        </Grid.ColumnDefinitions>
+                                        <TextBlock Foreground="#bbbbbb" Text="{Binding PackageName}" />
+                                        <TextBlock Grid.Column="1" Foreground="#bbbbbb" Text="{Binding Name}" />
```
mjkkirschner, line 78: can we use colors only from the dynamo resource dictionary and if we need new colors add them to the dictionary with clear names?

src/PythonMigrationViewExtension/IronPythonInfoDialog.xaml.cs
```diff
@@ -12,8 +16,50 @@ internal IronPythonInfoDialog(PythonMigrationViewExtension viewModel)
...
+                    CustomNodeItem customNodeItem = new CustomNodeItem
+                    {
+                        Name = customNodeInfo.Name,
+                        FunctionId = customNodeInfo.FunctionId,
+                        PackageName = customNodeInfo.PackageInfo != null ? customNodeInfo.PackageInfo.Name : "Definitions",
+                    };
+
+                    if (customNodeInfo.PackageInfo != null)
+                    {
+                        PackagedCustomNodesContainingPython.Items.Add(customNodeItem);
```
mjkkirschner, line 43: is there a reason to implement this manually instead of using a binding?

src/PythonMigrationViewExtension/IronPythonInfoDialog.xaml.cs
```diff
@@ -12,8 +16,50 @@ internal IronPythonInfoDialog(PythonMigrationViewExtension viewModel)
...
+                    }
+                    else
+                    {
+                        UserDefinitionCustomNodesContainingPython.Items.Add(customNodeItem);
+                    }
+                }
+
+                if (PackagedCustomNodesContainingPython.Items.Count > 0)
```
mjkkirschner, line 51: same question here?

src/PythonMigrationViewExtension/IronPythonInfoDialog.xaml.cs
mjkkirschner: also if you do decide to use a binding remember to use an observable collection.

src/PythonMigrationViewExtension/IronPythonInfoDialog.xaml.cs
```diff
@@ -24,5 +70,56 @@ private void OnMoreInformationButtonClicked(object sender, RoutedEventArgs e)
             this.ViewModel.OpenPythonMigrationWarningDocumentation();
             this.Close();
         }
+
+        private void OnCustomNodeClick(object sender, RoutedEventArgs e)
+        {
+            if (UserDefinitionCustomNodesContainingPython.SelectedItems.Count == 1)
+            {
+                CustomNodeItem selectedCNItem = UserDefinitionCustomNodesContainingPython.SelectedItem as CustomNodeItem;
```
mjkkirschner, line 78: could this ever be null after this `as`?

src/PythonMigrationViewExtension/IronPythonInfoDialog.xaml.cs
```diff
@@ -24,5 +70,56 @@ private void OnMoreInformationButtonClicked(object sender, RoutedEventArgs e)
...
+    // Represents the data for each Custom Node item in the listbox. 
+    internal class CustomNodeItem
+    {
+        public string Name { get; set; }
+        public string PackageName { get; set; }
+        internal Guid FunctionId { get; set; }
+
+        public override string ToString()
+        {
+            if (PackageName.Equals("Definitions"))
```
mjkkirschner, line 115: can we pull this into a shared constant?

src/PythonMigrationViewExtension/IronPythonInfoDialog.xaml
```diff
@@ -51,12 +50,57 @@
...
+                                    </Grid>
+                                </DataTemplate>
+                            </ListBox.ItemTemplate>
+                        </ListBox>
+                    </Expander>
+                    <Expander Grid.Row="1" x:Name="ExpanderForUserDefinitionDependencies" Header="Definitions"  Foreground="#bbbbbb" FontSize="15" Margin="20,5,0,0" Visibility="Collapsed">
```
mjkkirschner, line 84: The word definitions needs to be added to the localization resx.

src/PythonMigrationViewExtension/IronPythonInfoDialog.xaml
```diff
@@ -51,12 +50,57 @@
...
-                       TextWrapping="Wrap">
+                       TextWrapping="Wrap"
+                       Margin="0,0,0,10">
             </TextBlock>
 
+            <!-- CustomNodes that have IronPython2 Dependency -->
+            <Expander x:Name="MainExpander" Header="{x:Static w:Resources.CustomNodesPythonDependencyHeader}" Foreground="#bbbbbb"
+                      FontSize="15" Margin="-2,0,0,0" Visibility="Hidden" ScrollViewer.VerticalScrollBarVisibility="Visible">
+                <StackPanel Height="Auto">
+                    <Expander x:Name="ExpanderForPackagedDependencies" Header="Packaged" Foreground="#bbbbbb" FontSize="15" Margin="20,5,0,0" Visibility="Collapsed">
```
mjkkirschner, line 64: "Packaged" needs to be added to the localization resx.

mjkkirschner:
@reddyashish some comments - also tests 😉 ?

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
reddyashish: Internal because it is being used by other files.

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
reddyashish: yes, sounds good. will change.

reddyashish:
I have added tests for CustomNodesContainIronPythonDependency API in the previous PR, I thought we might need UI automation tests for this? What do you think?

src/PythonMigrationViewExtension/IronPythonInfoDialog.xaml.cs
reddyashish: I dont think it can be null here, as we are checking that the only one list item is selected. and the UserDefinitionCustomNodesContainingPython contains only user defined custom nodes.

reddyashish: Do you think the OpenCustomNodeWorkspace should be called from try-catch?

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
QilongTang: Sorry about being picky on this,, https://stackoverflow.com/questions/1370840/naming-conventions-what-to-name-a-method-that-returns-a-boolean

Maybe `ContainsIronPythonDependencyInCurrentWS`

src/PythonMigrationViewExtension/IronPythonInfoDialog.xaml
reddyashish: Using the colors from resource dictionary now.

src/PythonMigrationViewExtension/IronPythonInfoDialog.xaml.cs
reddyashish: As this list is static and doesn't have much functionality added to it, we decided to keep it simple.

reddyashish:
I think I addressed all the comments and would be good to test this as well on Monday's bug bash. So If this looks, we can merge it and will address any issues in a followup PR.

src/PythonMigrationViewExtension/IronPythonInfoDialog.xaml.cs
```diff
@@ -24,5 +67,65 @@ private void OnMoreInformationButtonClicked(object sender, RoutedEventArgs e)
...
+        }
+
+        private void ExecuteCtrlCCopyCommand(object sender, ExecutedRoutedEventArgs e)
+        {
+            string collectedText = "";
+            ListBox lb = (ListBox) sender;
+            
+            foreach (var item in lb.SelectedItems)
+            {
+                collectedText += item.ToString() + "\r\n";
```
mjkkirschner, line 88: just a small thing, but I think this should be `System.Environment.NewLine`

reddyashish:
Thanks @mjkkirschner and @QilongTang. Merging this.

PR #10865, opened 2020-07-08
Decode Python list into IList in function calls

mmisol:
### Purpose

Implements a decoder for Python sequence to transform them into generic
or non-generic IList. This relies on Python.NET 'PySequence.ToList'.

Also implements an encoder to perform a 'no-op'. This allows to treat
returned lists from function calls like .NET lists, which is the
behavior of Iron Python.

### Declarations

### Reviewers

@DynamoDS/dynamo


src/Libraries/DSCPython/Encoders/ListEncoder.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public bool TryDecode<T>(PyObject pyObj, out T value)
+        {
+            if (!PySequence.IsSequenceType(pyObj))
+            {
+                value = default;
+                return false;
+            }
+
+            using (var pyList = PyList.AsList(pyObj))
```
mjkkirschner, line 39: I'm confused by this using statement, does `AsList` return a .net list or a Python List?

src/Libraries/DSCPython/Encoders/ListEncoder.cs
```diff
@@ -0,0 +1,60 @@
...
+                }
+                return true;
+            }
+        }
+
+        public PyObject TryEncode(object value)
+        {
+            // This is a no-op to prevent Python.NET from encoding generic lists
+            // https://github.com/pythonnet/pythonnet/pull/963#issuecomment-642938541
+            return PyObject.FromManagedObject(value);
```
mjkkirschner, line 57: does this break examples where I want to call a python function with a list of ints?

```
def sum(data):
    for i in data:
```
where data is a List<int> ?

src/Libraries/DSCPython/Encoders/ListEncoder.cs
mmisol: It's a Python list

test/Libraries/DynamoPythonTests/PythonEvalTestsWithLibraries.cs
```diff
@@ -154,19 +158,16 @@ import clr
 r = range(0, 10, 2)
 # Python range => .NET array - Works in both
 a = DummyCollection.MakeArray(r)
-# Python range => .NET IList - Does not work in CPython
```
mjkkirschner, line 157: it does make me wonder if this should work - but through an exception - so that users don't expect all future engines to support all conversions...

IE when ironPython 3 exists, maybe we only want to support a limited set of conversions.

src/Libraries/DSCPython/Encoders/ListEncoder.cs
mmisol: Good question. Let me try it and come back with the answer.

test/Libraries/DynamoPythonTests/PythonEvalTestsWithLibraries.cs
mmisol: That one came for free because ranges can be iterated over and it just works. Iron Python 3 might add support for this conversion, it makes sense if they do it. I guess I see your point, but I don't think we should claim that at some point the engines will converge. I guess people will either stick to one or learn to live with their differences.

src/Libraries/DSCPython/Encoders/ListEncoder.cs
mmisol: It works. Do you want me to add it to the test?
Screen Shot 2020-07-08 at 12 39 42 PM

mjkkirschner: sure, nothing wrong with more tests?

mjkkirschner: okay, so l2 is a .net list because Flatten is a zt function?

mmisol: Yes, it's `List.Flatten` from our core nodes

mmisol:
The build passed here https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/695/. Merging

PR #10868, opened 2020-07-08
Decode Python Dictionary in function calls

mmisol:
### Purpose

Allows to decode Python dict to .NET IDictionary by using a Python.NET
decoder. Both generic and non-generic dictionaries are supported.
### Declarations

### Reviewers

@DynamoDS/dynamo


src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -111,8 +111,12 @@ static CPythonEvaluator()
...
         private static void InitializeEncoders()
         {
-            var encoders = new IPyObjectEncoder[] { new BigIntegerEncoder(), new ListEncoder() };
-            var decoders = encoders.Cast<IPyObjectDecoder>().ToArray();
+            var shared = new object[] { new BigIntegerEncoderDecoder(), new ListEncoderDecoder() };
+            var encoders = shared.Cast<IPyObjectEncoder>().ToArray();
+            var decoders = shared.Cast<IPyObjectDecoder>().Concat(new IPyObjectDecoder[]
+            {
+                new DictionaryDecoder()
+            }).ToArray();
```
aparajit-pratap, line 119: @mmisol so it looks like we have both an encoder and a decoder for List but not for Dictionary, correct?

mmisol: Yes, that's technically correct. In practice, the encoder we have for List only has the purpose of preventing the default encoding Python.NET does for lists. By doing that we can get a consistent behaviour between List and Dictionary conversions.

src/Libraries/DSCPython/Encoders/ListEncodeDecoder.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace DSCPython.Encoders
 {
-    internal class ListEncoder : IPyObjectEncoder, IPyObjectDecoder
+    internal class ListEncoderDecoder : IPyObjectEncoder, IPyObjectDecoder
```
aparajit-pratap, line 9: I see obsolete messages on `IPyObjectEncoder` and Decoder saying they are unstable and can be changed in the next minor release - something to keep note of unless there is an alternative interface already existing that we can use and replace right away.

mmisol: The custom encoder/decoder API appeared with 2.5, which is the latest minor release of Python.NET. My guess is they might make changes to it in version 3.0 so you bring a good point.

Before this API there was no way to do something like this AFAIK. We chose to use the API where possible, even if it is unstable, because otherwise we would have to modify Python.NET even more. Also, from these encoders/decoders you could target custom types, like DS Dictionary, which is another plus.

test/Libraries/DynamoPythonTests/PythonEvalTestsWithLibraries.cs
```diff
@@ -188,25 +187,35 @@ import clr
...
-d = DummyCollection.AcceptIDictionary(d)
-# .NET IDictionary => Python dict - Works in IronPython. Could not test in CPython due to previous bug
-d['five'] = 5
+# Python dict => .NET IDictionary
+untypedDictionary = DummyCollection.AcceptIDictionary(d)
+untypedDictionary['four'] = 4
 
-OUT = d
+# Python dict => .NET IDictionary<> - Does not work in IronPython
+typedDictionary = DummyCollection.AcceptDictionary(d)
```
aparajit-pratap, line 196: I understand that passing `d` to `AcceptDictionary` or IDictionary is the decoding part to convert it into the .NET dictionary but what about returning the .net dictionary from `AcceptDictionary`? Doesn't that need an encoding from the  returned .NET dictionary to the Python dictionary?

aparajit-pratap: Are the returned `untypedDictinoary` and `typedDictionary`  not Python dictionaries?

mmisol: It does trigger an encode operation but in this case we are not doing anything special. Python.NET will wrap .NET objects that it doesn't know how to encode with a generic wrapper. I haven't taken much of a look to their implementation, but they even support iterating like  this for example https://github.com/DynamoDS/Dynamo/blob/master/test/Libraries/DynamoPythonTests/PythonEvalTestsWithLibraries.cs#L77

PR #10869, opened 2020-07-09
Easily install missing IronPython extension package

aparajit-pratap:
### Purpose

Easily install missing IronPython extension package.
python_package

### Declarations

### Reviewers

@DynamoDS/dynamo 

### FYIs


src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
```diff
@@ -18,7 +20,11 @@ namespace Dynamo.WorkspaceDependency
     public class WorkspaceDependencyViewExtension : IViewExtension, ILogSource
     {
         internal MenuItem workspaceReferencesMenuItem;
-        private const String extensionName = "Workspace References";
+        private const string extensionName = "Workspace References";
+        private const string pythonPackage = "DSIronPython_Test";
+        private readonly Version pythonPackageVersion = new Version(1, 0, 7);
```
aparajit-pratap, line 25: Can we download the latest version of a package from PM by default because currently, AFAIK, we have to specify the exact version? Or better yet, can we just publish a 1.0.0 version of the IronPython extension package to prod?

src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.csproj
```diff
@@ -82,6 +82,10 @@
       <Project>{47533B7C-0E1A-44A4-8511-B438645F052A}</Project>
       <Name>DynamoPackages</Name>
     </ProjectReference>
+    <ProjectReference Include="..\PythonMigrationViewExtension\PythonMigrationViewExtension.csproj">
+      <Project>{10af430d-0d3a-49ce-a63d-848912959745}</Project>
```
aparajit-pratap, line 86: yuck!

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
```diff
@@ -252,27 +252,6 @@ public void AddDependent(Guid guid)
...
-            {
-                Nodes.Add(guid);
-            }
-        }
-
-        /// <summary>
-        /// Remove a dependent node
-        /// </summary>
-        /// <param name="guid"></param>
-        internal void RemoveDependent(Guid guid)
```
aparajit-pratap, line 271: unused code

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -124,7 +124,15 @@ private void OnWorkspacePropertyChanged(object sender, PropertyChangedEventArgs
         internal void DependencyRegen(WorkspaceModel ws)
         {
             this.RestartBanner.Visibility = Visibility.Hidden;
-            var packageDependencies = ws.NodeLibraryDependencies.Where(d => d is PackageDependencyInfo);
+            var packageDependencies = ws.NodeLibraryDependencies.Where(d => d is PackageDependencyInfo).ToList();
+
+            // This assumes it is a one-time addition to add the IronPython extension 
+            // as a package dependency. We might only want to give users the option to install
+            // the package on opening a workspace for the first-time. Even if any IronPython nodes are deleted
+            // from the workspace subsequently, it should not be necessary to uninstall/reinstall the package again later.
```
QilongTang, line 132: @aparajit-pratap Can you talk about it more? Let's say if the user was hesitate to address it first time opening, then the chance is gone?

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -287,6 +295,8 @@ private void Refresh_MouseLeftButtonDown(object sender, System.Windows.Input.Mou
         {
             DependencyRegen(currentWorkspace);
         }
+
+
```
QilongTang, line 299: would you remove the extra lines

src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
QilongTang: @aparajit-pratap I think the workspace references design is based on package name and version.. I do not anticipate user need to resolve a case later when they need to upgrade to a newer version. But who knows what could happen :) I would vote for being consistent with the existing behavior of other WS references

src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.csproj
QilongTang: We can discuss.. Any other work around you guys can see @DynamoDS/dynamo? @aparajit-pratap you mentioned moving `ContainsIronPythonDependencies` to DynamoCore does not help correct?

src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
aparajit-pratap: @QilongTang I'm not sure I understand what you mean by staying consistent in this context?

src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.csproj
aparajit-pratap: This function has a dependency on PythonNodeModels and DynamoCore doesn't.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
QilongTang: By "consistent" I mean WS Reference will display the package graph is depending on including name and version, if user local version is behind, user get a chance to choose to upgrade or not. Since you mentioned we always install the latest version, I want to make sure it is a consistent with what we have or not..

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
aparajit-pratap: If the user doesn't download it the first time, it will continue to display as a missing dependency in the workspace references view, just like any other missing package.

aparajit-pratap:
I'm currently looking at the failing selfserve tests ...

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
aparajit-pratap: I should make the comment clearer. What I meant to say is that if a python node is removed from the workspace, the workspace references may not be updated. I need to double-check.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
aparajit-pratap: I asked here if we can simply install the latest version that is available on the PM as an exception for the IronPython packages as otherwise, we'll have to hardcode this value in the code. I hope it is a one-time thing as I don't anticipate this package being updated in the near future.

mjkkirschner: I don' t think there is an API to download the latest version, but there are APIs for getting the versions of a package, and then we could download the latest.

@zeusongit can you advise here?

zeusongit: Yes, correct! We don't have an API that returns the latest version of a package directly, we may have to get the latest version from a separate request.
Although, as aparajit suggested, if it is a one-time thing, then we can just fetch the latest package by using the fixed version number.

But in case, we really don't want to hard code things here, we can think about adding a new API to the PM.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyViewExtension.cs
```diff
@@ -18,7 +20,11 @@ namespace Dynamo.WorkspaceDependency
     public class WorkspaceDependencyViewExtension : IViewExtension, ILogSource
     {
         internal MenuItem workspaceReferencesMenuItem;
-        private const String extensionName = "Workspace References";
+        private const string extensionName = "Workspace References";
+        internal readonly string pythonPackage = "DSIronPython_Test";
+        internal readonly Version pythonPackageVersion = new Version(1, 0, 7);
```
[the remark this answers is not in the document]
aparajit-pratap: By final, do you mean sealed?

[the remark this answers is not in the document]
mmisol: I don't mean modifiers of variables but their values. Maybe `name = "Iron Python script engine"` and `version = "2.7.9"`?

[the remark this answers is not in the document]
aparajit-pratap: Yes, we intend to update these names and versions before publishing the package to production PM. FYI: @QilongTang. They will then need to be updated here. At the moment this is referring to the package temporarily created on Dev PM.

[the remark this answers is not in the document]
QilongTang: Understood, this is meant to test on PM_dev from my understand

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -541,6 +541,12 @@ public HashSet<Guid> Dependencies
             }
         }
 
+        internal Func<WorkspaceModel, IEnumerable<INodeLibraryDependencyInfo>> RequestPackageDependencies;
+        internal IEnumerable<INodeLibraryDependencyInfo> OnRequestPackageDependencies()
+        {
+            return RequestPackageDependencies?.Invoke(this);
+        }
```
aparajit-pratap, line 548: Will be adding code comments.

src/PythonMigrationViewExtension/PythonMigrationViewExtension.cs
```diff
@@ -119,6 +120,7 @@ private void UnsubscribeFromDynamoEvents()
             LoadedParams.CurrentWorkspaceChanged -= OnCurrentWorkspaceChanged;
             DynamoViewModel.CurrentSpaceViewModel.Model.NodeAdded -= OnNodeAdded;
             DynamoViewModel.Model.Logger.NotificationLogged -= OnNotificationLogged;
+            //CurrentWorkspace.RequestPackageDependencies -= PythonDependencies.AddPythonPackageDependency;
```
mjkkirschner, line 123: why remove this and the other unsubscribe statements? It's hard to reason about if the extra handlers will cause memory leaks.

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
```diff
@@ -12,6 +12,9 @@ namespace Dynamo.PythonMigration
     public class GraphPythonDependencies
     {
         private ViewLoadedParams ViewLoaded { get; set; }
+        internal static readonly string PythonPackage = "DSIronPython_Test";
+        internal static readonly Version PythonPackageVersion = new Version(1, 0, 7);
```
aparajit-pratap, line 16: These will be updated once the IronPython extension is published to PM prod.

src/PythonMigrationViewExtension/PythonMigrationViewExtension.cs
mjkkirschner: sorry, missed your commit.

aparajit-pratap:
> What should be the strategy for releasing future versions of the Iron Python engine package? Should 2.7.9 be the version here and whenever 3.0.0 is released we deprecate the previous version of the package in PM (if that's possible)?

@mmisol, as discussed I think the strategy is to add new packages for new engine versions and make them available in the PM in the future.

PR #10878, opened 2020-07-10
Python class interop using shared global scope

mjkkirschner:
### Purpose

This PR starts to address the case where a custom python class or other unconvertible type is returned from a CPython engine to Dynamo as an output.

In these cases we wrap the pyObj.handle, which is an `intPtr` matching the python objects memory location (and also its `ID` which python docs state will be stable for the object lifetime) in a class `DynamoCPythonHandle`

We create the new class because we want some custom dispose logic, and also because any calls to the PyObj need to be wrapped in `GIL` and `Python Lock` calls and we don't want to add these to the VM.

~~I also thought it might be smart to avoid any thread contention just to get a string rep for the preview bubble, so we get the string rep before leaving the marshaler and cache that.~~

#### Manual tests have I done that need cases created:

- [x] object is returned, used by multiple nodes downstream, evaluation is correct for all nodes, one handle is created.
- [x] object is returned, used by multiple nodes downstream, node is deleted causing dispose, variable is removed from global scope.
- [ ] object is returned, node is deleted, object is disposed.
- [x] object is returned, downstream node modifies it in python, object is returned again from second node, handles are disposed corretly.
- [x] SamePythonObject returned multiple times from same node, multiple handles created but they have the same id, but disposal is done safely.

### Declarations


src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -10,6 +10,70 @@
 
 namespace DSCPython
 {
+    /// <summary>
+    /// This class wraps a PythonNet.PyObj and performs
+    /// disposal tasks to make sure the underlying object is removed from
+    /// the shared global scope between all CPython scopes.
+    /// </summary>
+    [IsVisibleInDynamoLibrary(false)]
+    public class DynamoCPythonHandle : IDisposable
```
mmisol, line 19: Ok. Now I see we need to do some special handling for disposal, so using plain `PyObject` won't do.

Anyway, this is a secondary discussion 🙂

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -244,8 +321,11 @@ public static DataMarshaler OutputMarshaler
                                 {
                                     if (unmarshalled.Equals(pyObj))
                                     {
-                                        // Object can't be unmarshalled. Prevent a stack overflow.
-                                        throw new InvalidOperationException(Properties.Resources.FailedToUnmarshalOutput);
+                                        var globalScope = PyScopeManager.Global.Get(globalScopeName);
+                                        //try moving object to global scope
+                                        globalScope.Set(pyObj.Handle.ToString(),pyObj);
```
mmisol, line 326: I think this might be unsafe. Basically we are using memory addresses to identify these variables. For different executions of script nodes (even the same one) you could get either the same memory space (lucky if it's the same node, unlucky if it's not), overlapping (you might get a different variable) or disjointed (you might not get the variable). The disposal of these variables by the VM also plays a part here. I'm not sure if it's triggered at a certain event or it can happen at any time. You might get inconsistent results depending on the answer to that too.

mjkkirschner: yes I think we'll have to think through the cases carefully, I originally used guids to identify these handles, instead of the python memory locations, but I thought perhaps we want only one handle for each `PyObj`.

I think like you brought up, we'll have to consider what happens when references or instances of the same handle are returned into the graph, and then one of these goes out of scope and is disposed, but another is not.

Disposal is handled by the DS garbage collector - I think it occurs when the VM resets, when a new graph run is started, and when a new workspace is opened or on shutdown. 

Someone could manually call dispose as well.

python docs do indicate that a particular object will retain the its `id` for its lifetime, and the `intptr` is created using that id as its memory address - though I guess it could move, even if the visible id does not, and the value could then change? I can't find docs on this.

Maybe it's okay if there are multiple DynamoHandles which really point to the same PyObj as long as we dispose them correctly... Maybe doing some reference counting?

mjkkirschner: I guess the question I want to answer is if we can reliably check if one pyObj points to the same underlying python object as another? Can we use hashcode or equality checks on these objects?

mjkkirschner: nice, it looks like we can - 
https://github.com/pythonnet/pythonnet/blob/master/src/runtime/pyobject.cs#L1063

what do you think about that solution instead?

mjkkirschner: another idea - 
we could expose an `is` method on the `pyObj` wrapper on our fork, I can't seem to find one - though I guess we could just call this using `InvokeMethod` ?

mmisol:
Played with this for a while. Looks pretty convincing. Even though I found a bug, I think it should be easy to fix by keeping a ref count on our side or something similar.

Good things:
- The disposal of instances seems to be triggered consistently
- The handle is retained from different executions

Limitations but same as IronPython
- It makes little sense to modify these objects in parallel branches, as you cannot guarantee which will modify the object first
- Even though the object might be available, the definition of its class is not

Bug
As different Python scripts return the same object, Dynamo will be create a different wrapper instances of the same handle. This can be seen debugging by placing a breakpoint in the wrapper constructor and taking a look at the handler id.
The disposal, however, seems to get rid of the object. That can lead to scenarios where the object is variable is deleted but other nodes are still using it.
Screen Shot 2020-07-10 at 2 34 26 PM


Repro instructions:
1. Open graph
2. Add a space in the second script node. Run
3. Remove the added space. Run again

[Python_Simple_Bug.dyn.txt](https://github.com/DynamoDS/Dynamo/files/4904874/Python_Simple_Bug.dyn.txt)

test/Libraries/DynamoPythonTests/PythonEditTests.cs
```diff
@@ -511,5 +511,88 @@ public void Python_CanReferenceDynamoServicesExecutionSession()
...
+
+            ViewModel.HomeSpace.Run();
+            AssertPreviewValue(downstream2.GUID.ToString(), "joe");
+
+            ViewModel.Model.CurrentWorkspace.Nodes.OfType<CodeBlockNodeModel>().First().UpdateValue(new UpdateValueParams("Code", "\"foo\";"));
+
+            ViewModel.HomeSpace.Run();
+            AssertPreviewValue(downstream2.GUID.ToString(), "foo");
+
+        }
```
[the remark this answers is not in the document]
mjkkirschner: thanks @mmisol will do!

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -10,6 +10,119 @@
...
+        /// A unique ID that identifies this python object. It's as a lookup
+        /// symbol within the global scope to find this python object instance.
+        /// </summary>
+        internal IntPtr PythonObjectID { get; set; }
+        /// <summary>
+        /// A string representation of this object.
+        /// TODO - it may be better to just call ToString - this was an attempt
+        /// to optimize needing to use locks and the GIL to get strings for preview bubbles.
+        /// </summary>
+        internal string StringRepresentation { get; set; }
```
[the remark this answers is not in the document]
mjkkirschner: this property is used in the `ToString` override method for this class - what do you think about this given the comment above? We could take the other approach, get rid of this property, and have every call to toString make a call to python.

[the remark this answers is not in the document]
mmisol: I think we should probably use it. That way we don't need to deal with locking every time.

One thing I'm not seeing though, is if this is getting updated as Python nodes change the object. Could be something to test.

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -10,6 +10,119 @@
...
+                {
+
+                    PyScopeManager.Global.Get(CPythonEvaluator.globalScopeName).Remove(PythonObjectID.ToString());
+                    HandleCountMap.Remove(this);
+                   
+                } 
+            }
+            catch (Exception E)
+            {
+                System.Diagnostics.Debug.WriteLine("error removing python object from global scope");
```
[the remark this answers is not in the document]
mjkkirschner: hmm @mmisol - because this mostly happens when the graph changes or when nodes are deleted - putting a warning on the node is basically useless -  the only thing I can think of here at the moment is to implement the `ILogSource` interface or pass a logger to the CPythonEvaluator class, that will take a bit of refactoring and the DSCPython binary will then reference DynamCore.

Any other suggestions?

src/Libraries/DSCPython/CPythonEvaluator.cs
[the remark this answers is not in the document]
mjkkirschner: yep, let me take a look at that, because we're just using the handle, and that address should not change I thought this was okay to cache for the object lifetime, but maybe it gives a false impression and could be useful for debugging to update it at some points.

src/Libraries/DSCPython/CPythonEvaluator.cs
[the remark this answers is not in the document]
mmisol: You are right, I forgot about that. I guess we can leave that for the future. Maybe log a ticket?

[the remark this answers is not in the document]
mjkkirschner: I'll add a TODO and a ticket number.

src/Libraries/DSCPython/CPythonEvaluator.cs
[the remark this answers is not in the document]
mmisol: Yes, but if you update the object, then its `ToString` representation might change, I guess. It depends on what `PyObject` does.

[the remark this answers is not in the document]
mjkkirschner: yes, you're right, this will get out of date if users implement `__repr__` or `__str__` on their classes. Will think on it. Might need to ditch the cacheing...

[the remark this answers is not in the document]
mjkkirschner: I'm leaning towards the following.
1. for now, just remove the property, remove the cacheing logic, and let the toString lock and get the most up to date value.

2. if we find in the future that this causes performance problems worth optimizing, we can add cacheing back and update ALL existing DSCPythonHandles string values every time any Cpython engine executes, so this update occurs all within the same lock statment -  We could try to make this stricter and check for references to avoid unnecessary updates, but I'm not sure we would detect all nested properties etc.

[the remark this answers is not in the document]
mmisol: Option 1 sounds ok. This should be an uncommon usage of Python anyway.

mjkkirschner:
I need to update graceful failure test.

mjkkirschner:
@mmisol PTAL, test is fixed.

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -10,6 +10,119 @@
...
+    class DynamoCPythonHandleComparer : IEqualityComparer<DynamoCPythonHandle>
+    {
+
+        public bool Equals(DynamoCPythonHandle x, DynamoCPythonHandle y)
+        {
+            if (x.PythonObjectID.Equals(y.PythonObjectID))
+            {
+                return true;
+            }
+            return false;
```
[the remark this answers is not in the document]
mmisol: I meant `return x.PythonObjectID.Equals(y.PythonObjectID)` 😉

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -10,6 +10,122 @@
...
+                using (Py.GIL())
+                {
+                    var scope = PyScopeManager.Global.Get(CPythonEvaluator.globalScopeName);
+                    var pyobj = scope.Get(PythonObjectID.ToString());
+                    return pyobj.ToString();
+                }
+            }
+            catch (Exception e)
+            {
+                //TODO(DYN-2941) implement Ilogsource or pass logger to python eval so can log to Dynamo console here.
```
mmisol, line 78: Nice!

test/Libraries/DynamoPythonTests/PythonEditTests.cs
```diff
@@ -21,7 +23,6 @@ protected override void GetLibrariesToPreload(List<string> libraries)
         {
             libraries.Add("DesignScriptBuiltin.dll");
             libraries.Add("DSCoreNodes.dll");
-            libraries.Add("DSIronPython.dll");
```
mmisol, line 24: Are you sure you can remove this? I see some test that seem to use IronPython, like the one that uses `IronPythonInfo_TestGraph.dyn`.

mjkkirschner: I believe it should be removed since these tests should be loading the IronPythonExtension which should load DSIronPython anyway, trying to load it twice just results in time spent catching exceptions during each test's startup.

mjkkirschner: these all pass locally.

src/Libraries/DSCPython/CPythonEvaluator.cs
aparajit-pratap: Does this class need to be imported into the VM?

mjkkirschner: yes it does.

test/Libraries/DynamoPythonTests/PythonEditTests.cs
```diff
@@ -511,5 +512,107 @@ public void Python_CanReferenceDynamoServicesExecutionSession()
...
+        [Test]
+        public void TwoCPythonHandlesReturnedFromSameNodeHaveSameHandleID()
+        {
+            // open test graph
+            var examplePath = Path.Combine(TestDirectory, @"core\python", "cpythoncustomclass_returnManyInstances.dyn");
+            ViewModel.OpenCommand.Execute(examplePath);
+
+            var classdef = ViewModel.Model.CurrentWorkspace.Nodes.First(x => x.Name == "classdef");
+
+            ViewModel.HomeSpace.Run();
```
aparajit-pratap, line 554: Isn't the graph run in line 550 itself?

test/Libraries/DynamoPythonTests/PythonEditTests.cs
```diff
@@ -511,5 +512,107 @@ public void Python_CanReferenceDynamoServicesExecutionSession()
...
+            var handles = classdef.CachedValue.GetElements().Select(x=>x.Data).Cast<DynamoCPythonHandle>().ToList<DynamoCPythonHandle>();
+            var firstMemLoc = handles.First().PythonObjectID;
+            Assert.IsTrue(handles.All(x => x.PythonObjectID == firstMemLoc));
```
aparajit-pratap, line 557: So does this mean that all handles have the same ID?

mjkkirschner: It means that all handles that point to the same pyObj have the same id.

test/Libraries/DynamoPythonTests/PythonEvalTests.cs
```diff
@@ -170,30 +170,46 @@ public void CPythonEngineIncludesTraceBack()
...
             var empty = new ArrayList();
-
-            try
+            Assert.DoesNotThrow(() =>
             {
-                DSCPython.CPythonEvaluator.EvaluatePythonScript(code, empty, empty);
-                Assert.Fail("An exception was expected");
-            }
-            catch (Exception exc)
+                Assert.IsTrue(DSCPython.CPythonEvaluator.EvaluatePythonScript(code, empty, empty).ToString().Contains("weakref at"));
```
aparajit-pratap, line 190: Is `ToString` expected to return the output from `__str__`?

mjkkirschner: yes, this is what ironPython does if `toString` is not defined, which it would not be for python objects.

test/Libraries/DynamoPythonTests/PythonEditTests.cs
mjkkirschner: yes this is probably overkill, but I want to guard against this graph being made manual at some point by another dev.

mjkkirschner: if it's confusing I can remove it.

mjkkirschner: or make the graph manual, so the runs are more explicit.

aparajit-pratap: I guess there's no harm as if there is no change in the graph, it will effectively not run.

test/Libraries/DynamoPythonTests/PythonEvalTests.cs
aparajit-pratap: I guess, I'm trying to understand where the `weakref at` comes from. I know there's a weak ref being returned but still. Is it supported now?

mjkkirschner: not really, the `weakref` type in python defines a __str__ like `weak ref to blah at xyz` and we see that here.

mjkkirschner: something like this:
https://github.com/python/cpython/blob/master/Lib/weakref.py#L615

mjkkirschner: to help clarify - a consequence of this PR is that all python types that we cannot marshall we return handles for, which are mostly useless except that they show a string rep, and you can pass them to other python nodes.

aparajit-pratap: Gotcha!

test/Libraries/DynamoPythonTests/PythonEditTests.cs
QilongTang: Right because IronPythonExtension is loaded by default..

PR #10879, opened 2020-07-10
Convert Python iterable to IList and IEnumerable

mmisol:
### Purpose

Adds support to convert any Python iterable to an IList or IEnumerable
in the context of a function call using the CPython engine. The decoder
now checks if the PyObject returned is iterable instead of a sequence,
which is supported by the function we are using in Python.NET. Also,
IEnumerable, both generic and non-generic, is declared as a target type
to allow more conversion scenarios.

This should now allow converting dictionary view types, set and
frozenset to IList. It should also allow these and any other types that
are iterable (list, tuple, dictionary) to be converted to IEnumerable.

### Declarations

### Reviewers

@mjkkirschner 

### FYIs

@DynamoDS/dynamo


test/Libraries/DynamoPythonTests/PythonEvalTestsWithLibraries.cs
```diff
@@ -66,21 +68,29 @@ import clr
 
 l3 = [[1,2],[3,4]]
 # Python list (nested) => .NET IList<IList<>>
-flatennedList = List.Flatten(l3)
+flattenedList = List.Flatten(l3)
```
QilongTang, line 71: Nice catch

test/Libraries/DynamoPythonTests/PythonEvalTestsWithLibraries.cs
```diff
@@ -196,17 +206,87 @@ import clr
...
+sortedKeys = List.Sort(d)
+
+OUT = untypedDictionary, typedDictionary, sortedKeys
 ";
             var empty = new ArrayList();
-            var expected = new Dictionary<string, int> {
+            var expectedDictionary = new Dictionary<string, int> {
                 { "one", 1 }, { "two", 2 }, { "three", 3 }, { "four", 4 }
             };
+            var expectedKeys = new List<string> { "one", "three", "two" };
```
QilongTang, line 218: Is it because `List` unordered so sequence does not matter here?

mmisol: The order is alphabetic

test/Libraries/DynamoPythonTests/PythonEvalTestsWithLibraries.cs
```diff
@@ -196,17 +206,87 @@ import clr
...
+            var code = @"
+import clr
+clr.AddReference('DSCoreNodes')
+from DSCore import List
+
+d = {'one': 1, 'two': 2, 'three': 3}
+
+dk = List.AddItemToEnd('four', d.keys())
+dv = List.AddItemToEnd(4, d.values())
+di = List.AddItemToEnd(('four', 4), d.items())
```
aparajit-pratap, line 243: Did we get all this supported by simply adding the `IsIterable`?

test/Libraries/DynamoPythonTests/PythonEvalTestsWithLibraries.cs
```diff
@@ -196,17 +206,87 @@ import clr
...
+        [Test]
+        public void TestSetDecodingCPython()
+        {
+            var code = @"
+import clr
+clr.AddReference('DSCoreNodes')
+from DSCore import List
+
+s = { 'hello' }
+fs = frozenset(s)
```
aparajit-pratap, line 276: What is a frozenset?

test/Libraries/DynamoPythonTests/PythonEvalTestsWithLibraries.cs
mmisol: Yes. They were all iterable, but we were checking if they were sequences, which they are not. Since our Python.NET code supported all iterables we are good!

test/Libraries/DynamoPythonTests/PythonEvalTestsWithLibraries.cs
mmisol: It's like an immutable set.

mmisol:
Got the usual 23 failures related to the visualization tests. Should this be good to go @QilongTang ?

QilongTang:
@mmisol Sure go for it

PR #10887, opened 2020-07-13
Refresh Python Engine when IronPython loaded

QilongTang:
Please Note:

### Purpose

Per @aparajit-pratap 's suggestion

### Declarations

### Reviewers

@DynamoDS/dynamo 

### FYIs

### DIFF
--- a/src/Libraries/PythonNodeModels/Properties/AssemblyInfo.cs
+++ b/src/Libraries/PythonNodeModels/Properties/AssemblyInfo.cs
@@ -13,3 +13,4 @@ using System.Runtime.InteropServices;
 // Remove this when PythonEngineSelector become public
 [assembly: InternalsVisibleTo("DynamoPythonTests")]
 [assembly: InternalsVisibleTo("DynamoCoreWpfTests")]
+[assembly: InternalsVisibleTo("IronPythonExtension")]
--- a/src/Libraries/PythonNodeModels/Properties/AssemblyInfo.cs
+++ b/src/Libraries/PythonNodeModels/Properties/AssemblyInfo.cs
@@ -14,3 +14,4 @@ using System.Runtime.InteropServices;
 [assembly: InternalsVisibleTo("DynamoPythonTests")]
 [assembly: InternalsVisibleTo("DynamoCoreWpfTests")]
 [assembly: InternalsVisibleTo("IronPythonExtension")]
+[assembly: InternalsVisibleTo("PythonMigrationViewExtension")]
--- a/src/Libraries/PythonNodeModels/Properties/AssemblyInfo.cs
+++ b/src/Libraries/PythonNodeModels/Properties/AssemblyInfo.cs
@@ -1,4 +1,5 @@
 ﻿using System.Reflection;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
 // General Information about an assembly is controlled through the following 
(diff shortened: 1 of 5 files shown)


aparajit-pratap:
@QilongTang, I guess the best and easiest way to test this is after publishing the latest package. I wasn't able to test it very well without a package install workflow. Thanks!

src/Libraries/PythonNodeModels/Properties/AssemblyInfo.cs
```diff
@@ -12,4 +12,5 @@
 [assembly: Guid("a0debf47-0cf2-46e8-90b9-0c3d9b808fd7")]
 // Remove this when PythonEngineSelector become public
 [assembly: InternalsVisibleTo("DynamoPythonTests")]
-[assembly: InternalsVisibleTo("DynamoCoreWpfTests")]
\ No newline at end of file
+[assembly: InternalsVisibleTo("DynamoCoreWpfTests")]
+[assembly: InternalsVisibleTo("IronPythonExtension")]
```
mmisol, line 16: I don't think this helps in the cause of externalizing the extension. We should use an event and a proper API.

mjkkirschner: `RequestUpdateLoadedPythonEngines`?

aparajit-pratap: For this Core would need to be aware of the event on the extension, in order to register its scan method with it, which does not seem right. An extension is and should be aware of Core, not the other way round.

aparajit-pratap:
> An extension that is going to become external should not have access to internals of a core library.

True, but in this case, we need to make an exception as this extension is maintained by us. I think we have made similar exceptions for other extensions that are written by the Dynamo team.

QilongTang:
> > An extension that is going to become external should not have access to internals of a core library.
> 
> True, but in this case, we need to make an exception as this extension is maintained by us. I think we have made similar exceptions for other extensions that are written by the Dynamo team.

@mmisol @mjkkirschner I would like to emphasize, the Engine Selector APIs were not finalized, that is the reason why we make them internal visible to other binaries for now so we do not have to obsolete API soon. But since you insist, I've made them public with the new commit, PTAL

QilongTang:
@mmisol @mjkkirschner I think since our API scope is not clear yet, it is still part of our API now. However it is ready in its current state. We would just need to make proper changes when we need to implement dynamic loading, but if we do not get to the dynamic loading task, the current state is still good to use for extension or 3rd party devs.

mjkkirschner:
Agree with @mmisol's final statement here:

> I guess I feel the last change rushes an API that may not be what we want. If that's the case, the outcome would be worse than without these changes, because at least we would not have made anything public.
Finalizing the API was the idea I had in my mind, but if this won't be done before this PR gets merged, I'd rather take the previous change than this.

It's not clear we can finalize the API until we finalize the dynamic loading - so my vote would be to merge this in a temporary internal state and move on to that to reasses.

QilongTang:
@aparajit-pratap Are you fine with me reverting the last commit then?

aparajit-pratap:
I'd like to point out that the `PythonEngineSelector` is also being used by the PythonMigration view extension now in https://github.com/DynamoDS/Dynamo/pull/10869 and since it was initially made internal, I had to make it visible to the python migration view extension as well. I think the crux of the matter is that these extensions that we have created recently (like the IronPython and Python migration extensions) are more internal in their core functionality and that it is unlikely that third party devs would be creating extensions like these in the future and so in that sense, these are in a class apart from general-purpose extensions and view extensions.

QilongTang:
> I'd like to point out that the `PythonEngineSelector` is also being used by the PythonMigration view extension now in #10869 and since it was initially made internal, I had to make it visible to the python migration view extension as well. I think the crux of the matter is that these extensions that we have created recently (like the IronPython and Python migration extensions) are more internal in their core functionality and that it is unlikely that third party devs would be creating extensions like these in the future and so in that sense, these are in a class apart from general-purpose extensions and view extensions.

I agree

QilongTang:
@aparajit-pratap @mmisol @mjkkirschner Reverted. Let me know

PR #10889, opened 2020-07-14
Fix preview hiding

mmisol:
### Purpose

Fixes a bug which would make the preview hide even when the mouse was
moved over a visible part of it. This happened only when the preview
area was wider than the node.

The cause of the bug I would argue is some very strange behavior of
HitTest. When run on the NodeView, this test would not consider the
right part of the preview, like if the box was as wide as the node but
displaced to the left.

Likewise, running a HitTest on the PreviewControl itself provided
inconsistent results. Initially this would work as expected, but after
the preview was expanded and recondensed, the box would remain forever
bloated as if still expanded. This resulted in inconsistencies with the
leave event, which used the real margins of the control.

The problem was fixed by checking explicitly that the preview visible
area is abandoned before hiding the preview.

ezgif com-video-to-gif(2)

### Declarations

### Reviewers

@QilongTang @mjkkirschner 

### FYIs

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCoreWpf/Views/Core/NodeView.xaml.cs
+++ b/src/DynamoCoreWpf/Views/Core/NodeView.xaml.cs
@@ -490,8 +490,6 @@ namespace Dynamo.Controls
         {
             ViewModel.ZIndex = oldZIndex;
 
-            Debug.WriteLine($"Mouse over preview: {PreviewControl.IsMouseOver}");
-
             // If mouse in over node/preview control or preview control is pined, we can not hide preview control.
             if (IsMouseOver || PreviewControl.IsMouseOver || PreviewControl.StaysOpen || IsMouseInsidePreview(e) ||
                 (Mouse.Captured is DragCanvas && IsMouseInsideNodeOrPreview(e.GetPosition(this)))) return;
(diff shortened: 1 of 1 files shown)


src/DynamoCoreWpf/Views/Core/NodeView.xaml.cs
```diff
@@ -633,14 +633,35 @@ private bool IsMouseInsideNodeOrPreview(Point mousePosition)
...
+        /// <summary>
+        /// Checks whether a mouse event occurred at a position matching the preview.
+        /// Alternatives attempted:
+        /// - PreviewControl.IsMouseOver => This is always false
+        /// - HitTest on NodeView => Anomalous region that skips right part of preview if larger than node
+        /// - HitTest on Preview => Bounds become irreversible larger than "mouse over area" after preview is expanded
+        /// </summary>
+        /// <param name="e">A mouse event</param>
+        /// <returns>Whether the mouse is over the preview or not</returns>
+        private bool IsMouseInsidePreview(MouseEventArgs e)
```
aparajit-pratap, line 652: How is this method different in functionality from the one with the similar name: `IsMouseInsideNodeOrPreview`

aparajit-pratap:
Nice fix to have. This was quite frustrating.

src/DynamoCoreWpf/Views/Core/NodeView.xaml.cs
mmisol: Yes, that's right @aparajit-pratap . I know it's kinda strange for a method summary, but why it was needed was hard to explain otherwise.

mjkkirschner: do we know why the `anomalous` region exists? 😉  💀

mmisol:
@mjkkirschner I haven't found a way to do that which would allow me to specify a position. Will search a little more tomorrow morning.

mmisol:
Merging as we are doing a UI automation test for this

PR #10897, opened 2020-07-16
Dummy Python evaluator for missing engine

mmisol:
### Purpose

When building the AST for a Python node an exception is no longer
thrown when the engine is missing. Instead the class and method of a
dummy evaluator are returned. This evaluator simply returns an error
message indicating that the configured engine is missing.

Doing this allows to recover from the absence of a selected Python
engine, as evaluation errors are retried on the next run. Also, since
this is now an evaluation error, it is displayed over the node as a
warning.

### Declarations

### Reviewers

@QilongTang 

### FYIs

@DynamoDS/dynamo

### DIFF
--- a/src/Libraries/CoreNodes/CoreNodes.csproj
+++ b/src/Libraries/CoreNodes/CoreNodes.csproj
@@ -72,6 +72,7 @@
     <Compile Include="Color.cs" />
     <Compile Include="Compare.cs" />
     <Compile Include="DateTime.cs" />
+    <Compile Include="DummyPythonEvaluator.cs" />
     <Compile Include="FileSystem.cs" />
     <Compile Include="Data.cs" />
     <Compile Include="List.cs" />
--- a/src/Libraries/CoreNodes/DummyPythonEvaluator.cs
+++ b/src/Libraries/CoreNodes/DummyPythonEvaluator.cs
@@ -2,7 +2,7 @@
 using System.Collections;
 using Autodesk.DesignScript.Runtime;
 
-namespace DSCPython
+namespace DSCore
 {
     /// <summary>
     /// Python evaluator that will only throw an error on evaluation.
--- a/src/Libraries/CoreNodes/Properties/Resources.Designer.cs
+++ b/src/Libraries/CoreNodes/Properties/Resources.Designer.cs
@@ -205,7 +205,7 @@ namespace DSCore.Properties {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to The configured Python engine could not be found.
+        ///   Looks up a localized string similar to The selected Python engine could not be found.
         /// </summary>
         internal static string MissingPythonEngine {
             get {
--- a/src/Libraries/CoreNodes/Properties/Resources.Designer.cs
+++ b/src/Libraries/CoreNodes/Properties/Resources.Designer.cs
@@ -205,6 +205,15 @@ namespace DSCore.Properties {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to The configured Python engine could not be found.
+        /// </summary>
+        internal static string MissingPythonEngine {
+            get {
+                return ResourceManager.GetString("MissingPythonEngine", resourceCulture);
+            }
+        }
+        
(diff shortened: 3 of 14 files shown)


QilongTang:
hi @mmisol I would put the dummy python engine as part of PythonNodeModels project so it is shared. Also pass in the Python engine name when we initialize it like we discussed at grooming so messages are more friendly to users

mmisol:
Hi @QilongTang . If I put it there it won't be imported in the VM, right?

aparajit-pratap:
> Hi @QilongTang . If I put it there it won't be imported in the VM, right?

Good point @mmisol.

mjkkirschner:
we could add this to one of the core libraries if we want it to always be loaded.

aparajit-pratap:
I'm currently trying another approach. Will update here if we can avoid adding the Dummy class.

aparajit-pratap:
Ok, the other approach I tried was to build a special AST to signal a build AST failure, but apparently that code is dead. I agree with this approach.

QilongTang:
> > Hi @QilongTang . If I put it there it won't be imported in the VM, right?
> 
> Good point @mmisol.

@mmisol You have a very good point! @aparajit-pratap @mjkkirschner and whoever watching, what I was worried is that if in the future we make both Python engine packages, I would like the dummy evaluator loaded for any of these options. In the current approach, if DSCPython.dll is missing or not loaded, the indication for IronPython engine will be gone.. Not ideal to me, I am not sure if we want to move it to another dll, but I can anticipate the issue will always be there for the dynamic loading. Can we use DSCoreNodes.dll for now?

aparajit-pratap:
DSCoreNodes sounds fine by me.

QilongTang:
Thanks @mmisol Can you attach a picture of what it looks like right now? I am still thinking about the warning message and inclusion of the Python engine name but since that is reflected on the node lower right corner, what we have might be sufficient. 👍

mmisol:
Screen Shot 2020-07-16 at 3 58 53 PM

QilongTang:
@mmisol And you verified locally the bug you found at bug bash should not longer exist with we putting it on warning state right?

aparajit-pratap:
I would vote for `The selected python engine could not be found`

mmisol:
@QilongTang Yes, that bug is fixed by this.

@aparajit-pratap Sounds good. I'll make the change.

QilongTang:
> I would vote for `The selected python engine could not be found`

Do we want to direct users to check WSReference panel in the message here?

aparajit-pratap:
I'll have to update PR https://github.com/DynamoDS/Dynamo/pull/10869 once again after this is merged :)

aparajit-pratap:
> > I would vote for `The selected python engine could not be found`
> 
> Do we want to direct users to check WSReference panel in the message here?

The panel is expected to show only when a workspace containing ironpython references is open and ironpython is missing.

QilongTang:
> > > I would vote for `The selected python engine could not be found`
> > 
> > 
> > Do we want to direct users to check WSReference panel in the message here?
> 
> The panel is expected to show only when a workspace containing ironpython references is open and ironpython is missing.

OK. Then changes here looks good to me..

aparajit-pratap:
> > I would vote for `The selected python engine could not be found`
> 
> Do we want to direct users to check WSReference panel in the message here? I think it goes hand in hand with #10869.

Or did you mean have a link in the warning bubble that would open the WS ref panel? Is that possible?

mmisol:
The build worked here https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/708/. Merging

QilongTang:
> > > I would vote for `The selected python engine could not be found`
> > 
> > 
> > Do we want to direct users to check WSReference panel in the message here? I think it goes hand in hand with #10869.
> 
> Or did you mean have a link in the warning bubble that would open the WS ref panel? Is that possible? That might be neat too.

That should be possible but I did not mean to drag this task. I just feel like we can start with asking user to check WSReference panel somehow

PR #10900, opened 2020-07-17
[DYN-2772] Disable IronPython Alerts by Selection

zeusongit:
### Purpose

To enable users to hide IronPython alerts.

- Added 'Do not show again' checkbox to IronPython alerts.
- Added global switch option to turn off/on IronPython alerts, this setting is preserved for the user.
- Added tests

**Note:** You will probably need to enable `Python2ObsoleteMode` debug mode **before** starting the Dynamo Sandbox by editing the `debug.config` file.

SS:
Screen Shot 2020-07-17 at 12 01 41 PM

### Declarations

### Reviewers

@DynamoDS/dynamo


src/PythonMigrationViewExtension/IronPythonInfoDialog.xaml.cs
```diff
@@ -68,6 +68,14 @@ private void OnMoreInformationButtonClicked(object sender, RoutedEventArgs e)
             this.Close();
         }
 
+        private void ToggleIronPythonDialog(object sender, RoutedEventArgs e)
+        {
+            if(DisableIronPythonDialogCheck.IsChecked.HasValue && DisableIronPythonDialogCheck.IsChecked.Value)
+                ViewModel.DynamoViewModel.IsIronPythonDialogDisabled = true;
+            else
+                ViewModel.DynamoViewModel.IsIronPythonDialogDisabled = false;
```
QilongTang, line 76: Could it be null? If not, how about
```
ViewModel.DynamoViewModel.IsIronPythonDialogDisabled = DisableIronPythonDialogCheck.IsChecked;
```

src/PythonMigrationViewExtension/PythonMigrationViewExtension.cs
```diff
@@ -71,8 +71,8 @@ public void Loaded(ViewLoadedParams p)
 
         private void DisplayIronPythonDialog()
         {
-            // we only want to create the dialog ones for each graph per Dynamo session
-            if (DialogTracker.ContainsKey(CurrentWorkspace.Guid))
+            // we only want to create the dialog ones for each graph per Dynamo session, if the global setting is not disabled
+            if (DialogTracker.ContainsKey(CurrentWorkspace.Guid) || DynamoViewModel.IsIronPythonDialogDisabled)
```
QilongTang, line 75: ah.. Not really related to this task but this line might be the reason why we have bug for dyf everytime we open..

mjkkirschner: yes, tracking guid and not function id?

QilongTang: @mjkkirschner yup

zeusongit: Do you mean this bug?
https://jira.autodesk.com/browse/DYN-2936

src/PythonMigrationViewExtension/IronPythonInfoDialog.xaml.cs
zeusongit: keeping the if in place, just in case

QilongTang: With the new code, it seems we can never update the value when `DisableIronPythonDialogCheck.IsChecked.Value` is `false`? Can you verify?

zeusongit: Ah.. sorry, fixed

src/PythonMigrationViewExtension/IronPythonInfoDialog.xaml.cs
```diff
@@ -68,6 +68,11 @@ private void OnMoreInformationButtonClicked(object sender, RoutedEventArgs e)
             this.Close();
         }
 
+        private void ToggleIronPythonDialog(object sender, RoutedEventArgs e)
+        {
+            ViewModel.DynamoViewModel.IsIronPythonDialogDisabled = DisableIronPythonDialogCheck.IsChecked.Value;
```
QilongTang, line 73: 👍

src/PythonMigrationViewExtension/PythonMigrationViewExtension.cs
QilongTang: Nope this one: https://jira.autodesk.com/browse/DYN-2937. Anyway, I think out of scope..

PR #10902, opened 2020-07-17
Python class interop global

mjkkirschner:
@aparajit-pratap fixing the failing test.


test/Libraries/DynamoPythonTests/PythonEditTests.cs
```diff
@@ -579,6 +579,8 @@ public void TwoCPythonHandlesReturnedFromDifferentNodesHaveSameHandleID()
         [Test]
         public void CPythonClassCanBeReturnedAndSafelyDisposedInDownStreamNode()
         {
+            //clear the handleCountMap so other tests don't effect the counts.
+            DynamoCPythonHandle.HandleCountMap.Clear();
```
aparajit-pratap, line 583: Shouldn't we be doing this before the start of each test?

mmisol: I think a better fix would be to actually check for the handles that are relevant in the test @mjkkirschner

mjkkirschner: we definitely could, I can add it to the setup. 
Alternatively, we could clear it on opening a new home workspace or something like that.

Right now objects are just removed when they get disposed by the GC - but the entire map is never cleared. What do you think?

aparajit-pratap: Wouldn't the dispose have kicked in even in the case of these tests? If so, then we wouldn't need to explicitly clear the map here, would we?

mjkkirschner: @mmisol yes, thats a better fix. Okay, I will make the test more precise - do you two have any thoughts on clearing the handle map at a specific time? Maybe at shutdown or new workspace creation?

mjkkirschner: @aparajit-pratap  I sneakily edited my response, the relative part is:
 
```
The previous test that created this extra object was just using the evaluator directly without Dynamo at all, so the GC was not even aware.
```

mmisol: > do you two have any thoughts on clearing the handle map at a specific time? Maybe at shutdown or new workspace creation - or does it seem unnecessary/leaky.

I would think this happens already when opening a new workspace. You fixed that a short time ago, right @aparajit-pratap ? You can probably do some manual tests by placing a breakpoint and closing the workspace and opening an empty one if you are worried that this may not be working @mjkkirschner

aparajit-pratap: @mmisol that is correct. A full cleanup now happens for every new workspace created or opened.

PR #10903, opened 2020-07-17
Respect the "Show Grid" setting when 3D background preview is disabled

Dewb:
### Purpose

When the 3D background preview is disabled, a 2D grid is shown behind the node graph. This grid cannot be disabled, while the 3D coordinate system grid can be turned off by unchecking View > Background Preview > Show Grid. This PR causes this setting to control the display of both the 3D and the 2D grid.

#### UI when 3D preview is off and Show Grid is off
image

### Declarations

- [X] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.

### DIFF
--- a/src/DynamoCore/Configuration/PreferenceSettings.cs
+++ b/src/DynamoCore/Configuration/PreferenceSettings.cs
@@ -167,7 +167,7 @@ namespace Dynamo.Configuration
                 if (value == isBackgroundGridVisible) return;
                 isBackgroundGridVisible = value;
 
-                RaisePropertyChanged("IsBackgroundGridVisible");
+                RaisePropertyChanged(nameof(IsBackgroundGridVisible));
             }
         }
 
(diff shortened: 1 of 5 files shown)


src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -155,7 +156,21 @@ private BackgroundPreviewActiveState GetBackgroundPreviewData(string name)
...
+            get
+            {
+                return isBackgroundGridVisible;
+            }
+            set
+            {
+                if (value == isBackgroundGridVisible) return;
+                isBackgroundGridVisible = value;
+
+                RaisePropertyChanged("IsBackgroundGridVisible");
```
mjkkirschner, line 170: can we use `nameof`?

QilongTang:
@Dewb Would you also attach a picture of UI diff just for future reference?

src/DynamoCore/Configuration/PreferenceSettings.cs
Dewb: Good catch. Now that we're on C# 6.0 we should add a task to switch all all our  `RaisePropertyChanged("")` calls to use `nameof`.

PR #10915, opened 2020-07-21
T1.2 accept/reject code changes proposed by the MigrationAssistant

SHKnudsen:
### Purpose

This PR adds functionality to the `MigrationAssistant` which allow users to accept or reject the proposed code changes. When code changes are accepted the Python Engine  will automatically switch to `CPython3`. The code migration is part of the Dynamo undo stack, so clicking the `Revert` button will change the code back to the original state.

Mechanism for saving a complete back-up of the existing .dyn file containing Ironpython2 code is being prepared on a separate PR
 
T1 2-ChangeCodeGif

### Declarations

### Reviewers

@mmisol 
@QilongTang 

### FYIs
@mjkkirschner 
@radumg


src/PythonMigrationViewExtension/PythonMigrationViewExtension.csproj
```diff
@@ -131,6 +131,11 @@
       <Name>DynamoCore</Name>
       <Private>False</Private>
     </ProjectReference>
+    <ProjectReference Include="..\Libraries\PythonNodeModelsWpf\PythonNodeModelsWpf.csproj">
+      <Project>{01DE9B06-0BCB-4D8A-862E-E8170F5D6B4F}</Project>
+      <Name>PythonNodeModelsWpf</Name>
+      <Private>False</Private>
+    </ProjectReference>
```
mmisol, line 138: From what I could see, this dependency is not really needed. The line that casts the parent window to a specific type `var parentWindow = Window.GetWindow(btn) as ScriptEditorWindow;` is not really necessary so we can remove this.

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -199,6 +213,33 @@ protected override bool UpdateValueCore(UpdateValueParams updateValueParams)
...
+        /// Updates the Script property of the node.
+        /// NOTE: This is a temporary method used during the Python 2 to Python 3 transistion period,
+        /// it will be removed when the transistion period is over.
+        /// </summary>
+        /// <param name="newCode">The new migrated code</param>
+        internal void MigrateCode(string newCode)
+        {        
+            var e = new PythonCodeMigrationEventArgs(Script, newCode); 
+            Script = newCode;
+            if(Engine  != PythonEngineVersion.CPython3)
```
mmisol, line 226: NIT: Spacing issue here

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -199,6 +213,33 @@ protected override bool UpdateValueCore(UpdateValueParams updateValueParams)
...
+                Engine = PythonEngineVersion.CPython3;
```
mmisol, line 227: One thing I'm seeing, when the node is using IronPython2 and I migrate the code, is that the execution happens twice, first with a warning and then successfully. My suspicion is that at the point this line runs, an execution is triggered with the old code, causing this weird effect.

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
```diff
@@ -42,6 +42,7 @@ public partial class ScriptEditorWindow : ModelessChildWindow
 
             completionProvider = new SharedCompletionProvider(nodeModel.Engine,dynamoViewModel.Model.PathManager.DynamoCoreDirectory);
             completionProvider.MessageLogged += dynamoViewModel.Model.Logger.Log;
+            nodeModel.CodeMigrated += OnNodeModelCodeMigrated;
```
mjkkirschner, line 45: can you unsubscribe this at some point in `nodeView` or `nodeModel's` lifetime? 
I guess it's possible it's not required, but I have to ask for proof - with all the trouble we find with memory leaks during testing.
I like https://blog.jetbrains.com/dotnet/2014/09/04/fighting-common-wpf-memory-leaks-with-dotmemory/ (eventhandler leak section in particular)

src/Libraries/PythonNodeModels/PythonNode.cs
SHKnudsen: This was happening as we where calling `UpdateScript` twice when migrating the code, form `OnEngineChanged` and `OnNodeModelCodeMigrated` i changed this so that `OnNodeModelCodeMigrated` doesnt call `UpdateScript` but instead changes the engine which calls `OnEngineChanged`. This seems to solve it.

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
SHKnudsen: unsubscribing from this when the ScriptEditorWindow closes now, would that work?

mjkkirschner: I don't think that's a good solution because the script editor window lives for the entire nodeView lifetime I believe. That means it closes multiples times, but is only subscribed once.

mmisol: @mjkkirschner Would you say we should move the event subscription code to the window show event rather than the constructor?

mjkkirschner: that sounds like a good place to put it - or alternatively, clean it up when the nodeView is disposed.

SHKnudsen: I looked at this and it seems like its removing the reference to the `ScriptEditorWindow` when the window is closed and creating a new instance when you double click the node again, or am i missing something?

```
        private void View_MouseDown(object sender, MouseButtonEventArgs e)
        {
            if (e.ClickCount >= 2)
            {
                EditScriptContent(null, null);
                e.Handled = true;
            }
        }

        public void editWindow_Closed(object sender, EventArgs e)
        {
            editWindow = null;
        }

        private void EditScriptContent(object sender, EventArgs e)
        {
            using (var cmd = Dynamo.Logging.Analytics.TrackCommandEvent("PythonEdit"))
            {
                if (editWindow != null)
                {
                    editWindow.Activate();
                }
                else
                {
                    editWindow = new ScriptEditorWindow(dynamoViewModel, pythonNodeModel, pythonNodeView, ref editorWindowRect);
                    editWindow.Initialize(workspaceModel.Guid, pythonNodeModel.GUID, "ScriptContent", pythonNodeModel.Script);
                    editWindow.Closed += editWindow_Closed;
                    editWindow.Show();
                }
            }
        }
```
is there a good way to know when the nodeView is disposed, could only see a way of knowing when the nodeModel is disposed which i guess will also dispose the view?

mjkkirschner: thanks for verifying @SHKnudsen - I think you're right! it looks like the scriptEditor is constructed on edit, if thats the case, I think your change for clearing on window close is good. Let's do that.

Sorry for the confusion.

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
```diff
@@ -193,5 +203,10 @@ private void OnEngineChanged(object sender, System.Windows.Controls.SelectionCha
                 UpdateScript(editText.Text);
             }
         }
+
+        private void OnScriptEditorWindowClosed(object sender, EventArgs e)
+        {
+            nodeModel.CodeMigrated -= OnNodeModelCodeMigrated;
```
mjkkirschner, line 209: I know it might seem overkill, but can you also unsubscribe the Closing event from this same method?

PR #10924, opened 2020-07-22
Allow printing in Python nodes with DynamoPrint

mmisol:
### Purpose

Adds basic logging functionality to Python node and Python String node.
Logging is done to the Dynamo console using a function added to the
Python scopes named DynamoPrint. The function prefixes the node name to
the log entry. These log entries are only added to the ConsoleLog.

Screen Shot 2020-07-22 at 12 48 08 PM

### Declarations

### Reviewers

@mjkkirschner @QilongTang 

### FYIs

@SHKnudsen @radumg @Amoursol

### DIFF
--- a/src/Libraries/DSCPython/CPythonEvaluator.cs
+++ b/src/Libraries/DSCPython/CPythonEvaluator.cs
@@ -192,16 +192,14 @@ namespace DSCPython
                     }
                     using (PyScope scope = Py.CreateScope())
                     {
+                        ProcessAdditionalBindings(scope, bindingNames, bindingValues);
+
                         int amt = Math.Min(bindingNames.Count, bindingValues.Count);
 
                         for (int i = 0; i < amt; i++)
                         {
                             scope.Set((string)bindingNames[i], InputMarshaler.Marshal(bindingValues[i]).ToPython());
                         }
-
-                        dynamic logger = ExecutionEvents.ActiveSession.GetParameterValue(ParameterKeys.Logger);
-                        Action<string> logFunction = msg => logger.Log($"USER: {msg}");
-                        scope.Set("DynamoPrint", logFunction.ToPython());
 
                         try
                         {
@@ -235,6 +233,31 @@ namespace DSCPython
             {
                 PythonEngine.ReleaseLock(gs);
             }
+        }
+
+        /// <summary>
+        /// Processes additional bindings that are not actual inputs.
+        /// Currently, only the node name is received in this way.
+        /// </summary>
+        /// <param name="scope">Python scope where execution will occur</param>
+        /// <param name="bindingNames">List of binding names received for evaluation</param>
+        /// <param name="bindingValues">List of binding values received for evaluation</param>
+        private static void ProcessAdditionalBindings(PyScope scope, IList bindingNames, IList bindingValues)
+        {
+            string nodeName;
+            if (!bindingNames[0].Equals("Name"))
+            {
+                // Defensive code to fallback in case the additional binding is not there for some reason
+                nodeName = "USER";
+            }
+            else
+            {
+                bindingNames.RemoveAt(0);
+                nodeName = (string)bindingValues[0];
+            }
+            dynamic logger = ExecutionEvents.ActiveSession.GetParameterValue(ParameterKeys.Logger);
+            Action<string> logFunction = msg => logger.Log($"{nodeName}: {msg}");
+            scope.Set("DynamoPrint", logFunction.ToPython());
         }
 
         /// <summary>
--- a/src/Libraries/DSCPython/CPythonEvaluator.cs
+++ b/src/Libraries/DSCPython/CPythonEvaluator.cs
@@ -244,8 +244,9 @@ namespace DSCPython
         /// <param name="bindingValues">List of binding values received for evaluation</param>
         private static void ProcessAdditionalBindings(PyScope scope, IList bindingNames, IList bindingValues)
         {
+            const string NodeNameInput = "Name";
             string nodeName;
-            if (bindingNames.Count == 0 || !bindingNames[0].Equals("Name"))
+            if (bindingNames.Count == 0 || !bindingNames[0].Equals(NodeNameInput))
             {
                 // Defensive code to fallback in case the additional binding is not there, like
                 // when the evaluator is called directly in unit tests.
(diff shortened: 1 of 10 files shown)


src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -231,6 +235,32 @@ static CPythonEvaluator()
...
+        /// Processes additional bindings that are not actual inputs.
+        /// Currently, only the node name is received in this way.
+        /// </summary>
+        /// <param name="scope">Python scope where execution will occur</param>
+        /// <param name="bindingNames">List of binding names received for evaluation</param>
+        /// <param name="bindingValues">List of binding values received for evaluation</param>
+        private static void ProcessAdditionalBindings(PyScope scope, IList bindingNames, IList bindingValues)
+        {
+            string nodeName;
+            if (!bindingNames[0].Equals("Name"))
```
mjkkirschner, line 248: can we extract this to a const somewhere?

src/Libraries/DSIronPython/IronPythonEvaluator.cs
```diff
@@ -148,6 +152,32 @@ private static string PythonStandardLibPath()
...
         }
 
+        /// <summary>
+        /// Processes additional bindings that are not actual inputs.
+        /// Currently, only the node name is received in this way.
+        /// </summary>
+        /// <param name="scope">Python scope where execution will occur</param>
+        /// <param name="bindingNames">List of binding names received for evaluation</param>
+        /// <param name="bindingValues">List of binding values received for evaluation</param>
+        private static void ProcessAdditionalBindings(ScriptScope scope, IList bindingNames, IList bindingValues)
```
mjkkirschner, line 162: is this code exactly the same? Does it make sense to add it to PythonNodeModel? I guess it's small enough the duplication is ok.

mjkkirschner: ah, I see it's a `scriptScope`

Dewb:
Is the choice to introduce a new function, rather than capturing output of `print()`, motivated by user concerns or by implementation concerns?

src/Libraries/DSCPython/CPythonEvaluator.cs
mmisol: I could define a const, but where do I put it? DSCPython and DSIronPython don't have anything in common now with PythonNodeModels...

src/Libraries/DSIronPython/IronPythonEvaluator.cs
mmisol: Not exactly the same. Adding something to the scope works different for each engine.

mmisol:
Hi @Dewb . I agree with your observation given the current status of this PR. I could take a look at changing the behavior of `print` as well, even if it was not part of the internal ticket. What do you think @mjkkirschner ?

I think the plan for the future is to make `DynamoPrint` accept a color argument that will colorize the output text. That justifies it's existence, although it may also be as you say, a good way to expose that function to customers.

mjkkirschner:
The original reason to create a new function was to avoid any unforeseen consequences of suddenly logging all print statements to the dynamo console. We don't know how many print statements may be scattered in user code, or in library code, and the dynamo console is not well profiled, and potentially not designed for constant logging.

We also did not want to pollute the log files we use for debugging of user issues with dynamo statements, though that can be solved in either case.

Third was the addition of future improvements for color or tagging capability.

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -245,9 +245,10 @@ static CPythonEvaluator()
         private static void ProcessAdditionalBindings(PyScope scope, IList bindingNames, IList bindingValues)
         {
             string nodeName;
-            if (!bindingNames[0].Equals("Name"))
+            if (bindingNames.Count == 0 || !bindingNames[0].Equals("Name"))
```
QilongTang, line 248: Would you define the "Name" here as a const string or something, it seems we do not have reference to node model here so seems no way to reference the property itself.

mmisol: Sure, I can do that.

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -231,6 +235,39 @@ static CPythonEvaluator()
...
+                nodeName = (string)bindingValues[0];
+                bindingValues.RemoveAt(0);
+            }
+
+            // Session is null when running unit tests.
+            if (ExecutionEvents.ActiveSession != null)
+            {
+                dynamic logger = ExecutionEvents.ActiveSession.GetParameterValue(ParameterKeys.Logger);
+                Action<string> logFunction = msg => logger.Log($"{nodeName}: {msg}");
+                scope.Set("DynamoPrint", logFunction.ToPython());
```
mjkkirschner, line 267: is `ToPython()` potentially slow? Does it make sense to cache this?

mmisol: It's marshalling/encoding the function so it can be used in Python. It's the same thing we are doing for all inputs, so I wouldn't say it's particularly slow.

mjkkirschner: okay, sounds good, I was just curious if function marshaling was slower than other types since we do this for each invocation of the python node.

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -231,6 +235,39 @@ static CPythonEvaluator()
...
         }
 
+        /// <summary>
+        /// Processes additional bindings that are not actual inputs.
+        /// Currently, only the node name is received in this way.
+        /// </summary>
+        /// <param name="scope">Python scope where execution will occur</param>
+        /// <param name="bindingNames">List of binding names received for evaluation</param>
+        /// <param name="bindingValues">List of binding values received for evaluation</param>
+        private static void ProcessAdditionalBindings(PyScope scope, IList bindingNames, IList bindingValues)
```
QilongTang, line 245: The two function body here looks very similar, are they identical? Can we somehow move the shared part to PythonNodeModels project?

mmisol: I think the key part is the `somehow`. A month ago doing that would have been easy, but now the engines don't reference `PythonNodeModels` anymore. I guess we can revisit this as something to take into account when factoring out the engines and trying to re-use code between them. What do you think @QilongTang ?

QilongTang: Ah sorry I keep forgetting about that we did so many changes to the projects.. Sure, let's consider this later

PR #10925, opened 2020-07-23
Opening graph with custom nodes from missing packages correctly loads in Dynamo.

reddyashish:
### Purpose

The regression mentioned in https://jira.autodesk.com/browse/DYN-2936 is already fixed by @aparajit-pratap here in this PR: 
https://github.com/DynamoDS/Dynamo/pull/10869/files#diff-f13147450a5b0038bab03d9f6493aff8R100

The reason it was failing was because if the package that contains the custom node is missing, the custom manager will not have the CN information and in that case we just want to skip it. In that case, it wont be shown in the PythonMigration dialog even if the CN has IronPython2 nodes in it. Users will need to resolve the custom nodes first from WorkspaceDependency viewer and then when they switch to that workspace, the dialog will be shown.

I am just adding a test for that regression in this PR as the fix is already in master. 

### Declarations

### Reviewers

### FYIs

@DynamoDS/dynamo

### DIFF
--- a/test/DynamoCoreWpfTests/ViewExtensions/PythonMigrationViewExtensionTests.cs
+++ b/test/DynamoCoreWpfTests/ViewExtensions/PythonMigrationViewExtensionTests.cs
@@ -353,6 +353,6 @@ namespace DynamoCoreWpfTests
             Assert.AreEqual(GraphPythonDependencies.PythonPackage, pkgDependencyInfo.Name);
             Assert.AreEqual(GraphPythonDependencies.PythonPackageVersion, pkgDependencyInfo.Version);
         }
-    }
 
+    }
 }
--- a/test/DynamoCoreWpfTests/ViewExtensions/PythonMigrationViewExtensionTests.cs
+++ b/test/DynamoCoreWpfTests/ViewExtensions/PythonMigrationViewExtensionTests.cs
@@ -354,4 +354,5 @@ namespace DynamoCoreWpfTests
             Assert.AreEqual(GraphPythonDependencies.PythonPackageVersion, pkgDependencyInfo.Version);
         }
     }
+
 }
--- a/test/DynamoCoreWpfTests/ViewExtensions/PythonMigrationViewExtensionTests.cs
+++ b/test/DynamoCoreWpfTests/ViewExtensions/PythonMigrationViewExtensionTests.cs
@@ -353,6 +353,5 @@ namespace DynamoCoreWpfTests
             Assert.AreEqual(GraphPythonDependencies.PythonPackage, pkgDependencyInfo.Name);
             Assert.AreEqual(GraphPythonDependencies.PythonPackageVersion, pkgDependencyInfo.Version);
         }
-
     }
 }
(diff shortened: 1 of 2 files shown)


test/DynamoCoreWpfTests/ViewExtensions/PythonMigrationViewExtensionTests.cs
```diff
@@ -353,6 +353,5 @@ public void IronPythonPackageLoadedTest()
             Assert.AreEqual(GraphPythonDependencies.PythonPackage, pkgDependencyInfo.Name);
             Assert.AreEqual(GraphPythonDependencies.PythonPackageVersion, pkgDependencyInfo.Version);
         }
-
```
aparajit-pratap, line 356: @reddyashish could you simply revert this file if it isn't involved in this PR.

aparajit-pratap:
@reddyashish does this mean that this bug is not reproducible in master and that the node positions as well as workspace dependency viewer work as expected?

reddyashish:
@aparajit-pratap yes it is fixed in master. Just to make sure, you need to enable the "Python2ObsoleteMode" debug-mode to reproduce this regression. I think that is the reason Mike was not able to reproduce it during the bug bash.

test/DynamoCoreWpfTests/ViewExtensions/PythonMigrationViewExtensionTests.cs
reddyashish: Yes, just thought I would clean this up as well because of the unnecessary extra new line.

test/DynamoCoreWpfTests/ViewExtensions/WorkspaceDependencyViewExtensionTests.cs
```diff
@@ -236,9 +236,18 @@ public void TestPropertiesWithCodeInIt()
...
+        public void VerifyDynamoLoadingOnOpeningWorkspaceWithMissingCustomNodes()
+        {
+            DebugModes.LoadDebugModesStatusFromConfig(Path.Combine(GetTestDirectory(ExecutingDirectory), "DynamoCoreWpfTests", "python2ObsoleteMode.config"));
+            DynamoModel.IsTestMode = false;
+            var examplePath = Path.Combine(@"core\packageDependencyTests\PackageDependencyStates.dyn");
+            Open(examplePath);
+            Assert.AreEqual(1, View.ExtensionTabItems.Count);
+            var workspaceViewExtension = View.viewExtensionManager.ViewExtensions.Where(x => x.Name.Equals("Workspace References")).Count();
+            Assert.AreEqual(1, workspaceViewExtension);
+        }
```
aparajit-pratap, line 251: @reddyashish can you also verify the name(s) of the missing dependencies that are displayed in WS references view?

reddyashish: yes, modified the test.

reddyashish:
Merging this.

PR #10927, opened 2020-07-23
Custom Node containing IronPython nodes shows dialog alert once per session

zeusongit:
### Purpose

Do not let IronPython dialog alerts show more than once for a custom node in a session.
Used Custom Node ID to identify previously opened nodes.

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/PythonMigrationViewExtension/PythonMigrationViewExtension.cs
+++ b/src/PythonMigrationViewExtension/PythonMigrationViewExtension.cs
@@ -73,7 +73,8 @@ namespace Dynamo.PythonMigration
         private void DisplayIronPythonDialog()
         {
             // we only want to create the dialog ones for each graph per Dynamo session, if the global setting is not disabled
-            if (DynamoViewModel.IsIronPythonDialogDisabled || DialogTracker.ContainsKey(CurrentWorkspace.Guid) || DialogTracker.ContainsKey((CurrentWorkspace as CustomNodeWorkspaceModel).CustomNodeId))
+            if (DynamoViewModel.IsIronPythonDialogDisabled || DialogTracker.ContainsKey(CurrentWorkspace.Guid)) return;
+            if (CurrentWorkspace is CustomNodeWorkspaceModel && DialogTracker.ContainsKey((CurrentWorkspace as CustomNodeWorkspaceModel).CustomNodeId))
                 return;
 
             var dialog = new IronPythonInfoDialog(this)
@@ -87,7 +88,7 @@ namespace Dynamo.PythonMigration
             }), DispatcherPriority.Background);
 
             DialogTracker[CurrentWorkspace.Guid] = dialog;
-            if (CurrentWorkspace.GetType() == typeof(CustomNodeWorkspaceModel)){
+            if (CurrentWorkspace is CustomNodeWorkspaceModel){
                 DialogTracker[(CurrentWorkspace as CustomNodeWorkspaceModel).CustomNodeId] = dialog;
             }
         }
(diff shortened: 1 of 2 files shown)


mjkkirschner:
@zeusongit looks like there are some test failures now.
https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/1131/

src/PythonMigrationViewExtension/PythonMigrationViewExtension.cs
```diff
@@ -73,7 +73,7 @@ public void Loaded(ViewLoadedParams p)
         private void DisplayIronPythonDialog()
         {
             // we only want to create the dialog ones for each graph per Dynamo session, if the global setting is not disabled
-            if (DialogTracker.ContainsKey(CurrentWorkspace.Guid) || DynamoViewModel.IsIronPythonDialogDisabled)
+            if (DynamoViewModel.IsIronPythonDialogDisabled || DialogTracker.ContainsKey(CurrentWorkspace.Guid) || DialogTracker.ContainsKey((CurrentWorkspace as CustomNodeWorkspaceModel).CustomNodeId))
```
mjkkirschner, line 76: can you make these lines shorter.

src/PythonMigrationViewExtension/PythonMigrationViewExtension.cs
```diff
@@ -87,6 +87,9 @@ private void DisplayIronPythonDialog()
             }), DispatcherPriority.Background);
 
             DialogTracker[CurrentWorkspace.Guid] = dialog;
+            if (CurrentWorkspace.GetType() == typeof(CustomNodeWorkspaceModel)){
```
mjkkirschner, line 90: you can just do `is`

test/DynamoCoreWpfTests/ViewExtensions/PythonMigrationViewExtensionTests.cs
```diff
@@ -334,6 +334,47 @@ public void CustomNodeContainsIronPythonDependencyTest()
...
         }
 
+        /// <summary>
+        /// This test verifies that the IronPython dialog wont show the second time a custom node is opened
+        /// even if it contains IronPython nodes
+        /// </summary>
+        [Test]
+        public void WillNotDisplayDialogWhenOpeningCustomNodeWithIronPythonNodesSecondTimeInSameSession()
+        {
+            DebugModes.LoadDebugModesStatusFromConfig(Path.Combine(GetTestDirectory(ExecutingDirectory), "DynamoCoreWpfTests", "python2ObsoleteMode.config"));
```
mjkkirschner, line 344: maybe your test should disable this config mode before the next test?

zeusongit: I guess each test start its own instance of the sandbox so this should not be a problem, also, it is not disabled for any other test, before it as well which are passing.

mjkkirschner: Each test does not start it's own instance of sandbox - they run in two different scenarios.

on master-15 all tests run in a single process.
on master-5 there are some number of parallel processes with multiple tests inside, which tests is not deterministic.

zeusongit: ok, got it.

zeusongit: The problem was with the casting of CurrentWS into CustomNodeWS. Still have to figure out the other 2 failing tests, looking into those.

zeusongit: All test pass now

PR #10928, opened 2020-07-23
T1.3 save python2 backup file

SHKnudsen:
### Purpose

This PR adds functionality to the MigrationAssistant which will save a complete backup of the dyn as `{dynName}.Python2.dyn`. This will only happen the first time the migration assistant is used to update code. As long as there is a `{dynName}.Python2.dyn` file in the backup folder no additional python 2 backups will be saved.

T1 3-SavePython2Backup

### Declarations

### Reviewers

@mmisol 

### FYIs

@mjkkirschner 
@QilongTang 
@radumg


src/PythonMigrationViewExtension/PythonMigrationAssistantViewModel.cs
```diff
@@ -23,7 +27,25 @@ private void MigrateCode()
...
+                return;
+
+            DynamoViewModel.SaveAs(path, true);
+        }
+
+        private string GetPythonMigrationBackupPath()
+        {
+            var workspaceName = DynamoViewModel.CurrentSpace.Name;
+            var backupDirectory = DynamoViewModel.Model.PathManager.BackupDirectory;
+            var fileName = Path.Combine(backupDirectory, workspaceName) + string.Concat(".", Properties.Resources.PythonMigrationBackupExtension, ".dyn");
```
mmisol, line 47: I guess for custom nodes we would want to keep the extension `dyf`.

src/PythonMigrationViewExtension/PythonMigrationAssistantViewModel.cs
```diff
@@ -23,7 +27,25 @@ private void MigrateCode()
...
             PythonNode.MigrateCode(NewCode);
         }
+
+        private void SavePythonMigrationBackup()
+        {
+            var path = GetPythonMigrationBackupPath();
+            if (File.Exists(path))
+                return;
+
+            DynamoViewModel.SaveAs(path, true);
```
mmisol, line 40: At this point, would it make sense to somehow notify the user that a backup was created? Otherwise they might never realize of their existence. What do you think @Amoursol ?

Amoursol: Definitely notify users, and we should make sure it persists between versions also.

radumg: How would we like to notify users ? 

1. simple message box 
2. custom notification (like those on Dynamo startup)
3. Dynamo notification system (does anyone see or look at those ?)

also @Amoursol , what do you mean by `persist between versions` ?

mjkkirschner:
I look at them. 😢

src/PythonMigrationViewExtension/PythonMigrationAssistantViewModel.cs
```diff
@@ -8,11 +10,13 @@ internal class PythonMigrationAssistantViewModel
         public string OldCode { get; set; }
         public string NewCode { get; set; }
         private PythonNode PythonNode { get; set; }
+        public DynamoViewModel DynamoViewModel { get; set; }
 
-        public PythonMigrationAssistantViewModel(PythonNode pythonNode)
+        public PythonMigrationAssistantViewModel(PythonNode pythonNode, DynamoViewModel dynamoViewModel)
```
mjkkirschner, line 15: This change worries me.

1. it would be best not to change a public method - even if I wouldn't really consider this the API.
2. Worse, we should not be using access to the viewModel if we don't need it.

I would really appreciate instead expanding the extensions API instead to allow saving of graphs... and I'm pretty sure thats already possible using commands.

src/PythonMigrationViewExtension/PythonMigrationAssistantViewModel.cs
mjkkirschner: The pathmanager can already be accessed without using the viewModel !!

src/PythonMigrationViewExtension/PythonMigrationViewExtension.csproj
```diff
@@ -37,6 +37,9 @@
     <Reference Include="DiffPlex.Wpf, Version=1.1.1.0, Culture=neutral, PublicKeyToken=1d35e91d1bd7bc0f, processorArchitecture=MSIL">
       <HintPath>..\packages\DiffPlex.Wpf.1.1.1\lib\net46\DiffPlex.Wpf.dll</HintPath>
     </Reference>
+    <Reference Include="Microsoft.Practices.Prism, Version=4.1.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35">
```
mjkkirschner, line 40: where is this coming from?

src/PythonMigrationViewExtension/PythonMigrationAssistantViewModel.cs
Amoursol: @radumg Can we look to add a `simple message box` that states to the user a message stating that it has been created and it's location (On the proviso they want to delete at some point). 

In terms of persist between versions: If a user uses Dynamo 2.7 Sandbox, then downloads 2.8 Sandbox then we should honour their choices and note this in the message above.

mjkkirschner: Are there any choices to honor here? Also moving preferences between 2.7 and 2.8 etc - is a whole other can of worms.

src/PythonMigrationViewExtension/PythonMigrationAssistantViewModel.cs
radumg: yep, looking at this now

src/PythonMigrationViewExtension/PythonMigrationAssistantViewModel.cs
radumg: added handling for `dyf` in e9c4d5506205e7b494eecbde2b12d13eab902a7a, will look at `PathManager` next

src/PythonMigrationViewExtension/PythonMigrationAssistantViewModel.cs
radumg: if we can avoid that 🥫🐍 then we should 😁 
Don't think there's anything to persist between versions anyway, backup location is version-agnostic. Unless you mean disabling this backup notification if they've also disabled the other migration warning?

this is what it looks like now :
image

The text is as below, with `{0}` replaced with path.
> Before applying the Python code migration, a backup of this graph has been saved to the configured backup location : {0}

src/PythonMigrationViewExtension/PythonMigrationAssistantViewModel.cs
radumg: @mjkkirschner I've changed the migration assistant viewmodel to depend on `WorkspaceModel` and `IPathManager` not `DynamoViewModel`

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -216,8 +216,9 @@ protected override bool UpdateValueCore(UpdateValueParams updateValueParams)
             return base.UpdateValueCore(updateValueParams);
         }
 
+        [Obsolete]
```
mjkkirschner, line 219: obsolete?

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
```diff
@@ -33,43 +33,52 @@ internal GraphPythonDependencies(ViewLoadedParams viewLoadedParams)
...
+            return evaluatorClass == PythonEngineSelector.Instance.IronPythonEvaluatorClass
+                && evaluationMethod == PythonEngineSelector.Instance.IronPythonEvaluationMethod;
         }
 
-        internal bool ContainsIronPythonDependencyInCurrentWS()
+        /// <summary>
+        /// Determines if the current workspace has any dependencies on IronPython engine.
+        /// </summary>
+        /// <returns>True if depencies are found, false otherwise.</returns>
+        internal bool CurrentWorkspaceHasIronPythonDepency()
```
mjkkirschner, line 49: @reddyashish please review these changes.

mjkkirschner, line 49: spelling

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
```diff
@@ -81,13 +90,17 @@ internal IEnumerable<INodeLibraryDependencyInfo> AddPythonPackageDependency()
             return new[] { packageDependencyInfo };
         }
 
-        // This function returns true, if any of the custom nodes in the input list has an IronPython dependency. 
-        // It traverses all CN's in the given list of customNodes and marks them as true in CustomNodePythonDependency, 
-        // if the prarent custom node or any of its child custom nodes contain an IronPython dependency.
+        /// <summary>
+        /// This recursive function returns true if any of the custom nodes in the input list has an IronPython dependency.
```
mjkkirschner, line 94: this seems to be a big change in behavior? Is the diff wrong? Why these changes?

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
radumg: 🙈

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
radumg: for future reference :
found it to be doing unnecessary cycles (recursively too!) to just say `yay` or `nay` at the end, but as discussed, the full dependency map dictionary is actually needed for displaying the affected nodes in the warning modal.

radumg: I'll revert behaviour to how it was

src/Libraries/PythonNodeModels/PythonNode.cs
radumg: as a a temporary feature, this will be removed, did this so it raises warnings in the compiler and in VS Intellisense, so anyone using this API will know not to depend on it and that it's already planned to go away.

wanted to ask you about it, let me know if you prefer with the attribute or without.

mjkkirschner: hmm, thats a good point - for now would you mind adding a small message indicating this API will only exist for a short time.

src/PythonMigrationViewExtension/PythonMigrationViewExtension.csproj
radumg: no idea, wasn't showing up in NuGet packages list in VS either, removed it and builds just fine.

src/Libraries/PythonNodeModels/PythonNode.cs
radumg: done 😄

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
radumg: will revert the behaviour as part of other comment below, but please have a second sanity check @reddyashish

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
radumg: - `IsIronPythonPackageLoaded` - logic is the same
- `ContainsIronPythonDependencyInCurrentWS` (now `CurrentWorkspaceHasIronPythonDependency`) - reverted to original logic
-

radumg:
> something along the lines of opening a file, migrating a node and checking that the backup is where it should be, would be a good test to have I guess. Thoughts @SHKnudsen @mjkkirschner ?

last comment to address - we could add this now, one question i would have is around the fact the migration now produces a `MessageBox` to warn user that a backup file has been created. 

Would this interfere with the test you think (before i delve into it) ?

mmisol:
@radumg I'm afraid opening a `MessageBox` could be a problem, but maybe there is a way to close it from code? I know you guys have created some tests in the past that deal with windows, like for the Python warnings. Not sure if something similar can be done.
You might be also be able to use `Moq` in your test to override the part of the class that shows the dialog, and do an assertion instead.
On a related track, if `MessageBox` is not our final choice, we might decide to leave this for later, and maybe the test will be simpler. For instance, if we chose to use the Dynamo notifications that appear on the top right instead of the `MessageBox`, then the problem would no longer exist. Copying @Amoursol to get his opinion regarding this last part.

radumg:
>  I know you guys have created some tests in the past that deal with windows, like for the Python warnings. Not sure if something similar can be done.
Think those were easier since we could get control of that as `Window`.

> You might be also be able to use `Moq` in your test to override the part of the class that shows the dialog, and do an assertion instead.
Those methods are `private` for now and non-virtual, not sure exposing them more just for the test would be great?

> On a related track, if `MessageBox` is not our final choice, we might decide to leave this for later, and maybe the test will be simpler. For instance, if we chose to use the Dynamo notifications that appear on the top right instead of the `MessageBox`, then the problem would no longer exist. Copying @Amoursol to get his opinion regarding this last part.
Defo, happy to come back to this one later on.

mmisol:
>Those methods are private for now and non-virtual, not sure exposing them more just for the test would be great?

You can do that with `internal virtual` methods. Here is an example: https://github.com/DynamoDS/Dynamo/blob/master/src/VisualizationTests/Preview3DMemoryOutageTest.cs#L35.
For that to work you need to use `[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")]`, so that `Moq` can access the `internal` members in that assembly.

Amoursol:
To confirm @mmisol @radumg - Creating a `Messagebox` that tells the user we are saving a backup of their Python2 graph causes issues with testing?

mjkkirschner:
for many message boxes that are modal we sometimes use `isTestMode` as a way to turn these off during testing.

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
```diff
@@ -81,12 +94,17 @@ internal IEnumerable<INodeLibraryDependencyInfo> AddPythonPackageDependency()
...
         }
 
-        // This function returns true, if any of the custom nodes in the input list has an IronPython dependency. 
-        // It traverses all CN's in the given list of customNodes and marks them as true in CustomNodePythonDependency, 
-        // if the prarent custom node or any of its child custom nodes contain an IronPython dependency.
-        private static bool CustomNodesContainIronPythonDependency(IEnumerable<Function> customNodes, ICustomNodeManager customNodeManager)
+        /// <summary>
+        /// This recursive function returns true if any of the custom nodes in the input list has an IronPython dependency.
+        /// Any custom nodes in the input list traversed during evalution have their dependencies cached in <see cref="CustomNodePythonDependencyMap"/>.
+        /// Custom odes are found to depend on IronPython if the parent custom node or any of its child custom nodes contain an IronPython dependency.
```
mjkkirschner, line 100: `Custom odes` spelling

src/PythonMigrationViewExtension/MigrationAssistant/PythonMigrationAssistantViewModel.cs
```diff
@@ -0,0 +1,62 @@
...
+using Dynamo.Interfaces;
+using PythonNodeModels;
+using System.IO;
+using System.Windows;
+
+namespace Dynamo.PythonMigration.MigrationAssistant
+{
+    internal class PythonMigrationAssistantViewModel
+    {
+        public string OldCode { get; set; }
```
mjkkirschner, line 11: do these need public setters?

src/PythonMigrationViewExtension/MigrationAssistant/PythonMigrationAssistantViewModel.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            MigrateCode();
+        }
+
+        private void MigrateCode()
+        {
+            this.NewCode = ScriptMigrator.MigrateCode(this.OldCode);
+        }
+
+        public void ChangeCode()
```
mjkkirschner, line 34: /// comment please

src/PythonMigrationViewExtension/MigrationAssistant/ScriptMigrator.cs
```diff
@@ -39,7 +39,7 @@ internal static string MigrateCode(string code)
                         scope.Exec(GetPythonMigrationScript());
 
                         var result = scope.Contains(RETURN_NAME) ? scope.Get(RETURN_NAME) : null;
-                        return result.ToString();
+                        return result?.ToString();
```
mjkkirschner, line 42: should this return an empty string if result is null?

src/PythonMigrationViewExtension/PythonMigrationViewExtension.cs
```diff
@@ -120,7 +121,7 @@ private void OnMigrationAssistantRequested(object sender, EventArgs e)
             var parentWindow = Window.GetWindow(btn);
 
             var node = sender as PythonNode;
-            var viewModel = new PythonMigrationAssistantViewModel(node);
+            var viewModel = new PythonMigrationAssistantViewModel(node, DynamoViewModel.CurrentSpace, DynamoViewModel.Model.PathManager);
```
mjkkirschner, line 124: @radumg you should be able to get both the current workspace and the path manager from the LoadedParams without referencing the DynamoViewModel.

mjkkirschner:
Screen Shot 2020-07-30 at 12 36 30 PM

7 failing tests.

mmisol:
@Amoursol Yes, but there are workarounds to not show them during the tests, which I think we could apply in the context of this PR. @radumg If you can please take a look at it, as well as the build failures and Mike's last comments. Thanks!

radumg:
When we fix the above, also move the `?` button to be next to `2->3` one :
https://github.com/DynamoDS/Dynamo/pull/10930#issuecomment-666482548

src/PythonMigrationViewExtension/MigrationAssistant/ScriptMigrator.cs
radumg: note : result?.ToString() ?? string.Empty;

radumg: var result = scope.Contains(RETURN_NAME) ? scope.Get(RETURN_NAME) : string.Empty;

SHKnudsen:
@mmisol fixed all of @mjkkirschner comments and added a test to check that the backup function works as expected. Also moved the "More Information" button on the ScriptEditor window so it now sits next to the 2to3 button.
T1 3-MoveMoreInfoButton

src/PythonMigrationViewExtension/MigrationAssistant/PythonMigrationAssistantViewModel.cs
```diff
@@ -8,8 +8,8 @@ namespace Dynamo.PythonMigration.MigrationAssistant
 {
     internal class PythonMigrationAssistantViewModel
     {
-        public string OldCode { get; set; }
-        public string NewCode { get; set; }
+        public string OldCode { get; private set; }
```
mjkkirschner, line 11: Sorry guys, I know this is obvious, but it's a public getter, can we please add /// comments.

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
aparajit-pratap: Is there any reason why this class has been touched? Are these changes relevant to the work for this PR?

mjkkirschner:
let's wait for the final test run, looks good though.

src/PythonMigrationViewExtension/GraphPythonDependencies.cs
aparajit-pratap: I see you have reverted most changes and only done some code cleanup preserving the logic. 👍

PR #10932, opened 2020-07-23
T1.5-Modify the VisualDifferenceViewer so text wraps instead of using scrollviewers

SHKnudsen:
# Based on PR #10928 , please merge that first

### Purpose

This PR addresses the wish to have the text in the Code Difference viewer wrap instead of using scrollviewers, to make it more in style with the Python script editor. I have looked into modifying the `DiffPlex` viewer but couldn't find a way to do this. This PR creates new views that behaves and looks like the original DiffPlex viewer but adds wrapping to the text.

The caveat of this approach is that it adds a lot more code than using the DiffPlex viewer. We still need the Dependence on DiffPlex but we can remove the reference to DiffPlex.Wpf

T1 5-DiffViewerWithTextWrap

### Declarations

### Reviewers

@mmisol 

### FYIs

@mjkkirschner 
@QilongTang 
@Amoursol 
@radumg


mjkkirschner:
why is wrapping preferable?  Comparing line by line seems more useful...

mmisol:
Copying @Amoursol . There is a long story to this discussion, starting from PR number 1 and at least two zoom meetings. Would make a summary here but might be faster if you guys discussed directly.

mjkkirschner:
If you guys already discussed I'm happy to go along. Just wanted to make sure.

src/PythonMigrationViewExtension/Controls/BaseDiffViewer.xaml.cs
```diff
@@ -0,0 +1,45 @@
...
+            {
+                ViewModel.ChangeViewModel(Differ.ViewMode.SideBySide);
+                Width = 1200;
+            }
+
+            else
+            {
+                ViewModel.ChangeViewModel(Differ.ViewMode.Inline);
+                Width = 600;
+            }
```
mmisol, line 31: As we discussed, if we want the code in the diff viewer to use the exact same wrapping as the code editor, we will need some fine tuning to these sizes. A good test for this is the line `print 'a very long line that will wrap and will allow us to test if the 2 3 Python migration window respects the same spacing as the Python script editor window'`. This line would break after `3` in the code editor but before `2` in the diff viewer initially.

mmisol: Also, the side by side viewers are a little smaller than the inline viewer, probably because we are not accounting the width of the separator. Also something to consider while fine tuning these sizes.

radumg:
This will have conflicts with the open PR #10928

radumg:
To solve conflicts ahead of time and incorporate fixes, i've rebased this on the branch for #10928 , so please merge that first and then this one.

radumg:
The width of the code (text) area can't be matched when window size is exactly the same as we're always going to have the additional space for the `+` or `-` indicators, so went ahead with subtly making the differ windows larger.

Also used the `lorem ipsum` text as a more accurate indicator on how the word wrap affects multi-line paragraphs.

Inline
image

Side by side
image

src/PythonMigrationViewExtension/Controls/BaseDiffViewer.xaml.cs
radumg: fixed now, please see screenshots below.

radumg:
Colours updated

image

radumg:
Also upgraded the `inline` view to now show the exact fragments of what's changed, instead of a simple deleted/changed for the entire line.

## Before 
image


## After
improved inline view

radumg:
That should wrap this one up as well, please have a look @mmisol , @mjkkirschner

src/PythonMigrationViewExtension/Controls/BaseDiffViewer.xaml.cs
mmisol: Hi @radumg . This is looking better, but I can still see a slight difference with the example I provided (the `3` appears on the first line in the code editor):
Screen Shot 2020-07-30 at 8 32 16 AM
However, after changing to inline view, the wrapping matches the expectations and switching back to side by side view now it also looks as expected. I guess there is something we are not accounting for when displaying the initial window but then we are when switching back and forth?

radumg: hm, that sounds curious as initial display and switching should follow same path and be initialised the same.

radumg: Just tested and can't replicate it, the `3` is on the next line for me when the script editor opens

image

radumg: could this be a high-DPI or scaling issue? Both my laptop and external monitor are 4K with scaling set to the recommended `150%`

Amoursol:
Looking pretty epic to me @radumg! Thank you.

src/PythonMigrationViewExtension/Controls/BaseDiffViewer.xaml.cs
mmisol: That is strange indeed. I wonder why it the script editor shows different for us 🤔 Here is my screenshot:
Screen Shot 2020-07-30 at 9 38 38 AM

radumg: i notice Safari in the background, you running Windows virtualised (Parallels, VmWare) on Mac ? Not sure if it'd make a difference, but it might affect visual fidelity somehow ?

radumg: there's something going on with the window borders on your screenshot ?
Might be skewing things a few pixels width, maybe enough to bump the `3` to next line ?

image

image

mmisol: @radumg My settings are different, so I guess it could be. I'm on 250% DPI. The funny thing is that WPF is supposed to isolate us from these problems, but maybe it's like a 1 pixel issue rounding differently for me than for you.

mmisol: @radumg Just checked and yes, with 150% DPI they both match. I don't know if there is something you can do on your end, but at least you have a way to reproduce it I guess. Let us know if you would like to try.

mmisol: I would say it's not related to parallels because I can reproduce it on 250% when running outside coherence too.

radumg: just tried and 250% produces the same result for me 🤷‍♂️

Also, don't think anyone with a Windows laptop would actually be on 250% scaling unless using an 8K display 😂

image

mmisol:
@Amoursol Bringing this to your attention because it got hidden in a conversation https://github.com/DynamoDS/Dynamo/pull/10932#pullrequestreview-454432122

To summarize, I can see a slight difference in wrapping that @radumg was not able to reproduce on his end. It could be related to a combination of DPI settings and screen resolution but we are not sure. How do you think we should deal with this?

Amoursol:
@mmisol Unless we can reproduce the issue, in which case we should definitely resolve, I am fine with minor visual differences. Do we have a 3rd machine we can test on? One of our Virtual test machines? It may be a Mac vs. Windows issue.

SHKnudsen:
@mmisol @radumg  tested this on my machine as well and everything seems to wrap as expected. I made a few small changes to make sure the sizing of the diff viewer is consistent.

mmisol:
Thanks for testing @SHKnudsen . Still wraps different for me but this should be good to go. I'll delay approval until we merge the other PR, just in case to prevent any issues.

src/PythonMigrationViewExtension/Controls/BaseDiffViewer.xaml.cs
```diff
@@ -19,6 +19,8 @@ internal BaseDiffViewer(PythonMigrationAssistantViewModel viewModel)
         {
             ViewModel = viewModel;
             DataContext = viewModel;
+            // The diff viewer is initialized with the SideBySide view
+            SetSideBySideWindowWidth();
```
radumg, line 23: @SHKnudsen , let's put this after the `initialize component` method call

mmisol:
@SHKnudsen @radumg #10928 got merged. Could you rebase this PR so that we can approve it and merge it too?

SHKnudsen:
@mmisol should be ready now.

PR #10940, opened 2020-07-27
Add LogLevel.ConsoleOnly for DynamoLogger

mmisol:
### Purpose

This log level allows to log messages only to the console, skipping
file logging. The addition of a new level valuie was required because
counterintuitively LogLevel.Console logs to both the console and file.

This new level is currently used only for logging with DynamoPrint, but
the overload was added as a public function in DynamoLogger, so it can
be used in arbitrary code, like ZT nodes.

### Declarations

### Reviewers

@mjkkirschner 

### FYIs

@smangarole

### DIFF
--- a/src/DynamoCore/Logging/DynamoLogger.cs
+++ b/src/DynamoCore/Logging/DynamoLogger.cs
@@ -12,7 +12,7 @@ namespace Dynamo.Logging
     /// <summary>
     /// Specifies the level for log messages. A log message could be a console or file or warning.
     /// </summary>
-    public enum LogLevel{Console, File, Warning}
+    public enum LogLevel{Console, File, Warning, ConsoleOnly}
 
     /// <summary>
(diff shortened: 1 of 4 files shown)


mjkkirschner:
@mmisol it looks good, do you think we should have a test for the new log level? It seems difficult to test because the log file is never generated during testing, and is only sent to std out anyway.

mmisol:
@mjkkirschner Good point. We can actually override the logger settings a test is using, by doing something like this https://github.com/DynamoDS/Dynamo/blob/master/test/Libraries/DynamoPythonTests/PythonTestsWithLogging.cs. Those  tests are already using the new `LogLevel.ConsoleOnly`. What they are not doing is checking that the content is not logged to the file. I couldn't figure out how to check that without trying to read the file from disk, which didn't sound like a good idea.

PR #10948, opened 2020-07-29
[DYN-2827] Added Show/Hide spaces and tabs option for Python script editor.

zeusongit:
### Purpose
[[DYN-2827]](https://jira.autodesk.com/browse/DYN-2827)
Add a global setting option `Show Python Editor Whitespace Characters`.
Show/Hide spaces and tabs in python script editor
Preserve setting in preferences.
Add Tests

spacesntabs

### Declarations

### Reviewers

@DynamoDS/dynamo


src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -504,6 +504,48 @@ public bool IsIronPythonDialogDisabled
             }
         }
 
+        private ICSharpCode.AvalonEdit.TextEditorOptions editTextOptions = new ICSharpCode.AvalonEdit.TextEditorOptions();
+
+        /// <summary>
+        /// Gets the text editor options for python script editor.
+        /// </summary>
+        public ICSharpCode.AvalonEdit.TextEditorOptions TextOptions
```
mmisol, line 512: We might want to avoid making this public, just in case at some point we want to change the editor control. Otherwise we would be in a situation similar to the one we had with the Helix upgrade, although with a lot less exposure.

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
```diff
@@ -65,6 +65,11 @@ internal void Initialize(Guid workspaceGuid, Guid nodeGuid, string propName, str
             editText.TextArea.TextEntering += OnTextAreaTextEntering;
             editText.TextArea.TextEntered += OnTextAreaTextEntered;
 
+            // Initialize editor with global settings for show/hide tabs and spaces
+            editText.Options = dynamoViewModel.TextOptions;
+            editText.Options.ShowTabs = dynamoViewModel.ShowTabsAndSpacesInScriptEditor;
+            editText.Options.ShowSpaces = dynamoViewModel.ShowTabsAndSpacesInScriptEditor;
```
mmisol, line 71: I'm confused by this. If we have `TextOptions` in our VM, why do we also need `ShowTabsAndSpacesInScriptEditor` as well? Can't `ShowTabs` and `ShowSpaces` come already assigned in `TextOptions` so that we can avoid exposing `ShowTabsAndSpacesInScriptEditor` too?
Also, should this be needed if we are using a binding for `Options`?

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
mjkkirschner: this is a valid concern, please do not expose types from our dependencies into our public apis. This was already done for the CodeCompletionProvider and it took significant effort and now technical debt to work around.

src/Libraries/PythonNodeModelsWpf/Resources/ICSharpCode.PythonBinding.Resources.Python.xshd
```diff
@@ -12,8 +12,8 @@
     <FoldMarker bold="false" italic="false" color="Gray" bgcolor="#242424" />
     <SelectedFoldLine bold="false" italic="false" color="Gray" />
     <EOLMarkers bold="false" italic="false" color="#999999" />
-    <SpaceMarkers bold="false" italic="false" color="#999999" />
-    <TabMarkers bold="false" italic="false" color="#999999" />
+    <SpaceMarkers bold="false" italic="false" color="#75715E" />
```
mjkkirschner, line 15: if you search the avalonEdit code base for this element - it appears to be used absolutely no where... which is I guess why it does nothing.

mjkkirschner: looks to me like they are hardcoded to white... https://github.com/icsharpcode/AvalonEdit/blob/28b887f78c821c7fede1d4fc461bde64f5f21bd1/ICSharpCode.AvalonEdit/Rendering/SingleCharacterElementGenerator.cs#L109

seems like a bug in avalonEdit maybe?

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -504,6 +504,48 @@ public bool IsIronPythonDialogDisabled
...
+            {
+                return editTextOptions;
+            }
+
+            set
+            {
+                if (editTextOptions != value)
+                {
+                    editTextOptions = value;
+                    RaisePropertyChanged("TextOptions");
```
aparajit-pratap, line 524: nameof

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
zeusongit: How do I make `TextOptions ` private and assign its value in viewmodel?
`TextOptions` is needed because `ShowSpaces` and `ShowTabs` are its child members, we cannot access them directly, and we need a binding to this property in order to reflect the changes in real-time.
On the other hand `ShowTabsAndSpacesInScriptEditor` is needed because we have a global setting option that is binded to its bool value.

mjkkirschner: You don't need to use this type directly in this property, you can use an interface or a class that hides the underlying type.

src/DynamoCoreWpf/ViewModels/Core/Converters/DynamoTextOptions.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// Sets the text editor options
+        /// </summary>
+        public void SetTextOptions(bool value) {
+            TextOptions.ShowSpaces = value;
+            TextOptions.ShowTabs = value;
+        }
+        /// <summary>
+        /// Gets the text editor options
+        /// </summary>
+        public ICSharpCode.AvalonEdit.TextEditorOptions GetTextOptions()
```
mmisol, line 31: If this ends up being `public` we'd still expose the type publicly in our API, right?

Can we maybe fire an event when the editor setting is changed and react to it on the Python side by changing a global `TextEditorOptions` that lives inside that assembly?

zeusongit: Will try to do this.

mjkkirschner: you can also maybe cheat and use internal and `internalsVisibleto`

src/DynamoCoreWpf/ViewModels/Core/Converters/DynamoTextOptions.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// </summary>
+        public DynamoTextOptions()
+        {
+            if(TextOptions == null)
+                TextOptions = new ICSharpCode.AvalonEdit.TextEditorOptions();
+        }
+        /// <summary>
+        /// Sets the text editor options
+        /// </summary>
+        public void SetTextOptions(bool value) {
```
mjkkirschner, line 24: I think this should be very clear since it's public - 

1. let's call it `ShowWhiteSpaceCharacters` etc
2. it needs /// summary and parameter tags / comments

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -504,6 +505,50 @@ public bool IsIronPythonDialogDisabled
             }
         }
 
+        //private ICSharpCode.AvalonEdit.TextEditorOptions editTextOptions = new ICSharpCode.AvalonEdit.TextEditorOptions() {
```
mjkkirschner, line 508: remove?

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -504,6 +505,50 @@ public bool IsIronPythonDialogDisabled
...
+        //    ConvertTabsToSpaces = false
+        //};
+        private DynamoTextOptions editTextOptions = new DynamoTextOptions();
+
+        /// <summary>
+        /// Gets the text editor options for python script editor.
+        /// </summary>
+        public DynamoTextOptions TextOptions
```
mjkkirschner, line 516: maybe this property name should more clearly describe that it's about python? Especially if it's public - what about PythonScriptEditorTextOptions?

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -504,6 +505,50 @@ public bool IsIronPythonDialogDisabled
...
+                if (editTextOptions != value)
+                {
+                    editTextOptions = value;
+                    RaisePropertyChanged(nameof(TextOptions));
+                }
+            }
+        }
+
+        /// <summary>
+        /// Indicates if whether the whitespaces and tabs should be visible in the python script editor.
```
mjkkirschner, line 534: I think you can drop `whether` from this sentence.

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -504,6 +505,50 @@ public bool IsIronPythonDialogDisabled
...
+        {
+            get
+            {
+                return model.PreferenceSettings.ShowTabsAndSpacesInScriptEditor;
+            }
+            set
+            {
+                TextOptions.SetTextOptions(value);
+                model.PreferenceSettings.ShowTabsAndSpacesInScriptEditor = value;
+                RaisePropertyChanged(nameof(ShowTabsAndSpacesInScriptEditor));
```
mjkkirschner, line 546: because both @mmisol  and I were confused when looking at this initially, maybe it is worth adding a comment that this property is for the menu item?

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml
```diff
@@ -83,7 +87,8 @@
                                    ShowLineNumbers="True"
                                    Background="#272822"
                                    Foreground="#F8F8F2"
-                                   HorizontalScrollBarVisibility="Disabled" />
+                                   HorizontalScrollBarVisibility="Disabled"
+                                   Options="{Binding TextOptions, FallbackValue={StaticResource TextEditOptions}}" />
```
mjkkirschner, line 91: @zeusongit If you remove this binding - does everything still work?

zeusongit: yes, it does -.-

src/DynamoCoreWpf/ViewModels/Core/Converters/DynamoTextOptions.cs
mmisol: Cheating would be fine too. If you do please add a comment stating it was made internal to avoid exposing AvalonEdit in our API.

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -504,6 +505,49 @@ public bool IsIronPythonDialogDisabled
             }
         }
 
+
+        private DynamoPythonScriptEditorTextOptions editTextOptions = new DynamoPythonScriptEditorTextOptions();
+        /// <summary>
+        /// Gets/Sets the text editor options for python script editor.
+        /// </summary>
+        public DynamoPythonScriptEditorTextOptions PythonScriptEditorTextOptions
```
mjkkirschner, line 513: hmm!
so - if the binding is no longer needed...
1. why is this public?
2. why are we raising a property changed event?

src/DynamoCoreWpf/Properties/Resources.en-US.resx
```diff
@@ -2267,4 +2267,7 @@ Uninstall the following packages: {0}?</value>
   <data name="psDynamoPoint" type="System.Resources.ResXFileRef, System.Windows.Forms">
     <value>..\ViewModels\Watch3D\compiledShaders\psDynamoPoint;System.Byte[], mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </data>
+  <data name="DynamoViewSettingsMenuShowTabsAndSpacesInScriptEditor" xml:space="preserve">
+    <value>Show Python Editor Whitespace Characters</value>
```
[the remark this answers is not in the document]
QilongTang: Echo that

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
zeusongit: I will remove the `set` block altogether as it is never being called.

zeusongit:
I am able to access the members after making them internal, and the script editor works fine.
Will merge.
Thank you for the review.

QilongTang:
@zeusongit Did you forgot to merge or waiting for the PR checks?

zeusongit:
Was waiting for the self serve, which just completed, merging.

PR #10950, opened 2020-07-29
Improve namespace listing using CPython3 engine

mmisol:
### Purpose

Namespace listing using dir would only provide non meaningful results
using the CPython3 engine. This is now fixed by preloading names by
default globally.

### Declarations

### Reviewers

@mjkkirschner 

### FYIs

@DynamoDS/dynamo

### DIFF
--- a/src/Libraries/DSCPython/CPythonEvaluator.cs
+++ b/src/Libraries/DSCPython/CPythonEvaluator.cs
@@ -189,7 +189,7 @@ namespace DSCPython
                 {
                     if (globalScope == null)
                     {
-                        globalScope = Py.CreateScope(globalScopeName);
+                        globalScope = CreateGlobalScope();
                     }
                     using (PyScope scope = Py.CreateScope())
                     {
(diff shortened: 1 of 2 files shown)


src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -236,6 +236,20 @@ static CPythonEvaluator()
...
         }
 
+        /// <summary>
+        /// Creates and initializaes the global Python scope.
+        /// </summary>
+        private static PyScope CreateGlobalScope()
+        {
+            var scope = Py.CreateScope(globalScopeName);
+            // Allows discoverability of modules by inspecting their attributes
+            scope.Exec(@"
```
mjkkirschner, line 246: I'm confused why this only needed to be run on this scope...
I guess it may be related to the `sys.path` issue @reddyashish is looking at and some confusion I have around the lifetime of ironPython engines vs PythonNet ones.

mmisol: Yes. From what I could see, loaded modules stay alive beyond a single Python node evaluation. It's like they stay alive with the process. Doing it here makes more sense `clr.setPreload(True)`, not only because it's easier, but also because once a module is loaded you can't reload it.

Issue #10816, opened 2020-06-29
Crash report from Dynamo 2.4.1.6394


MissmelissaL:
# Issue Description
Dynamo will not run the graph and every node has a null value

### What did you do?
Modified nodes with is input testing in dynamo player

### What did you expect to see?
Run completed or run with warnings and see values at each node 

### What did you see instead?
Nulls for every node including the first node which is an import or an excel file

### What packages or external references (if any) were used?
Data-shapes, Bake ++

### Stack Trace
```
Retrive attribute failed.

   at Autodesk.Civil.AttributeHelper.getAttributeGeneric<AeccAtom>(AeccAtom* , UInt32 attributeId, IAeccAttributeBin* pAttrBin, AeccUserParam* userParam)
   at Autodesk.Civil.AttributeHelper.getAttributeAeccAtom(AeccAtom* , UInt32 attributeId, IAeccAttributeBin* pAttrBin, AeccUserParam* userParam)
   at Autodesk.Civil.AttributeHelper.getAttributeAeccAtom(AeccAtom* , UInt32 attributeId, IAeccAttributeBin* pAttrBin)
   at Autodesk.Civil.DatabaseServices.Entity.get_Name()
   at Autodesk.Civil.DynamoNodes.CivilObject.ToString()
   at ProtoFFI.CLRObjectMarshaler.GetStringValue(StackValue dsObject)
   at ProtoCore.DSASM.Mirror.ExecutionMirror.GetClassTrace(StackValue val, Heap heap, Int32 langblock, Boolean forPrint)
   at ProtoCore.DSASM.Mirror.ExecutionMirror.GetStringValue(StackValue val, Heap heap, Int32 langblock, Int32 maxArraySize, Int32 maxOutputDepth, Boolean forPrint)
   at ProtoCore.Utils.StringUtils.ConvertToString(StackValue sv, RuntimeCore runtimeCore, RuntimeMemory rmem)
   at ProtoCore.Lang.BuiltInFunctionEndPoint.Execute(Context c, List`1 formalParameters, StackFrame stackFrame, RuntimeCore runtimeCore)
   at ProtoCore.CallSite.ExecWithZeroRI(List`1 functionEndPoint, Context c, List`1 formalParameters, StackFrame stackFrame, RuntimeCore runtimeCore, SingleRunTraceData previousTraceData, SingleRunTraceData newTraceData, FunctionEndPoint finalFep)
[...]

PR #10952, opened 2020-07-29
add try catch to CLR Object Marshaler getStringValue

mjkkirschner:
### Purpose

Address the situation where a `ToString` call on a clr object that is not null results in an exception - it turns out this happens quite frequently in Autocad and Revit.

### Declarations

### DIFF
--- a/src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
+++ b/src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
@@ -1256,21 +1256,28 @@ namespace ProtoFFI
 
             if (clrObject != null)
             {
-                if (!DumpXmlProperties)
+                try
                 {
-                    return clrObject.ToString();
-                }
-                else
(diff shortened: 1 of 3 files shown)


aparajit-pratap:
@mjkkirschner can we verify that the ToString calls on these Revit/AutoCAD objects are throwing the exceptions and that they are the reason for the crash?

mjkkirschner:
I can try to reproduce locally with Revit, but the stack trace seems to imply that:
https://github.com/DynamoDS/Dynamo/issues/10816

mjkkirschner:
Screen Shot 2020-07-30 at 5 01 25 PM

crashed before - after seems to execute ok.

aparajit-pratap:
@mjkkirschner it looks like the exception for occurring for only certain parameters and not all. In cases, where the exception is being thrown and caught, there should be empty strings, am I right?

mjkkirschner:
yep!

PR #10972, opened 2020-08-06
Allow integer slider to support 64 bit ints

aparajit-pratap:
### Purpose

https://jira.autodesk.com/browse/DYN-1085

- Added `IntegerSlider64Bit` that derives from `SliderBase<long>` to support 64 bit integers
- Deserialize a `CoreNodeModels.Input.IntegerSlider` type from JSON as a `CoreNodeModels.Input.IntegerSlider64Bit` so that the new 64 bit slider node is instantiated
- Serialize 64 bit node back as `CoreNodeModels.Input.IntegerSlider` type so that new DYN's can be opened by older versions of Dynamo
- ~~Deserialize 32 bit integer slider from old XML format as new 64 bit integer slider~~ - This does not seem to be straightforward since in order to not break the API I needed to keep the legacy class, which is why I had to retain the legacy behaviour of xml-based 32 bit integer sliders (to be deserialized as 32 bit). Now only when you save the XML file (when saved as JSON) and reopened, will the slider behave as a 64 bit integer slider node, not before.

### Declarations

### Reviewers

@DynamoDS/dynamo


mjkkirschner:
@aparajit-pratap an interesting set of failures, including the ones your predicted.

src/Libraries/CoreNodeModels/Input/IntegerSlider.cs
```diff
@@ -91,7 +91,7 @@ public IntegerSlider()
         //If the value field in the slider has a number greater than
         //In32.Maxvalue (or MinValue), the value will be changed to Int32.MaxValue (or MinValue)
         //The value will be changed, but to update the UI, this property is overridden here. 
-        public override int Value
+        public override long Value
```
mjkkirschner, line 94: ugh, this is an API break, what a new slider class that inherits from sliderBase<long> ?

src/Libraries/CoreNodeModels/Input/IntegerSlider.cs
```diff
@@ -22,7 +22,7 @@ namespace CoreNodeModels.Input
     [SupressImportIntoVM]
     [IsDesignScriptCompatible]
     [AlsoKnownAs("DSCoreNodesUI.Input.IntegerSlider")]
-    public class IntegerSlider : SliderBase<int>
+    public class IntegerSlider : SliderBase<long>
```
aparajit-pratap, line 25: @mjkkirschner is this an API break too?

mjkkirschner: I think so... 

I'm not sure if this would be considered a break or judgement call.

aparajit-pratap:
@QiltongTang do you know how to change the description of the output port tooltip for a UI node? The only thing remaining in this PR is to change the output port type of the integer slider to report `Int64`. `Int32` is obviously wrong in this case.
image

QilongTang:
@aparajit-pratap I would try look at when outport model constructor got called when placing integer slider: https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Graph/Nodes/PortModel.cs#L335

QilongTang:
image

aparajit-pratap:
Thanks @QilongTang!

src/Libraries/CoreNodeModels/Input/IntegerSlider.cs
```diff
@@ -272,6 +272,9 @@ public override NodeInputData InputData
             Step = 1;
             Value = 1;
             ShouldDisplayPreviewCore = false;
+
+            var outport = outPorts.First();
+            outport.ToolTip = "Int64";
```
QilongTang, line 277: If we do not add this, would the tooltip be "long"?

aparajit-pratap: The tooltip string is serialized as `Int32` in JSON files currently so when the DYN file is opened, it is read as `Int32`, which is not true anymore. If these files are saved again, the tooltip is correctly serialized as `Int64`.

QilongTang:
@aparajit-pratap Seems there are regressions with this PR?

aparajit-pratap:
> @aparajit-pratap Seems there are regressions with this PR?

@QilongTang there are 2 regressions remaining. I addressed many of them yesterday. Will address the remaining ones today. In the meantime do you have any comments on the rest of the review?

src/Libraries/CoreNodeModels/Input/BasicInteractive.cs
```diff
@@ -61,7 +61,6 @@ public override NodeInputData InputData
 
         protected BasicInteractive(IEnumerable<PortModel> inPorts, IEnumerable<PortModel> outPorts) : base(inPorts, outPorts)
         {
-            Type type = typeof(T);
```
QilongTang, line 64: Do we want to simply remove this function?

src/Libraries/CoreNodeModels/Input/IntegerSlider.cs
QilongTang: Thanks for the explanation

src/Libraries/CoreNodeModels/Input/SliderBase.cs
```diff
@@ -130,5 +131,27 @@ internal static int ConvertStringToInt(string value)
             }
             return result;
         }
+
+        internal static long ConvertStringToInt64(string value)
```
QilongTang, line 135: Can you add some function comments saying what would be the expected return value for string over the maximum or minimum value?

test/core/recorded/UpdateNodeContents.xml
```diff
@@ -1,4 +1,4 @@
-<Commands ExitAfterPlayback="true" PauseAfterPlaybackInMs="10" CommandIntervalInMs="1000">
+<Commands ExitAfterPlayback="true" PauseAfterPlaybackInMs="10" CommandIntervalInMs="100">
```
QilongTang, line 1: hmm this change is to save test time?

aparajit-pratap: Yes.

QilongTang: Thanks!

src/Libraries/CoreNodeModels/Input/BasicInteractive.cs
aparajit-pratap: Will check that.

src/Libraries/CoreNodeModels/Input/SliderBase.cs
aparajit-pratap: Sure.

src/Libraries/CoreNodeModels/Input/BasicInteractive.cs
aparajit-pratap: It actually has references so I cannot remove it.

PR #10974, opened 2020-08-06
CPython engine marshals python classes to lists when the expectation is to get a handle.

mjkkirschner:
### Purpose

While trying to get data-shapes working with the cpython engine it was discovered that some of the data-shapes python classes implement `__iter__/__getitem__` and sequence protocols.

These classes are marshaled into dynamo as lists, but for use in data-shapes they need to be marshaled as handles only.

To enable this workflow I've added a new "builtin" 👋   attribute the class author can choose to add to their class. `__dynamoskipconversion__`. If this attribute exists at all (the return value does not matter) then the object will skip all conversion attempts and immediately be marshaled as a handle.

For dataShapes this means a one line code change to get this class working.

### Declarations

- [ ] **When/If this PR is approved I will update the python3 wiki with information on this attribute.**

### DIFF
--- a/src/Libraries/DSCPython/CPythonEvaluator.cs
+++ b/src/Libraries/DSCPython/CPythonEvaluator.cs
@@ -147,6 +147,7 @@ namespace DSCPython
     [IsVisibleInDynamoLibrary(false)]
     public static class CPythonEvaluator
     {
+        private const string DynamoSkipAttributeName = "__dynamoskipconversion__";
         static PyScope globalScope;
         internal static readonly string globalScopeName = "global";
         static CPythonEvaluator()
@@ -390,9 +391,7 @@ clr.setPreload(True)
 
                             if (IsMarkedToSkipConversion(pyObj))
                             {
-                                var globalScope = PyScopeManager.Global.Get(globalScopeName);
-                                globalScope.Set(pyObj.Handle.ToString(), pyObj);
-                                return new DynamoCPythonHandle(pyObj.Handle);
+                                return GetDynamoCPythonHandle(pyObj);
                             }
 
                             // Dictionaries are iterable, so they should come first
@@ -447,11 +446,7 @@ clr.setPreload(True)
                                 {
                                     if (unmarshalled.Equals(pyObj))
                                     {
-                                        var globalScope = PyScopeManager.Global.Get(globalScopeName);
-                                        //try moving object to global scope
-                                        globalScope.Set(pyObj.Handle.ToString(), pyObj);
-
-                                        return new DynamoCPythonHandle(pyObj.Handle);
+                                        return GetDynamoCPythonHandle(pyObj);
                                     }
                                     else
                                     {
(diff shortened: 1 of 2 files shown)


test/Libraries/DynamoPythonTests/PythonEvalTests.cs
```diff
@@ -212,6 +212,46 @@ class myobj:
...
+        return iter([0,1,2,3])
+    def __getitem__(self,key):
+        return key
+
+o = notiterable()
+OUT = o
+";
+            var empty = new ArrayList();
+           
+           Assert.IsInstanceOf(typeof(IList),DSCPython.CPythonEvaluator.EvaluatePythonScript(code, empty, empty));
```
mmisol, line 249: Maybe also assert the value for this case?

mjkkirschner: will do

test/Libraries/DynamoPythonTests/PythonEvalTests.cs
```diff
@@ -212,6 +212,46 @@ class myobj:
             });
         }
 
+        [Test]
+        public void PythonObjectWithDynamoSKIPisNotMarshaled()
```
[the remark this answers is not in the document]
mjkkirschner: no good reason, will change it.

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -459,6 +467,11 @@ public static DataMarshaler OutputMarshaler
             }
         }
 
+        private static bool IsMarkedToSkipConversion(PyObject pyObj)
+        {
+            return pyObj.HasAttr("__dynamoskipconversion__");
```
aparajit-pratap, line 472: Can we make this string a constant member variable of the class?

test/Libraries/DynamoPythonTests/PythonEvalTests.cs
```diff
@@ -212,6 +214,47 @@ class myobj:
...
+        return iter([0,1,2,3])
+    def __getitem__(self,key):
+        return key
+
+o = notiterable()
+OUT = o
+";
+            var empty = new ArrayList();
+
+            Assert.IsInstanceOf(typeof(IList), DSCPython.CPythonEvaluator.EvaluatePythonScript(code, empty, empty));
```
aparajit-pratap, line 251: Why not just store the result of `DSCPython.CPythonEvaluator.EvaluatePythonScript(code, empty, empty)` in a local var and use it in both places?

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -387,6 +387,14 @@ public static DataMarshaler OutputMarshaler
                             {
                                 return outputMarshaler.Marshal(clrObj);
                             }
+
+                            if (IsMarkedToSkipConversion(pyObj))
+                            {
+                                var globalScope = PyScopeManager.Global.Get(globalScopeName);
+                                globalScope.Set(pyObj.Handle.ToString(), pyObj);
+                                return new DynamoCPythonHandle(pyObj.Handle);
```
aparajit-pratap, line 395: Minor but these 3 lines of code are repeated in 2 places - here and further down below. Might be worth refactoring into a helper method?

test/Libraries/DynamoPythonTests/PythonEvalTests.cs
mjkkirschner: sure, can do.

src/Libraries/DSCPython/CPythonEvaluator.cs
mjkkirschner: wil fix.

PR #10975, opened 2020-08-07
Expose CPython3 engine evaluation events

mmisol:
### Purpose

These events were created with internal visibility, which does not make
them available for integrators. Integrators need, for instance, to
react to the start and end of an evaluation to disable/re-enable
certain UI elements.

At the same time makes DynamoCPythonHandle internal, so that we do not
accidentally expose this type.

### Declarations

### Reviewers

@mjkkirschner @QilongTang 

### FYIs

@DynamoDS/dynamo

### DIFF
--- a/src/Libraries/DSCPython/CPythonEvaluator.cs
+++ b/src/Libraries/DSCPython/CPythonEvaluator.cs
@@ -40,7 +40,7 @@ namespace DSCPython
     /// then make sure to call Dispose when you are done with the instance.
     /// </summary>
     [IsVisibleInDynamoLibrary(false)]
-    public class DynamoCPythonHandle : IDisposable
+    internal class DynamoCPythonHandle : IDisposable
     {
         /// <summary>
         /// A static map of DynamoCPythonHandle counts, is used to avoid removing the underlying python objects from the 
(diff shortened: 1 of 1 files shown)


src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -40,7 +40,7 @@ public int GetHashCode(DynamoCPythonHandle obj)
     /// then make sure to call Dispose when you are done with the instance.
     /// </summary>
     [IsVisibleInDynamoLibrary(false)]
-    public class DynamoCPythonHandle : IDisposable
+    internal class DynamoCPythonHandle : IDisposable
```
mjkkirschner, line 43: hm - I feel this change should incur a failing test. I do not think I added adequate tests, if this type is made internal, it should not be imported into the VM - if I had written a test that opened and ran a graph, I think it may have failed.

@aparajit-pratap  please confirm.
@mmisol can you please try this in dynamo with a python class, is the handle still returned correctly?

mmisol: @mjkkirschner I'll give it a try, but we should have a test for this if I remember correctly

mmisol: This sample I had continues to work after the change. My guess is we don't really need it to be public if it's only used as a result. 

Screen Shot 2020-08-07 at 1 45 26 PM

aparajit-pratap: I'm not sure how this worked. As @mjkkirschner mentions, in order to be imported into the VM, this type needs to be public. I don't think the Dispose method is going to be invoked for this type from VM GC now that it is internal. @mmisol any specific reason to make this internal?

aparajit-pratap: Sorry for the late review.

aparajit-pratap: It strangely seems to work as expected including Dispose. My guess is that it probably is referenced somehow and is created as an "empty" type in the VM.

aparajit-pratap: Ok, it was imported as System.object, which has a Dispose method and at runtime it calls the exact Dispose method of the object polymorphically.

PR #10977, opened 2020-08-07
Sys.path value should be limited to the scope of the python node(CPython3 Engine)

reddyashish:
### Purpose

This PR is to address the task https://jira.autodesk.com/browse/DYN-2956. 

Issue: For CPython3 Engine, when you add a custom path to the 'sys.path' variable, its value was persistent for the whole Dynamo session. This wasn't the case with IronPython2. To make this consistent, we have decided to limit the scope of this variable to individual python node. 

To fix this, whenever a python node is evaluated, the 'sys.path' value is reset to contain only the default python values. For CPython engine, the first 3 paths correspond to the default python paths. 

To test this, I have added a custom path "C:\\Program Files\\dotnet" in the Python script editor and checked that this path is not reflected in the 2nd python node. I wasn't sure if using such paths in our test could cause any issues. My thought was this dotnot folder should be present on all Windows machines with dotnet installed. Looking for suggestions.

### Declarations

### Reviewers
@mjkkirschner @mmisol 

### FYIs
@DynamoDS/dynamo

### DIFF
--- a/src/Libraries/DSCPython/CPythonEvaluator.cs
+++ b/src/Libraries/DSCPython/CPythonEvaluator.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Reflection;
 using Autodesk.DesignScript.Runtime;
@@ -146,6 +147,7 @@ namespace DSCPython
     {
         static PyScope globalScope;
         internal static readonly string globalScopeName = "global";
+
         static CPythonEvaluator()
         {
             InitializeEncoders();
@@ -176,7 +178,6 @@ namespace DSCPython
             {
                 PythonEngine.Initialize();
                 PythonEngine.BeginAllowThreads();
-
             }
 
             IntPtr gs = PythonEngine.AcquireLock();
@@ -188,6 +189,10 @@ namespace DSCPython
                     {
                         globalScope = CreateGlobalScope();
                     }
+                   
+                    // Reset the 'sys.path' value to the default python paths on node evaluation. 
+                    code = "import sys" + Environment.NewLine + "sys.path = sys.path[0:3]" + Environment.NewLine + code;
+
                     using (PyScope scope = Py.CreateScope())
                     {
                         ProcessAdditionalBindings(scope, bindingNames, bindingValues);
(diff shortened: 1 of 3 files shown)


src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -188,6 +189,10 @@ static CPythonEvaluator()
                     {
                         globalScope = CreateGlobalScope();
                     }
+                   
+                    // Reset the 'sys.path' value to the default python paths on node evaluation. 
+                    code = "import sys" + Environment.NewLine + "sys.path = sys.path[0:3]" + Environment.NewLine + code;
```
mmisol, line 194: This should be done in its own scope. Otherwise the line numbers will be moved, which will affect traceback.

mjkkirschner: alternatively it can be done from the c# side - without injecting more code into python code string at all.

reddyashish:
Modified the code to set the sys.path in its own scope. Verified that the line numbers in the trace point to the correct line number. 
Tried doing the same from the C# side but the scope.Get("sys.path") throws an error saying that this attribute is not present in the scope. Even the PythonEngine.PythonPath variable is also not consistent with the value present in sys.path field, so PythonEngine.PythonPath cannot be used to reset the path.

reddyashish:
Self-serve passed with no failures: https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/715/

test/Libraries/DynamoPythonTests/PythonEditTests.cs
```diff
@@ -612,7 +613,31 @@ public void CPythonClassCanBeReturnedAndSafelyDisposedInDownStreamNode()
             ViewModel.Model.ExecuteCommand(deleteCmd2);
 
             Assert.IsEmpty(DynamoCPythonHandle.HandleCountMap.Where(x => x.ToString().Contains("myClass")));
+        }
+
+        [Test]
+        public void VerifySysPathValueForCPythonEngine()
+        {
+            // open test graph
```
mmisol, line 621: Can you also add an assertion or another simple test to make sure the number of paths is 3 before making any changes to it? Given those paths are added by Python.NET by default, and we are assuming they are 3 on our end, this would be useful to catch changes on their end.

reddyashish: Added it to the same test.

reddyashish:
Merging this.

PR #10978, opened 2020-08-07
Fix whitespace during Python 2to3 migration

mmisol:
### Purpose

The 2to3 migration script, by design, does not take care of normalizing
whitespaces. Python 3, as opposed to Python 2, enforces that spaces and
tabs cannot be mixed. In order to get a working script, a reindentation
is performed over the migrated content. This is only done in case the
code uses tabs, otherwise there is no need to do changes to indent.

### Declarations

### Reviewers

@QilongTang @mjkkirschner 

### FYIs

@DynamoDS/dynamo

### DIFF
--- a/src/PythonMigrationViewExtension/MigrationAssistant/reindent.py
+++ b/src/PythonMigrationViewExtension/MigrationAssistant/reindent.py
@@ -1,0 +1,182 @@
+#! /usr/bin/env python3
+
+# Based on a script by Tim Peters, 03 October 2000.
+
+import tokenize
+
+
+def reindent(code):
+    r = Reindenter(code)
+    r.run()
+    return r.getresult()
+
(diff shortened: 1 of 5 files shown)


src/PythonMigrationViewExtension/MigrationAssistant/reindent.py
```diff
@@ -0,0 +1,182 @@
+#! /usr/bin/env python3
+
+# Based on a script by Tim Peters, 03 October 2000.
```
mjkkirschner, line 3: what is the license for this code?
should it be added to license / about etc?

mmisol: Not sure. I got the original script from the Python repo, so I guess it's the same license.

PR #10983, opened 2020-08-10
CPython eval always fails

mjkkirschner:
### Purpose

Cpython eval should reflect actual evaluation results similar to DSIronPython engine.

### Declarations

### DIFF
--- a/test/Libraries/DynamoPythonTests/DynamoPythonTests.csproj
+++ b/test/Libraries/DynamoPythonTests/DynamoPythonTests.csproj
@@ -76,6 +76,7 @@
       <Private>False</Private>
     </Reference>
     <Reference Include="PresentationCore" />
+    <Reference Include="Python.Runtime, Version=2.5.1.0, Culture=neutral, PublicKeyToken=null" />
     <Reference Include="System" />
     <Reference Include="System.Core" />
     <Reference Include="Microsoft.CSharp" />
--- a/test/Libraries/DynamoPythonTests/PythonEvalTests.cs
+++ b/test/Libraries/DynamoPythonTests/PythonEvalTests.cs
@@ -3,9 +3,7 @@ using System.Collections;
 using System.Collections.Generic;
 using NUnit.Framework;
 using System.Linq;
-using ProtoCore.Lang;
 using DSCPython;
-using Python.Runtime;
 
 namespace DSPythonTests
 {
@@ -175,7 +173,7 @@ print 'hello'
         [Test]
         public void CPythonEngineWithErrorRaisesCorrectEvent()
         {
-            var code = @"1";
+          
             var count = 0;
             DSCPython.EvaluationEventHandler CPythonEvaluator_EvaluationEnd = (state, scope, codeString, bindings) =>
             {
@@ -192,6 +190,7 @@ print 'hello'
 
             CPythonEvaluator.EvaluationEnd += CPythonEvaluator_EvaluationEnd;
 
+            var code = @"1";
             try
             {
                 DSCPython.CPythonEvaluator.EvaluatePythonScript(code, new ArrayList(), new ArrayList());
@@ -200,8 +199,8 @@ print 'hello'
             {
                 Assert.AreEqual(1, count);
             }
-            code = @"1/a";
 
+            code = @"1/a";
             try
             {
                 DSCPython.CPythonEvaluator.EvaluatePythonScript(code, new ArrayList(), new ArrayList());
@@ -298,7 +297,7 @@ OUT = o
             Assert.IsInstanceOf(typeof(IList), result1);
             Assert.IsTrue(new List<object>() { 0L, 1L, 2L, 3L }
                 .SequenceEqual((IEnumerable<Object>)result1));
-            Assert.IsInstanceOf(typeof(DSCPython.DynamoCPythonHandle),result2 );
+            Assert.IsInstanceOf(typeof(DSCPython.DynamoCPythonHandle), result2);
 
         }
 
--- a/test/Libraries/DynamoPythonTests/PythonEvalTests.cs
+++ b/test/Libraries/DynamoPythonTests/PythonEvalTests.cs
@@ -5,6 +5,7 @@ using NUnit.Framework;
 using System.Linq;
 using ProtoCore.Lang;
 using DSCPython;
+using Python.Runtime;
 
 namespace DSPythonTests
 {
@@ -16,7 +17,6 @@ namespace DSPythonTests
             DSCPython.CPythonEvaluator.EvaluatePythonScript,
             DSIronPython.IronPythonEvaluator.EvaluateIronPythonScript
         };
-        static int count = 0;
 
         [Test]
         [Category("UnitTests")]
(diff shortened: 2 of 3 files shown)


test/Libraries/DynamoPythonTests/DynamoPythonTests.csproj
```diff
@@ -76,6 +76,7 @@
       <Private>False</Private>
     </Reference>
     <Reference Include="PresentationCore" />
+    <Reference Include="Python.Runtime, Version=2.5.1.0, Culture=neutral, PublicKeyToken=null" />
```
mjkkirschner, line 79: this is required because the evaluationHandlers require a reference to `pyScope` a type from pythonNet's Python.Runtime. 😢

test/Libraries/DynamoPythonTests/PythonEvalTests.cs
```diff
@@ -171,6 +172,39 @@ public void CPythonEngineIncludesTraceBack()
...
+                // Trace back is expected here. Line is 3 due to the line break from the code variable declaration
+                Assert.AreEqual(@"ZeroDivisionError : division by zero ['  File ""<string>"", line 3, in <module>\n']", exc.Message);
+                count = count + 1;
+            }
+            finally
+            {
+                DSCPython.CPythonEvaluator.EvaluationEnd -= CPythonEvaluator_EvaluationEnd;
+                Assert.AreEqual(1, count);
+            }
+        }
```
mmisol, line 201: This is not a good test because:
1. Even if the event was not fired it would still pass.
2. Does not test the event gives a success state when it runs the code without errors.

On a secondary level, this looks like copied from another test that checked the line number of the traceback. That doesn't help to see the intention of the test clearly, so I would recommend removing the non-relevant parts.

mjkkirschner: @mmisol thanks, I put the count assertion in the wrong place, will fix it up.

mjkkirschner: actually @mmisol - your other points are still valid, but I think 1. is false.
If this event was not raised, then the assertion of  count == 1 in the finally block would fail the test, would it not?

mmisol: `count` is increased in the `catch` block, not by the handler, right? My point is, how would you know the handler was executed if it does not change anything outside of himself?

mjkkirschner: 👍

test/Libraries/DynamoPythonTests/PythonEvalTests.cs
```diff
@@ -15,7 +16,7 @@ public class PythonEvalTests
             DSCPython.CPythonEvaluator.EvaluatePythonScript,
             DSIronPython.IronPythonEvaluator.EvaluateIronPythonScript
         };
-
+        static int count = 0;
```
mmisol, line 19: Could you instead define the handler as a delegate/lambda internal to the test so there is no need to define a static field that is only to be used by that test?

test/Libraries/DynamoPythonTests/PythonEvalTests.cs
```diff
@@ -171,6 +172,59 @@ public void CPythonEngineIncludesTraceBack()
             }
         }
 
+        [Test]
+        public void CPythonEngineWithErrorRaisesCorrectEvent()
+        {
+
+            var code = @"
+1
+";          count = 0;
```
mmisol, line 181: The code should be a single line for something as simple as this. Also, please don't do multiple sentences in a single line.

test/Libraries/DynamoPythonTests/PythonEvalTests.cs
```diff
@@ -171,6 +172,59 @@ public void CPythonEngineIncludesTraceBack()
...
+            DSCPython.CPythonEvaluator.EvaluationEnd += CPythonEvaluator_EvaluationEnd;
+          
+            try
+            {
+                DSCPython.CPythonEvaluator.EvaluatePythonScript(code, new ArrayList(), new ArrayList());
+            }
+           
+          
+            finally
```
mmisol, line 190: Please don't leave blank lines between related blocks

test/Libraries/DynamoPythonTests/PythonEvalTests.cs
```diff
@@ -171,6 +172,59 @@ public void CPythonEngineIncludesTraceBack()
...
+            {
+                Assert.AreEqual(1, count);
+            }
+             code = @"
+1/a
+";
```
mmisol, line 196: Single line please.

test/Libraries/DynamoPythonTests/PythonEvalTests.cs
```diff
@@ -171,6 +172,59 @@ public void CPythonEngineIncludesTraceBack()
...
+
+            try
+            {
+                DSCPython.CPythonEvaluator.EvaluatePythonScript(code, new ArrayList(), new ArrayList());
+            }
+            catch
+            {
+                //we anticipate a undefined var error.
```
mmisol, line 204: `a` => `an`

test/Libraries/DynamoPythonTests/PythonEvalTests.cs
```diff
@@ -171,6 +172,59 @@ public void CPythonEngineIncludesTraceBack()
...
+            }
+
+            finally
```
mmisol, line 207: No blank lines before finally

test/Libraries/DynamoPythonTests/PythonEvalTests.cs
mjkkirschner: I don't want to use a lambda because then I can't unsubscribe, but I will try a delegate.

test/Libraries/DynamoPythonTests/PythonEvalTests.cs
mjkkirschner: can you clarify this comment?
`Also, please don't do multiple sentences in a single line.`

test/Libraries/DynamoPythonTests/PythonEvalTests.cs
mmisol: Well, a lambda is a `Func` which pretty much works in the same way as a delegate. As long as you have a variable assignment to it you can unsubscribe it.

test/Libraries/DynamoPythonTests/PythonEvalTests.cs
mmisol: You are closing the string and initializing `count` in the same line.

test/Libraries/DynamoPythonTests/PythonEvalTests.cs
```diff
@@ -4,6 +4,8 @@
 using NUnit.Framework;
 using System.Linq;
 using ProtoCore.Lang;
+using DSCPython;
+using Python.Runtime;
```
mmisol, line 8: Please sort these

test/Libraries/DynamoPythonTests/PythonEvalTests.cs
```diff
@@ -171,6 +172,51 @@ public void CPythonEngineIncludesTraceBack()
             }
         }
 
+        [Test]
+        public void CPythonEngineWithErrorRaisesCorrectEvent()
+        {
+            var code = @"1";
```
mmisol, line 178: You might want to bring this closer to where it's executed for clarity.

test/Libraries/DynamoPythonTests/PythonEvalTests.cs
QilongTang: echo that

PR #10987, opened 2020-08-10
Fix line number for IronPython errors

mmisol:
### Purpose

The IronPython evaluator was adding an 'import sys' and a new line as a
prefix to the code entered in a Python script node. This was causing
errors to be reported in the the line next to where they actually
occurred.

The issue is fixed by removing the prefixed code, since it was not
really needed.

### Declarations

### Reviewers

@reddyashish @mjkkirschner 

### FYIs

@DynamoDS/dynamo

### DIFF
--- a/src/Libraries/DSIronPython/IronPythonEvaluator.cs
+++ b/src/Libraries/DSIronPython/IronPythonEvaluator.cs
@@ -104,7 +104,6 @@ namespace DSIronPython
                 ScriptEngine PythonEngine = Python.CreateEngine();
                 if (!string.IsNullOrEmpty(stdLib))
                 {
-                    code = "import sys" + System.Environment.NewLine + code;
                     paths = PythonEngine.GetSearchPaths().ToList();
                     paths.Add(stdLib);
                 }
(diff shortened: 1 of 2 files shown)


src/Libraries/DSIronPython/IronPythonEvaluator.cs
```diff
@@ -104,7 +104,6 @@ private static string PythonStandardLibPath()
                 ScriptEngine PythonEngine = Python.CreateEngine();
                 if (!string.IsNullOrEmpty(stdLib))
                 {
-                    code = "import sys" + System.Environment.NewLine + code;
```
mjkkirschner, line 107: why was this not required?

mmisol: Looking at the code, it did not make sense to me. I removed it and made sure the std lib is still loaded in the test.

reddyashish: This was being imported by default into all the Python scripts before. But If we remove it completely, I think this would break in the case where the users do not explicitly have this "import sys" in their code but try to access the "sys" data. They could just use sys.path without actually having to import.

mmisol:
The failing test is using `sys` without importing it.

Screen Shot 2020-08-11 at 9 19 15 AM

PR #10989, opened 2020-08-11
Dyn 3026 code coverage analysis folder

RobertGlobant20:
### Purpose

Increase the code coverage adding test cases for the Libraries/Analysis folder.
I created several unit tests methods using nUnit for the next classes:

- Label
- SurfaceData
- AnalysisExtensions
- Utils

Also I removed two methods from the SurfaceData.cs file since both have zero references and both are private.
The job DynamoSelfServe ran successfully in master-5.

### Declarations

### Reviewers

@QilongTang 

### FYIs

@alfredo-pozo

### DIFF
--- a/test/Libraries/AnalysisTests/AnalysisTests.csproj
+++ b/test/Libraries/AnalysisTests/AnalysisTests.csproj
@@ -56,9 +56,16 @@
     <Compile Include="..\..\..\src\AssemblySharedInfoGenerator\AssemblySharedInfo.cs">
       <Link>Properties\AssemblySharedInfo.cs</Link>
     </Compile>
+    <Compile Include="LabelTests.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
+    <Compile Include="SurfaceDataTests.cs" />
+    <Compile Include="UtilsTests.cs" />
   </ItemGroup>
   <ItemGroup>
+    <ProjectReference Include="..\..\..\src\DynamoCore\DynamoCore.csproj">
+      <Project>{7858FA8C-475F-4B8E-B468-1F8200778CF8}</Project>
+      <Name>DynamoCore</Name>
+    </ProjectReference>
     <ProjectReference Include="..\..\..\src\DynamoUtilities\DynamoUtilities.csproj">
       <Project>{b5f435cb-0d8a-40b1-a4f7-5ecb3ce792a9}</Project>
       <Name>DynamoUtilities</Name>
--- a/test/Libraries/AnalysisTests/AnalysisTests.csproj
+++ b/test/Libraries/AnalysisTests/AnalysisTests.csproj
@@ -36,9 +36,12 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="HelixToolkit.Wpf.SharpDX, Version=2.11.0.0, Culture=neutral, PublicKeyToken=52aa3500039caf0d" />
     <Reference Include="nunit.framework, Version=2.6.3.13283, Culture=neutral, PublicKeyToken=96d09a1eb7f44a77, processorArchitecture=MSIL">
       <HintPath>..\..\..\src\packages\NUnit.2.6.3\lib\nunit.framework.dll</HintPath>
     </Reference>
+    <Reference Include="PresentationCore" />
+    <Reference Include="PresentationFramework" />
     <Reference Include="ProtoGeometry, Version=1.0.0.0, Culture=neutral, processorArchitecture=MSIL">
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\..\..\extern\ProtoGeometry\ProtoGeometry.dll</HintPath>
@@ -50,6 +53,7 @@
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="System.Data" />
     <Reference Include="System.Xml" />
+    <Reference Include="WindowsBase" />
   </ItemGroup>
   <ItemGroup>
     <Compile Include="AnalysisTests.cs" />
@@ -62,6 +66,10 @@
     <Compile Include="UtilsTests.cs" />
   </ItemGroup>
   <ItemGroup>
+    <ProjectReference Include="..\..\..\src\DynamoCoreWpf\DynamoCoreWpf.csproj">
+      <Project>{51BB6014-43F7-4F31-B8D3-E3C37EBEDAF4}</Project>
+      <Name>DynamoCoreWpf</Name>
+    </ProjectReference>
     <ProjectReference Include="..\..\..\src\DynamoCore\DynamoCore.csproj">
       <Project>{7858FA8C-475F-4B8E-B468-1F8200778CF8}</Project>
       <Name>DynamoCore</Name>
@@ -75,6 +83,10 @@
       <Project>{76686ed6-e759-4772-81c2-768740be13fa}</Project>
       <Name>Analysis</Name>
     </ProjectReference>
+    <ProjectReference Include="..\..\..\src\Libraries\CoreNodeModels\CoreNodeModels.csproj">
+      <Project>{D8262D40-4880-41E4-91E4-AF8F480C8637}</Project>
+      <Name>CoreNodeModels</Name>
+    </ProjectReference>
     <ProjectReference Include="..\..\..\src\Libraries\CoreNodes\CoreNodes.csproj">
       <Project>{87550b2b-6cb8-461e-8965-dfafe3aafb5c}</Project>
       <Name>CoreNodes</Name>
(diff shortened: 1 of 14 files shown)


RobertGlobant20:
This is a screenshot about the code coverage for the Analysis project using DotCover command line:

analysis_code_coverage

mjkkirschner:
@RobertGlobant20 did you ever get these methods to actually draw labels into the background preview?

RobertGlobant20:
> @RobertGlobant20 did you ever get these methods to actually draw labels into the background preview?

@mjkkirschner No, I tried to find references to the Label class (or classes that inherit of Label class) and is only used in the Label.cs and the test class that i created, then I think there is no way to used in other places, probably the labels in the background preview is using a different class.

test/Libraries/AnalysisTests/LabelTests.cs
```diff
@@ -0,0 +1,38 @@
...
+        [Test, Category("UnitTests")]
+        public void LabelByPointAndStringTest()
+        {
+            //Arrange
+            var labelTest = Label.ByPointAndString(Point.ByCoordinates(100,100), "TestingLabel");
+            var factory = new DefaultRenderPackageFactory();
+            var package = factory.CreateRenderPackage();
+
+            //Act
+            labelTest.Tessellate(package, new TessellationParameters());
```
aparajit-pratap, line 28: What are you testing from the `Tessellate` call? I think the assertions below will pass even if you don't call `Tessellate`, wouldn't they?

test/Libraries/AnalysisTests/LabelTests.cs
```diff
@@ -0,0 +1,38 @@
...
+
+            //Assert
+            Assert.IsNotNull(labelTest);
```
aparajit-pratap, line 31: I think this assertion should be before the call to `Tessellate`. Also, can you check if certain properties of `Label` are set in addition to checking for it to be non-null?

test/Libraries/AnalysisTests/SurfaceDataTests.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// This test method will execute the BySurfaceAndPoints method from the SurfaceData class passing different parameters
+        /// </summary>
+        [Test, Category("UnitTests")]
+        public void SurfaceAnalysisDataBySurfaceAndPointsTest_Exceptions()
+        {
+            //Act
+            var sad = SurfaceData.BySurfaceAndPoints(TestSurface(),TestUvs());
+
+            //Assert
+            Assert.NotNull(sad);
```
aparajit-pratap, line 22: Unrelated, but what is the difference between `Assert.IsNotNull` and `NotNull`?

test/Libraries/AnalysisTests/SurfaceDataTests.cs
```diff
@@ -0,0 +1,96 @@
...
+            Assert.IsNull(sad.Values);
+
+            //It will raise an ArgumentNullException when we pass the Surface as null
+            Assert.Throws<ArgumentNullException>(() => SurfaceData.BySurfaceAndPoints(null, TestUvs()));
+
+            //It will raise an ArgumentNullException when we pass the UV values as null
+            Assert.Throws<ArgumentNullException>(() => SurfaceData.BySurfaceAndPoints(TestSurface(), null));
+
+            //It will raise an ArgumentNullException when we pass the UV values as empty IEnumerable
+            Assert.Throws<ArgumentException>(() => SurfaceData.BySurfaceAndPoints(TestSurface(), new UV[0]));
```
aparajit-pratap, line 34: If it throws `ArgumentNullException`, shouldn't you check for `ArgumentNullException` specifically?

test/Libraries/AnalysisTests/LabelTests.cs
RobertGlobant20: @aparajit-pratap 
The Tessellate method is a [IsVisibleInDynamoLibrary(false)], has no code coverage and is not used (inside the Label class o in other classes, event the whole Label class is not used), I was proposing to delete the whole Label.cs file but Michael said the next "if it's public, we cannot get rid of it as it will break semantic versioning."
You are correct, the assertions are checking the results of Label.ByPointAndString method then they don't depend of calling the  Tessellate method, then is it ok if I just delete the call to the Tessellate method (of course if won't have coverage)?

test/Libraries/AnalysisTests/SurfaceDataTests.cs
RobertGlobant20: @aparajit-pratap 
The SurfaceData.BySurfaceAndPoints method throws a ArgumentNullException when the first parameter or second parameter are null but when the IEnumerable<UV> parameter is empty it throws a ArgumentException, the I think is correct in this way.
Let me know if you agree.

test/Libraries/AnalysisTests/LabelTests.cs
RobertGlobant20: @aparajit-pratap 
This depends of the previous comment, if I delete the call to Tessellate method there is no need to do it, otherwise I will move it before the call.

test/Libraries/AnalysisTests/SurfaceDataTests.cs
RobertGlobant20: @aparajit-pratap 
I was reading that IsNotNull and NotNull are the same but IsNotNull is compatible to previous versions of nUnit then I updated two lines for using IsNotNull.

commit: https://github.com/DynamoDS/Dynamo/pull/10989/commits/0fbc2034d4f413ca7cfe9ae76c525eacb51be946

test/Libraries/AnalysisTests/LabelTests.cs
aparajit-pratap: It looks like the package properties are modified in Tessellate. You can check for those properties after the call to Tessellate.

mjkkirschner:
Hi @RobertGlobant20 - whats odd is that https://github.com/DynamoDS/Dynamo/blob/master/src/Libraries/Analysis/Label.cs#L42
should create a point in the background at least.

test/Libraries/AnalysisTests/LabelTests.cs
RobertGlobant20: @aparajit-pratap this test method was re-written due to some comments that Michael did, please review again.

RobertGlobant20:
> Hi @RobertGlobant20 - whats odd is that https://github.com/DynamoDS/Dynamo/blob/master/src/Libraries/Analysis/Label.cs#L42
> should create a point in the background at least.

@mjkkirschner The tests for the Label class were rewritten and also I applied changes in the Label.cs and HelixRenderPackage.cs files, please review again.

mjkkirschner:
hmm, @RobertGlobant20 it looks like this change causes the mono.sln build to fail - thats because now the analysis tests require sharpDX and WPF references.

I think we have two options.

1. remove the analysis tests from the mono.sln
2. Add those tests that require those references to the wpfvisualizationTests project instead of to the analysisTests project, so the analysis test project does not need these windows only references.

what do you think?

test/Libraries/AnalysisTests/AnalysisTests.csproj
```diff
@@ -68,6 +83,10 @@
       <Project>{76686ed6-e759-4772-81c2-768740be13fa}</Project>
       <Name>Analysis</Name>
     </ProjectReference>
+    <ProjectReference Include="..\..\..\src\Libraries\CoreNodeModels\CoreNodeModels.csproj">
+      <Project>{D8262D40-4880-41E4-91E4-AF8F480C8637}</Project>
+      <Name>CoreNodeModels</Name>
```
mjkkirschner, line 88: @RobertGlobant20 - this change is causing atleast some of the 600 failures you are getting.
Change this to `<Private>False<Private>` or in visual studio CopyLocal False, same thing.

This binary is being loaded twice, once from the /nodes folder, and now again from the /bin folder (in your PR). It should only be present in the /nodes folder.

To be safe, I would make all of these dependencies copy Local false, but thats probably not the problem.

RobertGlobant20:
> hmm, @RobertGlobant20 it looks like this change causes the mono.sln build to fail - thats because now the analysis tests require sharpDX and WPF references.
> 
> I think we have two options.
> 
> 1. remove the analysis tests from the mono.sln
> 2. Add those tests that require those references to the wpfvisualizationTests project instead of to the analysisTests project, so the analysis test project does not need these windows only references.
> 
> what do you think?

@mjkkirschner I took the option 1, I moved the LabelTests.cs to the DynamoCoreWpfTests project inside a Libraries folder, then I removed the sharpDX and Wpf references from the AnalysisTests project.

test/Libraries/AnalysisTests/AnalysisTests.csproj
RobertGlobant20: @mjkkirschner I removed the CoreNodeModels reference from the AnalysisTests project, then there shouldn't be any problem now.
Thanks

RobertGlobant20:
> let's wait for the tests, but otherwise looks good.
@mjkkirschner  The Tests were successful executed.
https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/master/

PR #10994, opened 2020-08-13
Add files via upload

Amoursol:
The Dynamo blog (Wordpress website) cannot host .DYN files and we need to do so for a blog post by Colin McCrone on post-COVID office layout.

### DIFF
--- a/doc/models/COVID occupancy by instance.dyn
+++ b/doc/models/COVID occupancy by instance.dyn
@@ -1,0 +1,7360 @@
+{
+  "Uuid": "16de11db-935d-4a5e-a79e-713a6f8911ba",
+  "IsCustomNode": false,
+  "Description": null,
+  "Name": "COVID occupancy by instance",
+  "ElementResolver": {
+    "ResolutionMap": {
+      "Vector": {
+        "Key": "Autodesk.DesignScript.Geometry.Vector",
+        "Value": "ProtoGeometry.dll"
+      },
+      "DSCore.Color": {
+        "Key": "DSCore.Color",
+        "Value": "DSCoreNodes.dll"
+      },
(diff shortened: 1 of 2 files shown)


mmisol:
HI @Amoursol . There are alternatives to uploading these files other than putting them in the main Dynamo GitHub repository. Given they are not related to the source and the need to upload files for users in the forum will continue to exist after this, I would choose not to go in this direction. Could we at least use another repo? Temporary storage for files can be achieved in a number of ways while we come up with a better solution to the problem.

Amoursol:
Totally cool by me - I can look for a better place to house these :)

QilongTang:
I actually think it is good and better live in Dynamo repo. Given 

- this repo already contains a lot of Dynamo test graphs
- dyn files do not take much space comparing to other binaries cases
- side repos are easy to create but always got forgotten later. 

One alternative is that @Amoursol You can use your personal Github storage for this. The way to do it is that you drag the file to PR which will generate a link for you and you use that link for blog post. But will appear less official than current approach. I am supportive of this way.

However, open for suggestions. @DynamoDS/dynamo Any other ways you can think of or comments?

mmisol:
@QilongTang I would argue this is in a different category than graphs used for tests. If we start taking graphs made by the community, I'm thinking it should not be up to us to maintain them. Also there may be legal implications just by accepting them in our repo.

QilongTang:
@mmisol I forgot to mention the author of this is from ADSK not community if that is your concern. Colin used to work for Dynamo team but now moved on to Forge. I do not think our team plan to store or maintain graphs from community in general.

Amoursol:
Closing this PR as it's not the ideal place to house Wordpress content.

PR #10998, opened 2020-08-13
Define default Python engine by system and user

mmisol:
### Purpose

Adds the ability to define a default Python engine for the system. This
is done through a new property in the model start configuration, named
'DefaultPythonEngine'.

Also adds the ability for users to define through a setting their own
default. This can be set to either engine or to fallback to the system
default, whatever it is.

Serialization is tweaked so that 'IronPython2' remains the
deserialization default but is not the absolute default. Engine will be
unspecified on node creation but become loaded as soon as the property
is accessed, to avoid errors on automatic execution.

In order to make system and user settings available on the node, and
avoiding further changes to the API, the properties were made static
and internal.

### Declarations

### Reviewers

@mjkkirschner @QilongTang 

### FYIs

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoApplications/StartupUtils.cs
+++ b/src/DynamoApplications/StartupUtils.cs
@@ -98,8 +98,6 @@ namespace Dynamo.Applications
 
                 var asmPath = string.Empty;
 
-                //var pythonEngine = string.Empty;
-
                 bool showHelp = false;
                 var optionsSet = new OptionSet().Add("o=|O=", "OpenFilePath, Instruct Dynamo to open headless and run a dyn file at this path", o => openfilepath = o)
                 .Add("c=|C=", "CommandFilePath, Instruct Dynamo to open a commandfile and run the commands it contains at this path," +
@@ -111,9 +109,8 @@ namespace Dynamo.Applications
                 .Add("h|H|help", "Get some help", h => showHelp = h != null)
                 .Add("g=|G=|geometry", "Geometry, Instruct Dynamo to output geometry from all evaluations to a json file at this path", g => geometryFilePath = g)
                 .Add("i=|I=|import", "Import, Instruct Dynamo to import an assembly as a node library. This argument should be a filepath to a single .dll" +
-                " - if you wish to import multiple dlls - use this flag multiple times: -i 'assembly1.dll' -i 'assembly2.dll' ", i => importPaths.Add(i))
-                .Add("gp=|GP=|geometrypath=|GeometryPath=", "relative or absolute path to a directory containing ASM. When supplied, instead of searching the hard disk for ASM, it will be loaded directly from this path.", gp => asmPath = gp);
-                //.Add("py=|PY=|pythonEngine=|PythonEngine=", "Default Python script engine", py => pythonEngine = py);
+                " - if you wish to import multiple dlls - use this flag multiple times: -i 'assembly1.dll' -i 'assembly2.dll' ", i => importPaths.Add(i)).
+                Add("gp=|GP=|geometrypath=|GeometryPath=", "relative or absolute path to a directory containing ASM. When supplied, instead of searching the hard disk for ASM, it will be loaded directly from this path.", gp => asmPath = gp);
 
                 optionsSet.Parse(args);
 
@@ -137,7 +134,6 @@ namespace Dynamo.Applications
                     GeometryFilePath = geometryFilePath,
                     ImportedPaths = importPaths,
                     ASMPath = asmPath
-                    //PythonEngine = pythonEngine
                 };
             }
 
@@ -155,10 +151,6 @@ namespace Dynamo.Applications
             public string GeometryFilePath { get; set; }
             public IEnumerable<String> ImportedPaths { get; set; }
             public string ASMPath { get; set; }
-            /// <summary>
-            /// Default Python script engine
-            /// </summary>
-            //public string PythonEngine { get; set; }
         }
 
         public static void PreloadShapeManager(ref string geometryFactoryPath, ref string preloaderLocation)
--- a/src/DynamoApplications/StartupUtils.cs
+++ b/src/DynamoApplications/StartupUtils.cs
@@ -98,6 +98,8 @@ namespace Dynamo.Applications
 
                 var asmPath = string.Empty;
 
+                //var pythonEngine = string.Empty;
+
                 bool showHelp = false;
                 var optionsSet = new OptionSet().Add("o=|O=", "OpenFilePath, Instruct Dynamo to open headless and run a dyn file at this path", o => openfilepath = o)
                 .Add("c=|C=", "CommandFilePath, Instruct Dynamo to open a commandfile and run the commands it contains at this path," +
@@ -109,8 +111,9 @@ namespace Dynamo.Applications
                 .Add("h|H|help", "Get some help", h => showHelp = h != null)
                 .Add("g=|G=|geometry", "Geometry, Instruct Dynamo to output geometry from all evaluations to a json file at this path", g => geometryFilePath = g)
                 .Add("i=|I=|import", "Import, Instruct Dynamo to import an assembly as a node library. This argument should be a filepath to a single .dll" +
-                " - if you wish to import multiple dlls - use this flag multiple times: -i 'assembly1.dll' -i 'assembly2.dll' ", i => importPaths.Add(i)).
-                Add("gp=|GP=|geometrypath=|GeometryPath=", "relative or absolute path to a directory containing ASM. When supplied, instead of searching the hard disk for ASM, it will be loaded directly from this path.", gp => asmPath = gp);
+                " - if you wish to import multiple dlls - use this flag multiple times: -i 'assembly1.dll' -i 'assembly2.dll' ", i => importPaths.Add(i))
+                .Add("gp=|GP=|geometrypath=|GeometryPath=", "relative or absolute path to a directory containing ASM. When supplied, instead of searching the hard disk for ASM, it will be loaded directly from this path.", gp => asmPath = gp);
+                //.Add("py=|PY=|pythonEngine=|PythonEngine=", "Default Python script engine", py => pythonEngine = py);
 
                 optionsSet.Parse(args);
 
@@ -133,7 +136,8 @@ namespace Dynamo.Applications
                     ConvertFile = convertFile,
                     GeometryFilePath = geometryFilePath,
                     ImportedPaths = importPaths,
-                    ASMPath = asmPath,
+                    ASMPath = asmPath
+                    //PythonEngine = pythonEngine
                 };
             }
 
@@ -151,6 +155,10 @@ namespace Dynamo.Applications
             public string GeometryFilePath { get; set; }
             public IEnumerable<String> ImportedPaths { get; set; }
             public string ASMPath { get; set; }
+            /// <summary>
+            /// Default Python script engine
+            /// </summary>
+            //public string PythonEngine { get; set; }
         }
 
         public static void PreloadShapeManager(ref string geometryFactoryPath, ref string preloaderLocation)
--- a/src/DynamoCore/Configuration/PreferenceSettings.cs
+++ b/src/DynamoCore/Configuration/PreferenceSettings.cs
@@ -334,9 +334,24 @@ namespace Dynamo.Configuration
         public bool ShowTabsAndSpacesInScriptEditor { get; set; }
 
         /// <summary>
-        /// Engine used by default for new Python script and string nodes. If defined this takes precedence over any system settings.
+        /// Engine used by default for new Python script and string nodes. If not empty, this takes precedence over any system settings.
         /// </summary>
-        public string DefaultPythonEngine { get; set; }
+        public string DefaultPythonEngine
+        {
+            get
+            {
+                return defaultPythonEngine;
+            }
+            set
+            {
+                defaultPythonEngine = value;
+            }
+        }
+
+        /// <summary>
+        /// Static field backing the DefaultPythonEngine setting property.
+        /// </summary>
+        private static string defaultPythonEngine;
 
         /// <summary>
         /// Indicates (if any) which namespaces should not be displayed in the Dynamo node library.
@@ -480,6 +495,15 @@ namespace Dynamo.Configuration
         public static string GetPythonTemplateFilePath()
         {
             return pythonTemplateFilePath;
+        }
+
+        /// <summary>
+        /// Provides access to the DefaultPythonEngine setting in a static context. Used from PythonNodeBase.
+        /// </summary>
+        /// <returns>DefaultPythonEngine setting value</returns>
+        internal static string GetDefaultPythonEngine()
+        {
+            return defaultPythonEngine;
         }
 
         internal void InitializeNamespacesToExcludeFromLibrary()
--- a/src/DynamoCore/Configuration/PreferenceSettings.cs
+++ b/src/DynamoCore/Configuration/PreferenceSettings.cs
@@ -334,6 +334,11 @@ namespace Dynamo.Configuration
         public bool ShowTabsAndSpacesInScriptEditor { get; set; }
 
         /// <summary>
+        /// Engine used by default for new Python script and string nodes. If defined this takes precedence over any system settings.
+        /// </summary>
+        public string DefaultPythonEngine { get; set; }
+
+        /// <summary>
         /// Indicates (if any) which namespaces should not be displayed in the Dynamo node library.
         /// String format: "[library name]:[fully qualified namespace]"
         /// </summary>
@@ -388,6 +393,7 @@ namespace Dynamo.Configuration
             PythonTemplateFilePath = "";
             IsIronPythonDialogDisabled = false;
             ShowTabsAndSpacesInScriptEditor = false;
+            DefaultPythonEngine = string.Empty;
         }
 
         /// <summary>
--- a/src/DynamoCore/Graph/Nodes/NodeModel.cs
+++ b/src/DynamoCore/Graph/Nodes/NodeModel.cs
@@ -2515,6 +2515,11 @@ namespace Dynamo.Graph.Nodes
 
         #endregion
 
+        public virtual void OnBeforeAdd(Dynamo.Models.DynamoModel dynamoModel, WorkspaceModel workspace)
+        {
+            // Purposely empty. For inheritors to override if needed.
+        }
+
         [JsonIgnore]
         public bool ShouldDisplayPreview
         {
--- a/src/DynamoCore/Graph/Nodes/NodeModel.cs
+++ b/src/DynamoCore/Graph/Nodes/NodeModel.cs
@@ -2515,11 +2515,6 @@ namespace Dynamo.Graph.Nodes
 
         #endregion
 
-        public virtual void OnBeforeAdd(Dynamo.Models.DynamoModel dynamoModel, WorkspaceModel workspace)
-        {
-            // Purposely empty. For inheritors to override if needed.
-        }
-
         [JsonIgnore]
         public bool ShouldDisplayPreview
         {
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -115,7 +115,7 @@ namespace Dynamo.Models
         private Timer backupFilesTimer;
         private Dictionary<Guid, string> backupFilesDict = new Dictionary<Guid, string>();
         internal readonly Stopwatch stopwatch = Stopwatch.StartNew();
-        private string defaultPythonEngine;
+        private static string defaultPythonEngine;
 
         #endregion
 
@@ -425,7 +425,7 @@ namespace Dynamo.Models
         /// </summary>
         public AuthenticationManager AuthenticationManager { get; set; }
 
-        internal string DefaultPythonEngine
+        internal static string DefaultPythonEngine
         {
             get
             {
@@ -511,14 +511,6 @@ namespace Dynamo.Models
             /// No update checks or analytics collection should be done.
             /// </summary>
             bool IsHeadless { get; set; }
-            /// <summary>
-            /// Default Python script engine
-            /// </summary>
-            string DefaultPythonEngine { get; set; }
-            /// <summary>
-            /// Provide values for startup options
-            /// </summary>
-            //IDictionary<string, string> StartupOptions { get; set; }
         }
 
         /// <summary>
@@ -540,8 +532,10 @@ namespace Dynamo.Models
             public TaskProcessMode ProcessMode { get; set; }
             public bool IsHeadless { get; set; }
             public string PythonTemplatePath { get; set; }
+            /// <summary>
+            /// Default Python script engine
+            /// </summary>
             public string DefaultPythonEngine { get; set; }
-            //public IDictionary<string, string> StartupOptions { get; set; }
         }
 
         /// <summary>
@@ -573,14 +567,12 @@ namespace Dynamo.Models
         /// <param name="config">Start configuration</param>
         protected DynamoModel(IStartConfiguration config)
         {
-            //if (config.StartupOptions != null)
-            //{
-            //    if (config.StartupOptions.ContainsKey("PythonEngine"))
-            //    {
-            //        defaultPythonEngine = config.StartupOptions["PythonEngine"];
-            //    }
-            //}
-            defaultPythonEngine = config.DefaultPythonEngine;
+            if (config is DefaultStartConfiguration)
+            {
+                // This is not exposed in IStartConfiguration to avoid a breaking change.
+                // This fact should probably be revisited in next major.
+                defaultPythonEngine = ((DefaultStartConfiguration)config).DefaultPythonEngine;
+            }
 
             ClipBoard = new ObservableCollection<ModelBase>();
 
@@ -2232,7 +2224,6 @@ namespace Dynamo.Models
         /// should be selected</param>
         internal void AddNodeToCurrentWorkspace(NodeModel node, bool centered, bool addToSelection = true)
         {
-            node.OnBeforeAdd(this, CurrentWorkspace);
             CurrentWorkspace.AddAndRegisterNode(node, centered);
 
             //TODO(Steve): This should be moved to WorkspaceModel.AddNode when all workspaces have their own selection -- MAGN-5707
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -563,7 +563,7 @@ namespace Dynamo.Models
             if (config is DefaultStartConfiguration)
             {
                 // This is not exposed in IStartConfiguration to avoid a breaking change.
-                // TODO: This fact should probably be revisited in next major.
+                // TODO: This fact should probably be revisited in 3.0.
                 DefaultPythonEngine = ((DefaultStartConfiguration)config).DefaultPythonEngine;
             }
 
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -115,7 +115,6 @@ namespace Dynamo.Models
         private Timer backupFilesTimer;
         private Dictionary<Guid, string> backupFilesDict = new Dictionary<Guid, string>();
         internal readonly Stopwatch stopwatch = Stopwatch.StartNew();
-        private static string defaultPythonEngine;
 
         #endregion
 
@@ -425,13 +424,7 @@ namespace Dynamo.Models
         /// </summary>
         public AuthenticationManager AuthenticationManager { get; set; }
 
-        internal static string DefaultPythonEngine
-        {
-            get
-            {
-                return defaultPythonEngine;
-            }
-        }
+        internal static string DefaultPythonEngine { get; private set; }
 
         #endregion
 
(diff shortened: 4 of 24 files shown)


mmisol:
Just wanted to make it clear this is a Work In Progress. Please ignore the commented out code. The idea is to get some feedback on how this looks before wrapping up, since there are several questionable decisions made.

src/DynamoApplications/StartupUtils.cs
```diff
@@ -109,8 +111,9 @@ public static CommandLineArguments Parse(string[] args)
...
                 .Add("i=|I=|import", "Import, Instruct Dynamo to import an assembly as a node library. This argument should be a filepath to a single .dll" +
-                " - if you wish to import multiple dlls - use this flag multiple times: -i 'assembly1.dll' -i 'assembly2.dll' ", i => importPaths.Add(i)).
-                Add("gp=|GP=|geometrypath=|GeometryPath=", "relative or absolute path to a directory containing ASM. When supplied, instead of searching the hard disk for ASM, it will be loaded directly from this path.", gp => asmPath = gp);
+                " - if you wish to import multiple dlls - use this flag multiple times: -i 'assembly1.dll' -i 'assembly2.dll' ", i => importPaths.Add(i))
+                .Add("gp=|GP=|geometrypath=|GeometryPath=", "relative or absolute path to a directory containing ASM. When supplied, instead of searching the hard disk for ASM, it will be loaded directly from this path.", gp => asmPath = gp);
+                //.Add("py=|PY=|pythonEngine=|PythonEngine=", "Default Python script engine", py => pythonEngine = py);
```
mmisol, line 116: At some point I thought this could also be a command line option, but then disregarded that. Integrators won't use that and not sure we would ever use it for Sandbox. It's also not fully implemented.

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -2515,6 +2515,11 @@ public static NodeMigrationData MigrateShortestLacingToAutoLacing(NodeMigrationD
 
         #endregion
 
+        public virtual void OnBeforeAdd(Dynamo.Models.DynamoModel dynamoModel, WorkspaceModel workspace)
+        {
+            // Purposely empty. For inheritors to override if needed.
+        }
```
mmisol, line 2521: In my mind it made sense to add this sort of life-cycle method to the node. It allowed to act at the right time and provides the required `DynamoModel`.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -501,6 +511,14 @@ public interface IStartConfiguration
...
             /// </summary>
             bool IsHeadless { get; set; }
+            /// <summary>
+            /// Default Python script engine
+            /// </summary>
+            string DefaultPythonEngine { get; set; }
+            /// <summary>
+            /// Provide values for startup options
+            /// </summary>
+            //IDictionary<string, string> StartupOptions { get; set; }
```
mmisol, line 521: This is like an alternative API I had in mind at first. It would allow to provide command-line options in a generic way. The only benefit I can think about it is that we don't add an explicit Python property to the config. On the other hand, it is messier to add entries to a dictionary than assigning properties.

src/DynamoCoreWpf/UI/Converters.cs
```diff
@@ -1108,6 +1108,48 @@ public object ConvertBack(object value, Type targetType, object parameter, Syste
...
+    //        return DependencyProperty.UnsetValue;
+    //    }
+
+    //    public object[] ConvertBack(object value, Type[] targetTypes, object parameter, CultureInfo culture)
+    //    {
+    //        // No way to implement this. Multi-binding can't achieve this
+    //        throw new NotImplementedException();
+    //    }
+    //}
+
```
mmisol, line 1152: Some experiments I did with WPF before ending up with the current UI.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml
```diff
@@ -724,6 +724,55 @@
                                   IsCheckable="True"
                                   IsChecked="{Binding ShowTabsAndSpacesInScriptEditor, Mode=TwoWay}"
                                   Header="{x:Static p:Resources.DynamoViewSettingsMenuShowTabsAndSpacesInScriptEditor}"/>
+                        <MenuItem Focusable="False"
+                                  Header="{x:Static p:Resources.DynamoViewSettingsDefaultPythonEngine}">
+                            <ComboBox ItemsSource="{Binding DefaultPythonEngineOptions}" SelectedValue="{Binding DefaultPythonEngine, Mode=TwoWay}" SelectedValuePath="Value" DisplayMemberPath="Name"/>
+                        </MenuItem>
```
mmisol, line 730: Used a drop down. As you can see from the commented markup I also tried child checkable menu items, which had some limitations. Open to discussion.

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -29,13 +32,15 @@ public PythonCodeMigrationEventArgs(string oldCode, string newCode)
 
     public abstract class PythonNodeBase : VariableInputNode
     {
-        private PythonEngineVersion engine = PythonEngineVersion.IronPython2;
+        private PythonEngineVersion? engine = null;
```
mmisol, line 35: The engine became nullable, as it cannot be initialized at this point. We have no access to `DynamoModel`. Luckily JSON serialization handles this change without problems.

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -54,6 +59,30 @@ protected PythonNodeBase()
...
+                }
+                else if (!string.IsNullOrEmpty(systemDefault) && Enum.TryParse(setting, out version))
+                {
+                    Engine = version;
+                }
+                else
+                {
+                    // In the absence of both a setting and system default, default to deserialization default.
+                    Engine = PythonEngineVersion.IronPython2;
+                }
```
mmisol, line 82: user setting > system default > deserialization default

src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -333,6 +333,11 @@ public string PythonTemplateFilePath
         /// </summary>
         public bool ShowTabsAndSpacesInScriptEditor { get; set; }
 
+        /// <summary>
+        /// Engine used by default for new Python script and string nodes. If defined this takes precedence over any system settings.
```
mjkkirschner, line 337: do you mean if not null? Does empty string qualify as defined?

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -2515,6 +2515,11 @@ public static NodeMigrationData MigrateShortestLacingToAutoLacing(NodeMigrationD
 
         #endregion
 
+        public virtual void OnBeforeAdd(Dynamo.Models.DynamoModel dynamoModel, WorkspaceModel workspace)
```
mjkkirschner, line 2518: this is quite a big change - in the rest of Dynamo it's common for models to only have references to their children - ie DynamoModel knows about NodeModels - and NodeModels know about portModels, but not the other way around.

Curious why you need this reference - maybe it will become clear.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -501,6 +511,14 @@ public interface IStartConfiguration
             /// No update checks or analytics collection should be done.
             /// </summary>
             bool IsHeadless { get; set; }
+            /// <summary>
+            /// Default Python script engine
+            /// </summary>
+            string DefaultPythonEngine { get; set; }
```
mjkkirschner, line 517: modifying this interface is an API break - we can add a new interface instead, and merge them together in 3.x

src/DynamoCore/Models/DynamoModel.cs
mjkkirschner: I've always thought it would be nice to be able to override the entire config with an xml or json file - but thats a discussion for another time I think.

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -46,6 +47,20 @@ public interface IDynamoViewModel : INotifyPropertyChanged
         ObservableCollection<WorkspaceViewModel> Workspaces { get; set; } 
     }
 
+    public class SimpleViewModel<T> : ViewModelBase, INotifyPropertyChanged
```
mjkkirschner, line 50: experiment?

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -29,13 +32,15 @@ public PythonCodeMigrationEventArgs(string oldCode, string newCode)
...
+        private PythonEngineVersion? engine = null;
 
         [JsonConverter(typeof(StringEnumConverter))]
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Populate)]
+        [DefaultValue("IronPython2")]
         /// <summary>
         /// Return the user selected python engine enum.
         /// </summary>
-        public PythonEngineVersion Engine
+        public PythonEngineVersion? Engine
```
mjkkirschner, line 43: I wonder if this qualifies as an API break...

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -88,7 +117,7 @@ protected override string GetInputTooltip(int index)
 
             // Here we switched to use the AstFactory.BuildFunctionCall version that accept
             // class name and function name. They will be set by PythonEngineSelector by the engine value. 
-            PythonEngineSelector.Instance.GetEvaluatorInfo(Engine, out string evaluatorClass, out string evaluationMethod);
+            PythonEngineSelector.Instance.GetEvaluatorInfo(Engine.Value, out string evaluatorClass, out string evaluationMethod);
```
mjkkirschner, line 120: I guess so?

src/DynamoCore/Graph/Nodes/NodeModel.cs
mjkkirschner: hmm - it makes a lot of sense that the python nodeModel needs to continue working even in headless Dynamo.

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -54,6 +59,30 @@ protected PythonNodeBase()
             ArgumentLacing = LacingStrategy.Disabled;
         }
 
+        public override void OnBeforeAdd(DynamoModel dynamoModel, WorkspaceModel workspace)
```
mjkkirschner, line 62: workspace looks unused - just thinking ahead?

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
mjkkirschner: apparently you can use old style tuple items as bindings - you cannot use new style value tuples though as those members are fields - anyhow, something I consider I guess.

mjkkirschner: https://stackoverflow.com/questions/43208852/is-it-possible-to-bind-to-a-valuetuple-field-in-wpf-with-c7

src/DynamoCore/Graph/Nodes/NodeModel.cs
mjkkirschner: I'm curious what you think of using the active `ExecutionSession` instead - storing the pythonEngine there and extracting it from the static map when the node is constructed.

src/DynamoCore/Configuration/PreferenceSettings.cs
mmisol: I'm pretty sure I'm using empty string for not defined.

src/DynamoCore/Graph/Nodes/NodeModel.cs
mmisol: I'm open to change this. `ExecutingSession` is looking uglier each time we add another new thing to it, but I guess it can be used.

This does expose more data, but we are not being super consistent with the visibility anyway. For view models the rules are pretty different than what you mention here.

I don't think this would break headless Dynamo, unless you added a node to a graph dynamically (can you do that headless?) and forgot to assign an `Engine` to it.

src/DynamoCore/Models/DynamoModel.cs
mmisol: An alternative can be not have this in the interface but keep it in `DefaultStartConfiguration`, and cast and see in `DynamoModel` constructor. I see some other properties on this boat already.

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
mmisol: Maybe not the right place for it, but I thought this could be a good way to to it. I'm not a fan of tuples, specially because the property names are positional and have no meaning.

src/Libraries/PythonNodeModels/PythonNode.cs
mmisol: It does.

src/Libraries/PythonNodeModels/PythonNode.cs
mmisol: In my mind, this "life-cycle method" made more sense if you only got the workspace where the node was being added. However, you have the same problem, `WorkspaceModel` has no visibility of the `DynamoModel` associated with it. In the end I kept both because only providing the `DynamoModel` made less sense as an API, but it's true it's not really used.

src/DynamoCore/Configuration/PreferenceSettings.cs
mjkkirschner: can you document that in this comment?

src/DynamoCore/Graph/Nodes/NodeModel.cs
mjkkirschner: you can add a node to a graph dynamically in headless mode for sure - consider some headless client of Dynamo that calls the createNodeCommand - but - I was actually trying to make the point that I thought your solution was better than my original suggestion because this doesn't require the NodeView to be around to correctly set the engine.

mjkkirschner: for some context - I believe historically, there was a giant static class called something like `DynamoSettings` or `DynamoModelEntryPoint` that was used from everywhere, this made the separation between core and wpf layers very difficult to enforce, and a big refactor was undertaken.

There was also the idea that one day we'd have multiple DynamoModels running at the same time in the same process, but this hit other roadblocks.

I'm not sure the pain for nodeModel developers (like you here) is worth the old ideas from that refactor any longer.

src/Libraries/PythonNodeModels/PythonNode.cs
mjkkirschner: that seems like a problem ;) Can we add a new enum like none?

src/DynamoCore/Graph/Nodes/NodeModel.cs
mmisol: Multiple DynamoModel sounds intriguing. In any case a node would belong to just one of those, right? 

Giving a node model access to its parent workspace model and that one having access to its parent dynamo model does not sound so bad to me. It feels having access to the context a node model is in should be more direct than relying on global classes. 

In any case, we probably don't want to make a refactor part of this PR. I'll stick to `ExecutingSession` for now.

src/Libraries/PythonNodeModels/PythonNode.cs
mmisol: I can do this. I just felt we weren't tied to this API yet since it was recently added, but I guess people may be using it already.

src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -388,6 +393,7 @@ public PreferenceSettings()
             PythonTemplateFilePath = "";
             IsIronPythonDialogDisabled = false;
             ShowTabsAndSpacesInScriptEditor = false;
+            DefaultPythonEngine = string.Empty;
```
QilongTang, line 396: Do we want to make PythonEngineVersion enum static so we can refer anywhere with known value options?

src/DynamoCore/Graph/Nodes/NodeModel.cs
mmisol: Now thinking again from a security perspective, accessing the `DynamoModel` opens the door for a node to do a bunch of things we may not want them to do. Note that the risk currently exists for a custom `NodeView` as they can access the `DynamoViewModel` and then to the `DynamoModel`.

It might be a  good idea to think this further and consider refactoring in 3.0.

src/DynamoCore/Configuration/PreferenceSettings.cs
mmisol: Do you mean make the `DefaultPythonEngine` preference setting static? I guess that could work, but we also have the system setting, so we'd have to make them both static I guess. It could be an alternative to putting stuff into `ExecutingSession`, which may be cleaner if we can use `InternalsVisibleTo`. Will check that out.

mmisol:
Based on the feedback, I went with internal static properties and got rid of `OnBeforeAdd`. Also cleaned up and added tests. Please take another look @mjkkirschner @QilongTang

src/DynamoCore/Models/DynamoModel.cs
Dewb: We already have multiple, partially overlapping systems to manage config, with StartConfiguration, IPreferences, StartupUtils.CommandLineArguments, properties copied into models, etc. Rationalizing all of this should be a priority for the next API-breaking release, would love someone to research and draft a proposal.

mmisol:
Tests are passing now https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/1284/

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -423,6 +425,14 @@ public IEnumerable<WorkspaceModel> Workspaces
         /// </summary>
         public AuthenticationManager AuthenticationManager { get; set; }
 
+        internal static string DefaultPythonEngine
+        {
+            get
+            {
+                return defaultPythonEngine;
```
mjkkirschner, line 432: just curious, why is this not using an auto property?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -553,6 +567,13 @@ public static DynamoModel Start(IStartConfiguration configuration)
         /// <param name="config">Start configuration</param>
         protected DynamoModel(IStartConfiguration config)
         {
+            if (config is DefaultStartConfiguration)
+            {
+                // This is not exposed in IStartConfiguration to avoid a breaking change.
```
mjkkirschner, line 572: do you want to add a TODO?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -553,6 +567,13 @@ public static DynamoModel Start(IStartConfiguration configuration)
...
+                // This fact should probably be revisited in next major.
+                defaultPythonEngine = ((DefaultStartConfiguration)config).DefaultPythonEngine;
```
mjkkirschner, line 574: would it ever happen that we've loaded an old xml config and this will throw?

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -67,7 +68,13 @@ public int ScaleFactorLog
         private readonly DynamoModel model;
         private Point transformOrigin;
         private bool showStartPage = false;
-        
+
+        private ObservableCollection<SimpleViewModel<string>> defaultPythonEngineOptions = new ObservableCollection<SimpleViewModel<string>>();
+        public ObservableCollection<SimpleViewModel<string>> DefaultPythonEngineOptions
```
mjkkirschner, line 73: why the need for the field?

src/DynamoCoreWpf/ViewModels/Core/SimpleViewModel.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.ComponentModel;
+
+namespace Dynamo.ViewModels
+{
+    /// <summary>
+    /// Dead simple view model for when you only need to have a different display name and underlying value.
+    /// </summary>
+    /// <typeparam name="T">The type of your value</typeparam>
+    public class SimpleViewModel<T> : ViewModelBase, INotifyPropertyChanged
```
mjkkirschner, line 9: do you expect this to ever grow to contain more properties?

Maybe we can call it `NameValuePairViewModel`?

src/DynamoCoreWpf/ViewModels/ViewModelBase.cs
```diff
@@ -20,15 +20,6 @@ public bool IsDebugBuild
             }
         }
 
-        [JsonIgnore]
-        public bool IsPython2ObsoleteDebugModeEnabled
```
mjkkirschner, line 24: why is this removed?

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -29,15 +32,25 @@ public PythonCodeMigrationEventArgs(string oldCode, string newCode)
 
     public abstract class PythonNodeBase : VariableInputNode
     {
-        private PythonEngineVersion engine = PythonEngineVersion.IronPython2;
+        private PythonEngineVersion engine = PythonEngineVersion.Unspecified;
 
         [JsonConverter(typeof(StringEnumConverter))]
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Populate)]
+        [DefaultValue("IronPython2")]
```
mjkkirschner, line 39: can we use nameof?

src/Libraries/PythonNodeModels/PythonEngineSelector .cs
```diff
@@ -9,6 +9,7 @@ namespace PythonNodeModels
     /// </summary>
     public enum PythonEngineVersion
     {
+        Unspecified,
```
mjkkirschner, line 12: this should go at the end of the list as this change re-orders all the existing enums - ie:
Ironpython2 used to = `0`, but after this change it is `1`.

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -29,15 +32,25 @@ public PythonCodeMigrationEventArgs(string oldCode, string newCode)
...
         /// <summary>
         /// Return the user selected python engine enum.
         /// </summary>
         public PythonEngineVersion Engine
         {
-            get { return engine; }
+            get
+            {
+                // This is a first-time case for newly created nodes only
+                if (engine == PythonEngineVersion.Unspecified)
```
mjkkirschner, line 48: 👍 , but please see comment about enum order

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -48,6 +61,47 @@ public PythonEngineVersion Engine
             }
         }
 
+        private ObservableCollection<PythonEngineVersion> availableEngines;
+        /// <summary>
+        /// Available Python engines.
+        /// </summary>
+        public ObservableCollection<PythonEngineVersion> AvailableEngines
```
mjkkirschner, line 68: same question about auto properties

src/DynamoCore/Models/DynamoModel.cs
mmisol: Can use with private set. Will change

test/DynamoCoreWpfTests/PythonNodeCustomizationTests.cs
```diff
@@ -15,7 +15,7 @@ namespace DynamoCoreWpfTests
 {
     public class PythonNodeCustomizationTests : DynamoTestUIBase
     {
-        private readonly List<string> expectedEngineMenuItems = Enum.GetNames(typeof(PythonNodeModels.PythonEngineVersion)).ToList();
+        private readonly List<string> expectedEngineMenuItems = Enum.GetNames(typeof(PythonNodeModels.PythonEngineVersion)).Skip(1).ToList();
```
mjkkirschner, line 18: ? comment on this maybe?

src/DynamoCore/Models/DynamoModel.cs
mmisol: The next line is a `TODO` without a prefix, so I'll add it there.

src/DynamoCore/Models/DynamoModel.cs
mmisol: I don't know in which context this gets serialized/deserialized. XML serialization OOTB does not throw in cases like this but I don't know what you are doing exactly. Can you point to where that happens?

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
mmisol: I didn't want to initialize this more than once.

src/DynamoCoreWpf/ViewModels/Core/SimpleViewModel.cs
mmisol: Ok

src/DynamoCoreWpf/ViewModels/ViewModelBase.cs
mmisol: That's a good question. I saw it was not being used. You could say removing it is a breaking change, but it was added a couple of weeks ago, probably by mistake.

src/Libraries/PythonNodeModels/PythonNode.cs
mmisol: Yes

src/Libraries/PythonNodeModels/PythonEngineSelector .cs
mmisol: Technically this only breaks binary compatibility. I put this in the balance with following best practices, and opted for the latter. Let me know if you'd rather change this anyway.

src/Libraries/PythonNodeModels/PythonNode.cs
mmisol: This is for the purpose of initializing it only once.

test/DynamoCoreWpfTests/PythonNodeCustomizationTests.cs
mmisol: I think I'd rather change the code to make it more clear.

src/Libraries/PythonNodeModels/PythonEngineSelector .cs
mjkkirschner: my experience with this type of change is pretty complex confusion - code which was previously compiled against Dynamo 2.7 and selected engine ironPython2 will now be selecting unspecified.

I guess it's unlikely that any clients besides our own would ever be in this situation though.

If you've already thought about it then I'm fine with your decision. 👍

src/Libraries/PythonNodeModels/PythonNode.cs
mjkkirschner: I guess I'm still a bit confused by this answer - 
can't you do `Public ObserverableCollection<PythonEngineVersion> {get; set;} = ObserverableCollection<PythonEngineVersion>(){ yada yada}` ?

mmisol: Yes, you are right. I guess I don't really like to initialize in that way when it takes more than one line. If it was just creating the collection I wouldn't mind.

mmisol: Oh, I see you put `{ yada yada}` in same line. I don't really like that either 😄

src/DynamoCore/Models/DynamoModel.cs
QilongTang: Thanks for adding it @mmisol , can you mark it specially for 3.0? A bit easier to track all the 3.0 issues in code base

mmisol: Will do 👍

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -583,6 +604,11 @@ public static DynamoViewModel Start(StartConfiguration startConfiguration = new
 
         protected DynamoViewModel(StartConfiguration startConfiguration)
         {
+            // These options are currently fixed. Eventually we should make them dynamic I guess.
+            defaultPythonEngineOptions.Add(new NameValuePairViewModel<string>(string.Empty, WpfResources.DefaultPythonEngineNone));
+            defaultPythonEngineOptions.Add(new NameValuePairViewModel<string>("IronPython2", "IronPython2"));
+            defaultPythonEngineOptions.Add(new NameValuePairViewModel<string>("CPython3", "CPython3"));
```
QilongTang, line 610: Sry to get to it late, this is the reason why I mentioned earlier if we want to just use the enums we already exposing as `PythonEngineVersion`, we just need to add a default entry

QilongTang: I guess what I mean is instead of defining a fix collection but loop through the enums and construct it to reduce defined strings. Not sure if you have thought about it already

mmisol: The thing is that while `IronPython2` and `CPython3` are engine names and can go as they are, `Unspecified` is the absence of an engine and should be translated. Probably in different ways according to the context.
Also, not sure how permanent our enum is going to live forever. If we are going to open up the possibility for new engines, we'll need a solution based on names/strings.

mmisol:
@QilongTang If you are ok I'll go ahead and merge this.

QilongTang:
@mmisol I am OK to merge. Since this piece of work will eventually go into D4R and Civil3D. I would like us do some testing and keep it in mind that we may revisit some details here. I like the fact we are using preference setting, just seems more manageable to me. Good work!

src/DynamoCoreWpf/Views/Core/DynamoView.xaml
QilongTang: @mmisol Looking back on this, what limitations are in this case? Is it because we want to be dynamic on this and not achievable using child menu items?

PR #11010, opened 2020-08-17
Tab with spaces when using Python 3

mmisol:
### Purpose

The Python text editor now inserts spaces when the tab key is pressed
if the engine is not 'IronPython2'. This is done in order to remain
consistent with the reindent that occurs in 2to3 migration, which will
avoid errors from Python 3 complaining about mixing spaces and tabs.

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
+++ b/src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
@@ -207,6 +207,8 @@ namespace PythonNodeModelsWpf
             {
                 UpdateScript(editText.Text);
             }
+
+            editText.Options.ConvertTabsToSpaces = nodeModel.Engine != PythonEngineVersion.IronPython2;
         }
 
         private void OnScriptEditorWindowClosed(object sender, EventArgs e)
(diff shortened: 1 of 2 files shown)


src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
```diff
@@ -207,6 +207,8 @@ private void OnEngineChanged(object sender, System.Windows.Controls.SelectionCha
             {
                 UpdateScript(editText.Text);
             }
+
+            editText.Options.ConvertTabsToSpaces = nodeModel.Engine != PythonEngineVersion.IronPython2;
```
zeusongit, line 211: What happens when a PythonScript node is saved with engine selected as CPython3 is opened for editing by the user? In that case the `OnEngineChanged` event won't be triggered, will this option still persist?

QilongTang: @mmisol In the case above this should still work from my reading right?

mmisol: I don't know if I understand what you mean. You are saying what happens on a newly open node? This event will also be fired when the editor is open, even if you don't change the drop down. Did that answer your question?

PR #4276, opened 2015-04-19
Add task merging for graph updates

lukechurch:
@Benglin @ikeough Any reason to not do graph update task merging?

wrt: http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-7078

Opinions solicited.

Issue #4229, opened 2015-04-12
Slider Breaking/Lagging in "Automatic" mode in 0.8.1 


archinate:
Hello all,
I am using a slider in "Automatic" mode in the latest 0.8 official release.  My geometry will initially update with the slider.  However, after a few iterations the slider will either fail to update any geometry OR there will be a significant delay.

I tried this on something as simple as controlling a point position.

Note the image below.... "Automatic" mode is active, the X coordinate is set to 0, yet my point remains in its previous position.
![image](https://cloud.githubusercontent.com/assets/1890663/7107891/b77cd8ba-e139-11e4-8a1b-aa6fa50ab86c.png)

rudderdon:
Slowest animation ever

jnealb:
@archinate @rudderdon What are your slider Max and Step values? If mine are 10 and 1 or 0.5 etc then with 0.8.0.950 there is no issue as I drag the slider with Run Automatically: the point tracks in real-time. Perhaps we could set up a session to debug this?

archinate:
Hey @jnealb .  I should have been more clear.  I am using the 0.8.1 official release (0.8.1.955).

All of my slider values are default in the above scenario (0-100, Step 0.1)  The slider will begin updating smoothly, but after moving the slider enough, the definition will start to delay and eventually stop working.

tatlin:
I saw the lack of red point updating on some machines today in a workshop in London. 0.8 only, older versions worked fine

-matt

jnealb:
@archinate @rudderdon @tatlin @ikeough I did reproduce the delay with 8.1 0.8.1.1025, also the preview bubble on Point.ByCoordinates is not updating its X coordinate. When the slow down occurs, which is not easily reproduced for me, I find that waiting a bit then clears the condition.

ikeough:
Andy noticed this when he was working on Firefly. It has to do with our task compaction mechanism. Currently, whenever the slider ticks, an execution is fired. These execution "tasks" are then scheduled to run and they are processed in the order they are received. Execution tasks, unlike rendering tasks, do not get compacted by the scheduler. That is, if sliding a slider quickly submits tasks A...B...C, the scheduler doesn't know to only keep task C, the result of which is what you are actually looking for. 

[...]

PR #11021, opened 2020-08-18
Fix crashes by undoing/redoing nodes deletion

mmisol:
### Purpose

Fixes two crashes whie doing this at a high pace:

One crash happened in HomeWorkspaceViewModel, when postprocessing the
nodes after execution. These changes were done from the scheduler
thread, which when timed just right conflicted with changes done from
the UI thread, by the undo and redo operations.

Another one happened in ChangeSetComputer, when trying to compute the
delta AST. The cause was an optimization that attempted to skip tasks
based on containment comparison but disrespecting causality. The path
taken was to prevent skipping tasks altogether, but some more elaborate
optimization can probably be put in place.

### Declarations

### Reviewers

### FYIs

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCore/Scheduler/UpdateGraphAsyncTask.cs
+++ b/src/DynamoCore/Scheduler/UpdateGraphAsyncTask.cs
@@ -152,34 +152,6 @@ namespace Dynamo.Scheduler
             }
         }
 
-        /// <summary>
-        /// Returns true if this task's graph sync data is a super set of the other
-        /// </summary>
-        /// <param name="other"></param>
-        /// <returns></returns>
-        private bool Contains(UpdateGraphAsyncTask other)
-        {
-            // Be more conservative here. Not only check ModifiedNodes, but
-            // also check *all* nodes in graph sync data. For example, consider
-            // a CBN outputs to a node, the node is a downstream node of cbn.
-            //
-            // Now if node is modified and request a run, task's ModifiedNodes
-            // will cotain this node; then CBN is modified and request a run,
-            // ModifiedNodes would contain both CBN and the node, and previous
-            // task will be thrown away.
-            if (!other.ModifiedNodes.All(ModifiedNodes.Contains))
-                return false;
-
-            if (graphSyncData == null)
-                return other.graphSyncData == null;
-            else if (other.graphSyncData == null)
-                return true;
-
-            return other.graphSyncData.AddedNodeIDs.All(graphSyncData.AddedNodeIDs.Contains) &&
-                   other.graphSyncData.ModifiedNodeIDs.All(graphSyncData.ModifiedNodeIDs.Contains) &&
-                   other.graphSyncData.DeletedNodeIDs.All(graphSyncData.DeletedNodeIDs.Contains);
-        }
-
         protected override TaskMergeInstruction CanMergeWithCore(AsyncTask otherTask)
         {
             // One cannot just simply merge these tasks on a compare one to another basis.
--- a/src/DynamoCore/Scheduler/UpdateGraphAsyncTask.cs
+++ b/src/DynamoCore/Scheduler/UpdateGraphAsyncTask.cs
@@ -180,23 +180,15 @@ namespace Dynamo.Scheduler
                    other.graphSyncData.DeletedNodeIDs.All(graphSyncData.DeletedNodeIDs.Contains);
         }
 
-        private bool IsScheduledAfter(UpdateGraphAsyncTask other)
-        {
-            return CreationTime > other.CreationTime;
-        }
-
         protected override TaskMergeInstruction CanMergeWithCore(AsyncTask otherTask)
         {
-            var theOtherTask = otherTask as UpdateGraphAsyncTask;
-            if (theOtherTask == null)
-                return base.CanMergeWithCore(otherTask);
-
-            if (theOtherTask.IsScheduledAfter(this) && theOtherTask.Contains(this))
-                return TaskMergeInstruction.KeepOther;
-            else if (this.IsScheduledAfter(theOtherTask) && this.Contains(theOtherTask))
-                return TaskMergeInstruction.KeepThis;
-            else
-                return TaskMergeInstruction.KeepBoth;
+            // One cannot just simply merge these tasks on a compare one to another basis.
+            // The reason is that the changeSetComputer internal state might get corrupted.
+            // Imagine a sequence of queued tasks with changes like these: +A | -A | +AB
+            // Imagine a new task with change -A, ending up with the following sequence: -A | +A | -A | +AB
+            // While the merge routine would simplify this sequence to: -A | +A | +AB
+            // That creates an inconsistent sequence of changes where subtree A is added when it already exists!
+            return TaskMergeInstruction.KeepBoth;
         }
 
         #endregion
--- a/src/DynamoCore/Scheduler/UpdateGraphAsyncTask.cs
+++ b/src/DynamoCore/Scheduler/UpdateGraphAsyncTask.cs
@@ -152,15 +152,61 @@ namespace Dynamo.Scheduler
             }
         }
 
-        protected override TaskMergeInstruction CanMergeWithCore(AsyncTask otherTask)
+        /// <summary>
+        /// Returns true if this task's graph sync data is a super set of the other
+        /// </summary>
+        /// <param name="other"></param>
+        /// <returns></returns>
+        private bool CanReplace(UpdateGraphAsyncTask other)
         {
-            // One cannot just simply merge these tasks on a compare one to another basis.
-            // The reason is that the changeSetComputer internal state might get corrupted.
+            // Be more conservative here. Not only check ModifiedNodes, but
+            // also check *all* nodes in graph sync data. For example, consider
+            // a CBN outputs to a node, the node is a downstream node of cbn.
+            //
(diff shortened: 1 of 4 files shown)


src/DynamoCore/Scheduler/UpdateGraphAsyncTask.cs
```diff
@@ -180,23 +180,15 @@ private bool Contains(UpdateGraphAsyncTask other)
...
         }
 
-        private bool IsScheduledAfter(UpdateGraphAsyncTask other)
-        {
-            return CreationTime > other.CreationTime;
-        }
-
         protected override TaskMergeInstruction CanMergeWithCore(AsyncTask otherTask)
         {
-            var theOtherTask = otherTask as UpdateGraphAsyncTask;
```
mjkkirschner, line 190: I think this change may have significant performance impacts - we should investigate the original reason for the implementation - my memory is that when moving a slider it kicked off hundreds of unnecessary runs and made dynamo unresponsive - if this is related then theres an argument that the crash is actually acceptable...?

mmisol: Strange that was fixed by doing this instead of throttling the requests queued by the slider 🤔 

Still, maybe we could restore a more elaborate merge strategy that accounted for this case in particular. What kind of `graphSyncData` would we expect? Do we have a test case?

mjkkirschner: heres part of the discussion I recall: 
https://github.com/DynamoDS/Dynamo/pull/4276

mjkkirschner: and a related github issue:
https://github.com/DynamoDS/Dynamo/issues/4229

src/DynamoCore/Scheduler/UpdateGraphAsyncTask.cs
```diff
@@ -180,23 +180,15 @@ private bool Contains(UpdateGraphAsyncTask other)
...
-
-            if (theOtherTask.IsScheduledAfter(this) && theOtherTask.Contains(this))
-                return TaskMergeInstruction.KeepOther;
-            else if (this.IsScheduledAfter(theOtherTask) && this.Contains(theOtherTask))
-                return TaskMergeInstruction.KeepThis;
-            else
-                return TaskMergeInstruction.KeepBoth;
+            // One cannot just simply merge these tasks on a compare one to another basis.
+            // The reason is that the changeSetComputer internal state might get corrupted.
+            // Imagine a sequence of queued tasks with changes like these: +A | -A | +AB
```
aparajit-pratap, line 187: Is +A the front of the queue here or the back?

mmisol: Back <= Front

src/DynamoCore/Scheduler/UpdateGraphAsyncTask.cs
mmisol: `Delay` seems to work just fine according to my manual testing. Is there any special validation you would like to make beyond that issue?

src/DynamoCoreWpf/ViewModels/Core/HomeWorkspaceViewModel.cs
```diff
@@ -91,10 +92,27 @@ public override void OnNodeModified(NodeModel obj)
...
+                UpdateNodesDeltaState(e.NodeGuidList, e.GraphExecuted);
+            }
+        }
+
+        private void UpdateNodesDeltaState(List<Guid> nodeGuids, bool graphExecuted)
+        {
             // if runsettings is manual, and if the graph is not executed, then turing on showrunpreview 
             //should turn on showexectionpreview on every node.
-            if (nodeGuids.Count == 0 && !e.GraphExecuted)
+            if (nodeGuids.Count == 0 && graphExecuted)
```
aparajit-pratap, line 115: Are you sure this needs to be negated here? Was it a bug?

mmisol: Oops. That's my mistake actually. Didn't mean to change that. Good catch!

src/DynamoCore/Scheduler/UpdateGraphAsyncTask.cs
aparajit-pratap: I'm concerned if there are other scenarios (other than the slider) where we are not dropping redundant tasks that are quickly accumulating as a result of this change thus hitting performance.

mjkkirschner: We don't control the UI that other package authors may build - delay for the slider is great, but it does not cover the case generally, though I guess we could document the idea somewhere to throttle their controls?

For example I am thinking of controls like:
image

and

image

mmisol: Granted. For those cases having a strategy to merge/discard would be useful. I'll try to come up with something.

mmisol:
Manually tested the slider without the `Delay` in the binding, to try to replicate what a faulty slider would do. The new code prevents that issue, while it also doesn't cause the crashes related to undoing/redoing. Please give it another look @mjkkirschner

aparajit-pratap:
Looks good to me.

src/DynamoCoreWpf/ViewModels/Core/HomeWorkspaceViewModel.cs
```diff
@@ -91,10 +92,27 @@ public override void OnNodeModified(NodeModel obj)
...
+            {
+                DynamoViewModel.UIDispatcher.Invoke(() => UpdateNodesDeltaState(e.NodeGuidList, e.GraphExecuted));
+            }
+            else
+            {
+                // However, when Dynamo is shutting down, a funny deadlock situation occurs.
+                // The main thread is waiting for the scheduler to finish executing.
+                // So in order to avoid it, we do the update on the scheduler thread.
+                // Concurrency should not be an issue now that the UI is blocked waiting for shutdown.
+                UpdateNodesDeltaState(e.NodeGuidList, e.GraphExecuted);
```
mjkkirschner, line 107: I'm curious about this because I thought `Nodes` had a lock around it? This is not sufficient?

mjkkirschner:
@mmisol I really like this explanation - thanks, and solution looks good.
* I'm curious about the change to dispatch to the UI thread though - can you forsee any issues if the scheduler thread is the UI thread? (like in Revit?)

* Did you mean to remove the delay, or keep it just to be safe?

src/DynamoCoreWpf/ViewModels/Core/HomeWorkspaceViewModel.cs
mmisol: I didn't see a lock around here. Maybe you are thinking about the Model?

mjkkirschner: yes, thats what I was thinking of. ✔️

mmisol:
@mjkkirschner I can give it a try in Revit to be sure.

I kept the `Delay` but maybe I should remove it. Transitions look "smoother" when it's not there.

mmisol:
@mjkkirschner Worked fine in Revit. I'll wait for https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/1315/ to make sure this is not causing anything weird in the CI.

In the end I removed the delay on the slider to keep it smooth like before.

mmisol:
In the end I had to use `lock` instead of dispatching to the UI. Too many corner cases. If you have a minute please take another look @mjkkirschner 

The last build I ran is https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/1319/

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
```diff
@@ -685,9 +688,13 @@ private void unsubscribeNodeEvents(NodeViewModel nodeViewModel)
 
         void Model_NodeRemoved(NodeModel node)
         {
-            NodeViewModel nodeViewModel = Nodes.First(x => x.NodeLogic == node);
-            Errors.Remove(nodeViewModel.ErrorBubble);
-            Nodes.Remove(nodeViewModel);
+            NodeViewModel nodeViewModel;
+            lock (Nodes)
+            {
+                nodeViewModel = Nodes.First(x => x.NodeLogic == node);
```
mjkkirschner, line 694: I guess this is safest, but is there any harm in scanning the collection outside the lock?

mjkkirschner:
Thanks @mmisol - though this makes me wonder how rigorous we should be in applying locks other mutual exclusion mechanisms in Dynamo - especially in the view - in the Revit context this rarely matters, but now that more integrations are using multiple threads this will become more necessary.

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
mmisol: The thing is you can run into errors by enumerating over it while it's being modified. That's the initial crash I found.

PR #11023, opened 2020-08-19
missed some property change notifications to update view

mjkkirschner:
### Purpose

During testing @smangarole noted that the UI was not being updated for `showLabels` or `IsInput` or `IsOutput` checkboxes when undoing - I found that the tests are passing because the state is actually being correctly undone - but the UI was not being updated.

Two properties were missing and one was misspelled.

### Declarations


QilongTang:
@mjkkirschner I guess the Python engine option is directly binded so it is handled correctly already on view level?

mjkkirschner:
The PythonNode works because it's bound directly to the property on the NodeModel instead of being bound to a proxy property on the ViewModel.

mjkkirschner:
@QilongTang did the tests pass here?

QilongTang:
@mjkkirschner The self CI did not get triggered. I only saw the SonarQube scan which I think is not necessary, did I miss something?

PR #11028, opened 2020-08-20
Initial set of Node Auto Complete View Extension Files

QilongTang:
Please Note:

### Purpose

TODO:
- [ ] ~~Add debug mode and loading logic~~

Added manifest file support to control if particular view extension is set to be loaded or not. This came from a suggestion from @Dewb and I've added in this PR for team to discuss. The good part is I can see it backward compatible the way I implemented it and also useful for future view extensions dev work. In theory, it builds on top of the existing config and do not pollute extension loading logic by doing special GUID or debug mode check. In the extension area, it is a bit more straight forward than debug.config, but when user have both debug mode and extension to disable, they may get questions? Some thing for us to all think about. 

You can see the new view extension by default is not loaded for now until user enable that. 
image

### Declarations

### Reviewers

### FYIs


src/NodeAutoCompleteViewExtension/NodeAutoCompleteViewExtension.cs
```diff
@@ -0,0 +1,74 @@
...
+using Dynamo.Logging;
+using Dynamo.Wpf.Extensions;
+
+namespace Dynamo.NodeAutoComplete
+{
+    /// <summary>
+    /// This sample view extension demonstrates a sample IViewExtension 
+    /// which tracks current clicked node port in Dynamo and 
+    /// try to suggest the next best node to connect that port.
+    /// </summary>
```
mmisol, line 12: Please replace this with a proper description of the extension.

QilongTang: hmm can you be more specific? Any word here not proper? I think this is what we doing for node autocomplete no?

mjkkirschner: hmm, I wouldn't call this extension a sample?

mmisol: Sorry @QilongTang , I guess I read only the first line 😄

Definitely not a sample and not demonstrating anything. The rest is probably OK.

QilongTang: Gotcha, will remove sample

QilongTang:
@mmisol See todo in PR description. I think we want that but I have a question to the team, the DebugModes class currently serve as an internal singleton in DynamoCore. In order to expose that, we will need to make it public or make DynamoCore internal visible to this extension, which on you prefer? @DynamoDS/dynamo

mmisol:
Thanks @QilongTang . Sorry, I missed that part. I guess we can just use `InternalsVisibleTo` for this extension. There are no plans to make the extension independent and we probably don't want to rush exposing `DebugModes`. My two cents.

src/DynamoCoreWpf/Extensions/ViewExtensionLoader.cs
```diff
@@ -60,6 +65,11 @@ public IViewExtension Load(string extensionPath)
                 {
                     definition.TypeName = item.InnerText;
                 }
+                else if (item.Name == "IsEnabled")
+                {
+                    // Because the default value for this property is true so we check if user specifically put it false
+                    definition.IsEnabled = !item.InnerText.Equals("false") && !item.InnerText.Equals("False");
```
mmisol, line 71: `bool.TryParse` could be used instead and only assign `definition.IsEnabled` if a boolean was actually parsed.

src/NodeAutoCompleteViewExtension/NodeAutoCompleteViewExtension.cs
```diff
@@ -0,0 +1,73 @@
+﻿using System;
+using Dynamo.Extensions;
+using Dynamo.Logging;
+using Dynamo.Wpf.Extensions;
+
+namespace Dynamo.NodeAutoComplete
+{
+    /// <summary>
+    /// This view extension tracks current clicked node port in Dynamo and 
+    /// try to suggest the next best node to connect that port.
```
mmisol, line 10: Syntax: `try` => `tries`

src/NodeAutoCompleteViewExtension/NodeAutoCompleteViewExtension.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        public void Ready(ReadyParams readyParams)
+        {
+            // Do nothing for now
+        }
+
+        public void Shutdown()
+        {
+            this.Dispose();
```
mmisol, line 53: While we have been doing this, the life cycle of an extension will make it so that `Shutdown` is called and then `Dispose` is called as well. That results in `Dispose` being called two times, which can lead to exceptions. Not that it's causing issues now, because this extension is empty, but may lead to issues in the future so I guess we should avoid it.

src/DynamoCoreWpf/Extensions/ViewExtensionLoader.cs
```diff
@@ -19,10 +20,14 @@ private IViewExtension Load(ViewExtensionDefinition viewExtension)
                     CertificateVerification.CheckAssemblyForValidCertificate(viewExtension.AssemblyPath);
                 }
 
-                var assembly = Assembly.LoadFrom(viewExtension.AssemblyPath);
-                var result = assembly.CreateInstance(viewExtension.TypeName) as IViewExtension;
-                ExtensionLoading?.Invoke(result);
-                return result;
+                if (viewExtension.IsEnabled)
```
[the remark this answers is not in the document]
QilongTang: Will do

src/DynamoCoreWpf/Extensions/ViewExtensionLoader.cs
QilongTang: Will do

src/NodeAutoCompleteViewExtension/NodeAutoCompleteViewExtension.cs
QilongTang: Hi @mmisol Can you point me where indicates Dispose() is going to be called twice?

mmisol: Took some time to dig it but found it. It's in `DynamoView.WindowClosed`. In that method you'll see `ext.Shutdown();` being called for each extension. Then `this.Dispose();` calls `viewExtensionManager.Dispose();` which will call `Dispose` on each extension.

QilongTang: Ah, I see this is introduced as part of https://github.com/DynamoDS/Dynamo/pull/10672. Seems @mjkkirschner was doing it to make sure shutdown() and dispose() on extension level always get called. We have multiple ways to handle this then, we can either revert part of the Dispose() code from view extension manager or remove all instances  Dispose() calls on view extension shutdown.. Will ask team input tomorrow at standup

QilongTang:
Thanks @mmisol for the comments, it is current covered in test `PackageManagerLoadsAndAddsViewExtension`. This view extension will not be loaded by DynamoSandbox or tests as a verification.

mmisol:
@QilongTang While what you say is true, we will want to enable this extension soon. We can't depend this on staying disabled. It will also make this lack of test harder to track in the future, as people will simply add the extension to the list never realizing it was part of a test.

QilongTang:
@mmisol Good point! Two more tests added

QilongTang:
@mmisol I will make another PR with the changes to get rid of calling Dispose() in Shutdown() for our view extensions. Merging this one in.

PR #11031, opened 2020-08-20
Avoid Helix crash due to edge double values

mmisol:
### Purpose

This prevents the display of geometry located in positions considered
invalid by Helix. This includes the special values PositiveInfinity,
NegativeInfinity and NaN.

Also NaN is checked as a special value when unmarhsalling data from an
FFI. Special values PositiveInfinity and NegativeInfinity were already
handled. This allows to generate a warning over the node with an
appropriate message.

Here is a screenshot of how the message looks:

NaN

### Declarations

### Reviewers

### FYIs

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
@@ -1607,6 +1607,8 @@ namespace Dynamo.Wpf.ViewModels.Watch3D
         /// <param name="packages">An <see cref="IEnumerable"/> of <see cref="HelixRenderPackage"/>.</param>
         internal virtual void AggregateRenderPackages(IEnumerable<HelixRenderPackage> packages)
         {
+            packages = FilterOutInvalidPackages(packages);
+
             IEnumerable<string> customNodeIdents = null;
             if (InCustomNode())
             {
@@ -1807,6 +1809,29 @@ namespace Dynamo.Wpf.ViewModels.Watch3D
                 }
 
             }
+        }
+
+        /// <summary>
+        /// Filters out packages that are considered invalid by Helix. This includes any package
+        /// that has a coordinate with a special value like NaN or Infinite.
+        /// </summary>
+        /// <param name="packages">Original packages to render</param>
+        /// <returns>List of packages considered valid</returns>
+        private IEnumerable<HelixRenderPackage> FilterOutInvalidPackages(IEnumerable<HelixRenderPackage> packages)
+        {
+            List<HelixRenderPackage> result = new List<HelixRenderPackage>();
+
+            foreach (var package in packages)
+            {
+                if (!package.Points.Positions.Any(v => v.IsInvalid()) &&
+                    !package.Lines.Positions.Any(v => v.IsInvalid()) &&
+                    !package.Mesh.Positions.Any(v => v.IsInvalid()))
+                {
+                    result.Add(package);
+                }
+            }
+
+            return result;
         }
 
         private void AddLabelPlace(string nodeId, Vector3 pos, IRenderPackage rp)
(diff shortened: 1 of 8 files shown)


src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -1809,6 +1811,29 @@ internal virtual void AggregateRenderPackages(IEnumerable<HelixRenderPackage> pa
...
+        private IEnumerable<HelixRenderPackage> FilterOutInvalidPackages(IEnumerable<HelixRenderPackage> packages)
+        {
+            List<HelixRenderPackage> result = new List<HelixRenderPackage>();
+
+            foreach (var package in packages)
+            {
+                if (!package.Points.Positions.Any(v => v.IsInvalid()) &&
+                    !package.Lines.Positions.Any(v => v.IsInvalid()) &&
+                    !package.Mesh.Positions.Any(v => v.IsInvalid()))
+                {
```
aparajit-pratap, line 1829: I wonder if this will cause a significant slowdown.

src/VisualizationTests/CustomGraphicItemVisualizationTest.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            OpenVisualizationTest("NaN.dyn");
+
+            // Only the default scene items should be present
+            Assert.AreEqual(3, BackgroundPreviewGeometry.Count());
+
+            // Nodes should be set to warning status with an appropriate message
+            var nanNode = Model.CurrentWorkspace.Nodes.First(n => n.Name == "NaN");
+            Assert.AreEqual(ElementState.Warning, nanNode.State);
+            StringAssert.Contains("'NaN' is being cast to 'Double'", nanNode.ToolTipText);
```
aparajit-pratap, line 28: @mmisol are these warning messages coming from .net or are we creating them?

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
mjkkirschner: it sure seems like it could - for an alternative, what about checking this during package creation?

mmisol: The checks should be quick. The amount will grow with geometry, but should be a very low percentage in comparison.

@mjkkirschner What do you mean by package creation? To check it as soon as the geometry is added to the package?

src/VisualizationTests/CustomGraphicItemVisualizationTest.cs
mmisol: They come from us, from `CLRObjectMarshaller`.

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
mjkkirschner: yes, exactly, since we already need to iterate over each vertex at that point. It may not be possible though if some render packages are created in LibG code.

mjkkirschner: `GetRenderPackagesFromMirrorData` is the function I am thinking of.

mjkkirschner: But - it's probably better to profile this than ask you to change your implementation without reason.

mmisol: Sounds good 👍

PR #11032, opened 2020-08-20
Fix multiple engine in Python editor bug

mmisol:
### Purpose

The source of the combo was set to a non-static property. This caused
a really strange behavior when multiple Python nodes were present in
the graph. In that case the engines would appear repeated multiple
times.

Changing the property to static fixed the issue and it actually makes
more sense.

### Declarations

### Reviewers

@mjkkirschner @QilongTang 

### FYIs

@DynamoDS/dynamo

### DIFF
--- a/src/Libraries/PythonNodeModels/PythonNode.cs
+++ b/src/Libraries/PythonNodeModels/PythonNode.cs
@@ -61,11 +61,11 @@ namespace PythonNodeModels
             }
         }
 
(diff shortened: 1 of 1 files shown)


mmisol:
This is my bad. Sorry about that.

mjkkirschner:
This should be cherry picked to 2.8 right?

mmisol:
Indeed. I'll do it right after I merge this.

PR #11037, opened 2020-08-21
fix isolate selected regression, and longstanding issue with selection highlight.

mjkkirschner:
### Purpose

In 2.7 there was a regression with isolateSelection, but I also found there was a long standing issue with overlapping geometry in some cases not highlighting correctly even during normal selection.

this pr does the following:
* inside the dynamo mesh vertex shader - selected geometry is moved along the normal of each vertex by `.01` - this is large enough to prevent z-fighting even when zoomed outed until the geometry is no longer visible.
    * geometry which requires vertex colors maintains its original movement of `.0001`
    * if a vertex is both vertex colored and selected it moves by `.0101`

* visualization tests are added for two spheres in exactly the same location in isolation mode and regular selection mode.

Screen Shot 2020-08-21 at 4 22 49 PM

The reason that we need an additional increment for selected geometry is because I found most mesh geometry we generate has white vertex colors so all verts are moved slightly... essentially nothing special was being done for selected geometry. These colors were applied in LibG many years ago, and I don't think it is worth the risk to modify that at this point.

For those who are curious I found that the reason this worked in 2.6 for isolated geometry, which was that after every interaction we would resort all geometry in the scene so that the selected geometry was rendered last... it's great to be able to skip this given that this sorting could be slow if there was a lot of geometry.

### Declarations


QilongTang:
@DynamoDS/dynamo Team decided to merge for Dynamo 2.8 and cherry-pick for now.

PR #11040, opened 2020-08-24
Translate DS array to NET array for description

mmisol:
### Purpose

This fixes a bug where the description of Curve.SplitByParameter in the
library would show up a description of an obsoleted overload.

The cause of the issue was the inability of XmlDocumentationExtensions
to handle conversion from primitve DS arrays to their equivalent .NET
arrays. This made the module unable to find the exact overload, so it
resorted to a fallback which returns the first overload available,
which in this case was the incorrect one.

The problem is fixed for this particular instance, as it is now able to
find the exact overload. Note, however, that the mechanism is not
bullet-proof, as there is no guaranteed way to translate back a type
from DS to the original .NET type.

### Declarations

### Reviewers

@aparajit-pratap @QilongTang 

### FYIs

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCore/Library/XmlDocumentationExtensions.cs
+++ b/src/DynamoCore/Library/XmlDocumentationExtensions.cs
@@ -154,6 +154,8 @@ namespace Dynamo.Engine
                 documentNode = documentNodes[fullyQualifiedName];
             else
             {
+                // Note that the following may take the incorrect overload.
+                // Unfortunately we can't map back to the exact .NET parameter types from the DS function descriptor.
                 var overloadedName = documentNodes.Keys.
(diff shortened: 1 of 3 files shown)


src/DynamoCore/Library/XmlDocumentationExtensions.cs
```diff
@@ -211,12 +217,20 @@ private static string PrimitiveMap(string s)
                     return "System.Object";
                 case "double":
                     return "System.Double";
+                case "double[]":
+                    return "System.Double[]";
```
aparajit-pratap, line 221: @mmisol this could also be an IEnumerable or IList, etc., right? Is that why you say this is not foolproof?

mmisol: @aparajit-pratap Exactly

PR #11042, opened 2020-08-24
Fix layout issue hiding Python editor buttons

mmisol:
### Purpose

This fixes an issue that partially hid the "More info" button in the
Python editor window when the UI was shown in German language. In that
case "Run" would display as "Ausführen" which pushed the content to the
right causing the "More info" button to be trimmed because it fell out
of the boundaries of its container.

The issue is fixed by removing the parent grid, resulting in a simpler
and more flexible layout.

Here is a screenshot of the button bar after the fix:
Screen Shot 2020-08-24 at 11 42 50 AM

### Declarations

### Reviewers

@QilongTang @reddyashish 

### FYIs

@DynamoDS/dynamo


QilongTang:
Thanks @mmisol Please also cherrypick after merging

PR #11048, opened 2020-08-24
Add tooltip to preview disabled nodes

zeusongit:
### Purpose

[DYN-3074 ](https://jira.autodesk.com/browse/DYN-3074)
Add description, out-ports, in-ports tooltip to nodes which had their preview disabled.

previewtooltip

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCoreWpf/Views/Core/NodeView.xaml
+++ b/src/DynamoCoreWpf/Views/Core/NodeView.xaml
@@ -283,6 +283,7 @@
                 Background="#99B8CA"
                 Opacity=".4"
                 Style="{StaticResource SZoomFadePreview}"
+                ToolTipService.ShowDuration="60000"
                 MouseDown="NameBlock_OnMouseDown">
             <Border.Visibility>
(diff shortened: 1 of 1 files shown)


Amoursol:
This was the problem submitted in the Problem Report, but I agree @mmisol that rather than bespoke solutions we should simply enable all node actions when the **Preview State** is turned off with the exception of the **Geometry**.

aparajit-pratap:
I'm not quite sure I understood why we needed to add another UI element for the tooltip when we already have it. Like @mmisol says, it should be a matter of keeping the existing tooltip enabled even in the preview-off state.

zeusongit:
Well, the existing tooltip element was part of the rectangle, and disabling the preview adds a border on top of the node with its bg color set to light blue, and that did not have the tooltip element which I added. I don't know if there is a way to create a separate tooltip element and bind it to both the rectangle and border.

Amoursol:
We can investigate these kind of issues in our **Node Refresh** work that is currently being explored by UX :)

PR #11057, opened 2020-08-26
Make side panel handles size to content

mmisol:
### Purpose

The handles for size panels were using a fixed size, which was not
enough to fit some translated texts. The handles are now allowed to
size to content by doing some extra gimnastics to rotate them in the
right way.

Here is a screenshot of how it looks with the German translation:
extension_handle

### Declarations

### Reviewers

@QilongTang @zeusongit 

### FYIs

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCoreWpf/Views/Core/DynamoView.xaml
+++ b/src/DynamoCoreWpf/Views/Core/DynamoView.xaml
@@ -1516,7 +1516,7 @@
                     VerticalAlignment="Top"
                     Margin="0,310,0,0">
             <Grid Background="#353535"
-                  Width="100"
+                  Width="Auto"
                   Cursor="Hand">
                 <Button Click="OnCollapsedLeftSidebarClick"
                         Template="{DynamicResource BackgroundButton}">
@@ -1533,8 +1533,8 @@
                             </Border>
                         </ControlTemplate>
                     </Button.Resources>
-                    <Grid Mouse.MouseEnter="Button_MouseEnter"
-                          Mouse.MouseLeave="Button_MouseLeave">
+                    <Grid Mouse.MouseEnter="LibraryHandle_MouseEnter"
+                          Mouse.MouseLeave="LibraryHandle_MouseLeave">
                         <Grid.ColumnDefinitions>
                             <ColumnDefinition Width="*" />
                             <ColumnDefinition Width="Auto" />
@@ -1581,9 +1581,10 @@
                     Background="{StaticResource HighlightedBrush}"
                     HorizontalAlignment="Right"
                     VerticalAlignment="Top"
-                    Margin="0,310,-79,0">
+                    Margin="0,310,0,0"
+                    RenderTransformOrigin="1,0">
             <Grid Background="#353535"
-                  Width="100"
+                  Width="Auto"
                   Cursor="Hand">
                 <Button Click="OnCollapsedRightSidebarClick"
                         Template="{DynamicResource BackgroundButton}">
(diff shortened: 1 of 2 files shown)


src/DynamoCoreWpf/Views/Core/DynamoView.xaml
```diff
@@ -1625,17 +1618,29 @@
                                Margin="5,5,5,5"
                                RenderTransformOrigin="0.5, 0.5">
                                 <Image.RenderTransform>
-                                    <RotateTransform Angle="90" />
+                                    <RotateTransform Angle="-90" />
```
aparajit-pratap, line 1621: Curious what this is for?

aparajit-pratap:
Does the library handle look better as well with German characters?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml
mmisol: That's for rotating a UI element. This is related to the gymnastics I mentioned. Since the rotation of the panel changed directions, this one needs to change as well.

mmisol:
@aparajit-pratap Do you mean German characters not in English? I did not try, as the German translation did not have any for this case. Is there any particular text you would like to try? I can just write it literally in the control to try it

aparajit-pratap:
> @aparajit-pratap Do you mean German characters not in English? I did not try, as the German translation did not have any for this case. Is there any particular text you would like to try? I can just write it literally in the control to try it

Yes, I meant any language as long as it is a long string, just so we can manually check that the library handle resizes as expected.

mmisol:
Here is a sample showing a very long text. It eventually overlaps some of the canvas controls, but the handle adapts just fine. In any case, this is an hypothetical case, we don't have such long translations.

long_extension_text

QilongTang:
@mmisol Thanks taking a look now

QilongTang:
@mmisol Can you check the left side panel for me? Do we have similar issue there?
image

src/DynamoCoreWpf/Views/Core/DynamoView.xaml
```diff
@@ -1516,7 +1516,7 @@
                     VerticalAlignment="Top"
                     Margin="0,310,0,0">
             <Grid Background="#353535"
-                  Width="100"
+                  Width="Auto"
```
QilongTang, line 1519: Seems this is the only changed needed to fix the left side?

mmisol: Yes. No further gymnastics required on that side.

mmisol:
Here is a longer example for the library side @QilongTang 

long_library_text

src/DynamoCoreWpf/Views/Core/DynamoView.xaml
```diff
@@ -1600,22 +1601,14 @@
...
                         <StackPanel Orientation="Horizontal" 
                                     HorizontalAlignment="Center">
-                            <TextBlock Grid.Column="0"
-                                   Name="ExtensionSidebarText"
-                                   VerticalAlignment="Center"
-                                   Foreground="#aaaaaa"
-                                   FontWeight="Normal"
-                                   Margin="5,0,0,5"
-                                   Text="{x:Static p:Resources.ExtensionsViewTitle}">
-                            </TextBlock>
```
QilongTang, line 1618: Curious why did we only move this textblock for the right side but keep the change for left?

mmisol: Without getting into much detail, the main difference is that the centre of the rotation is different for each side, which in turn requires to do things differently in order to show the handles as we expect them.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -1780,25 +1780,40 @@ private void SlideWindowToIncludeTab(int tabSelected)
...
                 imageUri = new Uri(@"pack://application:,,,/DynamoCoreWpf;component/UI/Images/expand_hover.png");
                 BitmapImage hover = new BitmapImage(imageUri);
-                collapseIcon.Source = hover;
+                icon.Source = hover;
             }
         }
 
+        private void ExtensionHandle_MouseEnter(object sender, MouseEventArgs e)
+        {
+            Grid g = (Grid)sender;
```
QilongTang, line 1809: Should we check the type of `sender` here `is Grid` before doing all of the following?

mmisol: This part did not really change. I would say it is fine like this given it is an event handler, and as such should not be reused.

PR #11063, opened 2020-08-27
Fix custom node small icon load in library

### LINKED ISSUE #10669: Package icons do not show up in library
## Dynamo version

image

## Operating system

Windows 10

## What did you do? 

Added Package.Images.dll to provide icons for custom nodes that are *dyf based. 

## What did you expect to see?

Small/Large icons show up in the library.

## What did you see instead?

Only Large icons come through. 

image

Just FYI, this used to work before in older versions of Dynamo. I have a PackageImages.resx in the project and a post build routine that wraps it into a customization. Everything following your instructions from the WIKI pages. What's confusing is that clearly the customization DLL is doing some part of the job, because the Large icons show up properly. Small icons have the exact same name (guid.Small.png) and are embedded just the same way. They are 32px in size just like recommended. I don't see anything here. 

Ideas?

mmisol:
### Purpose

Icons for packages custom nodes (dyf files) were not loaded in the
library. This was due to a bug that created a wrong url, which used a
default name, instead of the custom node id.

Fixes #10669

### Declarations

### Reviewers

@reddyashish @aparajit-pratap 

### FYIs

@DynamoDS/dynamo


mmisol:
Before:
customnode_icon_repro

After:
customnode_icon_fixed

mmisol:
@aparajit-pratap Will check those tests.

For the example I'm showing here I followed those steps so I'd say it's accurate, only that the bug was preventing it from working.

I did not check ZT nodes because the reported issue was not related to them. I can also do that but should not be related to this PR. As you can see in the code this is a custom node specific error condition.

PR #11065, opened 2020-08-27
Autocompletion feature for CPython3 Engine. 

reddyashish:
### Purpose

This PR adds the autocompletion feature to the CPython3 engine. 

Task: https://jira.autodesk.com/browse/DYN-2781

. 

### Declarations

### Reviewers
@aparajit-pratap @mmisol @QilongTang 

### FYIs

@DynamoDS/dynamo


mmisol (changes requested):
Hi @reddyashish . I will abstain from doing a full review for now, but will request changes for some issues I noticed. Still, I must admit I don't entirely understand where this PR leaves us with regards to auto-completion in Python 3. From what I see this is largely based on the Iron Python implementation and I can see some glaring failing cases, like getting CLR suggestions for Python native primitives. I do see some code related to the CPython3 engine, but it seems commented out or incomplete. Let me know if these are not your latest changes.

In general, I feel that before proceeding with this PR we should think about:
1. What does this improve over using the Iron Python code completion?
2. What are the use cases that change given this is not Iron Python?
3. What are the use cases that change given this is Python 3 and not Python 2?

mmisol:
PS: This is a screenshot showing what I meant about `getting CLR suggestions for Python native primitives`:
python3_dotnet_functions

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -255,6 +256,7 @@ private static PyScope CreateGlobalScope()
             // Allows discoverability of modules by inspecting their attributes
             scope.Exec(@"
 import clr
+import sys
```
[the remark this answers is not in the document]
reddyashish: This was when I was testing something. Not needed. Removing it.

src/Libraries/DSCPython/DSCPythonCodeCompletionDataCore.cs
```diff
@@ -0,0 +1,55 @@
...
+                                                ExternalCodeCompletionType completionType, IExternalCodeCompletionProviderCore providerCore)
+        {
+            Text = text;
+            Stub = stub;
+            IsInstance = isInstance;
+            provider = providerCore;
+            CompletionType = completionType;
+        }
+
+        public string Text { get; private set; }
```
QilongTang, line 24: @reddyashish Do you make these properties public because they are made so in IronPythonCodeCompletionData before? Can we make them internal for now, and maybe add some comments for each of them?

src/Libraries/DSIronPython/IronPythonCodeCompletionProviderCore.cs
```diff
@@ -235,7 +235,7 @@ private Type TryGetTypeFromFullName(string name)
             return importMatches;
         }
 
-        private static Dictionary<string, string> FindVariableStatementWithRegex(string code, string valueRegex)
+        public static Dictionary<string, string> FindVariableStatementWithRegex(string code, string valueRegex)
```
QilongTang, line 238: internal?

reddyashish:
The self-serve was failing intermittently when the Python tests were all run together. Individually they were all passing but a couple of them were failing when the whole test suite was run. I made some fixes to them and verified it locally. Waiting for the self-serve to confirm.

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,1277 @@
...
+                    }
+
+                    importMatches.Add(typeName, wholeLine.Replace(joinedTypeNames, typeName));
+                }
+            }
+
+            return importMatches;
+        }
+
+        public static Dictionary<string, string> FindVariableStatementWithRegex(string code, string valueRegex)
```
aparajit-pratap, line 239: I don't think this needs to be public as it's being obtained by reflection using a non-public flag.

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,1277 @@
...
+        internal static readonly Regex MATCH_FROM_IMPORT_STATEMENTS = new Regex(@"from\s+?([\w.]+)\s+?import\s+?([\w, *]+)", RegexOptions.Compiled | RegexOptions.Multiline);
+        internal static readonly Regex MATCH_VARIABLE_ASSIGNMENTS = new Regex(@"^[ \t]*?(\w+(\s*?,\s*?\w+)*)\s*?=\s*(.+)", RegexOptions.Compiled | RegexOptions.Multiline);
+
+        internal static readonly Regex STRING_VARIABLE = new Regex("[\"']([^\"']*)[\"']", RegexOptions.Compiled);
+        internal static readonly Regex DOUBLE_VARIABLE = new Regex("^-?\\d+\\.\\d+", RegexOptions.Compiled);
+        internal static readonly Regex INT_VARIABLE = new Regex("^-?\\d+", RegexOptions.Compiled);
+        internal static readonly Regex LIST_VARIABLE = new Regex("\\[.*\\]", RegexOptions.Compiled);
+        internal static readonly Regex DICT_VARIABLE = new Regex("{.*}", RegexOptions.Compiled);
+
+        internal static readonly string BAD_ASSIGNEMNT_ENDS = ",([{";
```
aparajit-pratap, line 49: Do all of these regex variables need to be internal, can they be made private?

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,1277 @@
...
+                {
+                    continue;
+                }
+
+                importMatches.Add(libName, wholeLine);
+            }
+
+            return importMatches;
+        }
+
```
aparajit-pratap, line 647: I think the `#endregion` for the `#region backup legacy class` should come somewhere here and not line 904, correct?

src/Libraries/DSCPython/DSCPythonCodeCompletionDataCore.cs
```diff
@@ -0,0 +1,55 @@
+﻿using Autodesk.DesignScript.Interfaces;
+
+namespace DSCPython
+{
+    /// <summary>
+    /// Concrete type that gets returned and converted to an Avalonedit type implementing
+    /// ICompletionData when used from WPF ScriptEditorContorl.
```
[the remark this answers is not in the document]
reddyashish: Yeah, changed it to IExternalCodeCompletionData.

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
reddyashish: Changed it to internal

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,1277 @@
...
+{
+
+    internal class DSCPythonCodeCompletionProviderCore : IExternalCodeCompletionProviderCore, ILogSource
+    {
+        #region internal constants
+        internal static readonly string commaDelimitedVariableNamesRegex = @"(([0-9a-zA-Z_]+,?\s?)+)";
+        internal static readonly string variableName = @"([0-9a-zA-Z_]+(\.[a-zA-Z_0-9]+)*)";
+        internal static readonly string spacesOrNone = @"(\s*)";
+        internal static readonly string atLeastOneSpaceRegex = @"(\s+)";
+        internal static readonly string dictRegex = "({.*})";
```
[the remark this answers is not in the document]
reddyashish: Only a couple of them were duplicated. I removed the duplicate ones from SharedCompletionProvider as they were only being used inside this file.

src/Libraries/PythonNodeModelsWpf/SharedCompletionProvider.cs
```diff
@@ -81,11 +95,21 @@ internal static IExternalCodeCompletionProviderCore FindMatchingCodeCompletionCo
...
-                Select(x => new IronPythonCompletionData(x)).ToArray();
+            if (engineVersion.Equals(PythonEngineVersion.IronPython2))
+            {
+                return this.providerImplementation?.GetCompletionData(code, expand).
+                    Select(x => new IronPythonCompletionData(x)).ToArray();
+            }
+            else if (engineVersion.Equals(PythonEngineVersion.CPython3))
+            {
+                return this.providerImplementation?.GetCompletionData(code, expand).
+                    Select(x => new IronPythonCompletionData(x)).ToArray();
```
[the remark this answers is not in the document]
reddyashish: The providerImplementation will point to the appropriate completion provider core based on the Engine type. So that object would be different. Changed this to a single if statement.

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,1277 @@
...
+            {
+                var name = matches[i].Groups[1].Value.Trim();
+                var val = matches[i].Groups[6].Value.Trim();
+                paramMatches.Add(name, val);
+            }
+
+            return paramMatches;
+        }
+
+        private static List<Tuple<string, string, string>> FindAllImportStatements(string code)
```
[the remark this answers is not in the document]
reddyashish: This function in IronPythonCompletionProvider will be removed after we get rid of this class. We would need this method in DSCPythonCodeCompletionProviderCore.cs.

reddyashish:
@aparajit-pratap Addressed the comments. 
@mmisol I have made changes so that the .NET functions/fields are not shown for string/int type variables. The Python inbuilt functions for these types are not available from the PyObject so we might need another approach to get them. Let me know if you have any ideas around it.

mmisol:
Hi @reddyashish . In my opinion that would be a pretty big limitation. I can probably investigate the issue but the path could be long and/or incompatible with this PR, so the timing to do it might not be the best. Probably something to bring up to @QilongTang and @Amoursol and discuss tomorrow.

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,1270 @@
...
+        //Even though the following region is full of private implementation details, these
+        // back the old IronPythonCompletionProvider class
+        //!!!- do not modify these signatures until that class is removed.
+        //We do not know if anyone was using that class, but we needed to remove the compile
+        //time references between PythonNodeModels and DSIronPython to support dynamically loading
+        //python versions.
+
+        //note that the public members below are not really public (this is an internal class)
+        //but must be marked that way to satisfy the legacy interface
+        #region BACKING LEGACY CLASS DO NOT MODIFY UNTIL 3
```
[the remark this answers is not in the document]
mjkkirschner: I think I wrote this comment, so it seems this was copy pasted from the other file... yeah we should be able to get rid of all of this unless it's used here?

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,1266 @@
...
+            {
+                PythonEngine.ReleaseLock(gs);
+            }
+        }
+
+        #endregion
+
+        private PyScope CreateGlobalScope()
+        {
+            PyScopeManager.Global.TryGet(globalScopeName, out PyScope Scope);
```
mjkkirschner, line 1212: so this is kind of worrying - doesn't this use the global scope (that we use for transferring python objects between other scopes) as the scope for all type checking required for autocomplete. 

This seems like it might have consequences or be prone to memory leaks.

src/Libraries/DSCPython/DSCPython.csproj
```diff
@@ -47,6 +47,14 @@
     <Reference Include="Python.Runtime, Version=2.4.1.0, Culture=neutral, processorArchitecture=MSIL">
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\..\..\extern\Python\Python.Runtime.dll</HintPath>
+    </Reference>
+	<Reference Include="Microsoft.Dynamic, Version=1.2.2.0, Culture=neutral, PublicKeyToken=7f709c5b713576e1, processorArchitecture=MSIL">
```
mjkkirschner, line 51: this seems wrong - why does DSCPython now require types from the IronPython DLR project?

src/Libraries/DSCPython/DSCPython.csproj
```diff
@@ -47,6 +47,14 @@
...
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\packages\DynamicLanguageRuntime.1.2.2\lib\net45\Microsoft.Dynamic.dll</HintPath>
+    </Reference>
```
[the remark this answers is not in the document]
mjkkirschner: I'm not clear why this was marked resolved - so I'm marking it unresolved. 😉

src/Libraries/PythonNodeModelsWpf/SharedCompletionProvider.cs
```diff
@@ -8,12 +8,26 @@
 
 namespace Dynamo.Python
 {
-    internal class SharedCompletionProvider : LogSourceBase
+    public class SharedCompletionProvider : LogSourceBase
```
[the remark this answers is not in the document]
mjkkirschner: why does this need to be done in 3.0? Wasn't it internal before?
But why should this be public?

src/Libraries/PythonNodeModelsWpf/SharedCompletionProvider.cs
```diff
@@ -8,12 +8,25 @@
...
 {
-    internal class SharedCompletionProvider : LogSourceBase
+    // TODO 3.0: Change this file name and class to CompletionProviderAdaptor. 
+    public class SharedCompletionProvider : LogSourceBase
     {
 
         #region Properties and Fields
         private readonly IExternalCodeCompletionProviderCore providerImplementation;
 
+        public static string doubleQuoteStringRegex = "(\"[^\"]*\")"; // Replaced w/ quotesStringRegex - Remove in Dynamo 3.0
```
mjkkirschner, line 18: why are these public?

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,1266 @@
...
+
+                importMatches.Add(libName, wholeLine);
+            }
+
+            return importMatches;
+        }
+
+        #endregion
+
+        public Type TryGetType(string name)
```
mjkkirschner, line 650: I think it's possible this code makes some incorrect assumptions about the types that are retrievable from PythonNet... it's been a while but I think this code assumes that types are real .net types... which they won't be of course in pythonNet.

src/Libraries/PythonNodeModelsWpf/SharedCompletionProvider.cs
reddyashish: Made them internal

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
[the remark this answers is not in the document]
reddyashish: They are being used for now because we do not have a common class for both of the CompletionProviderCore classes. My thinking was if only one of the IronPython/CPython is present, we would still need these helper functions.

src/Libraries/PythonNodeModelsWpf/SharedCompletionProvider.cs
```diff
@@ -9,23 +9,23 @@
 namespace Dynamo.Python
 {
     // TODO 3.0: Change this file name and class to CompletionProviderAdaptor.
```
mjkkirschner, line 11: @reddyashish since this was already internal, you can just change the name now!

mjkkirschner:
@reddyashish 
- was the followup filed to address the python object instances not correctly having autocomplete?
- I kicked off the tests again to see if the failure is related to ASM changes from other prs, you might want to update this branch with latest master.
- you should be able to make your name change now to the shared complete provider.
- I don't understand your comment about not having a shared implementation - any reason why you are choosing to not do that work now?

mjkkirschner:
@reddyashish - do you think that the `SharedCompletionProvider` will only ever be used for python? Could we use it for a c# scripting window for example?

If theres nothing that makes it python specific, then lets leave python out of the name-  what do you think?

reddyashish:
@mjkkirschner Yes that makes more sense. I have changed the name to CompletionProviderAdaptor.

I have created the followup task for the Python object instances completion functionality: https://jira.autodesk.com/browse/DYN-3198

Regarding the shared implementation, We do not have a common class that extends both the IronPythonCompletionProviderCore and DSCPythonCompletionProviderCore to have the shared implementation. I tired moving those to a the CompletionProviderAdaptor but was running into different issues. The CompletionProviderAdaptor class is more like a pointer to which CompletionProviderCore should be chosen. They were like 4-5 functions that have same implementation and the team decided we will move them to a different shared location once we have all the functionality up and running. What do you think?

reddyashish:
The self-serve validation has passed now. @mjkkirschner Let me know if you like any more changes.

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,1270 @@
...
+
+        internal static readonly string BAD_ASSIGNEMNT_ENDS = ",([{";
+
+        /// <summary>
+        /// A list of short assembly names used with the TryGetTypeFromFullName method
+        /// </summary>
+        private static string[] knownAssemblies = {
+            "mscorlib",
+            "RevitAPI",
+            "RevitAPIUI",
```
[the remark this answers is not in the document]
mjkkirschner: @reddyashish did you get a chance to test this?

mjkkirschner (changes requested):
@reddyashish

 1. please address/investiate the concerns with memory leaks by not reusing the global scope, but creating an autocomplete scope for each script editor.
I imagine the scope should be disposed when the editor window is closed, and recreated when opened so that python code executed in that context is not around for the application lifetime.

2. Please the conflicts on the scriptWindow files.

reddyashish:
@mjkkirschner I missed this point about refactoring the global scope, will be addressing it.

reddyashish:
@mjkkirschner Added the dispose method for the global Scope of completion provider. The Scope will be created for each completion provider instance with a GUID identifier and that will be disposed after the editor is closed. 

Also resolved the merge conflicts. Please take a look and Let me know if that looks good.

src/NodeServices/Interfaces.cs
```diff
@@ -235,6 +235,7 @@ internal interface IExternalCodeCompletionProviderCore
         /// <param name="dynamoCorePath"></param>
         void Initialize(string dynamoCorePath);
 
+        void Dispose();
```
mjkkirschner, line 238: hmmm, I think this is strange, can you try to implement `IDisposable` on the concrete types instead?
That seems much more idiomatic for c#.

mjkkirschner: One reason I say this is that changing interfaces after they are released is usually an API break - it's harder to judge here because this is an internal interface... so likely not an issue, but I'm unsure.

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,1292 @@
...
+        /// imported symbols.
+        /// </summary>
+        public object Engine
+        {
+            get { return engine; }
+            set { engine = (PythonEngine)value; }
+        }
+        private PyScope scope;
+
+        internal string globalScopeName = "globalScope";
```
mjkkirschner, line 892: this is really strange - you're not using this as a globalScopeName, you're using it as a unique scope per completion provider. I think this could be named more clearly.

I have some thoughts on how to get rid of this completely, will try to expand below.

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,1292 @@
...
+            }
+            finally
+            {
+                PythonEngine.ReleaseLock(gs);
+            }
+        }
+
+        #endregion
+
+        private PyScope CreateGlobalScope()
```
mjkkirschner, line 1209: this function again is named the wrong thing for what it does.

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,1292 @@
...
+        {
+            PyScopeManager.Global.TryGet(globalScopeName, out PyScope Scope);
+
+            if (Scope == null)
+            {
+                Scope = Py.CreateScope(globalScopeName);
```
mjkkirschner, line 1215: I think you don't need to use this name at all - I believe if you just call `Py.CreateScope()` you'll get a new scope, and you can assign that as a field of this class and use the reference when you dispose.

The only reason to use a named scope is if you need to lookup this particular scope at some point, does that ever happen?

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
reddyashish: Changed the name to CreateUniquePythonScope()

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
reddyashish: Changed the variable name accordingly.

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
reddyashish: Before creating a Python Scope we check for the Scope by its name from the PyScopeManager and only create it when it is not present. Otherwise we will run into the issue of creating many scopes for each provider if we do not set a name to it. What do you think?

src/NodeServices/Interfaces.cs
reddyashish: We can use IDisposable interface for DSCPythonCodeCompletionProviderCore and IronPythonCodeCompletionProviderCore but we would need the dispose() in the interface to call it on the completionProviderAdaptor in line 222 of ScriptEditorWindow.xaml.cs .

mjkkirschner: you can do a check like this
(if provider is IDisposable dis){
dis.Dispose()
}

uses c#7 pattern match to get rid of the `as` statement.

reddyashish:
@mjkkirschner Addressed the comments. Please take a look.

test/Libraries/DynamoPythonTests/CodeCompletionTests.cs
```diff
@@ -138,6 +152,10 @@ public void SetupPythonTests()
             var assemblyPath = Assembly.GetExecutingAssembly().Location;
             var moduleRootFolder = Path.GetDirectoryName(assemblyPath);
 
+            Assembly.LoadFrom(Path.Combine(moduleRootFolder, "DSCPython.dll"));
+          //  Assembly.LoadFrom(Path.Combine(moduleRootFolder, "Python.Included.dll"));
```
mjkkirschner, line 156: @reddyashish whats this commented out code?

test/Libraries/DynamoPythonTests/CodeCompletionTests.cs
```diff
@@ -290,62 +364,89 @@ public void CanMatchBasicImportStatementMultiLine()
...
+            completionList = completionData.Select(d => d.Text);
+            Assert.IsTrue(completionList.Any());
+            Assert.IsTrue(completionList.Intersect(new[] { "IO", "Console", "Reflection" }).Count() == 3);
             Assert.AreEqual(2, completionProvider.ImportedTypes.Count);
             Assert.IsTrue(defaultImports.SequenceEqual(completionProvider.ImportedTypes.Keys.ToList()));
         }
 
         [Test]
         [Category("UnitTests")]
+        [Category("Failure")]
```
mjkkirschner, line 417: so this a test that will be fixed in a followup?

mjkkirschner: are there any cpython completion tests that test new classes that are defined in the python blocks? I understand it would currently fail.

mjkkirschner:
LGTM if the tests all pass. - But please look through and don't leave any commented out code around unless there is a comment.

test/Libraries/DynamoPythonTests/CodeCompletionTests.cs
reddyashish: Ah, yes forgot to remove it. Was testing if we needed we these binaries for the tests.

test/Libraries/DynamoPythonTests/CodeCompletionTests.cs
reddyashish: Yes this test will be fixed in a followup task.
You mean tests for new classes in Python? or the tests for new CPython completion provider?

mmisol:
Hi @reddyashish . I did some more testing, this time with .NET imported types and I noticed a basic use case that does not seem to work. Using the default snippet that Dynamo includes in Python nodes:
```python
# Load the Python Standard and DesignScript Libraries
import sys
import clr
clr.AddReference('ProtoGeometry')
from Autodesk.DesignScript.Geometry import *

# The inputs to this node will be stored as a list in the IN variables.
dataEnteringNode = IN

# Place your code below this line
Point.ByCoordinates()

# Assign your output to the OUT variable.
OUT = 0
```
I didn't get any suggestions when writing `Point.` using the `CPython3` engine. However, if I write `Autodesk.DesignScript.Geometry.Point.` I do get suggestions, so there may be something that needs to be adjusted. Note that both options work fine when the completion provider is `IronPython2`.

Another thing I noticed is some weirdness regarding the provider used for suggestions and the drop down value. This is probably a side effect from recent changes that went into master. What I notice is that changing the drop down and hitting save is not enough to change the provider used for suggestions. It's always needed to close and re-open the window. This is probably minor with respect to the previous issue, but still would be nice if it worked.

reddyashish:
@mmisol Thank you for the testing and I looked into the issue you pointed out. I made a fix to handle those cases. 

Regarding your point, yes I noticed it as well. That behavior changed due to the design changes in this PR: https://github.com/DynamoDS/Dynamo/pull/11182. Currently the node's engine property doesn't update when the value in the dropdown is changed. The engine property is only updated when the save button is clicked. That is the reason, the completion provider is not updated. I understand that it can be confusing for users. We can get this point up with the team in the standup and make a decision on it.

mmisol:
> The engine property is only updated when the save button is clicked. That is the reason, the completion provider is not updated.

@reddyashish What do you mean by this? Shouldn't the provider used for suggestions also change with the engine? I would expect that both change simultaneously **_after_** saving.

mmisol:
I'm also noticing a hard crash of Dynamo, in a random but quite consistent fashion. Attaching a screenshot:

Screen Shot 2020-10-26 at 9 37 47 AM

This happens while I am testing this. I have a graph with several Python nodes with different code in them. Opening them, triggering auto complete and closing them around five times eventually leads to that error.

mmisol:
Another issue, although minor when compared with the previous, is that suggestions always seem to be identified as fields, i.e. they use the `ExternalCodeCompletionType.Field` enumeration, even when they are methods or classes like the screenshots show:
Screen Shot 2020-10-26 at 9 56 40 AM

Screen Shot 2020-10-26 at 10 01 52 AM

reddyashish:
@mmisol Can you share the graph you used to reproduce the crash?

mmisol:
@reddyashish I found I can reproduce it with just a new Python node, set with the `CPython3` engine. Just open the editor, type `Circle.` to trigger auto-completion and close it. After doing that several times, I get the crash.

reddyashish:
@mmisol I have fixed the Dynamo hard crash that was happening. Had to remove the wait() statement that was causing a threading issue. 
Also fixed the case to correctly identify the completion items as fields and methods. 
Added the condition to update the completion provider after the dropdown value is changed and saved. 

I have addressed the issues you pointed out. Appreciate the time you took in testing all these cases. Please take a look after these recent changes and let me know if I missed anything.

reddyashish:
The self-serve job: https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/737/.

The only failures showing up in the job are related to Excel tests, which are not related to these changes.

mmisol:
Thanks Ashish. I'm continuing with the tests. Related to the icons for suggested members, I still see members of a module listed as fields. See the screenshots for a comparison:
<table>
<tr>
<td>
IronPython2
</td>
<td>
CPython3
</td>
</tr>
<tr>
<td>
ironpython2
</td>
<td>
cpython3
</td>
</tr>
</table>

mmisol:
Also properties do not seem to be recognized when using CPython3. Rather than the property, what's shown is `get_` methods:
<table>
<tr>
<td>
IronPython2
</td>
<td>
CPython3
</td>
</tr>
<tr>
<td>
IronPython2
</td>
<td>
CPython3
</td>
</tr>
</table>

mmisol:
Another thing I noticed from the previous example is that CPython3 cannot recognize the type of `my_circle` in the following code sample:
```
import sys
import clr
clr.AddReference('ProtoGeometry')
from Autodesk.DesignScript.Geometry import Circle

my_circle = Circle.ByCenterPointRadius()
my_circle.
```
If you take a look at the screenshot in my previous message, that actually results in suggestions for IronPython2. For CPython3 it did not, so I had to change the code to use the class instead.

mmisol:
About the hard crash I reported earlier, I noticed that the underlying error still happens to me while testing, although not as frequently as possible, and now I see it associated to execution. I'll try to see if I can reproduce that in master, maybe it's related to calling `SetupPython` too much or too frequently. Will let you know what I find.

mmisol:
@reddyashish An update on the previous. Taking a look at what Python.Included is doing, it's pretty clear the cause of the problem is calling `SetupPython` too many times. Every time we do, it adds the Python installation location to the path, without looking if it was already there. I will make a PR to avoid calling it more than once in master, but please do the change for the autocompletion bits here.

reddyashish:
Fixed the case for the below code sample. It now handles the autocompletion for object instances as well. 

import sys
import clr
clr.AddReference('ProtoGeometry')
from Autodesk.DesignScript.Geometry import Circle

my_circle = Circle.ByCenterPointRadius()
my_circle.

mmisol:
@reddyashish What I'm seeing after the last change you made is Iron Python suggestions being used again, like I first reported here https://github.com/DynamoDS/Dynamo/pull/11065#issuecomment-690607360. This looks like it's using Iron Python code completion again.

I still didn't look in depth at the code, but I find it strange that with the small change you did in your last commit everything changed so drastically. I feel we are not converging with this PR and should probably revisit the approach. I propose we start by answering the questions I made here https://github.com/DynamoDS/Dynamo/pull/11065#pullrequestreview-486161514 and continue analyzing the different scenarios and investigating solutions for them. After that, decide how to scope the work so that it can be done in an incremental way, so that at every stage the code can be reviewed and we are aware of what is supported and what is not supported.

reddyashish:
@mmisol I forgot to add a condition to differentiate between Python variable types and internal variable types. That's why you were seeing the same suggestions as before. 

We can discuss the points you put forward in the meeting today.

QilongTang:
@reddyashish Changed the branch base for you. Once you resolve the merge conflicts, we can merge this PR

reddyashish:
Merging this PR to the feature branch.

PR #11072, opened 2020-08-27
Include rendering time in Perf Vis tests

mmisol:
### Purpose

Include rendering time in performance visualization tests.

### Declarations

### Reviewers

@aparajit-pratap 

### FYIs

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
@@ -14,6 +14,7 @@ using System.Windows.Input;
 using System.Windows.Media;
 using System.Windows.Media.Imaging;
 using System.Windows.Media.Media3D;
+using System.Windows.Threading;
 using System.Xml;
 using Autodesk.DesignScript.Interfaces;
 using Dynamo.Controls;
@@ -855,10 +856,13 @@ namespace Dynamo.Wpf.ViewModels.Watch3D
                 (dynamoModel as DynamoModel).Report3DPreviewOutage(summary, description);
             }
 #if DEBUG
-            renderTimer.Stop();
-            Debug.WriteLine(string.Format("RENDER: {0} ellapsed for compiling assets for rendering.", renderTimer.Elapsed));
-            renderTimer.Reset();
-            renderTimer.Start();
+            // Defer stopping the timer until after the rendering has occurred
+            Dispatcher.CurrentDispatcher.BeginInvoke(DispatcherPriority.Loaded, new Action(() =>
+            {
+                renderTimer.Stop();
+                Debug.WriteLine(string.Format("RENDER: {0} ellapsed time rendering.", renderTimer.Elapsed));
+                renderTimer.Restart();
+            }));
 #endif
 
             OnSceneItemsChanged();
@@ -1026,19 +1030,10 @@ namespace Dynamo.Wpf.ViewModels.Watch3D
             return true;
         }
 
-        #region internal methods
+#region internal methods
 
         internal void ComputeFrameUpdate()
         {
-#if DEBUG
-            if (renderTimer.IsRunning)
-            {
-                renderTimer.Stop();
-                Debug.WriteLine(string.Format("RENDER: {0} ellapsed for setting properties and rendering.", renderTimer.Elapsed));
-                renderTimer.Reset();
-            }
-#endif
-
             // Raising a property change notification for
             // the SceneItems collections causes a full
             // re-render including sorting for transparency.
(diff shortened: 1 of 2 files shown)


src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -1030,15 +1034,6 @@ protected override bool CanToggleCanNavigateBackground(object parameter)
 
         internal void ComputeFrameUpdate()
         {
-#if DEBUG
-            if (renderTimer.IsRunning)
```
aparajit-pratap, line 1034: Why was this deleted? I saw that control still went here some times during my debugging.

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -855,10 +856,13 @@ public override void GenerateViewGeometryFromRenderPackagesAndRequestUpdate(Rend
...
-            Debug.WriteLine(string.Format("RENDER: {0} ellapsed for compiling assets for rendering.", renderTimer.Elapsed));
-            renderTimer.Reset();
-            renderTimer.Start();
+            // Defer stopping the timer until after the rendering has occurred
+            Dispatcher.CurrentDispatcher.BeginInvoke(DispatcherPriority.Loaded, new Action(() =>
+            {
+                renderTimer.Stop();
+                Debug.WriteLine(string.Format("RENDER: {0} ellapsed time rendering.", renderTimer.Elapsed));
+                renderTimer.Restart();
+            }));
```
aparajit-pratap, line 865: 👍

src/VisualizationTests/PerformanceVisualizationTests.cs
```diff
@@ -18,14 +18,16 @@ namespace WpfVisualizationTests
...
+        const int SamplingAmount = 3;
 
         [Test]
         [TestCaseSource("PerformanceTestSource")]
         public void TestVisualizationPerformance(string filePath)
         {
             ViewModel.OpenCommand.Execute(filePath);
+            DispatcherUtil.DoEvents();
             RunCurrentModel();
+            DispatcherUtil.DoEvents();
```
aparajit-pratap, line 30: Do these tests now run as part of the other performance benchmark tests? It's been a long time since I ran those tests locally. It would be interesting to see what are the render time results from running these.

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
mmisol: Since we are scheduling the timer stop for after rendering occurs, this was actually disrupting the elapsed time obtained, because it used the same timer.
While we could change this to use another timer instead of deleting it, I didn't see this provide accurate results during my tests, so I just deleted it. Also, the new measure includes what this is supposed to include, so I don't think we want to track this separately.

src/VisualizationTests/PerformanceVisualizationTests.cs
mmisol: These tests relation with performance benchmark tests is only that by default they use the same graphs. These are aimed to just measure rendering time, which is something that gets mixed with a lot of other stuff in the performance benchmark.
Also these tests are not really a tool, you need to do some tweaking and do some extra steps to get results out of them.

aparajit-pratap: Can you include comments on how to run these tests and see the timing results?

mmisol: It's in the top of this file, in the class comment.

aparajit-pratap: `For that, you should make sure the renderTimer Stopwatch in HelixWatch3DViewModel is enabled and logging to Console. Output should be redirected to a file.`

Does one need to add a `TextWriterTraceListener` for this? Can you give an example of how would someone carry out the steps above?

mmisol: I haven't explored using `TextWriterTraceListener`. For my testing I just run the test using nunit from the command line and redirect output to a file using `>`.
If you want to run in `Release` mode, you need to do changes to make sure the Stopwatch is enabled. Removing the `#if DEBUG` surrounding the timer where it's defined and used should be needed. Also, to guarantee you are logging to Console in Release, you should use `Console.WriteLine` rather than `Debug.WriteLine`.

aparajit-pratap: Thanks. Should we just change `Debug.WriteLine` to `Console.WriteLine` in that case?

mmisol: I think `Debug.WriteLine` should be considered a better practice in general, even if under a `#if DEBUG` flag. The possibility to add trace listeners is there, even if we don't use it today. In general we are using `Debug.WriteLine` in other places to log stuff, so I think it's more consistent if we keep it that way.

PR #11075, opened 2020-08-28
Handle overflow in DesignScript and core nodes

mmisol:
### Purpose

Operations that may result in an integer overflow in DesignScript code
or core nodes now display a warning when one occurs. The result is
maintained as it was, in order to avoid introducing incompatibilities.

### Declarations

### Reviewers

@aparajit-pratap @reddyashish 

### FYIs

@DynamoDS/dynamo


mmisol:
Some inconsistencies I found while doing this:
- Math.Factorial is one of the few Math nodes that handled overflow. It returns null, in contrast with other Math nodes.
- Math.Abs gives does not handle overflow, but .NET seems to it internally because it throws. I would label this as an inconsistency in .NET.
- Math.Round returns a double, while ceiling and floor return integers. This means overflow cannot happen for Math.Round.

src/Libraries/CoreNodes/Properties/Resources.en-US.resx
```diff
@@ -156,6 +156,9 @@
   <data name="FromObjectObsolete" xml:space="preserve">
     <value>This node is obsolete, please use "String from Object"</value>
   </data>
+  <data name="IntegerOverflow" xml:space="preserve">
+    <value>The operation resulted in an integer overflow. Its result may be unexpected.</value>
```
aparajit-pratap, line 160: @Amoursol does this error message sound okay? Any scope for improvement?

Amoursol: The term 'overflow' will not make sense to an end user. I suggest we keep it to be technically correct, but change the second sentence:

"The operation resulted in an integer overflow. Integer overflow occurs when a numeric value falls outside of the range in which that integer can be represented - either higher than the maximum, or lower than the minimum. This may cause the integer value to wrap around the maximum or minimum value and generate an unexpected result."

mmisol: @Amoursol The error message seem rather long. Would you be ok if I added a linked explanation as an html document for the documentation browser?

Amoursol: Totally cool to have it as a verbose explanation in the Doc Browser, instead yes :)

mmisol: @Amoursol Added it to the doc browser, including how to fix and a code sample. Let me know if there is something you would like to change.
Part 1
doc_1
Part 2 (after scrolling down a little)
doc_2

Amoursol: All comments handled via Slack, but for logging reproduced here: 

Under **How to Fix**: I would add: “When working with very large numbers, try to use the **double** `(Decimal placed number)` type…” , and, “…type instead of the **int** `(Whole number)` type.”

image

mmisol: Thanks for the help with the docs @Amoursol ! I have just pushed the new version including your changes. I'll post just the end, the previous section did not change:
Screen Shot 2020-08-31 at 5 19 23 PM

Amoursol: LGTM! Did you also include the parenthesis callouts of `(Whole number)` for int, and `(Decimal place number)` for double?

mmisol: Yes, I did. Sorry, I forgot to include a screenshot of it.

Amoursol: Perfect, thank you @mmisol. No need to screenshot 😊

PR #11016, opened 2020-08-17
[DYN-3027] Libraries core node model coverage

Astul-Betizagasti:
### Purpose

Increase code coverage for the Libraries/CoreNodeModel project

### Declarations

Check these if you believe they are true

- [ ] The codebase is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.
- [ ] This PR modifies some build requirements and the readme is updated

### Reviewers

Aaron Tang (@QilongTang )
Michael Kirschner (@mjkkirschner )

### FYIs

Alfredo Pozo (@alfredo-pozo )

PR #11076, opened 2020-08-29
Revert "[DYN-3027] Libraries core node model coverage"

alfredo-pozo:
Reverts DynamoDS/Dynamo#11016


aparajit-pratap:
@alfredo-pozo @Astul-Betizagasti can you point out what specifically in PR #11016 broke the CoreNodeTests project? Is it possible to fix the issue instead of reverting the whole PR?

PR #11080, opened 2020-08-31
Guard against parser errors detected while migrating python code

aparajit-pratap:
### Purpose

https://jira.autodesk.com/browse/DYN-3102
Guard against crash due to parser errors detected while migrating python code. The following message is shown in the migration assistant in such a case:

image

### Declarations

- [ ] The level of testing this PR includes is appropriate - **WIP**

### Reviewers

@Amoursol - any inputs on the error message (see screenshot).
@mmisol

### DIFF
--- a/src/PythonMigrationViewExtension/Differ/IDiffViewViewModel.cs
+++ b/src/PythonMigrationViewExtension/Differ/IDiffViewViewModel.cs
@@ -1,34 +1,10 @@
-﻿using System;
-using System.Windows;
-using System.Windows.Data;
-
-namespace Dynamo.PythonMigration.Differ
+﻿namespace Dynamo.PythonMigration.Differ
 {
     public enum State
     {
         NoChanges,
         HasChanges,
         Error
-    }
-
-    public class DiffStateToVisibilityConverter : IValueConverter
-    {
-        public object Convert(object value, Type targetType, object parameter, System.Globalization.CultureInfo culture)
-        {
-            if (value != null)
-            {
-                var state = (State) value;
-                if (state == State.Error)
-                    return Visibility.Hidden;
-            }
-
-            return Visibility.Visible;
-        }
-
-        public object ConvertBack(object value, Type targetType, object parameter, System.Globalization.CultureInfo culture)
-        {
-            throw new NotSupportedException();
-        }
     }
 
     public interface IDiffViewViewModel
--- a/src/PythonMigrationViewExtension/Differ/IDiffViewViewModel.cs
+++ b/src/PythonMigrationViewExtension/Differ/IDiffViewViewModel.cs
@@ -1,4 +1,8 @@
-﻿namespace Dynamo.PythonMigration.Differ
+﻿using System;
+using System.Windows;
+using System.Windows.Data;
+
+namespace Dynamo.PythonMigration.Differ
 {
     public enum State
     {
@@ -7,11 +11,30 @@
         Error
     }
 
+    public class DiffStateToVisibilityConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, System.Globalization.CultureInfo culture)
+        {
+            if (value != null)
+            {
+                var state = (State) value;
+                if (state == State.Error)
+                    return Visibility.Hidden;
+            }
+
+            return Visibility.Visible;
+        }
+
+        public object ConvertBack(object value, Type targetType, object parameter, System.Globalization.CultureInfo culture)
+        {
+            throw new NotSupportedException();
+        }
+    }
+
     public interface IDiffViewViewModel
     {
         ViewMode ViewMode { get; }
-        State DiffState { get; }
-        bool Error { get; set; }
+        State DiffState { get; set; }
     }
 
     public enum ViewMode
--- a/src/PythonMigrationViewExtension/Differ/IDiffViewViewModel.cs
+++ b/src/PythonMigrationViewExtension/Differ/IDiffViewViewModel.cs
@@ -1,9 +1,17 @@
 ﻿namespace Dynamo.PythonMigration.Differ
 {
+    public enum State
+    {
+        NoChanges,
+        HasChanges,
+        Error
+    }
+
     public interface IDiffViewViewModel
     {
         ViewMode ViewMode { get; }
-        bool HasChanges { get; }
+        State DiffState { get; }
+        bool Error { get; set; }
     }
 
     public enum ViewMode
--- a/src/PythonMigrationViewExtension/MigrationAssistant/PythonMigrationAssistantViewModel.cs
+++ b/src/PythonMigrationViewExtension/MigrationAssistant/PythonMigrationAssistantViewModel.cs
@@ -8,6 +8,7 @@ using Dynamo.Graph.Workspaces;
 using Dynamo.Interfaces;
 using Dynamo.PythonMigration.Controls;
 using Dynamo.PythonMigration.Differ;
+using Python.Runtime;
 using PythonNodeModels;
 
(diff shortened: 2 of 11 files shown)


src/PythonMigrationViewExtension/Differ/IDiffViewViewModel.cs
```diff
@@ -1,9 +1,17 @@
...
+        HasChanges,
+        Error
+    }
+
     public interface IDiffViewViewModel
     {
         ViewMode ViewMode { get; }
-        bool HasChanges { get; }
+        State DiffState { get; }
+        bool Error { get; set; }
```
mmisol, line 14: Do we need this boolean if `DiffState` can already be `Error`?

aparajit-pratap:
> Just one comment.
> 
> On a general note, I don't think letting the user continue with the migration is a good idea in this case. We would create a backup with the invalid code, which is probably not useful.

Well, I guess the same could be said when there are no code changes needed between python2 and python3.

mmisol:
@aparajit-pratap In my mind it's a different case.

When there are no changes for the current node, you can still say things are "in order" and other nodes can still present changes, so the backup is still useful. Also, remember you are changing the engine, so the backup should happen at this point.

However, if the syntax is incorrect then most likely things are not "in order". Maybe the user typed something by mistake before lunching the migrator? Backing up the graph with that syntax error is what I believe is not useful. If the user needs to restore the backup, the graph will be invalid, which is probably not expected.

aparajit-pratap:
@mmisol I wanted to show a message to the user saying there is an error in the code when he/she clicked the 2-3 migration icon. If we simply don't do anything then the user could be confused in the absence of any feedback. Perhaps I can continue to show the message in the migration assistant but disable the file backup?

src/PythonMigrationViewExtension/Differ/IDiffViewViewModel.cs
aparajit-pratap: I'm using the `DiffState` instead of the previously used `HasChanges` boolean in the xaml view, which is why I needed to change it to an enum from a bool as it now represents 3 states - `HasChanges` - when the diff viewer shows code diff, `NoChanges` - when we show the message saying the code has no changes, and this new `Error` state, where I wanted to show the error message in the same UI.

mmisol:
@aparajit-pratap I'm not saying we should not give feedback to the user, I'm just saying we shouldn't let them proceed. For this purpose we could, for instance, disable the accept button.

Beyond the backup, we should also avoid changing the engine. Otherwise, when we do backup we start from invalid content anyway. Preventing the migration seems the safest approach in this situation.

aparajit-pratap:
@Amoursol do you have any feedback on the expected behaviour when there is a parser error in the code and the migration cannot proceed. Do we simply want to disable the migration assistant UI (not do anything), or show a message like the one above, or show some other message popup?

src/PythonMigrationViewExtension/Differ/IDiffViewViewModel.cs
mmisol: I understand that but my question remains. Why do we need the boolean `Error`?

aparajit-pratap: I'm using it as a flag to set it to `true` after catching the `PythonException`. I could probably just use the `DiffState` enum and use that instead. Let me try.

Amoursol:
@aparajit-pratap - This is in the use case where users have non-compliant IronPython2 code they wish to migrate? i.e adding in a single line of `blah blah blah` not explicitly called out as a string causing the Python engine to fail? 

If so, I agree that we should not allow them to run the `2->3` migration assistant (Disable 'accept' button) + provide a message stating why they cannot use the assistant (The message you have above is good if we change accept to cancel!)

src/PythonMigrationViewExtension/Differ/IDiffViewViewModel.cs
```diff
@@ -1,9 +1,40 @@
...
+namespace Dynamo.PythonMigration.Differ
 {
+    public enum State
+    {
+        NoChanges,
+        HasChanges,
+        Error
+    }
+
+    public class DiffStateToVisibilityConverter : IValueConverter
```
mmisol, line 14: Maybe this can be moved out to its own file since it's not part of the interface itself.

src/PythonMigrationViewExtension/MigrationAssistant/PythonMigrationAssistantViewModel.cs
```diff
@@ -42,37 +43,49 @@ public IDiffViewViewModel CurrentViewModel
...
+            diffModel = sidebyside.BuildDiffModel(OldCode, NewCode, false);
         }
 
         /// <summary>
         /// Replaces the code in the Python node with the code changes made by the Migration Assistant.
         /// </summary>
         public void ChangeCode()
         {
-            if (!this.CurrentViewModel.HasChanges)
+            if (CurrentViewModel.DiffState == State.Error || CurrentViewModel.DiffState == State.NoChanges)
```
mmisol, line 86: This code should not execute when the state is `Error` but please remove `CurrentViewModel.DiffState == State.Error` from here as it looks confusing. When the state is error, what's inside the `if` is something we don't want to do.

src/PythonMigrationViewExtension/MigrationAssistant/PythonMigrationAssistantViewModel.cs
```diff
@@ -42,37 +43,49 @@ public IDiffViewViewModel CurrentViewModel
...
             {
-                if (this.PythonNode.Engine == PythonEngineVersion.CPython3)
-                    return;
```
mmisol, line 74: Is this already being checked in the property setter? If not please add it back.

src/PythonMigrationViewExtension/Properties/Resources.resx
```diff
@@ -170,6 +170,12 @@ By clicking “Continue”, you are agreeing to implement the syntax changes to
   <data name="MigrationAssistantNoChangesStateMessage" xml:space="preserve">
     <value>Your code is ready to run with the Python 3 engine. Click accept and we'll switch to the Python 3 engine for you.</value>
   </data>
+  <data name="MigrationAssistantErrorStateHeader" xml:space="preserve">
+    <value>There is an error in your code!</value>
+  </data>
+  <data name="MigrationAssistantErrorStateMessage" xml:space="preserve">
+    <value>Fix all errors before attempting code migration. Clicking cancel will switch back to the Python 2 engine without any code changes.</value>
```
mmisol, line 177: Technically, the engine will not be switched back. It should simply remain the same. Even though it doesn't make much sense, you could also launch the migrator with the `CPython3` engine set.

How about something more on the lines of: `The code migration cannot continue. Please click cancel and fix all errors before re-attempting it.`

Amoursol: I'll file a new task - but we should disable the `2->3` button when you have the CPython3 engine set. It's confusing otherwise!

src/PythonMigrationViewExtension/MigrationAssistant/PythonMigrationAssistantViewModel.cs
aparajit-pratap: The setter has a check already. It doesn't set it unless it is a different engine.

src/PythonMigrationViewExtension/Properties/Resources.resx
aparajit-pratap: If the engine is already CPython, I'm not doing anything special currently like switching it back to IronPython. So I'm removing the second sentence altogether and leaving it at that.

PR #11092, opened 2020-09-02
Update pan/orbit state icon

mmisol:
### Purpose

Fixes issues related to the state of the pan icon not reflecting the
actual current state.

There were two problems preventing this from updating.

Some actions did not seem to trigger a property change on IsPlanning
and IsOrbiting at the BackgroundPreviewViewModel. There are properties
with the same name in the local view model that get updated more
consistently.

Some actions were triggering property change events in the wrong place.
These were moved to the right place.

While there were no reported issues with IsOrbiting, this property
should behave in the same way, so the same changes were applied.

### Declarations

### Reviewers

@QilongTang @zeusongit 

### FYIs

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
@@ -881,6 +881,7 @@ namespace Dynamo.ViewModels
 
             // Reset workspace state
             CurrentSpaceViewModel.CancelActiveState();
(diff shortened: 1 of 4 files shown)


mmisol:
@zeusongit I didn't actually remove those comments. Just moved them to the `RefreshState()` function which is now called from those other places.

zeusongit:
Ah.. ok, got it.

QilongTang (approved):

PR #11126, opened 2020-09-17
Style Workspace References consistently

mmisol:
### Purpose

Changed the styles of the workspace references extension window to be
more consistent with respect to Dynamo. Particularly, buttons that used
default WPF button style were switched to now use the same style as in
the "Geometry Scaling" window. Also, refresh is no longer a button, to
get rid of the undesired hover style, and the image foreground is
changed on hover.

Here is how it looks like now:

workspace_references

### Declarations

### Reviewers

@QilongTang @aparajit-pratap 

### FYIs

@Amoursol

### DIFF
--- a/src/DynamoCoreWpf/UI/Themes/Modern/DynamoModern.xaml
+++ b/src/DynamoCoreWpf/UI/Themes/Modern/DynamoModern.xaml
@@ -683,6 +683,68 @@
         </Setter>
     </Style>
 
+    <Style x:Key="SmallTextButton" TargetType="{x:Type Button}">
+        <Setter Property="Template">
+            <Setter.Value>
+                <ControlTemplate>
+                    <Border x:Name="container"
+                                Background="Transparent"
+                                BorderBrush="#3c3c3c"
+                                BorderThickness="1">
+                        <Grid x:Name="inner"
+                                    Background="#373737">
+                            <TextBlock x:Name="text"
+                                       HorizontalAlignment="Center"
+                                       VerticalAlignment="Center"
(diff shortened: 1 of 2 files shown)


Amoursol:
Hi @mmisol - Is it possible to make the button a little smaller? It currently takes up significant real-estate. Image shows before and after. 

image

QilongTang:
@mmisol About @Amoursol 's request, does that need a new template for button?

mmisol:
@QilongTang Unfortunately yes. While there is the possibility to create a style based on another, it does not allow to slightly change the template, so I'm creating a new one. It's an easy change anyway.

QilongTang:
> @QilongTang Unfortunately yes. While there is the possibility to create a style based on another, it does not allow to slightly change the template, so I'm creating a new one. It's an easy change anyway.

That is actually good, I mean still better than the inline style we have right now

mmisol:
It's looking like this now. What do you guys think? @Amoursol  @QilongTang 

smaller_button

QilongTang:
@mmisol Can you test in German and some language with long strings? If they can be displayed correctly, I am good with this change

mmisol:
@QilongTang I'm pretty sure I went back to the same size and spacing as before. Could you do the comparison again @Amoursol ?

Amoursol:
> @QilongTang I'm pretty sure I went back to the same size and spacing as before. Could you do the comparison again @Amoursol ?

Looks good to me now @mmisol ! 
image

QilongTang (approved):

PR #11129, opened 2020-09-18
Given an input port retrieve compatible upstream nodes

aparajit-pratap:
### Purpose

This contains APIs to retrieve the type of an input port and return matching nodes that output the same type for that port. 
node_autcomplete

Currently, we need to make note of the following limitations:
- `NodeModel` nodes need to specify the fully qualified types in both the `InPortTypes` and `OutPortTypes` attributes for node autocomplete to work properly
- Only zero-touch, builtin, and `NodeModel` nodes (provided the above condition is met) are supported currently
- Multi-return nodes aren't supported. Type information for multi-return nodes relies on whether the types are specified explicitly in the XML code comments for the node and this isn't always guaranteed.
- There could be specific cases such as `var` and `function` types that aren't fully tested. We may need to filter out `var` specifically to avoid returning a vast number of matching results.
- An exact match is currently being made for array types, for e.g. if a node accepts `Point[]` as input, only a node that returns `Point[]` will match and not one that returns `Point`. Ideally, this should be handled by returning all nodes that return `Point` since DS supports replication and array promotion. We can address this in a future PR.

### Declarations

### Reviewers

@DynamoDS/dynamo 

### FYIs
@Amoursol

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Core/PortViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/PortViewModel.cs
@@ -1,7 +1,11 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Windows;
+using Dynamo.Engine;
 using Dynamo.Graph.Nodes;
 using Dynamo.Models;
+using Dynamo.Search.SearchElements;
 using Dynamo.UI.Commands;
 using Dynamo.Utilities;
 
@@ -371,7 +375,9 @@ namespace Dynamo.ViewModels
         private void OnRectangleMouseEnter(object parameter)
         {
             if (MouseEnter != null)
+            {
                 MouseEnter(parameter, null);
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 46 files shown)


src/Libraries/CoreNodeModels/ColorRange.cs
```diff
@@ -20,7 +20,7 @@ namespace CoreNodeModels
     [NodeSearchTags("ColorRangeSearchTags", typeof(Resources))]
 
     [InPortNames("colors", "indices", "value")]
-    [InPortTypes("Color[]", "double[]", "double")]
+    [InPortTypes("DSCore.Color[]", "double[]", "double")]
```
aparajit-pratap, line 23: NodeModel nodes need to specify the fully qualified types in both the `InPortTypes` and `OutPortTypes` attributes for node autocomplete to work properly. We need to make this a guideline for NodeModel node authors if they wish for node autocomplete to work with their nodes. Without this requirement, it is difficult to get the feature working for such nodes.

src/Libraries/CoreNodeModels/WatchImageCore.cs
```diff
@@ -14,6 +14,7 @@ namespace CoreNodeModels
     [NodeCategory(BuiltinNodeCategories.CORE_VIEW)]
     [NodeSearchTags("WatchImageSearchTags", typeof(Resources))]
     [IsDesignScriptCompatible]
+    [InPortTypes("System.Drawing.Bitmap")]
```
aparajit-pratap, line 17: See comment above for NodeModel nodes.

Amoursol:
Excited to see this come through @aparajit-pratap ! 🥳

src/DynamoCoreWpf/ViewModels/Core/PortViewModel.cs
```diff
@@ -412,6 +417,49 @@ private void OnMouseLeftUseLevel(object parameter)
...
+        /// Returns a collection of node search elements for nodes
+        /// that output a type compatible with the port type if it's an input port.
+        /// These search elements can belong to either zero touch, NodeModel or Builtin nodes.
+        /// This method returns an empty collection if the input port type cannot be inferred or
+        /// there are no matching nodes found for the type. Currently the match is an exact match
+        /// done including the rank information in the type, e.g. Point[] or var[]..[].
+        /// The search elements can be made to appear in the node autocomplete search dialog.
+        /// </summary>
+        /// <returns>collection of node search elements</returns>
+        internal IEnumerable<NodeSearchElement> GetMatchingNodes()
```
QilongTang, line 430: It seems we can move this into `NodeAutoCompleteSearchViewModel` so everytime we show the control we update the list of entries since I have the target PortViewModel stored as a property already. What do you think？

src/Libraries/CoreNodeModels/ColorRange.cs
QilongTang: Agreed

src/DynamoCoreWpf/ViewModels/Core/PortViewModel.cs
aparajit-pratap: Sure, we can do that. Will move this method.

QilongTang: Also I have been thinking about maybe rename this method to `GetMatchingNodeSearchElements`? Since the API is not returning a collection of NodeModels. What do you think?

aparajit-pratap:
@QilongTang closing this as it has already been merged in `NodeAutoComplete` branch and reviewed.

PR #11132, opened 2020-09-21
Use node suggestion API in node autocomplete UI

aparajit-pratap:
### Purpose

Call node suggestion API from node autocomplete UI.

### Declarations

### Reviewers

@QilongTang 

### FYIs

### DIFF
--- a/src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml.cs
+++ b/src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml.cs
@@ -24,9 +24,9 @@ namespace Dynamo.UI.Controls
 
         internal event Action<ShowHideFlags> RequestShowNodeAutoCompleteSearch;
 
-        public SearchViewModel ViewModel
+        public NodeAutoCompleteSearchViewModel ViewModel
         {
-            get { return DataContext as SearchViewModel; }
+            get { return DataContext as NodeAutoCompleteSearchViewModel; }
         }
 
         public NodeAutoCompleteSearchControl()
@@ -87,7 +87,8 @@ namespace Dynamo.UI.Controls
             if (searchElement != null)
             {
                 searchElement.Position = ViewModel.InCanvasSearchPosition;
-                searchElement.ClickedCommand.Execute(null);
+                PortViewModel port = ViewModel.targetPortViewModel;
+                searchElement.CreateAndConnectCommand.Execute(port.PortModel);
             }
         }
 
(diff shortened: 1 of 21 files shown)


src/DynamoCoreWpf/ViewModels/Watch3D/DynamoGeometryModel3D.cs
```diff
@@ -48,7 +49,9 @@ public class DynamoGeometryModel3D : MaterialGeometryModel3D
     {
         protected override SceneNode OnCreateSceneNode()
         {
-            return new DynamoMeshNode();
+            var node = new DynamoMeshNode();
+            node.CullMode = CullMode.Back;
```
QilongTang, line 53: hmm any comments here?

aparajit-pratap: Reverted. This change was unrelated.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -32,5 +32,16 @@ internal void InitializeDefaultAutoCompleteCandidates()
             }
             FilteredResults = candidates;
         }
+
+        internal void PopulateAutoCompleteCandidates()
+        {
+            var searchElements = PortViewModel.GetMatchingNodes();
```
QilongTang, line 38: Thanks for moving it. But I was hoping we can move all of the code from `PortViewModel` to the customized search view model here given that 1. those code are not really related to PortView 2. I feel `PortViewModel` does not necessarily obtain knowledge about search elements 3. We still have access to DynamoViewModel from here. What do you think?

aparajit-pratap: I haven't moved the code yet. Doing so now.

src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml.cs
```diff
@@ -121,7 +121,8 @@ private void OnNodeAutoCompleteSearchControlVisibilityChanged(object sender, Dep
             Dispatcher.BeginInvoke(new Action(() =>
             {
                 SearchTextBox.Focus();
-                ViewModel.InitializeDefaultAutoCompleteCandidates();
+                //ViewModel.InitializeDefaultAutoCompleteCandidates();
+                ViewModel.PopulateAutoCompleteCandidates();
```
QilongTang, line 125: We probably want to consider returning the number of filtered results here, if the number is 0 which means we failed to find any match, then we still call `InitializeDefaultAutoCompleteCandidates()`.

aparajit-pratap: The reason I commented this out was I noticed issues with this. On a mouse up, the node suggestion list was reverting to this default list each time.

PR #11134, opened 2020-09-21
Adjust ExtensionsCollapsed after size change

mmisol:
### Purpose

Fixes the value provided by ExtensionsCollapsed to also account for the
initial state of the columns, where the sizes are 5* for the canvas and
2* for the extension bar. This in turn fixes a bug where the extension
bar wouldn't collapse if never resized.

### Declarations

### Reviewers

@QilongTang @reddyashish 

### FYIs

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
+++ b/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
@@ -1848,12 +1848,17 @@ namespace Dynamo.Controls
         {
             get
             {
-                // Threshold that determines if button should be displayed
-                if (RightExtensionsViewColumn.Width.Value < 20)
-                { extensionsCollapsed = true; }
-
+                // Special case: when the extension bar was never resized its size will be 2.
+                // While 2 is a valid size for the extension bar, 5 is not one for the canvas,
+                // so that's a safer check to be made.
+                if (CanvasColumn.Width.Value == 5)
+                {
+                    extensionsCollapsed = RightExtensionsViewColumn.Width.Value == 0;
+                }
                 else
-                { extensionsCollapsed = false; }
+                {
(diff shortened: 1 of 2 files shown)


src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -1848,12 +1848,17 @@ public bool ExtensionsCollapsed
         {
             get
             {
-                // Threshold that determines if button should be displayed
-                if (RightExtensionsViewColumn.Width.Value < 20)
-                { extensionsCollapsed = true; }
-
+                // Special case: when the extension bar was never resized its size will be 2.
```
QilongTang, line 1851: @mmisol Can we hold on on this change, I feel this assumption may not be true in the future, after we implement https://jira.autodesk.com/browse/DYN-3144

mmisol: @QilongTang I don't see why. The work here is done and it leaves the code in a better state than what it was. The fact we may need to change it later does not fix the bug. In any case, if you think it's necessary, you may want to link this PR as reference for future work.

QilongTang: @mmisol Let me ask in a different way. If `RightExtensionsViewColumn` was initialized with a different width, will this check still be valid？I was simply hoping we could fix two issues in a single shot

mmisol: @QilongTang The check would still be needed. Depending on how you do it, you may have to change the values here or not.

QilongTang (approved):
Looks good to me. Let's fix the regression first and consider the follow up in a different time. Thank you for working on this ahead @mmisol

PR #11147, opened 2020-09-24
[DYN-3027] Libraries CoreNodeModels Coverage

Astul-Betizagasti:
### Purpose

Increase code coverage for the CoreNodeMoldes

### Declarations

### Reviewers

Aaron Tang (@QilongTang )
Aparajit Pratap (@aparajit-pratap )

### FYIs

Alfredo Pozo (@alfredo-pozo )


Astul-Betizagasti:
@aparajit-pratap @QilongTang, you will see I removed some methods, those where all private and with no references, I assume they where left behind from past versions of those classes.

QilongTang (approved):
This looks good to me but would like to get confirmation from @aparajit-pratap  as well. He knows AST much better than me

test/Libraries/CoreNodesTests/CoreNodeModelsTests/BaseTypesTests.cs
```diff
@@ -0,0 +1,168 @@
...
+            testNode = DoubleInput.ParseValue("1:#TestText", new[] { '\n' }, new List<string>(), DoubleInputNode.Convert).FirstOrDefault();
+        }
+
+        [Test]
+        public void GetFSchemeValueTest()
+        {
+            Dictionary<string, double> idLookup = new Dictionary<string, double>();
+            idLookup.Add("TestText", 1);
+
+            IEnumerable<double> result = (IEnumerable<double>) testNode.GetFSchemeValue(idLookup);
```
aparajit-pratap, line 38: I don't think `GetFSchemeValue` is used any longer. It is from the old days when the Dynamo language was based on FScheme.

aparajit-pratap: Please don't add a test for this method and mark it obsolete.

test/Libraries/CoreNodesTests/CoreNodeModelsTests/BaseTypesTests.cs
```diff
@@ -0,0 +1,168 @@
...
+            OpenModel(openPath);
+
+            DoubleInputNode = CurrentDynamoModel.CurrentWorkspace.NodeFromWorkspace<DoubleInput>(
+                Guid.Parse("6fc905f8533f433a90fe4b9181463d53"));
+
+            testNode = DoubleInput.ParseValue("1..2:#n", new[] { '\n' }, new List<string>(), DoubleInputNode.Convert).FirstOrDefault();
+        }
+
+        [Test]
+        public void GetFSchemeValueTest()
```
aparajit-pratap, line 87: remove test

test/Libraries/CoreNodesTests/CoreNodeModelsTests/BaseTypesTests.cs
```diff
@@ -0,0 +1,168 @@
...
+        public void TestsSetup()
+        {
+            string openPath = Path.Combine(TestDirectory,
+                @"core\number\TestNumberDeserialization.dyn");
+            OpenModel(openPath);
+
+            DoubleInputNode = CurrentDynamoModel.CurrentWorkspace.NodeFromWorkspace<DoubleInput>(
+                Guid.Parse("6fc905f8533f433a90fe4b9181463d53"));
+
+            testNode = DoubleInput.ParseValue("1..2:#n", new[] { '\n' }, new List<string>(), DoubleInputNode.Convert).FirstOrDefault();
```
aparajit-pratap, line 83: I don't see how these expressions can be used in a `DoubleInput` node: `1..2:#n`, etc. I think this was a very old syntax used during the Fscheme days.

aparajit-pratap: These return an error now.

aparajit-pratap (changes requested):
@Astul-Betizagasti I don't think any of these tests are relevant anymore, unfortunately. `DoubleInput` node only takes numbers now, no expressions. Let's keep the deleted code in this PR and also mark the `GetFSchemeValue` methods for obsoletion.

test/Libraries/CoreNodesTests/CoreNodeModelsTests/BaseTypesTests.cs
Astul-Betizagasti: I see, I'll look into that and change it to the correct one

test/Libraries/CoreNodesTests/CoreNodeModelsTests/BaseTypesTests.cs
Astul-Betizagasti: ok I will update that on the next commit

src/Libraries/CoreNodeModels/Input/BaseTypes.cs
```diff
@@ -286,88 +286,22 @@ protected override void DeserializeCore(XmlElement element, SaveContext context)
...
 
         private static readonly Regex IdentifierPattern = new Regex(@"(?<id>[a-zA-Z_][^ ]*)|\[(?<id>\w(?:[^}\\]|(?:\\}))*)\]");
-        private static readonly string[] RangeSeparatorTokens = { "..", ":", };
 
         public static List<IDoubleSequence> ParseValue(string text, char[] seps, List<string> identifiers, ConversionDelegate convertToken)
         {
             var idSet = new HashSet<string>(identifiers);
-            return text.Replace(" ", "").Split(seps, StringSplitOptions.RemoveEmptyEntries).Select(
+            var doubleSequenceList = text.Replace(" ", "").Split(seps, StringSplitOptions.RemoveEmptyEntries).Select(
                 delegate(string x)
```
aparajit-pratap, line 294: I think we can just say `x =>`

src/Libraries/CoreNodeModels/Input/BaseTypes.cs
```diff
@@ -286,88 +286,22 @@ protected override void DeserializeCore(XmlElement element, SaveContext context)
...
-                        }
-
-                    }
+                    IDoubleInputToken startToken = ParseToken(x, idSet, identifiers);
 
                     return new OneNumber(startToken, convertToken) as IDoubleSequence;
                 }).ToList();
+
+            if (doubleSequenceList.Count == 0)
+                throw new Exception("No identifiers found.");
```
aparajit-pratap, line 302: When is this exception thrown, when the number node is empty? Can you reproduce it? If so, please add the error message as string resource.

aparajit-pratap (approved):

src/Libraries/CoreNodeModels/Input/BaseTypes.cs
Astul-Betizagasti: Yes it was supposed to cover that case but I just checked and it is already being checked somewhere else. I will delete those lines

src/Libraries/CoreNodeModels/Input/BaseTypes.cs
Astul-Betizagasti: It should be better now

aparajit-pratap:
@Astul-Betizagasti there are 4 test failures as a result of this change:

DynamoCoreWpfTests.RecordedTestsDSEngine.MAGN_7348_AttractorPoint
DynamoCoreWpfTests.RecordedTestsDSEngine.MAGN_7348_Count
DynamoCoreWpfTests.RecordedTestsDSEngine.MAGN_7348_MinMax
DynamoCoreWpfTests.RecordedTestsDSEngine.MAGN_7348_Transpose

I think the reason for these failures is that these tests are using range/sequence expressions in the number node and for some reason when the recorded tests are running, the numeric validation is getting skipped due to which the range and sequence expressions are passing in the number node, while they would fail in normal graph creation. I would suggest you verify that all 4 tests use expressions in number nodes and if so update the XML files used in these tests to use code block nodes instead of number nodes.

Issue #26, opened 2012-09-12
adaptive component node


tatlin:
update and finish version i started before in side branch

PR #11150, opened 2020-09-28
Drag after right click on annotation fix

zeusongit:
### Purpose

[DYN-2877](https://jira.autodesk.com/browse/DYN-2877)
Fixes the bug where dragging after a right click on an annotation detaches the annotation from the nodes it contained temporarily.

The problem discovered was on every right-click on the annotation group the selection was cleared, that removed all the nodes in the annotation space from the active selected nodes. Therefore, whenever any node in the annotation group was moved immediately after a context click it will only move the annotation box and nothing else. 
The fix was to not clear the selection on right click but only clear the selection for the required options, such as delete and ungroup. 

Before:
image-2020-07-02-12-41-01-104

After:
Kapture 2020-09-29 at 11 57 10

Related UI Automation Test: [#26](https://git.autodesk.com/Dynamo/DynamoUIAutomation/pull/26) 

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCoreWpf/Views/Core/AnnotationView.xaml.cs
+++ b/src/DynamoCoreWpf/Views/Core/AnnotationView.xaml.cs
@@ -112,6 +112,10 @@ namespace Dynamo.Nodes
         {
             if (ViewModel != null)
             {
+                DynamoSelection.Instance.ClearSelection();
+                System.Guid annotationGuid = this.ViewModel.AnnotationModel.GUID;
+                ViewModel.WorkspaceViewModel.DynamoViewModel.ExecuteCommand(
+                   new DynCmd.SelectModelCommand(annotationGuid, Keyboard.Modifiers.AsDynamoType()));
                 ViewModel.WorkspaceViewModel.DynamoViewModel.DeleteCommand.Execute(null);
             }
         }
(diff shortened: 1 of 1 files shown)


src/DynamoCoreWpf/Views/Core/AnnotationView.xaml.cs
```diff
@@ -165,10 +173,7 @@ private void AnnotationView_OnMouseLeftButtonDown(object sender, MouseButtonEven
 
         private void AnnotationView_OnMouseRightButtonDown(object sender, MouseButtonEventArgs e)
         {
-            DynamoSelection.Instance.ClearSelection();
-            System.Guid annotationGuid = this.ViewModel.AnnotationModel.GUID;
-            ViewModel.WorkspaceViewModel.DynamoViewModel.ExecuteCommand(
-               new DynCmd.SelectModelCommand(annotationGuid, Keyboard.Modifiers.AsDynamoType()));
```
QilongTang, line 171: I was just looking at the original commit by @ramramps who introduced this. Trying to think about any case we may miss by removing this. https://github.com/DynamoDS/Dynamo/commit/360fdcb9f764f2758fd7c5cb1719aaebe44d2a94

QilongTang: Thinking about this a bit, unselect on right click does seem a bit too aggressive. So looks good to me

QilongTang (approved):

src/DynamoCoreWpf/Views/Core/AnnotationView.xaml.cs
zeusongit: I thought this too, that is why I tested all the options available on the right click, and out of them 2 were making use of the unselect on right click code, so I handled each of those cases, which were delete and ungroup.

PR #11153, opened 2020-09-29
Node annotations in DocumentationBrowser

SHKnudsen:
### Purpose

This PR adds functionality to the `DocumentationBrowser` so it can show Markdown files with node annotations.

Markdown files gets associated with nodes by ~~specifying the folder where the doc are located in the *pkg json~~ placing the files in a `doc` folder in the package directory, each markdown doc should be named the same as the namespace of the node.

FileWatchers  are implemented on the annotation doc so they can be updated while loaded in Dynamo.

Node overloads is handled by specifying the input names in the markdown filename i.e. `TestPackage.TestCategory.TestCN(input1,input2)`

#### Node Annotation
CustomNodeDocumentation

NoDocumentationProvided

nodeAnnotationDocBrowser_2

#### Hot reload
nodeAnnotationDocBrowser_HotReload

### Declarations

### Reviewers

@Dewb 
@mmisol 
@QilongTang 
@aparajit-pratap 

### FYIs
@Amoursol 

For now this is implementation looks for a `doc` folder in the package to find documentation files.


src/DynamoPackages/Package.cs
```diff
@@ -98,6 +103,15 @@ public bool TypesVisibleInManager
         /// </summary>
         public IEnumerable<string> HostDependencies { get { return hostDependencies; } set { hostDependencies = value; RaisePropertyChanged("HostDependencies"); } }
 
+        private IEnumerable<string> nodeAnnotationPaths;
+
+        public IEnumerable<string> NodeAnnotationPaths
+        {
+            get { return nodeAnnotationPaths; }
+            set { nodeAnnotationPaths = value; RaisePropertyChanged(nameof(NodeAnnotationPaths)); }
+        }
```
SHKnudsen, line 112: This was to enable the user to add paths through the pkg.json. This is not being used right now, instead we are only looking for a `doc` folder in the package directory.

src/DynamoPackages/Package.cs
```diff
@@ -191,6 +205,8 @@ public static Package FromJson(string headerPath, ILogger logger)
                     SiteUrl = body.site_url,
                     RepositoryUrl = body.repository_url,
                     HostDependencies = body.host_dependencies,
+                    // hard coded for now, until we get "annotationPaths" in the package json
+                    NodeAnnotationPaths = new List<string> { @"C:\Users\SylvesterKnudsen\AppData\Roaming\Dynamo\Dynamo Core\2.9\packages\JsonData\doc" },
```
SHKnudsen, line 209: same here.

src/DocumentationBrowserViewExtension/DocumentationBrowserUtils.cs
```diff
@@ -0,0 +1,86 @@
...
+        adaptDPI() </script>";
+
+        /// <summary>
+        /// Clean up possible script tags from the content string.
+        /// </summary>
+        /// <param name="content"></param>
+        /// <returns>Returns true if any tags was removed from the content</returns>
+        internal static bool RemoveScriptTagsFromString(ref string content)
+        {
+            if (Regex.IsMatch(content, SCRIPT_TAG_REGEX, RegexOptions.IgnoreCase))
```
Dewb, line 49: Script tags aren't the only risk we need to sanitize, and we don't want to be doing our own HTML parsing -- there are a lot of edge cases. Consider a library like https://github.com/mganss/HtmlSanitizer instead.

src/DynamoPackages/PackageDocManager.cs
```diff
@@ -0,0 +1,132 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Dynamo.PackageManager
+{
+    /// <summary>
+    /// Creates a lookup for node annotation markdown files associated with a package.
+    /// Also creates file watchers for the package node annotation paths so new files can be hot reloaded in Dynamo.
+    /// </summary>
```
Dewb, line 13: Could these capabilities be moved to the `DocumentationBrowserViewExtension`? This is all triggered from the Package Manager's OnPackageLoaded handler; the doc browser extension could subscribe to that event as well. Also, this looks like we're monitoring all documentation for all packages at all times -- that seems excessive. We want to watch the minimum necessary to make hot reload work. It's okay to defer a lot of this work until the user requests help for a node.

src/DocumentationBrowserViewExtension/DocumentationBrowserUtils.cs
Dewb: More: http://htmlparsing.com/regexes.html

src/DynamoPackages/PackageDocManager.cs
SHKnudsen: @Dewb we are watching all packages `doc` folder so we know if the documentation is added or some is removed, but we are only watching the current open documentation file for changes. I cant see the way to defer that to when the user press help, or am i missing something?

src/DocumentationBrowserViewExtension/DocumentationBrowserUtils.cs
SHKnudsen: changed this to use `HtmlSanitizer` it seems like this library has a lot of configuration options, my knowledge on what might be considered dangerous in this is limited, so for know im just using the default settings.

src/DocumentationBrowserViewExtension/Docs/MarkdownStyling.html
```diff
@@ -0,0 +1,250 @@
+﻿<!-- These headers are required to force the browser embedded in the DocumentationBrowser extension to render in modern IE11 mode. -->
+<head>
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <style>
+        /* PrismJS 1.21.0
```
mjkkirschner, line 7: was this css extracted from prisim.js?

src/DocumentationBrowserViewExtension/Docs/NodeAnnotationNotFound.md
```diff
@@ -0,0 +1,3 @@
+## No further documentation provided for this node.
+ 
+For more information on how to add extended documentation to your custom nodes, have a look [here](https://github.com/DynamoDS/Dynamo/wiki).
```
mjkkirschner, line 3: is there a valid wiki entry on how to use the new functionality?

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
```diff
@@ -51,7 +55,7 @@ internal void OnMessageLogged(ILogMessage msg)
 
         public void Startup(ViewStartupParams viewStartupParams)
         {
-            // Do nothing for now
+            pmExtension = viewStartupParams.ExtensionManager.Extensions.OfType<PackageManagerExtension>().FirstOrDefault();
```
mjkkirschner, line 58: I think we should write code that interacts with extensions defensively - this code might return `null` - then later null needs to be checked for when accessing this field.

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
```diff
@@ -68,14 +72,29 @@ public void Loaded(ViewLoadedParams viewLoadedParams)
...
 
             // subscribe to the documentation open request event from Dynamo
             this.viewLoadedParamsReference.RequestOpenDocumentationLink += HandleRequestOpenDocumentationLink;
 
             // subscribe to property changes of DynamoViewModel so we can show/hide the browser on StartPage display
             (viewLoadedParams.DynamoWindow.DataContext as DynamoViewModel).PropertyChanged += HandleStartPageVisibilityChange;
+
+            // subscribe to package loaded so we can add the package documentation 
+            // to the Package documentation manager when a package is loaded
+            pmExtension.PackageLoader.PackgeLoaded += OnPackgeLoaded;
```
mjkkirschner, line 85: it's possible pmExtension is null.

src/DocumentationBrowserViewExtension/PackageDocumentationManager.cs
```diff
@@ -0,0 +1,127 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Dynamo.DocumentationBrowser
+{
+    /// <summary>
+    /// Creates a lookup for node annotation markdown files associated with a package.
+    /// Also creates file watchers for the package node annotation paths so new files can be hot reloaded in Dynamo.
+    /// </summary>
+    public class PackageDocumentationManager
```
mjkkirschner, line 14: curious why you did not use a static class?

src/DocumentationBrowserViewExtension/PackageDocumentationManager.cs
```diff
@@ -0,0 +1,127 @@
...
+            TrackDocumentationFiles(files);
+        }
+
+        private void MonitorDirectory(DirectoryInfo directoryInfo)
+        {
+            if (markdownFileWatchers.ContainsKey(directoryInfo.FullName))
+                return;
+            
+            var watcher = new FileSystemWatcher(directoryInfo.FullName, fileExtension) { EnableRaisingEvents = true };
+            watcher.Renamed += OnFileRenamed;
```
mjkkirschner, line 90: can we clean these events up when the extension is disposed?

src/DynamoPackages/Package.cs
```diff
@@ -53,6 +53,11 @@ public string ExtraDirectory
             get { return Path.Combine(RootDirectory, "extra"); }
         }
 
+        public string NodeDocumentaionDirectory
```
mjkkirschner, line 56: please add ///summary.

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModelEventArgs.cs
```diff
@@ -153,4 +159,142 @@ public OpenDocumentationLinkEventArgs(Uri link)
...
     }
+
+    /// <summary>
+    /// A node documentation request compatible with the Dynamo RequestOpenDocumentationLink event,
+    /// such as the namespace of the node used to lookup the .md documentation file
+    /// and additional Node Info gathered from the NodeModel properties.
+    /// </summary>
+    public class OpenNodeAnnotationEventArgs : OpenDocumentationLinkEventArgs
+    {
+        public string Namespace { get; }
```
mjkkirschner, line 170: please add ///summaries to public members

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModelEventArgs.cs
```diff
@@ -153,4 +159,142 @@ public OpenDocumentationLinkEventArgs(Uri link)
...
+        public string Type { get; }
+        public string Description { get; }
+        public string Category { get; }
+        public List<string> InputNames { get; private set; }
```
mjkkirschner, line 174: is there a reason these are actually `List<T>` in their public signature? If you can use IEnumerable<T> thats more flexible - especially if the underlying implementation doesn't need to be a list.

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModelEventArgs.cs
```diff
@@ -153,4 +159,142 @@ public OpenDocumentationLinkEventArgs(Uri link)
...
+
+            var inputs = nodeModel.InPorts;
+            foreach (var input in inputs)
+            {
+                InputNames.Add(input.Name);
+                InputDescriptions.Add(input.ToolTip);
+            }
+        }
+
+        static private string GetNamespace(NodeModel nodeModel, DynamoViewModel viewModel)
```
mjkkirschner, line 218: this function is called `GetNameSpace` but it seems like it returns the `namespace.FunctionName` and sometimes other names...

Maybe... GetMinimumQualifiedName? or something like that?

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModelEventArgs.cs
```diff
@@ -153,4 +159,142 @@ public OpenDocumentationLinkEventArgs(Uri link)
...
+                     
+
+                default:
+                    return string.Empty;
+            }
+        }
+
+        private static bool CustomNodeHasCollisons(string nodeName, string packageName, DynamoViewModel viewModel)
+        {
+            var pmExtension = viewModel.Model.GetPackageManagerExtension();
```
mjkkirschner, line 261: I know this function gets used in lots of places carelessly, but it's possible it will return null, can we check its value before proceeding?

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModelEventArgs.cs
```diff
@@ -153,4 +159,142 @@ public OpenDocumentationLinkEventArgs(Uri link)
...
+                .Select(x => x).ToList();
+
+            if (searchEntries.Count() > 1)
+                return true;
+            return false;
+        }
+
+        private static string GetPackageName(NodeModel node)
+        {
+            return node.Category.Split(new char[] { '.' }).FirstOrDefault();
```
mjkkirschner, line 291: This doesn't seem correct... what about if this is called on nodes that are not from a package?

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.csproj
```diff
@@ -30,16 +30,44 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="AngleSharp, Version=0.14.0.0, Culture=neutral, PublicKeyToken=e83494dcdc6d31ea, processorArchitecture=MSIL">
```
mjkkirschner, line 33: what is AngleSharp a dep of?

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.csproj
```diff
@@ -30,16 +30,44 @@
...
+    <Reference Include="Markdig, Version=0.21.1.0, Culture=neutral, processorArchitecture=MSIL">
+      <HintPath>..\packages\Markdig.0.21.1\lib\net452\Markdig.dll</HintPath>
+    </Reference>
     <Reference Include="Microsoft.Practices.Prism">
       <HintPath>..\..\extern\prism\Microsoft.Practices.Prism.dll</HintPath>
     </Reference>
     <Reference Include="PresentationCore" />
     <Reference Include="PresentationFramework" />
     <Reference Include="System" />
+    <Reference Include="System.Buffers, Version=4.0.3.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51, processorArchitecture=MSIL">
```
mjkkirschner, line 51: what is bringing in all of these system packages? Why are they such old versions?

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
```diff
@@ -63,10 +39,21 @@ public Uri Link
             get { return this.link; }
             private set
             {
+                var oldLink = link;
+                // if the link is changed we unsubscribe the current watcher and create a new
```
mjkkirschner, line 43: create a new watcher?

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.csproj
```diff
@@ -30,16 +30,44 @@
...
+      <HintPath>..\packages\AngleSharp.0.14.0\lib\net472\AngleSharp.dll</HintPath>
+    </Reference>
+    <Reference Include="AngleSharp.Css, Version=0.14.2.0, Culture=neutral, PublicKeyToken=e83494dcdc6d31ea, processorArchitecture=MSIL">
+      <HintPath>..\packages\AngleSharp.Css.0.14.2\lib\net472\AngleSharp.Css.dll</HintPath>
+    </Reference>
+    <Reference Include="HtmlSanitizer, Version=5.0.0.0, Culture=neutral, PublicKeyToken=61c49a1a9e79cc28, processorArchitecture=MSIL">
+      <HintPath>..\packages\HtmlSanitizer.5.0.353\lib\net46\HtmlSanitizer.dll</HintPath>
+    </Reference>
+    <Reference Include="Markdig, Version=0.21.1.0, Culture=neutral, processorArchitecture=MSIL">
+      <HintPath>..\packages\Markdig.0.21.1\lib\net452\Markdig.dll</HintPath>
```
mjkkirschner, line 43: ⚠️ 
I think to use markdig we need to update our license file and about box text:
https://github.com/lunet-io/markdig/blob/master/license.txt

bsd-2 requires atrribution.

mjkkirschner:
@SHKnudsen perhaps you could provide a small diagram that illustrates the logic of the file watchers, I'm trying to understand where / when they are created. It also seems there are file watchers for specific files created, and additionally watchers created for entire directories?

src/DocumentationBrowserViewExtension/DocumentationBrowserUtils.cs
mjkkirschner: this brings in a bunch of dependencies - @QilongTang @mmisol any thoughts?

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.csproj
mjkkirschner: I see, it comes in with `HTMLSanitizer`

src/DocumentationBrowserViewExtension/DocumentationBrowserUtils.cs
QilongTang: hmm actually I would rather not bring in more dependencies, is this a must have for locally hosted static web content? @Dewb

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModelEventArgs.cs
SHKnudsen: Is that actually possible? I use this to check if a custom node has any nodes with the same name in its package. Im simply extracting the first part of the category, i was under the impression that all custom nodes requires a category.

Maybe im wrong here?

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.csproj
SHKnudsen: Not sure why all of these was added, removed them and everything still functions as expected.

src/DocumentationBrowserViewExtension/Docs/MarkdownStyling.html
SHKnudsen: yes

src/DocumentationBrowserViewExtension/Docs/NodeAnnotationNotFound.md
SHKnudsen: its in progress, will have to update this link when it gets uploaded

SHKnudsen:
@mjkkirschner does this make it clearer? 

FileWatcherLogic

src/DocumentationBrowserViewExtension/PackageDocumentationManager.cs
SHKnudsen: hm, yea good question, i guess i had other ideas at the start. Could certainly make this into a static class instead if you prefer that?

mjkkirschner: nope, just curious.

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModelEventArgs.cs
mjkkirschner: My point is that the name of this function might be confusing since it's called get package name, but thats not what it does.

src/DocumentationBrowserViewExtension/DocumentationBrowserUtils.cs
mjkkirschner: @SHKnudsen - could you look into potential html sanitization options that have fewer dependencies?

mjkkirschner:
@SHKnudsen looks like some tests are failing with:
```
MESSAGE:
System.IO.FileNotFoundException : Could not load file or assembly 'System.Text.Encoding.CodePages, Version=4.1.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.
+++++++++++++++++++
STACK TRACE:
   at AngleSharp.Browser.EncodingMetaHandler..ctor()
   at AngleSharp.Configuration..ctor(IEnumerable`1 services)
   at Ganss.XSS.HtmlSanitizer.CreateParser()
   at Ganss.XSS.HtmlSanitizer.SanitizeDom(String html, String baseUrl)
   at Ganss.XSS.HtmlSanitizer.Sanitize(String html, String baseUrl, IMarkupFormatter outputFormatter)
   at Dynamo.DocumentationBrowser.DocumentationBrowserUtils.RemoveScriptTagsFromString(String& content) in c:\WorkspaceDynamo\src\DocumentationBrowserViewExtension\DocumentationBrowserUtils.cs:line 52
   at Dynamo.DocumentationBrowser.DocumentationBrowserViewModel.LoadContentFromResources(String name, Boolean injectDPI, Boolean removeScriptTags) in c:\WorkspaceDynamo\src\DocumentationBrowserViewExtension\DocumentationBrowserViewModel.cs:line 379
   at Dynamo.DocumentationBrowser.DocumentationBrowserViewModel.NavigateToContentMissingPage() in c:\WorkspaceDynamo\src\DocumentationBrowserViewExtension\DocumentationBrowserViewModel.cs:line 275
   at Dynamo.DocumentationBrowser.DocumentationBrowserViewModel.HandleLocalResource(OpenDocumentationLinkEventArgs e) in c:\WorkspaceDynamo\src\DocumentationBrowserViewExtension\DocumentationBrowserViewModel.cs:line 186
   at Dynamo.DocumentationBrowser.DocumentationBrowserViewModel.HandleOpenDocumentationLinkEvent(OpenDocumentationLinkEventArgs e) in c:\WorkspaceDynamo\src\DocumentationBrowserViewExtension\DocumentationBrowserViewModel.cs:line 140
   at DynamoCoreWpfTests.DocumentationBrowserViewExtensionTests.DisplaysInvariantContentWhenNoCompatibleLocalizedContentIsAvailable() in c:\WorkspaceDynamo\test\DynamoCoreWpfTests\ViewExtensions\DocumentationBrowserViewExtensionTests.cs:line 337
```

I wonder if instead of fixing these right now, we could investigate alternative html parsing/sanitization libraries with fewer dependencies... I'd like to know what are options are.

src/DocumentationBrowserViewExtension/DocumentationBrowserUtils.cs
SHKnudsen: @mjkkirschner i looked into this and it seems like there arent a lot of good options, here is what i found so far:

[**AntiXss**](https://docs.microsoft.com/en-us/dotnet/api/system.web.security.antixss.antixssencoder?view=netframework-4.8) - looks like this hasent been update since 2014, also read a few negative articles about it. (https://www.nuget.org/packages/AntiXSS/)

[**HtmlRuleSanitizer**](https://github.com/Vereyon/HtmlRuleSanitizer) - This seems to be the most used option after *HtmlSanitizer*, however this still has dependecies on an Html Parser, in this case that is [HtmlAgilityPack](https://www.nuget.org/packages/HtmlAgilityPack/). So this would probably bring in about the same dependencies as **HtmlSanitizer** 

[**HtmlSanitizer**](https://github.com/mganss/HtmlSanitizer) - A @Dewb suggested, this looks to be the most commonly used library. It depends on [AngelSharp](https://github.com/AngleSharp/AngleSharp) which also looks like its the most commonly use Html parser.

In addition to these three the only real option i found was to [build our own sanitizer](https://weblog.west-wind.com/posts/2012/Jul/19/NET-HTML-Sanitation-for-rich-HTML-Input), however this would probably still bring in a dependency on a HtmlParser.

I dont have a lot of experience on this area tbh, so not entirely sure what we actually need the sanitation to do except for disabling `<script>` tags.

mjkkirschner: thanks for looking into this @SHKnudsen - it looks like in the latest releases of HtmlSanitizer the `system.encoding.codepages dependency` was removed - perhaps we could use that newest release?

SHKnudsen: @mjkkirschner looks like its still there:

**HtmlSanitizer Dependencies**
SanitizerDep

**Markdig Dependencies**
MarkdigDep

mjkkirschner:
@SHKnudsen @Dewb @mmisol @QilongTang ugh - this does not inspire confidence:
https://github.com/mganss/HtmlSanitizer/issues/241

This was the commit/issue I was referring to:

It seems the build of htmlsanitizer was broken because the devs there added a dependency on `System.Text.Encoding.CodePages.dll` - but `angleSharp` which is a dependency of `htmlsanitizer` requires a version < 4.5.

mmisol:
@mjkkirschner Feels like I got to the discussion a little late. I thought we had discussed this when the Documentation Browser was added and decided to go for a less bullet-proof approach that did not depend on adding dependencies, because of all the possible conflicts this could bring with integrators, add-ins, and so on. Did we decide to reconsider?
I guess I'm not against it if we can be more or less certain this is not going to bring problems to integrators. In any case, we should check if that one is our best bet or if there are other options that are more up-to-date.

mjkkirschner:
Hi @mmisol -
@Dewb made the following point:
> Script tags aren't the only risk we need to sanitize, and we don't want to be doing our own HTML parsing -- there are a lot of edge cases. Consider a library like https://github.com/mganss/HtmlSanitizer instead.

I'm personally quite wary of the additional dependencies... so looking for discussion / opinions on the tradeoffs and understanding from @Dewb if a **proper** scrubbing of html is critical here since we are allowing user input as content.

mmisol:
Do you mean input because the files are included loosely with the package? Does the concern disappear if we switch to embedded resources? In any case, it might be better to organize a short meeting to discuss this.

mjkkirschner:
@mmisol - I don't think the concern disappears with embedded resources - though I guess it is a slightly more difficult vector to compromise.

I will organize a meeting.

mjkkirschner:
Hi @SHKnudsen - we are identifying how we would like to move forward with this, but it would really help us as we evaluate alternatives if:
1. this PR had associated test cases that we could run to see if our changes produce the same results
2. the existing tests ran okay (I think the current failures might just be a fluke - they are related to changing asm versions) - integrating latest master may solve this.

mjkkirschner:
@SHKnudsen - BTW, latest test run shows only two failures:

```
Stacktrace
MESSAGE:
  Expected: True
  But was:  False

+++++++++++++++++++
STACK TRACE:
   at DynamoCoreWpfTests.DocumentationBrowserViewExtensionTests.DPIScriptExists() in c:\WorkspaceDynamo\test\DynamoCoreWpfTests\ViewExtensions\DocumentationBrowserViewExtensionTests.cs:line 399
```

and 

```
Stacktrace
MESSAGE:
  Expected: True
  But was:  False

+++++++++++++++++++
STACK TRACE:
   at DynamoCoreWpfTests.DocumentationBrowserViewExtensionTests.RemovesScriptTagsFromLoadedHtml() in c:\WorkspaceDynamo\test\DynamoCoreWpfTests\ViewExtensions\DocumentationBrowserViewExtensionTests.cs:line 374
```

SHKnudsen:
Thanks @mjkkirschner, will look into this and add new test cases tomorrow.

SHKnudsen:
Hi @mjkkirschner got a bit caught up today but just pushed a fix for the failing tests and a few new tests that checks the content of node documentation.

If there are some specific tests cases that will help you guys move forward, let me know and i will have a look at it Tuesday.

mjkkirschner:
FYI @sm6srw 

> Hi @mjkkirschner got a bit caught up today but just pushed a fix for the failing tests and a few new tests that checks the content of node documentation. If there are some specific tests cases that will help you guys move forward, let me know and i will have a look at it Tuesday.

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
QilongTang: Seems we also forgot to unsubscribe this during disposing

mjkkirschner:
Hi @SHKnudsen - we are going to look at isolating the markdown and html conversion / sanitization as a separate process, and potentially make commits to this branch, or another based on it.

Have you had time to:
1. look at the failing test: https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/1671/testReport/junit/DynamoCoreWpfTests/DocumentationBrowserViewExtensionTests/CanCreateNodeDocumenationHtmlFromNodeAnnotationEventArgsWithPackageNodeWithAddtionalDocumentaion/

2. write up / link the wiki entry on using the node annotation docs?
`For more information on how to add extended documentation to your custom nodes, have a look [here](https://github.com/DynamoDS/Dynamo/wiki).`

FYI @jasonstratton @sm6srw @QilongTang

SHKnudsen:
Hi @mjkkirschner,

Didnt know there was a failing test, will have a look now.
I posted the wiki documentation on slack, will tag you there so you can have a look.

sm6srw:
@SHKnudsen Sorry, I pushed to the wrong branch. I will revert my changes. They where supposed to go into my fork!

src/DocumentationBrowserViewExtension/NodeDocumentationHtmlGenerator.cs
```diff
@@ -0,0 +1,86 @@
...
+            sb.AppendLine(string.Format("<p><i>{0}</i></p>", e.MinimumQualifiedName));
+            sb.AppendLine("<hr>");
+
+            return sb.ToString();
+        }
+
+        private static string CreateNodeInfo(OpenNodeAnnotationEventArgs e)
+        {
+            StringBuilder sb = new StringBuilder();
+            sb.AppendLine("<h2>Node Info</h2>");
```
sm6srw, line 48: @mjkkirschner Shouldn't these strings be resources?

mjkkirschner: yes - but AFAIK we don't currently have a great way to localize web view content - I guess we could put tokens here in the html, and before displaying them, replace those with data from a resource dictionary in the extension itself?

mjkkirschner: can we file this as an improvement?

mmisol: @mjkkirschner We have localized HTML documents for documentation browser so I guess the user expectation is that this content is localized too. Do you feel it's a big amount of work to externalize the strings here to resources?

mjkkirschner: actually - I recall that @mmisol did some work to enable localization - I think the team may localize the html files directly.

mjkkirschner: hah, thanks for chiming in Martin, can you remind me what the process is for html files currently?

mmisol: On our side, we should make sure HTML files go as embedded resources. That makes it a lot easier for the localization team. However, in this case it is just a string so it should go as a string resource.

sm6srw: Thanks!

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
```diff
@@ -186,6 +199,59 @@ private void HandleLocalResource(OpenDocumentationLinkEventArgs e)
...
+                return;
+
+            var nodeAnnotationArgs = openDocumentationLinkEventArgs as OpenNodeAnnotationEventArgs;
+            this.content = CreateNodeAnnotationContent(nodeAnnotationArgs);
+            this.Link = new Uri(e.FullPath);
+        }
+
+        private string CreateNodeAnnotationContent(OpenNodeAnnotationEventArgs e)
+        {
+            var writer = new StringWriter();
```
mjkkirschner, line 243: @sm6srw @SHKnudsen this writer implements IDisposable, it should probably be wrapped in a using, can it be fixed?

sm6srw: Sorry, missed this. I will take a look.

mjkkirschner:
@sm6srw should we move this out of draft PR state?

sm6srw:
@mjkkirschner Yes, I did that now.

PR #11163, opened 2020-10-05
Reorder Helix elements for Isolate Geometry Mode to display isolated geometry

aparajit-pratap:
### Purpose

https://jira.autodesk.com/browse/DYN-3130

This attempts to fix a regression with Isolate Selected Geometry feature where point-line geometry when selected does not appear when hidden by mesh geometry (solids and surfaces). The reason for this is that the geometry needs to be reordered so that selected geometries will have higher precedence and rendered closer to the camera compared to unselected geometries.

Caveat: This comes with some performance penalty when rendering selected geometry in isolated mode. 
isolate_pointline_geom

### Declarations

### FYIs


src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -523,6 +523,28 @@ protected HelixWatch3DViewModel(NodeModel model, Watch3DViewModelStartupParams p
...
+            if (Camera != null)
+            {
+                values.Sort(new Element3DComparer(Camera.Position));
+            }
+
+            if (sceneItems == null)
+                sceneItems = new ObservableElement3DCollection();
+
+            sceneItems.Clear();
+            sceneItems.AddRange(values);
```
aparajit-pratap, line 547: This logic has been lifted off of `UpdateSceneItems`, however, `UpdateSceneItems` seems to be doing some extra work like cycling through the list of elements and updating vertices, triangles, and colors, which I'm not sure if that is necessary in this case. Skipping that saves some time.

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -523,6 +523,28 @@ protected HelixWatch3DViewModel(NodeModel model, Watch3DViewModelStartupParams p
 
             SetupScene();
             InitializeHelix();
+            DynamoSelection.Instance.Selection.CollectionChanged += OnSelectionCollectionChanged;
```
mjkkirschner, line 526: I think we should find an appropriate spot to unsubscribe this handler.

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -523,6 +523,28 @@ protected HelixWatch3DViewModel(NodeModel model, Watch3DViewModelStartupParams p
...
+            if (Element3DDictionary == null)
+            {
+                sceneItems = new ObservableElement3DCollection();
+                return;
+            }
+
+            var values = Element3DDictionary.Values.ToList();
+            if (Camera != null)
+            {
+                values.Sort(new Element3DComparer(Camera.Position));
```
mjkkirschner, line 540: Why does this need to happen every time the selection changes? Couldn't it be done once when the isolation mode is enabled?

aparajit-pratap: Tried that. Doesn't work for some reason.

mjkkirschner:
nice! @aparajit-pratap 

would you also mind adding an image comparison test for this so it does not regress again? I added a few for the mesh mesh case in the last pr.

https://github.com/DynamoDS/Dynamo/commit/89818c3e9b1dd8a019980546e4c13793b9229c78

mjkkirschner (approved):

PR #11164, opened 2020-10-06
Node AutoLayout

QilongTang:
Please Note:

### Purpose

Auto layout the newly created node from node auto complete.

Instead of the existing auto layout algorithm we are using for clean up layout which makes the initial node jumpy, I am simply putting the new node right next to the input port and arrange some Y offset based on inputport index. 

AutoLayout_v3

### Declarations

### Reviewers

@DynamoDS/dynamo 

### FYIs


src/DynamoCoreWpf/ViewModels/Search/SearchViewModel.cs
```diff
@@ -1061,11 +1062,20 @@ public void OnSearchElementClicked(NodeModel nodeModel, Point position)
...
         {
+            // Do not auto connect code block node since default code block node do not have output port
             if (!nodeCreationName.Contains("Code Block"))
             {
+                //dynamoViewModel.ExecuteCommand(new DynamoModel.MakeConnectionCommand(nodeModel.GUID,))
+                var id = Guid.NewGuid();
                 // Create a new node based on node creation name and connect ports
-                dynamoViewModel.ExecuteCommand(new DynamoModel.CreateAndConnectNodeCommand(Guid.NewGuid(), portModel.Owner.GUID,
-                    nodeCreationName, 0, portModel.Index, portModel.CenterX - 50, portModel.CenterY + 200, false, false));
+                var initialNode = dynamoViewModel.CurrentSpaceViewModel.Nodes.Where(x => x.Id == portModel.Owner.GUID).FirstOrDefault();
```
aparajit-pratap, line 1071: isn't `initialNode` the same as `portModel.Owner`?

aparajit-pratap: Can we split this step into a `CreateNodeCommand` step and then add a new command called something like `PlaceAndConnectCommand`? That way we can first create the new node, get its height and width and place it relative to the original node according to its respective height and width. In other words the `PlaceAndConnectCommand` will be a combination of 2 steps: 1) place the node based on an calculated X and Y position and 2) call the `MakeConnectionCommand`.

aparajit-pratap: This way we might be able to do away with these hard-coded values.

src/DynamoCoreWpf/ViewModels/Search/SearchViewModel.cs
```diff
@@ -1061,11 +1062,20 @@ public void OnSearchElementClicked(NodeModel nodeModel, Point position)
...
+                // Placing the new node based on which input port it is connecting to.
+                // The larger index for the input port, the lower the new node will be placed. 60 and 200 and offset we can adjust.
+                double adjustedY = initialNode.Y + (portModel.Index - 1) * 60 ;
+                // Placing the new node to the left of initial node
+                var adjustedX = initialNode.X - 200; 
+                dynamoViewModel.ExecuteCommand(new DynamoModel.CreateAndConnectNodeCommand(id, portModel.Owner.GUID,
+                    nodeCreationName, 0, portModel.Index, adjustedX, adjustedY, false, false));
```
aparajit-pratap, line 1078: Can we split this step into a CreateNodeCommand step and then add a new command called something like ``PlaceAndConnectCommand``? That way we can first create the new node, get its height and width and use the height and width to place it relative to the original node. In other words, the `PlaceAndConnectCommand`
will be a combination of 2 steps: 1) place the node based on a calculated X and Y position, and 2) call the
`MakeConnectionCommand`. This way we might be able to do away with these hard-coded values.

QilongTang: @aparajit-pratap I did thought about that before implementation but the height and width for the new nodes are always default to 100 for node model..

QilongTang: image
@aparajit-pratap This is what I am referring to. I already considered this solution before finalizing the PR. For now there isn't a way I can see we can get the actual height and width for the newly placed node.

QilongTang: Another slightly less important reason is about undo/redo, the current implementation only requires one undo to get back to the previous state versus the split way will need to be done by two undos.

aparajit-pratap: I'm trying something else, will get back to you on this PR, I may commit directly to this PR if I can get something working.

aparajit-pratap: Couldn't get it to work. In any case this is the problem with this approach right now:
create_and_connect_nodes

QilongTang: Indeed, placing any slider or any giant node will result in such problem but I dont see a way to be smarter yet..

aparajit-pratap (commented):
@QilongTang I think we can close this :)

PR #11169, opened 2020-10-07
Fix input port type for instance properties

aparajit-pratap:
### Purpose

Fix `GetInputPortType` on `PortModel` to handle nodes that are properties.

### Declarations

### DIFF
--- a/src/DynamoCore/Graph/Nodes/PortModel.cs
+++ b/src/DynamoCore/Graph/Nodes/PortModel.cs
@@ -408,7 +408,7 @@ namespace Dynamo.Graph.Nodes
                 string type;
                 // In the case of a node for an instance method, the first port
                 // type is the declaring class type of the method itself.
-                if (fd.Type == FunctionType.InstanceMethod)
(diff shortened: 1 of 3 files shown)


src/DynamoCore/Graph/Nodes/PortModel.cs
```diff
@@ -408,7 +408,7 @@ internal string GetInputPortType()
                 string type;
                 // In the case of a node for an instance method, the first port
                 // type is the declaring class type of the method itself.
-                if (fd.Type == FunctionType.InstanceMethod)
+                if (fd.Type == FunctionType.InstanceMethod || fd.Type == FunctionType.InstanceProperty)
```
QilongTang, line 411: @aparajit-pratap I guess I am not too familiar with `FunctionType` enum, do we need to consider other options here?

aparajit-pratap: The only other option I need to check is `Constructor`. Let me check.

aparajit-pratap: I have verified that this works well with constructors too.

QilongTang (approved):

PR #11176, opened 2020-10-12
Expand matching nodes in node autocomplete results list using inheritance.

mjkkirschner:
**UPDATED:**
Oct-19-2020 14-36-41

UPDATED:

### Purpose

* This PR adds some methods to the `NodeAutoCompleteSearchViewModel` _I'm not sure that is the right place for these, but I didn't find a derived model of `NodeSearchModel` - it looks like it's just using the base class one... perhaps moving some logic and data into a derived model class and out of the viewModel makes sense... Maybe thats another PR._

* Both `DerivesFrom` and  `GetTypeDistance` use the `GetClassHierarchy` method on `ClassMirror` - this returns the base classes of a class - but it does not include the current class in the result.

* adds an additional getter to `ClassMirror` so that classMirrors can be compared based on the ID of the class in the classTable - this let's us compare two class mirrors for the same underlying type even if we created the class mirrors with different names - ie `Autodesk.DesignScript.Geometry.Geometry` and `Geometry` - the vm will find both of these and create separate `ClassMirror`(s) for them, even though they are backed by the same type - knowing that they are the same is very useful. 👍 .

* adds an internal getter to `ZTSearchElement` so the underlying functionDescriptor can be retrieved - this is so we can scan the loaded searchElements and then have access to the full information about the function descriptor. (like fully qualified return type)

* when we encounter a type that includes a rank we strip off the rank by parsing the `typedIdentifer` and returning only the typeName which does not include the rank. This let's us match nodes that accept arrays with nodes that return single items and vice versa. This makes sense to me because of replication and how it is easy it is for a user to create a list of items from a node that returns single items.

Logic is as follows:

1. when scanning ZT nodes, if the `type name of the return type matches the input type`  OR if the `return type derives from the input type`. then add to the result list - Also match the typename without rank.
2. sort the results using the distance between the two types (determined by the class hierarchy)
3. a second sort is used to group all the entries into constructors, actions, and queries - so all constructors are listed first, ranked by their type distance etc etc. see updated video.

TODO:
- [x] sort is implemented on `NodeSearchElements` but only deals with nodes with a single return value - fix this.
- [ ] consider cacheing of classmirrors, we recreate them over and over for the same types.
- [ ] considering cacheing type distance, we recompute it over and over for the same types.
- [x] tests.
- [x] I think there might be a leak in the viewModels for each search entry. (existing leak)

### Declarations


src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -75,9 +84,16 @@ internal IEnumerable<NodeSearchElement> GetMatchingNodes()
 
             foreach (var descriptor in functionDescriptors)
             {
-                if (descriptor.ReturnType.ToString() == inputPortType)
+                if ((descriptor.ReturnType.ToString() == inputPortType) || DerivesFrom(inputPortType, descriptor.ReturnType.ToString(), dynamoViewModel.Model.EngineController.LiveRunnerCore))
                 {
-                    elements.Add(new ZeroTouchSearchElement(descriptor));
+                    //TODO it's strange we create a new search element instead of using the existing ones in the model...
+                    var entry = new ZeroTouchSearchElement(descriptor);
```
mjkkirschner, line 90: @QilongTang @aparajit-pratap I'm curious if there is a reason we create a new ZT searchElement instead of accessing the model and using the existing ZTSearchElements, this has some benefits, one of them being it's more consistent with the next few lines where NodeModelSearchElements are extracted from the Model.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -37,12 +44,13 @@ internal void InitializeDefaultAutoCompleteCandidates()
...
-            if(PortViewModel == null) return;
+            if (PortViewModel == null) return;
 
             var searchElements = GetMatchingNodes();
             FilteredResults = searchElements.Select(e =>
             {
-                var vm  = new NodeSearchElementViewModel(e, this);
+                var vm = new NodeSearchElementViewModel(e, this);
+                //TODO I think this leaks.
                 vm.RequestBitmapSource += SearchViewModelRequestBitmapSource;
```
aparajit-pratap, line 54: Ok, I think its other usage also leaks. I think you can cycle over the `FilteredResults` property of the `SearchViewModel` in its Dispose method and unsubscribe from its `SearchViewModelRequestBitmapSource` method for each `NodeSearchElementViewModel`.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
aparajit-pratap: I'm not sure if it will be more work to search for an existing `ZTSearchElement` with the given descriptor as opposed to instantiating a new one.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -75,9 +84,16 @@ internal IEnumerable<NodeSearchElement> GetMatchingNodes()
 
             foreach (var descriptor in functionDescriptors)
             {
-                if (descriptor.ReturnType.ToString() == inputPortType)
+                if ((descriptor.ReturnType.ToString() == inputPortType) || DerivesFrom(inputPortType, descriptor.ReturnType.ToString(), dynamoViewModel.Model.EngineController.LiveRunnerCore))
```
aparajit-pratap, line 87: Is the `LiveRunnerCore` available in manual mode?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -87,8 +103,138 @@ internal IEnumerable<NodeSearchElement> GetMatchingNodes()
...
+            try
+            {
+                //TODO mirrors can be cached until new types are imported...
+                var mirror1 = new ClassMirror(typea, core);
+                var mirror2 = new ClassMirror(typeb, core);
 
-            return elements;
+
+                //TODO as we do this check we can cache the type distance...
+                if (mirror2.GetClassHierarchy().Any(x => x.ClassName == mirror1.ClassName))
```
QilongTang, line 128: @mjkkirschner @aparajit-pratap Can you explain what is type distance for me? I interpretate as levels of inheritance?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -87,8 +103,138 @@ internal IEnumerable<NodeSearchElement> GetMatchingNodes()
...
+            {
+                // TODO need to deal with the fact that type[] and type[]..[] are not classes - trying to create these up as ClassMirrors will yield an exception.
+                Console.WriteLine($"failed to create class mirror for {typea} or {typeb} during node autocomplete operation ");
+                return false;
+            }
         }
+
+
+        // TODO should we sort at the functiondescriptor stage?
+        private int CompareNodeSearchElementByTypeDistance(NodeSearchElement x, NodeSearchElement y)
```
QilongTang, line 145: @mjkkirschner Notice all of these are done at the click time, team also talked about doing these `grooming` at Dynamo startup level so we offset these operations. An interesting scale problem IMHO.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
mjkkirschner: yes I believe so, the `liveRunnerServices` and `core` constructed when the engineController is constructed - that should definitely be around in manual mode.

also, my manual testing performs fine in manual mode.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
mjkkirschner: yes, distance between classes in the inheritance chain.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
mjkkirschner: well, we don't need to search for anything if we use the loaded searchElements directly from the model instead of using function descriptors.

If that was the reason then - I'll make a small change to `ZTSearchElement` to expose the fully qualified output parameter name and I think we can unify this.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
aparajit-pratap: Code block node autocomplete uses the `LibraryManagementCore` on `LibraryServices`, which is the static core. I've noticed that we have existing issues due to these cores not matching up exactly but that is a separate VM discussion I guess. I'm okay with using the LiveRunnerCore, just that we should be cognizant of such issues.

mjkkirschner: hmm, let me take a look at using that core if you think it will be more stable? Which one is more "up to date" ?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
aparajit-pratap: Sure, that sounds like a good suggestion. I can't remember why I chose to use functiondescriptor but now that you point it out, it might just be easier to deal with search elements from the model.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
mjkkirschner: unfortunately, as far as I can tell, we only call dispose on the searchViewModel when the workspaceViewModel is disposed, when means that FilteredResults will have been set and reset many times - so we'll only end up calling dispose on the last set of `FilteredResults`, I think this is insufficient.

One day we really need to look into weak references for events...
https://docs.microsoft.com/en-us/dotnet/desktop/wpf/advanced/weak-event-patterns?view=netframeworkdesktop-4.8

I think one approach is to dispose the `FilteredResults` when the view is closed.... I think there might be some subtle differences in allocation between the in canvas search and the node autocomplete though...

I'll tread carefully here, I don't want to effect in canvas search in this PR, we can always handle investigating any leaks as a separate task if that works?

aparajit-pratap: How about making `SearchViewModel.SearchViewModelRequestBitmapSource` internal and then doing the following in the `NodeSearchElementViewModel.Dispose` method:
```
if (RequestBitmapSource != null)
{
                    RequestBitmapSource -= searchViewModel.SearchViewModelRequestBitmapSource;
 }
```

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -73,22 +63,178 @@ internal IEnumerable<NodeSearchElement> GetMatchingNodes()
...
+
+            public int Compare(NodeSearchElement x, NodeSearchElement y)
+            {
+               return CompareNodeSearchElementByTypeDistance(x, y, typeNameToCompareTo, core);
+            }
+
+            private int CompareNodeSearchElementByTypeDistance(NodeSearchElement x, NodeSearchElement y, string typeNameToCompareTo, ProtoCore.Core core)
+            {
+                /*
+                 * -1	x proceeds y.
```
aparajit-pratap, line 146: x precedes y

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -73,22 +63,178 @@ internal IEnumerable<NodeSearchElement> GetMatchingNodes()
...
+
+            foreach (var ztSearchElement in ztSearchElements)
             {
-                if (descriptor.ReturnType.ToString() == inputPortType)
+                var descriptor = ztSearchElement.Descriptor;
+                if ((descriptor.ReturnType.ToString() == inputPortType)
+                    || DerivesFrom(inputPortType, descriptor.ReturnType.ToString(), dynamoViewModel.Model.LibraryServices.LibraryManagementCore))
                 {
-                    elements.Add(new ZeroTouchSearchElement(descriptor));
+                    elements.Add(ztSearchElement);
```
aparajit-pratap, line 74: 👍

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -2557,7 +2557,7 @@ internal void HideUnhideNamespace(bool hide, string library, string namespc)
             }
         }
 
-        private void AddZeroTouchNodesToSearch(IEnumerable<FunctionGroup> functionGroups)
+        internal void AddZeroTouchNodesToSearch(IEnumerable<FunctionGroup> functionGroups)
```
mjkkirschner, line 2560: made internal so it can be called from certain tests to fill the searchModel with ZT nodes.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -53,42 +42,240 @@ internal void PopulateAutoCompleteCandidates()
...
+            //if parsing the type failed, revert to original string.
+            inputPortType = ast != null ? ast.datatype.Name : inputPortType;
 
-            foreach (var descriptor in functionDescriptors)
+            //gather all ztsearchelements that are visible in search and filter using inputPortType and zt return type name.
+            var ztSearchElements = Model.SearchEntries.OfType<ZeroTouchSearchElement>().Where(x => x.IsVisibleInSearch);
+            foreach (var ztSearchElement in ztSearchElements)
             {
-                if (descriptor.ReturnType.ToString() == inputPortType)
+                //for now, remove rank from descriptors
```
QilongTang, line 67: The reason is we using our own rank defined？

mjkkirschner: please see this note in the description:

`when we encounter a type that includes a rank we strip off the rank by parsing the typedIdentifer and returning only the typeName which does not include the rank. This let's us match nodes that accept arrays with nodes that return single items and vice versa. This makes sense to me because of replication and how it is easy it is for a user to create a list of items from a node that returns single items.`

I'm not sure what you mean by "our own rank"

this PR is currently disregarding rank completely.

test/DynamoCoreWpfTests/NodeAutoCompleteSearchTests.cs
```diff
@@ -110,7 +116,12 @@ public void NodeSuggestions_InputPortBuiltInNode_AreCorrect()
                 "FFITarget.FFITarget.SecondNamespace.ClassWithNameConflict.PropertyD",
                 "FFITarget.FFITarget.SecondNamespace.ClassWithNameConflict.PropertyE",
                 "FFITarget.FFITarget.SecondNamespace.ClassWithNameConflict.PropertyF",
-                "FFITarget.FFITarget.TestData.GetStringValue"
+                "FFITarget.FFITarget.TestData.GetStringValue",
```
mjkkirschner, line 119: these additional methods now match because they return `string[]`

src/DynamoCore/Search/SearchElements/ZeroTouchSearchElement.cs
```diff
@@ -27,7 +27,11 @@ public class ZeroTouchSearchElement : NodeSearchElement
         ///     The full name of entry which consists of assembly name and qualified name for function descriptor.
         /// </summary>
         public override string FullName { get { return fullname; } }
-        
+        /// <summary>
+        /// Retrieve undelrying Function Descriptor.
```
aparajit-pratap, line 31: typo: underlying

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -1,47 +1,37 @@
...
         {
-            if(PortViewModel == null) return;
+            if (PortViewModel == null) return;
 
             var searchElements = GetMatchingNodes();
             FilteredResults = searchElements.Select(e =>
             {
-                var vm  = new NodeSearchElementViewModel(e, this);
+                var vm = new NodeSearchElementViewModel(e, this);
+                //TODO I think this leaks.
```
aparajit-pratap, line 34: I think we can remove this TODO as I added the proposed change below to unsubscribe the event.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -43,53 +42,56 @@ internal void PopulateAutoCompleteCandidates()
...
-            // Builtin functions and zero-touch functions
-            // Multi-return ports for these nodes do not contain type information
-            // and are therefore skipped.
-            var functionGroups = libraryServices.GetAllFunctionGroups();
-            var functionDescriptors = functionGroups.SelectMany(fg => fg.Functions).Where(fd => fd.IsVisibleInLibrary);
 
-            var ztSearchElements = Model.SearchEntries.OfType<ZeroTouchSearchElement>();
+            if (inputPortType == null) return elements;
+            //if inputPortType is an array, use just the typename
+            var cbn = ParserUtils.ParseWithCore($"dummyName:{ inputPortType};", dynamoViewModel.Model.LibraryServices.LibraryManagementCore);
```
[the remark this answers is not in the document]
aparajit-pratap: You can call it `parseResult` or something similar.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -52,42 +42,240 @@ internal void PopulateAutoCompleteCandidates()
...
+                    //if inputPortType is an array, lets just use the class name to match 
+                    xTypeNames = xTypeNames.Select(type => ParserUtils.ParseWithCore($"dummyName:{ type};", core).CodeBlockNode.Children().ElementAt(0).Name);
+                }
+                if (y is ZeroTouchSearchElement yzt)
+                {
+                    yTypeNames = new string[] { yzt.Descriptor.ReturnType.Name };
+                }
+                else
+                {
+                    yTypeNames = yTypeNames.Select(type => ParserUtils.ParseWithCore($"dummyName:{ type};", core).CodeBlockNode.Children().ElementAt(0).Name);
```
aparajit-pratap, line 188: Is the `else` condition for NodeModel nodes?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -52,42 +42,240 @@ internal void PopulateAutoCompleteCandidates()
...
+
+                }
+
+                if (xTypeNames.SequenceEqual(yTypeNames))
+                {
+                    return 0;
+                }
```
aparajit-pratap, line 195: When we say multiple x types and y types, are we future-proofing for multi-output nodes?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -52,42 +42,240 @@ internal void PopulateAutoCompleteCandidates()
...
+                {
+                    return 1;
+                }
+                // null is further away, so y is at the end of the list.
+                if (yTypeNames.SequenceEqual(none))
+                {
+                    return -1;
+                }
+
+                // x preceeds y because it contains an exact match
```
aparajit-pratap, line 214: typo: precedes

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
QilongTang: Thanks for the description, understand better now!

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -43,53 +42,56 @@ internal void PopulateAutoCompleteCandidates()
...
             {
                 if (element.OutputParameters.Any(op => op == inputPortType))
                 {
                     elements.Add(element);
                 }
             }
-            //TODO if we intend to cache data instantiate this comparer at construction time
-            //and update the typeNameToCompare.
+
             var comparer = new NodeSearchElementComparer(inputPortType, dynamoViewModel.Model.LibraryServices.LibraryManagementCore);
```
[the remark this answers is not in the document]
mjkkirschner: done.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
mjkkirschner: yes, both else conditions are for `NodeModelSearchElements` which represent NodeModel nodes - I will add a comment

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
mjkkirschner: will do.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
mjkkirschner: yes, I did this just to future proof, it doesn't handle zeroTouch nodes, but it *should* handle nodeModel nodes that return multiple values. I think that depends on if the types are actually specified as attributes on the nodeModel class.

aparajit-pratap (approved):
LGTM after review comments are addressed as they've been resolved already.

mjkkirschner:
Thanks @aparajit-pratap I've fixed them locally, and will push the changes with tests.

mjkkirschner:
@QilongTang @aparajit-pratap - please take a look to see if you think tests are adequate, will merge when test results come back.

@Amoursol I think you should play with this if you have the chance to see if you find sorting satisfactory. For example, right now all constructors, actions, and queries are together, and in each group it's sub sorted by type distance. It's not clear to me which sorting should take precedence. My vote would be to get this in and test it out, then we easily tweak sorting.

Amoursol:
> @QilongTang @aparajit-pratap - please take a look to see if you think tests are adequate, will merge when test results come back.
> 
> @Amoursol I think you should play with this if you have the chance to see if you find sorting satisfactory. For example, right now all constructors, actions, and queries are together, and in each group it's sub sorted by type distance. It's not clear to me which sorting should take precedence. My vote would be to get this in and test it out, then we easily tweak sorting.

I also don't know the right answer on this, so agree that we should get it in and test during the MVP testing at AU on this. We'll want to review heavily the results versus what was expected regardless. I'll create a build and have a play too 😊

test/DynamoCoreWpfTests/NodeAutoCompleteSearchTests.cs
```diff
@@ -75,6 +93,25 @@ public void NodeSuggestions_InputPortZeroTouchNode_AreCorrect()
...
+            ViewModel.ExecuteCommand(new DynamoModel.CreateNodeCommand(
+               node, 0, 0, true, false));
+            DispatcherUtil.DoEvents();
+            var nodeView = NodeViewWithGuid(node.GUID.ToString());
+            var searchViewModel = (ViewModel.CurrentSpaceViewModel.NodeAutoCompleteSearchViewModel as NodeAutoCompleteSearchViewModel);
+            searchViewModel.PortViewModel = nodeView.ViewModel.InPorts.FirstOrDefault();
+
+            var suggestions = searchViewModel.GetMatchingNodes();
+            Assert.AreEqual(SearchElementGroup.Create, suggestions.FirstOrDefault().Group);
+            Assert.AreEqual(SearchElementGroup.Action, suggestions.ElementAt(45).Group);
```
aparajit-pratap, line 142: So this could break when we add new nodes that output a `Curve`, or its subtype? Are we okay with that? I prefer using FFITarget for this reason but I guess this is fine too.

test/DynamoCoreWpfTests/NodeAutoCompleteSearchTests.cs
```diff
@@ -97,7 +134,7 @@ public void NodeSuggestions_InputPortBuiltInNode_AreCorrect()
             var searchViewModel = (ViewModel.CurrentSpaceViewModel.NodeAutoCompleteSearchViewModel as NodeAutoCompleteSearchViewModel);
             searchViewModel.PortViewModel = inPorts[1];
             var suggestions = searchViewModel.GetMatchingNodes();
-            Assert.AreEqual(20, suggestions.Count());
+            Assert.AreEqual(22, suggestions.Count());
```
aparajit-pratap, line 169: Is this difference because you are stripping out array ranks now?

src/DynamoCoreWpf/ViewModels/Search/SearchViewModel.cs
```diff
@@ -897,7 +897,7 @@ private NodeSearchElementViewModel MakeNodeSearchElementVM(NodeSearchElement ent
             var elementVM = element != null
                 ? new CustomNodeSearchElementViewModel(element, this)
                 : new NodeSearchElementViewModel(entry, this);
-
+            //TODO lookout for leak.
```
aparajit-pratap, line 900: Is this still an issue?

mjkkirschner: I'm not sure.

test/DynamoCoreWpfTests/NodeAutoCompleteSearchTests.cs
mjkkirschner: the additional 2 from the last commit is because I'm also loading protogeometry into this test fixture.

test/DynamoCoreWpfTests/NodeAutoCompleteSearchTests.cs
mjkkirschner: I can try to rewrite - the 45 is just somewhere in the middle - not a specific node.

QilongTang:
> > @QilongTang @aparajit-pratap - please take a look to see if you think tests are adequate, will merge when test results come back.
> > @Amoursol I think you should play with this if you have the chance to see if you find sorting satisfactory. For example, right now all constructors, actions, and queries are together, and in each group it's sub sorted by type distance. It's not clear to me which sorting should take precedence. My vote would be to get this in and test it out, then we easily tweak sorting.
> 
> I also don't know the right answer on this, so agree that we should get it in and test during the MVP testing at AU on this. We'll want to review heavily the results versus what was expected regardless. I'll create a build and have a play too 😊

I also vote for getting in and test. Tests look good

Amoursol:
I've tested and ran through with @mjkkirschner now. Looks to go me for the MVP - We can now get in front of customers and ask for feedback.

PR #11177, opened 2020-10-13
Node auto layout for node autocomplete

aparajit-pratap:
### Purpose

https://jira.autodesk.com/browse/DYN-2978

This builds on top of PR https://github.com/DynamoDS/Dynamo/pull/11164. 

The node chosen from node autocomplete suggestions is created, connected, and then placed relative to the initial node using the graph auto layout. A single undo-recorder group is initialized for this entire operation so that it can be undone in a single step. The following are the sequence of steps:
1. A new undo-redo recorder group is first initialized
2. A new option is added to the `CreateAndConnectCommand` to make it use an existing undo-redo recorder group instead of creating a new one
3. The `NodeViewReady` event on DynamoViewModel is subscribed to. Only when this event is raised does the size of the new node created get set. Only once the size is set, can graph layout work correctly.
4. In the `NodeViewReady` event handler the graph auto layout is performed using only the input nodes to the initial node. This is to ensure that the initial node does not jump to a new location on graph layout.
5. The undo-redo group is finally disposed and the event unsubscribed.

_Please hide whitespace to review this PR._

place_and_connect_nodes_final

### Declarations

### FYIs
@Amoursol


src/DynamoCore/Graph/Workspaces/LayoutExtensions.cs
```diff
@@ -18,7 +18,7 @@ public static class LayoutExtensions
         /// This function wraps a few methods on the workspace model layer
         /// to set up and run the graph layout algorithm.
         /// </summary>
-        internal static List<GraphLayout.Graph> DoGraphAutoLayout(this WorkspaceModel workspace)
+        internal static List<GraphLayout.Graph> DoGraphAutoLayout(this WorkspaceModel workspace, bool reuseUndoRecorder = false)
```
QilongTang, line 21: Would you add comments for params here since they are missing anyway?

src/DynamoCore/Models/DynamoModelCommands.cs
```diff
@@ -83,43 +83,55 @@ private void CreateAndConnectNodeImpl(CreateAndConnectNodeCommand command)
...
+        /// This method assumes that there exists an undo-redo action group already
+        /// that can be used to record creation and deletion of models.
+        /// </summary>
+        /// <param name="command"></param>
+        private void CreateAndConnectNodeImplWithUndoGroup(CreateAndConnectNodeCommand command)
+        {
+            var newNode = CreateNodeFromNameOrType(command.ModelGuid, command.NewNodeName);
 
-                var models = GetConnectorsToAddAndDelete(inPortModel, outPortModel);
+            newNode.X = command.X;
```
QilongTang, line 99: I guess from command standpoint, newNode could be null when the node name is invalid? So it could fail or crash? This is not something this PR brings but would like to confirm what would be the behavior.

src/DynamoCore/Models/DynamoModelCommands.cs
```diff
@@ -83,43 +83,55 @@ private void CreateAndConnectNodeImpl(CreateAndConnectNodeCommand command)
...
-                            CurrentWorkspace.UndoRecorder.RecordCreationForUndo(modelPair.Key);
-                            break;
-                        case UndoRedoRecorder.UserAction.Deletion:
-                            CurrentWorkspace.UndoRecorder.RecordDeletionForUndo(modelPair.Key);
-                            break;
-                    }
+                    case UndoRedoRecorder.UserAction.Creation:
+                        CurrentWorkspace.UndoRecorder.RecordCreationForUndo(modelPair.Key);
+                        break;
+                    case UndoRedoRecorder.UserAction.Deletion:
```
QilongTang, line 132: @aparajit-pratap Wouldn't all of them creation?

src/DynamoCoreWpf/ViewModels/Search/SearchViewModel.cs
```diff
@@ -1061,14 +1073,56 @@ public void OnSearchElementClicked(NodeModel nodeModel, Point position)
...
+                undoRecorderGroup = dynamoViewModel.CurrentSpace.UndoRecorder.BeginActionGroup();
+            }
+
+            // Create a new node based on node creation name and connect ports
+            dynamoViewModel.ExecuteCommand(new DynamoModel.CreateAndConnectNodeCommand(id, portModel.Owner.GUID,
+                nodeCreationName, 0, portModel.Index, adjustedX, 0, false, false, true));
+
+            // Clear current selections and select all input nodes
+            DynamoSelection.Instance.ClearSelection();
+            var inputNodes = portModel.Owner.InputNodes.Values.Where(x => x != null).Select(y => y.Item2);
```
QilongTang, line 1107: Can you explain the Item2 part? I think I tried this one before by just calling Select and did not work.. Maybe i missed something

src/DynamoCore/Models/DynamoModelCommands.cs
aparajit-pratap: Yes, you're right, `newNode` can be null. I will add a null check.

src/DynamoCoreWpf/ViewModels/Search/SearchViewModel.cs
aparajit-pratap: Haha, nothing smart here. `InputNodes` property on `NodeModel` is of type `IDictionary<int, Tuple<int, NodeModel>>`. I'm just getting the `NodeModel` part of the tuple.

src/DynamoCore/Models/DynamoModelCommands.cs
aparajit-pratap: Honestly, not sure. I've left this unchanged from previously, just refactored the code a bit. Please hide whitespace to see actual diff.

src/DynamoCoreWpf/ViewModels/Search/SearchViewModel.cs
```diff
@@ -1061,14 +1073,62 @@ public void OnSearchElementClicked(NodeModel nodeModel, Point position)
...
+            dynamoViewModel.ExecuteCommand(new DynamoModel.CreateAndConnectNodeCommand(id, portModel.Owner.GUID,
+                nodeCreationName, 0, portModel.Index, adjustedX, 0, false, false, true));
+
+            // Clear current selections and select all input nodes as we need to perform Auto layout on only the input nodes.
+            DynamoSelection.Instance.ClearSelection();
+            var inputNodes = portModel.Owner.InputNodes.Values.Where(x => x != null).Select(y => y.Item2);
+
+            foreach (var inputNode in inputNodes)
+            {
+                DynamoSelection.Instance.Selection.AddUnique(inputNode);
```
QilongTang, line 1116: Oh @aparajit-pratap I see, you are touching the `Selection` directly rather than calling `Select()` on each inidividual node. That explains why my approach earlier may not work.. Good that you have a fresh eye on this!

QilongTang (approved):
LGTM

QilongTang:
If there is a chance, I would like to get some knowledge share around undo/redo recorder and what it means to reuse it. That is the only piece I am not sure if the way in this PR will be something we want to promote in the feature.

aparajit-pratap:
@QilongTang I have done some final refactor - I have moved the commands from `SearchViewModel` to `NodeSearchElementViewModel`. It would be great if you could review one last time.

src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
```diff
@@ -233,14 +241,85 @@ public ImageSource LargeIcon
...
+            }
+
+            var initialNode = portModel.Owner;
+            var initialNodeVm = dynamoViewModel.CurrentSpaceViewModel.Nodes.FirstOrDefault(x => x.Id == initialNode.GUID);
+            var id = Guid.NewGuid();
+
+            var adjustedX = initialNodeVm.X;
+
+            var createAsDownStreamNode = portModel.PortType == PortType.Output;
+            // Placing the new node based on which input port it is connecting to.
```
QilongTang, line 273: Which port?

src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
```diff
@@ -233,14 +241,85 @@ public ImageSource LargeIcon
...
+                // Placing the new node to the left of initial node
+                adjustedX -= initialNode.Width + 50;
+            }
+
+            // Create a new node based on node creation name and connection ports
+            dynamoViewModel.ExecuteCommand(new DynamoModel.CreateAndConnectNodeCommand(id, initialNode.GUID,
+                Model.CreationName, 0, portModel.Index, adjustedX, 0, createAsDownStreamNode, false, true));
+
+            // Clear current selections and select all input nodes as we need to perform Auto layout on only the input nodes.
+            DynamoSelection.Instance.ClearSelection();
```
QilongTang, line 290: @aparajit-pratap Do you know if selection is part of undo/redo sequence? I have been thinking about if there could be any side effects to this but in general feels OK

QilongTang (approved):

src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
aparajit-pratap: This is not part of the undo/redo sequence. I think you'll need to use one of the selection related commands to add it to undo/redo queue.

PR #11182, opened 2020-10-13
Implement UX design logic fixes for resolving inconsistencies of the Python editor window

zeusongit:
### Purpose
To align the behavior of python script editor window as per [DYN-2803](https://jira.autodesk.com/browse/DYN-2803)
In this PR:
- Changing the drop-down engine won't trigger graph run irrespective of auto/manual mode.
- Revert button will revert to the last saved engine as well.
- Save and revert won't close the editor window.
- Only exit button (x) closes the window.

Tests updated. [#31](https://git.autodesk.com/Dynamo/DynamoUIAutomation/pull/31)

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/Libraries/PythonNodeModels/PythonNode.cs
+++ b/src/Libraries/PythonNodeModels/PythonNode.cs
@@ -60,6 +60,32 @@ namespace PythonNodeModels
                 }
             }
         }
+        [JsonConverter(typeof(StringEnumConverter))]
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Populate)]
+        [DefaultValue(nameof(PythonEngineVersion.IronPython2))]
+        /// <summary>
+        /// Return the user selected python engine enum.
+        /// </summary>
+        public PythonEngineVersion CachedEngine
+        {
+            get
+            {
+                // This is a first-time case for newly created nodes only
+                if (engine == PythonEngineVersion.Unspecified)
+                {
+                    SetEngineByDefault();
+                }
+                return engine;
+            }
+            set
+            {
+                if (engine != value)
+                {
+                    engine = value;
+                    RaisePropertyChanged(nameof(CachedEngine));
+                }
+            }
+        }
 
         private static ObservableCollection<PythonEngineVersion> availableEngines;
         /// <summary>
@@ -173,7 +199,7 @@ namespace PythonNodeModels
                 PythonEngineVersion result;
                 if (Enum.TryParse<PythonEngineVersion>(value, out result))
                 {
-                    Engine = result;
+                    CachedEngine = result;
                     return true;
                 }
 
@@ -347,7 +373,7 @@ namespace PythonNodeModels
 
             if (engineNode != null)
             {
-                this.Engine = (PythonEngineVersion)Enum.Parse(typeof(PythonEngineVersion),engineNode.InnerText);
+                this.CachedEngine = (PythonEngineVersion)Enum.Parse(typeof(PythonEngineVersion),engineNode.InnerText);
             }
         }
 
--- a/src/Libraries/PythonNodeModels/PythonNode.cs
+++ b/src/Libraries/PythonNodeModels/PythonNode.cs
@@ -33,7 +33,7 @@ namespace PythonNodeModels
     public abstract class PythonNodeBase : VariableInputNode
     {
         private PythonEngineVersion engine = PythonEngineVersion.Unspecified;
-        private PythonEngineVersion cached_engine = PythonEngineVersion.Unspecified;
+        private PythonEngineVersion cachedEngine = PythonEngineVersion.Unspecified;
 
         [JsonConverter(typeof(StringEnumConverter))]
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.Populate)]
@@ -68,13 +68,13 @@ namespace PythonNodeModels
         {
             get
             {
-                return cached_engine;
+                return cachedEngine;
             }
             set
             {
-                if (cached_engine != value)
+                if (cachedEngine != value)
                 {
-                    cached_engine = value;
+                    cachedEngine = value;
                     RaisePropertyChanged(nameof(CachedEngine));
                 }
             }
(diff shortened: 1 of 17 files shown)


src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -47,7 +48,7 @@ public PythonEngineVersion Engine
                 // This is a first-time case for newly created nodes only
                 if (engine == PythonEngineVersion.Unspecified)
                 {
-                    SetEngineByDefault();
+                    SetEngineByDefault(ref engine);
```
mjkkirschner, line 51: I'm very confused why you would use a `ref` parameter here.

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -60,6 +61,24 @@ public PythonEngineVersion Engine
                 }
             }
         }
+        /// <summary>
+        /// Return the user selected python engine enum.
+        /// </summary>
+        public PythonEngineVersion CachedEngine
```
mjkkirschner, line 67: This does not seem like something which needs to be exposed as a public property... it instead seems like something that is an internal implementation detail.

src/Libraries/PythonNodeModels/PythonNode.cs
zeusongit: I am using the same function to update different instances of PythonEngineVersion, therefore passing by reference.

zeusongit: I am sorry, I just saw that I am no longer doing that, will update this

src/Libraries/PythonNodeModels/PythonNode.cs
zeusongit: making this internal makes the initial dropdown value to be blank, this is accessed directly by the dropdown

QilongTang: Yup. I believe direct binding on WPF assumes the property to be public..

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -33,6 +33,7 @@ public PythonCodeMigrationEventArgs(string oldCode, string newCode)
     public abstract class PythonNodeBase : VariableInputNode
     {
         private PythonEngineVersion engine = PythonEngineVersion.Unspecified;
+        private PythonEngineVersion cached_engine = PythonEngineVersion.Unspecified;
```
QilongTang, line 36: I think the consensus around C# naming convention these days is do not user underscore at all, would you try `cachedEngine`?

QilongTang (approved):
LGTM with one comment

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
```diff
@@ -200,14 +204,6 @@ private void OnMoreInfoClicked(object sender, RoutedEventArgs e)
 
         private void OnEngineChanged(object sender, System.Windows.Controls.SelectionChangedEventArgs e)
         {
-            //removedItems will be empty during the first binding
-            //as the window is constructed, we don't want to execute the node just
-            //as a consequence of opening the editor.
-            if (e.RemovedItems.Count > 0)
-            {
-                UpdateScript(editText.Text);
-            }
```
QilongTang, line 209: Neat, no more auto execution or debt

src/Libraries/PythonNodeModels/PythonNode.cs
mjkkirschner: I understand that, but why are we binding to this at all? Why not continue binding to the previous engine property. It's Adding public properties to a NodeModel type (its API) is not something that should be done without a good reason.

aparajit-pratap:
@zeusongit I see that UX design states that nothing should run the graph other than the Run button and remember you had a question during standup about whether the Save/Revert buttons should run the graph. So, I would like to make sure that that is the case - we are not running the graph during Save or Revert, correct?

zeusongit:
> @zeusongit I see that UX design states that nothing should run the graph other than the Run button and remember you had a question during standup about whether the Save/Revert buttons should run the graph. So, I would like to make sure that that is the case - we are not running the graph during Save or Revert, correct?

Currently Save/Revert will run the graph in Automatic mode and won't run in Manual mode, same as it was before. This was my confusion initially and after discussion with aaron and in standups it was decided to stick with the ACs only, as the AC was the subset of what Jingyi proposed initially. 
@Amoursol any thoughts on this?

Amoursol:
> > @zeusongit I see that UX design states that nothing should run the graph other than the Run button and remember you had a question during standup about whether the Save/Revert buttons should run the graph. So, I would like to make sure that that is the case - we are not running the graph during Save or Revert, correct?
> 
> Currently Save/Revert will run the graph in Automatic mode and won't run in Manual mode, same as it was before. This was my confusion initially and after discussion with aaron and in standups it was decided to stick with the ACs only, as the AC was the subset of what Jingyi proposed initially.
> @Amoursol any thoughts on this?

Subset AC's as agreed at standup are totally cool @zeusongit :)

src/Libraries/PythonNodeModels/PythonNode.cs
mjkkirschner: It sounds to me, like you don't want to use a binding on the NodeModel - this is purely a `View` level property correct? I think you should bind to the view or viewModel for this property, and add this property there.

It feels strange to me to put a public property on the pythonNode which has NO impact on execution, only on view, meaning if the view was not present this property is useless, and also confusing.

QilongTang:
OK, seems we are all on board. We can try to confirm with users but I do know Revit users would not put Dynamo into automatic mode often, plus automatic mode literally means any changes to the current workspace should trigger a run so in this case I feel it is still legit/consistent in the current state after this PR. Maybe worth bring up to the forum though but since the ACs are fulfilled, I encourage us to do this in agile way :)

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
```diff
@@ -46,6 +47,7 @@ public partial class ScriptEditorWindow : ModelessChildWindow
             nodeModel.CodeMigrated += OnNodeModelCodeMigrated;
 
             InitializeComponent();
+            this.DataContext = this;
```
zeusongit, line 50: DataContext was null before

PR #11184, opened 2020-10-15
Translate package publishing error messages

mmisol:
### Purpose

These errors originate from PackageManager. In order to translate them,
these are identified, parsed and then translated using local resources.

### Declarations

### Reviewers

@QilongTang @zeusongit

### DIFF
--- a/test/DynamoCoreWpfTests/PackageManagerUnitTests.cs
+++ b/test/DynamoCoreWpfTests/PackageManagerUnitTests.cs
@@ -1,0 +1,40 @@
+﻿using Dynamo;
+using Dynamo.PackageManager;
+using NUnit.Framework;
+
+namespace DynamoCoreWpfTests
+{
+    [TestFixture]
+    public class PackageManagerUnitTests : UnitTestBase
+    {
+        #region PackageManager error translation
+
+        [Test]
+        public void ReturnsTranslationForUserIsNotAMaintainerError()
+        {
+            var message = "The user sending the new package version, DynamoTeam, is not a maintainer of the package Clockwork for Dynamo 1.x";
+            var actual = PublishPackageViewModel.TranslatePackageManagerError(message);
+            var expected = "The current user, 'DynamoTeam', is not a maintainer of the package 'Clockwork for Dynamo 1.x'.";
+            Assert.AreEqual(expected, actual);
+        }
+
+        [Test]
+        public void ReturnsTranslationForPackageAlreadyExistsError()
+        {
+            var message = "A package with the given name and engine already exists.";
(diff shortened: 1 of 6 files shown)


test/DynamoCoreWpfTests/PackageManagerUnitTests.cs
```diff
@@ -0,0 +1,40 @@
...
+            var message = "The user sending the new package version, DynamoTeam, is not a maintainer of the package Clockwork for Dynamo 1.x";
+            var actual = PublishPackageViewModel.TranslatePackageManagerError(message);
+            var expected = "The current user, 'DynamoTeam', is not a maintainer of the package 'Clockwork for Dynamo 1.x'.";
+            Assert.AreEqual(expected, actual);
+        }
+
+        [Test]
+        public void ReturnsTranslationForPackageAlreadyExistsError()
+        {
+            var message = "A package with the given name and engine already exists.";
```
QilongTang, line 24: maybe use resource string here?

mmisol: This is a test. Why would we use one?

test/DynamoCoreWpfTests/PackageManagerUnitTests.cs
```diff
@@ -0,0 +1,40 @@
...
+            var actual = PublishPackageViewModel.TranslatePackageManagerError(message);
+            var expected = "A package with the given name already exists.";
```
QilongTang, line 26: I meant to suggest using the resource string you defined `Dynamo.Properties.Resources.PackageManagerPackageAlreadyExists` here

mmisol: Ah, now it makes sense. While I could do that to avoid repeating the string, it would also mean to use part of the solution for the assertion, which is kind of fishy for a test to do in my opinion. Maybe this case it is not so evident, but for `ReturnsTranslationForUserIsNotAMaintainerError` one would also have to format the string by filling the placeholder, which is pretty close to the solution. I think I'm inclined to leave the test independent because the repetition is just a single string and it feels cleaner to go this path. What do you think @QilongTang ?

QilongTang: Thinking about it now, I think we could go either way. If you prefer to leave it as it is, let's do that

QilongTang:
This looks good to me. @zeusongit any comments?

zeusongit (approved):
LGTM

QilongTang (approved):

PR #11192, opened 2020-10-20
Add search functionality to Node Autocomplete window.

reddyashish:
### Purpose

This PR adds the search functionality to the Node Autocomplete window.
Task: https://jira.autodesk.com/browse/DYN-3136

TODO:
1) Working on the tests. 

search node autocompletion

We are caching the list of matching nodes after they are calculated for the first time and re-use them. 

As of now, we are just matching the node name with the user input. If we want to match for any other node property, we can add the necessary condition. 

### Declarations

### Reviewers

@DynamoDS/dynamo


mjkkirschner:
@reddyashish could you add info to this PR about how this works?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -48,6 +48,23 @@ internal void PopulateAutoCompleteCandidates()
             });
         }
 
+        internal void SearchAutoCompleteCandidates(string input)
+        {
+            if (PortViewModel == null) return;
+
+            var searchElements = GetMatchingNodes();
```
mjkkirschner, line 55: I think this has the potential to be very slow - this seems to be recalculating the matched nodes every time a user enters a new character...
I think this should be calculated only once

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -48,6 +48,23 @@ internal void PopulateAutoCompleteCandidates()
...
+            FilteredResults = searchElements.Select(e =>
+            {
+                if (e.Name.Contains(input))
```
mjkkirschner, line 58: it also seems not to use the regular search algorithm.... that might be a good thing.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
reddyashish: Yes, I have changed this. Will be updating with new commits. Running into some errors with some workflows.

aparajit-pratap: Yes, I agree, the matched nodes should have already been calculated just before the autocomplete popup shows up so this looks wasteful to re-calculate the matched nodes each time you type a search character. Can we cache them and re-use that cache for search?

aparajit-pratap: Also, I think this API has been renamed to `GetMatchingSearchElements` in latest master.

src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml.cs
```diff
@@ -67,8 +67,14 @@ private void OnSearchTextBoxTextChanged(object sender, TextChangedEventArgs e)
...
                 binding.UpdateSource();
 
-            if (ViewModel != null)
-                ViewModel.SearchCommand.Execute(null);
+            if (ViewModel != null) 
+            {
+                Dispatcher.BeginInvoke(new Action(() =>
+                {
+                    // SearchTextBox.Focus();
+                    ViewModel.SearchAutoCompleteCandidates(SearchTextBox.Text);
```
QilongTang, line 75: Would like to know the consequence of not using ViewModel command here

mjkkirschner: I'll let @reddyashish add more info - but my thought is that this means the search is a special search (as evidenced by the fact that it uses just a `string.contains` on the name of the node.

As a consequence it also means the search cannot be started from a ViewExtension.

Commands make a lot of sense for operations we want to automate via tests, or which should be apis.
Don't have a strong feeling on this one.

QilongTang: I thought whether implementing a viewcommand does not relate to which search algorithm we are using. If we do, I expect us instead of using the existing command but implement a new one dedicated for our needs with search elements already replaced for search view model so that we are not going over every searchElements in the pool but we still encounter for any search key words defined. For example, I can still search for cube to get to cuboid, etc.

I have no strong feeling about view commands but would like to validate what we missing if not using the search algorithm.

mjkkirschner: descriptions will not be searched, weights will not be used, keywords will not be used.
Theres probably more.

reddyashish: As of now, just matching the name of the node. Do we want to match the description or any other node property?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
reddyashish: I am not clear on this. The regular search algorithm that we use for library search of in-canvas search?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -51,9 +53,17 @@ private void InitializeDefaultAutoCompleteCandidates()
...
+            searchElementsCache.Clear();
+
             if (PortViewModel == null) return;
 
             var searchElements = GetMatchingSearchElements();
+
+            foreach (NodeSearchElement element in searchElements)
+            {
+                searchElementsCache.Add(element);
+            }
```
QilongTang, line 65: Instead of hard copy here, seems we can just do `searchElementsCache = GetMatchingSearchElements();` above?

src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml.cs
QilongTang: I will review tonight again. it does seem there's gap not covered by naming comparison

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -70,6 +80,31 @@ internal void PopulateAutoCompleteCandidates()
...
+            if (PortViewModel == null) return;
+
+            FilteredResults = searchElementsCache.AsEnumerable().Where(e =>
+                                QuerySearchElements(e, input)).Select(e =>
+                                {
+                                      var vm = new NodeSearchElementViewModel(e, this);
+                                      vm.RequestBitmapSource += SearchViewModelRequestBitmapSource;
+                                      return vm;
+                                });
+        }
```
QilongTang, line 94: I think you can return search elements here and make another function including the part to convert searchelements to FilteredResults so we have less duplicated code, what do you think?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -70,6 +80,31 @@ internal void PopulateAutoCompleteCandidates()
             }
         }
 
+        internal void SearchAutoCompleteCandidates(string input)
```
QilongTang, line 83: Would you add some description?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -70,6 +80,31 @@ internal void PopulateAutoCompleteCandidates()
...
+            FilteredResults = searchElementsCache.AsEnumerable().Where(e =>
+                                QuerySearchElements(e, input)).Select(e =>
+                                {
+                                      var vm = new NodeSearchElementViewModel(e, this);
+                                      vm.RequestBitmapSource += SearchViewModelRequestBitmapSource;
+                                      return vm;
+                                });
+        }
+
+        private bool QuerySearchElements(NodeSearchElement e, string input)
```
QilongTang, line 96: Some description would be good

QilongTang (commented):
Looking solid, would love to see some clean up and a unit test to cover the search function

src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml.cs
QilongTang: Thinking about this more, I'm fine with this approach given we may not want to be limited to the existing search and unclear if user would like to search through description yet

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
QilongTang: I guess I mean if there is a way to convert the duplicate code her (with line 76) to a shared function

QilongTang:
One last comment

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -17,6 +18,7 @@ public class NodeAutoCompleteSearchViewModel : SearchViewModel
     {
 
         internal PortViewModel PortViewModel { get; set; }
+        private List<NodeSearchElement> searchElementsCache = new List<NodeSearchElement>();
```
aparajit-pratap, line 21: Do we need this initialization?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -51,17 +53,25 @@ private void InitializeDefaultAutoCompleteCandidates()
         
         internal void PopulateAutoCompleteCandidates()
         {
+            searchElementsCache = new List<NodeSearchElement>();
```
aparajit-pratap, line 56: I don't think we need to initialize it here as it is being assigned on line 60 anyway.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -51,17 +53,25 @@ private void InitializeDefaultAutoCompleteCandidates()
...
             if (PortViewModel == null) return;
 
-            var searchElements = GetMatchingSearchElements();
+            searchElementsCache = GetMatchingSearchElements().ToList();
+
             // If node match searchElements found, use default suggestions
-            if (!searchElements.Any())
+            if (!searchElementsCache.Any())
             {
+                foreach (NodeSearchElementViewModel element in DefaultResults)
```
aparajit-pratap, line 65: `searchElementsCache = DefaultResults.Select(e => e.Model);`

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -70,6 +80,38 @@ internal void PopulateAutoCompleteCandidates()
...
+
+        /// <summary>
+        /// Returns true if the user input matches the name of the filtered node element. 
+        /// </summary>
+        /// <returns>True or false</returns>
+        private bool QuerySearchElements(NodeSearchElement e, string input) 
+        {
+            StringComparison stringComparison = StringComparison.CurrentCultureIgnoreCase;
+
+            if (e.Name.IndexOf(input, stringComparison) >= 0)
```
aparajit-pratap, line 107: `return e.Name.IndexOf(input, stringComparison) >= 0 ? true : false;`

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -70,6 +80,38 @@ internal void PopulateAutoCompleteCandidates()
...
 
+        /// <summary>
+        /// Filters the matching node search elements based on user input in the search field. 
+        /// </summary>
+        internal void SearchAutoCompleteCandidates(string input)
+        {
+            if (PortViewModel == null) return;
+
+            FilteredResults = searchElementsCache.Where(e =>
+                                QuerySearchElements(e, input)).Select(e =>
```
aparajit-pratap, line 91: Can we not use the same search algorithm as the one used for in-canvas search?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
reddyashish: Since we are storing the search results in a list, when we call the PopulateAutoCompleteCandidates API, we want to clear the previous matching node elements and update it with the new list.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
reddyashish: This can be removed.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
reddyashish: For the incanvas search we use the search command on the model and search for all the nodes in the library. It doesn't search a specific list of nodes.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
reddyashish: Yes changing this to a shared function.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
aparajit-pratap: Yes, I got that but can we not use the same algorithm/methods, refactor some code, etc?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
aparajit-pratap: Won't the previous cache be cleared when you point this cache to a new list?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
QilongTang: @aparajit-pratap Can you be more specific? We are not using the same algorithm here right?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
reddyashish: Yes, I missed that part. Removing this initialization.

reddyashish:
@QilongTang Refactored the code and fixed the tests.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
reddyashish: Yes, Aparajit is asking if we could use the same algorithm to search. https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Search/SearchDictionary.cs#L327

We can use it with some changes but since the list of search elements won't be that long and we are checking just the name, I kept it simple without using the weights and time. What do you guys think?

QilongTang: I think this looks good for now.

QilongTang (approved):
LGTM

QilongTang:
@reddyashish Can we merge and cherry-pick to RC2.9.0?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
aparajit-pratap: I think we should keep the search algorithm consistent irrespective of how many search elements there might be. I think one of @Amoursol's points of emphasis was that search is key for the node autocomplete feature. We should at least create an improvement task for this to address in the near future.

Amoursol: Indeed @aparajit-pratap - now that I'm back we will spin up a conversation around Search :) For Node Autocomplete to be as successful as we want it to be, Search also needs to be robust.

PR #11201, opened 2020-10-23
Skip primitive type input ports from node-autocomplete

zeusongit:
### Purpose

[DYN-3184](https://jira.autodesk.com/browse/DYN-3184)

Skip primitive input ports when displaying node autocomplete suggestions.

nautof

### Declarations

### Reviewers

@DynamoDS/dynamo 

### FYIs

@QilongTang


src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -85,6 +85,12 @@ internal IEnumerable<NodeSearchElement> GetMatchingSearchElements()
         {
             var elements = new List<NodeSearchElement>();
             var inputPortType = PortViewModel.PortModel.GetInputPortType();
+            if (inputPortType == null) return elements;
+
+            //List of nodes that are skipped temporarily, and will display list of default suggestions instead.
+            var skippedInputPorts = new List<string>() { "var", "object", "string", "bool", "int", "double" };
+
+            var libraryServices = dynamoViewModel.Model.LibraryServices;
```
QilongTang, line 93: Is this line still needed?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -102,7 +108,9 @@ internal IEnumerable<NodeSearchElement> GetMatchingSearchElements()
                 var returnTypeName = ztSearchElement.Descriptor.ReturnType.Name;
 
                 var descriptor = ztSearchElement.Descriptor;
-                if ((returnTypeName == inputPortType)
+                if (descriptor.ReturnType.ToString() == inputPortType)
+                    if (!skippedInputPorts.Any(s => s == returnTypeName.ToString())
```
QilongTang, line 112: @zeusongit @mjkkirschner Seems @aparajit-pratap  The goal here is to let `elements` remain empty for these specific inputPortTypes. Should we move the check out side of the scanning to save performance? e.g. moving to line 96 since we know the results are not going to be useful?

zeusongit: seems like a good idea, will be updating

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -102,7 +107,8 @@ internal IEnumerable<NodeSearchElement> GetMatchingSearchElements()
                 var returnTypeName = ztSearchElement.Descriptor.ReturnType.Name;
 
                 var descriptor = ztSearchElement.Descriptor;
-                if ((returnTypeName == inputPortType)
+                if (descriptor.ReturnType.ToString() == inputPortType)
```
mjkkirschner, line 110: can we use `{` around if statements - it's in our coding style, if we're going to stop doing this let's update the style guide.
https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards#conditional-blocks

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -85,9 +85,14 @@ internal IEnumerable<NodeSearchElement> GetMatchingSearchElements()
         {
             var elements = new List<NodeSearchElement>();
             var inputPortType = PortViewModel.PortModel.GetInputPortType();
+
+            //List of nodes that are skipped temporarily, and will display list of default suggestions instead.
+            var skippedInputPorts = new List<string>() { "var", "object", "string", "bool", "int", "double" };
```
mjkkirschner, line 90: how do you all feel about removing string from this list? How many items does it return?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -102,10 +107,13 @@ internal IEnumerable<NodeSearchElement> GetMatchingSearchElements()
                 var returnTypeName = ztSearchElement.Descriptor.ReturnType.Name;
 
                 var descriptor = ztSearchElement.Descriptor;
-                if ((returnTypeName == inputPortType)
-                    || DerivesFrom(inputPortType, returnTypeName, core))
+                if (descriptor.ReturnType.ToString() == inputPortType)
```
aparajit-pratap, line 110: Why has this extra conditional been added?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
aparajit-pratap, line 90: I would rather name this as `skippedInputTypes`

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -85,9 +85,14 @@ internal IEnumerable<NodeSearchElement> GetMatchingSearchElements()
         {
             var elements = new List<NodeSearchElement>();
             var inputPortType = PortViewModel.PortModel.GetInputPortType();
+
+            //List of nodes that are skipped temporarily, and will display list of default suggestions instead.
+            var skippedInputPorts = new List<string>() { "var", "object", "string", "bool", "int", "double" };
+          
+            if (inputPortType == null || (skippedInputPorts.Any(s => s == inputPortType.ToString()))) return elements;
```
aparajit-pratap, line 92: @QilongTang shouldn't we be returning the default search elements in this case?

QilongTang: @aparajit-pratap Yes, it depends on where you would like to implement. I think @zeusongit decides to put it inside of the API itself, we can also implement as part of PopulateAutoCompleteCandidates(). That's why I was asking your preference if you have any

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -107,13 +107,10 @@ internal IEnumerable<NodeSearchElement> GetMatchingSearchElements()
                 var returnTypeName = ztSearchElement.Descriptor.ReturnType.Name;
 
                 var descriptor = ztSearchElement.Descriptor;
-                if (descriptor.ReturnType.ToString() == inputPortType)
+                if ((returnTypeName == inputPortType)
+                || DerivesFrom(inputPortType, returnTypeName, core))
```
aparajit-pratap, line 111: Please indent this line for readability

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
QilongTang: `String` currently already return default list, so I guess it could be optional. It is defined by @aparajit-pratap in original AC. That's why @zeusongit added it. If we all agree, I can update the AC
image

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
zeusongit: Yes, by returning `null` here, it is handled in the `PopulateAutoCompleteCandidates` that fills the default values

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -85,9 +85,14 @@ internal IEnumerable<NodeSearchElement> GetMatchingSearchElements()
         {
             var elements = new List<NodeSearchElement>();
             var inputPortType = PortViewModel.PortModel.GetInputPortType();
+
+            //List of nodes that are skipped temporarily, and will display list of default suggestions instead.
+            var skippedInputTypes = new List<string>() { "var", "object", "string", "bool", "int", "double" };
+          
+            if (inputPortType == null || (skippedInputTypes.Any(s => s == inputPortType.ToString()))) return elements;
```
QilongTang, line 92: I think @mjkkirschner also suggested using {} and follow our own coding style documented.

QilongTang (approved):
LGTM with one comment

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
QilongTang: If no one have strong opinions, let's leave it as it is as the end result is same

test/DynamoCoreWpfTests/NodeAutoCompleteSearchTests.cs
```diff
@@ -164,35 +164,82 @@ public void NodeSuggestions_InputPortBuiltInNode_AreCorrect()
             Assert.AreEqual("string", type);
 
             var searchViewModel = (ViewModel.CurrentSpaceViewModel.NodeAutoCompleteSearchViewModel as NodeAutoCompleteSearchViewModel);
-            searchViewModel.PortViewModel = inPorts[1];
+            searchViewModel.PortViewModel = inPorts[0];
```
mjkkirschner, line 167: I think if you make this change you should change the assertion above on line 164 so we have a clue as to what this is testing...

aparajit-pratap, line 167: I don't understand this change.

test/DynamoCoreWpfTests/NodeAutoCompleteSearchTests.cs
```diff
@@ -164,35 +164,82 @@ public void NodeSuggestions_InputPortBuiltInNode_AreCorrect()
...
             var suggestions = searchViewModel.GetMatchingSearchElements();
-            Assert.AreEqual(22, suggestions.Count());
+            Assert.AreEqual(70, suggestions.Count());
```
aparajit-pratap, line 169: Or this one?

test/DynamoCoreWpfTests/NodeAutoCompleteSearchTests.cs
```diff
@@ -164,35 +164,82 @@ public void NodeSuggestions_InputPortBuiltInNode_AreCorrect()
...
+                "Operators.*",
+                "Operators./",
+                "Operators.||",
+                "Operators.+",
+                "Operators.<",
+                "Operators.<=",
+                "Operators.==",
+                "Operators.>",
+                "Operators.>=",
+                "Operators.Not"
```
aparajit-pratap, line 242: Why are there so many results?

mjkkirschner: the port thats being matched has changed.

test/DynamoCoreWpfTests/NodeAutoCompleteSearchTests.cs
aparajit-pratap: It looks like the inPorts[0] corresponds to a `var[]..[]` type, which should be skipped, correct? In which case, shouldn't it return 0 results?

test/DynamoCoreWpfTests/NodeAutoCompleteSearchTests.cs
zeusongit: `string` type is now skipped

aparajit-pratap: I think the first input type is var[]..[] and the second is string, so in both cases, the type is skipped, which should mean 0 results, or am I missing something?

aparajit-pratap: Or this should return only the default types, which are around 6 or so.

test/DynamoCoreWpfTests/NodeAutoCompleteSearchTests.cs
zeusongit: `var[]..[]` is not in the skipped type list. `var` is skipped as of now.
Do you want to add `var[]..[]` as well?

test/DynamoCoreWpfTests/NodeAutoCompleteSearchTests.cs
zeusongit: there are now 70 suggestions for `var[]..[]`

test/DynamoCoreWpfTests/NodeAutoCompleteSearchTests.cs
zeusongit: `var[]..[]` is not skipped

test/DynamoCoreWpfTests/NodeAutoCompleteSearchTests.cs
aparajit-pratap: I think @mjkkirschner made a change that strips the array rank from the type. But in any case, I think you should strip the rank and skip just the type. That way it will skip all rank variations, meaning var[], var[][], etc.

test/DynamoCoreWpfTests/NodeAutoCompleteSearchTests.cs
aparajit-pratap: Well, as per my comment above, it should be skipped.

test/DynamoCoreWpfTests/NodeAutoCompleteSearchTests.cs
zeusongit: But isn't that(`var[]..[]`) already being asserted in line 160?
Or are you suggesting changing their order?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -85,9 +85,17 @@ internal IEnumerable<NodeSearchElement> GetMatchingSearchElements()
...
             var elements = new List<NodeSearchElement>();
             var inputPortType = PortViewModel.PortModel.GetInputPortType();
+
+            //List of nodes that are skipped temporarily, and will display list of default suggestions instead.
+            var skippedInputTypes = new List<string>() { "var", "object", "string", "bool", "int", "double" };
+
+            if (inputPortType == null || (skippedInputTypes.Any(s => s == inputPortType.ToString())))
+            {
+                return elements; 
+            }
```
aparajit-pratap, line 95: I think these lines of code need to be moved down to line 104, i.e. after lines:
```
//if parsing the type failed, revert to original string.
inputPortType = ast != null ? ast.datatype.Name : inputPortType;
```
At this point, the rank is stripped from the type info.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -85,9 +85,17 @@ internal IEnumerable<NodeSearchElement> GetMatchingSearchElements()
...
+
             var core = dynamoViewModel.Model.LibraryServices.LibraryManagementCore;
 
-            if (inputPortType == null) return elements;
```
aparajit-pratap, line 90: Once you move the lines above, this line would need to be restored.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
QilongTang: @aparajit-pratap Can you explain why you prefer moving the code in details? If we are simply ignoring these types, why run the algorithm against these types at all?

test/DynamoCoreWpfTests/NodeAutoCompleteSearchTests.cs
zeusongit: in this case, do you want me to include another node, that does show some suggestions and then assert it, or asserting 0 suggestions would suffice?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -102,6 +102,12 @@ internal IEnumerable<NodeSearchElement> GetMatchingSearchElements()
             //if parsing the type failed, revert to original string.
             inputPortType = ast != null ? ast.datatype.Name : inputPortType;
 
+            //check if the input port return type is in the skipped input types list
+            if (skippedInputTypes.Any(s => s == inputPortType.ToString()))
```
aparajit-pratap, line 106: inputPortType is already a string

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -85,15 +85,29 @@ internal IEnumerable<NodeSearchElement> GetMatchingSearchElements()
         {
             var elements = new List<NodeSearchElement>();
             var inputPortType = PortViewModel.PortModel.GetInputPortType();
+
+            //List of nodes that are skipped temporarily, and will display list of default suggestions instead.
```
QilongTang, line 89: List of types?

QilongTang:
@zeusongit Nice job, once the comment updated, I think we can get this in

aparajit-pratap (approved):

QilongTang:
@zeusongit Do you mind cherry pick this one to 2.9.0 branch?

PR #11215, opened 2020-10-28
Avoid PATH overflow from SetupPython calls

mmisol:
### Purpose

Calling Installer.SetupPython from Python.Included makes sure Python is
installed and also adds the install location to the PATH env variable.
The location is added to the PATH even if it was already there, so this
can potentially lead to an env variable overflow with the message
'Environment variable name or value is too long', after which the only
way to get CPython3 back to work would be to restart Dynamo.

The problem is avoided by remembering if Python was setup using a local
field. This should reduce the number of calls to a maximum of 2, which
is not the best (1 would be) but is far from resulting in an overflow.

### Declarations

### DIFF
--- a/src/Libraries/DSCPython/CPythonEvaluator.cs
+++ b/src/Libraries/DSCPython/CPythonEvaluator.cs
@@ -178,7 +178,7 @@ namespace DSCPython
                 return null;
             }
 
-            Python.Included.Installer.SetupPython().Wait();
+            InstallPython();
 
             if (!PythonEngine.IsInitialized)
             {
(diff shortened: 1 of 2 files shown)


mjkkirschner:
@mmisol is it worth adding a test for this functionality to preserve this behavior in the future or you are satisfied with the comment?

mmisol:
Reproducing this with a unit test would be easy too, you would only need to execute code with CPython3 in a loop until you run into the overflow. I thought about it and went with the comment, since it seemed cleaner than that test. Let me know what you think.

mjkkirschner (approved):

reddyashish (approved):

mmisol:
Strange. The build is this https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/1689/ and it passed but somehow it got disconnected from the PR. Merging.

PR #11224, opened 2020-11-02
Fix autocomplete popup - #3

aparajit-pratap:
### Purpose

Attempt at fixing autocomplete popup placement one final time. Based on discussion with @zeusongit we decided that the `CustomPopupPlacementCallback` approach seemed most promising that would work on most machines. This also fixes an existing issue where the autocomplete popup doesn't show up on opening a new workspace.

### Declarations


src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml.cs
```diff
@@ -37,21 +37,9 @@ public NodeAutoCompleteSearchControl()
...
         }
 
-        private void NodeAutoCompleteSearchControl_Loaded(object sender, RoutedEventArgs e)
-        {
-            if (ViewModel != null && ViewModel.PortViewModel != null)
-            {
-                ViewModel.PortViewModel.PlaceNodeAutocompleteWindow(this, e);
-                Analytics.TrackEvent(
-                Dynamo.Logging.Actions.Open,
-                Dynamo.Logging.Categories.NodeAutoCompleteOperations);
```
QilongTang, line 51: @aparajit-pratap @zeusongit I'm fine removing the loaded function but we need to move the analytics call to a different place

aparajit-pratap: I noticed an issue with this loaded function. It wasn't getting called for some reason after opening a new workspace, which is why I decided to remove it. I didn't realize analytics info was added here. Sorry for overlooking that.

aparajit-pratap:
I will also take a look at the test failure.

src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml.cs
QilongTang: @aparajit-pratap @zeusongit No problem. There is a show/hide place of code we can move to, see `OnNodeAutoCompleteSearchControlVisibilityChanged()`. May need to do the same change for in-canvas search as well.. if this loaded function is not reliable, because @zeusongit added that for in canvas search following your example here

zeusongit (approved):

src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml.cs
aparajit-pratap: @zeusongit what event exactly does the `analytics.trackevent` aim to track here - the autocomplete popup popping up?

QilongTang: @aparajit-pratap Yes, so that we know how frequent users using this comparing to in canvas search so we can make improvements

src/DynamoCoreWpf/Views/Core/WorkspaceView.xaml.cs
```diff
@@ -186,7 +187,14 @@ private void ShowHidePopup(ShowHideFlags flag, Popup popup)
...
                 case ShowHideFlags.Show:
                     // Show InCanvas search just in case, when mouse is over workspace.
-                    popup.IsOpen = DynamoModel.IsTestMode || IsMouseOver;
+                    var displayPopup = DynamoModel.IsTestMode || IsMouseOver;
+                    if (displayPopup && popup == NodeAutoCompleteSearchBar)
+                    {
+                        ViewModel.NodeAutoCompleteSearchViewModel.PortViewModel.SetupNodeAutocompleteWindowPlacement(popup);
+                    }
+                    popup.IsOpen = displayPopup;
+                    popup.CustomPopupPlacementCallback = null;
```
QilongTang, line 196: Is this a must have or just sanity cleanup?

aparajit-pratap: Yes, otherwise the `CustomPopupPlacementCallback` continues pointing to `PlaceAutocompletePopup`, which can be invoked unexpectedly, say when in-canvas search popup is opened.

QilongTang (approved):

QilongTang:
Changes here looks good to me, but we should fix the analytics somehow and make sure we unify the implementation between incanvas search and node autocomplete. @aparajit-pratap @zeusongit

aparajit-pratap:
> Can we leave this on the control level? I think it may live better there which in canvas control may just follow same example.

Where exactly would you suggest moving this to?

QilongTang:
> > Can we leave this on the control level? I think it may live better there which in canvas control may just follow same example.
> 
> Where exactly would you suggest moving this to?

https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml.cs#L136

QilongTang:
@aparajit-pratap Didn't you say the loaded function is not always called? Then why bring that implementation back?

aparajit-pratap:
> @aparajit-pratap Didn't you say the loaded function is not always called? Then why bring that implementation back?

I'm not sure - it seems like it's now being invoked because the popup does appear after moving the placement from the loaded function to explicitly calling it [here]([url]([url](https://github.com/DynamoDS/Dynamo/pull/11224/files#diff-52d5dfb966e6deb4517a57cbd86209a747642ee01574f300bc0cab06452285c0R196))).

QilongTang:
@aparajit-pratap Regardless of that, the current implementation for analytics looks wrong in this PR because the if check below can't be removed, otherwise initialization of the control will also trigger analytics which gives us an extra use of node autocomplete everytime user launches Dynamo. Please add it back and to the cherry-pick PR. Thanks
```
if (ViewModel != null && ViewModel.PortViewModel != null)
```

aparajit-pratap:
@QilongTang the above condition will always be true in the new case.

QilongTang:
> @QilongTang the above condition will always be true in the new case.

OK, then maybe because I am debugging the old version. All good

QilongTang:
> @QilongTang the above condition will always be true in the new case.

Must be I was debugging the old version, but anyway this brings a bug that extra use is reported from analytics..

aparajit-pratap:
My debugging indicates that the `loaded` callback is being called just once for every port click and at all times both `ViewModel` and `ViewModel.PortViewModel` are non-null so if you're still seeing extra uses of analytics being reported, then even if I add the null-check, it might not fix the extra reporting. This could mean that I need to try adding the analytics tracking at some other place.

QilongTang:
> My debugging indicates that the `loaded` callback is being called just once for every port click and at all times both `ViewModel` and `ViewModel.PortViewModel` are non-null so if you're still seeing extra uses of analytics being reported, then even if I add the null-check, it might not fix the extra reporting. This could mean that I need to try adding the analytics tracking at some other place.

The loaded function is always called for me when Dynamo is started, see below gif
NodeAutoCompleteInitialization

I think that is why I was say earlier that let's move the tracking to https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml.cs#L136 because visibility change is a safer place IMHO. Let me know, if you are not comfortable doing it, I can make a PR because the change need to be done for in canvas search as well.

aparajit-pratap:
@QilongTang I'm not confident of how many times and when `OnNodeAutoCompleteSearchControlVisibilityChanged` is invoked. As a correction to my above statement, I'm adding back the null checks to the `loaded` method. Please let me know if that is acceptable.

QilongTang:
> @QilongTang I'm not confident of how many times and when `OnNodeAutoCompleteSearchControlVisibilityChanged` is invoked. As a correction to my above statement, I'm adding back the null checks to the `loaded` method. Please let me know if that is acceptable.

From my debugging, that is acceptable
image

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
```diff
@@ -563,6 +563,9 @@ public double Y
             }
         }
 
+        internal double ActualHeight { get; set; }
```
mjkkirschner, line 566: Just curious, can you explain why these were needed?

aparajit-pratap: These were needed to get the exact height and width from the node view and exposed to the portviewmodel via the nodeviewmodel. The width and height are needed to compute the popup placement positioning.

PR #11232, opened 2020-11-03
Change analytics code block position for both controls, Nodeautocomplete and Incanvas search

zeusongit:
### Purpose

ADP tracking for in-canvas search control was being recorded incorrectly as it was also triggered right after startup.
Therefore, it is now moved to the function that handles visibility change for both the controls to maintain uniformity and accurate hits.

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml.cs
+++ b/src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml.cs
@@ -37,18 +37,7 @@ namespace Dynamo.UI.Controls
             {
                 Application.Current.Deactivated += currentApplicationDeactivated;
             }
-            Loaded += NodeAutoCompleteSearchControl_Loaded;
             Unloaded += NodeAutoCompleteSearchControl_Unloaded;
-        }
-
-        private void NodeAutoCompleteSearchControl_Loaded(object sender, RoutedEventArgs e)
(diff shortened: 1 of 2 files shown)


src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml.cs
```diff
@@ -134,6 +123,10 @@ private void OnNodeAutoCompleteSearchControlVisibilityChanged(object sender, Dep
             // When launching this control, always start with clear search term.
             SearchTextBox.Clear();
 
+            Analytics.TrackEvent(
```
aparajit-pratap, line 126: Doesn't this get called twice - once when the control becomes visible and once when it gets hidden?

zeusongit: `if (!(bool)e.NewValue)`
this takes care of those cases when control hides.

QilongTang: @aparajit-pratap See line 120

QilongTang (approved):
LGTM

QilongTang:
@zeusongit Let's merge this and cherry-pick to 2.9.0 branch

PR #11244, opened 2020-11-12
CustomNodes that replicate do not access trace correctly.

mjkkirschner:
### Purpose
image

https://jira.autodesk.com/browse/DYN-1726

I found while investigating this issue that during execution of a nested node which accesses trace and creates elements (lets call this `nodeT`) - that the callsite which is created for `nodeT` contains an `invokeCount` field - this field is incremented each time the callsite is accessed "within a single run".

Unfortunately, when `nodeT` is nested inside a customNode or any DSFunction which is represented by a `JILFunctionEndpoint` - the `invokeCount` field is reset to 0 when the function is called.

This means that when a `JILFunctionEndPoint` is replicated, each replicated call, results in nested function calls reseting their callsites's `invokeCount` to 0.

When this occur the resulting execution results and trace data are both replaced with the last function call result.

Update:

I've fallen back to adding a new internal field to RuntimeCore - `LastDispatchedCallSite` which is set to the `this` inside of `DispatchNew`- before resetting the `invokeCount` we check if the last callsite was the same as the current callsite - if this is the case we're very likely in some type of replicated or nested replicated call - so don't reset the invoke count. 

When the callsites are out of sync we've moved on to a new callsite, and we **should reset** the invoke count - this takes care of the cases where we execute the graph more than once, or execute a function using associative update (without the execution actually ending, and yet still returning to a previous callsite).

I tried to use the existing property `RuntimeCore.DSExecutable.ExecutinGraphNode` - but by the time we can check this, it's already been set to the same graphnode that represents the current callsite, so it's not useful for this check.

I'd like to return to this in the future to investigate why the entire contents of `InterperterProperties` is always null, because it contains a `stack` of `executingGraphNodes` which should be able to tell us what `graphNode`s executed in what order - but it seems to be reset to null for a reason in multiple places, and appears to hold up lots of other things, like callsite ids... I think refactoring that is out of scope for now, and I will file a task to investigate it. Unfortunately, it looks like it is unchanged from when Luke first merged it into Dynamo in 2014, so no helpful git history.

### Declarations

### DIFF
--- a/src/Engine/ProtoCore/Lang/CallSite.cs
+++ b/src/Engine/ProtoCore/Lang/CallSite.cs
@@ -708,7 +708,7 @@ namespace ProtoCore
         /// </summary>
         /// <param name="methodName"></param>
         /// <param name="core"></param>
-        internal void UpdateCallsiteExecutionState(Object callsiteData, RuntimeCore runtimeCore)
+        private void UpdateCallsiteExecutionState(Object callsiteData, RuntimeCore runtimeCore)
         {
         
             Debug.WriteLine($"resetting callsite invoke count for {this.methodName}");
--- a/src/Engine/ProtoCore/Lang/CallSite.cs
+++ b/src/Engine/ProtoCore/Lang/CallSite.cs
@@ -1409,14 +1409,18 @@ namespace ProtoCore
             DominantListStructure domintListStructure,
             StackFrame stackFrame, RuntimeCore runtimeCore)
         {
-
-            if(runtimeCore.FunctionCallDepth < 2)
+           
+            // if the last dispatched callsite is this callsite then we are repeatedly making calls
+            // to this same callsite (for example replicating over an outer function that contains this callsite)
+            // and should not reset the invoke count.
+            if (runtimeCore.LastDispatchedCallSite != this)
             {
                 UpdateCallsiteExecutionState(null, runtimeCore);
             }
-           
+            runtimeCore.LastDispatchedCallSite = this;
 
-            //TODO reuse this.
+
+            //TODO reuse this when we have time to profile it.
             Stopwatch sw = new Stopwatch();
             sw.Start();
 
--- a/src/Engine/ProtoCore/Lang/CallSite.cs
+++ b/src/Engine/ProtoCore/Lang/CallSite.cs
@@ -76,7 +76,7 @@ namespace ProtoCore
 
                     if (HasData)
                         return true;
-                    
+
                     //Not empty, and doesn't have data so test recursive
                     Validity.Assert(NestedData != null,
                         "Invalid recursion logic, this is a VM bug, please report to the Dynamo Team");
@@ -245,10 +245,10 @@ namespace ProtoCore
                 var assemblyNameObj = new AssemblyName(assemblyName);
                 //find matching assemblies by name, version is not used.
                 var matchingAssembly = assemblies.FirstOrDefault(x => x.GetName().Name == assemblyNameObj.Name);
-                if(matchingAssembly!= null)
+                if (matchingAssembly != null)
                 {
-                   var matchingType = matchingAssembly.GetType(typeName);
-                    if(matchingType != null)
+                    var matchingType = matchingAssembly.GetType(typeName);
+                    if (matchingType != null)
                     {
                         return matchingType;
                     }
@@ -708,9 +708,11 @@ namespace ProtoCore
         /// </summary>
         /// <param name="methodName"></param>
         /// <param name="core"></param>
-        private void UpdateCallsiteExecutionState(Object callsiteData, RuntimeCore runtimeCore)
+        internal void UpdateCallsiteExecutionState(Object callsiteData, RuntimeCore runtimeCore)
         {
+
             invokeCount = 0;
+            Debug.WriteLine("resetting callsite invoke count");
 
             /*
             if (core.EnableCallsiteExecutionState)
@@ -907,7 +909,7 @@ namespace ProtoCore
                             if (replicationInstructions == null ||
                                 IsSimilarOptionButOfHigherRank(replicationInstructions, replicationOption))
                             {
-                                resolvedFeps = new List<FunctionEndPoint>() {compliantTarget};
+                                resolvedFeps = new List<FunctionEndPoint>() { compliantTarget };
                                 replicationInstructions = replicationOption;
                                 matchFound = true;
                             }
@@ -948,7 +950,7 @@ namespace ProtoCore
                         if (replicationInstructions == null ||
                             IsSimilarOptionButOfHigherRank(replicationInstructions, replicationOption))
                         {
-                            resolvedFeps = new List<FunctionEndPoint>() {compliantTarget};
+                            resolvedFeps = new List<FunctionEndPoint>() { compliantTarget };
                             replicationInstructions = replicationOption;
                             matchFound = true;
                         }
@@ -1406,10 +1408,9 @@ namespace ProtoCore
             DominantListStructure domintListStructure,
             StackFrame stackFrame, RuntimeCore runtimeCore)
         {
-            // Update the CallsiteExecutionState with 
-            // TODO: Replace this with the real data
-            UpdateCallsiteExecutionState(null, runtimeCore);
 
+
+            //TODO reuse this.
             Stopwatch sw = new Stopwatch();
             sw.Start();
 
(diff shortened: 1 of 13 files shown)


mjkkirschner:
I'm moving this PR to open - because it looks like the API used to check for all pullrequests does not treat draft prs as "open"

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -1406,10 +1408,9 @@ private FunctionGroup FirstFunctionGroupInInheritanceChain(RuntimeCore runtimeCo
             DominantListStructure domintListStructure,
             StackFrame stackFrame, RuntimeCore runtimeCore)
         {
-            // Update the CallsiteExecutionState with 
-            // TODO: Replace this with the real data
-            UpdateCallsiteExecutionState(null, runtimeCore);
```
mjkkirschner, line 1411: this is the first real change in this PR.

src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -1352,6 +1352,21 @@ private void PostExecution()
         {
             ApplyUpdate();
             HandleWarnings();
+            //reset callsite invocation count after run.
+            ResetCallsiteExecutionState();
```
mjkkirschner, line 1356: second real change.

mjkkirschner:
There are some failing test cases here - that seem to use associative update to trigger executions of those same FEPs - the test has a comment that seems to conflict with what I understand about callsites - it needs more through if this solution will work.

I'm not sure supporting associative update matters any longer as you can no longer redefine variables in associative codeblocks - obviously the test still runs though when run through the protoscript runner just fine... 😕

src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -1352,6 +1352,21 @@ private void PostExecution()
...
+        /// <summary>
+        /// Resets Callsites Execution state - does not reset trace data.
+        /// </summary>
+        private void ResetCallsiteExecutionState()
+        {
+            //TODO what about other ds runners?
+            //TODO move to internal function in RuntimeData object.
+            runtimeCore.RuntimeData.CallsiteCache.Values.ToList().ForEach(x =>
+            {
+                x.UpdateCallsiteExecutionState(null, runtimeCore);
```
aparajit-pratap, line 1368: How about simply resetting for each call site just before returning from `DispatchNew`?

mjkkirschner: unfortunately as discussed this does not work.

mjkkirschner:
much easier to review this with whitespace changes off:
https://github.com/DynamoDS/Dynamo/pull/11244/files?diff=unified&w=1

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -1406,10 +1409,18 @@ private FunctionGroup FirstFunctionGroupInInheritanceChain(RuntimeCore runtimeCo
...
+           
+            // if the last dispatched callsite is this callsite then we are repeatedly making calls
+            // to this same callsite (for example replicating over an outer function that contains this callsite)
+            // and should not reset the invoke count.
+            if (runtimeCore.LastDispatchedCallSite != this)
```
aparajit-pratap, line 1416: Did this not work:
```
if(this.id != runtimeCore.DSExecutable.ExecutingGraphnode.CallsiteIdentifier) {...}
```

mjkkirschner: yes, as stated above, I found this did not work - the `ExecutingGraphNode` on the executable has already been set to the current callsite - so it's too late, we really need the `LastGraphNode` - not the current one.

mjkkirschner:
I'm looking into the test failure - it seems unrelated to these changes as it deals with serialization of inputs... but will track it down.

mjkkirschner:
@aparajit-pratap @pinzart - test failure should be resolved latest commit - I believe what was happening was as follows:

The serialization tests grab the first 300 dyn files in the repo - I added new .dyn files, so this changed which tests ran.
Some of the tests don't fully run, for example if they have dummy nodes.
There is a map that contains guids and ids that we some tests may update to fully test parts of id deserialization.
This map was not being cleared so previous test runs were affecting other tests -
This could happen for a few different reasons - like if a test did not fully run, or if the last test run had changed between test methods - because the act of adding new test files changed which tests actually ran.

In any case, to avoid this in the future, all serialization tests now clear this map before the run.

pinzart90:
@mjkkirschner @aparajit-pratap  I may have this wrong, but here's my 2 cents:

Each function call (represented by a unique callsite), from running a graph, will generate trace data (persisted in callsite). This trace data is used to enable binding to elements.
So if a function call - callsite X - (with replication or without) generates n number of traces, then it could bind to n number of elements. - I think....
If you have a subsequent call to callsite X that generates n + 1 traces, then we should reuse the first n preexisting traces, and generate 1 additional one.
I think this is where the invokeCount mechanism steps in.

Regarding the invokeCount mechanism: 
From what I can tell, invokeCount is supposed to represent the number of times a callsite has been called into during one execution of the graph. 
This is a bit weird because invokeCount is persisited in callSites, but it seems to be more suited as transitory data (execution context). Anyway...

I think the design was to set invokeCount to 0 when first calling into a function or callsite. Then keep incrementing invokeCount for however many times we replicate that call.
I guess no one accounted for the fact that recursion can reset the invokeCount.

Regarding your change:
If you take the simple case of a non replicating function call foo().
Because invokeCount is persited data (callsites are persisted too) then each Graph run will keep incrementing invokeCount with no reset (because this === LastDispatchedCallSite).
This in turn will cause the callsite to keep adding trace data after each execution.
Execute the same function 5 times will result in invokeCount=5 and 5 traceData. 
Before your change,  invokeCount was always reset to 0.

sm6srw (approved):
@pinzart90 summed it up quite good. To me, it looks like this scenario has never worked as intended and was never considered when invokeCount was implemented.  I think this proposed fix is OK for now and we can decide to do a bigger refactoring later if and when we run into any new regressions because of it. LGTM!

pinzart90:
What I would highlight is that if you have a graph with a single node (with single callsite) that is supposed to do element binding, the your PR will change the behavior (with your changes, each execution of the graph will also create a new trace data - which in turn would mean that a new element will be created). Can you test this case out @mjkkirschner  ?

aparajit-pratap (commented):
> Regarding your change:
> If you take the simple case of a non replicating function call foo().
> Because invokeCount is persited data (callsites are persisted too) then each Graph run will keep incrementing invokeCount with no reset (because this === LastDispatchedCallSite).

@pinzart ~~in this case, each graph run, even if it's to the same function will result in a new callsite so it will not match with `LastDispatchedCallSite`~~ I was wrong about this. Sorry.

mjkkirschner:
@pinzart I can test that case - that could be a problem yes, I will add a test -  one solution I guess is to reset `LastDispatchedCallsite` to null after the execution or at the start of an execution.

pinzart90:
> > Regarding your change:
> > If you take the simple case of a non replicating function call foo().
> > Because invokeCount is persited data (callsites are persisted too) then each Graph run will keep incrementing invokeCount with no reset (because this === LastDispatchedCallSite).
> 
> @pinzart in this case, each graph run, even if it's to the same function will result in a new callsite so it will not match with `LastDispatchedCallSite`

From what I've seen, callsites are cached and stored in a map. If a subsequent run is made to the same function, it should re use an existing callsite. So a single function foo called multiple times with different input params should go to the same callsite. Is this not correct ?

aparajit-pratap:
> From what I've seen, callsites are cached and stored in a map. If a subsequent run is made to the same function, it should re use an existing callsite. So a single function foo called multiple times with different input params should go to the same callsite. Is this not correct ?

No, that isn't true.

mjkkirschner:
@aparajit-pratap - isn't it? 
The callsites are not recreated - the callsiteCache is used from inside the GetCallSite function and those existing callsites are retrieved, isn't that how trace works?

mjkkirschner:
@pinzart your example case fails - I will look into solutions, only thing that comes to mind is resetting the LastCallSite between executions. Any other ideas? 

I guess as you suggested above moving the invoke count out of the callsite altogether might be a refactor worth doing in the future?

src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -1354,6 +1354,7 @@ private void PostExecution()
             HandleWarnings();
         }
 
+
```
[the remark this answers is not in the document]
mjkkirschner: will do

pinzart90 (approved):
LGTM

aparajit-pratap (approved):

PR #11245, opened 2020-11-13
[SPIKE] Use Helix instancing to save memory for large number of mesh geometry

aparajit-pratap:
### Purpose

Trial implementation exploring using Helix instancing to render large number of mesh geometry.
This goes hand-in-hand with corresponding LibG changes here: https://git.autodesk.com/Dynamo/LibG/pull/1080

#### Assumptions
- This assumes that any node producing mesh geometry (solids and surfaces) and participating in replication can potentially benefit from instancing while being rendered.

#### Issues
- Instances do not currently display in the background preview for some reason
- Geometry transforms are incorrect in LibG (e.g. if a Sphere is located away from the origin, it's transform matrix does not reflect its exact position)

### Declarations

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
@@ -37,6 +37,7 @@ using SharpDX;
 using Color = SharpDX.Color;
 using ColorConverter = System.Windows.Media.ColorConverter;
 using GeometryModel3D = HelixToolkit.Wpf.SharpDX.GeometryModel3D;
+using Matrix = SharpDX.Matrix;
 using MeshBuilder = HelixToolkit.Wpf.SharpDX.MeshBuilder;
 using MeshGeometry3D = HelixToolkit.Wpf.SharpDX.MeshGeometry3D;
 using PerspectiveCamera = HelixToolkit.Wpf.SharpDX.PerspectiveCamera;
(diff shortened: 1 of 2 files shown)


src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -1629,6 +1630,45 @@ internal virtual void AggregateRenderPackages(IEnumerable<HelixRenderPackage> pa
...
+                            ? p.Description : baseId) + MeshKey;
+
+                        InstancingMeshGeometryModel3D instancingModel;
+                        if (Element3DDictionary.TryGetValue(id, out var element3D))
+                        {
+                            instancingModel = element3D as InstancingMeshGeometryModel3D;
+                        }
+                        else
+                        {
+                            instancingModel =  new InstancingMeshGeometryModel3D();
```
mjkkirschner, line 1655: hmm - we should inspect the EffectsManager at runtime to see if theres an effect for instancing meshes - it may try to use our shader.

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -1629,6 +1630,45 @@ internal virtual void AggregateRenderPackages(IEnumerable<HelixRenderPackage> pa
...
+                {
+                    var p = pkgs[0];
+                    if (p.Mesh != null && p.Mesh.Positions.Any())
+                    {
+                        var baseId = p.Description;
+                        if (baseId.IndexOf(":", StringComparison.Ordinal) > 0)
+                        {
+                            baseId = baseId.Split(':')[0];
+                        }
+                        var id = ((p.RequiresPerVertexColoration || p.Colors != null || p.RequiresCustomTransform)
```
mjkkirschner, line 1645: does this somehow conflict with line this line below:
` id = ((rp.RequiresCustomTransform) ? rp.Description : baseId) + PointsKey;`

mjkkirschner:
@aparajit-pratap I think in our vertex shaders we'll need to add something like this:

https://github.com/DynamoDS/Dynamo/blob/RC2.5.0_master/src/DynamoCoreWpf/ViewModels/Watch3D/Shaders/Custom.fx#L157

mjkkirschner:
Another thought on this type of approach - opposed to just replicated calls, it would also be good to try this on calls like move or scale, rotate etc- where instancing and transforming is exactly what the node does and instancing will definitely work.

PR #11246, opened 2020-11-13
Undock/redock extension tabs to windows

mmisol:
### Purpose

Adds the ability to undock an extension tab from the side bar to a
window. This way the extension control can be moved and resized without
limits beyond actual screen size. Once in window state, the content can
also be docked back to the side bar or just closed.

Here is a screenshot of the change:
Screen Shot 2020-11-13 at 2 14 12 PM

### Declarations


src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -222,7 +228,7 @@ internal TabItem AddExtensionTabItem(IViewExtension viewExtension, ContentContro
                 // creates a new tab item
                 tab = new TabItem();
                 tab.Header = viewExtension.Name;
-                tab.Tag = viewExtension.GetType();
+                tab.Tag = viewExtension;
```
mjkkirschner, line 231: Can we consider if this causes any memory leak issues?

mjkkirschner:
What happens if a user invokes a menu item which adds the extension to the sidebar again while it is already popped out?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
mmisol: I don't see how to be honest. Did you see a problem?

mjkkirschner: Not really - tab's lifetime should be the same as the extension in most cases, maybe shorter.
In a testing scenario I could see it being an issue if the UI is already leaking, and as a consequence tab is never cleaned up, now viewExtension will not be cleaned up either.

mmisol:
> What happens if a user invokes a menu item which adds the extension to the sidebar again while it is already popped out?

This is a good point. For the extensions I have seen, they actually change their menu item behaviour to show/hide according to whether the extension is already shown. However, that does not necessarily mean that they can't misbehave and try to add another tab/window. Looking at the code, I think I don't currently cover this case, so I'll see to do it.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
mmisol: If we are leaking tabs we are probably also leaking the controls within them. The `viewExtension` object might add a bit more, but I'd say most of the harm would already be done.

mmisol:
Should be good now. Thanks for reviewing @mjkkirschner

mjkkirschner (approved):

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -273,6 +290,7 @@ internal void CloseExtensionTabItem(IViewExtension viewExtension)
             TabItem tabitem = ExtensionTabItems.OfType<TabItem>().SingleOrDefault(n => n.Header.ToString() == tabName);
             CloseExtension?.Invoke(tabName);
             CloseExtensionTab(tabitem);
+            CloseExtensionWindow(tabName);
```
QilongTang, line 293: Can you update the function description as well?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -322,6 +340,86 @@ private void CloseExtensionTab(TabItem tabitem)
...
+        /// Undocks the extension with the given name.
+        /// Made internal for testing purposes only.
+        /// </summary>
+        /// <param name="name">Name of the extension</param>
+        internal void UndockExtension(string name)
+        {
+            var tabItem = ExtensionTabItems.OfType<TabItem>().SingleOrDefault(tab => tab.Header.ToString() == name);
+            CloseExtensionTab(tabItem);
+
+            var ext = new ExtensionWindow();
```
QilongTang, line 369: Curious if we can reuse the extension window so that we do not need to always create new ones, but given it is only a frame, seems no harm yet?

mmisol: They should be GC'ed just like tabs. Recycling them would be cumbersome and wouldn't bring any benefit IMO.

QilongTang: @mmisol I'm cool with this implementation for now.

QilongTang:
@mmisol Can you check the build? 
```
"D:\a\Dynamo\Dynamo\Dynamo\src\DynamoCoreWpf\DynamoCoreWpf.csproj" (default target) (18:2) ->
(CoreCompile target) -> 
  Extensions\ViewLoadedParams.cs(120,24): error CS1061: 'DynamoView' does not contain a definition for 'CloseExtensionTabItem' and no accessible extension method 'CloseExtensionTabItem' accepting a first argument of type 'DynamoView' could be found (are you missing a using directive or an assembly reference?) [D:\a\Dynamo\Dynamo\Dynamo\src\DynamoCoreWpf\DynamoCoreWpf.csproj]
```

QilongTang (approved):

PR #11049, opened 2020-08-24
Changes to switch the LibG references

alfredo-pozo:
Please Note:
1. Before submitting the PR, please review [How to Contribute to Dynamo](https://github.com/DynamoDS/Dynamo/blob/master/CONTRIBUTING.md)
2. Dynamo Team will meet 1x a month to review PRs found on Github (Issues will be handled separately)
3. PRs will be reviewed from oldest to newest
4. If a reviewed PR requires changes by the owner, the owner of the PR has 30 days to respond. If the PR has seen no activity by the next session, it will be either closed by the team or depending on its utility will be taken over by someone on the team
5. PRs should use either Dynamo's default PR template or [one of these other template options](https://github.com/DynamoDS/Dynamo/wiki/Choosing-a-Pull-Request-Template) in order to be considered for review.
6. PRs that do not have one of the Dynamo PR templates completely filled out with all declarations satisfied will not be reviewed by the Dynamo team.
7. PRs made to the `DynamoRevit` repo will need to be cherry-picked into all the DynamoRevit Release branches that Dynamo supports. Contributors will be responsible for cherry-picking their reviewed commits to the other branches after a `LGTM` label is added to the PR.

### Purpose

This PR has the purpose to use the new LibG nuget packages instead of the references under "extern" folder https://jira.autodesk.com/browse/DYN-1889

### Declarations

Check these if you believe they are true

- [X] The codebase is in a better state after this PR
- [X] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [X] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [X] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.
- [ ] This PR modifies some build requirements and the readme is updated

### Reviewers

@aparajit-pratap 

### FYIs

@QilongTang

PR #11254, opened 2020-11-16
Revert "Fix for missing file ProtoGeometry.xml"

QilongTang:
Reverts DynamoDS/Dynamo#11218

FYI: @mjkkirschner Need to revert this one then https://github.com/DynamoDS/Dynamo/pull/11049


mjkkirschner:
I'm not clear we need to revert this one @QilongTang - I thought we still needed this to correctly consume protogeometry from nuget even without the PackageReferences changes?

aparajit-pratap:
@QilongTang I don't think we need to revert these PR's (#11218 and #11049) as these are still required for LibG NuGets.

QilongTang:
@aparajit-pratap Understood, but the master build is not in a release ready state after https://github.com/DynamoDS/Dynamo/pull/11049. We have known issues as:
1. extra binaries
2. Devs not easy to build it

After discussion with @mjkkirschner , we would like to revert it for now and put a more robust implementation in Dynamo 2.11.

aparajit-pratap:
@qilongtang I'm not sure I understand. We had been integrating libg from nugets for several weeks without any issues before we decided to go with the package references. And the only reason we explored package refs was because we thought we could avoid having to make manual PRs to Dynamo to integrate the latest LibG builds.

QilongTang:
Correction. We intend to revert https://github.com/DynamoDS/Dynamo/pull/11186 and deliver it part of Dynamo 2.11, not the libG nuget part

PR #11260, opened 2020-11-17
Adding a public API on the View extension that gets triggered when the view extension is closed.

reddyashish:
### Purpose

This PR is for the task: https://jira.autodesk.com/browse/DYN-3214

Closed the initial PR: https://github.com/DynamoDS/Dynamo/pull/11242 and created this based on the review comments.

We now have a ViewExtensionBaseClass which has the OnViewExtensionClosed method that will be triggered only for the view extension that is closed. So any view extension that extends this base class will have to implement its own OnViewExtensionClosed method and the extension authors can perform any actions in here when the view extension is closed. 

### Declarations

### Reviewers

@QilongTang @mmisol @mjkkirschner 

### FYIs

@DynamoDS/dynamo @saintentropy


src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
```diff
@@ -14,7 +14,7 @@ namespace Dynamo.DocumentationBrowser
     /// The DocumentationBrowser view extension displays web or local html files on the Dynamo right panel.
     /// It reacts to documentation display request events in Dynamo to know what and when to display documentation.
     /// </summary>
-    public class DocumentationBrowserViewExtension : IViewExtension, ILogSource
+    public class DocumentationBrowserViewExtension : IViewExtensionBaseClass, IViewExtension, ILogSource
```
reddyashish, line 17: Extending a new class to a public class wouldn't be a breaking change, right?

QilongTang: Should be fine, I think that's how we used to do it. @mjkkirschner Please confirm?

mjkkirschner: `IViewExtensionBaseClass` is a strange name the `I` usually signifies Interface - I think this is against c# convention.

mjkkirschner: hmmm- it doesn't sound good...

❓ REQUIRES JUDGMENT: Introducing a new base class

A type can be introduced into a hierarchy between two existing types if it doesn't introduce any new abstract members or change the semantics or behavior of existing types. For example, in .NET Framework 2.0, the DbConnection class became a new base class for SqlConnection, which had previously derived directly from Component.


we'll have to do some reading to understand why - @mmisol any thoughts?

mjkkirschner: ahh, I think I understand - if we do this - we should provide empty default implementations for all methods in the abstract base class, or other extensions that derive from our extensions will fail to implement that abstract class correctly.

On the other hand - I don't know if we consider our Extension classes as part of our API - that would be a great thing to document somewhere.

reddyashish: Yes, changed the name to ViewExtensionBaseClass. This has empty default implementations for all the methods from the interface.

src/DynamoCoreWpf/Extensions/IViewExtension.cs
```diff
@@ -51,4 +51,77 @@ public interface IViewExtension: IDisposable
...
+        /// </summary>
+        public void Startup(ViewStartupParams viewStartupParams)
+        {
+           
+        }
+
+        /// <summary>
+        /// Action to be invoked when the view extenion is closed. 
+        /// </summary>
+        public abstract void OnViewExtensionClosed(String extensionTabName);
```
mjkkirschner, line 125: I mean it should also have empty default implementations for abstract members.

reddyashish: May be I am missing something, can an abstract method have an implementation body?

mjkkirschner: you're right, it's not possible - I'm not sure we want to use an abstract method then - maybe just virtual with a default implementation.

mmisol: I think it would be better if we followed what @mjkkirschner proposed and have something like `public virtual void Closed()` with an empty implementation. Maybe we also want to include a parameters object for future extensibility?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -320,7 +322,7 @@ private void CloseExtensionTab(TabItem tabitem)
             // get the selected tab
             TabItem selectedTab = tabDynamic.SelectedItem as TabItem;
 
-            if (tabToBeRemoved != null)
+            if (tabToBeRemoved != null && ExtensionTabItems.Count > 0)
```
mmisol, line 325: Is there a reason for adding this condition?

mmisol (changes requested):

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
reddyashish: When you close the documentation browser, it un-checks the menu item and that will in-turn call the MenuItemUnCheckedHandler() here: https://github.com/DynamoDS/Dynamo/blob/1e77d23ebd980ed42932ecc7e47dcae5234a486e/src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs#L83. So when it tries to close the side bar extension again, it was failing. This would just make sure there are extension tabs present before it closes any.

reddyashish:
Addressed the comments. Please take a look again.

mmisol:
I think we still want to address https://github.com/DynamoDS/Dynamo/pull/11260#discussion_r525554953.

Also, we should have unit tests for this.

reddyashish:
Added unit test.

mmisol:
Sorry I guess Mike's comment got lost in the other PR: https://github.com/DynamoDS/Dynamo/pull/11242#issuecomment-726315828

The thing is, at it stands now, the method does not feel idiomatic as `Loaded` and `Started` do. That's why `Closed` sounds cleaner.

Also, the string argument does not make sense currently as the implementation was fixed with respect to #11242. There is no need to check what's the name of the extension tab that was closed anymore. It should always be the one that belongs to the extension.

reddyashish:
Thanks Martin for the clarification. Changed it accordingly. Makes sense to not have the string argument anymore.

src/DynamoCoreWpf/Extensions/ViewExtensionBase.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// Exceptions thrown from this method will be caught by Dynamo and 
+        /// displayed.
+        /// </summary>
+        public virtual void Startup(ViewStartupParams viewStartupParams)
+        {
+
+        }
+
+        /// <summary>
+        /// Action to be invoked when the view extenion is closed.
```
mmisol, line 77: There is a small typo here: `extenion` => `extension`

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -302,10 +304,10 @@ internal void CloseExtensionControl(IViewExtension viewExtension)
         internal void CloseExtensionTab(object sender, RoutedEventArgs e)
         {
             string tabName = (sender as Button).DataContext.ToString();
-
-            CloseExtension?.Invoke(tabName);
-
             TabItem tabitem = ExtensionTabItems.OfType<TabItem>().SingleOrDefault(n => n.Header.ToString() == tabName);
+
+            var viewExtensionBaseClass = (ViewExtensionBase) tabitem.Tag;
```
mmisol, line 309: Thinking about casts like this one. We can't really be sure the extension is going to implement the base class. That means this can fail. What we should do is instead use the `is` or `as` operator and call the `Closed` function conditionally, based on whether the extension implements the base class or not. We should add a test for cases like this as well.

mmisol (commented):
Thanks @reddyashish . Sorry, but I just noticed an issue regarding backwards compatibility we should get fixed before delivering this.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
mjkkirschner: I've been liking this syntax lately:
```
 if (tabItem.Tag is ViewExtensionBase base)
        {
            yada yada with base
        }

```

reddyashish: Good point. Thanks for catching it. Changed it.

reddyashish: Regarding the test for these cases, since we removed the CloseExtension event implementation, I don't think we can add a test for this extensions workflow that do not implement this base class.

mmisol: @reddyashish Whether they implement the base class or not, extensions should still be able to call `CloseExtensionInSideBar`. While that won't result in a call to `Closed`, because it is not implemented, the extension should still be closed.

reddyashish: Yeah, we do have a test for that already. 
https://github.com/DynamoDS/Dynamo/blob/2453d84f8628fd5b764011913de5006839f848c7/test/DynamoCoreWpfTests/ViewExtensions/WorkspaceDependencyViewExtensionTests.cs#L118

src/DynamoCoreWpf/Extensions/IViewExtension.cs
reddyashish: Changed this.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
mmisol: But the workspace dependency extension now implements the base class with this PR. I'm asking for a test where that's not the case.

reddyashish: Added it.

mmisol (approved):
Thanks @reddyashish . This looks good to me.

QilongTang:
Hi @reddyashish Can you double check the unit test `APItoCloseViewExtensionTabTest` pass? I saw it failing on self CI

reddyashish:
@QilongTang Yes made the fix to that. Thanks for verifying it. Just noticed that, on the self-serve CI, the DynamoCoreWPFTests fixture was not running completely as it was throwing a socket exception so I missed this failing test. Tested it locally, will wait for the self-serve validation.

QilongTang:
This looks good. I would like validation from @saintentropy @nate-peters to see if it fits what you need

saintentropy:
@QilongTang This will work for us.  Just out of curiosity where this might get documented.  I assume release notes but anywhere else.  Updating the internal usage is helpful

QilongTang:
> @QilongTang This will work for us. Just out of curiosity where this might get documented. I assume release notes but anywhere else. Updating the internal usage is helpful

I think Release notes. We dont have a plan for updating the developer docs yet.

mjkkirschner:
It would be nice to update the DynamoSamples repo to show how to use this API and the new base class.

QilongTang:
@mjkkirschner I think we need a sample to demo view extension into right side bar anyway. I will log a Jira task

QilongTang:
> @mjkkirschner I think we need a sample to demo view extension into right side bar anyway. I will log a Jira task

https://jira.autodesk.com/browse/DYN-3278

PR #11261, opened 2020-11-18
update load asm binaries from registry logic

pinzart90:
### Purpose

https://jira.autodesk.com/browse/DYN-3180

Problem:
New adsk installer no longer adds records in the win registry under the product name key.
Ex: 
`Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Revit 2021` will no longer exist.

Acceptance Criteria:
1. API users should be able to find ASM install path by using the ProductCode (i.e GUID) of adsk products.
2. Dynamo Sandbox default asm loader behavior should find ASM locations by using the Product Names (so older installer and new installer behaviors should be handled)
3. New lookup behavior should not have a performance penalty.

Solutions:
1. Follwing method should allow API users to find asm by product code.
https://github.com/DynamoDS/Dynamo/blob/1d186a5c6e35b2ce0d103b152a5ab9140cf7a52c/src/Tools/DynamoInstallDetective/ProductLookUp.cs#L57-L63
2. Default ASM install locations will look through all registry records (at `Uninstall` reg key) and match against the DisplayName.
https://github.com/DynamoDS/Dynamo/blob/186aedf9659837edd7fe1116f9b7c53e1124af3a/src/Tools/DynamoInstallDetective/Utilities.cs#L67
3. Added Caching mechanism for Registry records.

Additional changes : moved Registry static methods into new RegUtils class

### Declarations

### Reviewers

### FYIs

### DIFF
--- a/src/Tools/DynamoInstallDetective/ProductLookUp.cs
+++ b/src/Tools/DynamoInstallDetective/ProductLookUp.cs
@@ -17,12 +17,20 @@ namespace DynamoInstallDetective
         {
             public RegistryCacher()
             {
-                EnableCache();
+                lock (mutex)
+                {
+                    cacheEnabled = true;
+                }
             }
 
             public void Dispose()
             {
-                DisableCache();
+                lock (mutex)
+                {
+                    cacheEnabled = false;
+                    cachedRecords.Clear();
+                    cachedRecords = null;
+                }
             }
         }
 
@@ -102,24 +110,6 @@ namespace DynamoInstallDetective
         public static IDisposable StartCache()
         {
             return new RegistryCacher();
-        }
-
-        public static void EnableCache()
-        {
-            lock (mutex)
-            {
-                cacheEnabled = true;
-            }
-        }
-
-        public static void DisableCache()
-        {
-            lock (mutex)
-            {
-                cacheEnabled = false;
-                cachedRecords.Clear();
-                cachedRecords = null;
-            }
         }
     }
 
--- a/src/Tools/DynamoInstallDetective/ProductLookUp.cs
+++ b/src/Tools/DynamoInstallDetective/ProductLookUp.cs
@@ -9,6 +9,117 @@ using Microsoft.Win32;
 
 namespace DynamoInstallDetective
 {
+    // Utility class for interacting with the windows registry.
+    internal static class RegUtils
+    {
+        // Utility class to enable/disable registry caching within a scope.
+        class RegistryCacher : IDisposable
+        {
+            public RegistryCacher()
+            {
+                EnableCache();
+            }
+
+            public void Dispose()
+            {
+                DisableCache();
+            }
+        }
+
+        public static string REG_KEY64 = @"SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\";
+        public static string REG_KEY32 = @"SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\Uninstall\";
+
+        // Cached data from windows registry. Consists of a dictionary that maps The product name/code to Display Name (Tuple.item1) and Install Path(Tuple.item2)
+        // Dictionary<ProductCode, (DisplayName, InstallPath)>
+        private static Dictionary<string, Tuple<string, string>> cachedRecords;
+        private static bool cacheEnabled = false;
+
+        private static readonly object mutex = new object();
+
+        public static RegistryKey OpenKey(string key)
+        {
+            var regKey = RegistryKey.OpenBaseKey(RegistryHive.LocalMachine, RegistryView.Registry64);
+            return regKey.OpenSubKey(key);
+        }
+        public static string GetInstallLocation(RegistryKey key)
+        {
+            if (key != null)
+                return key.GetValue("InstallLocation") as string;
+
+            return string.Empty;
+        }
+
+        public static string GetDisplayName(RegistryKey key)
+        {
+            if (key != null)
+                return key.GetValue("DisplayName") as string;
+
+            return string.Empty;
+        }
+
+        public static Dictionary<string, Tuple<string, string>> GetInstalledProducts()
+        {
+            lock (mutex)
+            {
+                if (cacheEnabled && cachedRecords != null)
+                {
+                    return cachedRecords;
+                }
+            }
+
+            var productInfo = new Dictionary<string, Tuple<string, string>>();
+
+            var key = OpenKey(REG_KEY64);
+            foreach(string s in key.GetSubKeyNames())
+            {
+                try
+                {
+                    var prodKey = key.OpenSubKey(s);
+                    var displayName = GetDisplayName(prodKey);
+                    var installLocation = GetInstallLocation(prodKey);
+                    if (displayName.Length > 0 && installLocation.Length > 0)
+                    {
+                        productInfo.Add(s, Tuple.Create(displayName, installLocation));
+                    }
+                }
+                catch (Exception)
+                { }
+            }
+
+            lock (mutex)
+            {
+                if (cacheEnabled)
+                {
+                    cachedRecords = new Dictionary<string, Tuple<string, string>>(productInfo);
+                }
+            }
+            return productInfo;
+        }
+
+        public static IDisposable StartCache()
+        {
+            return new RegistryCacher();
+        }
+
+        public static void EnableCache()
+        {
+            lock (mutex)
+            {
+                cacheEnabled = true;
+            }
+        }
+
+        public static void DisableCache()
+        {
+            lock (mutex)
+            {
+                cacheEnabled = false;
+                cachedRecords.Clear();
+                cachedRecords = null;
+            }
+        }
+    }
+
     /// <summary>
     /// Specifies an installed product
     /// </summary>
@@ -63,7 +174,7 @@ namespace DynamoInstallDetective
         IInstalledProduct GetProductFromProductCode(string productCode);
 
         /// <summary>
-        /// Returns product name list based on lookup criteria
+        /// Returns product name list using on lookup criteria based on product names.
         /// </summary>
         /// <returns>Product name list</returns>
         IEnumerable<string> GetProductNameList();
@@ -120,37 +231,12 @@ namespace DynamoInstallDetective
     /// </summary>
     public class InstalledProductLookUp : IProductLookUp
     {
-        const string REG_KEY64 = @"SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\";
-        const string REG_KEY32 = @"SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\Uninstall\";
-        
         /// <summary>
         /// Product name for lookup
         /// </summary>
         public string ProductLookUpName { get; private set; }
 
         private readonly Func<string, string> fileLocator;
-
-        static RegistryKey OpenKey(string key)
-        {
-            var regKey = RegistryKey.OpenBaseKey(RegistryHive.LocalMachine, RegistryView.Registry64);
-            return regKey.OpenSubKey(key);
-        }
-
-        static string GetInstallLocation(RegistryKey key)
-        {
-            if (key != null)
-                return key.GetValue("InstallLocation") as string;
-
-            return string.Empty;
-        }
-
-        static string GetDisplayName(RegistryKey key)
-        {
-            if (key != null)
-                return key.GetValue("DisplayName") as string;
-
-            return string.Empty;
-        }
 
         /// <summary>
         /// Implements a product look up algorithm based on registry key.
@@ -202,8 +288,9 @@ namespace DynamoInstallDetective
 
         public virtual IEnumerable<string> GetProductNameList()
         {
-            var key = OpenKey(REG_KEY64);
-            return key.GetSubKeyNames().Where(s => s.Contains(ProductLookUpName));
+            return RegUtils.GetInstalledProducts().Select(s => s.Value.Item1).Where(s => {
+                return s?.Contains(ProductLookUpName) ?? false;
+            });
         }
 
         public virtual bool ExistsAtPath(string basePath)
@@ -216,21 +303,21 @@ namespace DynamoInstallDetective
 
         public virtual string GetInstallLocationFromProductName(string name)
         {
-            var key = OpenKey(REG_KEY64 + name);
-            return GetInstallLocation(key);
+            var key = RegUtils.OpenKey(RegUtils.REG_KEY64 + name);
+            return RegUtils.GetInstallLocation(key);
         }
 
         public virtual string GetInstallLocationFromProductCode(string productCode, out string productName)
         {
-            string issProdKey = REG_KEY32 + productCode + "_is1";
-            var key = OpenKey(issProdKey);
+            string issProdKey = RegUtils.REG_KEY32 + productCode + "_is1";
+            var key = RegUtils.OpenKey(issProdKey);
             if (null == key)
             {
-                issProdKey = REG_KEY64 + productCode;
-                key = OpenKey(issProdKey);
+                issProdKey = RegUtils.REG_KEY64 + productCode;
+                key = RegUtils.OpenKey(issProdKey);
             }
-            productName = GetDisplayName(key);
-            return GetInstallLocation(key);
+            productName = RegUtils.GetDisplayName(key);
+            return RegUtils.GetInstallLocation(key);
         }
 
         public virtual string GetCoreFilePathFromInstallation(string installPath)
--- a/src/Tools/DynamoInstallDetective/ProductLookUp.cs
+++ b/src/Tools/DynamoInstallDetective/ProductLookUp.cs
@@ -39,7 +39,7 @@ namespace DynamoInstallDetective
 
         // Cached data from windows registry. Consists of a dictionary that maps The product name/code to Display Name (Tuple.item1) and Install Path(Tuple.item2)
         // Dictionary<ProductCode, (DisplayName, InstallPath)>
-        private static Dictionary<string, Tuple<string, string>> cachedRecords;
+        private static Dictionary<string, (string DisplayName, string InstallLocation)> cachedRecords;
         private static bool cacheEnabled = false;
 
         private static readonly object mutex = new object();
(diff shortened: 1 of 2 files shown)


src/Tools/DynamoInstallDetective/Utilities.cs
```diff
@@ -88,19 +88,22 @@ public static IEnumerable FindProductInstallations(string productSearchPattern,
...
-                        new KeyValuePair<string, Tuple<int, int, int, int>>(
-                        p.InstallLocation,
-                        p.VersionInfo));
+                return
+                    installs.Products.Select(
+                        p =>
+                            new KeyValuePair<string, Tuple<int, int, int, int>>(
+                            p.InstallLocation,
+                            p.VersionInfo));
+            }
```
pinzart90, line 106: I wrapped existing code in a `using (RegUtils.StartCache())` that will allow us to cache the Registry records within the scope

src/Tools/DynamoInstallDetective/ProductLookUp.cs
```diff
@@ -9,6 +9,118 @@
...
+        public static string GetDisplayName(RegistryKey key)
+        {
+            if (key != null)
+                return key.GetValue("DisplayName") as string;
+
+            return string.Empty;
+        }
+
+        // Returns all the products registered under "SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\" that have a valid DisplayName and an InstallLocation.
+        public static Dictionary<string, Tuple<string, string>> GetInstalledProducts()
```
pinzart90, line 61: this is the main change in this PR. 
We now iterate through all the registry subkeys and we return only those that have a DIsplayName and an InstalledLocation.

We will find asm location in by matching the DisplayName in the above records

src/Tools/DynamoInstallDetective/ProductLookUp.cs
```diff
@@ -202,8 +289,9 @@ public IInstalledProduct GetProductFromProductCode(string productCode)
 
         public virtual IEnumerable<string> GetProductNameList()
         {
-            var key = OpenKey(REG_KEY64);
-            return key.GetSubKeyNames().Where(s => s.Contains(ProductLookUpName));
+            return RegUtils.GetInstalledProducts().Select(s => s.Value.Item1).Where(s => {
```
pinzart90, line 292: Match against the displayName (i.e `s.Value.Item1`)

src/Tools/DynamoInstallDetective/ProductLookUp.cs
```diff
@@ -9,6 +9,118 @@
...
+
+        public static string REG_KEY64 = @"SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\";
+        public static string REG_KEY32 = @"SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\Uninstall\";
+
+        // Cached data from windows registry. Consists of a dictionary that maps The product name/code to Display Name (Tuple.item1) and Install Path(Tuple.item2)
+        // Dictionary<ProductCode, (DisplayName, InstallPath)>
+        private static Dictionary<string, Tuple<string, string>> cachedRecords;
+        private static bool cacheEnabled = false;
+
+        private static readonly object mutex = new object();
```
pinzart90, line 37: mutex is used to make static cache thread safe

src/Tools/DynamoInstallDetective/ProductLookUp.cs
```diff
@@ -120,38 +232,13 @@ public interface IProductCollection
...
-        const string REG_KEY32 = @"SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\Uninstall\";
-        
         /// <summary>
         /// Product name for lookup
         /// </summary>
         public string ProductLookUpName { get; private set; }
 
         private readonly Func<string, string> fileLocator;
 
-        static RegistryKey OpenKey(string key)
```
pinzart90, line 133: default access modifiers should be private...so we should be good with this change

src/Tools/DynamoInstallDetective/ProductLookUp.cs
```diff
@@ -9,6 +9,118 @@
...
+            }
+            return productInfo;
+        }
+
+        public static IDisposable StartCache()
+        {
+            return new RegistryCacher();
+        }
+
+        public static void EnableCache()
```
aparajit-pratap, line 105: Why not make `EnableCache`, `DisableCache`, and `cacheEnabled` private members of `RegistryCacher`? I think `RegistryCacher` should also have access to private members of `RegUtils`?

src/Tools/DynamoInstallDetective/ProductLookUp.cs
```diff
@@ -63,7 +175,7 @@ public interface IProductLookUp
         IInstalledProduct GetProductFromProductCode(string productCode);
 
         /// <summary>
-        /// Returns product name list based on lookup criteria
+        /// Returns product name list using on lookup criteria based on product names.
```
aparajit-pratap, line 178: We can omit the `on` after `using`.

src/Tools/DynamoInstallDetective/ProductLookUp.cs
```diff
@@ -9,6 +9,118 @@
...
+        public static void DisableCache()
+        {
+            lock (mutex)
+            {
+                cacheEnabled = false;
+                cachedRecords.Clear();
+                cachedRecords = null;
+            }
+        }
+    }
```
aparajit-pratap, line 122: I think all methods of `RegUtils` can also be made internal as the class itself is internal and does not derive from any other classes or interfaces.

src/Tools/DynamoInstallDetective/Utilities.cs
```diff
@@ -88,19 +88,22 @@ public static IEnumerable FindProductInstallations(string productSearchPattern,
         /// info.</returns>
         public static IEnumerable FindMultipleProductInstallations(List<string> productSearchPatterns, string fileSearchPattern)
         {
-            var installs = new InstalledProducts();
-            // Look up products with ASM installed on user's computer
-            foreach (var productSearchPattern in productSearchPatterns)
+            using (RegUtils.StartCache())
```
pinzart90, line 91: I wrapped existing code in a using (RegUtils.StartCache()) that will allow us to cache the Registry records within the scope

src/Tools/DynamoInstallDetective/ProductLookUp.cs
```diff
@@ -9,6 +9,118 @@
...
+
+            var key = OpenKey(REG_KEY64);
+            foreach(string s in key.GetSubKeyNames())
+            {
+                try
+                {
+                    var prodKey = key.OpenSubKey(s);
+                    var displayName = GetDisplayName(prodKey);
+                    var installLocation = GetInstallLocation(prodKey);
+                    if (displayName.Length > 0 && installLocation.Length > 0)
```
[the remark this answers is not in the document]
pinzart90: good suggestion 👍

src/Tools/DynamoInstallDetective/ProductLookUp.cs
```diff
@@ -9,6 +9,118 @@
...
+                    {
+                        productInfo.Add(s, Tuple.Create(displayName, installLocation));
+                    }
+                }
+                catch (Exception)
```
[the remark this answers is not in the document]
pinzart90: It is supposed to catch the exceptions that the registry api calls inside the GetDisplayName and GetInstallLocation might throw. Ex(`System.Security.SecurityException`)

src/Tools/DynamoInstallDetective/ProductLookUp.cs
```diff
@@ -9,6 +9,118 @@
...
+                { }
+            }
+
+            lock (mutex)
+            {
+                if (cacheEnabled)
+                {
+                    cachedRecords = new Dictionary<string, Tuple<string, string>>(productInfo);
```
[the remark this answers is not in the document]
pinzart90: Yes

src/Tools/DynamoInstallDetective/ProductLookUp.cs
pinzart90: 👍

src/Tools/DynamoInstallDetective/ProductLookUp.cs
pinzart90: If the Class is internal then it means, regardless of the method accessor, methods will not be accessible outside the current assembly. 
Having public specifier for certain methods in an internal class only helps to clarify the intention that those methods would be safe to use if the class was ever to be made public.

src/Tools/DynamoInstallDetective/ProductLookUp.cs
mjkkirschner, line 37: maybe not worth investigating for the simple case, but have you looked at concurrentDictionary? 
https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2?view=net-5.0

src/Tools/DynamoInstallDetective/ProductLookUp.cs
```diff
@@ -9,6 +9,118 @@
...
+                    var prodKey = key.OpenSubKey(s);
+                    var displayName = GetDisplayName(prodKey);
+                    var installLocation = GetInstallLocation(prodKey);
+                    if (displayName.Length > 0 && installLocation.Length > 0)
+                    {
+                        productInfo.Add(s, Tuple.Create(displayName, installLocation));
+                    }
+                }
+                catch (Exception)
+                { }
```
mjkkirschner, line 87: we should really log something I think- or are we worried there might be tons of errors?

src/Tools/DynamoInstallDetective/ProductLookUp.cs
mjkkirschner: Even those these are public - the class is internal, so these are really internal methods AFAIK.

src/Tools/DynamoInstallDetective/ProductLookUp.cs
mjkkirschner: hmm, If I'm first coming to this code I wouldn't understand what item1 was - but I think we can now create named tuple fields?
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/value-tuples#tuple-field-names

Maybe that or adding a comment?

src/Tools/DynamoInstallDetective/ProductLookUp.cs
pinzart90: I guess that would work.
The only concern is that it would make the code harder to understand ...
The only purpose of the RegistryCacher class was to make it easier to enable/disable caching (by using a Disposable object)

src/Tools/DynamoInstallDetective/ProductLookUp.cs
pinzart90: Totally forgot about those...thanks @mjkkirschner 
Will take a look and try to get rid of mutex

src/Tools/DynamoInstallDetective/ProductLookUp.cs
pinzart90: well, and error for each or some of the keys under `uninstall`. 
Might be too many

src/Tools/DynamoInstallDetective/ProductLookUp.cs
pinzart90: Hmm I did add comments here 
https://github.com/DynamoDS/Dynamo/blob/c1f14a59c18d622c1327c934b33f46d51fa6d081/src/Tools/DynamoInstallDetective/ProductLookUp.cs#L32-L33

Did you mean that I should add comments in the Select() above >?

Could not figure out how to define  a dictionary of named Tuples ....

src/Tools/DynamoInstallDetective/ProductLookUp.cs
aparajit-pratap: The fact that the `RegistryCacher` deals with caching, I think it makes sense to do anything related to caching part of the same class, meaning including `EnableCache` and `DisableCache`.

src/Tools/DynamoInstallDetective/ProductLookUp.cs
aparajit-pratap: I think it could be confusing to make the class internal and the methods public. It's more explicit to make the methods internal as well, but that's just me.

src/Tools/DynamoInstallDetective/ProductLookUp.cs
mjkkirschner: I think like this:
`var dict = new Dictionary<string, (string Foo, bool Bar)>();`

src/Tools/DynamoInstallDetective/ProductLookUp.cs
aparajit-pratap: In any case, I don't think it makes sense to have `Enable/DisableCache` public.

src/Tools/DynamoInstallDetective/ProductLookUp.cs
mjkkirschner: please add a comment then, otherwise I think sonarQube will complain.

src/Tools/DynamoInstallDetective/ProductLookUp.cs
pinzart90: Hmm..since I am allocating the dictionary and assigning null to it (and cacheEnabled boolean) , I do not think I can get rid of the lock altogether.
So it might not be very beneficial to introduce concurrentDictionary since that is meant to handle Add/Remove items in a concurrent scenario. The critical sections are only when we allocate the Dictionary, return a copy of it and  when we clear it up/make it null.

mjkkirschner: thanks for considering and ruling it out!

src/Tools/DynamoInstallDetective/ProductLookUp.cs
pinzart90: actually that is a good point. we don't really need the EnableCache, DisableCache methods at the moment.
With that in mind it would make sense to move the inner logic to the RegistryCacher

src/Tools/DynamoInstallDetective/ProductLookUp.cs
pinzart90: nice. Thanks

src/Tools/DynamoInstallDetective/ProductLookUp.cs
```diff
@@ -202,8 +281,9 @@ public IInstalledProduct GetProductFromProductCode(string productCode)
 
         public virtual IEnumerable<string> GetProductNameList()
         {
-            var key = OpenKey(REG_KEY64);
-            return key.GetSubKeyNames().Where(s => s.Contains(ProductLookUpName));
+            return RegUtils.GetInstalledProducts().Select(s => s.Value.DisplayName).Where(s => {
+                return s?.Contains(ProductLookUpName) ?? false;
+            });
```
aparajit-pratap, line 286: Minor point but I was wondering if we could just have a property on `InstalledProductLookup` or `IProductLookup` that stored a list of installed products called `InstalledProduct`:
```
private List<string> installedProducts;
List<string> InstalledProducts 
{
    get 
    {
         if(installedProducts != null) return installedProducts;
         installedProducts = RegUtils.GetInstalledProducts();
         return installedProducts;
    }
}

File: src/Tools/DynamoInstallDetective/ProductLookUp.cs (hunk 8 above)
Comment (pinzart90) [reply src/Tools/DynamoInstallDetective/ProductLookUp.cs:114]: To me it sounds a bit overkill to make all members internal if a class is specified as internal.
I guess if we consider this as a coding convention I could get behind it. 
What do you think @mjkkirschner @aparajit-pratap @sm6srw ?

Comment (aparajit-pratap) [reply src/Tools/DynamoInstallDetective/ProductLookUp.cs:114]: I'm fine with it as-is.

File: src/Tools/DynamoInstallDetective/ProductLookUp.cs (hunk 14 above)
Comment (aparajit-pratap) [reply src/Tools/DynamoInstallDetective/ProductLookUp.cs:286]: Then say this on line 284:
```
return InstalledProducts.Select(s => s.Value.DisplayName).Where(s => {
                return s?.Contains(ProductLookUpName) ?? false;
```

pinzart90: hmm...this might work but I have some concerns:
I would change the code a bit to have the `installedProducts` as private static (static so that the installedProducts can be reused across instances of InstalledProductLookup )
I do not think we can add properties to Interfaces like IProductLookup .

The benefit of creating the cache in a specific scope is that we can erase the cache at the end. I don't know if we can control when a static property in the InstalledProductLookup would get cleaned up. The state of the installedProducts  data would stay the same for the rest of the process lifetime (maybe that is not an issue...). And I don;t think we can control when the static property can get cleaned up..

aparajit-pratap: I don't think it needs to be static as `InstalledProductLookup` is initialized only once, I think, at least for ASM lookup. This simplifies the implementation.

pinzart90: In the example bellow :
`
foreach (var productSearchPattern in productSearchPatterns) { installs.LookUpAndInitProducts(new InstalledProductLookUp(productSearchPattern, fileSearchPattern)); }
`
We have multiple `InstalledProductLookup` instances. How do you mean `InstalledProductLookup` is initialized only once ?

aparajit-pratap: Oh, in that case, scratch that.

aparajit-pratap (approved):

src/Tools/DynamoInstallDetective/ProductLookUp.cs
pinzart90: Using it as static will make it work...but the only (biggest) downside I see is that the data in the static will live on for the entire lifetime of the process. If there are a lot of records in the windows registry, it will be useless data that Dynamo will carry around...

Issue #8885, opened 2018-05-23
Dynamo 2.0 Dictionary watch window beneath ZT nodes doesn't display


ThomasMahon:
## Dynamo version

![image](https://user-images.githubusercontent.com/18601851/40433901-154c8e22-5ea5-11e8-874c-fdd7d0026c4c.png)

## Operating system
Win 10

## What did you do? 

Dictionary watch window doesn't display beneath a node.

![recording 12](https://user-images.githubusercontent.com/18601851/40433936-299e13fa-5ea5-11e8-8586-ad9cce436d02.gif)

Seems to only be affecting either ZT nodes, or generic dictionaries:
![image](https://user-images.githubusercontent.com/18601851/40434162-ac29ad34-5ea5-11e8-8ec6-e1a6ecd8bd09.png)

mjkkirschner:
@aparajit-pratap was this fixed?
https://github.com/DynamoDS/Dynamo/pull/8734

mjkkirschner:
@ThomasMahon is it possible to test on the latest daily build?

mjkkirschner:
also @ThomasMahon just curious if you see this with your own nodes only or other ZT nodes as well? I feel like Déjà vu.

ThomasMahon:
@mjkkirschner no nested dictionaries, 2D list of curves, and 1D list of strings. Maybe its caused by using generic types in the dictionary declaration due to the mixed return types, given that the OOTB node shown above is working?

![image](https://user-images.githubusercontent.com/18601851/40435468-9abd6d9e-5ea8-11e8-94bc-ac5d936644e1.png)

mjkkirschner:
@ThomasMahon tough to say without the source for the nodes - but can you try removing the generics so type your dictionary <string,object> and then also or in another experiment remove the `[]` from the key names.

aparajit-pratap:
@ThomasMahon I don't see there's a problem with `[]` in the port names either as long as they are written in the `MultiReturnAttribute` as well.

aparajit-pratap:
Hmm, I tried it with a generic type too: 
![image](https://user-images.githubusercontent.com/5710686/40444398-4dce7d9a-5e97-11e8-8a99-b4a896cb937e.png)
![image](https://user-images.githubusercontent.com/5710686/40444410-58e7ea68-5e97-11e8-871d-927e5f3527c4.png)

ThomasMahon:
[...]

PR #11266, opened 2020-11-19
Fix for regression in multi-output node preview 

aparajit-pratap:
### Purpose

Fixes:
https://jira.autodesk.com/browse/DYN-3255
https://jira.autodesk.com/browse/DYN-1016 (GH issue: https://github.com/DynamoDS/Dynamo/issues/8885)

Node previews such as that for `List.SortByKey` had regressed after Martin Stacey's port renaming PR's.

Output port names are aligned with the names of the XML return tags specified on the node/method and the objective of Martin's work was to rename the port names by modifying the XML tag names. That was fine. However, for multi-return nodes, the multi-output node preview logic for ZT nodes (which is based on keys of a dictionary returned from the node) also followed the output port names, and these were therefore affected by the XML tag renaming changes, which is why this caused the node preview to break.

The fix is to make the returned dictionary keys to be based on the multi-return attribute inputs and not the output port names (aka XML tags).

### Declarations

### FYIs
@QilongTang

### DIFF
--- a/src/Libraries/CoreNodes/List.cs
+++ b/src/Libraries/CoreNodes/List.cs
@@ -444,7 +444,7 @@ namespace DSCore
         /// <returns name="sortedList">type: var[]..[]</returns>
         /// <returns name="sortedKeys">type: var[]..[]</returns>
         /// <search>sort;key</search>
-        [MultiReturn(new[] { "sorted list", "sorted keys" })]
+        [MultiReturn("sortedList", "sortedKeys")]
         [IsVisibleInDynamoLibrary(true)]
         public static IDictionary SortByKey(IList list, IList keys)
         {
@@ -484,8 +484,8 @@ namespace DSCore
 
             return new Dictionary<object, object>
             {
-                { "sorted list", sortedList },
-                { "sorted keys", sortedKeys }
+                { "sortedList", sortedList },
+                { "sortedKeys", sortedKeys }
             };
         }
 
--- a/test/DynamoCoreWpfTests/PreviewBubbleTests.cs
+++ b/test/DynamoCoreWpfTests/PreviewBubbleTests.cs
@@ -32,6 +32,7 @@ namespace DynamoCoreWpfTests
         {
             libraries.Add("DesignScriptBuiltin.dll");
             libraries.Add("DSCoreNodes.dll");
+            libraries.Add("FFITarget.dll");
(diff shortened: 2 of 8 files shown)


src/Libraries/CoreNodes/List.cs
```diff
@@ -484,8 +484,8 @@ public static IDictionary SortByKey(IList list, IList keys)
 
             return new Dictionary<object, object>
             {
-                { "sorted list", sortedList },
-                { "sorted keys", sortedKeys }
+                { "sortedList", sortedList },
+                { "sortedKeys", sortedKeys }
```
aparajit-pratap, line 488: These name changes are unrelated to the fix and are made only for consistency so that the same is reflected in the preview bubble.

mjkkirschner:
awesome @aparajit-pratap - is there a way to add a regression test for this?

aparajit-pratap:
@mjkkirschner I've added the test, let's wait for PR checks. I also need to update docs for node authors.

aparajit-pratap:
@mjkkirschner any other comments?

test/DynamoCoreWpfTests/PreviewBubbleTests.cs
```diff
@@ -354,6 +355,30 @@ public void PreviewBubble_ShowExpandedPreviewOnPinIconHover()
...
+            Assert.IsTrue(previewBubble.IsCondensed, "Compact preview bubble should be shown");
+            Assert.AreEqual(Visibility.Collapsed, previewBubble.bubbleTools.Visibility, "Pin icon should not be shown");
+
+            // hover preview bubble to see pin icon
+            RaiseMouseEnterOnNode(previewBubble);
+            Assert.AreEqual(Visibility.Visible, previewBubble.bubbleTools.Visibility, "Pin icon should be shown");
+
+            // expand preview bubble
+            RaiseMouseEnterOnNode(previewBubble.bubbleTools);
+            Assert.IsTrue(previewBubble.IsExpanded, "Expanded preview bubble should be shown");
```
mjkkirschner, line 379: so this line failed previously?

mjkkirschner (approved):
LGTM

test/DynamoCoreWpfTests/PreviewBubbleTests.cs
aparajit-pratap: Yes

PR #11268, opened 2020-11-20
Display input port type specific default suggestions only

zeusongit:
### Purpose

[DYN-3251](https://jira.autodesk.com/browse/DYN-3251)
This PR only displays the node autocomplete suggestion for skipped primitive types that corresponds to the clicked input port type.
Test updated

SS:

### Declarations

### Reviewers

@DynamoDS/dynamo


src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -140,8 +146,29 @@ internal IEnumerable<NodeSearchElement> GetMatchingSearchElements()
...
+                        DefaultResults = AllDefaultResults.Where(e => e.Name == "Boolean").ToList();
+                        break;
+                    default:
+                        DefaultResults = AllDefaultResults;
+                        break;
+                }
                 return elements;
             }
+            else {
+                DefaultResults = AllDefaultResults;
```
QilongTang, line 170: @zeusongit Why are we modifying `DefaultResults` (class property) in this Utility or API function? I think previously we are only returning elements. I would like to be consistent with that. Without that, I feel we do not need to create a new property `AllDefaultResults`, what do you think?

zeusongit: That was my initial intention, but for doing that we will have to query the default nodes from the list of all the nodes, which could be a time consuming process for a trivial task. So searching a type out of a list of five nodes made more sense to me.
Or 
an alternative approach that I thought of was to define a dictionary of those 5 default types one time and assign those from the dictionary into the elements collection based on the input port.

mjkkirschner: yeah, this feels pretty weird 😉 - I think a dictionary with the proper list is more straightforward - why keep state if you don't have to.

QilongTang: @zeusongit What about touch `InitializeDefaultAutoCompleteCandidates()` for those input types, make the default results correct in the first place so we do not need to touch this API?

zeusongit: @QilongTang InitializeDefaultAutoCompleteCandidates() is only called once when the constructor is called, not every time the user clicks on an input port.

QilongTang: @zeusongit I see, then produce a dictionary sounds proper. Would you update the code following what you described earlier?

QilongTang:
Hi @zeusongit Here is the change I suggested earlier so we do not touch the behavior of `DefaultResults` or  the API that get matching search elements, I tested myself a bit. What do you think?

```
 internal void PopulateAutoCompleteCandidates()
        {
            if (PortViewModel == null) return;

            searchElementsCache = GetMatchingSearchElements().ToList();

            // If node match searchElements found, use default suggestions
            if (!searchElementsCache.Any())
            {
                searchElementsCache =  DefaultResults.Select(e => e.Model).ToList();
                switch (PortViewModel.PortModel.GetInputPortType())
                {
                    case "int":
                        FilteredResults = DefaultResults.Where(e => e.Name == "Number Slider" || e.Name == "Integer Slider").ToList();
                        break;
                    case "double":
                        FilteredResults = DefaultResults.Where(e => e.Name == "Number Slider" || e.Name == "Integer Slider").ToList();
                        break;
                    case "string":
                        FilteredResults = DefaultResults.Where(e => e.Name == "String").ToList();
                        break;
                    case "bool":
                        FilteredResults = DefaultResults.Where(e => e.Name == "Boolean").ToList();
                        break;
                    default:
                        FilteredResults = DefaultResults;
                        break;
                }
            }
            else
            {
                FilteredResults = GetViewModelForNodeSearchElements(searchElementsCache);
            }
        }
```

zeusongit:
What I proposed on slack was searching the zt nodes once, but this seems better. Updating the code after test, Thanks.

QilongTang:
@zeusongit Once all regressions pass, looks good. The regression fix you did earlier may be still valid.

QilongTang (approved):
LGTM, once tests pass, feel free to merge and cherry-pick

PR #11274, opened 2020-11-23
Cherry-pick: Fix for regression in multi-output node preview  (#11266)

aparajit-pratap:


test/Engine/FFITarget/FFITarget.csproj
```diff
@@ -25,6 +25,7 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <Prefer32Bit>false</Prefer32Bit>
+    <DocumentationFile>..\..\..\bin\AnyCPU\Debug\en-US\FFITarget.xml</DocumentationFile>
```
mjkkirschner, line 28: hmm - I guess I missed this before, or didn't see it, how can we make sure this does not end up in the delivered zip? We must somehow be removing FFITarget.dll right?

mjkkirschner: oh - it looks like we do deliver it, okay, so I guess no harm...

mjkkirschner (approved):

test/Engine/FFITarget/FFITarget.csproj
mjkkirschner: Screen Shot 2020-11-23 at 3 11 59 PM

aparajit-pratap: I suppose we should make sure this doesn't get shipped in the future.

mjkkirschner: yes, actually, now would be the time... but it's a bit tricky because it needs to be present for the tests, I guess we would remove it from the harvest scripts or potentially from the zipping code itself in dynamoBuildScripts repo.

aparajit-pratap: Tried looking for code in DynamoBuildScripts repo where it may be filtering out test dll's before zipping but couldn't figure out where that logic is. Will try asking Sherlyn.

PR #11278, opened 2020-11-24
Save/Load view extension size and location

mmisol:
### Purpose

Information of where an extension control was last shown is saved and
retrieved as preference settings. These are actually not directly
visible for the user and are updating automatically, just like the main
Dynamo window size. These settings are used when showing the extension
control, to restore it to how it used to be before closing it.

The available information for each view extension is:

1. How it is shown. For now this can be either as a floating window or
docked to the right-side panel.

2. In case it is shown as a floating window, the following information
about the window is saved:
- Location by Top/Left coordinates
- Size by Width/Height
- Whether the window is maximized

### Declarations

### DIFF
--- a/src/DynamoCore/Configuration/PreferenceSettings.cs
+++ b/src/DynamoCore/Configuration/PreferenceSettings.cs
@@ -370,6 +370,11 @@ namespace Dynamo.Configuration
         [XmlIgnore]
         public bool NamespacesToExcludeFromLibrarySpecified { get; set; }
 
+        /// <summary>
+        /// Settings that apply to view extensions.
+        /// </summary>
+        public List<ViewExtensionSettings> ViewExtensionSettings { get; set; }
+
         #endregion
 
         /// <summary>
@@ -415,6 +420,7 @@ namespace Dynamo.Configuration
             ShowTabsAndSpacesInScriptEditor = false;
             EnableNodeAutoComplete = false;
             DefaultPythonEngine = string.Empty;
+            ViewExtensionSettings = new List<ViewExtensionSettings>();
         }
 
         /// <summary>
--- a/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
+++ b/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
@@ -287,17 +287,13 @@ namespace Dynamo.Controls
                     Width = windowSettings.Width,
                     Height = windowSettings.Height
                 };
-                var wasMaximized = windowSettings.Status == WindowStatus.Maximized;
-                if (wasMaximized)
+                window = new ExtensionWindow(this, ref windowRect);
+                if (windowSettings.Status == WindowStatus.Maximized)
                 {
                     // In case the window was maximized, its width and height will be max. That makes restore pretty
                     // useless so instead we set the size to 0 which means to restore to the default size.
-                    windowRect.Width = 0;
-                    windowRect.Height = 0;
-                }
-                window = new ExtensionWindow(this, ref windowRect);
-                if (wasMaximized)
-                {
+                    //windowRect.Width = 0;
+                    //windowRect.Height = 0;
                     window.WindowState = WindowState.Maximized;
                 }
             }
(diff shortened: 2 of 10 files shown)


src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -212,71 +212,165 @@ public DynamoView(DynamoViewModel dynamoViewModel)
...
+                    Top = windowSettings.Top,
+                    Width = windowSettings.Width,
+                    Height = windowSettings.Height
+                };
+                window = new ExtensionWindow(this, ref windowRect);
+                if (windowSettings.Status == WindowStatus.Maximized)
+                {
+                    // Rather than setting the WindowState here, this is delayed to the Loaded event.
+                    // This helps overcome a bug which makes the window appear always on the primary screen.
+                    window.ShouldMaximize = true;
```
zeusongit, line 295: Nice 😅

zeusongit (approved):
LGTM

src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -370,6 +370,11 @@ public string DefaultPythonEngine
         [XmlIgnore]
         public bool NamespacesToExcludeFromLibrarySpecified { get; set; }
 
+        /// <summary>
+        /// Settings that apply to view extensions.
+        /// </summary>
+        public List<ViewExtensionSettings> ViewExtensionSettings { get; set; }
```
QilongTang, line 376: The setting here is good design. I also wonder when we about do make internal window like Python editor dockable, do we apply same setting for them or use a list of different but similar setting? Or we make this setting more generic? Let me know what you think. I think it might be something we want to consider

mmisol: I don't think this can be reused just like that. `ViewExtensionSettings` by itself contains view extension specific properties like `UniqueId` and `Name` that don't apply to script editors. Still, a lot of things in this contribution can be reused.

QilongTang: @mmisol Yes, and thinking more on this. We may not need to support remembering the positions of python editor at all. Will take a look at the rest of code

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -212,71 +212,165 @@ public DynamoView(DynamoViewModel dynamoViewModel)
...
+            
+            // Icon is passed from DynamoView (respecting Host integrator icon)
+            SetApplicationIcon();
+            window.Icon = this.Icon;
+            window.ExtensionContent.Content = contentControl;
+            window.Title = viewExtension.Name;
+            window.Tag = viewExtension;
+            window.Uid = viewExtension.UniqueId;
+            window.Closing += ExtensionWindow_Closing;
+            window.Closed += ExtensionWindow_Closed;
```
QilongTang, line 307: These reminds me, do we need to unsubscribe these when docking?

src/DynamoCoreWpf/Windows/ExtensionWindow.xaml
```diff
@@ -7,6 +7,8 @@
                            xmlns:ui="clr-namespace:Dynamo.UI"
                            mc:Ignorable="d" 
                            d:DesignHeight="450" d:DesignWidth="800"
+                           MinHeight="32"
+                           MinWidth="160"
```
QilongTang, line 11: Would you attach an example of the extension window with minimal size to this PR so we have a record?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
mmisol: No because these windows no longer live after that.

src/DynamoCoreWpf/Windows/ExtensionWindow.xaml
mmisol: That's only big enough so to make the title bar not disappear.

QilongTang (approved):
@mmisol Thanks, LGTM

PR #11291, opened 2020-11-30
DYN-1819 ExportExcel Write as String Option

StudioLE:
### Purpose

JIRA: [DYN-1819](https://jira.autodesk.com/browse/DYN-1819)

Issues: https://github.com/DynamoDS/DynamoWishlist/issues/13

Added an additional `writeAsString` parameter to the `Data.ExportExcel` node. This parameter defaults to `false` ensuring the existing behaviour is preserved. When set to `true` the NumberFormat for the written range is set to `text`, therefore preserving the string rather than letting Excel automatically determine the format.

Behaviour before:
DYN-1819-Export-Excel-Before

Behaviour after `writeAsString: false`:
DYN-1819-Export-Excel-After-False

Behaviour after `writeAsString: true`:
DYN-1819-Export-Excel-After-True

### Declarations

### Reviewers

### FYIs

@Amoursol


mmisol:
Hi @StudioLE . Could you please add a unit test for this?

mjkkirschner:
I don't think these changes are allowed...
see
`Adding a parameter with a default value.`

see:

❌ DISALLOWED: Adding, removing, or changing the order of parameters

while this might work pretty well for nodes - I don't think it will work for c# code that calls this code.

mmisol:
@mjkkirschner Right, but this PR is doing what's described in the ticket. Could it be that we didn't consider this part of our API or disregarded the compatibility break? Copying @Amoursol

Amoursol:
> @mjkkirschner Right, but this PR is doing what's described in the ticket. Could it be that we didn't consider this part of our API or disregarded the compatibility break? Copying @Amoursol

I had indeed not been aware of the compatibility break - assumed that adding a new node port (As we have done in other cases?) would be acceptable and a non-breaking change if default values adhered to existing graph conditions.

@mjkkirschner - What kind of issues will arise from this in a non-node basis, assuming my above assertions are correct?

mjkkirschner:
code which calls this node from c# will throw a `MissingMethodException` - so if a user has written another zero touch node or extension which references this will likely result in a failing node or crash from the extension.

What should likely be done is to:
1. obsolete these nodes.
2. Make sure they are hidden in the library (obsolete should do this AFAIK)
3. Create a new node with the new input.
4. Consider looking into using the `[AlsoKnownAs]` attribute - I can't remember the details of this attribute but it's made for migration/loading older graphs with new nodes.

mjkkirschner:
it may make sense to use the migrations.xml file instead since theres already once for dsoffice assembly: 
https://github.com/DynamoDS/Dynamo/blob/master/src/Libraries/DSOffice/DSOffice.Migrations.xml

Either of these migration options will need manual and automated testing.

mmisol:
@StudioLE Based on the comments made by @mjkkirschner , could we ask you to change the contribution to use a new node/function so that we can remain backwards compatibility? Thank you

SHKnudsen:
Hi guys, im looking into getting this wrapped up, Sorry for the delay! 

Im having some issues getting the migrations to work. I tried the following:

- Create new node (ExportExcelWithStringParameter) <- name was just for testing btw..
- Make a new migration:
```xml
  <priorNameHint>
    <oldName>DSOffice.Data.ExportExcel</oldName>
    <newName>DSOffice.Data.ExportExcelWithStringParameter</newName>
  </priorNameHint>
```
- obsolete the old node so it dosent show in the library

Is there something im doing wrong here, it dosent really work at least. When i open a graph using the old `ExportExcel` node it dosent use the new `ExportExcelWithStringParameter` it just shows the old in a warning state as it is obsolete. For some reason copy and pasting the obsolete node will make a `ExportExcelWithStringParameter` node but renamed as `ExportExcel`.

mmisol:
Hi @SHKnudsen . I'm not familiar with the migrations but maybe @mjkkirschner and @aparajit-pratap can chime in?

aparajit-pratap:
@SHKnudsen could you also try including the list of parameters with both node names in the migrations.xml file since the new node also has an additional parameter. Refer to this for an example: https://github.com/DynamoDS/Dynamo/blob/b39ac2a78512a4214522c9da7f7ca331d22471aa/extern/ProtoGeometry/ProtoGeometry.Migrations.xml#L76

mjkkirschner:
@SHKnudsen @StudioLE - for now can you try the following:
1. Obsolete the existing node.
2. Add a new node with the extra parameter - make sure this node has a new name - do not use the same name.
3. Nevermind the migration for now.

SHKnudsen:
@mjkkirschner done.

Called the node `ExportToExcel` for now, hard to find a different name than the original. Let me know if there are other suggestions for the name.

src/Libraries/DSOffice/Excel.cs
```diff
@@ -838,7 +862,7 @@ public static object[][] ImportExcel(FileInfo file, string sheetName, bool readA
         /// <param name="writeAsString">Toggle to switch between writing Excel file as strings.</param>
         /// <returns name="data">Data written to the spreadsheet.</returns>
         /// <search>office,excel,spreadsheet</search>
-        public static object[][] ExportExcel(string filePath, string sheetName, int startRow, int startCol, object[][] data, bool overWrite = false, bool writeAsString = false)
+        public static object[][] ExportToExcel(string filePath, string sheetName, int startRow, int startCol, object[][] data, bool overWrite = false, bool writeAsString = false)
```
aparajit-pratap, line 865: @SHKnudsen how about naming it `ExportExcelAsString`?

src/Libraries/DSOffice/Excel.cs
```diff
@@ -820,6 +820,30 @@ public static object[][] ImportExcel(FileInfo file, string sheetName, bool readA
...
+        /// <param name="sheetName">Name of the workseet to write data to.</param>
+        /// <param name="startRow">Start row for writing data. Enter 0 for Row 1, 1 for Row 2, etc.</param>
+        /// <param name="startCol">
+        ///     Start column for writing data. Enter 0 for Column A, 1 for Column B, etc.
+        /// </param>
+        /// <param name="data">Data to write to the spreadsheet.</param>
+        /// <param name="overWrite"></param>
+        /// <returns name="data">Data written to the spreadsheet.</returns>
+        /// <search>office,excel,spreadsheet</search>
+        [Obsolete("Use ExportToExcel instead.")]
```
aparajit-pratap, line 840: Does adding the `Obsolete` attribute also hide it from the library or do we also need to add the `[IsVisibleInDynamoLibrary(false)]` attribute?

src/Libraries/DSOffice/Excel.cs
SHKnudsen: didnt want to do that as its just an option to export as a string, `ExportExcelAsString` sound like it will export as string by default to me?

src/Libraries/DSOffice/Excel.cs
SHKnudsen: Yes obsolete hides it from the Library.

src/Libraries/DSOffice/Excel.cs
aparajit-pratap: Agree, but it still says something about the node - that it at least makes that option (to write as string) available. It looks less confusing than `ExportExcel` and `ExportToExcel` IMO.

SHKnudsen: @aparajit-pratap updated the name of the node.

aparajit-pratap (commented):
@SHKnudsen thanks! You'll need to do one last thing - add an entry for the new node to the `libraryItems.json` file so that the remaining failing test passes. See: https://github.com/DynamoDS/Dynamo/wiki/Adding-or-Updating-Built-In-Library-Icons#step-5-update-testing-resources

src/Libraries/DSOffice/Excel.cs
aparajit-pratap: @SHKnudsen I'm so sorry, I just realized - the old node `ExportExcel` will be obsoleted and not appear in the library, in which case, I totally agree with you. It makes sense to have the new name, `ExportToExcel` since there will be just this node going forwards. Can we change that back? Really sorry.

SHKnudsen: No worries, I reverted the change so its back to `ExportToExcel`.

src/Libraries/DSOffice/DSOfficeImages.resx
```diff
@@ -120,159 +120,158 @@
...
   <data name="Excel.Small" type="System.Drawing.Bitmap, System.Drawing" mimetype="application/x-microsoft.net.object.bytearray.base64">
     <value>
-        iVBORw0KGgoAAAANSUhEUgAAACAAAAAgCAYAAABzenr0AAAAAXNSR0IArs4c6QAAAARnQU1BAACxjwv8
-        YQUAAAAZdEVYdFNvZnR3YXJlAEFkb2JlIEltYWdlUmVhZHlxyWU8AAAATElEQVRYR+3QsQkAIAxE0Uwr
-        juBsLqeWKc5CrjHwH6T5hXAGAOBmzL7yue2ZesRpeHa+ruVzWz1qhdPqUSucVo9a4bR61AqnAcCnIjZ8
-        fvMh4IFW4wAAAABJRU5ErkJggg==
+        iVBORw0KGgoAAAANSUhEUgAAACAAAAAgCAYAAABzenr0AAAABGdBTUEAALGPC/xhBQAAABl0RVh0U29m
+        dHdhcmUAQWRvYmUgSW1hZ2VSZWFkeXHJZTwAAABMSURBVFhH7dCxCQAgDETRTCuO4Gwup5YpzkKuMfAf
+        pPmFcAYA4GbMvvK57Zl6xGl4dr6u5XNbPWqF0+pRK5xWj1rhtHrUCqcBwKciNnx+8yHggVbjAAAAAElF
+        TkSuQmCC
```
mmisol, line 126: Did this icon change?

src/Libraries/DSOffice/DSOfficeImages.resx
```diff
@@ -120,159 +120,158 @@
...
+        vC7k127Ywu2XGwvC93pyWHBdqgUtLDeWw8FRHAl/cWKhdan8OXAQbSwvHsIoi4Sv11loXSp/J1icCWBL
+        uChYd2Ah+B5/4t3229dsGemyYRbmNMvQRu37ADs7iEKb/Yc1HwVrAfJIsABdNpSH2Ip5o0JNC+sxCi32
+        H9Z8FKwF/71/066CNp7evjpBiK2YNyrznzI1DFiPUWhxsbEw8h3QLjbgxinfmNPOaIOF4SdfDf4bPcm+
+        /7SK+0E6KcuxQPKjx0smFuIs8ms2DHBb0WChxLOALt+I/cgy5whuK3IsmHZL2vCfiy4mwa+R//Rs4HYi
+        YuG07wRg1kmI4TuLubJhS7goWAtyz83HV35CUAk/nHn0z8M79pEFPHYZ6jVeax5G+gnK7ztOqcH9AGs+
+        CtaC6IuTAHwPxw9VdqyQ8tD9FM0Pc/yN2sYi8f9FR+J9mdBi/2HNR8FawLwXf/3hu/yMOOD7Ob6qWfvz
+        xeDuk3//vu9nyGkkgJc832/yXdf8GTP0iWb3zoUWlxPLyEMb9wetSfDa9u+J/tm0YxIMrYh2w0I6YPIw
+        J/mbonv8D7zpnw8R+16TMAsWlk+Gvzl7yLn8Oxp6xHyahHljgWsS5o0FrkmYNxa4JmHeWOCahHljgWsS
+        5o0FrkmYNxZ4nASdnH1uLPRmEhT+vPBJwEchhBBCCCGEEEIIIYQQQgghxBSsrHwC4O+ISk6Usf4AAAAA
+        SUVORK5CYII=
```
mmisol, line 235: The content for this one looks different too. Is that expected?

src/Libraries/DSOffice/Excel.cs
```diff
@@ -321,14 +321,15 @@ public static object[][] Read(string filePath, string sheetName)
         /// </param>
         /// <param name="data">Data to write to the spreadsheet.</param>
         /// <param name="overWrite"></param>
+        /// <param name="writeAsString">Toggle to switch between writing Excel file as strings.</param>
         /// <returns name="data">Data written to the spreadsheet.</returns>
         /// <search>office,excel,spreadsheet</search>
         [IsVisibleInDynamoLibrary(false)]
-        public static object[][] WriteToFile(string filePath, string sheetName, int startRow, int startCol, object[][] data, bool overWrite = false)
+        public static object[][] WriteToFile(string filePath, string sheetName, int startRow, int startCol, object[][] data, bool overWrite = false, bool writeAsString = false)
```
mmisol, line 328: This function is having a binary breaking change because optional parameters are a compile-time feature. Can we use an overload instead?

src/Libraries/DSOffice/DSOfficeImages.resx
```diff
@@ -120,159 +120,158 @@
...
+        zZfrj3/3Oq+9ar7/fHnpfp8WfZn1cw+fo2rhujmqFrn3kMtv0ZdZP/fwOaoWrpujanG/s8/yw+3c8lv0
+        ZTAnFso/uMcSyzzva90Ny39osdDTSPtfyzksfymx2LlLYPlLiwVvuwSWfyix6HwJT2P57hUuR9XCdXNU
+        bQs+i9x954LrXty8ufj24/br9J6vauFmc1RdP/fwOaoWrpujaruAV5GmX4LrtrRLmN7z+6CRZ1xUXT/3
+        8DmqFq6bo2q7gPsfL1eum6PRwnVzVEUTC28fPxnfSHUosezp4yfjEg4hFj16z3+nGpYQC3YfP+0V8ypy
+        rhqWEkuePn5Y+jHEwtsFsPRt3CtcjqqF6+aoWrhujqqF6+aoun7u4XNULVw3R9XCdXNULVw3R9X1cw+f
+        o2rhujmqFq6bo2rhujmqAgAAAAAAAAAAAAAAAAAAAACSk5M/qVE78xNHtssAAAAASUVORK5CYII=
+</value>
   </data>
-  <data name="DSOffice.Data.ExportExcel.Small" type="System.Drawing.Bitmap, System.Drawing" mimetype="application/x-microsoft.net.object.bytearray.base64">
+  <data name="DSOffice.Data.ExportToExcel.Small" type="System.Drawing.Bitmap, System.Drawing" mimetype="application/x-microsoft.net.object.bytearray.base64">
```
mmisol, line 267: Since we are not doing the migration yet, I think we should keep the icon for the old node too, rather than replace it.

src/Libraries/DSOffice/DSOfficeImages.resx
SHKnudsen: I had to rename the icon. Guess thats whats triggering the change?

src/Libraries/DSOffice/DSOfficeImages.resx
SHKnudsen: Okay, but we still use the same icon for both right?

src/Libraries/DSOffice/Excel.cs
SHKnudsen: Sorry, hadnt realized this was public too, will fix that.

src/Libraries/DSOffice/DSOfficeImages.resx
mmisol: It could be some randomness in the `resx` file introduced by Visual Studio, but it is strange because this is another icon, not the one that was renamed.

src/Libraries/DSOffice/DSOfficeImages.resx
mmisol: Yes, that should be fine.

src/Libraries/DSOffice/DSOfficeImages.resx
SHKnudsen: Oh right, hmm that is strange. Not even sure what this icon is used for, there is only a `Small` version of it

src/Libraries/DSOffice/DSOfficeImages.resx
SHKnudsen: Yea not sure whats going on here, haven't touched this either

mmisol: I checked and in Dynamo the icons of the node look ok, so this is harmless.

mmisol (approved):
Looks good to me

src/Libraries/DSOffice/Excel.cs
aparajit-pratap: This new method doesn't need to be public. This logic can actually be folded into the new `ExportToExcel` node. The only reason there existed an old `WriteToFile` method was that it was originally a node, which was later renamed to `ExportExcel`.

aparajit-pratap (commented):
One last comment about the redundancy of the new `WriteToFile` method.

src/Libraries/DSOffice/Excel.cs
SHKnudsen: So one thing before i make this change. Would you prefer to have one private method that `WriteToFile` and `ExportToExcel` uses? other option would be to duplicate the logic from `WriteToFile` into `ExportToExcel` with the additional parameter.

mmisol: I think option 1 sounds good. I'd rather not duplicate the code if possible.

mmisol:
@aparajit-pratap Could I ask you take one final look here? If you are ok with the last change I think we are good to merge.

src/Libraries/DSOffice/Excel.cs
```diff
@@ -864,7 +845,7 @@ public static object[][] ImportExcel(FileInfo file, string sheetName, bool readA
```
aparajit-pratap, line 845: Is there a line white space here? Can we delete this?

aparajit-pratap (approved):

aparajit-pratap:
@mmisol are we going to add any tests or have Sylvester add tests?

mmisol:
Good point @aparajit-pratap . I didn't realize we didn't have them. Could you take care of adding one @SHKnudsen ?

SHKnudsen:
Yes sorry guys, actually thought there was tests on this one. Ill have a look at it now.

SHKnudsen:
Added a simple test that checks if the exported value is converted to `string` when the `writeAsString` is set to true.
image

mmisol:
Thanks @SHKnudsen . I'll wait for self serve and merge.

test/core/excel/ExportToExcelAsString.dyn
```diff
@@ -0,0 +1,284 @@
...
+  "Name": "ExportToExcelAsString",
+  "ElementResolver": {
+    "ResolutionMap": {}
+  },
+  "Inputs": [],
+  "Outputs": [],
+  "Nodes": [
+    {
+      "ConcreteType": "CoreNodeModels.Input.Filename, CoreNodeModels",
+      "HintPath": "C:\\Users\\SylvesterKnudsen\\Documents\\GitHub\\StudioLEDynamo\\test\\core\\excel\\WriteFile.xlsx",
```
QilongTang, line 14: I think we missed this during the PR review.. Not sure how self CI passed with this...

mmisol: @QilongTang Now that you mention it, I think @alfredo-pozo said `ExcelTests` are not run in self-serve CI. This is my bad, sorry for forgetting.

mjkkirschner: @QilongTang @mmisol - the weird part is why did we not get a master-15 job until today that contained this code?

PR #11292, opened 2020-11-30
DYN-2089 Copy To Clipboard from Preview and Watch Node

StudioLE:
### Purpose

JIRA: [DYN-2089](https://jira.autodesk.com/browse/DYN-2089)

Added the option to `Copy To Clipboard` via the context menu of the preview bubble and the watch node.

DYN-2089-Copy-Preview-v2

### Declarations

### Reviewers

### FYIs

@Amoursol


mmisol:
Hi @StudioLE . Could you please add a unit test for `GetNodeLabelTree`? Something similar to what you show in the gif should be fine.

src/DynamoCoreWpf/ViewModels/Preview/WatchViewModel.cs
```diff
@@ -274,5 +275,40 @@ public void CountLevels()
...
         }
+
+        /// <summary>
+        /// Get the NodeLabel for this WatchViewModel and any children.
+        /// Each level of children will be indented by 2 spaces.
+        /// </summary>
+        /// <param name="depth">The number of levels of indentation.</param>
+        /// <param name="includeKey">If true the list or dictionary key will be included in the string.</param>
+        /// <returns></returns>
+        public string GetNodeLabelTree(int depth = 0, bool includeKey = false)
```
mjkkirschner, line 286: @mmisol do you think this should be public or is internal fine?

mmisol: It's accessed from another project, so it should be either `public` or `internal` using `InternalsVisibleTo`. I think I have expressed it before but I don't think any `public` member is part of our API. We should document exactly which parts are.

StudioLE:
> Hi @StudioLE . Could you please add a unit test for `GetNodeLabelTree`? Something similar to what you show in the gif should be fine.

I've added a test for `GetNodeLabelTree`:
https://github.com/DynamoDS/Dynamo/pull/11292/files#diff-f5b5b36385405e835255f574028d8843b62365cf3cccb9c8b07dbc7e46aae3b2R334

A test that simulates clicking the `Copy` context menu item and checks the clipboard content:
https://github.com/DynamoDS/Dynamo/pull/11292/files#diff-cdd9f8e180ca08c4fef4e7bf3a80dc4171e731007b403f14b14c6058b2f2f296R373

The two tests use the following `.dyn`:
image

I've also amended the `PreviewControl` behaviour so that it will copy the value of the condensed preview if the expanded preview isn't open:
 https://github.com/DynamoDS/Dynamo/pull/11292/files#diff-320ce46a69cb61e366fce132975a7813669a2a386921870e8bda49714c774a47R726-R732
DYN-2089-Copy-Condensed-Preview

Finally, it looks like the `PreviewControl_HiddenDummyVerticalBoundaries` test is failing on my local machine. But this occurs even before the commits in this PR so I imagine it might be an issue with my local environment as the Jenkins SelfServe passed before.

MicrosoftTeams-image

mmisol:
I'm seeing the `PreviewBubble_CopyToClipboard` failing on the server with:
```
System.Runtime.InteropServices.COMException : OpenClipboard Failed (Exception from HRESULT: 0x800401D0 (CLIPBRD_E_CANT_OPEN))
+++++++++++++++++++
STACK TRACE:
   at System.Runtime.InteropServices.Marshal.ThrowExceptionForHRInternal(Int32 errorCode, IntPtr errorInfo)
   at System.Windows.Clipboard.Flush()
   at System.Windows.Clipboard.CriticalSetDataObject(Object data, Boolean copy)
   at System.Windows.Clipboard.SetDataInternal(String format, Object data)
   at DynamoCoreWpfTests.PreviewBubbleTests.PreviewBubble_CopyToClipboard() in c:\WorkspaceDynamo\test\DynamoCoreWpfTests\PreviewBubbleTests.cs:line 407
```
From what I read, it could be related to accessing the clipboard on multiple threads, which might happen because our tests run in parallel in our CI environment. I'm thinking we could maybe mark this test with the `[Category("Failure")]` attribute, just to prevent it from running in CI but still keep it, as it works locally. Thoughts @mjkkirschner ?

The good thing is we have the unit test working, so I think we should be covered.

mjkkirschner:
@alfredo-pozo are there any attributes to use to mark tests as incompatible with multithreading?

@mmisol  in general that sounds fine for a UI interaction that would not be accessed from multiple processes or threads at the same time... but I guess we wouldn't want to go in this direction for node functionality that raised the same error?

mmisol:
I agree. In this case this is a user interaction, so it should be no problem. Accessing the clipboard from a node's execution sounds like a bad idea to me :)

mmisol:
@StudioLE Could you add the tag to the test as mentioned here: https://github.com/DynamoDS/Dynamo/pull/11292#issuecomment-738231679

With that change I think this PR should be good to go.

SHKnudsen:
@mmisol just wrapping these up. I added the `[Category("Failure")]` to the test.

PR #11294, opened 2020-11-30
DYN-2381 Exclude code block nodes from SetArgumentLacing

StudioLE:
### Purpose

JIRA: [DYN-2381](https://jira.autodesk.com/browse/DYN-2381)

Exclude Code Block Nodes when setting the lacing by selecting multiple nodes and right clicking on the canvas.

The previous behaviour did nothing but causes some visual anomalies and can be confusing to a user.

Previous Behaviour:

DYN-2381-CBN-Lacing-Before

Revised Behaviour

DYN-2381-CBN-Lacing-Complete

### Declarations

### Reviewers

### FYIs

@Amoursol

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
@@ -1126,8 +1126,10 @@ namespace Dynamo.ViewModels
 
         private void SetArgumentLacing(object parameter)
         {
-            var modelGuids = DynamoSelection.Instance.Selection.
-                OfType<NodeModel>().Select(n => n.GUID);
+            var modelGuids = DynamoSelection.Instance.Selection
+                .OfType<NodeModel>()
+                .Where(n => !(n is CodeBlockNodeModel))
+                .Select(n => n.GUID);
 
             if (!modelGuids.Any())
                 return;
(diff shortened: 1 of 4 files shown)


mmisol (commented):
While this PR does what it was asked for, this is not only a problem with Code Block nodes. Python Script nodes suffer the same issue and probably others do.

By taking a look at the code, I could see `NodeModel` has a property named `ArgumentLacing` that should be set to `LacingStrategy.Disabled` when lacing is not supported by the node. Using that, rather than excluding by type like in https://github.com/DynamoDS/Dynamo/pull/11294/files#diff-14fa24535a2d0aa34f993496cc35eb02dbadcffb7576a54cbeef2e9821edba8cR1131, should be the right way to solve this IMO.

StudioLE:
> While this PR does what it was asked for, this is not only a problem with Code Block nodes. Python Script nodes suffer the same issue and probably others do.
> 
> By taking a look at the code, I could see `NodeModel` has a property named `ArgumentLacing` that should be set to `LacingStrategy.Disabled` when lacing is not supported by the node. Using that, rather than excluding by type like in https://github.com/DynamoDS/Dynamo/pull/11294/files#diff-14fa24535a2d0aa34f993496cc35eb02dbadcffb7576a54cbeef2e9821edba8cR1131, should be the right way to solve this IMO.

Thanks @mmisol. I've addressed that in def9416

mmisol (approved):
Looks good. Thanks @StudioLE

PR #11301, opened 2020-12-01
Update libG for 2.10 RC

mjkkirschner:
### Purpose

bring in latest libG nugets fixing some crash bugs.
I will list change log here.

changelog:

* fix a crash when creating an arc or ellipse when doing a projection to a plane
* improve roboustness of parameter at point on curves

### Declarations

### DIFF
--- a/src/Config/Cleanup.targets
+++ b/src/Config/Cleanup.targets
@@ -8,6 +8,7 @@
       <FilesToDelete Include="$(OutputPath)nodes\SharpDX*"/>
       <FilesToDelete Include="$(OutputPath)nodes\HelixToolkit*"/>
       <FilesToDelete Include="$(OutputPath)nodes\Cyotek*"/>
+      <FilesToDelete Include="$(OutputPath)nodes\ProtoGeometry*"/>
     </ItemGroup> 
     <Delete Files="@(FilesToDelete)"/>
   </Target>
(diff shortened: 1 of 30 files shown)


aparajit-pratap:
Do we not want to wait for the LibG build fixes?

mjkkirschner:
@aparajit-pratap - sure, but should it matter? Because the actual builds here are done with a specified version of the compilers there should be no issues.

aparajit-pratap:
> @aparajit-pratap - sure, but should it matter? Because the actual builds here are done with a specified version of the compilers there should be no issues.

I suppose so.

src/Config/Cleanup.targets
```diff
@@ -8,6 +8,7 @@
       <FilesToDelete Include="$(OutputPath)nodes\SharpDX*"/>
       <FilesToDelete Include="$(OutputPath)nodes\HelixToolkit*"/>
       <FilesToDelete Include="$(OutputPath)nodes\Cyotek*"/>
+      <FilesToDelete Include="$(OutputPath)nodes\ProtoGeometry*"/>
```
aparajit-pratap, line 11: 👍

aparajit-pratap (approved):

src/Config/Cleanup.targets
mjkkirschner: so one thing I forgot to mention about this - is that these are not foolproof - these targets are only imported into a few projects, so if another project that builds after copies the specified binaries *again* we'll have the same trouble - we really should have a way to add a target for post build to the entire .sln - but I think thats not straight forward.

mjkkirschner: alternatively, we could import them into all projects as part of our standard import I guess.

mjkkirschner:
cherry picking to 2.10

PR #11318, opened 2020-12-03
CP: Handle errors when loading application icon

mmisol:
When loading the application icon to be displayed in the extension
windows, the method used would result in an unhandled error when the
application was started from a network folder. This error is now
handled, resulting only in the absence of the icon for the extension
window. Also, the method used to get the application path is changed
for a more reliable one.

### DIFF
--- a/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
+++ b/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
@@ -488,13 +488,21 @@ namespace Dynamo.Controls
         {
             if (this.Icon == null && !DynamoModel.IsTestMode)
             {
-                var icon = System.Drawing.Icon.ExtractAssociatedIcon(System.Reflection.Assembly.GetEntryAssembly().Location);
-                var bmp = icon.ToBitmap();
-                MemoryStream stream = new MemoryStream();
-                bmp.Save(stream, System.Drawing.Imaging.ImageFormat.Png);
-                stream.Seek(0, SeekOrigin.Begin);
-                PngBitmapDecoder pngDecoder = new PngBitmapDecoder(stream, BitmapCreateOptions.None, BitmapCacheOption.Default);
-                this.Icon = pngDecoder.Frames[0];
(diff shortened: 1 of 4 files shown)


src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -488,13 +488,21 @@ private void SetApplicationIcon()
...
-                var bmp = icon.ToBitmap();
-                MemoryStream stream = new MemoryStream();
-                bmp.Save(stream, System.Drawing.Imaging.ImageFormat.Png);
-                stream.Seek(0, SeekOrigin.Begin);
-                PngBitmapDecoder pngDecoder = new PngBitmapDecoder(stream, BitmapCreateOptions.None, BitmapCacheOption.Default);
-                this.Icon = pngDecoder.Frames[0];
+                var applicationPath = Process.GetCurrentProcess().MainModule.FileName;
+                try
+                {
+                    var icon = System.Drawing.Icon.ExtractAssociatedIcon(applicationPath);
```
mjkkirschner, line 494: @mmisol - does this result in the existing behavior for Revit icon loading?

mmisol: This gets the icon from the running application. Since the extension window does not have the standard window frame, one does not get an icon on it automatically. This is the way I found to actually get it, which as I tested works for Sandbox and Revit.

QilongTang: @mmisol @mjkkirschner Looking at this code again. We may have also fixed a bug that Dynamo4C3D showing icon as ACAD. I bet with the old code, entry Assembly is ACAD since Civil3D is on top of that. I cant really test it right now before C3D move up to .Net48. But will check at Global launch.

mmisol: @QilongTang This code is very recent. It was added with the first PR of the extension window #11246. It was not working for Revit, which is the main reason I changed it.

PR #11322, opened 2020-12-04
AGD-1875 As a user I need a node that generates a list of random numbers where the seed value can be set

saintentropy:
### Purpose

The purpose of this PR is to augment the existing Math.Random, Math.RandList, and List.Shuffle nodes with version which include a seed values.  While some workflows require a node that generates a new random number per Dynamo session, other workflows require random generated values that can be reproduced between Dynamo session. 

The new nodes with their non-seed pairs are illustrated bellow

image

Note the seed is set with a default value of 1.

The node thumbnails follow the same graphic pattern of the existing non-seed / seed pair utilizing an _N_ within the graphic to signify the seed value.

image

image

### Declarations

### Reviewers

@mrahmaniasl @LongNguyenP @QilongTang 

### FYIs

@mjkkirschner


src/Libraries/CoreNodes/Math.cs
```diff
@@ -17,7 +17,7 @@ namespace DSCore
     public static class Math
     {
         /// <summary>
-        ///     Generates a random double in the range of [0, 1).
+        ///     Generates a random double in the range of [0, 1].
```
QilongTang, line 20: Are we sure about this change? I thought it means we would never get 1 as a result but could get 0.

QilongTang:
Please double check the self CI error. Also @mjkkirschner I forgot if new node needs to be added to Libary.json file..

src/Libraries/CoreNodes/Math.cs
saintentropy: You are correct.  Will fix that.

Amoursol:
@saintentropy most likely a new PR, but given you are in the space right now: **List.Shuffle** currently works from the global time, and as such, if you try to use two lists to shuffle incoming data the results will always be the same. I wonder if it's worth looking at a slightly improvement there also for "True Random"? I think Matterlab have such a node that we could consume.

saintentropy:
@Amoursol I agree.  I don't think the behavior around the existing Random nodes is ideal.  Would the issue above be considered a bug and we would fix the existing node?

Amoursol:
> @Amoursol I agree. I don't think the behavior around the existing Random nodes is ideal. Would the issue above be considered a bug and we would fix the existing node?

That's tough, because it being 'true random' is exactly how users expect it to behave, but given it's been like this since inception users may have this design built into their scripts now. I do know that many users had to resort to Python or other random generator methods to get around this (As I personally did too!). 

Question: When you update the nodes above, if a user chose the same seed it would replicate like this existing behavior? If we release in tandem, the "fix" for the original broken behavior would be these nodes with an identical seed number?

mjkkirschner:
@saintentropy there is a failing test here - I think because these nodes have not been added explicitly to the layout json - they just work because they are overloads.

In general overloads are not so great in Dynamo, I think we should use other names that are similar.

saintentropy:
@mjkkirschner The test failure was due to the test/ViewExtensionLibraryTests/resources/libraryItems.json.  Now passes after updating that test resource.

jnealb:
@saintentropy Do we know which pseudo-random generator is used here? The standard C# lib?

saintentropy:
@Amoursol I added the fix for the timing bug related to shuffle.

image

saintentropy:
> @saintentropy Do we know which pseudo-random generator is used here? The standard C# lib?

Standard C# lib.  No change for the original implementation.

mjkkirschner:
@saintentropy please read the section on overloads and default params:
https://codeblog.jonskeet.uk/2018/04/13/backward-compatibility-and-overloading/

src/Libraries/CoreNodes/Math.cs
```diff
@@ -58,6 +72,29 @@ public static IList RandomList(int amount)
...
+        ///     Produces a list containing the given amount of random doubles
+        ///     in the deffined range of [0, 1) based on an intial seed value.
+        /// </summary>
+        /// <param name="amount">Amount of random numbers the result list will contain.</param>
+        /// <param name="value1">One end of the range for the random number.</param>
+        /// <param name="value2">One end of the range for the random number.</param>
+        /// <param name="seed">Seed value for the random number generator.</param>
+        /// <returns name="number">List of random numbers in the range.</returns>
+        /// <search>random,listcontains,seed</search>
+        public static IList RandomList(int amount, [DefaultArgument("0.0")] double value1, [DefaultArgument("1.0")] double value2, [DefaultArgument("1")] int seed)
```
mjkkirschner, line 85: What attribute is this?
Is it the designscript one?
If it is our own attribute then I think you can disregard my comment about the overload/default paramter issue.

mjkkirschner: I do recall a concerted effort at 2.x timeframe to remove overloads from the library - maybe @aparajit-pratap can say more there.

QilongTang (approved):

aparajit-pratap:
> @Amoursol I added the fix for the timing bug related to shuffle.
> 
> image

Wouldn't this break existing workflows even though it's the expected behavior?

src/Libraries/CoreNodes/List.cs
```diff
@@ -1257,7 +1257,20 @@ public static object LastItem(IList list)
         [IsVisibleInDynamoLibrary(true)]
         public static IList Shuffle(IList list)
         {
-            var rng = new Random();
+            return list.Cast<object>().OrderBy(_ => mRandom.Next()).ToList();
```
aparajit-pratap, line 1260: As commented earlier, is this not a breaking change in terms of breaking existing graph behaviour?

src/Libraries/CoreNodes/Math.cs
```diff
@@ -41,6 +41,20 @@ public static double Random(double value1, double value2)
             return result;
         }
 
+        /// <summary>
+        ///     Produce a random number in the range [lower_number, higher_number) based on an intial seed value.
+        /// </summary>
+        /// <param name="value1">One end of the range for the random number.</param>
```
aparajit-pratap, line 47: I think this should read `Lower end of the range ...`

src/Libraries/CoreNodes/Math.cs
```diff
@@ -41,6 +41,20 @@ public static double Random(double value1, double value2)
...
+        /// <param name="value2">One end of the range for the random number.</param>
```
aparajit-pratap, line 48: I think this should read `Higher end of the range ...`

aparajit-pratap (commented):
@saintentropy few questions.

src/Libraries/CoreNodes/List.cs
Amoursol: Existing behavior is non-breaking in any singular situation (i.e. using the List.Shuffle node once). 

In any dual behavior situations, the current node is **not working as expected** by the vast majority of users and shuffling uniquely. It is also incredibly difficult to find out why (For the average user) as you need to look into the codebase itself and understand that the .NET Framework default seed value is derived from the system clock - meaning any concurrently executing list.shuffle nodes with the same input data will shuffle to the same results. 

While technically this can be a breaking change, it's currently more expensive for users to try and resolve than necessary. You have to resort to Python, a custom node or create your own C# node which is beyond many users, either through creation or simply finding it. 

With the addition of the nodes above from @saintentropy, the 'fix' to retain existing behavior is simply to add the same seed value.

PR #11336, opened 2020-12-07
Autocomplete autogenerated static properties

aparajit-pratap:
### Purpose

[DYN-996](https://jira.autodesk.com/browse/DYN-996)

You can refer to `ClassFunctionality` class defined in `FFITarget` to verify these examples.

Instance properties now appear in autocomplete by typing `.` after the class name and they can be invoked like methods by passing the instance as the first parameter:
image
image

Functions names starting with a `get_` can be invoked in a CBN just like getter properties, e.g. `obj.get_Method()` can be invoked as `Class.Method(obj)` even though this isn't allowed in C#. CBN autocomplete is made consistent with this.
image

### Declarations

### FYIs


src/DynamoCore/Engine/CodeCompletion/CodeCompletionServices.cs
```diff
@@ -298,7 +300,10 @@ internal CompletionData(string text, CompletionType type)
             if (property != null)
             {
                 string propertyName = property.PropertyName;
-                return new CompletionData(propertyName, CompletionType.Property);
+                return new CompletionData(propertyName, CompletionType.Property)
```
mjkkirschner, line 303: would it make sense to add a new constructor overload?

src/Engine/ProtoCore/Reflection/Mirror.cs
```diff
@@ -391,16 +391,24 @@ public IEnumerable<PropertyMirror> GetProperties()
...
                     {
-                        properties.Add(new PropertyMirror(pNode));
+                        // Auto-generated static property that takes the instance as the only input parameter.  
+                        var isAutoGeneratedThisProperty = pNode.IsAutoGeneratedThisProc && pNode.ArgumentInfos.Count == 1 &&
+                            pNode.ArgumentTypes[0].UID == pNode.ClassID;
+
+                        if (pNode.ArgumentInfos.Count == 0 || isAutoGeneratedThisProperty)
+                        {
+                            properties.Add(new PropertyMirror(pNode));
+                        }
```
mjkkirschner, line 411: is there a case where a static property might somehow have 2 parameters? Doesn't seem possible?

src/Engine/ProtoCore/Reflection/Mirror.cs
```diff
@@ -391,16 +391,24 @@ public IEnumerable<PropertyMirror> GetProperties()
...
                     name = pNode.Name;
                     bool hidden = pNode.MethodAttribute == null ? false : pNode.MethodAttribute.HiddenInLibrary;
-                    if (!hidden && name.Contains(getterPrefix) && pNode.ArgumentInfos.Count == 0)
+
+                    if (!hidden && name.Contains(getterPrefix))
                     {
-                        properties.Add(new PropertyMirror(pNode));
+                        // Auto-generated static property that takes the instance as the only input parameter.  
+                        var isAutoGeneratedThisProperty = pNode.IsAutoGeneratedThisProc && pNode.ArgumentInfos.Count == 1 &&
+                            pNode.ArgumentTypes[0].UID == pNode.ClassID;
```
mjkkirschner, line 406: curious if this last check if required - I think it's good! - I just want to know what case is being filtered out?

src/Engine/ProtoCore/Reflection/Mirror.cs
```diff
@@ -650,6 +659,36 @@ internal PropertyMirror(ProcedureNode procNode, bool isSetter = false)
                 this.Name = PropertyName;
                 this.isSetter = isSetter;
             }
+
+            private string Argument
```
mjkkirschner, line 663: would you mind adding a comment for what this is supposed to produce?

src/Engine/ProtoCore/Reflection/Mirror.cs
```diff
@@ -650,6 +659,36 @@ internal PropertyMirror(ProcedureNode procNode, bool isSetter = false)
...
+            /// "Property : 'return type' (this : 'class type')."
+            /// For a static property of a class,  this returns
+            /// "Property : 'return type'."
+            /// </summary>
+            /// <returns></returns>
+            public override string ToString()
+            {
+                var sb = new StringBuilder();
+
+                var returnType = " : " + procNode.ReturnType.ToString().Split('.').Last();
```
mjkkirschner, line 687: is this standard, that we only return the typeName not the fully qualified type name? Or am I misinterpreting this?

src/Engine/ProtoCore/Reflection/Mirror.cs
aparajit-pratap: There could be methods beginning with a `get_` that may take arguments that do not match with the class type. This is to filter out those methods.

src/Engine/ProtoCore/Reflection/Mirror.cs
aparajit-pratap: I don't think so.

src/Engine/ProtoCore/Reflection/Mirror.cs
aparajit-pratap: The static getter equivalent of a property will have exactly one argument.

src/Engine/ProtoCore/Reflection/Mirror.cs
aparajit-pratap: We display just the type in the function signature insight window just to be less verbose.

src/Engine/ProtoCore/Reflection/Mirror.cs
mjkkirschner: makes sense.

src/Engine/ProtoCore/Reflection/Mirror.cs
mjkkirschner: it just seems this could be confusing when dealing with multiple `Point` types for example - I guess we could use the elementResolver or VM to determine the shortest qualified name that is explicit? Maybe a future improvment?

mjkkirschner (approved):

src/Engine/ProtoCore/Reflection/Mirror.cs
aparajit-pratap: Yes, we could. For now, I've made it consistent with method overload completion in the insight window:
image

mjkkirschner:
@aparajit-pratap I was trying out this part of your PR description:

>  names starting with a get_ can be invoked in a CBN just like getter properties, e.g. obj.get_Method() can be invoked as Class.Method(obj) even though this isn't allowed in C#. CBN autocomplete is made consistent with this.

but I seem to get a missing method exception:
Screen Shot 2020-12-09 at 6 16 26 PM

maybe I misunderstanding?

mjkkirschner:
yep, I read this wrong, for some reason I interpreted this to mean that you could now call properties like methods - but I see you are stating that we can always could call properties as methods on classes when passing instances as arguments, and now the PR enables autocomplete to be consistent with that ability.

- apologies for the confusion.

mjkkirschner:
In the image below, `ClassFuncChild` is a class that derives from ClassFunctionality:

`ClassFuncChild : ClassFunctionality`

It's neat that this works, but should it?

Screen Shot 2020-12-09 at 6 39 32 PM

aparajit-pratap:
> In the image below, `ClassFuncChild` is a class that derives from ClassFunctionality:
> 
> `ClassFuncChild : ClassFunctionality`
> 
> It's neat that this works, but should it?
> 
> Screen Shot 2020-12-09 at 6 39 32 PM
> 
> also, it shows up in autocomplete which is interesting... should static properties and methods be callable like that from a child class name?

@mjkkirschner yes, I think this is supported in DS and in order to make autocomplete aligned with this behaviour completing a class also suggests all methods and properties on not just that class but also those that it derives from its base class(es).

mjkkirschner:
neat, what I didn't try was implementing "StaticMethod" or "StaticProp" on the derived class to see what happened in autocomplete.

PR #11339, opened 2020-12-07
CP: #11322 #11338 #11342

QilongTang:
Please Note:

### Purpose

CP: #11322 #11338 #11342

Validated
image

### Declarations

### FYIs

@DynamoDS/dynamo


mmisol (changes requested):
I think we should make sure this get tested before we merge it in 2.10. Specially since it doesn't contain unit tests

QilongTang:
@mmisol Yes, just create the PR to get self CI running for now. Behavior is getting tested

QilongTang:
Validated by me and @mjkkirschner. Merging and asking @saintentropy to append unit test to master branch for these. This is to unblock the RC2.10.0 release and localization.

mmisol:
If this was tested manually that sounds good. No need to ask for unit tests retroactively.

src/Libraries/CoreNodes/Math.cs
```diff
@@ -17,7 +17,7 @@ namespace DSCore
     public static class Math
     {
         /// <summary>
-        ///     Generates a random double in the range of [0, 1).
+        ///     Generates a random double in the range of (0, 1).
```
mmisol, line 20: Actually, `[` was right here because according to the docs the value can be 0: https://docs.microsoft.com/en-us/dotnet/api/system.random.nextdouble?view=net-5.0

src/Libraries/CoreNodes/Math.cs
```diff
@@ -29,21 +29,35 @@ public static double Random(int? seed = null)
...
 
         /// <summary>
-        ///     Produce a random number in the range [lower_number, higher_number).
+        ///     Produce a random number in the range (lower_number, higher_number).
         /// </summary>
-        /// <param name="value1">One end of the range for the random number.</param>
-        /// <param name="value2">One end of the range for the random number.</param>
-        /// <returns name="number">Random number in the range [lowValue, highValue).</returns>
+        /// <param name="value1">Lower range for the random number.</param>
+        /// <param name="value2">Higher range for the random number.</param>
```
mmisol, line 35: This should be `Lower/Higher end of the range for the random number.`

src/Libraries/CoreNodes/Math.cs
QilongTang: @mmisol We did see this doc but from our own testing with millions of numbers the 0 was never hit.

mmisol: @QilongTang The value space is huge. I can say for sure you didn't get all possible values 😄

QilongTang: After numerous testing, @mjkkirschner and I decided to update the range here as majority of our users would benefit from this..

mmisol: I don't see how giving misleading documentation could benefit users.

mjkkirschner: @mmisol you could be right of course - I am not a statistician -  my hypothesis was that we should see a 0 once or twice in a million results - but... I guess I am super wrong:

https://stackoverflow.com/questions/58241868/can-random-uniform0-1-ever-generate-0-or-1

the response here is a few quadrillion... I'm not so sure about that either...

I think the result of system.random and our own method is confused by the fact that we do some scaling - so it's not clear if we actually have returned 0 and the result is loosing precision.

PR #11341, opened 2020-12-07
DYN-3314: Simplify html sanitization so that html sanitization always returns a valid body element

sm6srw:
### Purpose

[DYN-3314](https://jira.autodesk.com/browse/DYN-3314) simplify html sanitization so that html sanitization always returns a valid body element

This pull request does:
* change the markdown to html conversion so it always returns a full html document.
* change the `md2html` tool so it only sanitize full documents.
* allow some new schemes/attributes/properties when sanitizing that reflects how our embedded content is setup.
* force the use of the `table--noborder` class when generating the node info doc table in a way that I think always works.
* add dynamic tests that checks if anything gets removed when sanitizing our embedded content 

### Declarations

### Reviewers

@mjkkirschner

### DIFF
--- a/src/Tools/Md2Html/Md2HtmlSanitizer.cs
+++ b/src/Tools/Md2Html/Md2HtmlSanitizer.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using AngleSharp.Css.Dom;
 using Ganss.XSS;
 
 namespace Md2Html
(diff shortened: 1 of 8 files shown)


src/Tools/Md2Html/Md2HtmlSanitizer.cs
```diff
@@ -20,8 +21,24 @@ internal Md2HtmlSanitizer()
...
+            AllowedCssProperties.Add(@"-o-tab-size");
+            AllowedCssProperties.Add(@"tab-size");
+            AllowedCssProperties.Add(@"-webkit-hyphens");
+            AllowedCssProperties.Add(@"-moz-hyphens");
+            AllowedCssProperties.Add(@"-ms-hyphens");
+            AllowedCssProperties.Add(@"hyphens");
+            AllowedCssProperties.Add(@"background-position-x");
+            AllowedCssProperties.Add(@"background-position-y");
+
+            AllowedSchemes.Add(@"file");
```
mjkkirschner, line 39: this is for images loading?

sm6srw: Yes

mjkkirschner: these look fine to me in general - I guess I would like us to do a bit of research on "file" to see if this could somehow be used to load and execute arbitrary js - to me it looks like you'd need another script tag to do that, but I'm not sure.

sm6srw: We only need this for loading images from markdown files. I think markdown do support embedded images so we could make that an requirement for all images? We do not allow script tags at all so there is no way to misuse `file` there. I will do some digging to see if there is a way to load js outside a script tag.

mjkkirschner (approved):

PR #11347, opened 2020-12-08
[AGD-1799] Support persisting extension and view extension data

SHKnudsen:
### Purpose

This PR adds a new ExtensionStorage interface which can be implemented on `Extensions` and `ViewExtension` to give the author of the extension the possibility to store extension data directly in the DYN file. 

The interface adds two new methods `OnWorkspaceOpen` and `OnWorkspaceSaving` which are triggered when a graph is opened and when its saved. Both methods passes a `Dictionary<string,string>` which can be used to store extension data in, `OnWorkspaceSaving` also passes a `SaveContext` so the implementer can react to different save contexts (i.e. save, save as).

The `WorkspaceModel` stores a collection of `ExtensionData` objects which is serialized to the DYN file on save. The `OnWorkspaceOpen` passes a copy of the extension data dictionary therefor modifications done to this will not be reflected in the `ExtensionData` object (FYI @saintentropy). `OnWorkspaceSaving` passes a direct reference to the `ExtensionData` objects dictionary and there for modification made to this will be reflected in the DYN file.

**TODO**

- Determine the max size of the dictionary stored in the DYN file
- Add extension version to the `ExtensionData` model

image

### Declarations

### Reviewers

@saintentropy 
@nate-peters 

### FYIs

@mjkkirschner 
@QilongTang


src/DynamoCore/Extensions/ExtensionManager.cs
```diff
@@ -139,6 +146,12 @@ public void Remove(IExtension extension)
                 Log(fullName + " extension cannot be disposed properly: " + ex.Message);
             }
 
+            if (extension is IExtensionStorageAccess storageAccess &&
+                storageAccessExtensions.Find(x => (x as IExtension).UniqueId == extension.UniqueId) is null)
```
mjkkirschner, line 150: shouldn't that be not null? I may be coffee deprived.

src/DynamoCore/Extensions/ExtensionManager.cs
```diff
@@ -33,6 +33,7 @@ public interface IExtensionSource
     public class ExtensionManager: IExtensionManager, ILogSource
     {
         private readonly List<IExtension> extensions = new List<IExtension>();
+        private readonly List<IExtensionStorageAccess> storageAccessExtensions = new List<IExtensionStorageAccess>();
```
mjkkirschner, line 36: maybe use a hashset or dictionary of guid to extensions?

src/DynamoCore/Extensions/ExtensionManager.cs
```diff
@@ -154,6 +167,14 @@ public IEnumerable<IExtension> Extensions
             get { return extensions; }
         }
 
+        /// <summary>
+        /// Returns the collection of registered extensions implementing IExtensionStorageAccess
+        /// </summary>
+        public IEnumerable<IExtensionStorageAccess> StorageAccessesExtensions
+        {
+            get { return storageAccessExtensions; }
```
mjkkirschner, line 175: why not use an autoproperty?

src/DynamoCore/Graph/ModelBase.cs
```diff
@@ -13,7 +13,7 @@ namespace Dynamo.Graph
     /// <summary>
     /// SaveContext represents several contexts, in which node can be serialized/deserialized.
     /// </summary>
-    public enum SaveContext { File, Copy, Undo, Preset, None };
+    public enum SaveContext { Save, SaveAs, File, Copy, Undo, Preset, None };
```
mjkkirschner, line 16: these need to go at the end of the enum - otherwise this is an API break.

mjkkirschner: I wonder if `File` is used already to mean both save and saveas?

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
```diff
@@ -640,21 +642,33 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
...
                     info, verboseLogging, isTestMode);
-            }
+            };
 
             ws.NodeLibraryDependencies = nodeLibraryDependencies.ToList();
+            ws.ExtensionData = GetExtensionData(serializer, obj);
 
             return ws;
         }
+                        //{ ExtensionData = GetExtensionData(serializer, obj) };
```
mjkkirschner, line 661: ?

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
```diff
@@ -759,6 +773,10 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
             writer.WritePropertyName(WorkspaceReadConverter.NodeLibraryDependenciesPropString);
             serializer.Serialize(writer, ws.NodeLibraryDependencies);
 
+            // ExtensionData
+            writer.WritePropertyName("ExtensionWorkspaceData");
```
mjkkirschner, line 777: maybe use `nameOf` or pull this name into a const.

src/DynamoCore/Extensions/IExtensionStorageAccess.cs
```diff
@@ -0,0 +1,29 @@
...
+        void OnWorkspaceOpen(Dictionary<string, string> extensionData);
+
+        /// <summary>
+        /// Action to be invoked when the workspace has begun its saving process.
+        /// The passed extensionData dictionary is a direct reference to the stored <see cref="ExtensionData"/> data dictionary, modifications to this dictionary
+        /// will be reflected in the stored <see cref="ExtensionData"/> data dictionary
+        /// </summary>
+        /// <param name="extensionData"></param>
+        /// <param name="saveContext"></param>
+        void OnWorkspaceSaving(Dictionary<string, string> extensionData, SaveContext saveContext);
```
mjkkirschner, line 27: have you checked what happens during backup save? Does anything special need to be handled for that case?

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -1158,6 +1177,7 @@ public virtual void Save(string filePath, bool isBackup = false, EngineControlle
                 var json = this.ToJson(engine);
 
                 // Stage 2: Save
+                OnSaving(SaveContext.None);
```
mjkkirschner, line 1180: None?

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -2096,5 +2116,34 @@ private Guid IdToGuidConverter(string id)
...
+            var extensionData = ExtensionData.Where(x => x.ExtensionGuid == uniqueId).FirstOrDefault();
+            if (extensionData is null)
+                return;
+            extensionData.Data = data;
+        }
+
+        internal void CreateNewExtensionData(string uniqueId, string name, Dictionary<string, string> data)
+        {
+            // TODO: Figure out how to add extension version when creating new ExtensionData 
+            var extensionData = new ExtensionData(uniqueId, name, "NotSureHowWeShouldDoThis...", data);
```
mjkkirschner, line 2145: I forgot what we decided here, I think we said grab the version number from the extension assembly?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1704,8 +1776,10 @@ private void OpenWorkspace(WorkspaceModel ws)
 
             AddWorkspace(ws);
             CurrentWorkspace = ws;
+            OnWorkspaceOpend(ws);
```
mjkkirschner, line 1779: spelling (opened)

src/DynamoCore/Extensions/ExtensionManager.cs
SHKnudsen: this is probably a case of me over complicating stuff, i wanted to follow the existing pattern of adding to the `extensions` list. The reason its checking if its null is to make sure its not already in the `storageAccessExtensions` List, so it only gets added if it isnt there already.

But as you mention a dict or hashset might be more suitable, this was mainly done to just follow the existing implementation.

src/DynamoCore/Extensions/ExtensionManager.cs
SHKnudsen: No reason not to, just wanted to follow same pattern that is used for the `extensions` collection.

src/DynamoCore/Extensions/IExtensionStorageAccess.cs
SHKnudsen: im not raising this on backups, same as with `OnSaved()`

https://github.com/DynamoDS/Dynamo/pull/11347/files#diff-14fa24535a2d0aa34f993496cc35eb02dbadcffb7576a54cbeef2e9821edba8cR558-R559

src/DynamoCore/Extensions/ExtensionManager.cs
mjkkirschner: If theres no reason - use the autoproperty - less code, same effect. I'm not sure if the extensions backing field is accessed directly somewhere for some reason?

src/DynamoCore/Extensions/ExtensionManager.cs
mjkkirschner: right - but isn't this the remove section? You'd want the extension to be present before removing it right?

src/DynamoCore/Extensions/IExtensionStorageAccess.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Threading.Tasks;
+using Dynamo.Graph;
+
+namespace Dynamo.Extensions
+{
+    public interface IExtensionStorageAccess
+    {
+        /// <summary>
+        /// Action to be invoked when the workspace is opened. The passed extensionData dictionary is a copy of the <see cref="ExtensionData"/> data dictionary.
+        /// Modifying the extensionData from this action will not modify the stored <see cref="ExtensionData"/>.
```
mjkkirschner, line 14: is that actually true? Do you make a copy of the data first to prevent modifications to the underlying dictionary? This would be a good test.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -674,6 +684,14 @@ public string Description
             }
         }
 
+        /// <summary>
+        /// List of user defined data from extensions and view extensions stored in the graph
+        /// </summary>
+        internal ICollection<ExtensionData> ExtensionData
```
mjkkirschner, line 690: This is an `ICollection` because you need to Add to it?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -2589,6 +2663,8 @@ public void AddWorkspace(WorkspaceModel workspace)
 
             Action savedHandler = () => OnWorkspaceSaved(workspace);
             workspace.Saved += savedHandler;
+            Action<SaveContext> savingHandler = (c) => OnWorkspaceSaving(workspace,c);
+            workspace.WorkspaceSaving += savingHandler;
```
mjkkirschner, line 2673: can this be unsubscribed somewhere?

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -1578,6 +1579,33 @@ public void SaveAs(string path, bool isBackup = false)
...
+        /// <param name="path">The path to the file.</param>
+        /// <param name="saveContext">The context of the save operation.</param>
+        /// <param name="isBackup">Indicates if an automated backup save has called this function.</param>
+        /// <exception cref="IOException"></exception>
+        /// <exception cref="UnauthorizedAccessException"></exception>
+        public void SaveAs(string path, SaveContext saveContext, bool isBackup = false)
+        {
+            try
+            {
+                Model.Logger.Log(String.Format(Properties.Resources.SavingInProgress, path));
```
mjkkirschner, line 1595: it seems one of these methods could just call the other instead of duplicating code, or use a shared private method.

mjkkirschner:
please add tests to your TODO list.

src/DynamoCore/Extensions/IExtensionStorageAccess.cs
SHKnudsen: yes making a copy here https://github.com/DynamoDS/Dynamo/pull/11347/files#diff-d730c0ae93cb1da6c95f0ba453dd94553067feddaedf354cb63a6bfcde71ffcfR990-R991, and passing that to the `OnWorkspaceOpen`

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
SHKnudsen: yep

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -930,6 +956,55 @@ private void RemoveExtension(IExtension ext)
...
+            foreach (var extension in extensionManager.StorageAccessesExtensions)
+            {
+                var ext = extension as IExtension;
+                RaiseIExtensionStorageAccessWorkspaceOpened(workspace, extension, ext.UniqueId);
+            }
+        }
+
+        private void HandleStorageExtensionsOnWorkspaceSaveing(WorkspaceModel workspace, SaveContext saveContext)
+        {
+            if (!extensionManager.StorageAccessesExtensions.Any())
```
mjkkirschner, line 974: this check doesn't seem very useful - it's small potatoes since the number of extensions should be small, but we might have to iterate until the last entry in the extensions array for no reason, nothing will happen in the for loop if the StorageAccessExtensions collection is empty.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -2096,5 +2118,39 @@ private Guid IdToGuidConverter(string id)
 
             return deterministicGuid;
         }
+
+        internal bool GetMatchingWorkspaceData(string uniqueId, out Dictionary<string, string> data)
```
mjkkirschner, line 2122: this seems like it should be called `TryGetMatchingWorkspaceData`

src/DynamoCore/Extensions/ExtensionManager.cs
SHKnudsen: Actually when thinking about it, we dont want to enable adding new `IExtensionStorageAccess` anywhere else than from the managers so therefor it makes sense to return an IEnumerable like this.

If we have it as an autoproperty we would have to return the actual collection which would enable adding IExtensionStorageAccess to the collection from other classes. So seems to me this is the way to do it with less code. Unless there is some other way to achieve that which im not familiar with?

src/DynamoCore/Extensions/ExtensionManager.cs
SHKnudsen: Oh right!! 🤦‍♂️

My bad, you are absolutely right!

src/DynamoCore/Extensions/ExtensionManager.cs
mjkkirschner: can you explain why a property without a backing field would force you to return the actual collection?

SHKnudsen: Sorry if im misunderstanding this, but we cant add directly to an `IEnumerable`.

So the reason for having a backing field is that we in the `ExtensionManager` can add new extensions to the `List<IExtensionStorageAccess>` but we only expose the `IEnumerable` to the outside. 

I think im struggling to see how we can do it with less code than
```csharp
private List<IExtensionStorageAccess> storageAccessExtensions = new List<IExtensionStorageAccess>(); 

public IEnumerable<IExtensionStorageAccess> StorageAccessExtensions { get { return storageAccessExtensions; }}
```

I guess we could use an `ReadOnlyCollection` but havent really looked into them before, and not sure it would minimize the amount code very much.

mjkkirschner: I *think* you can write this:

```
public IEnumerable<IExtensionStorageAccess> StorageAccessExtensions {get;} = new List<IExtensionStorageAccess>();

```

or 

```

some internal method(){

StorageAccessExtensions = new List<IExtensionStorageAccess>();

}
```

`IEnumerable<T>` is just an interface, the concrete type can still be a list. BUT I see that we do a bunch of internal stuff to that backing field which needs to be an `ICollection` and I *think* we'd have to do a bunch of funny casts for no reason so this looks good to me. 👍

src/DynamoCore/Graph/ModelBase.cs
SHKnudsen: @mjkkirschner i can find `File` being used in these locations, seems to be mostly used with Xml stuff. Not sure what `File` represents though?

https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Graph/NodeGraph.cs#L61
https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Graph/NodeGraph.cs#L130
https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Graph/NodeGraph.cs#L149
https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Graph/NodeGraph.cs#L187
https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Models/RecordableCommands.cs#L789
https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Models/DynamoModelCommands.cs#L151
https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs#L1597

SHKnudsen:
@saintentropy @mjkkirschner moved this out of `draft` state. Let me know if there are any other comments that needs to be addressed.

src/DynamoCore/Graph/ModelBase.cs
mjkkirschner: thanks for digging these up, it appears to represent any save (or loading in some cases) action being done from files - IE load some nodes from a file or save some nodes - as opposed to undo redo etc because those all used to use xml serialization and deserialization.


It's not great that we now have File and Save,SaveAs since these now overlap and it will not be clear to anyone using this Enum in the future which to use - I think the simplest solution is just to obsolete `File` and replace it with `Save` ... anywhere we would have checked for `File` we can check for `File | Save | SaveAs` - what do you think?

SHKnudsen: Yep that makes sense!

Just made a new update with these changes.

mjkkirschner:
@SHKnudsen there are around ~80 test failures in the last build - I'll kick it off again incase it's a fluke, but you may want to take a look at the  PR test validation link above.

SHKnudsen:
Turned out it was just me that had messed a few `if` statements up, should be fixed now.

mjkkirschner:
@SHKnudsen only one failure left 👍 
https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/1955/testReport/DynamoCoreWpfTests/SerializationTests/JSONisSameBeforeAndAfterSaveWithDummyNodes/

SHKnudsen:
@mjkkirschner  you where right, the test was failing because we are now adding a `ExtensionWorkspaceData` property on save. Im ignoring that property when comparing the two json files in the test, so that should be fixed now.

mjkkirschner (approved):

mjkkirschner:
@nate-peters @saintentropy - any other comments?

mjkkirschner:
Is there scope to add a sample extension to the dynamo samples repo illustrating use of this interface?

SHKnudsen:
> Is there scope to add a sample extension to the dynamo samples repo illustrating use of this interface?

I think i can find some scope to do this, if you guys are happy with that.

@saintentropy

src/DynamoCore/Graph/ModelBase.cs
```diff
@@ -13,7 +13,7 @@ namespace Dynamo.Graph
     /// <summary>
     /// SaveContext represents several contexts, in which node can be serialized/deserialized.
     /// </summary>
-    public enum SaveContext { File, Copy, Undo, Preset, None };
+    public enum SaveContext { [Obsolete("Use Save or SaveAs, instead of File")] File, Copy, Undo, Preset, None, Save, SaveAs };
```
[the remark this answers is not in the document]
mjkkirschner: I believe this is safe as long as the new enums are only added to the end of the enum. The reason is because these are simply compiled to `int`s in increasing order. (as long as the ints are not specified - I think that is actually even safer)

PR #11350, opened 2020-12-09
Do not implement props in ViewExtensionBase

mmisol:
### Purpose

This changes the definition of ViewExtensionBase to use abstract rather
than virtual for the definition of required properties like Name and
UniqueId. The use of virtual was creating a backing field which was not
used in ViewExtensionBase and could not be assigned in the inheriting
class, making it pretty much useless.

Also the Dispose method is declared as abstract, as it was already
required to implement and having a default empty implementation could
lead developers to oversight the need to dispose resources in their
extension.

### Declarations

### DIFF
--- a/src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
+++ b/src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
@@ -1,13 +1,13 @@
-﻿using Dynamo.DocumentationBrowser.Properties;
-using Dynamo.Logging;
-using Dynamo.PackageManager;
-using Dynamo.ViewModels;
-using Dynamo.Wpf.Extensions;
-using System;
+﻿using System;
 using System.ComponentModel;
 using System.Linq;
 using System.Windows;
 using System.Windows.Controls;
+using Dynamo.DocumentationBrowser.Properties;
+using Dynamo.Logging;
+using Dynamo.PackageManager;
+using Dynamo.ViewModels;
+using Dynamo.Wpf.Extensions;
 
 namespace Dynamo.DocumentationBrowser
 {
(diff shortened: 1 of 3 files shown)


mmisol:
While I was making this PR @QilongTang told me the release of 2.10 is already done. Releasing this in 2.11 would make this a technically a breaking change with respect to 2.10. However, that breaking change will actually reveal issues in the extension developed, so it could be argued whether this is actually beneficial.

aparajit-pratap (commented):
~~While this looks like a good move in the right direction, it could be a breaking change, right?~~ Ok, just saw your comment above.

aparajit-pratap:
I actually am confused about the purpose of `ViewExtensionBase` as opposed to the interface, `IViewExtension`. Not all view extensions derived from ViewExtensionBase.

mmisol:
I guess the discussion for the need of `ViewExtensionBase` came from the discussions [here](https://github.com/DynamoDS/Dynamo/pull/11242) and [here](https://github.com/DynamoDS/Dynamo/pull/11260). As a summary, it is used as a way to extend the view extension API without adding a new member in the `IViewExtension` interface. Extension developers can implement these members by inheriting from the `ViewExtensionBase` class. Dynamo dynamically checks whether the extension actually implements the base class and, if that case, it calls them. For now this is used for the `Closed` function that was recently added.

aparajit-pratap:
> I guess the discussion for the need of `ViewExtensionBase` came from the discussions [here](https://github.com/DynamoDS/Dynamo/pull/11242) and [here](https://github.com/DynamoDS/Dynamo/pull/11260). As a summary, it is used as a way to extend the view extension API without adding a new member in the `IViewExtension` interface. Extension developers can implement these members by inheriting from the `ViewExtensionBase` class. Dynamo dynamically checks whether the extension actually implements the base class and, if that case, it calls them. For now this is used for the `Closed` function that was recently added.

Ah, now I remember. Thanks for the context. In that case, it seems less risky to break this at this stage.

mjkkirschner:
Is this a big enough issue that we should cherry pick to 2.10 and rtm that build?

mjkkirschner (approved):

PR #11353, opened 2020-12-09
Adding a view element to let users expand the watch node.

reddyashish:
### Purpose

This PR adds a functionality to let the users expand the watch node so that they can view the whole content. 

Task: https://jira.autodesk.com/browse/DYN-2457

After discussing with @Amoursol and @QilongTang, we thought it would be useful to add a thumb control to expand the watch node to a desirable size. This thumb control to similar to the one, that is present in watch3D node. 

watch node expansion

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
+++ b/src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
@@ -240,7 +240,7 @@
         </ResourceDictionary>
     </UserControl.Resources>
 
-    <Grid Name ="inputGrid" MinHeight="100" MinWidth="100">
+    <Grid Name ="inputGrid">
         <Grid.RowDefinitions>
             <RowDefinition Height ="*" />
             <RowDefinition Height="Auto" />
--- a/src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
+++ b/src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
@@ -17,7 +17,6 @@
                 <ui:SharedResourceDictionary Source="{x:Static ui:SharedDictionaryManager.DynamoConvertersDictionaryUri}" />
                 <ui:SharedResourceDictionary Source="{x:Static ui:SharedDictionaryManager.DynamoModernDictionaryUri}" />
             </ResourceDictionary.MergedDictionaries>
-            <controls:BoolToVisibilityCollapsedConverter x:Key="BoolToVisibilityCollapsedConverter" />
             <Style x:Key="BorderlessButton"
                    TargetType="{x:Type Button}">
                 <Setter Property="Padding"
--- a/src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
+++ b/src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
@@ -7,8 +7,6 @@
              xmlns:ui="clr-namespace:Dynamo.UI"
              xmlns:controls="clr-namespace:Dynamo.Controls"
              mc:Ignorable="d"
-             Width="300"
-             Height="300"
              HorizontalAlignment="Stretch"
              VerticalAlignment="Stretch"
              BorderBrush="Black">
@@ -392,7 +390,7 @@
                HorizontalAlignment="Right"
                Margin="5"
                DragDelta="ThumbResizeThumbOnDragDeltaHandler"
-               Visibility="{Binding IsResizable, Converter={StaticResource BoolToVisibilityCollapsedConverter}}"/>
+               Visibility="Hidden"/>
           </Grid>
         </Border>
     </Grid>
--- a/src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
+++ b/src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
@@ -19,7 +19,6 @@
                 <ui:SharedResourceDictionary Source="{x:Static ui:SharedDictionaryManager.DynamoConvertersDictionaryUri}" />
                 <ui:SharedResourceDictionary Source="{x:Static ui:SharedDictionaryManager.DynamoModernDictionaryUri}" />
             </ResourceDictionary.MergedDictionaries>
-            <controls:BoolToVisibilityConverter x:Key="BoolToVisibilityConverter" />
             <controls:BoolToVisibilityCollapsedConverter x:Key="BoolToVisibilityCollapsedConverter" />
             <Style x:Key="BorderlessButton"
                    TargetType="{x:Type Button}">
--- a/src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
+++ b/src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
@@ -335,7 +335,7 @@
                 <ColumnDefinition Width="Auto" />
                 <ColumnDefinition Width="*" />
                 <ColumnDefinition Width="Auto" />
-                <ColumnDefinition Width="8" />
+                <ColumnDefinition Width="Auto" />
             </Grid.ColumnDefinitions>
             
             <!-- Shows counts of all items in List -->
@@ -388,10 +388,9 @@
                 </ListView.ItemTemplate>
             </ListView>
 
-            <Thumb Name ="resizeThumb" Grid.Column ="1"
+            <Thumb Name ="resizeThumb" Grid.Column ="3"
                Width="10" Height="10" 
-               HorizontalAlignment="Right" 
-               VerticalAlignment="Bottom" 
+               HorizontalAlignment="Right"
                Margin="5"
                DragDelta="ThumbResizeThumbOnDragDeltaHandler"
                Visibility="{Binding IsResizable, Converter={StaticResource BoolToVisibilityCollapsedConverter}}"/>
--- a/src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
+++ b/src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
@@ -5,13 +5,13 @@
              xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
              xmlns:viewModels="clr-namespace:Dynamo.ViewModels"
              xmlns:ui="clr-namespace:Dynamo.UI"
+             xmlns:controls="clr-namespace:Dynamo.Controls"
              mc:Ignorable="d"
-             d:DesignHeight="161"
-             d:DesignWidth="410"
+             Width="300"
+             Height="300"
              HorizontalAlignment="Stretch"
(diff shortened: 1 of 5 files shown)


Amoursol:
@reddyashish looking great as a first pass!

Can you please switch the location of the **item count** and **resize box** in these? So that the box is justified to the lower right, and that the count is justified to the bottom, but to the left hand side of the box?

mjkkirschner:
@reddyashish - I think you'll have to consider opening a graph and how to serialize/deserialize the width and height now. This should probably be in the view data for the watch node.

reddyashish:
Looks like this now. 
Screen Shot 2020-12-09 at 4 50 28 PM

QilongTang:
> @reddyashish - I think you'll have to consider opening a graph and how to serialize/deserialize the width and height now. This should probably be in the view data for the watch node.

Yes, I believe this is part of the AC as well.

src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
```diff
@@ -5,20 +5,22 @@
...
-             MaxHeight="300">
+             BorderBrush="Black">
     <UserControl.Resources>
         <ResourceDictionary>
             <ResourceDictionary.MergedDictionaries>
                 <ui:SharedResourceDictionary Source="{x:Static ui:SharedDictionaryManager.DynamoColorsAndBrushesDictionaryUri}" />
                 <ui:SharedResourceDictionary Source="{x:Static ui:SharedDictionaryManager.DynamoConvertersDictionaryUri}" />
                 <ui:SharedResourceDictionary Source="{x:Static ui:SharedDictionaryManager.DynamoModernDictionaryUri}" />
             </ResourceDictionary.MergedDictionaries>
+            <controls:BoolToVisibilityConverter x:Key="BoolToVisibilityConverter" />
```
QilongTang, line 22: Where is this converter used?

src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
```diff
@@ -5,20 +5,22 @@
              xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
              xmlns:viewModels="clr-namespace:Dynamo.ViewModels"
              xmlns:ui="clr-namespace:Dynamo.UI"
+             xmlns:controls="clr-namespace:Dynamo.Controls"
              mc:Ignorable="d"
-             d:DesignHeight="161"
-             d:DesignWidth="410"
+             Width="300"
+             Height="300"
```
QilongTang, line 11: Would this change affect default watch node size?

src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
reddyashish: We just need the BoolToVisibilityCollapsedConverter. Will remove this reference.

src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
```diff
@@ -240,8 +242,8 @@
             </Style>
         </ResourceDictionary>
     </UserControl.Resources>
-    <Grid>
 
+    <Grid Name ="inputGrid" MinHeight="100" MinWidth="100">
```
QilongTang, line 246: Would this change the current behavior?

reddyashish: No, this will only set the min height and width so that when resizing the node, the thumb icon doesn't get collapsed.

src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
reddyashish: Yes, before this change we would have a maximum height of 300 but now we don't want to put any limitation on that. 
So when we remove that, the height of the watch node would be dependent on the number of elements in the list and it would become much longer. So now we will have a default width and height same as the size of watch3d node and it can be expanded with the thumb icon.

reddyashish:
@mjkkirschner As discussed with @QilongTang, I will be opening up a new PR to address the deserialization of the watch node height and width because there seems to be some missing pieces and the team can have a separate discussion on that draft PR.

src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
mjkkirschner: hmm - I think it would be nice if the default stayed the same as it was - so that existing graphs do not change in appearance when opened.

QilongTang:
Can you check Regression:

`DynamoCoreWpfTests.PreviewBubbleTests.PreviewBubble_HiddenDummyVerticalBoundaries`

src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
reddyashish: So whats happening is, previously there was no height and width set at runtime for the watch node. Only the maximum height of 300 was set. So the height of the node would depend on the number of elements but once the max height is reached, the scroll bar appears. Now we don't have any max height set to the watch node and if no default height or width is set explicitly, then the watch node will be too long in case where input has a lot of elements. To avoid it, I thought we could make it consistent with the watch3d node.

Thomas84:
It would be nice to have the same functionality at the node preview:
image

Amoursol:
> It would be nice to have the same functionality at the node preview:
> image

Hello @Thomas84 - We have to be careful of the Node Preview as this state is often 'working' and/or left on and costs system resources, versus a Watch node is a user-driven action. We see them as two separable elements. 

Is there a use case that you would prefer to use an expandable preview bubble over an expandable watch node?

Thomas84:
> > It would be nice to have the same functionality at the node preview:
> > image
> 
> Hello @Thomas84 - We have to be careful of the Node Preview as this state is often 'working' and/or left on and costs system resources, versus a Watch node is a user-driven action. We see them as two separable elements.
> 
> Is there a use case that you would prefer to use an expandable preview bubble over an expandable watch node?

Normally I use the node preview to explore a script or when i build a new script. I use watch nodes rarely, just if I want to shift the focus of someone to a specific result or to show the result in the dynamo player. I place all my nodes in a way, that the node preview will not obstruct anything, if it will be expanded by someone. It is just time consuming to place extra watch nodes. In my opinion there should be an explorer mode, which expands all node previews at once and pin it. The height of the node preview is not the real problem, for me it is most often the width. And the best option would be, if the width adjust automatically, so I don't have to use the horizontal scroll bar or the drag handle (of the new watch node).

Amoursol:
> ..., I use Code Blocks with a variable, as description of the data. Thus if I need to, I can show the result in the node preview bubble. Especially if the script is very large it is a good method, to know what data is coming (without scrolling to another area of the script). In the last screenshot at the bottom, the width of the preview bubble can be expanded (there is a large maximum width of the preview bubble at the Code Block, thus a long variable name can expand the width (trick :P), and at the screenshot at the top of my post, the width is too small.

@Thomas84 - Gotcha! We have noted this down for future works as we are starting to look at the UX of Dynamo again in a fresh pass. For now, this task will be limited to Watch Nodes as we try to keep our improvements small, but we will file a task into our backlog for the preview bubble state as our UX designers can take a look at it.

@reddyashish please make sure to file a Jira task for the 'Preview Bubble' with a [UX] prefix and ping me.

reddyashish:
The self-service job looks good. Only the excel tests are failing. https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/764/

src/DynamoCoreWpf/Views/Preview/WatchTree.xaml.cs
```diff
@@ -25,6 +26,15 @@ void WatchTree_Loaded(object sender, RoutedEventArgs e)
             _vm = this.DataContext as WatchViewModel;
         }
 
+        internal void SetWatchNodeProperties() 
+        {
+            resizeThumb.Visibility = Visibility.Visible;
+            this.Width = 200;
+            this.Height = 200;
```
QilongTang, line 33: Do we need to respect `Configurations.MaxWatchNodeWidth` and `Configurations.MaxWatchNodeHeight` here? If not at least we should define these magic numbers as const or readonly somewhere

reddyashish: Since we have the ability to expand the watch node manually to any desired size, we cannot set max height and width for this node now. 
I updated the PR to define these magic numbers as a const.

src/Libraries/CoreNodeModelsWpf/NodeViewCustomizations/Watch.cs
```diff
@@ -40,12 +40,12 @@ public void CustomizeView(Watch nodeModel, NodeView nodeView)
 
             var watchTree = new WatchTree();
 
+            watchTree.SetWatchNodeProperties();
+
             // make empty watchViewModel
             rootWatchViewModel = new WatchViewModel(dynamoViewModel.BackgroundPreviewViewModel.AddLabelForPath);
 
             // Fix the maximum width/height of watch node.
-            nodeView.PresentationGrid.MaxWidth = Configurations.MaxWatchNodeWidth;
```
QilongTang, line 47: Should we also obsolete these two properties? Seems the only references are in this file here which will be removed. 

Or we can still set a much bigger maximum height and width, like half window size or something. This might be a UX question. FYI: @Amoursol If we choose to give user total control, then I prefer these two obsolete now and removed in Dynamo 3.0

QilongTang (approved):
LGTM with one question

src/Libraries/CoreNodeModelsWpf/NodeViewCustomizations/Watch.cs
QilongTang: I discussed this with UX-Jingyi about if we should make a max size of watch node still. Jingyi did not feel that we need. So let's obsolete `MaxWatchNodeWidth` and `MaxWatchNodeHeight` and move on.

Amoursol: @QilongTang I'm cool with what you and Jingyi came up with!

CongzhengZOU: I think the "watch node" from Grasshopper can be a very good reference. 
Imagine in Dynamo, after double click, we got a Panel, you can write number, string, or any code you want. If you don't write anything, it is a "watch node". With a default size that can display correctly for most of the time. For the special case, you can resize it very easily. Also, you can change the color of the panel.
image

PR #11355, opened 2020-12-10
Do not aggregate node execution analytics anymore

dynamo-ci-user:
Please Note:
1. Before submitting the PR, please review [How to Contribute to Dynamo](https://github.com/DynamoDS/Dynamo/blob/master/CONTRIBUTING.md)
2. Dynamo Team will meet 1x a month to review PRs found on Github (Issues will be handled separately)
3. PRs will be reviewed from oldest to newest
4. If a reviewed PR requires changes by the owner, the owner of the PR has 30 days to respond. If the PR has seen no activity by the next session, it will be either closed by the team or depending on its utility will be taken over by someone on the team
5. PRs should use either Dynamo's default PR template or [one of these other template options](https://github.com/DynamoDS/Dynamo/wiki/Choosing-a-Pull-Request-Template) in order to be considered for review.
6. PRs that do not have one of the Dynamo PR templates completely filled out with all declarations satisfied will not be reviewed by the Dynamo team.
7. PRs made to the `DynamoRevit` repo will need to be cherry-picked into all the DynamoRevit Release branches that Dynamo supports. Contributors will be responsible for cherry-picking their reviewed commits to the other branches after a `LGTM` label is added to the PR.

### Purpose

Previously, node execution analytics data was aggregated with comma, e.g.
```
Code Block, Python Script, ...
```
This makes data less straight forward to parse on backend with all kinds of combos of modified nodes during execution.

After this PR, each executed node will be tracked individually.

### Declarations

Check these if you believe they are true

- [ ] The codebase is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.
- [ ] This PR modifies some build requirements and the readme is updated

### Reviewers

(FILL ME IN) Reviewer 1  (If possible, assign the Reviewer for the PR)

(FILL ME IN, optional) Any additional notes to reviewers or testers.

### FYIs

(FILL ME IN, Optional) Names of anyone else you wish to be notified of

PR #11356, opened 2020-12-11
Do not aggregate node execution analytics anymore

QilongTang:
Please Note:

### Purpose

Example Graph
image

Previously, node execution analytics data was aggregated with comma, e.g.

```
Code Block, String from Array, Watch
```

This makes data much less straight forward on backend with all kinds of combos of modified nodes during execution.

After this PR, each executed node will be tracked individually and no longer using timed event. Here are the cases:

1. Adding a new node, `ModifiedNodes` only contains the new node
2. Deleting a node, depending on which node we are deleting, `ModifiedNodes` only contains the downstream subgraph
3. Moving a node, no graph run will be triggered
4. Undo moving a node, `ModifiedNodes` only contains the node moved and downstream subgraph. This is a bug tracked.
5. Modifying a node, `ModifiedNodes` only contains the node modified and downstream subgraph

Individual node execution event will be tracked as `DYN.NodeOperations.Run` instead of the previous aggregated `DYN.Performance.UpdateGraphAsyncTask`. See related PR: https://git.autodesk.com/Dynamo/Analytics.NET/pull/26

About performance impact, I used DotTrace by Jetbrains to profile with a file with ~760 nodes, I did file opening and some other actions above which would trigger rerun with a lot of nodes executed. The profiling results did not mark any hot spots on analytics call but marked Python node view customization as hot spots. Since the analytics call do not involve system I/O by default which matches my expectation, I think adding this loop should be safe. FYI: @mmisol 
image

### Declarations

### Reviewers

@DynamoDS/dynamo 

### FYIs

### DIFF
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -1044,18 +1044,18 @@ namespace Dynamo.Models
 
                         if (Logging.Analytics.ReportingAnalytics)
                         {
-                            var modifiedNodes = "";
                             if (updateTask.ModifiedNodes != null && updateTask.ModifiedNodes.Any())
                             {
-                                modifiedNodes = updateTask.ModifiedNodes
-                                    .Select(n => n.GetOriginalName())
-                                    .Aggregate((x, y) => string.Format("{0}, {1}", x, y));
+                                // Send analytics for each of modified nodes so they are counted individually
+                                foreach (var node in updateTask.ModifiedNodes)
+                                {
+                                    // Notice the executionTimeSpan is still for the whole delta computation
+                                    Dynamo.Logging.Analytics.TrackTimedEvent(
+                                        Categories.Performance,
+                                        e.Task.GetType().Name,
+                                        executionTimeSpan, node.GetOriginalName());
+                                }
                             }
-
-                            Dynamo.Logging.Analytics.TrackTimedEvent(
-                                Categories.Performance,
-                                e.Task.GetType().Name,
-                                executionTimeSpan, modifiedNodes);
                         }
 
                         Debug.WriteLine(String.Format(Resources.EvaluationCompleted, executionTimeSpan));
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -1049,11 +1049,12 @@ namespace Dynamo.Models
                                 // Send analytics for each of modified nodes so they are counted individually
                                 foreach (var node in updateTask.ModifiedNodes)
                                 {
-                                    // Notice the executionTimeSpan is still for the whole delta computation
-                                    Dynamo.Logging.Analytics.TrackTimedEvent(
-                                        Categories.Performance,
-                                        e.Task.GetType().Name,
-                                        executionTimeSpan, node.GetOriginalName());
+                                    // Starting from Dynamo 2.11.0, tracking node execution as generic event so
+                                    // it is distinguished with the previous aggregated performance event
+                                    Dynamo.Logging.Analytics.TrackEvent(
+                                        Actions.Run, 
+                                        Categories.NodeOperations,
+                                        node.GetOriginalName());
                                 }
                             }
                         }
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -1049,8 +1049,8 @@ namespace Dynamo.Models
                                 // Send analytics for each of modified nodes so they are counted individually
                                 foreach (var node in updateTask.ModifiedNodes)
                                 {
-                                    // Starting from Dynamo 2.11.0, tracking node execution as generic event so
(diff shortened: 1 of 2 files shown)


src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1044,18 +1044,18 @@ private void OnAsyncTaskStateChanged(DynamoScheduler sender, TaskStateChangedEve
...
-                                    .Aggregate((x, y) => string.Format("{0}, {1}", x, y));
+                                // Send analytics for each of modified nodes so they are counted individually
+                                foreach (var node in updateTask.ModifiedNodes)
+                                {
+                                    // Notice the executionTimeSpan is still for the whole delta computation
+                                    Dynamo.Logging.Analytics.TrackTimedEvent(
+                                        Categories.Performance,
+                                        e.Task.GetType().Name,
+                                        executionTimeSpan, node.GetOriginalName());
+                                }
```
QilongTang, line 1057: @mmisol @aparajit-pratap Thank you for your input. I would like to take more time to think through how we track real changes to nodes executed.

From what I can tell, adding/deleting/moving/modifying a node will all trigger execution so even a node showing as popularly executed, it might be just users modify them regularly.. Like Code Block node

QilongTang: Certain cases, undo/redo may trigger multiple runs..

QilongTang: Also @mmisol I dont think I can only look for ModifiedNodes.count = 1 case because when upstream node modified, we usually make downstream nodes dirty as well in Dynamo. e.g. with the following example, if user touches CBN, the all three nodes will be in ModifiedNodes set during execution.. Maybe the `ModifiedNodes` name itself is misleading..
image

QilongTang:
I looked at this change with @aparajit-pratap together and found it works great without Tuneup view extension, I will try to do some profiling to estimate the performance impact.

src/DynamoCore/Models/DynamoModel.cs
mjkkirschner: maybe a stupid question... doesn't this send `n` http requests for each graph run? Doesn't that seem potentially slow with large graphs?

QilongTang: @mjkkirschner ADPSDK will handle the requests with optimization on their end. From my dottrace profiling, this do not seem impact Dynamo performance that much. I have a screen shot of that in PR description using DotTrace.

aparajit-pratap (approved):

zeusongit (approved):
LGTM

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1044,18 +1044,19 @@ private void OnAsyncTaskStateChanged(DynamoScheduler sender, TaskStateChangedEve
...
+                                // Send analytics for each of modified nodes so they are counted individually
+                                foreach (var node in updateTask.ModifiedNodes)
+                                {
+                                    // Starting from Dynamo 2.11.0, tracking node execution as generic event so
+                                    // it is distinguished with the previous aggregated performance event
+                                    Dynamo.Logging.Analytics.TrackEvent(
+                                        Actions.Run, 
+                                        Categories.NodeOperations,
+                                        node.GetOriginalName());
+                                }
```
mmisol, line 1058: I see this is not a timed event anymore, which makes sense to me given we cannot separate the individual execution time for each node. Just checking if this was an intended change since it wasn't announced in the description @QilongTang

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1044,18 +1044,19 @@ private void OnAsyncTaskStateChanged(DynamoScheduler sender, TaskStateChangedEve
...
                             if (updateTask.ModifiedNodes != null && updateTask.ModifiedNodes.Any())
                             {
-                                modifiedNodes = updateTask.ModifiedNodes
-                                    .Select(n => n.GetOriginalName())
-                                    .Aggregate((x, y) => string.Format("{0}, {1}", x, y));
+                                // Send analytics for each of modified nodes so they are counted individually
+                                foreach (var node in updateTask.ModifiedNodes)
+                                {
+                                    // Starting from Dynamo 2.11.0, tracking node execution as generic event so
+                                    // it is distinguished with the previous aggregated performance event
```
mmisol, line 1053: Regarding this comment, I would abstain from including `Starting from Dynamo 2.11.0,`. Since we are using git it's easy to backtrack changes without having to add timestamps or version marks.

mmisol (commented):
Just a couple of questions on my end. Maybe my comment in #11355 was misunderstood because I didn't say I had concerns about performance.

src/DynamoCore/Models/DynamoModel.cs
QilongTang: Thanks，yes it is intended and let me update description

QilongTang:
@mmisol Thanks for review, Addressed the comments, updated PR description and merging.

PR #11358, opened 2020-12-11
Implement ToString for DS Dictionary

mmisol:
### Purpose

This is done to provide meaningful output when a dictionary is provided
to nodes like 'String from Array' and 'String from Object'. These nodes
call built-in DS functions that end up calling ToString on the
dictionary, which until now used the default implementation inherited
from Object which just returned the class name.

### Declarations


mmisol:
The format should be like this: `{A:1,B:2,C:3,D:4}`

aparajit-pratap (approved):
@mmisol thanks for fixing this!

QilongTang (approved):

PR #11359, opened 2020-12-11
Issues GitHub workflow update

Astul-Betizagasti:
### Purpose

Update the GitHub workflows related to Issues, all workflows removed have their behavior included in the new Issues_workflow.yaml file.

### Declarations

### Reviewers

Aaron Tang (@QilongTang )

### FYIs

Alfredo Pozo (@alfredo-pozo )


.github/workflows/move_issue.yaml
```diff
@@ -4,19 +4,34 @@ on:
...
           - label: 'Advance-Steel'
             repoName: 'Dynamo-Advance-Steel'
+          - label: 'Wishlist'
+            repoName: 'DynamoWishlist'
     name: Move issues to corresponding repository
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+
+      #Adds a comment to the issue before moving it
```
QilongTang, line 23: I was thinking about after moving it, can you describe or give an example how would it looks like?

QilongTang (approved):
LGTM with one comment

.github/workflows/move_issue.yaml
Astul-Betizagasti: image

This is a test I did on the repo I was testing the workflow, The creator of the issue will remain the same as in the original issue, and this will be added to the new issue thread. The message left by the bot is missing the "@" as this was a test but it is included for this pr. The author of the transfer will be the DynamoBot account created for that purpose.

QilongTang: Since you mentioned this will be added to the new issue thread, should we update the comment saying `#Adds a comment to the issue after moving it`

Astul-Betizagasti: The comment is actually added before the issue is transferred and then everything gets moved with the issue to the new repo

QilongTang: @Astul-Betizagasti OK, got you

PR #11360, opened 2020-12-14
Make IntegerSlider the only AKA for migration

mmisol:
### Purpose

There were two classes that were declaring an 'AlsoKnownAs' attribute
for 'DSCoreNodesUI.Input.IntegerSlider' in 'CoreNodeModels.Input':
'IntegerSlider' and 'IntegerSlider64Bit'. While the latter obsoletes
the former, it is not possible to migrate to it from XML, so only 
'IntegerSlider' should continue declaring an 'AlsoKnownAs' attribute.
Note that currently both classes are trying to be registered, which
results in only one of them being registered but not really making any
guarantees of which one it is. This also generates an exception catched
internally but shown in the console.

### Declarations

### DIFF
--- a/src/Libraries/CoreNodeModels/Input/IntegerSlider.cs
+++ b/src/Libraries/CoreNodeModels/Input/IntegerSlider.cs
@@ -22,7 +22,6 @@ namespace CoreNodeModels.Input
     [OutPortTypes("int")]
     [SupressImportIntoVM]
     [IsDesignScriptCompatible]
-    [AlsoKnownAs("DSCoreNodesUI.Input.IntegerSlider")]
     [IsVisibleInDynamoLibrary(false)]
     public class IntegerSlider : SliderBase<int>
     {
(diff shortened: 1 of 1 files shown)


aparajit-pratap:
@mmisol could you share a reproducible case? I suspect you see an exception when you try to open a legacy graph using the `DSCoreNodesUI.Input.IntegerSlider` node? After this fix, does that node automatically migrate to the int64 node? If so, then this probably fixes the last item that was not possible in [this PR](https://github.com/DynamoDS/Dynamo/pull/10972), which was to `Deserialize 32 bit integer slider from old XML format as new 64 bit integer slider`.

mmisol:
After discussing with @aparajit-pratap and seeing the migration from XML to `IntegerSlider64Bit` does not work, this PR has been repurposed to make `IntegerSlider` the only class defining the `AlsoKnownAs` attribute for `DSCoreNodesUI.Input.IntegerSlider`.

aparajit-pratap (approved):

QilongTang (approved):

PR #11373, opened 2020-12-18
Fix for regression in Node2Code.

aparajit-pratap:
### Purpose

Fix for [regression ](https://autodesk.slack.com/archives/C1866AFQB/p1608302596349400)in Node2Code.

n2c

### Declarations

### FYIs
@saintentropy

### DIFF
--- a/src/DynamoCore/Engine/ShortestQualifiedNameReplacer.cs
+++ b/src/DynamoCore/Engine/ShortestQualifiedNameReplacer.cs
@@ -106,7 +106,7 @@ namespace Dynamo.Engine
             // return IdentifierNode if identifier string is not separated by '.'
             if (strIdentList.Length == 1)
             {
-                return new IdentifierNode(strIdentList[0]);
+                return node;
             }
 
             // create IdentifierListNode from string such that
--- a/test/DynamoCoreTests/NodeToCodeTest.cs
+++ b/test/DynamoCoreTests/NodeToCodeTest.cs
@@ -1399,6 +1399,24 @@ namespace Dynamo.Tests
         }
 
         [Test]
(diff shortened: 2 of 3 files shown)


mjkkirschner:
@aparajit-pratap thanks for fixing this so quickly, do you think we can add a test case for `+` or a similar class of errors?

If you prefer to make this into a task let me know (or @mmisol  may have already filed one?)

src/DynamoCore/Engine/ShortestQualifiedNameReplacer.cs
```diff
@@ -106,7 +106,7 @@ public override AssociativeNode VisitIdentifierNode(IdentifierNode node)
             // return IdentifierNode if identifier string is not separated by '.'
             if (strIdentList.Length == 1)
             {
-                return new IdentifierNode(strIdentList[0]);
+                return node;
```
mjkkirschner, line 109: If my memory serves correctly, then the AST walkers are a bit inconsistent when it comes to returning new AST nodes or the same node - maybe this should return a copy of the node instead?

aparajit-pratap: The AST walkers are meant to return new nodes _only_ if they are rewritten and return the original node if there is no change.

mjkkirschner: ah! Is that documented somewhere? Would it make sense to add a note about that to the base class or something? Not that it needs to be part of this pr.

aparajit-pratap: Can do.

mmisol (approved):
Looks good to me. Thanks for fixing @aparajit-pratap !

PR #9306, opened 2018-12-11
Enable package node migration for Dynamo Core 2.0+ graphs

alfarok:
### Purpose

[QNTM-5547](https://jira.autodesk.com/browse/QNTM-5547)

This PR restores the previous (1.3 and older) behavior for package node migration for JSON graphs (2.0+).  For example, you can now migrate a packaged node that was built against Core 2.0 but was later renamed.  We previously handled 1.3 (XML) to 2.0 (JSON) but never 2.0 to 2.0+.

**Package Migration Strategies:**
ZeroTouch nodes => Use a `Migrations.XML` file located in the packages `bin` folder
NodeModel derived nodes => Use the `AlsoKnownAs` attribute on the class

**Examples:**

`MyZeroTouchLib.MyNodes.SayHello` to `MyZeroTouchLib.MyNodes.SayHelloRENAMED`
```XML
<?xml version="1.0"?>
<migrations>
  <priorNameHint>
    <oldName>MyZeroTouchLib.MyNodes.SayHello</oldName>
    <newName>MyZeroTouchLib.MyNodes.SayHelloRENAMED</newName>
  </priorNameHint>
</migrations>
```

`SampleLibraryUI.Examples.DropDownExample` to `SampleLibraryUI.Examples.DropDownExampleRENAMED`
```C#
namespace SampleLibraryUI.Examples
{
    [NodeName("Drop Down Example")]
    [NodeDescription("An example drop down node.")]
    [IsDesignScriptCompatible]
    [AlsoKnownAs("SampleLibraryUI.Examples.DropDownExample")]
    public class DropDownExampleRENAMED : DSDropDownBase
    {
        ...
    }
{
```

### Testing
1) Created a test package with several node types (not committed)
2) Created a legacy graph including these nodes (committed)
3) Renamed the classes of all included nodes
4) Included `migrations.xml` for ZeroTouch nodes
5) Added `[AlsoKnownAs]` attribute for the NodeModel node
6) Rebuilt package (committed)
7) Assert all nodes migrate to the appropriate types and no dummy nodes are found

Test Cases:
- [x] ZeroTouch single input (overload)
- [x] ZeroTouch multi input (overload)
- [x] ZeroTouch no input multi output
- [x] NodeModel dropdown

### Declarations

Check these if you believe they are true

- [x] The code base is in a better state after this PR
- [x] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [x] The level of testing this PR includes is appropriate
[...]

PR #11379, opened 2021-01-05
use existing library function to lookup previous name hints

mjkkirschner:
### Purpose

I found that this PR:
https://github.com/DynamoDS/Dynamo/pull/9306

did indeed get migrations working using the migration.xml and AKA attributes - but not consistently with xml migrations.

This PR uses an existing function in library services to first check for an exact match using the full function signature (including parameter types) - and then falls back to checking for shortNames without parameters. This is more consistent with xml migration.

I've manually tested changing the name of a ZT node in the DSOffice namespace, adding that new function(without parameters) to the DSOffice.migrations.xml and all works as expected.

### TODO:
- [x] add a test for this migration
- [x] investigate how old tests pass, ~~likely they are not precise enough.~~ The tests did not include a case of migrating a type with parameters using a migration name hint that did not include those parameters - modified the test migration xml to include this case.

### Declarations

### DIFF
--- a/src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
+++ b/src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
@@ -230,15 +230,13 @@ namespace Dynamo.Graph.Workspaces
             else if (typeof(DSFunctionBase).IsAssignableFrom(type))
             {
                 var mangledName = obj["FunctionSignature"].Value<string>();
-                var priorNames = libraryServices.GetPriorNames();
-                var functionDescriptor = libraryServices.GetFunctionDescriptor(mangledName);
-                string newName;
-
-                // Update the function descriptor if a newer migrated version of the node exists
-                if (priorNames.TryGetValue(mangledName, out newName))
+                var hintSignature = libraryServices.FunctionSignatureFromFunctionSignatureHint(mangledName);
+                var lookupSignature = mangledName;
+                if(hintSignature != null)
                 {
-                    functionDescriptor = libraryServices.GetFunctionDescriptor(newName);
+                    lookupSignature = hintSignature;
                 }
+                var functionDescriptor = libraryServices.GetFunctionDescriptor(lookupSignature);
 
                 // Use the functionDescriptor to try and restore the proper node if possible
                 if (functionDescriptor == null)
--- a/src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
+++ b/src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
@@ -230,12 +230,7 @@ namespace Dynamo.Graph.Workspaces
             else if (typeof(DSFunctionBase).IsAssignableFrom(type))
             {
                 var mangledName = obj["FunctionSignature"].Value<string>();
-                var hintSignature = libraryServices.FunctionSignatureFromFunctionSignatureHint(mangledName);
-                var lookupSignature = mangledName;
-                if(hintSignature != null)
-                {
-                    lookupSignature = hintSignature;
-                }
+                var lookupSignature = libraryServices.GetFunctionSignatureFromFunctionSignatureHint(mangledName) ?? mangledName;
                 var functionDescriptor = libraryServices.GetFunctionDescriptor(lookupSignature);
 
                 // Use the functionDescriptor to try and restore the proper node if possible
--- a/test/Engine/FFITarget/ClassWithMigratedMembers.cs
+++ b/test/Engine/FFITarget/ClassWithMigratedMembers.cs
@@ -1,0 +1,37 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace FFITarget
+{
+    public class ClassWithMigratedMembers
+    {
+
+        public static ClassWithMigratedMembers ByNothing()
+        {
+            return new ClassWithMigratedMembers();
+        }
+
+        //we test by opening a file that references a node with three params
+        public string AMemberWithTwoParams(string a, string b)
+        {
+            return a + b;
+        }
+        public string AMemberWithThreeParams(string a, string b,string c ="abc")
+        {
+            return "migrated";
+        }
+
+        //we test a file that uses the node AMemberWithAChangedName0
+        public string AMemberWithAChangedName2(string a)
+        {
+            return a;
+        }
+      /*  public string AMemberWithAChangedName1(string a)
+        {
+            return a;
+        }*/
+    }
+}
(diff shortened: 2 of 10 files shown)


src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
```diff
@@ -230,15 +230,13 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
             else if (typeof(DSFunctionBase).IsAssignableFrom(type))
             {
                 var mangledName = obj["FunctionSignature"].Value<string>();
-                var priorNames = libraryServices.GetPriorNames();
-                var functionDescriptor = libraryServices.GetFunctionDescriptor(mangledName);
-                string newName;
-
-                // Update the function descriptor if a newer migrated version of the node exists
-                if (priorNames.TryGetValue(mangledName, out newName))
+                var hintSignature = libraryServices.FunctionSignatureFromFunctionSignatureHint(mangledName);
```
aparajit-pratap, line 233: How about renaming this existing method to `GetFunctionSignatureFromFunctionSignatureHint`?

test/pkgs/Dynamo Samples/bin/SampleLibraryZeroTouch.Migrations.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0"?>
 <migrations>
   <priorNameHint>
-    <oldName>Examples.BasicExample.Create@Autodesk.DesignScript.Geometry.Point</oldName>
-    <newName>Examples.NEWBasicExample.Create@Autodesk.DesignScript.Geometry.Point</newName>
+    <oldName>Examples.BasicExample.Create</oldName>
+    <newName>Examples.NEWBasicExample.Create</newName>
```
aparajit-pratap, line 5: Should we add another test with the parameter names included just to make sure that works as well? 
I think we should also add a test to test migration to a new method with the same name but with new parameters?

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
pinzart90, line 233: could simplify code here with coalescing operator :
`
var lookupSignature = libraryServices.FunctionSignatureFromFunctionSignatureHint(mangledName) ?? mangledName;
`
Unless your way is easier to read/debug.

test/pkgs/Dynamo Samples/bin/SampleLibraryZeroTouch.Migrations.xml
mjkkirschner: Theres an existing case below this one that tests your first requested case.

I'm not sure that your second requested case will actually work - if you change the parameters - this migration functionality has no way of wiring up the new parameters so they make sense... I can definitely try it out though and add a test.

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
mjkkirschner: can do.

mjkkirschner: will update.

test/pkgs/Dynamo Samples/bin/SampleLibraryZeroTouch.Migrations.xml
aparajit-pratap: What about cases like this: https://github.com/DynamoDS/Dynamo/blob/d1195bea43e6d6f0570670b5ed85d1bf723e9427/extern/ProtoGeometry/ProtoGeometry.Migrations.xml#L77?

aparajit-pratap: And what about the case with the same namespace but different method name only?

mjkkirschner: 1. I think this magically works by reflection right? The order of the parameters is unchanged and only curve as become curve[] ?  - you are interested in a test that checks for this but with a migration that does not include parameter types in the migration right?

2. I can definitely add that.

mjkkirschner:
@aparajit-pratap I've added a new test case using FFITarget with a new class - it checks:

1. that a node with extra parameters is migrated correctly to a new method, with a new name - both methods still exist.
2. a method name is changed and the old method is missing. 
3. A node is migrated from one overload to another with the same name but a different number of parameters.

test/Engine/FFITarget/ClassWithMigratedMembers.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            return "migrated";
+        }
+
+        //we test a file that uses the node AMemberWithAChangedName0
+        public string AMemberWithAChangedName2(string a)
+        {
+            return a;
+        }
+      /*  public string AMemberWithAChangedName1(string a)
```
aparajit-pratap, line 40: Does this have to be commented out in order for the migration to work?

test/Engine/FFITarget/ClassWithMigratedMembers.cs
```diff
@@ -0,0 +1,37 @@
...
+        public string AMemberWithTwoParams(string a, string b)
+        {
+            return a + b;
+        }
+        public string AMemberWithThreeParams(string a, string b,string c ="abc")
+        {
+            return "migrated";
+        }
+
+        //we test a file that uses the node AMemberWithAChangedName0
```
aparajit-pratap, line 35: I think you mean `AMemberWithAChangedName1`?

test/Engine/FFITarget/FFITarget.Migrations.xml
```diff
@@ -0,0 +1,11 @@
+﻿<?xml version="1.0"?>
+<migrations>
+  <priorNameHint>
+    <oldName>FFITarget.ClassWithMigratedMembers.AMemberWithAChangedName1</oldName>
+    <newName>FFITarget.ClassWithMigratedMembers.AMemberWithAChangedName2</newName>
+  </priorNameHint>
+  <priorNameHint>
+    <oldName>FFITarget.ClassWithMigratedMembers.AMemberWithTwoParams@string,string</oldName>
+    <newName>FFITarget.ClassWithMigratedMembers.AMemberWithThreeParams@string,string,string</newName>
```
aparajit-pratap, line 9: Does this work if you keep the method name the same?

test/Engine/FFITarget/ClassWithMigratedMembers.cs
mjkkirschner: no, I just thought it was more explicit to illustrate what we were testing, a bit weird I admit.

test/Engine/FFITarget/ClassWithMigratedMembers.cs
mjkkirschner: yes, thanks, will change.

aparajit-pratap (approved):

PR #11383, opened 2021-01-06
DYN-3320: Add basic standard library support

sm6srw:
### Purpose

Packages should be loadable/shippable with core runtime zip and daily builds. 

This pull request does:
* introduce a new `Standard Library\Packages` folder that lives in the same folder as the `DynamoPackages.dll`
* add this folder as a folder to load packages from in each `PackageLoader` constructor.
* require that packages loaded from this location is signed.

### Declarations

### Reviewers

### FYIs

### DIFF
--- a/src/DynamoPackages/PackageLoader.cs
+++ b/src/DynamoPackages/PackageLoader.cs
@@ -76,9 +76,20 @@ namespace Dynamo.PackageManager
 
         private readonly List<string> packagesDirectoriesToVerifyCertificates = new List<string>();
 
+        // The standard library directory is located in the same directory as the DynamoPackages.dll
+        private string StandardLibraryDirectory =>
+            Path.Combine(Path.GetDirectoryName(Assembly.GetAssembly(GetType()).Location),
+                @"Standard Library", @"Packages");
+
         public PackageLoader(string overridePackageDirectory)
-            : this(new[] { overridePackageDirectory })
         {
+            if (overridePackageDirectory == null)
+                throw new ArgumentNullException("overridePackageDirectory");
+
+            packagesDirectories.Add(overridePackageDirectory);
+            this.packagesDirectories.Add(StandardLibraryDirectory);
+
+            this.packagesDirectoriesToVerifyCertificates.Add(StandardLibraryDirectory);
         }
 
         public PackageLoader(IEnumerable<string> packagesDirectories)
@@ -87,10 +98,14 @@ namespace Dynamo.PackageManager
                 throw new ArgumentNullException("packagesDirectories");
 
             this.packagesDirectories.AddRange(packagesDirectories);
+            this.packagesDirectories.Add(StandardLibraryDirectory);
+            
             var error = PathHelper.CreateFolderIfNotExist(DefaultPackagesDirectory);
 
             if (error != null)
                 Log(error);
+
+            this.packagesDirectoriesToVerifyCertificates.Add(StandardLibraryDirectory);
         }
 
         /// <summary>
@@ -99,11 +114,22 @@ namespace Dynamo.PackageManager
         /// <param name="packagesDirectories">Default package directories</param>
         /// <param name="packageDirectoriesToVerify">Default package directories where node library files require certificate verification before loading</param>
         public PackageLoader(IEnumerable<string> packagesDirectories, IEnumerable<string> packageDirectoriesToVerify)
-            : this(packagesDirectories)
         {
+            if (packagesDirectories == null)
+                throw new ArgumentNullException("packagesDirectories");
+
             if (packageDirectoriesToVerify == null)
                 throw new ArgumentNullException("packageDirectoriesToVerify");
 
+            this.packagesDirectories.AddRange(packagesDirectories);
+            this.packagesDirectories.Add(StandardLibraryDirectory);
+
+            var error = PathHelper.CreateFolderIfNotExist(DefaultPackagesDirectory);
+
+            if (error != null)
+                Log(error);
+
+            this.packagesDirectoriesToVerifyCertificates.Add(StandardLibraryDirectory);
             this.packagesDirectoriesToVerifyCertificates.AddRange(packageDirectoriesToVerify);
         }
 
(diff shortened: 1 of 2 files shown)


src/DynamoPackages/PackageLoader.cs
```diff
@@ -76,9 +76,20 @@ public string DefaultPackagesDirectory
...
+            Path.Combine(Path.GetDirectoryName(Assembly.GetAssembly(GetType()).Location),
+                @"Standard Library", @"Packages");
+
         public PackageLoader(string overridePackageDirectory)
-            : this(new[] { overridePackageDirectory })
         {
+            if (overridePackageDirectory == null)
+                throw new ArgumentNullException("overridePackageDirectory");
+
+            packagesDirectories.Add(overridePackageDirectory);
```
pinzart90, line 89: add `this` to be consistent. Or remove it everywhere..

sm6srw: What is our preference, use `this` or not using `this`?

aparajit-pratap: Not using.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -76,9 +76,20 @@ public string DefaultPackagesDirectory
 
         private readonly List<string> packagesDirectoriesToVerifyCertificates = new List<string>();
 
+        // The standard library directory is located in the same directory as the DynamoPackages.dll
+        internal string StandardLibraryDirectory =>
+            Path.Combine(Path.GetDirectoryName(Assembly.GetAssembly(GetType()).Location),
+                @"Standard Library", @"Packages");
+
         public PackageLoader(string overridePackageDirectory)
-            : this(new[] { overridePackageDirectory })
```
aparajit-pratap, line 80: Why can't we continue to call this constructor? That way the new directory needs to be added to only that constructor.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -99,11 +114,22 @@ public PackageLoader(IEnumerable<string> packagesDirectories)
         /// <param name="packagesDirectories">Default package directories</param>
         /// <param name="packageDirectoriesToVerify">Default package directories where node library files require certificate verification before loading</param>
         public PackageLoader(IEnumerable<string> packagesDirectories, IEnumerable<string> packageDirectoriesToVerify)
-            : this(packagesDirectories)
```
aparajit-pratap, line 102: Same here.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -76,9 +76,20 @@ public string DefaultPackagesDirectory
 
         private readonly List<string> packagesDirectoriesToVerifyCertificates = new List<string>();
 
+        // The standard library directory is located in the same directory as the DynamoPackages.dll
+        internal string StandardLibraryDirectory =>
```
aparajit-pratap, line 80: Should this be readonly?

src/DynamoPackages/PackageLoader.cs
sm6srw: Thanks, I totally misread this code.

src/DynamoPackages/PackageLoader.cs
sm6srw: It is a property so no.

src/DynamoPackages/PackageLoader.cs
sm6srw: Thanks

aparajit-pratap (approved):

src/DynamoPackages/PackageLoader.cs
```diff
@@ -87,10 +92,14 @@ public PackageLoader(IEnumerable<string> packagesDirectories)
...
 
             this.packagesDirectories.AddRange(packagesDirectories);
+            this.packagesDirectories.Add(StandardLibraryDirectory);
+            
             var error = PathHelper.CreateFolderIfNotExist(DefaultPackagesDirectory);
 
             if (error != null)
                 Log(error);
+
+            packagesDirectoriesToVerifyCertificates.Add(StandardLibraryDirectory);
```
mjkkirschner, line 102: so I'm curious about this decision?

pinzart90 (approved):

src/DynamoPackages/PackageLoader.cs
aparajit-pratap: @mjkkirschner do you mean packages that we own do not really need to be signed?

mjkkirschner: I'm not sure, I want to discuss it, if other teams will use this loading mechanism then it's an additional cost, also they may load custom nodes from this location - I'm not sure if our currrent cert checks actually check all files or only `.dlls` - I'm also not sure if our build steps sign all files or only `.dlls` 

Definitely we should consider the security of this loading location given that this package loading location is local to dynamo and so could be *anywhere* - not in a protected folder like %appData% or %programData% that requires user or admin permissions.

In addition, if we want to "protect" this loading location perhaps there are better ways - or additional mechanisms.

mjkkirschner: another thought - we may not want anything but ADSK packages to be loaded from here - if that was the case how would we enforce it?

aparajit-pratap: These points bring another question to my mind - will we be telling integrators that they could save their product-specific packages to this loading location if they want those packages to be loaded at startup or will we give them an option to add custom package paths in addition to this one?

mjkkirschner: I imagined we would reuse this location - but I guess ultimately that depends what type of restrictions we want to put on this location, and if those same restrictions are acceptable/sensible for both use cases. Off the top of my head it seems to make sense to use the same set of rules because they'll have the same security concerns and many overlapping requirements - and if we can simplify, why not.

mjkkirschner (approved):
approving for now, because I think discussion regarding cert checks is likely out of scope and we can schedule an internal discussion and iterate on it with future tasks.

src/DynamoPackages/PackageLoader.cs
sm6srw: Well, these are binaries that we ship and they should always be signed. That was my thinking. And dynamo binaries will always live in a protected location when an integrator use it because they will install Dynamo in some way or another. So they need to sign any binaries they add to this location regardless. 

This needs more discussions.

PR #11387, opened 2021-01-07
[DYN-3342] Close stale issues

Astul-Betizagasti:
### Purpose

Add a new workflow that will close issues that have been stale for 30 days or more and have the "need more info" label.
This workflow uses the [Stale](https://github.com/actions/stale) action that is maintained by GitHub.

### Declarations

### Reviewers

Aaron Tang (@QilongTang )

### FYIs

Alfredo Pozo (@alfredo-pozo )


.github/workflows/close_stale_issues.yaml
```diff
@@ -0,0 +1,21 @@
...
+jobs:
+  close-stale-issues:
+    runs-on: ubuntu-latest
+    steps:
+    - name: close stale issues
+      uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        #This message is requiered but never used.
+        stale-issue-message: "."
```
QilongTang, line 14: What does this mean?

Astul-Betizagasti: By default the action will leave a message when marking the issue as stale and another-one when closing the issue. As we only want the message to be added when the issue is closed the "skip-stale-issue-message" parameter was set to true but the action wont work if the "stale-issue-message" doesn't have a value.

QilongTang: @Astul-Betizagasti does empty string works in this case?

QilongTang: Thanks for the updates

QilongTang (approved):
LGTM

PR #11395, opened 2021-01-11
WIP DYN-2538 Generate single RenderPackage per Node

saintentropy:
### Purpose

Note: I was cleaning up my Dynamo git stash over the holiday break and this was too good to leave on the shelf.  This builds on some work that @LongNguyenP, @Dewb, @aparajit-pratap, and I looked at a while back.  I picked up and fixed the regressions we noted previously. 

The purpose of this PR is a performance and memory optimization for the render pipeline.  The main architectural change is modifying the assumption that we generate a single `RenderPackage` object per every individual return object for a node. Instead we aggregate all node tessellation data into a single `RenderPackage` recursively.  This improves tessellation performance as the `HelixRenderPackage` is a large .NET object with a non-trivial init time in addition to a large saving in memory; specifically one render package with all the tessellation data is much smaller then many render packages with partial tessellation data.  This approach also creates performance benefit and memory savings when the tessellated data is added to the Helix scene.  Processing one RenderPackage per `AggregateRenderPackages` call is dramatically more efficient than numerous render packages as there are numerous collection interactions and temporary allocation.  In code, this change was enabled by generating the RenderPackage first in `GetRenderPackagesFromMirrorData` before the recursive loop to fill the render package data.  Previously it was generated within the loop.

There are a few things that require changes to support features which assumed a single item = a single RenderPackage.

* Labels (ie the labels that appear in the scene context by selecting items in the node preview).  The existing implementation encoded label information in the `Description` property of the `RenderPackage`.  That encoding was not well documented making it nearly impossible to add labels to the your own tessellation method overrides.  Here I introduce a new interface, `IRenderLabels`, which provides methods on the render package to add label data and placement.  The new underlying storage implementation matches the existing `LabelPlaces` data structure in the `HelixWatch3DViewModel` so that adding labels requires much less string parsing and processing. 

* Texture maps.  The original texture map implementation in the `Colors` and `ColorStride` property assumed only one texture map per the mesh geometry in the `RenderPackage`.  To support aggregation of multiple mesh geometries within a single `RenderPackage` with unique textures, I introduce a new interface, `IRenderPackageSupplement`, which supports adding lists of Colors and ColorStrides along with their associated mesh vertex ranges.  The `Colors` and `ColorStride` property as well as the `SetColors` method are marked as obsolete with notes to use the new methods in the `IRenderPackageSupplement`.

* Color adding methods on `IRenderPackage`.  While methods to add vertex color like  `AddPointVertexColor` still work fine with this change, when utilized in tessellation calls, methods like `ApplyPointVertexColors` overwrite the previous data vs appending.  That can cause data loss as they are currently used.  I introduce new methods to `IRenderPackageSupplement` like `InsertPointVertexColorRange` and `AppendPointVertexColorRange` to allow additions without overwriting existing data.  The old methods are also marked as obsolete with notes to use the new methods in the `IRenderPackageSupplement`.  The GeometryColor nodes are updated to use the new methods.

These new features are being introduced with a minor release and not Dynamo 3.0.  While we can update the tessellation implementation in Dynamo and Dynamo integrations, Third Party tessellation functions also need to stay working.  For each of the Obsolete methods above I introduced an Exception to allow the `GetRenderPackagesFromMirrorData` implementation to detect their usage.  Specifically within `IRenderPackageSupplement` there is a boolean `AllowLegacyColorOperations` that can be set to false which will cause the `LegacyRenderPackageMethodException` to be thrown when the obsolete methods are called.  In that case, we roll back any changes to the `RenderPackage` that occurred within the `Tesselate` call and generate a new `RenderPackage` with `AllowLegacyColorOperations` set to true and re-Tesselate.  This individual render package is added to the RenderPackageCache and the outer loop resumes.  In the transition to Dynamo 3.0, the obsolete methods can be removed as well as this fallback.

Todo
* Add tests for the new functionality
* Update Tessellate implementation in DynamoRevit, other integrations, samples, or internal packages.  This is not strictly required due to the fallback mechanism. 
* Fix the Label.ByPointAndString node to work with the new label mechanism so that the node actually work.
* ITransformable needs to be refactored similarly to the Texture Map implementation.  This may be covered under a separate PR as we can possibly take advantage of the refactor to introduce instancing mechanism. 

Examples

https://user-images.githubusercontent.com/2145751/104148265-98006400-539f-11eb-980e-595a01f6b856.mp4

https://user-images.githubusercontent.com/2145751/104148270-9df64500-539f-11eb-9c8f-20ed9d0c27e5.mp4

https://user-images.githubusercontent.com/2145751/104148276-a2baf900-539f-11eb-9edf-a803ab32acc3.mp4

https://user-images.githubusercontent.com/2145751/104148280-a64e8000-539f-11eb-997d-f8235882d69a.mp4

For most point and curve tessellation the speed and memory improvement will be linear based on the number of tessellated items.  However for meshes the speed and memory improvement is dramatic. 

### Declarations

### Reviewers

### DIFF
--- a/src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
+++ b/src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
@@ -149,6 +149,27 @@ namespace Dynamo.Scheduler
                 return;
             }
 
+            var package = factory.CreateRenderPackage();
+            package.DisplayLabels = displayLabels;
+            package.Description = tag;
+            package.IsSelected = isNodeSelected;
+
+            GetRenderPackagesFromMirrorDataImp(mirrorData, package, tag);
+
+            if (package.MeshVertexColors.Any())
+            {
+                package.RequiresPerVertexColoration = true;
+            }
+
+            renderPackageCache.Add(package, outputPortId);
+        }
+
+        private void GetRenderPackagesFromMirrorDataImp(
+            MirrorData mirrorData,
+            IRenderPackage package,
+            string labelKey)
+        {
+
             if (mirrorData.IsCollection)
             {
                 int count = 0;
@@ -156,69 +177,29 @@ namespace Dynamo.Scheduler
                 {
                     if (el.IsCollection || el.Data is IGraphicItem)
                     {
-                        string newTag = tag + ":" + count;
-                        GetRenderPackagesFromMirrorData(outputPortId, el, newTag, displayLabels);
+                        string newLabel = labelKey + ":" + count;
+                        GetRenderPackagesFromMirrorDataImp(el, package, newLabel);
                     }
-                    count = count + 1;
+                    count += 1;
                 }
             }
             else
             {
-                var graphicItem = mirrorData.Data as IGraphicItem;
-                if (graphicItem == null)
+                if (!(mirrorData.Data is IGraphicItem graphicItem))
                 {
                     return;
                 }
 
-                var package = factory.CreateRenderPackage();
                 var packageWithTransform = package as ITransformable;
-                package.Description = tag;
 
                 try
                 {
-                    graphicItem.Tessellate(package, factory.TessellationParameters);
-                    if (package.MeshVertexColors.Any())
-                    {
-                        package.RequiresPerVertexColoration = true;
-                    }
+                    var previousPointVertexCount = package.PointVertexCount;
+                    var previousLineVertexCount = package.LineVertexCount;
+                    var previousMeshVertexCount = package.MeshVertexCount;
 
-                    //If the package has a transform that is not the identity matrix
-                    //then set requiresCustomTransform to true.
-                    if (packageWithTransform != null && packageWithTransform.Transform.SequenceEqual(
-                        new double[] { 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1 }) == false)
-                    {
-                        (packageWithTransform).RequiresCustomTransform = true;
-                    }
-
-                    if (factory.TessellationParameters.ShowEdges)
-                    {
-                        var topology = graphicItem as Topology;
-                        if (topology != null)
-                        {
-                            var surf = graphicItem as Surface;
-                            if (surf != null)
-                            {
-                                foreach (var curve in surf.PerimeterCurves())
-                                {
-                                    curve.Tessellate(package, factory.TessellationParameters);
-                                    curve.Dispose();
-                                }
-                            }
-                            else
-                            {
-                                var edges = topology.Edges;
-                                foreach (var geom in edges.Select(edge => edge.CurveGeometry))
-                                {
-                                    geom.Tessellate(package, factory.TessellationParameters);
-                                    geom.Dispose();
-                                }
-                                edges.ForEach(x => x.Dispose());
-                            }
-                        }
-                    }
-                    
-                    var plane = graphicItem as Plane;
-                    if (plane != null)
+                    //Plane tessellation needs to be handled here vs in LibG currently
+                    if (graphicItem is Plane plane)
                     {
                         package.RequiresPerVertexColoration = true;
 
@@ -229,9 +210,7 @@ namespace Dynamo.Scheduler
                         var b = Point.ByCartesianCoordinates(cs, -s, s, 0);
                         var c = Point.ByCartesianCoordinates(cs, -s, -s, 0);
                         var d = Point.ByCartesianCoordinates(cs, s, -s, 0);
-
-                        // Get rid of the original plane geometry.
-                        package.Clear();
+                        //Todo Dispose cs, a, b, c, d?
 
                         package.AddTriangleVertex(a.X, a.Y, a.Z);
                         package.AddTriangleVertex(b.X, b.Y, b.Z);
@@ -263,67 +242,81 @@ namespace Dynamo.Scheduler
                         var nEnd = plane.Origin.Add(plane.Normal.Scale(2.5));
                         package.AddLineStripVertex(nEnd.X, nEnd.Y, nEnd.Z);
 
-                        for (var i = 0; i < package.LineVertexCount / 2; i++)
+                        for (var i = 0; i < 5; i++)
                         {
                             package.AddLineStripVertexCount(2);
-                        }
-
-                        for (var i = 0; i < package.LineVertexCount; i ++)
-                        {
+                            package.AddLineStripVertexColor(MidTone, MidTone, MidTone, 255);
                             package.AddLineStripVertexColor(MidTone, MidTone, MidTone, 255);
                         }
 
-                        for (var i = 0; i < package.MeshVertexCount; i++)
+                        for (var i = 0; i < 6; i++)
                         {
                             package.AddTriangleVertexNormal(plane.Normal.X, plane.Normal.Y, plane.Normal.Z);
-                        }
-
-                        for (var i = 0; i < package.MeshVertexCount; i++)
-                        {
                             package.AddTriangleVertexColor(0, 0, 0, 10);
                         }
                     }
-
-                    // The default color coming from the geometry library for
-                    // curves is 255,255,255,255 (White). Because we want a default
-                    // color of 0,0,0,255 (Black), we adjust the color components here.
-                    if (graphicItem is Curve || graphicItem is Surface || graphicItem is Solid || graphicItem is Point)
+                    else
                     {
-                        if (package.LineVertexCount > 0 && !package.LineStripVertexColors.Any())
-                        {
-                            package.ApplyLineVertexColors(CreateColorByteArrayOfSize(package.LineVertexCount, DefR, DefG, DefB, DefA));
-                        }
+                        graphicItem.Tessellate(package, factory.TessellationParameters);
+                    }
 
-                        if (package.PointVertexCount > 0 && !package.PointVertexColors.Any())
+                    //If the package has a transform that is not the identity matrix
+                    //then set requiresCustomTransform to true.
+                    if (packageWithTransform != null && !packageWithTransform.RequiresCustomTransform && packageWithTransform.Transform.SequenceEqual(
+                        new double[] { 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1 }) == false)
+                    {
+                        (packageWithTransform).RequiresCustomTransform = true;
+                    }
+
+                    if (factory.TessellationParameters.ShowEdges)
+                    {
+                        if (graphicItem is Topology topology)
                         {
-                            package.ApplyPointVertexColors(CreateColorByteArrayOfSize(package.PointVertexCount, DefR, DefG, DefB, DefA));
+                            if (graphicItem is Surface surf)
+                            {
+                                foreach (var curve in surf.PerimeterCurves())
+                                {
+                                    curve.Tessellate(package, factory.TessellationParameters);
+                                    curve.Dispose();
+                                }
+                            }
+                            else
+                            {
+                                var edges = topology.Edges;
+                                foreach (var geom in edges.Select(edge => edge.CurveGeometry))
+                                {
+                                    geom.Tessellate(package, factory.TessellationParameters);
+                                    geom.Dispose();
+                                }
+                                edges.ForEach(x => x.Dispose());
+                            }
                         }
                     }
+
+                    if (!package.DisplayLabels && package is IRenderLabels packageLabels)
+                    {
+                        if (package.MeshVertexCount > previousMeshVertexCount)
+                        {
+                            packageLabels.AddLabel(labelKey,VertexType.Mesh, package.MeshVertexCount);
+                        }
+                        else if (package.LineVertexCount > previousLineVertexCount)
+                        {
+                            packageLabels.AddLabel(labelKey, VertexType.Line, package.LineVertexCount);
+                        }
+                        else if (package.PointVertexCount > previousPointVertexCount)
+                        {
+                            packageLabels.AddLabel(labelKey, VertexType.Point, package.PointVertexCount);
+                        }
+                    }
+                    
+                    //Todo Do we need to validate that vertex color counts match vertex counts?
                 }
                 catch (Exception e)
                 {
                     Debug.WriteLine(
                         "PushGraphicItemIntoPackage: " + e);
                 }
-
-                package.DisplayLabels = displayLabels;
-                package.IsSelected = isNodeSelected;
-
-                renderPackageCache.Add(package, outputPortId);
             }
-        }
-
-        private static byte[] CreateColorByteArrayOfSize(int size, byte red, byte green, byte blue, byte alpha)
-        {
-            var arr = new byte[size * 4];
-            for (var i = 0; i < arr.Count(); i += 4)
-            {
-                arr[i] = red;
-                arr[i + 1] = green;
-                arr[i + 2] = blue;
-                arr[i + 3] = alpha;
-            }
-            return arr;
         }
 
         protected override void HandleTaskCompletionCore()
--- a/src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
+++ b/src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
@@ -154,7 +154,16 @@ namespace Dynamo.Scheduler
             package.Description = tag;
             package.IsSelected = isNodeSelected;
 
-            GetRenderPackagesFromMirrorDataImp(mirrorData, package, tag);
+            //Initialize the package here assuming we will only generate a single renderPackage for this node
+            //We set the AllowLegacyColorOperations so that we can catch tessellation implementations which 
+            //use the deprecated calls.  At that point we will roll back the changes to the renderPackage
+            //and call tessellate with a new renderPackage object with AllowLegacyColorOperations set to true.
+            if (package is IRenderPackageSupplement packageSupplement)
+            {
+                packageSupplement.AllowLegacyColorOperations = false;
+            }
+
+            GetRenderPackagesFromMirrorDataImp(outputPortId, mirrorData, package, tag);
 
             if (package.MeshVertexColors.Any())
             {
@@ -165,6 +174,7 @@ namespace Dynamo.Scheduler
         }
 
         private void GetRenderPackagesFromMirrorDataImp(
+            Guid outputPortId,
             MirrorData mirrorData,
             IRenderPackage package,
             string labelKey)
@@ -178,7 +188,7 @@ namespace Dynamo.Scheduler
                     if (el.IsCollection || el.Data is IGraphicItem)
                     {
                         string newLabel = labelKey + ":" + count;
-                        GetRenderPackagesFromMirrorDataImp(el, package, newLabel);
+                        GetRenderPackagesFromMirrorDataImp(outputPortId, el, package, newLabel);
                     }
                     count += 1;
                 }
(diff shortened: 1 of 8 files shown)


mjkkirschner:
@aparajit-pratap one downside of marking this PR as a draft is that the tests do not run - we'd have to modify the code that runs on the self serve pull job - and I think it makes use of https://github.com/microsoft/PowerShellForGitHub ... I gave up once I got that far.

mjkkirschner:
@saintentropy it would be good to run the wpf visualization test jobs - there are versions of that job that run with warp and another that runs on a real GPU.

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
```diff
@@ -212,104 +280,173 @@ protected override void HandleTaskExecutionCore()
...
+            package.AddLineStripVertex(d.X, d.Y, d.Z);
+            package.AddLineStripVertex(d.X, d.Y, d.Z);
+            package.AddLineStripVertex(a.X, a.Y, a.Z);
+
+            // Draw normal
+            package.AddLineStripVertex(plane.Origin.X, plane.Origin.Y, plane.Origin.Z);
+            var nEnd = plane.Origin.Add(plane.Normal.Scale(2.5));
+            package.AddLineStripVertex(nEnd.X, nEnd.Y, nEnd.Z);
+
+            for (var i = 0; i < 5; i++)
```
aparajit-pratap, line 366: An explanation of why this is 5 would help.

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
```diff
@@ -212,104 +280,173 @@ protected override void HandleTaskExecutionCore()
...
+                package.AddLineStripVertexCount(2);
+                package.AddLineStripVertexColor(MidTone, MidTone, MidTone, 255);
+                package.AddLineStripVertexColor(MidTone, MidTone, MidTone, 255);
+            }
 
-                        for (var i = 0; i < package.MeshVertexCount; i++)
-                        {
-                            package.AddTriangleVertexNormal(plane.Normal.X, plane.Normal.Y, plane.Normal.Z);
-                        }
+            for (var i = 0; i < 6; i++)
```
aparajit-pratap, line 373: An explanation of why this is 6?

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
```diff
@@ -149,54 +149,122 @@ protected override void HandleTaskExecutionCore()
...
+                    try
                     {
-                        package.RequiresPerVertexColoration = true;
+                        graphicItem.Tessellate(package, factory.TessellationParameters);
                     }
-
-                    //If the package has a transform that is not the identity matrix
-                    //then set requiresCustomTransform to true.
-                    if (packageWithTransform != null && packageWithTransform.Transform.SequenceEqual(
-                        new double[] { 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1 }) == false)
```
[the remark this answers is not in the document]
saintentropy: I think that case is valid.... I just don't think that is what our implementation supports currently.  Today you add your tessellated geometry and then you can additionally add a transform.  That transform does not support instancing it only transforms all the geometry in the render package.  I think the original use case being able to tesselate an object once and then move it around.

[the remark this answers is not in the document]
saintentropy: I think instancing would be much more interesting and we should add that as an enhancement.  I think we should do it as a follow up PR.  It would need a similar data structure to the color aspects where you can apply numerous transforms to any range of geometry data.  The advantage would be that we could and tessellated geometry data once and supply a transform matrix for each instance.  There would also need to changes to the helix implementation to support real instancing

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
```diff
@@ -149,54 +149,122 @@ protected override void HandleTaskExecutionCore()
...
+                    catch (LegacyRenderPackageMethodException)
                     {
-                        (packageWithTransform).RequiresCustomTransform = true;
+                        //At this point we have detected an implementation of Tessellate which is using legacy color methods
```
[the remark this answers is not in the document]
saintentropy: We do not use the API's that throw `LegacyRenderPackageMethodException` in LibG so no changes are required.  Only some of the Color methods fall in this category.  `void ApplyPointVertexColors(byte[] colors)` for example as it replaces all the vertex colors in the render packages.  This would overwrite any color data that had previously been added.  `void AddPointVertexColor(byte red, byte green, byte blue, byte alpha)` is ok as it just continues to append data.

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
saintentropy: Yep that would be good.  5 lines

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
saintentropy: Yes I can... 6 mesh vertex

saintentropy: Also we can move this into LibG too.  Not sure why it lives here.

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
```diff
@@ -212,104 +280,173 @@ protected override void HandleTaskExecutionCore()
...
+                if (count > 0)
+                {
+                    packageSupplement.AppendMeshVertexColorRange(CreateColorByteArrayOfSize(count, DefR, DefG, DefB, DefA));
                 }
-                catch (Exception e)
+            }
+        }
+
+        private static void LegacyPackageEnsureColorExistsPerVertex(IRenderPackage pkg, int previousPointVertexCount, int previousLineVertexCount, int previousMeshVertexCount)
+        {
```
aparajit-pratap, line 421:
```
var colorCount = pkg.PointVertexColors.Count() / 4;
if (pkg.PointVertexCount > previousPointVertexCount
                && colorCount > pkg.PointVertexCount) 
{
    for (var i = colorCount; i < pkg.PointVertexCount; i++)
    { ...
```

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
```diff
@@ -212,104 +280,173 @@ protected override void HandleTaskExecutionCore()
...
+                LegacyPackageEnsureColorExistsPerVertex(package, previousPointVertexCount, previousLineVertexCount, previousMeshVertexCount);
+                return;
+            }
+            
+            if (package.PointVertexCount > previousPointVertexCount)
+            {
+                var count = package.PointVertexCount - packageSupplement.PointVertexColorCount;
+                if (count > 0)
+                {
+                    packageSupplement.AppendPointVertexColorRange(CreateColorByteArrayOfSize(count, DefR, DefG, DefB, DefA));
```
aparajit-pratap, line 396: How is this different from `AddPointVertexColor` called in a for loop?

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
```diff
@@ -212,104 +280,173 @@ protected override void HandleTaskExecutionCore()
...
+                var count = package.MeshVertexCount - packageSupplement.MeshVertexColorCount;
+                if (count > 0)
+                {
+                    packageSupplement.AppendMeshVertexColorRange(CreateColorByteArrayOfSize(count, DefR, DefG, DefB, DefA));
                 }
-                catch (Exception e)
+            }
+        }
+
+        private static void LegacyPackageEnsureColorExistsPerVertex(IRenderPackage pkg, int previousPointVertexCount, int previousLineVertexCount, int previousMeshVertexCount)
```
aparajit-pratap, line 420: So this method essentially does the same thing as `EnsureColorExistsPerVertex` but with the old APIs? I suppose this check is needed only now after the change to a single render package per node? If so, when will there be a case where the package will not be an `IRenderPackageSupplement` but there would still be a single package per node?

aparajit-pratap: Is this method needed for custom third party `IRenderPackage `types that do not implement `IRenderPackageSupplement` but will still be running Dynamo after these changes?

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
```diff
@@ -212,104 +280,173 @@ protected override void HandleTaskExecutionCore()
...
                 }
+            }
 
-                package.DisplayLabels = displayLabels;
-                package.IsSelected = isNodeSelected;
+            if (pkg.LineVertexCount > previousLineVertexCount
+                && pkg.LineStripVertexColors.Any()
+                && pkg.LineStripVertexColors.Count() / 4 > pkg.LineVertexCount)
+            {
+                for (var i = pkg.LineStripVertexColors.Count() / 4; i < pkg.LineVertexCount; i++)
```
aparajit-pratap, line 436: Similar refactor as above will be useful.

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
```diff
@@ -212,104 +280,173 @@ protected override void HandleTaskExecutionCore()
...
+                    pkg.AddLineStripVertexColor(DefR, DefG, DefB, DefA);
+                }
+            }
 
-                renderPackageCache.Add(package, outputPortId);
+            if (pkg.MeshVertexCount > previousMeshVertexCount
+                && pkg.MeshVertexColors.Any()
+                && pkg.MeshVertexColors.Count() / 4 > pkg.MeshVertexCount)
+            {
+                for (var i = pkg.MeshVertexColors.Count() / 4; i < pkg.MeshVertexCount; i++)
```
aparajit-pratap, line 446: Similar refactor as above will be useful.

src/DynamoCoreWpf/Rendering/HelixRenderPackage.cs
```diff
@@ -555,12 +521,281 @@ public IEnumerable<int> PointIndices
...
+        }
+
+        #endregion
+
+        #region IRenderPackageSupplement implementation
+
+        /// <summary>
+        /// The number of point vertices colors in the package.
+        /// </summary>
+        public int PointVertexColorCount => points.Colors.Count;
```
aparajit-pratap, line 620: Why do these implementations differ from those in `DefaultRenderPackage`?

src/VisualizationTests/HelixWatch3DViewModelTests.cs
```diff
@@ -1189,7 +1189,7 @@ private Watch3DView FindFirstWatch3DNodeView()
             return views.Last();
         }
 
-        [Test]
+        [Test, Category("Failure")]
```
aparajit-pratap, line 1192: Why is this failing now?

src/Libraries/GeometryColor/GeometryColor.cs
```diff
@@ -241,7 +241,9 @@ private static Color[][] ComputeColorMap(Surface surface, IEnumerable<UV> uvs, C
         private void CreateColorMapOnSurface(Color[][] colorMap , IRenderPackage package, TessellationParameters parameters)
```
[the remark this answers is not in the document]
saintentropy: 👍

src/NodeServices/GraphicsDataInterfaces.cs
```diff
@@ -196,6 +202,124 @@ public interface IRenderPackage
...
+
+        /// <summary>
+        /// Add a label position to the render package.
+        /// </summary>
+        void AddLabel(string label, double x, double y, double z);
+
+        /// <summary>
+        /// A flag indicating whether the render package should auto generate labels
+        /// </summary>
+        bool AutoGenerateLabels { get; set; }
```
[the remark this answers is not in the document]
saintentropy: This is the classic replication index label that is used to indicate the position of the item when you select it in the preview list.

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
```diff
@@ -212,104 +280,173 @@ protected override void HandleTaskExecutionCore()
...
+            if (packageSupplement == null)
+            {
+                LegacyPackageEnsureColorExistsPerVertex(package, previousPointVertexCount, previousLineVertexCount, previousMeshVertexCount);
+                return;
+            }
+            
+            if (package.PointVertexCount > previousPointVertexCount)
+            {
+                var count = package.PointVertexCount - packageSupplement.PointVertexColorCount;
+                if (count > 0)
```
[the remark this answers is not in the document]
saintentropy: The count in 391 is to check if there has been a change to the geometry data in the render package when the tessellation method was called.  It could be that this specific tessellation object did not have any point vertex data and then no action is needed.  The second check on line 393 is verify that within the package the color and point vertex information is synchronized.  If the the tessellation method added point vertex data but did not add color, this method would add default colors for the new geometry data.

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
```diff
@@ -212,104 +280,173 @@ protected override void HandleTaskExecutionCore()
...
-                    if (graphicItem is Curve || graphicItem is Surface || graphicItem is Solid || graphicItem is Point)
-                    {
-                        if (package.LineVertexCount > 0 && !package.LineStripVertexColors.Any())
-                        {
-                            package.ApplyLineVertexColors(CreateColorByteArrayOfSize(package.LineVertexCount, DefR, DefG, DefB, DefA));
-                        }
+            if (package.LineVertexCount > previousLineVertexCount)
+            {
+                var count = package.LineVertexCount - packageSupplement.LineVertexColorCount;
+                if (count > 0)
```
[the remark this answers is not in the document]
saintentropy: same situation as point vertex.  Default colors have to be added if tessellation does not add color info.   This was actually also true in before the refactor.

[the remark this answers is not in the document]
saintentropy: Current implementation is here -> https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs#L290

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
saintentropy: Yes this is for third party implementation of `IRenderPackage`.  Our Helix and Default implementations in Dynamo  would add `IRenderPackageSuppliment`.  The color check is done regardless of render package per `IGraphicItem` or per node.  This was previously handled here -> https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs#L290.  This just ensures color info is added to the RenderPackage before being passed to the Helix View

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
saintentropy: This would be more clear...  Since this is on the hot path and the `PointVertexColors` is and IEnumerable<byte> I was a bit allergic to calling `Count()` unless I really had to.  Let me do some more reading on the performance impact

src/VisualizationTests/HelixWatch3DViewModelTests.cs
saintentropy: The two meshes do not exists any more within the `Element3DDictionary`.   Previously this RenderPackageCache would have added `var_88f09e6c057f4a5c95a7f4f0b25161e2:0:0:0":mesh` and `var_88f09e6c057f4a5c95a7f4f0b25161e2:0:0:1:mesh`.  Now as far as the Helix view is concerned there is only one mesh object related to this node identifier in the `Element3DDictionary`.  That is `var_88f09e6c057f4a5c95a7f4f0b25161e2:mesh` I think.  The indexes are no longer added here as this new implementation only adds one bucket for mesh, curve, and vertex data to the Helix dictionary per node.  In fact that is why this is so much faster.  1 point bucket vs potentially 100K is much much faster.

saintentropy: And much more memory efficient.

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
aparajit-pratap: My suggestion is to reuse the count instead of calling `Count()` on the innumerable in multiple places.

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
```diff
@@ -149,54 +149,122 @@ protected override void HandleTaskExecutionCore()
...
-                    //If the package has a transform that is not the identity matrix
-                    //then set requiresCustomTransform to true.
-                    if (packageWithTransform != null && packageWithTransform.Transform.SequenceEqual(
-                        new double[] { 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1 }) == false)
+                    catch (LegacyRenderPackageMethodException)
                     {
-                        (packageWithTransform).RequiresCustomTransform = true;
+                        //At this point we have detected an implementation of Tessellate which is using legacy color methods
+                        //We roll back the primary renderPackage to it previous state before calling tessellation again.
+                        package = RollBackPackage(package, previousPointVertexCount, previousLineVertexCount,
```
aparajit-pratap, line 235: I still don't understand why this needs to be rolled back if it's going to be discarded anyway when we return in line 254. I don't see `package `is being used anywhere beyond this point inside this condition?

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
```diff
@@ -212,104 +280,173 @@ protected override void HandleTaskExecutionCore()
...
+            package.RequiresPerVertexColoration = true;
+
+            var s = 2.5;
+
+            var cs = CoordinateSystem.ByPlane(plane);
+            var a = Point.ByCartesianCoordinates(cs, s, s, 0);
+            var b = Point.ByCartesianCoordinates(cs, -s, s, 0);
+            var c = Point.ByCartesianCoordinates(cs, -s, -s, 0);
+            var d = Point.ByCartesianCoordinates(cs, s, -s, 0);
+            //Todo Dispose cs, a, b, c, d?
```
aparajit-pratap, line 334: Can we add these Dispose statements in this PR while we are at it?

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
```diff
@@ -212,104 +280,173 @@ protected override void HandleTaskExecutionCore()
...
                 }
-                catch (Exception e)
+            }
+        }
+
+        private static void LegacyPackageEnsureColorExistsPerVertex(IRenderPackage pkg, int previousPointVertexCount, int previousLineVertexCount, int previousMeshVertexCount)
+        {
+            if (pkg.PointVertexCount > previousPointVertexCount
+                && pkg.PointVertexColors.Any()
+                && pkg.PointVertexColors.Count() / 4 > pkg.PointVertexCount)
```
aparajit-pratap, line 424: Shouldn't this check be `pkg.PointVertexCount > pkg.PointVertexColors.Count() / 4`

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
```diff
@@ -212,104 +280,173 @@ protected override void HandleTaskExecutionCore()
...
-                        "PushGraphicItemIntoPackage: " + e);
+                    pkg.AddPointVertexColor(DefR, DefG, DefB, DefA);
                 }
+            }
 
-                package.DisplayLabels = displayLabels;
-                package.IsSelected = isNodeSelected;
+            if (pkg.LineVertexCount > previousLineVertexCount
+                && pkg.LineStripVertexColors.Any()
+                && pkg.LineStripVertexColors.Count() / 4 > pkg.LineVertexCount)
```
aparajit-pratap, line 434: Shouldn't this check be` pkg.LineVertexCount > pkg.LineVertexColors.Count() / 4`?

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
```diff
@@ -212,104 +280,173 @@ protected override void HandleTaskExecutionCore()
...
+                for (var i = pkg.LineStripVertexColors.Count() / 4; i < pkg.LineVertexCount; i++)
+                {
+                    pkg.AddLineStripVertexColor(DefR, DefG, DefB, DefA);
+                }
+            }
 
-                renderPackageCache.Add(package, outputPortId);
+            if (pkg.MeshVertexCount > previousMeshVertexCount
+                && pkg.MeshVertexColors.Any()
+                && pkg.MeshVertexColors.Count() / 4 > pkg.MeshVertexCount)
```
aparajit-pratap, line 444: Shouldn't this check be `pkg.MeshVertexCount > pkg.MeshVertexColors.Count() / 4`?

src/DynamoCoreWpf/Rendering/HelixRenderPackage.cs
```diff
@@ -555,12 +521,281 @@ public IEnumerable<int> PointIndices
...
+        /// </summary>
+        public void InsertPointVertexColorRange(int startIndex, int endIndex, byte red, byte green, byte blue, byte alpha)
+        {
+            var message = string.Empty;
+            if (!HasValidStartEnd(startIndex, endIndex, points, out message))
+            {
+                if (!string.IsNullOrEmpty(message))
+                {
+                    throw new Exception(message);
+                }
```
aparajit-pratap, line 643: Why throw an exception only if there is a message? In any case I think the message existing and `HasValidStartEnd` returning false go hand-in-hand.

src/VisualizationTests/HelixWatch3DViewModelTests.cs
aparajit-pratap: Why not remove this test in that case?

saintentropy: That is good.  I just marked them as fail initially as to clearly communicate behavioral changes

src/DynamoCoreWpf/Rendering/HelixRenderPackage.cs
```diff
@@ -555,12 +521,281 @@ public IEnumerable<int> PointIndices
...
+        }
+
+        #endregion
+
+        #region IRenderLabels implementation
+
+        ///// <summary>
+        ///// Get label data; label string and associated position
+        ///// </summary>
+        public List<Tuple<string, float[]>> LabelData
```
aparajit-pratap, line 551: Can we just have one property - either `LabelPlaces` or `LabelData` instead of having both? I think they are essentially the same data, just formatted differently, I'm I'm not wrong?

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
```diff
@@ -149,54 +149,122 @@ protected override void HandleTaskExecutionCore()
...
                 }
 
-                var package = factory.CreateRenderPackage();
-                var packageWithTransform = package as ITransformable;
-                package.Description = tag;
+                GetRenderPackageFromGraphicItem(outputPortId, graphicItem, package, labelKey);
+            }
+        }
+
+        private void GetRenderPackageFromGraphicItem(Guid outputPortId, IGraphicItem graphicItem, IRenderPackage package, string labelKey)
```
aparajit-pratap, line 210: Are you going to rename this to something more appropriate like you discussed earlier?

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
saintentropy: functionally there isn't a huge difference... under the hood it uses `AddRange`.  I think it just is clearer code.

src/DynamoCoreWpf/Rendering/HelixRenderPackage.cs
saintentropy: Because the helix render package is highly optimized to store data in Helix data structure.  Color, Vertex, etc...   This makes for a faster transfer to the View

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
```diff
@@ -149,54 +149,121 @@ protected override void HandleTaskExecutionCore()
...
                 }
 
-                var package = factory.CreateRenderPackage();
-                var packageWithTransform = package as ITransformable;
-                package.Description = tag;
+                GetTessellationDataFromGraphicItem(outputPortId, graphicItem, package, labelKey);
+            }
+        }
+
+        private void GetTessellationDataFromGraphicItem(Guid outputPortId, IGraphicItem graphicItem, IRenderPackage package, string labelKey)
```
aparajit-pratap, line 210: I think `package `should be passed as a `ref` parameter as it can be assigned a new object?

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
```diff
@@ -364,13 +363,20 @@ private static void CreatePlaneTessellation(IRenderPackage package, Plane plane)
...
                 package.AddLineStripVertexColor(MidTone, MidTone, MidTone, 255);
                 package.AddLineStripVertexColor(MidTone, MidTone, MidTone, 255);
             }
+
+            //dispose helper geometry
+            a.Dispose();
+            b.Dispose();
+            c.Dispose();
+            d.Dispose();
+            cs.Dispose();
```
aparajit-pratap, line 379: How about using `using` statements instead?

aparajit-pratap:
@saintentropy thanks for addressing the review comments and adding inline documentation. Also thanks for clearer names for the new APIs. I have just two remaining questions, then LGTM.

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
saintentropy: I tried this and it was more convoluted then the dispose.  I think just keep it simple here.

src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs
saintentropy: Definitely!  Good catch

saintentropy:
Hi @aparajit-pratap Added some of fixes from the comments and a lot more tests

saintentropy:
I also added the single item label fix.  The regression was introduced in this PR -> https://github.com/DynamoDS/Dynamo/pull/10989.  That PR also added the non-functional DisplayLabel node back into core so this PR fixes both the node and the label regression.

src/VisualizationTests/HelixRenderPackageTests.cs
```diff
@@ -157,6 +158,398 @@ public void HelixRenderPackage_MeshColorBufferCheck()
...
+            var testColors = WhiteByteArrayOfLength(5);
+
+            try
+            {
+                p.AppendMeshVertexColorRange(testColors);
+            }
+            catch(Exception e)
+            {
+                Assert.AreEqual(e.GetType(), typeof(Exception));
+            }
```
aparajit-pratap, line 199: `Assert.ThrowsException<Exception>(() => p.AppendMeshVertexColorRange(testColors));`

aparajit-pratap: How are these exceptions dealt with in Dynamo in a real situation?

src/VisualizationTests/HelixRenderPackageTests.cs
```diff
@@ -157,6 +158,398 @@ public void HelixRenderPackage_MeshColorBufferCheck()
...
+
+            Assert.AreEqual(0, p.LineVertexColorCount);
+
+            testColors = WhiteByteArrayOfLength(3);
+
+            try
+            {
+                p.AppendLineVertexColorRange(testColors);
+            }
+            catch (Exception e)
```
aparajit-pratap, line 331: Can we make these exceptions thrown be more specific?

src/VisualizationTests/HelixRenderPackageTests.cs
saintentropy: All the exceptions (my new ones and the old ones) within the `RenderPacakge` implementation are caught here -> https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs#L303.  The net affect is that the node does not display geometry in the background preview.

saintentropy: The only difference in the new implementation is the specific handling of the `LegacyRenderPackageMethodException

aparajit-pratap (approved):

src/VisualizationTests/HelixRenderPackageTests.cs
saintentropy: I think I can cover these in the next PR as I add more test coverage.

PR #11398, opened 2021-01-12
[AGD-1883] Support storing user defined tooltip info for nodes

SHKnudsen:
### Purpose

This pull request addresses JIRA task [AGD-1883](https://jira.autodesk.com/browse/AGD-1883).

**This pull request does**

- Adds `NodeInputData` to `DropDown` and `Selection` nodes
- Adds a new `NodeInputType` _(itemSelection)_ to use with DropDown nodes.
- Removes the `IsInputNode` override on the `ColorPalette` node
- Adds a new `UserDescription` to `NodeModel` and `NodeViewModel` which can be modified by user.

The new `UserDescription` will eventually have a UI where these descriptions can be controlled on Input and Output nodes. This is demonstrated below (disregard the **horrible** UI, this was just used for testing.)
InputOutputModifier

### Declarations

### Reviewers

@saintentropy @nate-peters 

### FYIs


src/Libraries/CoreNodeModels/Input/ColorPalette.cs
```diff
@@ -169,14 +169,5 @@ public override IEnumerable<AssociativeNode> BuildOutputAst(List<AssociativeNode
 
             return new[] { AstFactory.BuildAssignment(GetAstIdentifierForOutputIndex(0), colorNode) };
         }
-
-        /// <summary>
-        ///     Indicates whether node is input node. 
-        ///     Used to bind visibility of UI for user selection.
-        /// </summary>
-        public override bool IsInputNode
```
mjkkirschner, line 177: does this node actually execute heedlessly?

saintentropy: as apposed to other input nodes?

saintentropy: We use these in RDC and get tessellation data with color.  They are not input nodes but they run

mjkkirschner: Why are these not input nodes? I'm just confused by this change.

mjkkirschner: there is a color input type:
https://github.com/DynamoDS/Dynamo/pull/11398/files/c374680730481e808d9235df4ace55254a3c0d6b#diff-4ed48c5f49bfda18fef9e3525ec827e474feb30050cf2ce12dce35566bfb970aR20

src/Libraries/CoreNodeModels/DropDown.cs
```diff
@@ -63,13 +63,24 @@ public ObservableCollection<DynamoDropDownItem> Items
...
-            get { return null; }
+            get
+            {
+                return new NodeInputData
+                {
+                    Id = this.GUID,
+                    Name = this.Name,
+                    Type = NodeInputTypes.itemSelection,
+                    Description = this.Description,
+                    Value = this.SelectedString
```
mjkkirschner, line 81: shouldn't this also serialize the `SelectedIndex`? @QilongTang I know you ran into trouble and needed both index and selected value when deserializing dropdowns - is that correct?

src/DynamoCore/Graph/Nodes/NodeInputData.cs
```diff
@@ -22,7 +22,9 @@ public enum NodeInputTypes
         [EnumMember(Value = "date")]
         dateInput,
         [EnumMember(Value = "selection")]
-        selectionInput
+        selectionInput,
+        [EnumMember(Value = "itemSelection")]
```
mjkkirschner, line 26: what is  the difference between itemSelection and selection? I think selection also just returns a string...

src/Libraries/CoreNodeModels/DropDown.cs
QilongTang: Yes, I would also recommend saving index for additional safe guard. This is because there is no guarantee that in the new context the list options will be same or the serialized selection is still available.

src/DynamoCore/Graph/Nodes/NodeInputData.cs
SHKnudsen: I think the idea here was to have a way of distinguish between selecting something in the host environment (i.e. `Select Model Elements`) and selecting an item from a list (i.e. a dropdown selection).

@saintentropy can probably elaborate on this.

src/Libraries/CoreNodeModels/DropDown.cs
SHKnudsen: @mjkkirschner @QilongTang - Would you propose that we add a new `SelectedIndex` to the [`NodeInputData` model](https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Graph/Nodes/NodeInputData.cs)?

```csharp
        /// <summary>
        /// The index of the selected index.
        /// </summary>
        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
        public int SelectedIndex { get; set; }

Comment (QilongTang) [reply src/Libraries/CoreNodeModels/DropDown.cs:81]: Yes you can refer to https://github.com/DynamoDS/Dynamo/blob/master/src/Libraries/CoreNodeModels/DropDown.cs#L80 to see what kind of problems we are guarding

File: src/DynamoCore/Graph/Nodes/NodeInputData.cs (hunk 3 above)
Comment (nate-peters) [reply src/DynamoCore/Graph/Nodes/NodeInputData.cs:26]: @saintentropy and I just talked about this for a bit. The real issue is that the existing "selection" \ `selectionInput` type is ambiguously named, and in a Player UI context it could be interpreted as either a button (like a "Select Elements" button in the Revit context) or as a dropdown list.

@Dewb suggested that we just deprecate the existing "selection" member, and introduce two new members, one for buttons and one for dropdowns. I think "hostElementSelection" or just "hostSelection" for the first, and "listSelection" or "dropdownSelection" for the second.

File: src/Libraries/CoreNodeModels/Input/ColorPalette.cs (hunk 1 above)
Comment (saintentropy) [reply src/Libraries/CoreNodeModels/Input/ColorPalette.cs:177]: @mjkkirschner So the color type was originally designed and implemented in `NodeInputData` -> https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Graph/Nodes/NodeInputData.cs#L20.   This node also implemented an override to store that data -> https://github.com/DynamoDS/Dynamo/blob/master/src/Libraries/CoreNodeModels/Input/ColorPalette.cs#L50.  The only change here is we set this node to not be able to set as an input node.  Not sure exactly why but I think it was because the player had not implemented a picker.  Since then the Revit player has added ability to show warnings for any nodes that are not supported.  So in practice the only thing we are changing is the ability of the user to set this node as an input..

Comment (mjkkirschner) [reply src/Libraries/CoreNodeModels/Input/ColorPalette.cs:177]: Right - but we don't know who else is using this behavior, and I don't think it's a bug. No particularly strong opinion on this case but changing behavior should be done carefully.

* What will happen to previous graphs that set this as an input?
* If some other extension or integration is using this behavior - what is the consequence for them?
* What is the benefit of this change?

Comment (mjkkirschner) [reply src/Libraries/CoreNodeModels/Input/ColorPalette.cs:177]: @saintentropy can you or @nate-peters answer these questions? I'm interested in what the benefit is and if it's worth any potential breaks because of the change in behavior.

**Comment by nate-peters** [comment]
Hi @SHKnudsen - this is looking good. Can you update this comment to reflect the current set of supported input node types? We should add color, and remove the bit about excluding path nodes, since File Input nodes that return a path are also valid input types.
https://github.com/DynamoDS/Dynamo/blob/6cacdf3ebc65862159eb5a83b609b865f7073f5b/src/DynamoCore/Graph/Nodes/NodeModel.cs#L226

File: src/DynamoCore/Graph/Nodes/NodeModel.cs [hunk 4]
```diff
@@ -715,6 +715,16 @@ public string Description
             }
         }
 
+        /// <summary>
+        /// Node description that can be customized by the user.
+        /// </summary>
+        [JsonIgnore]
+        public string UserDescription
```
mjkkirschner, line 722: @nate-peters @SHKnudsen @saintentropy what do you guys think about making this internal and only exposing the viewmodel wrapper?

src/Libraries/CoreNodeModels/DropDown.cs
```diff
@@ -63,13 +63,25 @@ public ObservableCollection<DynamoDropDownItem> Items
                 RaisePropertyChanged("Items");
             }
         }
+
         public override NodeInputData InputData
         {
             //TODO There is not yet an appropriate input type
```
mjkkirschner, line 69: @SHKnudsen update this comment?

src/Libraries/CoreNodeModels/Input/ColorPalette.cs
mjkkirschner: @nate-peters cleared this up.

mjkkirschner:
This PR definitely needs some tests!

* new node input types need serialization tests - if required we may need test classes that implement the base classes, like dropdown, I think theres one floating around in the test assemblies somewhere.
* tests for the properties - userDescription

src/DynamoCore/Graph/Nodes/NodeInputData.cs
```diff
@@ -83,6 +88,12 @@ public class NodeInputData
         [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
         public string Description { get; set; }
 
+        /// <summary>
+        /// The index of the selected index.
```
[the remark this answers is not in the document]
SHKnudsen: haha, yea i think the intention was to say `Index of the selected item` 

will update that.

src/Libraries/CoreNodeModels/DropDown.cs
SHKnudsen: Not entirely sure what this comment referred to with `cogs graph schema` support, so didn't want to remove anything. 

Are we okay to remove the comment?

mjkkirschner: hmm- okay, we could change the TODO to say:
extend cogs graph schema with new `NodeInputTypes.dropdownSelection` support.

src/Libraries/CoreNodeModels/Selection.cs
```diff
@@ -32,7 +32,23 @@ public abstract class SelectionBase<TSelection, TResult> : NodeModel
         
         #region public properties
 
-        public IEnumerable<TSelection> Selection { get { return selection; } } 
+        public IEnumerable<TSelection> Selection { get { return selection; } }
+
+
+        public override NodeInputData InputData
```
mjkkirschner, line 38: @sm6srw this *should* support most Revit selection nodes automatically right?

sm6srw: > @sm6srw this _should_ support most Revit selection nodes automatically right?

Yes, I think so.

SHKnudsen:
Been trying to track down what is causing some tests to fail, haven't been able to solve it yet, but will look at it again tomorrow.

mjkkirschner (approved):

PR #11404, opened 2021-01-14
test overlap

zeusongit:
Please Note:
1. Before submitting the PR, please review [How to Contribute to Dynamo](https://github.com/DynamoDS/Dynamo/blob/master/CONTRIBUTING.md)
2. Dynamo Team will meet 1x a month to review PRs found on Github (Issues will be handled separately)
3. PRs will be reviewed from oldest to newest
4. If a reviewed PR requires changes by the owner, the owner of the PR has 30 days to respond. If the PR has seen no activity by the next session, it will be either closed by the team or depending on its utility will be taken over by someone on the team
5. PRs should use either Dynamo's default PR template or [one of these other template options](https://github.com/DynamoDS/Dynamo/wiki/Choosing-a-Pull-Request-Template) in order to be considered for review.
6. PRs that do not have one of the Dynamo PR templates completely filled out with all declarations satisfied will not be reviewed by the Dynamo team.
7. PRs made to the `DynamoRevit` repo will need to be cherry-picked into all the DynamoRevit Release branches that Dynamo supports. Contributors will be responsible for cherry-picking their reviewed commits to the other branches after a `LGTM` label is added to the PR.

### Purpose

(FILL ME IN) This section describes why this PR is here. Usually it would include a reference 
to the tracking task that it is part or all of the solution for.

### Declarations

Check these if you believe they are true

- [ ] The codebase is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.
- [ ] This PR modifies some build requirements and the readme is updated

### Reviewers

(FILL ME IN) Reviewer 1  (If possible, assign the Reviewer for the PR)

(FILL ME IN, optional) Any additional notes to reviewers or testers.

### FYIs

(FILL ME IN, Optional) Names of anyone else you wish to be notified of

PR #11409, opened 2021-01-15
Fix node overlap on node autocomplete

aparajit-pratap:
### Purpose

An alternative approach for https://github.com/DynamoDS/Dynamo/pull/11404.
autocompleteOverlap

### Declarations

### FYIs

@QilongTang

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
@@ -12,6 +12,7 @@ using Dynamo.Logging;
 using Dynamo.Models;
 using Dynamo.Search.SearchElements;
 using Dynamo.Selection;
+using Dynamo.Utilities;
 using Dynamo.ViewModels;
 using FontAwesome.WPF;
 using Microsoft.Practices.Prism.Commands;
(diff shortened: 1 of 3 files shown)


src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
```diff
@@ -306,18 +307,37 @@ private void AutoLayoutNodes(object sender, EventArgs e)
...
+            var dynamoViewModel = n.ViewModel.DynamoViewModel;
+
+            DynamoSelection.Instance.Selection.Add(n.ViewModel.NodeModel);
+
+            n.ViewModel.WorkspaceViewModel.BeginDragSelectionWithUndoRecorder(new Point2D(n.ViewModel.X, n.ViewModel.Y));
+
+            dynamoViewModel.ExecuteCommand(new DynamoModel.DragSelectionCommand(
+                new Point2D(n.ViewModel.X - n.ActualWidth / 2.0, n.ViewModel.Y),
+                DynamoModel.DragSelectionCommand.Operation.EndDrag));
+
```
zeusongit, line 330: Can you explain what you are doing here?

QilongTang: By reading, this seems simply moving the newly placed node half the node width to the left to make sure it does not cover. You can see it's triggered when `OnSizeChanged` so the actual width is correct for the new node.  I am curious about the effect on other cases because this should also impact other regular cases. And also output node case is not covered here

aparajit-pratap: @QilongTang your understanding is correct. I'm simply moving the node to the left by half its width when its correct width is known, which happens on the `OnSizeChanged` event. However, there is still an issue here as this event doesn't get triggered for all nodes consistently. As per my observation, I think it happens for only NodeModel (UI) nodes. For regular nodes, the `ActualWidth` obtained in `NodeViewReady` is the real width of the node. 

So the issue is if the `OnSizeChanged` event is not triggered consistently for all nodes, the `RepositionNode` method will not be called in all cases, and we cannot unsubscribe events or close the undo action group consistently, which is a problem. I'm still trying to figure that out.

QilongTang:
It seems there is a bug currently on this branch that if I use node autocomplete and then use undo, Dynamo crashes right away.
SliderAutoCompleteCrash

PR #11414, opened 2021-01-20
Fix node overlap issue by skipping the original node from AutoLayout repositioning

zeusongit:
### Purpose

Wider nodes tend to overlap the parent node, the fix is to include the parent node in the auto layout group, and then restrict original node's repositioning to prevent jumping effect.

Problem:
node_placement_bug
 
Fix:
ovrlp2

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCore/Graph/Workspaces/LayoutExtensions.cs
+++ b/src/DynamoCore/Graph/Workspaces/LayoutExtensions.cs
@@ -20,7 +20,7 @@ namespace Dynamo.Graph.Workspaces
         /// </summary>
         /// <param name="workspace">Workspace on which graph layout will be performed.</param>
         /// <param name="reuseUndoRedoGroup">If true, skip initializing new undo action group.</param>
-        internal static List<GraphLayout.Graph> DoGraphAutoLayout(this WorkspaceModel workspace, bool reuseUndoRedoGroup = false)
+        internal static List<GraphLayout.Graph> DoGraphAutoLayout(this WorkspaceModel workspace, bool reuseUndoRedoGroup = false, bool isNodeAutoComplete = false, Guid? originalNodeGUID = null)
         {
             if (workspace.Nodes.Count() < 2) return null;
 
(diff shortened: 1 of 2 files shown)


QilongTang:
Seems in certain case, it is still covering a little bit. More testing on the way
SliderAutoComplete

QilongTang:
This solution does come with an advantage that if user autocomplete not based on the order of the ports, they will see our auto layout clean up the sequence of the nodes for them instead of have the two result nodes mingled..
image

src/DynamoCore/Graph/Workspaces/LayoutExtensions.cs
```diff
@@ -426,6 +433,46 @@ private static void SaveLayoutGraph(this WorkspaceModel workspace, List<GraphLay
...
+                            note.ReportPosition();
+                        }
+                    }
+                }
+            }
+        }
+        /// <summary>
+        /// This method pushes changes from the GraphLayout.Graph objects
+        /// back to the workspace models, but only for nodes placed by NodeAutocomplete.
+        /// </summary>
```
QilongTang, line 453: Both versions of the function are missing comments for input params, do you mind adding them back?

QilongTang (approved):
LGTM with one comment. By comparing this solution with the other one by @aparajit-pratap, I see advantage for both. Due to the following reasons, I would like to ship this one for now:
- Node auto complete no longer depending on the order of the ports
- This solution is a bit simpler although less precise on the right amount to move, but do deliver a solution with less maintaining cost

aparajit-pratap:
IMHO this solution is not clean as it tampers with the original graph layout algorithm by adding a condition that if laying out for node-auto complete then do this other layout logic. Also, the placement of the nodes is not clean either as the nodes are placed on the top of the given node and not alongside it. I'm not sure if @Amoursol has any opinion on this but that's just me. Furthermore, I proposed a solution that I was hoping that the Morpheus team could take and continue working on/refining but it's unfortunate that no such attempt has not been made.

PR #11415, opened 2021-01-20
[DYN-3364] Cherry-pick workflow

Astul-Betizagasti:
### Purpose

Includes a GitHub workflow that automates cherry-picking commits from master to other branches on the repository.

### How it works 

Include the following command on the description of the "squash and merge" box when merging a pull request

`Cherry-pick to: 'branch-name'`

Example: 
image

Other valid variations of the command are:

- `cherry-pick to: 'branch-name'`
- `cherry-pick to:'branch-name'`
- `Cherry-pick to:'branch-name'`

### Declarations

### Reviewers

Aaron Tang (@QilongTang )

### FYIs

Alfredo Pozo (@alfredo-pozo )


.github/workflows/cherry-picking.yaml
```diff
@@ -0,0 +1,38 @@
...
+        replace-with: "-"    
+    #Checks the message looking for a cherry-pick request and extracts the target branch name
+    - name: Check Information
+      id: check-info
+      run: |
+        echo "destination_branch=$(pwsh .\\.github\\scripts\\cherry_pick_check.ps1 "${{ steps.remove_quotations.outputs.replaced }}" )" >> $GITHUB_ENV
+    - name: Create PR to branch
+      #If a target branch was found will run the action
+      if: env.destination_branch != 'invalid'
+      uses: gorillio/github-action-cherry-pick@master
```
alfredo-pozo, line 32: The only concern for me is if this action will be maintained on the long term, because is not supported by someone like GitHub or Microsoft, on that case I will prefer to create our own script that will do the logic.

What do you think @QilongTang ?

QilongTang: @alfredo-pozo You have a good point. I am less worried about this given Git cli commands are very backward compatible. I am hesitate to invest too much in writing our own action not because it's not useful but because of the effort to maintain them. We will rely on a lot of actions to finish our automations so I say we leave it as it is for now and focus on other Dynamo improvements. Thank you for the comment though

.github/scripts/cherry_pick_check.ps1
```diff
@@ -0,0 +1,15 @@
+param([string]$commitMessage)
+
+#Looks for the cherry-pick command on the commit message
+$commitData = $commitMessage | Select-String -Pattern "Cherry-pick to:\s?'*'"
+
+#Extracts the branch name
+if($commitData){
+    $splitedData = $commitData.Line.split("'")
+    $branchName = $splitedData[1]
```
QilongTang, line 9: So this means the Cherry-pick to `branchname` needs to be in the first line of the commit message right? Otherwise, other commit message may rune this condition, would you add comments?
e.g.
* commit message add resource string 'xxx'
* ...

Cherry-pick to `target branch`

Astul-Betizagasti: The "cherry-pick to: 'branch'" command is captured on line 4 and currently it can be in any part of the description. The lines you selected (8,9) are the ones that extract the branch name from the command. Maybe "commitData" isn't the best name for that variable as it can generate some confusion.

QilongTang: Oh I see, so the `commitData` will be the string `Cherry-pick to 'target branch'`? Would you add the format as comment on line 4?

Astul-Betizagasti: That is correct, added the comment and changed that variable name for something more representative of its value

.github/workflows/cherry-picking.yaml
```diff
@@ -0,0 +1,38 @@
...
+  cherry_pick:
+    runs-on: ubuntu-latest
+    env:
+      #Variable for the name of the branch to cherry-pick into.
+      #It will remain 'invalid' if no branch is specified
+      destination_branch: 'invalid'
+    steps:
+    - name: checkout
+      uses: actions/checkout@v2
+    #Removes posible conflicting characters on the commit message
```
QilongTang, line 16: conflicting characters are？

QilongTang (approved):

QilongTang:
Looks good, just curious what conflicting characters will be replaced

.github/workflows/cherry-picking.yaml
Astul-Betizagasti: In this case, quotation marks, this is because the content of the message will be passed to a script as a parameter (line 28) and quotation marks will split the text as if it where multiple parameters.

QilongTang: Thanks, please put this as comment below this line. After that we can merge this and give it a try.

PR #11420, opened 2021-01-21
Update libg nugets

aparajit-pratap:
### Purpose

Update libg nugets.

Change log:
- Some memory leak fixes
- Added SetAttributes and GetAttributes internal methods
- Optimizations to Surface methods

### Declarations

### FYIs

@QilongTang

### DIFF
--- a/src/DynamoCore/packages.config
+++ b/src/DynamoCore/packages.config
@@ -1,8 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="DynamoVisualProgramming.LibG_225_0_0" version="2.9.0.3356" targetFramework="net48" developmentDependency="true" />
-  <package id="DynamoVisualProgramming.LibG_226_0_0" version="2.9.0.3355" targetFramework="net48" developmentDependency="true" />
-  <package id="DynamoVisualProgramming.LibG_227_0_0" version="2.9.0.3355" targetFramework="net48" developmentDependency="true" />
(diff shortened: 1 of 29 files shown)


mjkkirschner:
LGTM

src/DynamoCore/packages.config
```diff
@@ -1,8 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="DynamoVisualProgramming.LibG_225_0_0" version="2.9.0.3356" targetFramework="net48" developmentDependency="true" />
-  <package id="DynamoVisualProgramming.LibG_226_0_0" version="2.9.0.3355" targetFramework="net48" developmentDependency="true" />
-  <package id="DynamoVisualProgramming.LibG_227_0_0" version="2.9.0.3355" targetFramework="net48" developmentDependency="true" />
+  <package id="DynamoVisualProgramming.LibG_226_0_0" version="2.9.0.3865" targetFramework="net48" developmentDependency="true" />
```
mjkkirschner, line 3: @aparajit-pratap one question about this change -  does this actually remove libG_225 from the bin folder? We just need to consider two things when doing this:
1. should we also remove 225 from the supported version lists at the same time.
2. Does this effect patch installers - has anyone included multiple libG folders in their patch installers? Should we just keep 225 around even if we dont support it, just in case.

aparajit-pratap: It might result in removing the libg_225 folder with a fresh build of Dynamo, I'll need to test to make sure. For the time being, I can leave libg_225 around with the last NuGet package version we have for it. I can file a task to actually remove it later (including supported versions).

mjkkirschner: thanks!

PR #11428, opened 2021-01-26
Switch to use sync version of CPython setup/installation method

QilongTang:
Please Note:

### Purpose

Regarding https://jira.autodesk.com/browse/REVIT-173181.

Switch to use sync version of CPython setup/installation method as suggested by @mmisol . 

ALittleHangInstallingCpython

If team is OK with this, let's ship the hotfix this way. Notice the little hang only happens first time user trying to evaluate a CPython node

### Declarations

### Reviewers

@DynamoDS/dynamo 

### FYIs

@mmisol @mjkkirschner @saintentropy

### DIFF
--- a/src/Libraries/DSCPython/CPythonEvaluator.cs
+++ b/src/Libraries/DSCPython/CPythonEvaluator.cs
@@ -3,7 +3,6 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
-using System.Threading.Tasks;
 using Autodesk.DesignScript.Runtime;
 using DSCPython.Encoders;
 using Dynamo.Events;
@@ -256,10 +255,8 @@ namespace DSCPython
         {
             if (!isPythonInstalled)
             {
-                Task.Run(() => {
-                   Python.Included.Installer.SetupPython().Wait();
-                   isPythonInstalled = true;
-               }).Wait();
+                Python.Included.Installer.SetupPythonSync();
+                isPythonInstalled = true;
             }
         }
 
--- a/src/Libraries/DSCPython/CPythonEvaluator.cs
+++ b/src/Libraries/DSCPython/CPythonEvaluator.cs
@@ -3,6 +3,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Threading.Tasks;
 using Autodesk.DesignScript.Runtime;
 using DSCPython.Encoders;
 using Dynamo.Events;
@@ -255,8 +256,10 @@ namespace DSCPython
         {
             if (!isPythonInstalled)
             {
-                Python.Included.Installer.SetupPython().Wait();
-                isPythonInstalled = true;
+                Task.Run(() => {
+                   Python.Included.Installer.SetupPython().Wait();
+                   isPythonInstalled = true;
+               }).Wait();
             }
         }
 
--- a/src/PythonMigrationViewExtension/MigrationAssistant/ScriptMigrator.cs
+++ b/src/PythonMigrationViewExtension/MigrationAssistant/ScriptMigrator.cs
@@ -80,7 +80,7 @@ namespace Dynamo.PythonMigration.MigrationAssistant
         {
             if (!isPythonInstalled)
             {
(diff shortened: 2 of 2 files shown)


src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -255,8 +256,10 @@ private static void InstallPython()
         {
             if (!isPythonInstalled)
             {
-                Python.Included.Installer.SetupPython().Wait();
-                isPythonInstalled = true;
+                Task.Run(() => {
+                   Python.Included.Installer.SetupPython().Wait();
+                   isPythonInstalled = true;
+               }).Wait();
```
mmisol, line 262: Hey Aaron. I must admit when I heard about the hang I remember I also happened to experience it. Back then, what I did to get rid of it is to make `SetupPython` fully synchronous, which is actually a simple change in our `Python.Included` fork https://git.autodesk.com/Dynamo/pythonnet/pull/14/files#diff-4cc075ba367f8ff85f7aff06fd5c060aR30. That change didn't make it into master because it was part of the draft for Python module installation support, which is still a RFC.
IMO, making `SetupPython` synchronous is a simpler solution. Besides, since the unzip action is almost instantaneous and runs only once it doesn't make much sense for it to be async. That would mean that we would just call `Python.Included.Installer.SetupPython();` here.
An alternative, if we don't want to change the existing `SetupPython` function, could be to create a synchronous variation named `SetupPythonSync` that basically uses the same code I linked.
What do you think about this @QilongTang @mjkkirschner ?

mjkkirschner: I like adding the new method!

mjkkirschner: what do you guys think about actually wrapping pythonIncluded in our own namespace? I guess I am concerned that if we need custom functionality that if we a regular build of PythonIncluded were loaded into Revit's process we would hit a missing method exception.

mmisol: I guess we could do that but not sure if now is a good time, given it would represent a breaking change. Python .NET could also potentially present clashes. Not sure if we have tested Revit + pyRevit + Dynamo, but we should probably do it to have an idea of how urgent this change should be. What do you think @mjkkirschner ?

mjkkirschner: @mmisol I'm fine waiting as well and we can discuss those changes together instead - but I am curious - in what sense is it a breaking change?

QilongTang: @mmisol I'm fine with that, do you mind make that change and I can help testing

QilongTang: Hi @mmisol You just reminded me, we tried that yesterday and found we need to wait for the unzip to finish before executing the graph. If we just call `Python.Included.Installer.SetupPython();` the install code in our PythonNet is async so user will get this error 
image

mmisol: @mjkkirschner Changing the namespace of a public class is technically a breaking change. Probably no-one is using it directly except us, so it could be low risk.

@QilongTang We don't just need a change in Dynamo, but we also need to change Python.Included. If you check the link I shared earlier you'll see the `SetupPython` function was changed to being synchronous, though It's probably best if we add a new one named `SetupPythonSync` and leave `SetupPython` unchanged.

QilongTang: @mmisol Cool. Go ahead for that

mjkkirschner:
This also needs to consume the new binary right? (I think manually copy to extern)

QilongTang:
> This also needs to consume the new binary right? (I think manually copy to extern)
Indeed, @mmisol helped https://github.com/DynamoDS/Dynamo/pull/11429

mmisol:
@QilongTang Could you search for `SetupPython().Wait` in the code base? I'm pretty sure this was being done someplace else. It would be nice to make the fix there too.

mmisol (approved):
Looks good to me. Thanks!

PR #11454, opened 2021-02-03
Option for users to exclude specified list of packages from loading

aparajit-pratap:
### Purpose

DYN-3325
This explores adding an option to exclude loading certain packages selected by the user. A use case could be where a user doesn't want specific packages to load by default from the standard library. Proposed here is a solution where a package exclusion list is added to the global preference settings file. This is the format it will look like in the Dynamo preference settings XML:
```
<PackagesToExcludeLoading>
	<Package>
		<Name>MeshToolkit</Name>
		<Version>2.0.1</Version>
	</Package>
</PackagesToExcludeLoading>
```

### Declarations

### Reviewers

### FYIs

### DIFF
--- a/src/DynamoCore/Configuration/PreferenceSettings.cs
+++ b/src/DynamoCore/Configuration/PreferenceSettings.cs
@@ -9,6 +9,13 @@ using Dynamo.Interfaces;
 
 namespace Dynamo.Configuration
 {
+    public class Package
+    {
+        public string Name { get; set; }
+
+        public string Version { get; set; }
+    }
+
     /// <summary>
     /// PreferenceSettings is a class for GUI to persist certain settings.
     /// Upon running of the GUI, those settings that are persistent will be loaded
@@ -233,6 +240,12 @@ namespace Dynamo.Configuration
 
         #region Dynamo application settings
 
+        // TODO: Add to IPreferences in Dynamo 3.0
+        /// <summary>
+        /// List of packages referenced by (name, version) pair to exclude from loading.
+        /// </summary>
+        public List<Package> PackagesToExcludeLoading { get; set; }
+
         /// <summary>
         /// The decimal precision used to display numbers.
         /// </summary>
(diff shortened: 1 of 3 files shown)


mjkkirschner:
@aparajit-pratap looks pretty, I like that the prefs are also stored in a location that requires privileged user access to modify.

I'm curious - were you also going to add an "allowed" List? Or only an exclusion list? Allowed actually seems safer...


I was thinking about the other issue some more - regarding packageConfig per workspace - just a few ideas:

* defer package loading until a workspace is loaded - this would need to be some kind of "safe mode" - maybe entered by the presence of a file or a preference bool.
* much larger project - look into loading packages with reflection for inspection and then lazily load them later.

aparajit-pratap:
@mjkkirschner the reason I chose an exclusion list and not an allowed list, which is what I started off with but later changed my mind, is because I thought it would be easier for users to enumerate what packages they didn't want deliberately rather than know upfront all of the packages that they wanted and inadvertently limit themselves from getting certain essential standard packages that are otherwise shipped out of the box as part of the standard library.

aparajit-pratap:
@mjkkirschner I like the idea of deferred package loading and lazy loading or loading only those packages required by the workspace + workspace-level allowed/excluded list. I did not consider these initially. I can try and come up with a poc for one/both of these ideas as well.

src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -9,6 +9,13 @@
 
 namespace Dynamo.Configuration
 {
+    public class Package
```
pinzart90, line 12: does this need to be public ?

pinzart90: Also a description could be useful

aparajit-pratap: Yes, I think so, for XML serialization.

Thomas84:
Sometimes we need just some specific nodes from old packages. So it would be nice to define a whitelist for specific package nodes instead of not loading a whole package. A similar concept is described here here: https://github.com/DynamoDS/DynamoWishlist/issues/169

aparajit-pratap:
Closing this as the team has collectively decided not to proceed with the package exclusion list approach.

PR #11462, opened 2021-02-08
Use checkboxes to select node libraries

mmisol:
### Purpose

Changes the publish package dialog to use checkboxes rather than a
context menu to select which dlls are node libraries. Also adds an
instructive text on top to further guide users.

Snapshot:
Screen Shot 2021-02-08 at 9 45 09 AM

### Declarations

### DIFF
--- a/src/DynamoCoreWpf/Properties/Resources.Designer.cs
+++ b/src/DynamoCoreWpf/Properties/Resources.Designer.cs
@@ -4543,6 +4543,15 @@ namespace Dynamo.Wpf.Properties {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to Select Node Libraries.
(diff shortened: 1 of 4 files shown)


QilongTang (approved):
LGTM. Curious if you had the chance to test it locally to check if the pkg.json has IsNodeLib set correctly

mmisol:
@QilongTang I tested it but I had to do some code changes to run it outside Revit (had some errors otherwise) and also avoid the actual publishing. It should be equivalent to actual usage I think.

QilongTang:
> @QilongTang I tested it but I had to do some code changes to run it outside Revit (had some errors otherwise) and also avoid the actual publishing. It should be equivalent to actual usage I think.

Great!

PR #11464, opened 2021-02-08
Change If node behaviour to avoid replication over true and false scopes

aparajit-pratap:
### Purpose

DYN-1976
Proposed a solution where the following DS code is translated as AST output for the `If` NodeModel node:
```
[t, f][cond ? 0 : 1];
```
These changes cause the node to behave as per common user expectation as well as match the `ScopeIf+` clockwork package node behaviour.

This translates to building a conditional expression, an array, and an indexing expression in the AST for the NodeModel node.
Lacing and partial function support do not seem to make much sense for this new `If` node and have been omitted.

WIP for @reddyashish :
- Add this functionality to a new class
- Deprecate the old node while keeping it to maintain backward compatibility
- Add tests

### Declarations

### FYIs

@reddyashish

### DIFF
--- a/src/Libraries/CoreNodeModels/Logic/If.cs
+++ b/src/Libraries/CoreNodeModels/Logic/If.cs
@@ -35,58 +35,11 @@ namespace CoreNodeModels.Logic
             var lhs = GetAstIdentifierForOutputIndex(0);
             AssociativeNode rhs;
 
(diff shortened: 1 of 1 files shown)


reddyashish:
Due to some authentication issue, I couldn't push my changes to this remote branch. So created a new one and closing this.

PR #11479, opened 2021-02-08
WIP use product code to lookup product install when displayName returns null

mjkkirschner:
### Purpose

I found that even though we started cacheing the product keys during asm product lookup - we were not actually using them to look in the registry to find the matching product - we were only using display name.

certain products no longer use their display name, and instead use their product code.

The following PR first tries to lookup the product in the cached registry data using the product name, then it uses the product key if the former returns null.

After this PR DynamoSandbox on my machine finds Revit 2022 and Revit 2021 as products with asm.

### Declarations


pinzart90 (approved):
LGTM
Unit test for this case?

mjkkirschner:
@pinzart90 please take another look - I think I've added a sufficient test and fixed the other tests that this broke (their mocks needed the new method added)

I also added a fallback if someone passes an interface type to this method.

pinzart90 (approved):

mjkkirschner:
@pinzart90 I think i'd like to wait on merging this until the build is back up.

PR #11481, opened 2021-02-09
Adding a new "IF" node that would simply return the truevalue or falsevalue based on the test condition. 

reddyashish:
### Purpose

Task: https://jira.autodesk.com/browse/DYN-1976

This PR is made from the original one by @aparajit-pratap : https://github.com/DynamoDS/Dynamo/pull/11464.

Created a new if node that would create a AST output for the following DS code

`[t, f][cond ? 0 : 1];`

We will have to figure out a new name for this node so that it would be clear to the users and can be found easily.
Added a test to verify the functionality of this new node. 
The current if node can be removed in the next major version if we feel it would not be needed.

If the input condition has a list of boolean values then the output will be replicated based on the test input value.

new if node

### Declarations

### Reviewers

@aparajit-pratap @QilongTang @mmisol @Amoursol @mjkkirschner

### DIFF
--- a/src/Libraries/CoreNodeModels/Logic/If.cs
+++ b/src/Libraries/CoreNodeModels/Logic/If.cs
@@ -35,58 +35,11 @@ namespace CoreNodeModels.Logic
             var lhs = GetAstIdentifierForOutputIndex(0);
             AssociativeNode rhs;
 
-            if (IsPartiallyApplied)
-            {
-                // Build partial function for If node if it were a single function call:
-                // Get.ValueAtIndex([t, f], cond ? 0 : 1);
-                var connectedInputs = new List<AssociativeNode>();
-                var inputs = new List<AssociativeNode>();
-
-                if (InPorts[1].IsConnected && InPorts[2].IsConnected)
-                {
-                    connectedInputs.Add(new IntNode(0));
-                    inputs.Add(AstFactory.BuildExprList(new List<AssociativeNode>
-                        {inputAstNodes[1], inputAstNodes[2]}));
-                }
-                if (InPorts[0].IsConnected)
-                {
-                    connectedInputs.Add(new IntNode(1));
-                    inputs.Add(AstFactory.BuildConditionalNode(inputAstNodes[0], new IntNode(0), new IntNode(1)));
-                }
-
-                var functionNode = new IdentifierListNode
-                {
-                    LeftNode = new IdentifierNode(ProtoCore.AST.Node.BuiltinGetValueAtIndexTypeName),
-                    RightNode = new IdentifierNode(ProtoCore.AST.Node.BuiltinValueAtIndexMethodName)
-                };
-                var paramNumNode = new IntNode(2);
-                var positionNode = AstFactory.BuildExprList(connectedInputs);
-
-                var args = new List<AssociativeNode>();
-                args.Add(AstFactory.BuildExprList(new List<AssociativeNode>
-                    {inputAstNodes[1], inputAstNodes[2]}));
-                args.Add(AstFactory.BuildConditionalNode(inputAstNodes[0], new IntNode(0), new IntNode(1)));
-                
-                var arguments = AstFactory.BuildExprList(args);
-                var inputParams = new List<AssociativeNode>
-                {
-                    functionNode,
-                    paramNumNode,
-                    positionNode,
-                    arguments,
-                    AstFactory.BuildBooleanNode(true)
-                };
-
-                rhs = AstFactory.BuildFunctionCall("__CreateFunctionObject", inputParams);
-            }
-            else
-            {
-                var conditional = AstFactory.BuildConditionalNode(inputAstNodes[0], new IntNode(0), new IntNode(1));
-                var trueStmt = inputAstNodes[1];
-                var falseStmt = inputAstNodes[2];
-                var list = AstFactory.BuildExprList(new List<AssociativeNode> {trueStmt, falseStmt});
-                rhs = AstFactory.BuildIndexExpression(list, conditional);
-            }
+            var conditional = AstFactory.BuildConditionalNode(inputAstNodes[0], new IntNode(0), new IntNode(1));
+            var trueStmt = inputAstNodes[1];
+            var falseStmt = inputAstNodes[2];
+            var list = AstFactory.BuildExprList(new List<AssociativeNode> {trueStmt, falseStmt});
+            rhs = AstFactory.BuildIndexExpression(list, conditional);
 
             return new[]
             {
--- a/src/Libraries/CoreNodeModels/Logic/If.cs
+++ b/src/Libraries/CoreNodeModels/Logic/If.cs
@@ -36,7 +36,7 @@ namespace CoreNodeModels.Logic
 
         public override IEnumerable<AssociativeNode> BuildOutputAst(List<AssociativeNode> inputAstNodes)
         {
-            Warning("This node will be removed in a future version of Dynamo. Please use the new 'If' node instead.", true);
+            Warning(Resources.IFNodeWarningMessage, true);
 
             var lhs = GetAstIdentifierForOutputIndex(0);
             AssociativeNode rhs;
--- a/src/Libraries/CoreNodeModels/Logic/If.cs
+++ b/src/Libraries/CoreNodeModels/Logic/If.cs
@@ -101,6 +101,14 @@ namespace CoreNodeModels.Logic
             RegisterAllPorts();
         }
 
+        /// <summary>
+        /// This node will translate the following DS code into an AST output for the If NodeModel node. 
+        ///   result  = [trueValue, falseValue][boolCondition ? 0 : 1]
+        /// This node will just return the trueValue or the falseValue based on the input condition
+        /// and would be different to the conditional "If" node that was present before.
+        /// </summary>
+        /// <param name="inputAstNodes"> List of input AST nodes. </param>
+        /// <returns></returns>
         public override IEnumerable<AssociativeNode> BuildOutputAst(List<AssociativeNode> inputAstNodes)
         {
             var lhs = GetAstIdentifierForOutputIndex(0);
--- a/src/Libraries/CoreNodeModels/Logic/If.cs
+++ b/src/Libraries/CoreNodeModels/Logic/If.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using Autodesk.DesignScript.Runtime;
 using CoreNodeModels.Properties;
 using Dynamo.Graph.Nodes;
 using Newtonsoft.Json;
@@ -15,6 +16,7 @@ namespace CoreNodeModels.Logic
     [OutPortTypes("Function")]
     [IsDesignScriptCompatible]
     [AlsoKnownAs("DSCoreNodesUI.Logic.If")]
+    [IsVisibleInDynamoLibrary(false)]
     [Obsolete("This node will be removed in a future version of Dynamo. Please use the new if node instead.")]
     public class If : NodeModel
     {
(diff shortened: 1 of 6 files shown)


reddyashish:
@Amoursol Yes I have tested this with Empty lists and null values and it works as expected. 
Screen Shot 2021-02-09 at 10 02 28 AM

You can pull in the changes from this branch and test it locally.

Amoursol:
> @Amoursol Yes I have tested this with Empty lists and null values and it works as expected.
> Screen Shot 2021-02-09 at 10 02 28 AM
> 
> You can pull in the changes from this branch and test it locally.

Cool - I'll build and test! :)

aparajit-pratap:
@Amoursol please note that this new node does not have lacing or partial function support. It differs from the old one in that the old one did have partial function support. We did not feel it was very useful for this node as the usage of the If node as a partial function seemed ambiguous. Please let us know if you have other thoughts.

Amoursol:
> @Amoursol please note that this new node does not have lacing or partial function support. It differs from the old one in that the old one did have partial function support. We did not feel it was very useful for this node as the usage of the If node as a partial function seemed ambiguous. Please let us know if you have other thoughts.

@aparajit-pratap - gotcha! I need to test to see if we do need **Lacing** or **List@Level** - both were direct requests from our users when requesting feedback here: https://forum.dynamobim.com/t/request-for-feedback-if-node/45257 

My concern without implementing this would be we are in the same position at a later date, but I'll test to see if the use cases in that thread are covered. The two bugs that appear to be fixed in this are definitely a step in the right direction: Empty List control and various length True/False lists not having data truncated.

aparajit-pratap:
If you're comparing this with the Clockwork node, I'd like to add that it is almost identical to that. Secondly, about replication and list-at-level, in order to make this node match with user expectation of its use, we needed to disable replication for the `true` and `false` inputs (2nd and 3rd resp.). The node can only replicate on the 1st input, the test condition.

aparajit-pratap:
From that forum thread to me it sounds like users are confused as to what Lacing or List@Level can do to help in the context of this node. Whatever output they expect from the node can be achieved without those additions.

aparajit-pratap:
If lacing or list-at-level support is required for this node for consistency sake, I think it might be possible to still add it, however, for partial function support, I think we might need a different approach where we create a special builtin function and call that from this node:
```
def if_func(cond : bool, t : var[]..[], f : var[]..[])
{
    return [Imperative]
    {
        return cond ? t : f;
    }
}
```

Amoursol:
> If lacing or list-at-level support is required for this node for consistency sake, I think it might be possible to still add it, however, for partial function support, I think we might need a different approach where we create a special builtin function and call that from this node:
> 
> ```
> def if_func(cond : bool, t : var[]..[], f : var[]..[])
> {
>     return [Imperative]
>     {
>         return cond ? t : f;
>     }
> }
> ```

I've been busy all day so far, going to download the build and test today/tomorrow :) Just need to see what kind of data I can emulated from a messy Revit source.

src/Libraries/CoreNodeModels/Logic/If.cs
```diff
@@ -72,4 +74,44 @@ public override IEnumerable<AssociativeNode> BuildOutputAst(List<AssociativeNode
             };
         }
     }
+
+    [NodeName("NewIf")]
```
aparajit-pratap, line 78: Why do we need to rename it here?

reddyashish: We still need to decide on the final name for this node. I have just added this name as a draft. Since, the old "IF" node still exists in the Dynamo library, we will need a different name for the node.

Amoursol:
Can we get a build of Dynamo including the **NewIf** node? That way we get get some feedback from users prior to merge.

src/Libraries/CoreNodeModels/Logic/If.cs
aparajit-pratap: I mean to ask if we can keep the same name for the new node while deprecating the old one.

QilongTang:
> @reddyashish Can we get a build of Dynamo including the **NewIf** node? That way we get get some feedback from users prior to merge.

 We can get a branch build out of this work for sure. Is this PR in a good state? If yes, we can merge into a feature branch and start from there

reddyashish:
@Amoursol Yes, we can get the branch build like Aaron suggested. I have added tests and this PR is pending review. Just to figure out a proper name for this name.

Amoursol:
> @Amoursol Yes, we can get the branch build like Aaron suggested. I have added tests and this PR is pending review. Just need to figure out a proper name for this name.

We can also crowd-source that name from the Forum with a testable build 😄  Can you let me know when you have a Branch build up?

src/Libraries/CoreNodeModels/Logic/If.cs
aparajit-pratap: @reddyashish @QilongTang I'm still not sure about this. To give an example, I introduced a new `IntegerSlider64Bit` `NodeModel` class but could keep the same name `Integer Slider` for the node. I think we may be able to do something similar here so that new `If` nodes have new behaviour but the same name. We might just have to figure out a way for old `If` nodes to continue to deserialize as the old NodeModel class.

mjkkirschner: @aparajit-pratap what about if the old IF node actually has its name migrated to "replicating if" or "legacy if" and the new node becomes "If"?

reddyashish: @aparajit-pratap we can have both the nodes with the same name but I feel it might cause some confusion to the users who search for that node as both will be visible now. I feel we should have a different name till we completely obsolete the old node. @mjkkirschner's suggestion looks good to me. What do you think?

aparajit-pratap: @reddyashish I think you're getting confused. The old node will be hidden from the library.

mjkkirschner: @reddyashish - AFAIK @aparajit-pratap is correct that the use of the `[obsolete]` attribute should hide the node - it's possible you need to use `[NodeObsolete]` for that though - I know there is some discrepancy between zero touch and nodeModel nodes though - so watch out for that.

Maybe a good first step is to decide what should happen?

1. should the original node be hidden, but still load when in an existing graph to stay backward compatible?
2. If a user places a new node what should it be called?
3. Is there any way to tell the difference between the old and new nodes in the same graph? (I assume the old ones will be in error state because of the obsolete attr - if it's working right?)

Amoursol: IMHO, from a user perspective they want the following

> @reddyashish - AFAIK @aparajit-pratap is correct that the use of the `[obsolete]` attribute should hide the node - it's possible you need to use `[NodeObsolete]` for that though - I know there is some discrepancy between zero touch and nodeModel nodes though - so watch out for that.
> 
> Maybe a good first step is to decide what should happen?
> 
> 1. should the original node be hidden, but still load when in an existing graph to stay backward compatible?
> 2. If a user places a new node what should it be called?
> 3. Is there any way to tell the difference between the old and new nodes in the same graph? (I assume the old ones will be in error state because of the obsolete attr - if it's working right?)

@mjkkirschner IMHO, from a user perspective:
1. Yes this should happen
2. It should simply be called "If" but reference the updated, new node. 
3. Is the error message obvious if we do this? Or will it simply say it's missing something? Ideally, users would understand that they need to replace that node. Would Copy + Paste of that node into the same place do the trick?

reddyashish: I have updated the code to hide the old "If" node from the library and rename the new one to have the same name. When the users open an old graph with the if node, it will behave like the old one(conditional if). Any new "If" node that will be placed will build on the new functionality. The [NodeObsolete] attribute can only be used for a method or a property and the [obsolete] attribute will throw a warning when the old "if" class is used in any of the code. It wouldn't throw an error on the Dynamo side. Do we want to throw an error or warning for the old "if" nodes?

reddyashish: After discussing with Sol, we will be showing a warning message on the "If" node when an old graph is opened..

Screen Shot 2021-02-16 at 10 52 22 AM

src/Libraries/CoreNodeModels/Logic/If.cs
```diff
@@ -32,13 +36,15 @@ public If()
 
         public override IEnumerable<AssociativeNode> BuildOutputAst(List<AssociativeNode> inputAstNodes)
         {
+            Warning("This node will be removed in a future version of Dynamo. Please use the new 'If' node instead.", true);
```
QilongTang, line 39: 👍🏻

src/Libraries/CoreNodeModels/Logic/If.cs
```diff
@@ -72,4 +78,44 @@ public override IEnumerable<AssociativeNode> BuildOutputAst(List<AssociativeNode
...
+        public override IEnumerable<AssociativeNode> BuildOutputAst(List<AssociativeNode> inputAstNodes)
+        {
+            var lhs = GetAstIdentifierForOutputIndex(0);
+            AssociativeNode rhs;
+
+            var conditional = AstFactory.BuildConditionalNode(inputAstNodes[0], new IntNode(0), new IntNode(1));
+            var trueStmt = inputAstNodes[1];
+            var falseStmt = inputAstNodes[2];
+            var list = AstFactory.BuildExprList(new List<AssociativeNode> { trueStmt, falseStmt });
+            rhs = AstFactory.BuildIndexExpression(list, conditional);
```
QilongTang, line 113: Would you add some comments to these lines indicating the difference with the old node?

QilongTang (approved):
LGTM with one comment

src/Libraries/CoreNodeModels/Logic/If.cs
reddyashish: Added it.

reddyashish:
Just waiting for the self-serve to confirm no regressions.

src/Libraries/CoreNodeModels/Logic/If.cs
QilongTang: @reddyashish Awesome. :shipit:

aparajit-pratap:
⚠️ 
I don't think we have got input from @Amoursol about 3 things for this node:
1. partial function support
2. Lacing
3. list-at-level
These would need more work. Also if we need partial function support, we might not be able to go with the approach in this PR ⚠️

Amoursol:
> ⚠️
> I don't think we have got input from @Amoursol about 3 things for this node:
> 
> 1. partial function support
> 2. Lacing
> 3. list-at-level
>    These would need more work. Also if we need partial function support, we might not be able to go with the approach in this PR ⚠️

Can we please get a working build to let users test against? We can put that into the forum thread and get direct feedback before merging the PR, as noted above in a previous comment :)

reddyashish:
@aparajit-pratap @Amoursol To get a build, we will need to merge it to a feature branch atleast. We will not be merging this to master yet as will do that after the feedback. What do you guys think?

reddyashish:
Changed the base to point to a new branch "RefactoredIfNode".

Amoursol:
> Changed the base to point to a new branch "RefactoredIfNode".

Awesome. Are you able to create a Sandbox build for users to test please @reddyashish?

QilongTang:
@reddyashish Go for it. Let me know if you have any question in terms of creating branch build after merging into feature branch

aparajit-pratap:
> @aparajit-pratap @Amoursol To get a build, we will need to merge it to a feature branch atleast. We will not be merging this to master yet as will do that after the feedback. What do you guys think?

Ok, I thought this was ready to be merged into master. If this is going into a feature branch for some user testing, then that sounds good.

PR #11483, opened 2021-02-09
Catch exceptions from IExtensionStorageAccess

SHKnudsen:
### Purpose

While implementing the new `IExtensionStorageAccess` in GD i noticed a nasty little bug when calling `OnWorkspaceSaving`. Because i wasn't catching any exceptions thrown from the `IExtensionStorageAccess` implementer, any errors on the extension side would break the Dynamo save process, and make it impossible to save the current graph.

I wrapped `OnWorkspaceSaving` and `OnWorkspaceOpen` in a try/catch, so the Dynamo save and open process will continue even if the extension throws any exceptions.

For now im just logging the error to the debug console but would appreciate your inputs on if/how we should surface this to the user?

### Declarations

### Reviewers

@mjkkirschner 

### FYIs

@saintentropy

### DIFF
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -977,7 +977,16 @@ namespace Dynamo.Models
         {
             workspace.TryGetMatchingWorkspaceData(extension.UniqueId, out Dictionary<string, string> data);
             var extensionDataCopy = new Dictionary<string, string>(data);
-            extension.OnWorkspaceOpen(extensionDataCopy);
+
+            try
+            {
+                extension.OnWorkspaceOpen(extensionDataCopy);
+            }
+            catch (Exception ex)
+            {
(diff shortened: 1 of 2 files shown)


src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -977,7 +977,16 @@ internal static void RaiseIExtensionStorageAccessWorkspaceOpened(WorkspaceModel
...
             var extensionDataCopy = new Dictionary<string, string>(data);
-            extension.OnWorkspaceOpen(extensionDataCopy);
+
+            try
+            {
+                extension.OnWorkspaceOpen(extensionDataCopy);
+            }
+            catch (Exception ex)
+            {
+                Debug.WriteLine(ex.Message + " : " + ex.StackTrace);
```
mjkkirschner, line 987: please use the logger and log these exceptions to the console - I think this will get stripped out in release builds.

mjkkirschner (approved):
LGTM

PR #11487, opened 2021-02-10
Prevent node autocomplete from autohiding

mmisol:
### Purpose

This happened when the amount of suggestions for an input surpassed a
certain number of suggestions. The cause was the following condition.

It turns out that popups, like autocomplete, close when the left mouse
button up event happens. The problem is that autocomplete itself is
launched by pressing the left mouse button. This would normally mean
that the popup would hide as soon as the button was released, but this
was actually prevented through a time condition, which ignored the
event if it was 'soon enough'.

Unfortunately, in cases where there were a lot of suggestions, the left
mouse button up event can come after the expected timespan. This
resulted in the workaround failing to prevent the popup from hiding
immediately.

The newly implemented solution replaces the previous time-based
workaround with a more robust solution, which is to ignore the next
event after the call to show the popup is made.

### Declarations

### DIFF
--- a/src/DynamoCoreWpf/Views/Core/WorkspaceView.xaml.cs
+++ b/src/DynamoCoreWpf/Views/Core/WorkspaceView.xaml.cs
@@ -62,6 +62,7 @@ namespace Dynamo.Views
         private double currentNodeCascadeOffset;
         private Point inCanvasSearchPosition;
         private List<DependencyObject> hitResultsList = new List<DependencyObject>();
+        private bool isAutoCompleteLoading;
 
         public WorkspaceViewModel ViewModel
         {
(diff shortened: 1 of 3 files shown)


src/DynamoCoreWpf/Views/Core/WorkspaceView.xaml.cs
```diff
@@ -390,7 +394,6 @@ void OnWorkspaceViewDataContextChanged(object sender, DependencyPropertyChangedE
                 ViewModel.RequestAddViewToOuterCanvas += vm_RequestAddViewToOuterCanvas;
                 ViewModel.WorkspacePropertyEditRequested += VmOnWorkspacePropertyEditRequested;
                 ViewModel.RequestSelectionBoxUpdate += VmOnRequestSelectionBoxUpdate;
-                ViewModel.RequestNodeAutoCompleteSearch += ShowHideNodeAutoCompleteControl;
```
QilongTang, line 393: I remember this was added so that Node AutoComplete works for custom node workspace. Any reason of removing this? If intended, would you test using node autocomplete in a new custom node?

mmisol: I see. This was causing the `Show` function to be called two times in a row. I can add it back to avoid affecting custom nodes, but it does seem suspicious.

reddyashish (approved):
Looks good to me.

src/DynamoCoreWpf/Views/Core/WorkspaceView.xaml.cs
mmisol: Instead of removing this handler assignment I removed the other one. Tested in both standard and custom node workspace  and it's working. This also prevents the existing behaviour of the handler being called twice on standard workspaces.

QilongTang: Thank you @mmisol . Looks reasonable,

QilongTang (approved):

PR #11500, opened 2021-02-12
[DYN-3343] Automatically add loaded packages as crash detail

Astul-Betizagasti:
### Purpose

Automatically include loaded packages information when creating a crash report as a Github Issue through Dynamo

### Declarations

### Reviewers

Aaron Tang (@QilongTang )

### FYIs

Alfredo Pozo (@alfredo-pozo )


src/DynamoCoreWpf/Utilities/CrashUtilities.cs
```diff
@@ -70,11 +76,13 @@ internal static string BuildMarkdownContent(string dynamoVersion, string stackTr
...
                 "## What packages or external references (if any) were used?" + Environment.NewLine +
-                "(Fill in here)" + Environment.NewLine + Environment.NewLine + Environment.NewLine +
+                "(Fill in here)" + Environment.NewLine + Environment.NewLine +
+
+                markdownPackages + Environment.NewLine + Environment.NewLine +
 
                 "## Stack Trace" + Environment.NewLine +
                 "```" + Environment.NewLine +
-                "(From the Dynamo crash window select 'Details' -> 'Copy' and paste here)" + Environment.NewLine +
+                stackTrace + Environment.NewLine +
```
Astul-Betizagasti, line 85: @QilongTang I notice the stack trace information was already available to be included in the issue template but was never used so I modified it to be automatically added instead of asking the user to copy and paste the contents of the crash prompt. If there was any specific reason this wasn't done already I can revert the change, just let me know.

QilongTang: hi @Astul-Betizagasti I think certain case that the crash detail is too long to be sent automatically. That was what I remember. @alfarok should be able to confirm why we did not include it then

Astul-Betizagasti: ah I see, well let me know if it should be reverted, and in case it does, should I remove the stack trace being sent to the method all together? It could make for a cleaner code if we are not going to use it

src/DynamoCoreWpf/UI/Prompts/CrashPrompt.xaml.cs
```diff
@@ -38,10 +43,15 @@ public CrashPrompt(DynamoViewModel dynamoViewModel)
             txtOverridingText.Text = string.Format(Wpf.Properties.Resources.CrashPromptDialogCrashMessage, productName);
         }
 
-        public CrashPrompt(CrashPromptArgs args, DynamoViewModel dynamoViewModel)
+        public CrashPrompt(CrashPromptArgs args, DynamoViewModel dynamoViewModel, PackageLoader packageLoader)
```
aparajit-pratap, line 46: API break ⚠️

src/DynamoCoreWpf/Utilities/CrashUtilities.cs
aparajit-pratap: I can't think of why we would not need the stack trace. I think in almost all cases we ask reporters to paste the stack trace in the issue where it's missing so if it can be done automatically there's nothing better than that.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -1272,7 +1272,8 @@ private void Selection_CollectionChanged(object sender, NotifyCollectionChangedE
 
         private void Controller_RequestsCrashPrompt(object sender, CrashPromptArgs args)
         {
-            var prompt = new CrashPrompt(args, dynamoViewModel);
+            var packageLoader = dynamoViewModel.Model.GetPackageManagerExtension().PackageLoader;
+            var prompt = new CrashPrompt(args, dynamoViewModel, packageLoader);
```
aparajit-pratap, line 1276: How about adding the list of packages to `CrashPromptArgs`?

src/DynamoCoreWpf/Utilities/CrashUtilities.cs
Astul-Betizagasti: Yes I agree, I didn't meant removing it from the crash prompt, I meant not sending it through the event when clicking on the button to generate the GitHub issue as a parameter
image
At the moment that's how the method looks and the `this.CrashDetailsContent.Text` parameter has the stack trace information that will not be used, while it could have some other info like the loaded packages for example

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
Astul-Betizagasti: Will check on that. That would also solve the API break right? Thanks for the comments

src/DynamoCoreWpf/Utilities/CrashUtilities.cs
aparajit-pratap: Ah ok, I remember there was an issue with this. @QilongTang is right.

src/DynamoCoreWpf/UI/Prompts/CrashPrompt.xaml.cs
QilongTang: hmm purely UI code here, I understand we are never clear about our API scope but would like to argue that we should be able to break our own UI layer logic.. especially about dialogs not top Dynamo view layer

src/DynamoCoreWpf/Utilities/CrashUtilities.cs
QilongTang: Thanks, I would remove the stacktrace here. Crashes callstacks or VM crashes callstacks are covered by analytics as a backup, I would not include here just to make sure user can always log an issue when they want. If anything interrupts that, it may discourage them from doing that later.

QilongTang:
@Astul-Betizagasti I noticed your fork head is not clean, you may want to clean your master in the future before creating the branch on your fork

test/DynamoCoreWpfTests/CrashReportingTests.cs
```diff
@@ -66,7 +67,7 @@ public void CanReportBugWithContent()
             var dynamoVersion = "2.1.0";
 
             // Create a crash report to submit
-            var crashReport = Wpf.Utilities.CrashUtilities.BuildMarkdownContent(dynamoVersion, StackTrace);
+            var crashReport = Wpf.Utilities.CrashUtilities.BuildMarkdownContent(dynamoVersion, StackTrace, "");
```
QilongTang, line 70: Please use string.empty where you can

test/DynamoCoreWpfTests/CrashReportingTests.cs
```diff
@@ -80,7 +81,7 @@ public void StackTraceIncludedInReport()
             var dynamoVersion = AssemblyHelper.GetDynamoVersion().ToString();
 
             // Create a crash report to submit
-            var crashReport = Wpf.Utilities.CrashUtilities.BuildMarkdownContent(dynamoVersion, StackTrace);
+            var crashReport = Wpf.Utilities.CrashUtilities.BuildMarkdownContent(dynamoVersion, StackTrace, "");
```
QilongTang, line 84: Same here

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
QilongTang: > How about adding the list of packages to `CrashPromptArgs`?

That is worth looking at I agree

src/DynamoCoreWpf/UI/Prompts/CrashPrompt.xaml.cs
```diff
@@ -69,6 +79,34 @@ public CrashPrompt(CrashPromptArgs args, DynamoViewModel dynamoViewModel)
...
+            if (packages.Any())
+            {
+                pckgsNames.AddRange(packages.Select(pckg => pckg.Name));
+            }
+            else
+            {
+                pckgsNames.Add("No packages where loaded.");
+            }
+            return pckgsNames;
+        }
```
QilongTang, line 109: I would maybe consider moving these functions to https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/Utilities/CrashUtilities.cs

src/DynamoCoreWpf/UI/Prompts/CrashPrompt.xaml.cs
QilongTang: Instead of passing it in, get it by `dynamoViewModel.Model.GetPackageManagerExtension().PackageLoader;`

Astul-Betizagasti:
@QilongTang @aparajit-pratap, Just pushed changes that should address all the previous comments, let me know what you think, thanks.

src/DynamoCoreWpf/Utilities/CrashUtilities.cs
```diff
@@ -26,28 +29,48 @@ internal static string GithubNewIssueUrlFromCrashContent(object crashContent)
...
 
-            return BuildMarkdownContent(dynamoVersion, stackTrace);
+            return BuildMarkdownContent(dynamoVersion, markdownPackages);
         }
 
         /// <summary>
         /// Builds a Markdown string that will be posted to our new GitHub issue
         /// </summary>
         /// <param name="dynamoVersion">Dynamo version that should be recorded in the issue report</param>
         /// <param name="stackTrace">The crash stack trace to be included in the issue report</param>
```
QilongTang, line 70: Let's remove this line of the param is no longer passed

src/DynamoCoreWpf/UI/Prompts/CrashPrompt.xaml.cs
```diff
@@ -92,7 +103,7 @@ private void Continue_Click(object sender, RoutedEventArgs e)
 
         private void PostOnGithub_Click(object sender, RoutedEventArgs e)
         {
-            DynamoViewModel.ReportABug(this.CrashDetailsContent.Text);
+            DynamoViewModel.ReportABug(markdownPackages);
```
QilongTang, line 106: Wouldn't this cause that we lose certain details?

src/DynamoCoreWpf/Utilities/CrashUtilities.cs
Astul-Betizagasti: oops missed that, done!

src/DynamoCoreWpf/UI/Prompts/CrashPrompt.xaml.cs
Astul-Betizagasti: previously only the stack trace was being sent there but it was not being used

QilongTang: I see. Please attach a testing gif. I think these looks good for me

QilongTang (approved):

src/DynamoCoreWpf/UI/Prompts/CrashPrompt.xaml.cs
```diff
@@ -42,6 +47,12 @@ public CrashPrompt(CrashPromptArgs args, DynamoViewModel dynamoViewModel)
         {
             InitializeComponent();
 
+            var packageLoader = dynamoViewModel.Model.GetPackageManagerExtension().PackageLoader;
```
mjkkirschner, line 50: be careful here - `GetPackageManagerExtension` might return null in certain circumstances - you can do
` dynamoViewModel.Model.GetPackageManagerExtension()?.PackageLoader` - and then proceed if packageLoader is not null.

Astul-Betizagasti: good catch, will fix

Astul-Betizagasti:
All changes made, here are some gifs showing the behavior: 
No packages
packages1

Some loaded packages
packages2

In the case that @mjkkirschner mentioned, when the package loader is null, the behavior will be similar to the first gif but it will say "(Fill in here)" instead of "No packages where found."

QilongTang:
LGTM

src/DynamoCoreWpf/Utilities/CrashUtilities.cs
```diff
@@ -26,28 +29,61 @@ internal static string GithubNewIssueUrlFromCrashContent(object crashContent)
...
+                {
+                    markdownText = "### Loaded Packages" + Environment.NewLine;
+                    foreach (var name in packagesNames)
+                    {
+                        markdownText += "- " + name + Environment.NewLine;
+                    }
+                }
+                else
+                {
+                    markdownText = "No loaded packages where found.";
```
QilongTang, line 55: were instead of where

src/DynamoCoreWpf/Utilities/CrashUtilities.cs
```diff
@@ -26,28 +29,61 @@ internal static string GithubNewIssueUrlFromCrashContent(object crashContent)
...
+                {
+                    markdownText = "### Loaded Packages" + Environment.NewLine;
+                    foreach (var name in packagesNames)
+                    {
+                        markdownText += "- " + name + Environment.NewLine;
+                    }
+                }
+                else
+                {
+                    markdownText = "No loaded packages were found.";
```
QilongTang, line 55: So I thought about if we need to convert these into resource strings but think it may the customer but not us. What do you guys think @DynamoDS/dynamo

src/DynamoCoreWpf/Utilities/CrashUtilities.cs
```diff
@@ -26,28 +29,61 @@ internal static string GithubNewIssueUrlFromCrashContent(object crashContent)
...
+                //List of the names of all the loaded packages
+                var packagesNames = packageLoader.LocalPackages.Select(o => o.Name).ToList();
+                //Package's issue section in markdown format
+                string markdownText;
+                if (packagesNames.Any())
+                {
+                    markdownText = "### Loaded Packages" + Environment.NewLine;
+                    foreach (var name in packagesNames)
+                    {
+                        markdownText += "- " + name + Environment.NewLine;
```
QilongTang, line 50: Also do you think we can easily get package versions as well? @Astul-Betizagasti

Astul-Betizagasti:
image
@QilongTang Added the packages versions, let me know of this looks good or do we want to include something like "ver" or "version" in between the name and the version?

QilongTang:
@Astul-Betizagasti Can you add a unit test in `CrashReportingTests.cs` covering the new API you added?

test/DynamoCoreWpfTests/CrashReportingTests.cs
```diff
@@ -80,7 +85,7 @@ public void StackTraceIncludedInReport()
             var dynamoVersion = AssemblyHelper.GetDynamoVersion().ToString();
 
             // Create a crash report to submit
-            var crashReport = Wpf.Utilities.CrashUtilities.BuildMarkdownContent(dynamoVersion, StackTrace);
+            var crashReport = Wpf.Utilities.CrashUtilities.BuildMarkdownContent(dynamoVersion, Packages);
```
QilongTang, line 88: I would assert below that  `decoded` contains package info. And also make a new test that covers the default state of the API

QilongTang:
Thanks @Astul-Betizagasti Looks good. Once the unit test updated, we should be able to wrap this one up

test/DynamoCoreWpfTests/CrashReportingTests.cs
Astul-Betizagasti: Did the change suggested on the test, but can you expand on the other test you where suggesting? The API should work the same as before the changes with the exception of it including more information (which is being checked on that test you linked). Are you referring to the method `PackagesToMakrdown(PackageLoader packageLoader)` in CrashUtilities.cs?

QilongTang: yes. I would like a test covering the default state so we can assert it defaults to `No loaded packages were found.`

PR #11502, opened 2021-02-17
Defer loading of workspace referenced packages until workspace open

aparajit-pratap:
### Purpose

This is an attempt to load packages at the time of opening a workspace by looking at the workspace references and only loading the necessary packages that are downloaded.

### Declarations

I have not been able to get this to work after spending a few days as it seems it will require bigger changes to the engine. However, this PR should serve as a reference for future work.


src/DynamoPackages/PackageLoader.cs
```diff
@@ -91,7 +92,7 @@ public PackageLoader(IEnumerable<string> packagesDirectories)
             if (packagesDirectories == null)
                 throw new ArgumentNullException("packagesDirectories");
 
-            this.packagesDirectories.AddRange(packagesDirectories);
+            //this.packagesDirectories.AddRange(packagesDirectories);
```
aparajit-pratap, line 95: Don't load all packages, just the ones in the standard lib.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -102,6 +103,36 @@ public PackageLoader(IEnumerable<string> packagesDirectories)
...
+                    var headerPath = Path.Combine(dir, "pkg.json");
+                    if (PathHelper.IsValidPath(headerPath))
+                    {
+                        var discoveredPkg = Package.FromJson(headerPath, AsLogger());
+                        if(discoveredPkg == null) continue;
+
+                        foreach (var info in packageDependencies)
+                        {
+                            // if workspace package dependency is present in downloaded packages, load it.
+                            if (info.Name == discoveredPkg.Name && info.Version.ToString() == discoveredPkg.VersionName)
```
aparajit-pratap, line 124: Collect all downloaded packages that match with the ones referenced in the workspace.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -102,6 +103,36 @@ public PackageLoader(IEnumerable<string> packagesDirectories)
...
+                            {
+                                pkgs.Add(discoveredPkg);
+                                break;
+                            }
+                        }
+                    }
+                }
+            }
+            LoadPackages(pkgs);
```
aparajit-pratap, line 133: I barely got this to defer-load packages but the workspace doesn't open for some reason :(

src/Engine/ProtoCore/Core.cs
```diff
@@ -828,7 +828,7 @@ public void ResetSSASubscript(Guid guid, int subscript)
 
         public void Cleanup()
         {
-            CLRModuleType.ClearTypes();
+            //CLRModuleType.ClearTypes();
```
aparajit-pratap, line 831: When a new workspace is opened, this static list of types was cleared which caused class redefinition errors downstream as some of the same types were being loaded again due to this list being empty.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1693,6 +1693,8 @@ private bool OpenJsonFileFromPath(string fileContents, string filePath, bool for
                             OnComputeModelDeserialized();
 
                             SetPeriodicEvaluation(ws);
+
+                            CurrentWorkspace = ws;
```
aparajit-pratap, line 1697: Update the current workspace here, after the VM is re-initialized and all libs/packages are reloaded. This will trigger the property changed event for `CurrentWorkspace` at the right time for the `DynamoPackages` extension to react to it and load packages for that new workspace.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1771,7 +1773,7 @@ private void OpenWorkspace(WorkspaceModel ws)
             }
 
             AddWorkspace(ws);
-            CurrentWorkspace = ws;
+            //CurrentWorkspace = ws;
```
aparajit-pratap, line 1776: Hope this doesn't cause any side-effects. As noted above I have moved this so that it happens a little later in the workspace opening sequence.

src/DynamoPackages/PackageManagerExtension.cs
```diff
@@ -224,6 +224,12 @@ private void OnCurrentWorkspaceChanged(IWorkspaceModel ws)
                 (ws as WorkspaceModel).CollectingCustomNodePackageDependencies += GetCustomNodePackageFromID;
                 (ws as WorkspaceModel).CollectingNodePackageDependencies += GetNodePackageFromAssemblyName;
                 currentWorkspace = ws;
+
+                if (PackageLoader != null && ws is HomeWorkspaceModel model)
+                {
+                    PackageLoader.OnWorkspaceRead(model.NodeLibraryDependencies,
+                        ReadyParams.StartupParams.PathManager.PackagesDirectories);
```
aparajit-pratap, line 231: Called once a new workspace is opened.

PR #11508, opened 2021-02-18
Update to new version of python net

sm6srw:
### Purpose

This PR will bump `pythonnet` to version `2.5.1.7`. This is related to DYN-3476.

### Declarations


pinzart90 (approved):

mjkkirschner:
@sm6srw what about the references to Python.Included:
https://github.com/DynamoDS/Dynamo/blob/fddfe07b13483a704a145283a9b8aac93b8d75ee/src/Libraries/DSCPython/DSCPython.csproj#L43

also I may have lead you astray - do we also need to update the python.Runtime.dll? Where is that coming from?

mjkkirschner (changes requested):
see above comment about other references in the DSCPython project.

sm6srw:
> see above comment about other references in the DSCPython project.

It works without changing them (they are both out of sync as it is already) but I will go ahead and update them.

mjkkirschner (approved):

PR #11510, opened 2021-02-21
update libg with find and replace

mjkkirschner:
### Purpose

update libG to latest:
change log:
* reduced memory leaks in solid loft workflow
* reduced memory leaks in polycurve workflows
* reduced memory leaks when throwing exceptions
* swig 4.0
* bug fixes in point projection and point prune duplicates
* faster pt/pt distance calculations
* updated docs for `projectInputOnto` nodes
* NurbsSurfaces operations should return exact surface types in more cases

### Declarations

### DIFF
--- a/src/DynamoCore/DynamoCore.csproj
+++ b/src/DynamoCore/DynamoCore.csproj
@@ -83,8 +83,8 @@ limitations under the License.
       <HintPath>..\packages\Newtonsoft.Json.8.0.3\lib\net45\Newtonsoft.Json.dll</HintPath>
       <Private>True</Private>
     </Reference>
-    <Reference Include="ProtoGeometry, Version=2.9.0.3837, Culture=neutral, processorArchitecture=AMD64">
-      <HintPath>..\packages\DynamoVisualProgramming.LibG_225_0_0.2.9.0.3837\lib\net48\ProtoGeometry.dll</HintPath>
+    <Reference Include="ProtoGeometry, Version=2.11.0.4156, Culture=neutral, processorArchitecture=AMD64">
+      <HintPath>..\packages\DynamoVisualProgramming.LibG_225_0_0.2.11.0.4156\lib\net48\ProtoGeometry.dll</HintPath>
     </Reference>
     <Reference Include="RestSharp, Version=105.2.3.0, Culture=neutral, processorArchitecture=MSIL">
       <HintPath>..\packages\RestSharp.105.2.3\lib\net46\RestSharp.dll</HintPath>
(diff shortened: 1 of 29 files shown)


src/DynamoCore/DynamoCore.csproj
```diff
@@ -83,8 +83,8 @@ limitations under the License.
       <HintPath>..\packages\Newtonsoft.Json.8.0.3\lib\net45\Newtonsoft.Json.dll</HintPath>
       <Private>True</Private>
     </Reference>
-    <Reference Include="ProtoGeometry, Version=2.9.0.3837, Culture=neutral, processorArchitecture=AMD64">
-      <HintPath>..\packages\DynamoVisualProgramming.LibG_225_0_0.2.9.0.3837\lib\net48\ProtoGeometry.dll</HintPath>
+    <Reference Include="ProtoGeometry, Version=2.11.0.4156, Culture=neutral, processorArchitecture=AMD64">
+      <HintPath>..\packages\DynamoVisualProgramming.LibG_225_0_0.2.11.0.4156\lib\net48\ProtoGeometry.dll</HintPath>
```
aparajit-pratap, line 87: Not that it matters much but we try to match this version with that for ASM227 (or whichever is the current latest).

mjkkirschner: Sorry @aparajit-pratap I'm not sure I get it - LibG227 nuget had a different version number, and was pointing to a newer commit, I wanted all nugets to contain the same code... 

atleast I think thats what I did :) - grabbed everything on nuget from yesterday - 227 had an extra commit I believe.

aparajit-pratap: What I'm saying is that we should reference the protogeometry version that's coming from ASM227 here like we're doing in all other projects.

mjkkirschner: ahh, okay, will make the change.

mjkkirschner:
@aparajit-pratap - I'm rerunning the tests, not sure why these 2 tests failed...

mjkkirschner:
@mjkkirschner these tests fail locally, as well, very strange, not sure how modifying libG could effect them - will investigate.

aparajit-pratap:
I guess we shouldn't wait that long before pulling libg nugets into Dynamo. Especially in this case, where we have some big changes in libg, it could be hard to track down which libg change is actually causing the regression.

aparajit-pratap (approved):

mjkkirschner:
this passed here:
https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/2095/

I'll merge after the final test job completes.

PR #11513, opened 2021-02-22
Add new configuration option to disable ADP

mjkkirschner:
### Purpose

* Adds a configuration option for clients to use to disable ADP for dynamo and any other API access within the process.
* Does some simple checks for invalid configuration options, as well as asserting that the disable call actually worked.
* updates ADPSDK to 3.x and updates analytics.net to 2.11.1 binaries.

### TODO:
~- [ ] add test~  I found it's difficult to add a test here without refactoring the analyticsServices classes because static types cannot be mocked, and during testing we may not actually have the ADP core binaries around. Because we have tests in the ADP project, and the test here would basically just be asserting the configuration bool is set correctly, I'm thinking it's not worth it - open to suggestions though.
- [x] adpwrapper binary must be updated in this pr

### Declarations


src/DynamoCore/Logging/AnalyticsService.cs
```diff
@@ -17,9 +17,24 @@ class AnalyticsService
...
-        internal static void Start(DynamoModel model)
+        /// <param name="disableADPForProcessLifetime">Pass true to disable ADP for the lifetime of the Dynamo or host process</param>
+        internal static void Start(DynamoModel model, bool disableADPForProcessLifetime)
         {
             var client = new DynamoAnalyticsClient(model);
+            if (disableADPForProcessLifetime)
+            {
+                //if this returns false something has gone wrong.
+                //the client requested ADP be disabled, but we cannot disable it.
+                if (!ADPProcessSession.DisableADPForProcessLifetime())
```
mjkkirschner, line 28: build will fail because this API doesn't exist until we merge the Analytics net wrapper PR

src/DynamoCore/Logging/AnalyticsService.cs
```diff
@@ -17,8 +17,33 @@ class AnalyticsService
...
+                {
+                    //TODO consider throwing instead - that will cause a crash.
+                    model.Logger.LogNotification("Dynamo",
+                        "Dynamo Startup Error",
+                        "Analytics could not be disabled",
+                        "Dynamo was started with configuration requesting ADP be disabled - but ADP could not be disabled.");
+                }
+            }
+
+            if (isHeadless)
```
mjkkirschner, line 43: I modified how late we bail out if `isHeadless` is true because some clients need to disable ADP and still use Dynamo as a headless runner.

mjkkirschner:
@pinzart90 @aparajit-pratap  - please see my note above on difficulties testing this - one approach I do think would work is to create an internal enum to record the state of the disable call - something like:

enum disableResult { success, failure, not_attempted} - then check this from a test, the issue is that during a test we can't guarantee we will have access to adp core, but we might still want to make sure the disable call was attempted.

what do you think?

src/DynamoCore/Logging/AnalyticsService.cs
```diff
@@ -17,8 +17,33 @@ class AnalyticsService
         /// the Analytics service and application life cycle start is tracked.
         /// </summary>
         /// <param name="model">DynamoModel</param>
-        internal static void Start(DynamoModel model)
+        /// <param name="disableADPForProcessLifetime">Pass true to disable ADP for the lifetime of the Dynamo or host process</param>
```
pinzart90, line 20: `disableADP` + descriptive might be easier on the eyes :)

mjkkirschner: @pinzart90 do you mean change the name to `disableADP` and provide a good description?

pinzart90: yes...just a thought

src/DynamoCore/Logging/AnalyticsService.cs
```diff
@@ -10,15 +10,40 @@ namespace Dynamo.Logging
...
-        private static IAnalyticsUI adpAnalyticsUI = new ADPAnalyticsUI();
+        private static ADPAnalyticsUI adpAnalyticsUI = new ADPAnalyticsUI();
 
         /// <summary>
         /// Starts the client when DynamoModel is created. This method initializes
         /// the Analytics service and application life cycle start is tracked.
         /// </summary>
         /// <param name="model">DynamoModel</param>
-        internal static void Start(DynamoModel model)
+        /// <param name="disableADP">Pass true to disable ADP for the lifetime of the Dynamo or host process</param>
```
pinzart90, line 20: sorry to pester on this...but I just saw you added a new serializable field called `disableADPForProcess`
I think the param here should be consistent with that. So either disableADPForProcess or disableADPin both places...
Asides from that...PR looks good

pinzart90 (approved):
See comment ..then looks good

src/DynamoCore/Logging/AnalyticsService.cs
mjkkirschner: @pinzart90 - what makes this field serializable? 
Also to be clear I added a private field to dynamoModel - and a public property to the startup configuration (which it not serialized)

I don't think either of them is serialized anywhere - but I can make them all match if you think the shorter, less specific name is better, and hopefully clients read the docs.

pinzart90: Oh. It's private ..ok then.
I think it is better if they match...easier on debugging...
Since everything is internal...naming does not matter that much...
Any way you wish. I do not want to slow this PR down ...

pinzart90 (approved):

src/DynamoCore/Logging/AnalyticsService.cs
QilongTang: Can you remind me, the `isHeadless` is set to true only when using command line, right? For Dynamo Player and GD, analytics is on because we are using Dynamo as a server.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -668,6 +674,20 @@ protected DynamoModel(IStartConfiguration config)
             {
                 // Do nothing for now
             }
+
+            // these configuration options are incompatible, one requires loading ADP binaries
+            // the other depends on not loading those same binaries.
+
+            if (areAnalyticsDisabledFromConfig && disableADP)
+            {
+                throw new ConfigurationException("Incompatible configuration: could not start Dynamo with both [Analytics disabled] and [ADP disabled] config options enabled");
```
QilongTang, line 683: 👍🏻

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -668,6 +674,20 @@ protected DynamoModel(IStartConfiguration config)
...
+            }
+
+            if(IsTestMode && disableADP)
+            {
+                this.Logger.Log("Incompatible configuration: [IsTestMode] and [ADP disabled] ");
```
QilongTang, line 688: what does this case mean?

mjkkirschner: this is the case where testmode is active (during most tests) and some test might try to disable ADP - but we can't guarantee that we have access to the core sdks in different testing scenarios, and I could not find a nice way to mock this dependency, so thought it would be valuable to log the case.

QilongTang: That is exactly what I was going to mention. We always bail out during test mode so I dont think ADP will be loaded at all? https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Models/DynamoModel.cs#L1429

mjkkirschner: we don't *need* it - but I think it might help avoid wasted time. It would be nice if Dynamo provided feedback on incompatible options since to a client it's not obvious what headless and testmode actually control.

src/DynamoCore/Logging/AnalyticsService.cs
mjkkirschner: I believe player does not use headless, I am not sure about GD.

src/DynamoCore/Models/DynamoModel.cs
QilongTang: Thanks @mjkkirschner If that's the case, it seems this warning can be moved into beginning of `AnalyticsService.Start()`

mjkkirschner: yes, can do.

QilongTang:
Thank you. Looks good to me

aparajit-pratap:
Why are these empty files added:
extern/Analytics.NET/AdpSDKCSharpWrapper.dll, ...

mjkkirschner:
@aparajit-pratap they are not empty or added - they are updated binaries:
1. new adpsdk
2. our updated wrapper 

check the history of the files, it shows "empty file" since it was added, not sure why.

PR #11519, opened 2021-02-25
Do not set execution hint on undo move

mmisol:
### Purpose

There were a couple of issues that were causing a node to be set to be
executed on undoing an action that consisted of just a move:

1. The 'UsingDefaultValue' property was triggering a property changed
even if the deserialized value was the same as the current. This was
fixed by checking if the value is the same and not doing anything in
that case.

2. Even if the 'IsFrozen' property for the current node had not changed
an undo would mark the current and all downstream nodes as if they were
modified. This was fixed by relying on the 'IsFrozen' property
setter, which correctly handles the cases when the previous should be
done.

### Declarations

### DIFF
--- a/src/DynamoCore/Graph/Nodes/NodeModel.cs
+++ b/src/DynamoCore/Graph/Nodes/NodeModel.cs
@@ -2234,7 +2234,7 @@ namespace Dynamo.Graph.Nodes
             argumentLacing = helper.ReadEnum("lacing", LacingStrategy.Disabled);
             IsSetAsInput = helper.ReadBoolean("isSelectedInput", false);
             IsSetAsOutput = helper.ReadBoolean("isSelectedOutput", false);
-            isFrozenExplicitly = helper.ReadBoolean("IsFrozen", false);
+            IsFrozen = helper.ReadBoolean("IsFrozen", false);
             PreviewPinned = helper.ReadBoolean("isPinned", false);
 
             var portInfoProcessed = new HashSet<int>();
@@ -2306,13 +2306,6 @@ namespace Dynamo.Graph.Nodes
                 RaisePropertyChanged(nameof(ArgumentLacing));
                 RaisePropertyChanged(nameof(IsVisible));
                 RaisePropertyChanged(nameof(DisplayLabels));
-                RaisePropertyChanged(nameof(IsSetAsInput));
-                RaisePropertyChanged(nameof(IsSetAsOutput));
-                //we need to modify the downstream nodes manually in case the
-                //undo is for toggling freeze. This is ONLY modifying the execution hint.
-                // this does not run the graph.
-                RaisePropertyChanged(nameof(IsFrozen));
-                MarkDownStreamNodesAsModified(this);
 
                 // Notify listeners that the position of the node has changed,
                 // then all connected connectors will also redraw themselves.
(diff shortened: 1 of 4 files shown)


src/DynamoCore/Graph/Nodes/PortModel.cs
```diff
@@ -189,9 +189,12 @@ public bool UsingDefaultValue
...
             set
             {
-                usingDefaultValue = value;
-                RaisePropertyChanged("UsingDefaultValue");
-                RaisePropertyChanged("ToolTip");
+                if (usingDefaultValue != value)
+                {
+                    usingDefaultValue = value;
+                    RaisePropertyChanged("UsingDefaultValue");
+                    RaisePropertyChanged("ToolTip");
```
QilongTang, line 196: I like the fix here. Also curious does similar problem not exist for other properties, e.g. X, Y etc?

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -347,7 +347,14 @@ protected override void DeserializeCore(XmlElement nodeElement, SaveContext cont
...
             if (engineNode != null)
             {
-                this.Engine = (PythonEngineVersion)Enum.Parse(typeof(PythonEngineVersion),engineNode.InnerText);
+                var oldEngine = Engine;
+                Engine = (PythonEngineVersion)Enum.Parse(typeof(PythonEngineVersion), engineNode.InnerText);
+                if (context == SaveContext.Undo && oldEngine != this.Engine)
+                {
+                    // For Python nodes, changing the Engine property does not set the node Modified flag.
+                    // This is done externally in all event handlers, so for Undo we do it here.
+                    OnNodeModified();
```
QilongTang, line 356: 👍🏻

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -2306,13 +2306,6 @@ protected override void DeserializeCore(XmlElement nodeElement, SaveContext cont
                 RaisePropertyChanged(nameof(ArgumentLacing));
                 RaisePropertyChanged(nameof(IsVisible));
                 RaisePropertyChanged(nameof(DisplayLabels));
-                RaisePropertyChanged(nameof(IsSetAsInput));
-                RaisePropertyChanged(nameof(IsSetAsOutput));
-                //we need to modify the downstream nodes manually in case the
-                //undo is for toggling freeze. This is ONLY modifying the execution hint.
-                // this does not run the graph.
-                RaisePropertyChanged(nameof(IsFrozen));
-                MarkDownStreamNodesAsModified(this);
```
QilongTang, line 2315: Hi @mmisol Did you test that downstream node can be also reverted back to the previous state?

QilongTang (approved):
LGTM with two questions

src/DynamoCore/Graph/Nodes/PortModel.cs
mmisol: Yes, this pattern is used in most places. For `X` and `Y`, I noticed they are deprecated and there is a `Position` which is a way to update both. Maybe that's the reason why it's different.

src/DynamoCore/Graph/Nodes/NodeModel.cs
mmisol: I'm not sure I follow. Could you mention the steps of the test?

QilongTang: Sure, if a node is frozen, the downstream nodes are also frozen. The original code seems taking care of correctly marking downstream nodes when undo freezing node as well. Just want to make sure the behavior is the same.

mmisol: Oh ok. Yes, I tested that and it works

PR #11530, opened 2021-03-03
set IsFirstRun to false if hide report options is true

mjkkirschner:
### Purpose

when `hide report options` is enabled, the prompt to request user analytics is also hidden  - this function has the side effect that `IsFirstRun` is set to false so the gallery is not shown again.

If  `hide report options` is enabled we set `IsFirstRun` to false after this check, but before trying to raise a gallery window.

### TODO
- [x] tests

### Declarations

### DIFF
--- a/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
+++ b/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
@@ -909,16 +909,9 @@ namespace Dynamo.Controls
             //Backing up IsFirstRun to determine whether to show Gallery
             var isFirstRun = dynamoViewModel.Model.PreferenceSettings.IsFirstRun;
 
-            if (!dynamoViewModel.HideReportOptions)
-            {
-                // If first run, Collect Info Prompt will appear
-                UsageReportingManager.Instance.CheckIsFirstRun(this, dynamoViewModel.BrandingResourceProvider);
-            }
-            else
-            {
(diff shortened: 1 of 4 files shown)


src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -914,6 +914,11 @@ private void DynamoView_Loaded(object sender, EventArgs e)
                 // If first run, Collect Info Prompt will appear
                 UsageReportingManager.Instance.CheckIsFirstRun(this, dynamoViewModel.BrandingResourceProvider);
```
QilongTang, line 915: It's a bit hard to understand from first glance, to me it seems we may want to fold the `dynamoViewModel.HideReportOptions` check into `CheckIsFirstRun()` at https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/Services/UsageReportingManager.cs#L154

mjkkirschner: @QilongTang let me give that a shot.

mjkkirschner:
@QilongTang PTAL

QilongTang (approved):

QilongTang:
@mjkkirschner I think this one will need cherry-pick right?

mjkkirschner:
doing it now 😉

PR #11536, opened 2021-03-08
Dyn 3456 - Explicit NodeModel nodes deserialize port tooltips and descriptions instead of loading them at runtime from type.

mjkkirschner:
### background

FormIt team reported that upon updating the tooltips for one of their existing nodemodel/UI nodes - the tooltips of existing graphs were still displaying the old value.

I will refer to tooltips and descriptions interchangeably.

The issue is that port tooltips and port names are deserialized - this can cause the following bugs:

1. tooltips that are updated in the node class definition are not updated in existing graphs.
2. tooltips from one localized language will not appear in a different localized language if an existing graph is transferred between two locales. This may not happen very often.

I was hopeful this bug would be easy to fix, similar to: 
https://jira.autodesk.com/browse/DYN-2002, for node description toolitps, but it's not as simple.
In the case of node descriptions - the description is always set on the specific NodeModel type via an attribute
`[NodeDescription(somestring)]` - this makes it very simple to lookup the attribute instead of using the deserialized description.

Unfortunately, that is not how *all* port descriptions and names are added. Instead in most implementations of NodeModel nodes, our samples, and user code, the ports are only instantiated at runtime for new nodes in the constructors, and instead are deserialized from the file directly in the json constructors.

In many cases there is no attribute to lookup the port information - it's just a call in the constructor to add a new port.

For base `NodeMode/UI` nodes a viable strategy is to request clients to update their code to supply the `[PortDescription]` and `[PortName]` attributes on their nodeModel, these exist already! yay!

For `VariableInputNodes` this is not possible for *all* ports, some are static, and this may work for those ports. Some are variable and can be added and deleted at runtime - for this case, a virtual method `GetToolTip(index)` and `GetName(index)` exist on `VariableInputPorts`

### What does this do:

~~I believe this PR contains the best experience for clients - at the risk of being difficult to reason about.~~

It does the following: 
* **A** - if port name and tooltip attributes exist, it loads those and sets those on the deserialized ports
~~* **B** -  if the number of attributes does not match the number of ports that have been deserialized, then we call the constructor of the node type we are deserializing, grab its ports, and store those names and tooltips on the deserialized ports. I think this might have performance and memory consequences when loading graphs with many nodeModel nodes that do not have attributes.~~
~~* **C**- If we encounter a `variableInput node` we call the `GetPortToolTip` and `GetPortName` methods to get the appropriate port names given the index of the port - we skip any ports that the constructor set.~~

~~### request for feedback 
One idea I would like feedback on is to simplify this solution - by removing the fallback logic of steps B and C and only handling the case where the attributes exist - this is more work for clients, they need to add attributes - but I think it's less of a memory or performance risk. It does not handle the variable Input Node case though, which we would need to address as a followup.~~

### Declarations


src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -1083,6 +1084,103 @@ protected NodeModel(IEnumerable<PortModel> inPorts, IEnumerable<PortModel> outPo
...
+                //modify the data coming from deserialization - risk is that these nodes may 
+                //create ports on the fly or do other transformations - just ask the node authors
+                //to handle this inside the json constructor - we've done the same for some variable input nodes.
+
+                // create a temporary node and extract data from it.
+                //TODO it would be great to avoid doing this for node types that we've already done it for...
+                //IE static or shared cache between node constructors - dumped/recreated after workspace load.
+
+                //TODO possible to create this node in another appdomain or somehow isolated?
+                var tempNode = this.GetType().GetConstructor(Type.EmptyTypes)?.Invoke(null) as NodeModel;
```
pinzart90, line 1124: How about using `Activator.CreateInstance` ?

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -1083,6 +1084,103 @@ protected NodeModel(IEnumerable<PortModel> inPorts, IEnumerable<PortModel> outPo
...
+            //try to get portData from attributes on the nodeModel type.
+            //to assign the correct tooltips.
+            //This might fail because some nodes don't use attributes to define 
+            //port data and instead add them using PortModel constructors.
+            var inportDatas = GetPortDataFromAttributes(PortType.Input).ToList();
+            var outportDatas = GetPortDataFromAttributes(PortType.Output).ToList();
+
+            // if the above attempt failed, for example because the node type does not have any port data attributes
+            // then try other approaches
+            if (inportDatas.Count() != inPorts.Count() || outportDatas.Count() != outPorts.Count())
```
pinzart90, line 1111: so if port data (tooltip/description) is not supplied for InputPorts but is supplied for OutputPorts...then we try fallback for OutputPorts too ?
is it possible to do it only for the ports (input/output) that do not have the required data ?

src/DynamoCore/Graph/Nodes/NodeModel.cs
pinzart90: this might throw MissingMethodException if no matching constructor exists

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -1083,6 +1084,103 @@ protected NodeModel(IEnumerable<PortModel> inPorts, IEnumerable<PortModel> outPo
...
+                        var data = thisAsVarInput.GetInputTooltipAndName(i);
+                        var newPortData = new PortData(data.name, data.tooltip);
+                        //add a new portdata for the existing port we deserialized.
+                        if (i >= inportDatas.Count())
+                        {
+                            inportDatas.Add(newPortData);
+                        }
+                        else
+                        {
+                            inportDatas[i] = newPortData;
```
pinzart90, line 1150: will this `else` block ever be hit ?

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -1083,6 +1084,103 @@ protected NodeModel(IEnumerable<PortModel> inPorts, IEnumerable<PortModel> outPo
...
+                if (outportDatas.Count > i)
+                {
+                    var portData = outportDatas.ElementAt(i);
+                    var port = outPorts.ElementAt(i);
+                    port.Name = portData.Name == string.Empty ? port.Name : portData.Name;
+                    port.ToolTip = portData.ToolTipString == string.Empty ? port.ToolTip : portData.ToolTipString;
+                }
+            }
+
+            for (var i = 0; i < inPorts.Count(); i++)
```
pinzart90, line 1170: how about `for (var i = 0; i < inPorts.Count() && i < inportDatas.Count(); i++)`  without the bounds check ? ...same for `outPorts` above

src/DynamoCore/Graph/Nodes/NodeModel.cs
pinzart90: now that I think about it, the targeted constructor could throw ...even d4r has node constructors  that can throw

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -1049,6 +1049,7 @@ protected NodeModel(IEnumerable<PortModel> inPorts, IEnumerable<PortModel> outPo
         {
             inputNodes = new Dictionary<int, Tuple<int, NodeModel>>();
             outputNodes = new Dictionary<int, HashSet<Tuple<int, NodeModel>>>();
+            FindPortData(inPorts, outPorts);
```
pinzart90, line 1052: how was the tooltip info retrieved before this PR ?

mjkkirschner: tooltip and name info was retrieved directly from the deserialized ports from the .dyn or .dyf - which leads to the two bugs mentioned in the description.

mjkkirschner:
@pinzart thank you for the code review- what are your thoughts on: 
```
request for feedback
One idea I would like feedback on is to simplify this solution - by removing the fallback logic of steps B and C and only handling the case where the attributes exist - this is more work for clients, they need to add attributes - but I think it's less of a memory or performance risk. It does not handle the variable Input Node case though, which we would need to address as a followup.

```

mjkkirschner:
discussed with Tibi - I will be falling back to only performing operation A and filing two followups 
1. for the proper use of the attributes in our samples
2. better design unification of ways to describe ports for variable input nodes

src/Libraries/CoreNodeModels/CreateList.cs
```diff
@@ -14,6 +14,10 @@ namespace CoreNodeModels
     [NodeCategory(BuiltinNodeCategories.CORE_LISTS_CREATE)]
     [IsDesignScriptCompatible]
     [AlsoKnownAs("DSCoreNodesUI.CreateList", "List.Create")]
+    [InPortNames("item0")]
+    [InPortDescriptions(typeof(Resources),"CreateListPortDataIndex0ToolTip")]
+    [OutPortNames("list")]
+    [OutPortDescriptions(typeof(Resources), "CreateListPortDataResultToolTip")]
```
QilongTang, line 20: These changes are for?

mjkkirschner: they were part of implementing variable input node port data lookup, it won't be too useful in the PR's current state - it can be removed if you want to make this PR more clear -

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -1083,6 +1084,51 @@ protected NodeModel(IEnumerable<PortModel> inPorts, IEnumerable<PortModel> outPo
...
+            var outportDatas = GetPortDataFromAttributes(PortType.Output).ToList();
+
+            // if the above attempt failed, for example because the node type does not have any port data attributes
+            // then give up for all ports.
+            if (inportDatas.Count() != inPorts.Count() || outportDatas.Count() != outPorts.Count())
+            {
+                return;
+            }
+
+            for (var i = 0; i < outPorts.Count() && i < outportDatas.Count; i++)
```
aparajit-pratap, line 1112: At this point in time, won't the `outPorts `count and `outportDatas `count be the same?

src/DynamoCore/Graph/Nodes/NodeModel.cs
aparajit-pratap: Same question here.

test/DynamoCoreTests/SerializationTests.cs
```diff
@@ -686,6 +686,61 @@ public void NodeDescriptionDeserilizationTest()
...
+        [Test]
+        [Category("Failure")]
+        [Category("TechDebt")]
+        public void OutPortDescriptionDeserilizationTest_VariableInputNodes()
+        {
+            //similar to above test, uses a modified graph to assert port tooltips not deserialized when possible.
+            var testFile = Path.Combine(TestDirectory, @"core\serialization\PortTooltipDeserilizationTest.dyn");
+            OpenModel(testFile);
+            var variableInputNode = this.CurrentDynamoModel.CurrentWorkspace.Nodes.Where(x => x.GUID == new Guid("db8fd7b97be1413e91897316ae75b51a")).First();
+            Assert.AreEqual(variableInputNode.OutPorts.First().ToolTip, "Combined lists");
```
aparajit-pratap, line 712: Is this for a future task to handle tooltips for dynamically added ports?

src/DynamoCore/Graph/Nodes/NodeModel.cs
mjkkirschner: yes @aparajit-pratap you're right, now that we're bailing out if the counts don't match I should be able to simplify this.

test/DynamoCoreTests/SerializationTests.cs
mjkkirschner: yep exactly.

aparajit-pratap (approved):
Just one more question - is it possible to update the tooltips for only those ports that have attributes instead of bailing out completely even if they are specified for only a few of them?

mjkkirschner:
@aparajit-pratap, yes I think it's possible, though maybe not too useful. I think it would only happen in two cases:
1. A node author has only added attributes for some of the ports - seems unlikely for most nodes. (see case 2)
2. This is a variableInput node. where the node author has used attributes for the static ports.

Do you think it's worth looking into?

aparajit-pratap:
> @aparajit-pratap, yes I think it's possible, though maybe not too useful. I think it would only happen in two cases:
> 
> 1. A node author has only added attributes for some of the ports - seems unlikely for most nodes. (see case 2)
> 2. This is a variableInput node. where the node author has used attributes for the static ports.
> 
> Do you think it's worth looking into?

I'm ok with this PR as-is. It may not be worth looking into if significantly more complex.

pinzart90 (approved):

PR #11554, opened 2021-03-18
DYN-3457: Dynamo should alert the user if there are conflicts loading packages from different locations

sm6srw:
### Purpose

This pull request does:
* Standard Library is now the first directory in the list.
* ignore packages with invalid versions.
* load the first version of a package found regardless of version.
* throw a `LibraryLoadFailedException` for packages with invalid versions.
* throw a `LibraryLoadFailedException' when a duplicated package is found 
* throw a `LibraryLoadFailedException' when a duplicated new package is found 
* throw a `LibraryLoadFailedException' when a duplicated old package is found 

### Declarations


aparajit-pratap (commented):
I think the intention of this task was to make Dynamo scan all package locations to see if there are going to be potential package conflicts before loading any single one. With these changes, it looks like Dynamo is going to continue loading them one by one, then when it encounters a package conflict with a second location, take appropriate action but by then it's too late - at least one package would already have been loaded.

sm6srw:
> I think the intention of this task was to make Dynamo scan all package locations to see if there are going to be potential package conflicts before loading any single one. With these changes, it looks like Dynamo is going to continue loading them one by one, then when it encounters a package conflict with a second location, take appropriate action but by then it's too late - at least one package would already have been loaded.

@aparajit-pratap The intention of this code is to always load the newest version of a package when there are version conflicts. The existing implementation is already doing a pre-scan of all packages but is only removing duplicated packages based on the package name. I extend that logic to also include the version. The actual loading happens later based on the cleaned up LocalPackages property. Lets talk at standup.

aparajit-pratap:
> @aparajit-pratap The intention of this code is to always load the newest version of a package when there are version conflicts. The existing implementation is already doing a pre-scan of all packages but is only removing duplicated packages based on the package name. I extend that logic to also include the version. The actual loading happens later based on the cleaned up LocalPackages property. Lets talk at standup.

@sm6srw if that's the case, then it's good but I'd like to test your PR once to convince myself that it works especially with the std library.

sm6srw:
> > @aparajit-pratap The intention of this code is to always load the newest version of a package when there are version conflicts. The existing implementation is already doing a pre-scan of all packages but is only removing duplicated packages based on the package name. I extend that logic to also include the version. The actual loading happens later based on the cleaned up LocalPackages property. Lets talk at standup.
> 
> @sm6srw if that's the case, then it's good but I'd like to test your PR once to convince myself that it works especially with the std library.

Sure! Go ahead.

src/DynamoCore/Properties/Resources.resx
```diff
@@ -735,4 +735,10 @@ Parameter name: {0}</value>
     <value>The current user, '{0}', is not a maintainer of the package '{1}'.</value>
     <comment>{0} = user name (i.e. 'DynamoTeam'), {1} = package name (i.e. 'Clockwork for Dynamo 1.x')</comment>
   </data>
+  <data name="DuplicatedOlderPackage" xml:space="preserve">
+    <value>An older version of the package called {0} version {2} was found at {1} with version {3}. Ignoring it.</value>
+  </data>
+  <data name="InvalidPackageVersion" xml:space="preserve">
+    <value>The version of the package called {0} found at {1} is invalid (version: "{2}"). Ignoring it.</value>
+  </data>
```
pinzart90, line 743: both Resources.en-US.resx  and Resources.resx are needed ?

src/DynamoCore/Properties/Resources.en-US.resx
```diff
@@ -735,4 +735,10 @@ Parameter name: {0}</value>
     <value>The current user, '{0}', is not a maintainer of the package '{1}'.</value>
     <comment>{0} = user name (i.e. 'DynamoTeam'), {1} = package name (i.e. 'Clockwork for Dynamo 1.x')</comment>
   </data>
+  <data name="DuplicatedOlderPackage" xml:space="preserve">
+    <value>An older version of the package called {0} version {2} was found at {1} with version {3}. Ignoring it.</value>
```
pinzart90, line 739: `Ignoring it.` is kind of unclear to me...Which one is ignored...and I guess "to ignore" in this context is something obvious to the user...

src/DynamoCore/Properties/Resources.resx
sm6srw: As far as I know, yes. @mjkkirschner?

src/DynamoCore/Properties/Resources.en-US.resx
sm6srw: What about changing `Ignoring it.` to `The older version has been ignored.`?

pinzart90: I think that is better ...

src/DynamoPackages/PackageLoader.cs
```diff
@@ -438,6 +467,25 @@ public Package ScanPackageDirectory(string directory, bool checkCertificates)
...
+        /// </summary>
+        /// <param name="version">the version string</param>
+        /// <param name="name">name of the package</param>
+        /// <param name="directory">package directory</param>
+        /// <returns>Returns a valid Version</returns>
+        private Version CheckAndGetPackageVersion(string version, string name, string directory)
+        {
+            try
+            {
+                return new Version(version);
```
pinzart90, line 481: There is an `VersionUtilities.PartialParse()` ..not sure if that does anything else ...

sm6srw: That method is still using the Version class. We might want to use it if we decide that we don't want to do a full compare of the package version.  Package version `1.0.0.1` is now newer than version `1.0.0.0`. I have noticed that the packages I have looked at usually only defines the version as `1.0.0`. That will automatically turn into `1.0.0.0` before comparison in my current implementation. Have we defined somewhere how we deal with packages versions? E.g. should the last number always be ignored?

src/DynamoCore/Properties/Resources.resx
mjkkirschner: yes

src/DynamoPackages/PackageLoader.cs
mjkkirschner: I'm confused as is often the case - AFAIK when you upload a package you can only specify major,minor,and revision - not build.

see:
image

I don't think we can really encounter a package that sets the last version number unless the header is modified manually.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -418,16 +418,45 @@ public Package ScanPackageDirectory(string directory, bool checkCertificates)
...
                 {
-                    CheckPackageNodeLibraryCertificates(directory, discoveredPkg);
+                    CheckPackageNodeLibraryCertificates(directory, discoveredPackage);
                 }
 
-                // prevent duplicates
-                if (LocalPackages.All(pkg => pkg.Name != discoveredPkg.Name))
+                var discoveredVersion = CheckAndGetPackageVersion(discoveredPackage.VersionName, discoveredPackage.Name, discoveredPackage.RootDirectory);
+
+                var existingPackage = LocalPackages.FirstOrDefault(package => package.Name == discoveredPackage.Name);
```
mjkkirschner, line 426: @sm6srw will this necessarily discover ALL conflicts, what if there are 3 versions of the package?

mjkkirschner: also - isn't `LocalPackages` only populated once a package is actually loaded?

src/DynamoPackages/PackageLoader.cs
```diff
@@ -418,16 +418,45 @@ public Package ScanPackageDirectory(string directory, bool checkCertificates)
...
+                // Is the existing version newer?
+                if (existingVersion > discoveredVersion)
+                {
+                    // Newer version already exist
+                    Log(String.Format(Properties.Resources.DuplicatedOlderPackage, discoveredPackage.Name, discoveredPackage.RootDirectory, existingVersion.ToString(), discoveredVersion.ToString()), WarningLevel.Moderate);
+                    return null;
+                }
+
+                // Older version exist, replace with newer version
+                Remove(existingPackage);
```
mjkkirschner, line 454: if `existingPackage` is already installed/loaded then this won't do anything during this session - right?

src/DynamoPackages/PackageLoader.cs
sm6srw: Thanks, I think we should ignore the last version number then and use the utility method we have.

src/DynamoPackages/PackageLoader.cs
sm6srw: Yes, it will find all conflicts as it make sure that a package can only exist once in LocalPackages. A package will be added the first time and then replaced with a new version or ignored if older going forward.

src/DynamoPackages/PackageLoader.cs
sm6srw: Correct.

src/DynamoPackages/PackageLoader.cs
sm6srw: Lock at the `LoadAll` method. It first scan all package directories and populate `LocalPackages`. It then uses `LocalPackages` for loading all packages.

src/DynamoPackages/PackageLoader.cs
sm6srw: This happens when you add/remove package directories. I think we should do something about it but it feels like it should be a separate task.

mjkkirschner: okay, so this workflow only works well during first startup of Dynamo after this PR? Thats okay with me as long as we file a followup

mjkkirschner:
@sm6srw it doesn't look like on first glance that the tests you have added are testing the package conflicts? Am I missing it? Also does it make sense to add a failing test for the workflow when a new package dir is added?

src/DynamoPackages/PackageLoader.cs
sm6srw: I will file something.

sm6srw:
> @sm6srw it doesn't look like on first glance that the tests you have added are testing the package conflicts? Am I missing it? Also does it make sense to add a failing test for the workflow when a new package dir is added?

The last two tests are testing package conflicts (The involved packages have the same name but different versions, I am using the description field as a way to separate in addition to the version number and that is probably what's confusing you). Yes, adding a couple of tests for covering adding/removing directories is a good idea.

src/DynamoPackages/PackageLoader.cs
sm6srw: This was actually easier to fix than I expected. I have added the case when a new version of a package is found but the old version is already loaded. This will show up as an notification for now.

pinzart90: LGTM

pinzart90 (approved):

src/DynamoPackages/PackageLoader.cs
sm6srw: @pinzart90 It happens when everything reloads after adding a new package directory.

src/DynamoCore/Properties/Resources.en-US.resx
```diff
@@ -735,4 +735,13 @@ Parameter name: {0}</value>
...
     <comment>{0} = user name (i.e. 'DynamoTeam'), {1} = package name (i.e. 'Clockwork for Dynamo 1.x')</comment>
   </data>
+  <data name="DuplicatedOlderPackage" xml:space="preserve">
+    <value>An older version of the package called {0} version {2} was found at {1} with version {3}. The older version has been ignored.</value>
+  </data>
+  <data name="InvalidPackageVersion" xml:space="preserve">
+    <value>The version of the package called {0} found at {1} is invalid (version: "{2}"). Ignoring it.</value>
+  </data>
+  <data name="DuplicatedPackageAlreadyLoaded" xml:space="preserve">
+    <value>A newer version of the package called {0} version {2} was found at {1} with version {3}. The old version of the package is already loaded. Please restart Dynamo for loading the new version of the package.</value>
```
pinzart90, line 745: is the application always "Dynamo" ? ..in ` Please restart Dynamo for`

sm6srw: I think so but it is a good point.

mjkkirschner: yeah, what if we're in Revit - you actually need to reboot Revit...

sm6srw: Ok, I was following the messaging that we have for package manager that currently says "Restart Dynamo to complete the uninstall.". I can change to "Please restart for loading"?

src/DynamoPackages/PackageLoader.cs
```diff
@@ -418,16 +418,51 @@ public Package ScanPackageDirectory(string directory, bool checkCertificates)
...
+                // Is the existing version newer?
+                if (existingVersion > discoveredVersion)
+                {
+                    // Newer version already exist
+                    Log(String.Format(Properties.Resources.DuplicatedOlderPackage, discoveredPackage.Name, discoveredPackage.RootDirectory, existingVersion.ToString(), discoveredVersion.ToString()), WarningLevel.Moderate);
+                    return null;
+                }
+
+                // Is the existing package already loaded?
+                if (existingPackage.Loaded)
```
aparajit-pratap, line 454: What does this case mean?

src/DynamoPackages/PackageLoader.cs
```diff
@@ -418,16 +418,51 @@ public Package ScanPackageDirectory(string directory, bool checkCertificates)
...
+
+                var existingVersion = CheckAndGetPackageVersion(existingPackage.VersionName, existingPackage.Name, existingPackage.RootDirectory);
+
+                // Is this a duplicated package?
+                if (existingVersion == discoveredVersion)
                 {
-                    Add(discoveredPkg);
-                    return discoveredPkg; // success
+                    // Duplicated with the same version
+                    throw new LibraryLoadFailedException(directory, String.Format(Properties.Resources.DulicatedPackage, discoveredPackage.Name, discoveredPackage.RootDirectory));
```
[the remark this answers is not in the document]
sm6srw: Sure..

src/DynamoPackages/PackageLoader.cs
sm6srw: I will add more comments.

aparajit-pratap: Why do we have a separate code path for this case? Shouldn't we continue to line 475 even for the case where the user has added a new package path?

sm6srw:
Redesigned version where the first found package always wins but with enhanced notifications. PTAL!

src/DynamoPackages/PackageLoader.cs
```diff
@@ -71,7 +71,7 @@ public IEnumerable<IExtension> RequestedExtensions
         private readonly List<string> packagesDirectories = new List<string>();
         public string DefaultPackagesDirectory
```
mjkkirschner, line 72: hmm - I'm thinking this could start gettng a bit confusing, could we add a /// summary to this property to explain what we mean by default?

test/pkgs/BadVersion/PackageWithBadVersion/pkg.json
```diff
@@ -0,0 +1 @@
+{"license":"","file_hash":null,"name":"Package","version":"","description":"Bad package","group":"","keywords":null,"dependencies":[],"contents":"","engine_version":"2.1.0.7840","engine":"dynamo","engine_metadata":"","site_url":"","repository_url":"","contains_binaries":true,"node_libraries":["Package, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null"]}
```
mjkkirschner, line 1: what is in the package.dll?

mjkkirschner (approved):

src/DynamoPackages/PackageLoader.cs
sm6srw: Will do.

test/pkgs/BadVersion/PackageWithBadVersion/pkg.json
sm6srw: Just a dummy Hello  implementation.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -69,9 +69,13 @@ public IEnumerable<IExtension> RequestedExtensions
...
 
         private readonly List<string> packagesDirectories = new List<string>();
+
+        // Returns the default package directory where new packages will be installed
+        // The default package directory is currently the second entry in the list
+        // The first entry is the standard library
         public string DefaultPackagesDirectory
         {
-            get { return packagesDirectories[0]; }
+            get { return packagesDirectories[1]; }
```
sm6srw, line 78: We need to do something about this property but I think that should be part of @mjkkirschner's PR.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -418,16 +422,50 @@ public Package ScanPackageDirectory(string directory, bool checkCertificates)
...
+                    throw new LibraryLoadFailedException(directory, String.Format(Properties.Resources.DuplicatedOlderPackage,
+                            existingPackage.Name,
+                            discoveredPackage.RootDirectory,
+                            existingVersion.ToString(),
+                            discoveredVersion.ToString()));
                 }
-                throw new LibraryLoadFailedException(directory, String.Format(Properties.Resources.DulicatedPackage, discoveredPkg.Name, discoveredPkg.RootDirectory));
+
+                // Newer version found, show notification.
+                throw new LibraryLoadFailedException(directory, String.Format(Properties.Resources.DuplicatedNewerPackage,
```
aparajit-pratap, line 464: I still don't understand this case. Are we rejecting the new version over here and keeping the old one? If so, why?

sm6srw: Yes, this is the where we keep the old one and reject the new one. This was the major change I made after the feedback yesterday. Instead of using the new version we just inform the user that we have found a newer version that we will ignore because we have already found another version of the package. That should give them enough information so they can decide  what directory order they want.

aparajit-pratap: How about raising the existing `mark for uninstall` dialog here asking the user if they'd like to mark the existing package for uninstall and install the newer version upon restart or to simply ignore the new version?

aparajit-pratap: Or does that not make sense at this stage?

sm6srw: I think we should wait with that for now and add that as a separate task if we need it.

pinzart90: I agree...save it for later. Adding msg boxes when starting up dynamo could be very impactful to UX.

aparajit-pratap (approved):

src/DynamoPackages/PackageLoader.cs
aparajit-pratap: That's okay, but just to clarify, I don't think this case is at startup, it's at the time you add a package path if I'm not wrong.

sm6srw: It can happen at startup. Package X version 1.0.0 is installed in standard library. User download version 2.0.0. The user will get the message above every time they start after that as long as the standard library is first in the directory list. That is one of those cases when the user can't uninstall the old package because most of them will not have write access to the standard library. So lets see how it plays out and go from there.

aparajit-pratap: I doubt there's gonna be a case where we deliver a package in the std lib with a version that's older than that available on the PM, since that's the decision that we're making as a team, but I agree, it's not impossible. It could happen with host integrator packages, etc.

mjkkirschner: I can imagine we might create a new version of `meshtoolkit` say for release to PM to get early feedback before incorporation into std.lib.

aparajit-pratap: 🙏

src/DynamoPackages/PackageLoader.cs
mjkkirschner: @sm6srw what were you thinking 😉 ?

PR #11566, opened 2021-03-23
Add internal preferences for disabling loading packages from std lib and custom paths

mjkkirschner:
### Purpose

As part of the effort around std library package loading we decided to add an escape hatch to disable the std library. This can be useful for
* development
* to avoid package conflicts
* to harden security posture in certain scenarios 

At the same time as it's related I added a similar internal preference to disable loading packages from custom package paths. These preferences are bools that work independently.

As part of this work I needed a signed package to test loading packages from the standard library location - so I setup this repo:
https://git.autodesk.com/Dynamo/SignedDynamoTestingPackages

it is built manually, and creates 2 packages that it zips up along with the build:
* signedpackage
* signedpackage2

signedpackage is a package that @saintentropy created when testing the cert check workflow.
signedpackage2 is used for tests in this pr.

The package is manually copied to the test/ folder.

Tests are added for each preference. 

At some point, we may expose these options in the UI, but no plans at this moment. We should document them, but I'm not sure where / how yet @Amoursol @QilongTang 

### Declarations


src/DynamoPackages/PackageLoader.cs
```diff
@@ -76,30 +78,53 @@ public string DefaultPackagesDirectory
 
         private readonly List<string> packagesDirectoriesToVerifyCertificates = new List<string>();
 
-        // The standard library directory is located in the same directory as the DynamoPackages.dll
-        internal string StandardLibraryDirectory =>
```
mjkkirschner, line 80: sorry @sm6srw I had to mess with your elegant getter to be able test this.

src/DynamoCore/Configuration/IPreferences.cs
```diff
@@ -166,6 +166,22 @@ public interface IRenderPrecisionPreference
...
+    {
+        /// <summary>
+        /// If enabled Dynamo Standard Library packages will not be loaded.
+        /// </summary>
+        bool DisableStandardLibrary { get; set; }
+        /// <summary>
+        /// If enabled user's custom package locations will not be loaded.
+        /// </summary>
+        bool DisableCustomPackageLocations { get; set; }
+    }
```
pinzart90, line 183: Could these 2 bools be turned into a single list like `IEnumerable<string> DisablePackageLocations ` ?
API user could add std lib location and customPackagesLocation to this list.

mjkkirschner: This doesn't work well from a security/hardening standpoint because you would need to know all custom package paths to disable them - unless you mean literally adding the strings
"standardLibrary" and "customPackagePaths" ?

pinzart90: I was actually thinking about having a more generic API that could handle future changes to the package locations.
If we add more locations or remove one of them...we will have to deal with these booleans for a while...

mjkkirschner: To that end I think the the bit flags or an IEnumerable<string> where each string is a package location type(not path) - could work.
doyou prefer one or the other?

pinzart90: I was originally thinking of list of paths...but you bring up a good point...
I think the current way (booleans) is easier to use/understand than enums or list of types.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -76,30 +78,53 @@ public string DefaultPackagesDirectory
...
+                    return Path.Combine(Path.GetDirectoryName(Assembly.GetAssembly(GetType()).Location),
+                        stdLibName, @"Packages");
+                }
+                else
+                {
+                    return stdLibDirectory;
+                }
+            }
+            set
+            {   if(stdLibDirectory!= value)
```
aparajit-pratap, line 102: the `if` could be on the next line

mjkkirschner:
The two new tests were failing because I forgot to force add the new package files.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -76,30 +78,53 @@ public string DefaultPackagesDirectory
...
+            {
+                if (stdLibDirectory == null)
+                {
+                    return Path.Combine(Path.GetDirectoryName(Assembly.GetAssembly(GetType()).Location),
+                        stdLibName, @"Packages");
+                }
+                else
+                {
+                    return stdLibDirectory;
+                }
```
aparajit-pratap, line 99: A ternary operator might be more concise.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -337,7 +378,21 @@ private void ScanAllPackageDirectories(IPreferences preferences)
...
+
+                if (preferences is IDisablePackageLoadingPreferences disablePrefs
+                    &&
+                    //if this directory is the standard library location
+                    //and loading from there is disabled, don't scan the directory.
+                    ((disablePrefs.DisableStandardLibrary && packagesDirectory == StandardLibraryDirectory)
+                    //or if custom package directories are disabled, and this is a custom package directory, don't scan.
+                    || (disablePrefs.DisableCustomPackageLocations && preferences.CustomPackageFolders.Contains(packagesDirectory))))
+                {
+                    continue;
```
aparajit-pratap, line 390: I wonder if it might be a good idea to log a message in the Dynamo console here saying that so and so package location has been skipped because of the preference setting?

test/Libraries/PackageManagerTests/PackageLoaderTests.cs
```diff
@@ -657,6 +646,92 @@ public void HasValidStandardLibraryPath()
...
+
+            var loaderParams = new LoadPackageParams()
+            {
+                PathManager = CurrentDynamoModel.PathManager,
+                Preferences = settings
+            };
+            //invoke the load
+            loader.LoadAll(loaderParams);
+
+            //assert the package in std lib was not loaded.
```
aparajit-pratap, line 665: `assert the package in std lib was loaded.`?

test/Libraries/PackageManagerTests/PackageLoaderTests.cs
```diff
@@ -657,6 +646,92 @@ public void HasValidStandardLibraryPath()
...
+            var loaderParams = new LoadPackageParams()
+            { PathManager = CurrentDynamoModel.PathManager,
+                Preferences = settings };
+            //then invoke load
+
+            loader.LoadAll(loaderParams);
+
+            //assert the package in std lib was not loaded.
+            Assert.AreEqual(0, loader.LocalPackages.Count());
+            Assert.IsFalse(loader.LocalPackages.Any(x => x.Name.Contains("SignedPackage2")));
```
aparajit-pratap, line 688: Is this assertion really necessary since we have already asserted that the count is zero?

test/Libraries/PackageManagerTests/PackageLoaderTests.cs
```diff
@@ -657,6 +646,92 @@ public void HasValidStandardLibraryPath()
...
+            //delete the package we copied.
```
aparajit-pratap, line 689: ?

test/Libraries/PackageManagerTests/PackageLoaderTests.cs
```diff
@@ -657,6 +646,92 @@ public void HasValidStandardLibraryPath()
...
+            {
+                PathManager = CurrentDynamoModel.PathManager,
+                Preferences = settings
+            };
+            //invoke the load
+            loader.LoadAll(loaderParams);
+
+            //assert the package in the custom package path was not loaded
+            Assert.AreEqual(0, loader.LocalPackages.Count());
+            Assert.IsFalse(loader.LocalPackages.Any(x => x.BinaryDirectory.Contains("SignedPackage2")));
```
aparajit-pratap, line 733: same here, is this necessary?

test/Libraries/PackageManagerTests/PackageLoaderTests.cs
mjkkirschner: no, these assertions are not necessary, but I can either move them up a line, or just turn this into a comment - I thought it would be useful to encode which package the test is expected to work with incase someone adds another package to that location.

mjkkirschner:
This is ready for another look!

aparajit-pratap (approved):

mjkkirschner:
@aparajit-pratap @sm6srw two failing tests now after latest merge with master - will take a look.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -70,40 +72,62 @@ public IEnumerable<IExtension> RequestedExtensions
...
+        /// <summary>
+        /// Returns the default package directory where new packages will be installed
+        /// The default package directory is currently the second entry in the list
+        /// The first entry is the standard library.
+        /// </summary>
+        /// <returns>Returns the path to the DefaultPackagesDirectory if found - or null if something has gone wrong.</returns>
         public string DefaultPackagesDirectory
         {
-            get { return packagesDirectories[1]; }
+            get { return packagesDirectories.Count > 1 ? packagesDirectories[1] : null; }
```
mjkkirschner, line 84: @sm6srw I took a shot at fixing this like this...it works for my tests - but lets see if it breaks anything else in the tests.

In the long term I don't think this type of solution will work very well - it's leading to bugs like:
https://jira.autodesk.com/browse/DYN-3571
so we may end up changing it soon anyway.

sm6srw: Agreed.

sm6srw (approved):

PR #11578, opened 2021-03-26
[DYN-3542] Preferences Features Tab

Astul-Betizagasti:
### Purpose

Add the Features tab content for the Preferences window

### Declarations

### Reviewers

Aaron Tang (@QilongTang )

### FYIs


src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -143,11 +149,116 @@ public ObservableCollection<string> NumberFormatList
...
+
         /// <summary>
         /// The PreferencesViewModel constructor basically initialize all the ItemsSource for the corresponding ComboBox in the View (PreferencesView.xaml)
         /// </summary>
         public PreferencesViewModel()
         {
+            PythonEnginesList = new ObservableCollection<string>();
+            PythonEnginesList.Add(Wpf.Properties.Resources.DefaultPythonEngineNone);
+            PythonEnginesList.Add("IronPython2");
+            PythonEnginesList.Add("CPython3");
```
QilongTang, line 259: I dont really like this but seems there is not an easy way to get the enums defined in PythonNodeModels.dll without adding the reference.. Which is not ideal either..

QilongTang (approved):
LGTM with comments

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
Astul-Betizagasti: That is correct, I had issues adding that reference

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -143,11 +149,116 @@ public ObservableCollection<string> NumberFormatList
...
+        public ObservableCollection<string> PythonEnginesList
+        {
+            get
+            {
+                return _pythonEngineList;
+            }
+            set
+            {
+                _pythonEngineList = value;
+                RaisePropertyChanged("PythonEnginesList");
```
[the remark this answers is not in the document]
Astul-Betizagasti: Thanks for the comments, I just updated the PR and include this changes, I also changed some other examples of this that where previous to this PR, let me know if thats ok

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
mjkkirschner: one way is to look for the enum type at runtime, if the lookup fails, then you can do this. The only benefit to that approach is to have the menu items update automatically - both of them are brittle in their own ways.

src/DynamoCoreWpf/UI/Themes/Modern/DynamoModern.xaml
```diff
@@ -2392,6 +2394,88 @@
...
+                BorderThickness="1" 
+                CornerRadius="1" >
+            <Grid>
+                <Grid.ColumnDefinitions>
+                    <ColumnDefinition Width="*" />
+                    <ColumnDefinition Width="Auto" />
+                </Grid.ColumnDefinitions>
+                <Path x:Name="Sign"
+                      Grid.Column="1"
+                      Data="M 0,6 H 12 M 6,0 V 12 Z"
```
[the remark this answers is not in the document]
Astul-Betizagasti: That is the data form that represents the + (plus) sign in the extender, you will see something similar a few lines downwards on the triggers section for when it changes into the - (minus) symbol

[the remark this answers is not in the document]
QilongTang: Thanks, can you add a comment about it just for future references?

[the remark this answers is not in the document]
Astul-Betizagasti: Sure, just added those in the new commit

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
QilongTang: @mjkkirschner I see, I think it is worth a try, especially with the new preferences panel, when user access it, node libs should be loaded 100%. @Astul-Betizagasti can you look at the code example provided by Mike and give it a try? There maybe other ways, like https://stackoverflow.com/questions/5005931/listing-values-within-enums-using-reflection-in-c-sharp which also worth a look

Astul-Betizagasti: Just added a commit that should cover this changes, let me know if you have any comments, thanks!
@QilongTang @mjkkirschner

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -246,6 +249,17 @@ public bool EnableTSplineIsChecked
...
             }
         }
+
+        /// <summary>
+        /// Gets the different Python Engine versions availables from PythonNodeModels.dll
+        /// </summary>
+        /// <returns>Strings array with the different names</returns>
+        private string[] GetPythonEngineOptions()
+        {
+            var assembly = Assembly.LoadFile(AppDomain.CurrentDomain.BaseDirectory + "nodes\\PythonNodeModels.dll");
```
mjkkirschner, line 259: I think this is likely to cause fewer conflicts / be better isolated
https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.reflectiononlyloadfrom?view=net-5.0

if you give this a read:
https://docs.microsoft.com/en-us/dotnet/framework/deployment/best-practices-for-assembly-loading#:~:text=of%20load%20contexts-,Avoid%20binding%20on%20partial%20assembly%20names,to%20the%20default%20load%20context

note the big warnings on top:
if you use LoadFile you will potentially be loading that assembly into multiple contexts.

My other suggestion, which I think I was not clear enough about -  was to just look for the type in the already loaded assemblies - instead of trying to load it manually. If it's not already loaded, give up to avoid all the issues with potentially loading an assembly twice into different contexts, that causes all sorts of weird runtime behavior.

Astul-Betizagasti: Ah I see, I will give those links a look and make some changes, thanks for the explanation

mjkkirschner: @Astul-Betizagasti some more thoughts :)

1. I think the safest thing is still to just check if the type is already around in the loaded assemblies - it should be since this is UI code and the engine should have been loaded when the DynamoModel started up.
2. I don't know if the reflectionOnly context counts as a real context - in which case maybe it's just as bad as the non default context? I'm not sure if we normally load the python assembly into the default or loadFrom contexts.
3. To be super safe if you want to load the assembly for certain - is to load the assembly into a new app domain, check the enum, then dispose the app domain and unload the assembly, this will make sure it is not interfering with dynamo. This could be a good pattern for dynamic operations we may need to do in the future. For example - inspecting a library of nodes before loading them.
4. I also think hardcoding the path is a bad idea - it's fragile and will break when we modify the repo structure, but we won't get compile errors - maybe you can use the pathManager to get this path - I think the `nodes` folder is a known location in the `PathManager`

Astul-Betizagasti: Updated

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -139,15 +148,168 @@ public ObservableCollection<string> NumberFormatList
...
+                return Array.Empty<string>();
+            }
+        }
+
+        private void AddPythonEnginesOptions()
+        {
+            var pythonEngineOptions = GetPythonEngineOptions();
+            if (pythonEngineOptions.Length != 0)
+            {
+                for (int i = 0; i < pythonEngineOptions.Length; i++)
```
QilongTang, line 287: This can probably be simplified by `foreach`

Astul-Betizagasti: Changed it. I initially went with a "for" as it is a little bit more efficient but I understand that in this case clarity may have priority, thanks for the feedback

QilongTang (approved):

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
QilongTang: @mjkkirschner What do you think of the update?

mjkkirschner: Hi @QilongTang @Astul-Betizagasti - I think the PR is good as far as porting over the settings functionality to the preferences dialog.

I do think that the dynamic nature of the python engines needs to be addressed sooner rather than later, and at that point, this preference dialog should be changed to take advantage of it but that seems out of scope for this PR.

some things to think about for that work:
* when we no longer ship IronPython 2.x - how will the preference and enum change?
* If a user wrote their own ironpython3.x engine (it seems to be getting further along) - https://github.com/IronLanguages/ironpython3 - how would the preferences and engine dropdown react? - How would they add their implementation?
* What happens if a users preferences set a default python engine that is not loaded?

mjkkirschner (approved):

QilongTang:
The self serve is failing for known issue. Merging

PR #11598, opened 2021-04-08
Add list level and lacing functionality to the new "If" node.

reddyashish:
### Purpose

This PR is to add list level and lacing functionality to the new PR by introducing a new custom function which would accept "true" and "false" input parameters of arbitrary ranks. 

### Declarations

### Reviewers

@aparajit-pratap

### DIFF
--- a/src/Engine/ProtoCore/DSASM/Defs.cs
+++ b/src/Engine/ProtoCore/DSASM/Defs.cs
@@ -383,6 +383,7 @@ namespace ProtoCore.DSASM
         public const string kFunctionRangeExpression = "%generate_range";
         public const string kDotMethodName = "%dot";
         public const string kInlineConditionalMethodName = "%inlineconditional";
+        public const string IfConditionalMethodName = "%conditionalIf";
         public const string kGetTypeMethodName = "%get_type";
         public const string kNodeAstFailed = "%nodeAstFailed";
         public const string kWatchResultVar = "watch_result_var";
--- a/src/Engine/ProtoCore/DSASM/Defs.cs
+++ b/src/Engine/ProtoCore/DSASM/Defs.cs
@@ -383,7 +383,7 @@ namespace ProtoCore.DSASM
         public const string kFunctionRangeExpression = "%generate_range";
         public const string kDotMethodName = "%dot";
         public const string kInlineConditionalMethodName = "%inlineconditional";
-        public const string IfConditionalMethodName = "%conditionalIf";
+        public const string kIfConditionalMethodName = "%conditionalIf";
         public const string kGetTypeMethodName = "%get_type";
         public const string kNodeAstFailed = "%nodeAstFailed";
         public const string kWatchResultVar = "watch_result_var";
--- a/src/Engine/ProtoCore/Lang/BuiltInFunctionEndPoint.cs
+++ b/src/Engine/ProtoCore/Lang/BuiltInFunctionEndPoint.cs
@@ -307,136 +307,12 @@ namespace ProtoCore.Lang
                     break;
                 case ProtoCore.Lang.BuiltInMethods.MethodID.InlineConditional:
                     {
-                        StackValue svCondition = formalParameters[0];
-                        if (!svCondition.IsBoolean)
-                        {
-                            // Comment Jun: Perhaps we can allow coercion?
-                            Type booleanType = TypeSystem.BuildPrimitiveTypeObject(PrimitiveType.Bool, 0);
-                            svCondition = TypeSystem.Coerce(svCondition, booleanType, runtimeCore);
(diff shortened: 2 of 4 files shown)


src/Engine/ProtoCore/Lang/BuiltInMethods.cs
```diff
@@ -60,6 +60,7 @@ public enum MethodID
             Transpose,
             Union,
             InlineConditional,
+            ConditionalIf,
```
mjkkirschner, line 63: because this enum is public - to avoid reordering the values in the list- this new enum should added at the end of the list, or assigned an int value greater than the last value in the list.

ie 
`ConditionalIf = 1000`

src/Libraries/CoreNodeModels/Logic/If.cs
```diff
@@ -114,11 +114,33 @@ public override IEnumerable<AssociativeNode> BuildOutputAst(List<AssociativeNode
...
+                    paramNumNode,
+                    positionNode,
+                    arguments,
+                    AstFactory.BuildBooleanNode(true)
+                };
+
+                rhs = AstFactory.BuildFunctionCall("__CreateFunctionObject", inputParams);
+            }
+            else
+            {
```
aparajit-pratap, line 140: Add a call to `UseLevelAndReplicationGuide(inputAstNodes);` here to enable list at level.

aparajit-pratap (commented):
Set `ArgumentLacing` to `LacingStrategy.Auto` in the constructors to enable lacing.

src/Engine/ProtoCore/Lang/BuiltInFunctionEndPoint.cs
```diff
@@ -372,7 +372,73 @@ public override StackValue Execute(ProtoCore.Runtime.Context c, List<StackValue>
...
+
+
+                        blockCaller = runtimeCore.DebugProps.CurrentBlockId;
+                        StackFrame bounceStackFrame = new StackFrame(svThisPtr, ci, fi, returnAddr, blockDecl, blockCaller, callerType, type, depth, framePointer, 0, registers, 0);
+
+                        ret = interpreter.runtime.Bounce(blockId, 0, bounceStackFrame, 0, false, runtimeCore.CurrentExecutive.CurrentDSASMExec, runtimeCore.Breakpoints);
+
+                        runtimeCore.RunningBlock = oldRunningBlockId;
+                        break;
+                    }
```
aparajit-pratap, line 441: This entire logic seems to be identical to the one for the inline conditional builtin. If so, then you can refactor the code into a helper function that can be called from both places to eliminate duplicate code.

src/Engine/ProtoCore/DSASM/Defs.cs
```diff
@@ -383,6 +383,7 @@ public struct Constants
         public const string kFunctionRangeExpression = "%generate_range";
         public const string kDotMethodName = "%dot";
         public const string kInlineConditionalMethodName = "%inlineconditional";
+        public const string IfConditionalMethodName = "%conditionalIf";
```
aparajit-pratap, line 386: Prefix this name with `k` like all other strings.

aparajit-pratap (commented):
I think after my comments are addressed we'll have all that we need for the new `If` node:
- partial function support
- lacing
- list at level 🥳

src/Libraries/CoreNodeModels/Logic/If.cs
```diff
@@ -114,11 +114,33 @@ public override IEnumerable<AssociativeNode> BuildOutputAst(List<AssociativeNode
...
+                var arguments = AstFactory.BuildExprList(inputAstNodes);
```
aparajit-pratap, line 141: This line of code can be removed.

reddyashish:
@aparajit-pratap Thank you, addressing your comments.

src/Engine/ProtoCore/Lang/BuiltInMethods.cs
gregmarr: If you put this one in the middle of the list, even if you set the value, you will still need to explicitly set the following one as well to keep its current value.
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/enums

reddyashish: I will add it to the end of the list which seems safer.

reddyashish:
Addressed all comments.

aparajit-pratap (commented):
@reddyashish thanks for making the changes. I think this is good to go once we have an exhaustive set of tests for the new node.

reddyashish:
Merging this to feature branch. Adding tests in a new PR.

PR #11599, opened 2021-04-08
Add OpenXML nodes to import/export data

mmisol:
### Purpose

Two new nodes are added that allow to import and export data from an
Excel spreadsheet. These nodes use the OpenXML SDK, so they do not
make use of the Excel application, thus can be used without Office
installed locally.

These nodes should behave pretty similarly to the existing Excel nodes.
In order to prove that, the tests for the existing nodes have been
copied and adapted for the new nodes.

At the function signature level, there are the following differences:

OpenXMLImportExcel
- Provides inputs for 'startRow' and 'startColumn' (optional)

OpenXMLExportExcel
- Inputs for 'startRow' and 'startColumn' are optional

### Screenshots

Library search

Screen Shot 2021-04-16 at 3 02 58 PM

Nodes on canvas

Screen Shot 2021-04-16 at 3 03 34 PM

### Declarations

### FYIs

@Amoursol


src/Libraries/DSOffice/DSOffice.csproj
```diff
@@ -37,25 +37,31 @@
...
+      <HintPath>..\..\packages\DocumentFormat.OpenXml.2.12.3\lib\net46\DocumentFormat.OpenXml.dll</HintPath>
+    </Reference>
     <Reference Include="Microsoft.Office.Interop.Excel, Version=15.0.0.0, Culture=neutral, PublicKeyToken=71e9bce111e9429c, processorArchitecture=MSIL">
       <HintPath>..\..\packages\Microsoft.Office.Interop.Excel.15.0.4795.1000\lib\net20\Microsoft.Office.Interop.Excel.dll</HintPath>
       <SpecificVersion>False</SpecificVersion>
       <Private>False</Private>
     </Reference>
     <Reference Include="System" />
     <Reference Include="System.Core" />
+    <Reference Include="System.Drawing" />
```
mjkkirschner, line 50: why is system drawing required? That may complicate/negate some of the benefits of moving to a cross platform solution.

mjkkirschner:
@mmisol I have not reviewed in depth yet, but I have some high level questions about approach that I think may help me review better.

1. Are the files produced by these nodes openable by excel without user intervention or training?
2. If 1 is true - why not replace the internal implementation of the existing nodes with open.xml functionality and remove the dependency on excel (of course we need to keep the unused binaries around for a while).
3. It would also mean the tests would not need to be duplicated, just kept as is I believe.
4. What about obsoleting the excel nodes and adding migrations to the new ones?
5. why the need for the windows specific binaries?

IMO a big benefit of this work, in addition to performance gains will be the removal of a source of conflicts and non working excel interop for users. I think, given the familiarity with execl, if those nodes are still present, most users will choose them.

Are there sufficient differences in behavior that you thought adding the new nodes was the best option? What are those differences?

src/Libraries/DSOffice/DSOffice.csproj
```diff
@@ -37,25 +37,31 @@
...
     <Reference Include="System.Xml.Linq" />
     <Reference Include="System.Data.DataSetExtensions" />
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="System.Data" />
     <Reference Include="System.Xml" />
+    <Reference Include="WindowsBase" />
```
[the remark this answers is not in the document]
mmisol: From what I read in the change log of the NuGet package, for .NET Standard this should depend on System.IO.Packaging.

```
Moved to using System.IO.Packaging from dotnet/corefx for .NET Standard 1.3 and WindowsBase for .NET 4.5.
```

mmisol:
Hi @mjkkirschner 

Regarding compatibility, I don't think we can make the guarantee these nodes work exactly like the existing ones. True, I was able to make the tests pass, but given the differences between the two approaches I wouldn't be surprised that something we have overlooked works differently. That's also why I don't think we should be so quick to obsolete the existing nodes. Also, there is plenty of functionality the obsolete Excel nodes cover that I cannot, like opening the Excel application or dealing with workbooks, worksheets, etc. as objects.

The windows-specific binaries seem to appear when targeting .NET only. As I read, if we targeted .NET standard another dependency would be used instead.

I have not reviewed if the dependency on DocumentFormat.OpenXML is safe to be made private. To be honest I don't know the exact list of products/plugins we should check or how to check them.

License and about box, that is something I can do.

src/Libraries/DSOffice/DSOffice.csproj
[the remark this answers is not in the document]
QilongTang: @mjkkirschner Given the current excel node is Windows dependent, I am fine with this. Are you worried about any headless Dynamo env in linux?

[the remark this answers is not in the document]
mjkkirschner: yes, I misunderstood the requirements of this library, it does seem that if we move to .net core / .net 5 we can take advantage of the cross platform packages.

Though I'm still confused by the need for `system.drawing`.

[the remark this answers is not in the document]
mmisol: It seems `System.Drawing` is not needed, as I was able to delete the reference and things continue to work as expected. I'll push the change in a second.

src/Libraries/DSOffice/OpenXmlHelper.cs
```diff
@@ -0,0 +1,647 @@
...
+                        rowData.Add(value);
+                        currentColumnIndex++;
+                    }
+
+                    // Fill remaining columns with empty cells
+                    while (currentColumnIndex <= columnCount + startColumn)
+                    {
+                        rowData.Add(null);
+                        currentColumnIndex++;
+                    }
```
QilongTang, line 92: Curious what would happen if we do not do this step?

mmisol: There are some tests that exercise this. If you have a sparse spreadsheet with lots of empty cells Dynamo is expected to still return empty values for those cells. The following would be an example, where `null` is the value to return but the cell is actually empty:
```
   1    null    null
null       2    null 
null   null       3
```
Notice that `row.Elements<Cell>` would not return empty cells, so in order to return 3 elements for the first two rows, that final loop adds the remaining values as null.

src/Libraries/DSOffice/OpenXmlHelper.cs
```diff
@@ -0,0 +1,647 @@
...
+        /// </summary>
+        /// <param name="columnIndex">Column index (1-based)</param>
+        /// <param name="rowIndex">Row index (1-based)</param>
+        /// <returns>The cell reference that matches the provided column and index</returns>
+        private static string GetCellReference(int columnIndex, uint rowIndex)
+        {
+            var letters = new Stack<char>();
+            do
+            {
+                letters.Push((char)('A' + ((columnIndex - 1) % 26)));
```
QilongTang, line 278: 👍🏻

src/Libraries/DSOffice/OpenXmlHelper.cs
```diff
@@ -0,0 +1,647 @@
...
+        /// <returns>A spreadsheet document</returns>
+        private static SpreadsheetDocument OpenOrCreate(string filePath, string sheetName)
+        {
+            if (File.Exists(filePath))
+            {
+                return SpreadsheetDocument.Open(filePath, true);
+            }
+            else
+            {
+                var document = SpreadsheetDocument.Create(filePath, SpreadsheetDocumentType.Workbook);
```
QilongTang, line 461: Curious if this function would be interrupted with exception that user only have read but not write access to the target path

mmisol: I think either of the code paths of this function should fail in that case, because the above path opens the file to write it.

src/Libraries/DSOffice/OpenXmlHelper.cs
```diff
@@ -0,0 +1,647 @@
...
+                
+                var workbookStylesPart = workbookPart.AddNewPart<WorkbookStylesPart>();
+                workbookStylesPart.Stylesheet = new Stylesheet();
+                AddDefaultStyles(workbookStylesPart.Stylesheet);
+                return document;
+            }
+        }
+
+        /// <summary>
+        /// Adds default styles to the stylesheet so that Office does not identify the spreadsheet as corrupt.
```
QilongTang, line 483: Is there any reference to the changes in this function? Do you foresee MS change how they identify spreadsheet as corrupt?

mmisol: That was something I noticed in practice when the styles were absent, even if they were not needed. This code is the result of reviewing the styles of a newly created spreadsheet in Excel and adding some of those elements until Excel would accept the file created by Dynamo. Unfortunately, there is no documentation for this AFAIK, so the process was pretty hard.

src/Libraries/DSOffice/OpenXmlHelper.cs
QilongTang: Thanks @mmisol for the detailed explanation!

src/Libraries/DSOffice/OpenXmlHelper.cs
QilongTang: @mmisol OK. I am not sure if you have a local env to test this. I can make a note to QA this case if you want?

src/Libraries/DSOffice/OpenXmlHelper.cs
QilongTang: @mmisol I can only imagine... Well, that is a rough start and I hope MS dont change these criteria often, otherwise we are in big trouble..

src/Libraries/DSOffice/OpenXmlHelper.cs
mmisol: I will give this a try but yes, it would be a good thing to try by part of the QA team.

mmisol: The message I get is `Access to the path '<PATH>' is denied.` which must come from Windows.

QilongTang: @mmisol Thanks, as long as Dynamo does not crash and we display this as a warning or so. Appreciate the checking!

test/Libraries/DynamoMSOfficeTests/OpenXmlTests.cs
```diff
@@ -0,0 +1,770 @@
...
+            var readNode = ViewModel.Model.CurrentWorkspace.Nodes.First(n => n.Name == "Data.ImportOpenXml");
+            ViewModel.HomeSpace.Run();
+
+            Assert.IsTrue(readNode.CachedValue.IsCollection);
+            var list = readNode.CachedValue.GetElements();
+            Assert.AreEqual(0, list.Count());
+        }
+
+        [Test]
+        public void CanWriteNullValuesToExcel1()
```
QilongTang, line 681: Can you comment on this test the diff with the test above or use longer name to make it self explain? From reading the code, we are making sure not single value but `null` in the middle of array can be written?

mmisol: I think I can, let me take a look at it to make sure. The "undocumented" aspect is inherited from the existing tests which I copied, but let me see if I can improve that here.

QilongTang (approved):
After some reading of this PR, @mmisol really raised my confidence level and a wonderful piece of work! I cant wait to play with these two nodes. @reddyashish @zeusongit Feel free to take a look as well.

QilongTang:
@mmisol @mjkkirschner Thank you guys for the about box discussion, I will make a note to update internal dependency tracking docs.

test/Libraries/DynamoMSOfficeTests/OpenXmlTests.cs
mmisol: I renamed the tests a little bit to remark the difference. The previous test writes `null` as data, while this one includes `null` as a value in a list.

QilongTang: @mmisol Great! I will leave some review period today for team and come back tomorrow to check more comments, if not expect our team to merge and bang on these this sprint!

QilongTang:
Also @mmisol if I remember correctly, OpenXML can handle other types of docs other than excel, like word and ppt as well. If that is the case, I believe `Data.ImportOpenXml` and `Data.ExportOpenXml` might be still ambiguous? How about `Data.ImportExcelInOpenXml` and `Data.ExportToExcelInOpenXml` or something else makes more sense?

mmisol:
Thanks for the kind words @QilongTang ! I must admit it was more work than I expected initially but hopefully users will appreciate this. It should also ease processing Excel files on the server which can enable some new type of applications.

Regarding the node names, it's true that there may be some ambiguity, but since the function names are under `Data.` I think that should point that the underlying file is a spreadsheet. Anyway, if I had to choose a longer name I would maybe go for `ImportOpenXmlSpreadsheet`. I can change it but may be something to discuss with @Amoursol .

Also, I copied the icons, but it's maybe a good idea to use slightly different ones. What do you think?

QilongTang:
@mmisol Yes, I have discussed the icon thing with @Amoursol and he prefer to use a different color than Green to keep same visual but indicate slight difference. We will need help from UX on touching color, I will raise that at standup.

Amoursol:
Hello @mmisol - Jingyi, Elizabeth and I have discussed. Could you please name the nodes the following?

- Data.OpenXMLImportExcel
- Data.OpenXMLExportExcel

This honors the layout of the library with the data structure and is as clear as possible in as short a name as possible.

mmisol:
@Amoursol @QilongTang I renamed the nodes and updated the icons. The PR description was also updated to reflect those changes.

Let me know if this looks good and if that's the case I'll merge after CI has passed.

QilongTang:
@mmisol This PR looks good, the failing test `Dynamo.Tests.Configuration.PreferenceSettingsTests.TestSettingsSerialization` is due to a different PR.. I am rerunning the self CI

QilongTang:
The latest self CI passed although not updating the status check here seems. This PR looks good to me and we would require some intensive testing this sprint.

PR #11606, opened 2021-04-13
Linter manager

SHKnudsen:
### Purpose

Still a few things to do here, but opening this as a draft for now so its easier to discuss. 

This PR does:
- Adds new LinterManager to DynamoCore. 
    - The LinterManager keeps track of the currently active and all available linter extensions using LinterExtensionDescriptors. 
    - The LinterManager is in charge of subscribing to the nessercary node/graph events and requestion evaluation of the Active linters rules.
    - The LinterManager gets notified when a rule has been evaluated by the `LinterRule`s static event `RuleEvaluated`, if the result is failed it gets stored in the linter manager and if it passes it will be removed.
- Adds Linter rule classes. Linter rules will all inherit from the `LinterRule` base class, currently there is an implementation for `Graph` and `Node` rules, which extensions can use to make their own rules.
- Adds `LinterExtensionBase` base class which new Linter extensions needs to inherit from. The base class handles evaluation of Rules when they are requested by the manager

### Declarations

### Reviewers

@saintentropy @nate-peters @BogdanZavu

### FYIs

- The `TestLinterExtension` is only here for testing purposes and will be removed before merging. The test extension should go into the DynamoSamples project at some point
- _**The most pressing concern right now is probably how we decide which events each rule reacts to, the current implementation evaluates all rules no matter what event triggers the request, which is not great. Also still looking at a good way to make it possible to request a rule evaluation outside of the events specified in the manager**_
- In the process of adding tests


src/Dynamo.All.sln
```diff
@@ -237,6 +237,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NodeAutoCompleteViewExtensi
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Md2Html", "Tools\Md2Html\Md2Html.csproj", "{0893F745-CB1A-427A-8E87-CA927273039A}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "TestLinterExtension", "TestLinterExtension\TestLinterExtension.csproj", "{74A99E4A-671E-4B99-B8E9-602AE5A86323}"
```
BogdanZavu, line 240: I think the project TestLinterExtension can live in Tests folder next to ViewExtensionLibraryTests and to preserve the naming convention there let's name it LintterExtensionTests or similar.

src/DynamoCore/Linting/Rules/LinterRule.cs
```diff
@@ -0,0 +1,79 @@
...
+{
+    /// <summary>
+    /// Base class for all linting rules
+    /// </summary>
+    public abstract class LinterRule
+    {
+        /// <summary>
+        /// Unique id of this rule
+        /// </summary>
+        public abstract string Id { get; }
```
BogdanZavu, line 19: Let's use the Guid class here and everywhere we need to deal with an id.

src/DynamoCore/Extensions/LinterExtensionBase.cs
```diff
@@ -0,0 +1,259 @@
...
+using Dynamo.Graph.Workspaces;
+using Dynamo.Linting;
+using Dynamo.Linting.Rules;
+
+namespace Dynamo.Extensions
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public abstract class LinterExtensionBase : IExtension
```
BogdanZavu, line 16: I think we can remove the Base word and just leave it LinterExtension

src/DynamoCore/Extensions/LinterExtensionBase.cs
```diff
@@ -0,0 +1,259 @@
...
+        {
+            if (linterRules.Contains(linterRule))
+                linterRules.Remove(linterRule);
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public void Activate()
+        {
```
BogdanZavu, line 79: Should we check if is already active here too ?

mjkkirschner:
@SHKnudsen @BogdanZavu @nate-peters @saintentropy @pinzart90 @QilongTang @sm6srw @aparajit-pratap 
 - here are some high level thoughts/a rough proposal regarding where this works lives, how it's introduced as an API, and using it as a jumping off point for other improvements in our process / API organization.

DynamoCore  / DynmoCoreWPF is quite large - to write an extension currently you must reference DynamoCore nuget packages. 

A couple issues here:
* the boundary between extensions and the rest of DynamoCore quickly become vague. Clients get access to things like the entire DynamoView or cast interfaces to concrete implementations  - this means that we don't get feedback that the API should be expanded and the extensions become coupled to those fuzzy boundaries.
* the nuget package supports many use cases, IMO it would be nice to grab a nuget based on the use case so it was more obvious which ones to reference and what assemblies to look through for useful methods. Especially since we don't have much in the way of written tutorials  or guides on these integration subjects.
*  Dynamo core will continue to grow and grow - making deliberate definition of the APIs more difficult.
_(adskers see links below)_

### Okay, why am I pointing this out now?

This work appears to me to be geared almost exclusively to extension authors - which is great, an expansion of our extension API! 🎉 

IMO it could serve as a nice starting point for the creation of a Dynamo.Extension (Dynamo.Extension.WPF might one day exist as well) nuget package.

My rough proposal is as follows:

* this work (linter manager) can be merged to core, everything that is public should be made internal - the interfaces might stay public - but maybe they are moved to the proposed `Dynamo.Extension.dll` binary or perhaps `DynamoServices.dll` which contains many interfaces and has no dependency on `DynamoCore.dll`
* A new .csproject and nuget/nuspec package are created `DynamoVisualProgramming.Extensions` - we only release this with `-beta` flag at first to experiment with it.
* the nuget is published with dynamo's build.
* the new binary is given access to the internal members of DynamoCore via `InternalVisibleTo` attribute.
* The new binary exposes the public apis of the Linter extension.
* We slowly do the same for other extension apis, obsoleting them, migrating them to internal,  and wrapping them deliberately in the new binary.
* Extensions that need to use the linter will reference the new binary, they will likely reference DynamoCore as well for a time, maybe a long time - but I think it will be instructive to see which use cases can be fully migrated to the new package.


thoughts?


for some more (and possibly outdated info) please see:
https://wiki.autodesk.com/pages/viewpage.action?pageId=692803375
https://wiki.autodesk.com/pages/viewpage.action?pageId=554340892

src/DynamoCore/Linting/Rules/LinterRule.cs
```diff
@@ -0,0 +1,79 @@
...
+        internal delegate void RuleEvaluatedHandler(IRuleEvaluationResult result);
+
+        internal static event RuleEvaluatedHandler RuleEvaluated;
+
+        public void OnRuleEvaluated(IRuleEvaluationResult result)
+        {
+            RuleEvaluated?.Invoke(result);
+        }
+
+        public virtual void Dispose() { }
```
BogdanZavu, line 77: do we need to keep this ?

BogdanZavu:
@mjkkirschner let's have a meeting tomorrow about this since making this architectural change in the way Dynamo handles extensions will increase the scope of our work a little bit.
We plan to have very little public API and the linter manager and the rules will not make use of anything from viewmodel. We will provide a base extension with almost all functionality already implemented and an API extension dev that will subclass our default extension will only be allowed to add these linting rules and not much more.

src/Dynamo.All.sln
mjkkirschner: if this project is going to stick around we'll need to update the test jobs to run it, we should also add it to the mono.sln.

src/DynamoCore/Extensions/LinterExtensionBase.cs
```diff
@@ -0,0 +1,259 @@
...
+            ReadyParamsRef.CurrentWorkspaceChanged += OnCurrentWorkspaceChanged;
```
mjkkirschner, line 80: I don't see this unsubscribed anywhere.

src/DynamoCore/Linting/Rules/LinterRule.cs
SHKnudsen: Can do it for the this LinterRule if we prefer that, on the `LinterExtensionBase` which i think is the only other place we use an Id the string type is dictated by `IExtension`

src/DynamoCore/Extensions/LinterExtensionBase.cs
SHKnudsen: think its nice to easily see that this is only serving as a base for other classes? I believe that this is also the pattern used for other base classes in Dynamo like: `ModelBase`, `ViewModelBase`, `ViewExtensionBase`

But happy to change it if we think that is better?

src/DynamoCore/Linting/Rules/LinterRule.cs
SHKnudsen: Nope probably not. Will remove it 👍

src/Dynamo.All.sln
SHKnudsen: The idea was to pull this out when moving out of draft. I really only included it so it was easy to do some manual testing. I was thinking that it would serve as a good starting point for the sample extension that would go into `DynamoSamples`

src/DynamoCore/Extensions/LinterExtensionBase.cs
BogdanZavu: No, you are right, let's keep it as it is :)

src/Dynamo.All.sln
BogdanZavu: Other than samples I would like to keep this around to serve as a minimal set of tests for LinterExtension, but we can deal with this a little bit later, add it to mono and maybe add more tests.

src/DynamoCore/Linting/Interfaces/IRuleEvaluationResult.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Dynamo.Linting.Interfaces
+{
+    public interface IRuleEvaluationResult
```
BogdanZavu, line 9: I'm wondering if we really need this as an interface. I'm thinking to be more restrictive and maybe only use a simple structure which along with RuleEvaluationStatusEnum and SeverityCodesEnum should cover what we need in terms of evaluation results.

src/DynamoCore/Linting/LinterExtensionDescriptor.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Dynamo.Linting
+{
+    public class LinterExtensionDescriptor
```
mjkkirschner, line 9: this looks pretty similar to: https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Extensions/ExtensionData.cs#L6

src/DynamoCore/Linting/LinterManager.cs
```diff
@@ -0,0 +1,124 @@
...
+            {
+                if (RuleEvaluationResults.Contains(result))
+                    return;
+                RuleEvaluationResults.Add(result);
+            }
+        }
+
+        private LinterExtensionBase GetLinterExtension(LinterExtensionDescriptor activeLinter)
+        {
+            return this.dynamoModel.ExtensionManager.
```
mjkkirschner, line 116: why pass make this type depend on DynamoModel when all you use is the ExtensionManager?

src/DynamoCore/Linting/LinterManager.cs
```diff
@@ -0,0 +1,124 @@
...
+                Extensions.
+                Where(x => x.UniqueId == activeLinter.Id).
```
mjkkirschner, line 118: maybe do `OfType()` first?

src/DynamoCore/Extensions/LinterExtensionBase.cs
```diff
@@ -151,22 +155,30 @@ private void InitializeRules()
 
         #region Extension Lifecycle
 
+        ///<inheritdoc/>
         public virtual void Ready(ReadyParams sp)
```
BogdanZavu, line 159: I would say to seal all the virtuals unless absolutely necessary and we can decide to expose them later

src/DynamoCore/Linting/LinterManager.cs
```diff
@@ -0,0 +1,127 @@
...
+using Dynamo.Extensions;
+using Dynamo.Graph.Nodes;
+using Dynamo.Graph.Workspaces;
+using Dynamo.Linting.Interfaces;
+using Dynamo.Linting.Rules;
+using Dynamo.Models;
+
+namespace Dynamo.Linting
+{
+    public class LinterManager : NotificationObject
```
BogdanZavu, line 17: Regarding the lifespan and subscription events of LinterManager and Linter rules.
We can have a dispose method here which is going to : 
unsubscribe from the static events and also call dispose on all the rules 
There is a dispose method in DynamoModel where I think we could do the disposal of Linter manager just before the extension manager gets disposed. Unless there is better place to do that ?

src/DynamoCore/Linting/LinterManager.cs
```diff
@@ -0,0 +1,127 @@
...
+    {
```
BogdanZavu, line 18: Also I see it is possible to remove an extension during the session ( only when an extension gets uninstalled or upgraded ? ). We should handle this case and remove the LinterExtension descriptor and call dispose on it as needed.

src/DynamoCore/Extensions/LinterExtensionBase.cs
SHKnudsen: i think there is some issues with not making these `virtual`, we would essentially remove all entry points for the implementer. For example, I'm using `Ready` to add the linter rules in the `TestLinterExtension` not sure where or how we would add these if these are not accessible for the implementer.

Unless we add a new `abstract List<LinterRule> AddLinterRules()` to the base that gets called in `Ready`?

src/DynamoCore/Extensions/LinterExtensionBase.cs
```diff
@@ -0,0 +1,274 @@
...
+            linterRules.Add(linterRule);
+        }
+
+        /// <summary>
+        /// Remove a LinterRule
+        /// </summary>
+        /// <param name="linterRule"></param>
+        public void RemoveLinterRule(LinterRule linterRule)
+        {
+            if (linterRules.Contains(linterRule))
```
BogdanZavu, line 66: I think we can call Remove directly, it will return false if it doesn't contain the rule and we avoid one search.

src/DynamoCore/Extensions/LinterExtensionBase.cs
BogdanZavu: Yeah, I think we can keep this one

src/DynamoCore/Extensions/LinterExtensionBase.cs
```diff
@@ -0,0 +1,274 @@
...
+        {
+            EvaluateGraphRules(node, NODE_REMOVED_PROPERTY);
+
+            var nodeRules = LinterRules.
+                Where(x => x is NodeLinterRule).
+                Cast<NodeLinterRule>().
+                ToList();
+
+            if (nodeRules is null)
+                return;
```
BogdanZavu, line 262: I would say to unsubscribe at the beginning on the function

src/DynamoCore/Linting/Interfaces/IRuleEvaluationResult.cs
BogdanZavu, line 9: since we made the subclasses internal let's have this one as internal as well

src/DynamoCore/Linting/Interfaces/IRuleEvaluationResult.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
```
BogdanZavu, line 1: I think we can remove all the using statements on all files in all files in Interfaces folder.
And since IRuleEvaluationResult, RuleEvaluationStatusEnum & SeverityCodesEnum are very small maybe we can just use a single file.

src/DynamoCore/Linting/LinterManager.cs
BogdanZavu: Disregard the above, we don't have to worry about this for now

src/DynamoCore/Linting/LinterManager.cs
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        private void OnRuleEvaluated(IRuleEvaluationResult result)
+        {
+            if (result is null)
+                return;
+
+            if (result.Status == RuleEvaluationStatusEnum.Passed)
+            {
+                if (!RuleEvaluationResults.Contains(result))
```
BogdanZavu, line 103: same, we can remove only

src/DynamoCore/Linting/Rules/GraphLinterRule.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// <summary>
+        /// Method to call when this rule needs to be evaluated.
+        /// This will use <see cref="EvaluateFunction(WorkspaceModel, NodeModel)"/> to evaluate the rule.       
+        /// </summary>
+        /// <param name="workspaceModel"></param>
+        /// <param name="modifiedNode"></param>
+        internal void Evaluate(WorkspaceModel workspaceModel, NodeModel modifiedNode = null)
+        {
+            var pair = EvaluateFunction(workspaceModel, modifiedNode);
+            var result = new GraphRuleEvaluationResult(this.Id, pair.Item1, pair.Item2);
```
BogdanZavu, line 26: null pair check

src/DynamoCore/Linting/Rules/GraphRuleEvaluationResult.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
```
BogdanZavu, line 3: lines 3,4,5 can be removed

src/DynamoCore/Linting/Rules/LinterRule.cs
BogdanZavu: we can leave it as it is, it's similar to many other places

test/DynamoCoreTests/Linting/LinterManagerTests.cs
```diff
@@ -0,0 +1,204 @@
...
+using Dynamo.Graph.Workspaces;
+using Dynamo.Linting.Interfaces;
+using Dynamo.Linting.Rules;
+using Dynamo.Models;
+using Dynamo.Scheduler;
+using Moq;
+using Moq.Protected;
+using NUnit.Framework;
+
+namespace DynamoCoreWpfTests.Linter
```
BogdanZavu, line 15: Maybe this should be Dynamo.Core.Tests.Linting

src/Dynamo.All.sln
BogdanZavu: Ok, let's do like you said, the tests under LinterManager should suffice for now I think.

src/DynamoCore/Linting/LinterManager.cs
```diff
@@ -0,0 +1,133 @@
...
+using Dynamo.Linting.Rules;
+using Dynamo.Models;
+
+namespace Dynamo.Linting
+{
+    public class LinterManager : NotificationObject, IDisposable
+    {
+        #region Private fields
+        private readonly IExtensionManager extensionManager;
+        private LinterExtensionDescriptor activeLinter;
```
BogdanZavu, line 21: We said we are going to serialize the active linter here right ?

BogdanZavu:
@SHKnudsen overall I think this is looking really good! Let's clarify the serialization aspect tomorrow and I think we can merge it.

src/DynamoCore/Linting/Interfaces/IRuleEvaluationResult.cs
SHKnudsen: if i make the interface internal i get issues with the linter rules `InitializeRule` (https://github.com/DynamoDS/Dynamo/pull/11606/files#diff-07729abac4d567c8b6ac2d6adff4f79e04715870506e8098e5cc1f3ca1afbcf0R45) as this will then throw an error because the return type is less accessible than the method.

src/DynamoCore/Linting/LinterManager.cs
BogdanZavu: We discussed about this internally and plan is to have another entry in the dynamo file : 
`Linting : { activeLinter: adadasdds, issuesCount: .... }`

SHKnudsen:
Hmm, i cant access the SelfServe details, so not sure what this issue is..

BogdanZavu:
[
[linter-tests.txt](https://github.com/DynamoDS/Dynamo/files/6394837/linter-tests.txt)
](url)

BogdanZavu:
Only 1 test remaining :

DynamoCoreWpfTests.SerializationTests.JSONisSameBeforeAndAfterSaveWithDummyNodes

MESSAGE:
  Expected: True
  But was:  False

+++++++++++++++++++
STACK TRACE:
   at DynamoCoreWpfTests.SerializationTests.JSONisSameBeforeAndAfterSaveWithDummyNodes() in c:\WorkspaceDynamo\test\DynamoCoreWpfTests\SerializationTests.cs:line 444

BogdanZavu (approved):

src/DynamoCore/Linting/LinterExtensionDescriptor.cs
mjkkirschner: @BogdanZavu @SHKnudsen 
1. does this class need to exist? See the very similarExtensionData class, could this just be a tuple instead? Do we expect this to grow
2. Does this class really need to be public?

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -347,6 +395,78 @@ internal virtual void OnSetNodeDeltaState(DeltaComputeStateEventArgs e)
...
+            IEnumerable<AnnotationModel> annotations,
+            IEnumerable<PresetModel> presets,
+            ElementResolver resolver,
+            WorkspaceInfo info,
+            bool verboseLogging,
+            bool isTestMode,
+            LinterManager linterManager)
+            : base(nodes, notes, annotations, info, factory, presets, resolver, linterManager)
+        {
+            Debug.WriteLine("Creating a home workspace...");
```
mjkkirschner, line 432: please don't copy and paste this, if if it has to exist in two spots refactor it out into a shared private method, this section of code is confusing enough.

src/DynamoCore/Linting/Interfaces/IRuleEvaluationResult.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace Dynamo.Linting.Interfaces
+{
+    public interface IRuleEvaluationResult
```
mjkkirschner, line 3: do you intend that this interface and related enums is used by anyone but Dynamo? If not, can it be internal?

src/DynamoCore/Linting/LinterManager.cs
```diff
@@ -0,0 +1,150 @@
...
+using Dynamo.Extensions;
+using Dynamo.Graph.Nodes;
+using Dynamo.Graph.Workspaces;
+using Dynamo.Linting.Interfaces;
+using Dynamo.Linting.Rules;
+using Dynamo.Models;
+
+namespace Dynamo.Linting
+{
+    public class LinterManager : NotificationObject, IDisposable
```
mjkkirschner, line 17: please add a summary for public type.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -323,6 +324,8 @@ public string Version
         /// </summary>
         public IExtensionManager ExtensionManager { get { return extensionManager; } }
 
+        public LinterManager LinterManager { get; }
```
mjkkirschner, line 327: summary on public property

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -861,6 +866,9 @@ protected DynamoModel(IStartConfiguration config)
 
                     try
                     {
+                        if (ext is LinterExtensionBase linter)
```
mjkkirschner, line 869: please wrap with `{`:
https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards#conditional-blocks

src/DynamoCore/Linting/Interfaces/IRuleEvaluationResult.cs
SHKnudsen: No intention for anyone else but Dynamo to use it.

The issue with making it internal is explained here
https://github.com/DynamoDS/Dynamo/pull/11606#discussion_r618430760

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -247,6 +248,7 @@ internal virtual void OnSetNodeDeltaState(DeltaComputeStateEventArgs e)
         /// <param name="verboseLogging">Indicates if detailed descriptions should be logged</param>
         /// <param name="isTestMode">Indicates if current code is running in tests</param>
         /// <param name="fileName">Name of file where the workspace is saved</param>
+        [Obsolete("please use the version with linterManager parameter.")]
```
mjkkirschner, line 251: some questions on this:
1. have you verified that all these constructors can be replaced by your new constructor? - What I mean is, do they cover any other use cases?
2. If a file is deserialized with no linterManager - should null be passed?

src/DynamoCore/Linting/Interfaces/IRuleEvaluationResult.cs
mjkkirschner: @SHKnudsen - is `InitalizeRule` virtual and protected so it can be implemented by third parties or is it just for our use? 

You can make members both internal and protected.

SHKnudsen: `InitializeRule` is just for us.

internal protected seems to have the same issue (https://stackoverflow.com/questions/49250829/c-sharp-protected-internal-method-with-internal-return-type)

However `private protected` seems to do the trick, I was not aware that was a possibility.

src/DynamoCore/Linting/LinterExtensionDescriptor.cs
BogdanZavu: So ExtensionData has a few more things in it and I feel that using it instead of LinterDescriptor will make things more confusing. I guess the most valuable thing of LinterDescriptor is the actually the name. I would expect that there are many similar or almost similar simple structs across dynamo. I think we can have a tuple here but I would prefer the named structure instead. Most likely it will not grow or if it will grow it will grow very little.
I think it can be internal.

mjkkirschner: @BogdanZavu - sounds good to me if thats possible.

BogdanZavu: I forgot about the view, we need it there (in the other [PR](https://github.com/DynamoDS/Dynamo/pull/11634/files) ) where the active linter is being set from the UI.

mjkkirschner: ah

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
SHKnudsen: - i would say I've tested this as much as possible, also all different constructors have gotten a new equivalent just with the linterManager param. Before the changes there were 3 ctors, I've add 3 new ones that replicate the old ones but with that new param.
- Essentially the only thing that is serialized from the linter manager is the active linter (name and id) and issue counts (the counts are only for things like GD and player). If there is no active linter when a graph is saved we just serialize `Linting : []`, when deserializing all we are doing is setting the active linter to whatever was serialized (if it's available on the manager). The linter manager is created on the DynamoModel, so if a workspace is deserialized without an active linter value we just don't set any active linter on the LinterManager.

test/pkgs/PackageThatWillBeModified/dyf/ANodeToModify.dyf
```diff
@@ -86,7 +86,7 @@
         {
           "Id": "52c34d69fe42408ca5c857f371168587",
           "Name": "x",
-          "Description": "x value.\n\nvar[]..[]",
+          "Description": "Integer or double value\n\nvar[]..[]",
```
mjkkirschner, line 89: @SHKnudsen - I think these changes were likely committed by accident? This seems unrelated to this PR?

PR #11611, opened 2021-04-14
Display Python dep info to users on package manager window

zeusongit:
### Purpose

This is a separate branch that will include all the visual changes until the backend is wired up, that includes a way to populate package dependency field with Python versions used in that package.

pyspy3

### DIFF
--- a/src/DynamoCoreWpf/UI/Converters.cs
+++ b/src/DynamoCoreWpf/UI/Converters.cs
@@ -77,6 +77,63 @@ namespace Dynamo.Controls
         }
     }
 
+    /// <summary>
+    /// Converts the list of package dependencies to a string
+    /// </summary>
+    public class DependencyListToStringConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            string depString = string.Empty;
+            bool flag = false;
+            if (value != null)
+            {
+                List<Greg.Responses.Dependency> depList = (List<Greg.Responses.Dependency>)value;
+                for (int i = 0; i < depList.Count(); i++)
+                {
+                    if (depList[i].name == "CPython" || depList[i].name == "IronPython")
+                    {
+                        depString += depList[i].name + " " + depList[i]._id + ", ";
+                        flag = true;
+                    }
+                }
+                return flag ? depString.Remove(depString.Length - 2) : null;
+            }
+            return null;
+        }
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            return null;
+        }
+    }
+
+    public class EmptyDepStringToCollapsedConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter,
+          CultureInfo culture)
+        {
+            if (value != null)
+            {
+                List<Greg.Responses.Dependency> depList = (List<Greg.Responses.Dependency>)value;
+                for (int i = 0; i < depList.Count(); i++)
+                {
+                    if (depList[i].name == "CPython" || depList[i].name == "IronPython")
+                    {
+                        return Visibility.Visible;
+                    }
+                }
+            }
+            return Visibility.Collapsed;
+        }
+
+        public object ConvertBack(object value, Type targetType, object parameter,
+          CultureInfo culture)
+        {
+            return null;
+        }
+    }
+
     public class PackageSearchStateToStringConverter : IValueConverter
     {
         public object Convert(object value, Type targetType, object parameter,
--- a/src/DynamoCoreWpf/UI/Converters.cs
+++ b/src/DynamoCoreWpf/UI/Converters.cs
@@ -36,7 +36,7 @@ namespace Dynamo.Controls
     {
         private const int MaxChars = 100;
         private const double MinFontFactor = 7.0;
-        
+
         public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
         {
             var tooltip = value as string;
@@ -82,6 +82,9 @@ namespace Dynamo.Controls
     /// </summary>
     public class DependencyListToStringConverter : IValueConverter
     {
+
+        private const string CPython = "CPython";
+        private const string IronPython = "IronPython";
         public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
         {
             string depString = string.Empty;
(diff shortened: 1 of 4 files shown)


Amoursol:
The UX implementation and colors LGTM @zeusongit !

QilongTang:
@zeusongit So this PR is about using dependency field. Is there a different PR using the HostDependency implementation for us to compare?

src/DynamoPackages/PackageManagerSearchElement.cs
```diff
@@ -64,6 +64,11 @@ public string HostsString
                         hostsString += host + "  ";
                     }
                 }
+                //added to this dummy package for representation, to be removed and replace space at the end with newline above
+                if (Name == "dronovBIM_v2")
+                {
+                    hostsString += "CPython3"+ Environment.NewLine + "IronPython2"+ Environment.NewLine + "Revit";
```
zeusongit, line 70: @QilongTang  Here `HostsString` is derived from `host_dependencies` that we are using  to display dependencies, including python deps.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
```diff
@@ -573,6 +573,15 @@ public void RefreshAndSearchAsync()
...
         {
             element.RequestDownload += this.PackageOnExecuted;
+
+            //added dummy values for representation, to be removed
+            if (element.Model.Name == "dronovBIM_v2")
+            {
+                element.Versions[0].Item1.full_dependency_ids.Add(new Dependency() { name = "CPython", _id = "3.0" });
+                element.Versions[1].Item1.full_dependency_ids.Add(new Dependency() { name = "IronPython", _id = "2.0" });
+                element.Versions[2].Item1.full_dependency_ids.Add(new Dependency() { name = "IronPython", _id = "2.0" });
+                element.Versions[2].Item1.full_dependency_ids.Add(new Dependency() { name = "CPython", _id = "3.0" });
```
zeusongit, line 583: We can add another property such as `version` or `vers` to record version instead of `_id` from PMClient class while implementating.

zeusongit:
I would like to conclude this spike, so that we can move further with discussion/implementation on this topic.
In this PR i have utilised both the approaches i.e to use `host_dependencies` as well as `full_dependency_ids` to display information.

Screen Shot 2021-04-16 at 4 22 59 PM


There are some observations/concerns:
- If using 'host_dependencies' we lose version info. 
- Do we display the latest versions dependency as the primary dependency in the hyperlink tooltip? Or a cumulative list of dependencies from all versions?
- If using 'full_dependency_ids' we need to figure out a way to populate exact versions of dependencies.
- Also we need to provide a way for these properties to be set, by the user or by some implicit process while publishing a new package.
- There also has to be a one-time function that scans all the packages and fill in dependency data for the existing packages.

src/DynamoCoreWpf/UI/Converters.cs
```diff
@@ -77,6 +77,63 @@ public object ConvertBack(object value, Type targetType, object parameter, Cultu
...
+    public class DependencyListToStringConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            string depString = string.Empty;
+            bool flag = false;
+            if (value != null)
+            {
+                List<Greg.Responses.Dependency> depList = (List<Greg.Responses.Dependency>)value;
+                for (int i = 0; i < depList.Count(); i++)
```
reddyashish, line 92: Use foreach loop here instead of indexing the list.

src/DynamoCoreWpf/UI/Converters.cs
```diff
@@ -77,6 +77,63 @@ public object ConvertBack(object value, Type targetType, object parameter, Cultu
...
+                {
+                    if (depList[i].name == "CPython" || depList[i].name == "IronPython")
```
reddyashish, line 94: Use string constants instead of string literals.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
reddyashish: Yes i think `version` would be a better choice.

src/DynamoCoreWpf/Views/PackageManager/PackageManagerSearchView.xaml
```diff
@@ -292,8 +304,18 @@
...
+                                                                <ToolTip Style="{StaticResource DependencyTooltip}">
+                                                                    <StackPanel Background="#535353" Height="Auto" Width="100" Orientation="Horizontal" HorizontalAlignment="Left" VerticalAlignment="Top" >
+                                                                        <TextBlock TextWrapping="Wrap" Padding="8" FontSize="12" Foreground="WhiteSmoke" Text="{Binding Path=Model.HostsString}"/>
+                                                                    </StackPanel>
+                                                                </ToolTip>
+                                                            </TextBlock.ToolTip>
+                                                            <Hyperlink Foreground="LightBlue">Dependencies</Hyperlink>
+                                                        </TextBlock>
+                                                        <!--<TextBlock Width ="90" FontSize="10"  Name="HostDependency" Text="{Binding Path=Model.HostsString}" Margin="5,0,0,0" HorizontalAlignment="Right" Foreground="LightBlue"
```
reddyashish, line 317: Remove the commented code.

src/DynamoPackages/PackageManagerSearchElement.cs
reddyashish: Avoid string literals wherever possible. Makes it easy to change it.

reddyashish:
@zeusongit just one more comment, forgot to mention it earlier. LGTM. What about the dummy values added in the PR? Will they be removed once the followup changes are made? If yes, please make sure the next task mentions that.

zeusongit:
> @zeusongit just one more comment, forgot to mention it earlier. LGTM. What about the dummy values added in the PR? Will they be removed once the followup changes are made? If yes, please make sure the next task mentions that.

Yes @reddyashish the dummy values will be removed, if you see the base branch of the PR  that is not `master`, this will be merged in a separate branch until other parts of this task are worked on.

QilongTang (approved):

PR #11613, opened 2021-04-15
Update NodeAutocomplete search criteria to also include nodes keywords and description

zeusongit:
[DYN-3357](https://jira.autodesk.com/browse/DYN-3357)

The node-autocomplete search used only the node name to match the search key. It has been updated to include node's keywords and description as well.

In the example below `radius` is the search key that is present in `ByFillet` node's description and therefore did not use to appear in the search results before.

Screen Shot 2021-04-15 at 10 28 04 AM

Before:
Screen Shot 2021-04-15 at 10 21 19 AM

After:

Screen Shot 2021-04-15 at 10 21 28 AM

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCore/Search/SearchDictionary.cs
+++ b/src/DynamoCore/Search/SearchDictionary.cs
@@ -295,7 +295,7 @@ namespace Dynamo.Search
         /// </summary>
         /// <param name="subset">The subset that will be used to match elements in tagDictionary.</param>
         /// <returns></returns>
-        private static bool MatchWithSubset2(IGrouping<string, Tuple<V, double>> pair, IEnumerable<SearchElements.NodeSearchElement> subset)
+        private static bool MatchWithSubset(IGrouping<string, Tuple<V, double>> pair, IEnumerable<SearchElements.NodeSearchElement> subset)
         {
             foreach (var eleAndWeight in pair)
             {
--- a/src/DynamoCore/Search/SearchDictionary.cs
+++ b/src/DynamoCore/Search/SearchDictionary.cs
@@ -23,10 +23,16 @@ namespace Dynamo.Search
             this.logger = logger;
         }
 
+        /// <summary>
+        ///     Dictionary of searchElement:(tag, weight)
+        /// </summary>
         protected readonly Dictionary<V, Dictionary<string, double>> entryDictionary =
             new Dictionary<V, Dictionary<string, double>>();
 
-        private List<IGrouping<string, Tuple<V, double>>> tagDictionary;
+        /// <summary>
+        /// Dictionary of tag:(list(searchelement, weight)) which contains all nodes that share a tag 
+        /// </summary>
+        private List<IGrouping<string, Tuple<V, double>>> tagDictionary;        
 
         /// <summary>
         ///     All the current entries in search.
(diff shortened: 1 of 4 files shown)


mjkkirschner:
Is there a reason this uses its own search instead of the search code that already exists for in canvas and library search?

QilongTang:
@zeusongit Did you made any comparison of in-canvas search yet? I am not sure if string comparison of these three cates will be a 100% match, I think originally we were leaning towards to just reuse the existing search code in in-canvas search.

zeusongit:
@mjkkirschner The in canvas and library searches are conducted on complete set of nodes, whereas this is done on a subset of that, which is defined by the target input port i.e all the nodes which are compatible with it. 
I had a discussion with @QilongTang and will try to use the existing search algorithm by providing it our own subset to search from.

src/DynamoCore/Search/SearchDictionary.cs
```diff
@@ -283,6 +289,29 @@ private static bool MatchWithQueryString(string key, string[] subPatterns)
             return (double)numberOfMatchSymbols / numberOfAllSymbols > 0.8;
         }
 
+        /// <summary>
+        /// Check if any element in the given subset
+        /// matches any element in the tagDictionary.
+        /// </summary>
+        /// <param name="subset">The subset that will be used to match elements in tagDictionary.</param>
+        /// <returns></returns>
+        private static bool MatchWithSubset2(IGrouping<string, Tuple<V, double>> pair, IEnumerable<SearchElements.NodeSearchElement> subset)
```
QilongTang, line 298: Is there a version1 of this function, if not, let's get rid of the `2` in function name?

test/DynamoCoreWpfTests/NodeAutoCompleteSearchTests.cs
```diff
@@ -349,7 +349,7 @@ public void SearchNodeAutocompletionSuggestions()
 
             // Filter the node elements using the search field.
             searchViewModel.SearchAutoCompleteCandidates("ar");
-            Assert.AreEqual(2 , searchViewModel.FilteredResults.Count());
+            Assert.AreEqual(4 , searchViewModel.FilteredResults.Count());
```
QilongTang, line 352: @zeusongit Looks like the number diff is expected due to?

src/DynamoCoreWpf/ViewModels/Search/SearchViewModel.cs
```diff
@@ -866,13 +866,14 @@ private void IsSelectedChanged(object sender, PropertyChangedEventArgs e)
         }
 
         /// <summary>
-        ///     Performs a search using the given string as query.
+        ///     Performs a search using the given string as query and subset, if provided.
         /// </summary>
         /// <returns> Returns a list with a maximum MaxNumSearchResults elements.</returns>
         /// <param name="search"> The search query </param>
-        internal IEnumerable<NodeSearchElementViewModel> Search(string search)
+        /// <param name="subset">Subset of nodes that should be used for the search instead of the complete set of nodes</param>
```
QilongTang, line 873: hmm can you specify these are NodeSearchElements?

test/DynamoCoreWpfTests/NodeAutoCompleteSearchTests.cs
zeusongit: Because the search term `ar` now returns more than 2 results as the search scope has been increased to cover description and keywords.

QilongTang (approved):

PR #11617, opened 2021-04-15
WIP: DYN-3625 Add shutdown handling for IExtensions

saintentropy:
### Purpose

The purpose of this PR is to add invocation of the Shutdown() method defined in the IExtension interface.  Currently the Shutdown() method is called for IViewExtension when the DynamoWindowView is being closed -> https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs#L1616 but the Extension Shutdown is never called.

### Declarations

### FYIs

@QilongTang @mjkkirschner

### DIFF
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -477,6 +477,18 @@ namespace Dynamo.Models
             ShutdownRequested = true;
 
             OnShutdownStarted(); // Notify possible event handlers.
+            
+            foreach (var ext in ExtensionManager.Extensions)
+            {
+                try
+                {
+                    ext.Shutdown();
+                }
+                catch (Exception exc)
+                {
+                    Logger.Log($"{ext.Name} :  {exc.Message} during shutdown");
+                }
+            }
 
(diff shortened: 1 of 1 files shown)


mjkkirschner:
Doesn't appear to be any test failures https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/2386/

I think theres the possibility of double dispose happening because previous examples all show dispose being called from shutdown (without sideeffect currently, since it's not called) - but this is probably still a worthwhile change, we should just update the samples and all built in extensions.

As part of this PR would you mind also removing all shutdown implementations that call dispose for the extensions in the repo?

saintentropy:
@mjkkirschner you mean like this -> https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoPackages/PackageManagerExtension.cs#L188 correct?

mjkkirschner:
@saintentropy yep yep.

QilongTang:
this should be addressed by https://github.com/DynamoDS/Dynamo/pull/11730. Closing.

PR #11620, opened 2021-04-15
[DYN-3577] Preferences Features sub-panel implementation

Astul-Betizagasti:
### Purpose

Add functionality to the Features sub-panel in the Preferences windows.
Redesign the current toggle button used in the Preferences window to solve a bug with its behavior.

### Declarations

### Reviewers

Aaron Tang (@QilongTang )

### FYIs


src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -292,24 +292,6 @@ public bool ShowCodeBlockLineNumber
             }
         }
 
-        /// <summary>
-        /// Indicates whether to make T-Spline nodes (under ProtoGeometry.dll) discoverable
-        /// in the node search library.
-        /// </summary>
-        public bool EnableTSpline
```
QilongTang, line 299: Unfortunately we can only Obsolete this

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -552,23 +534,6 @@ public bool ShowTabsAndSpacesInScriptEditor
                 RaisePropertyChanged(nameof(ShowTabsAndSpacesInScriptEditor));
             }
         }
-
-        /// <summary>
-        /// Engine used by default for new Python script and string nodes. If not empty, this takes precedence over any system settings.
-        /// </summary>
-        public string DefaultPythonEngine
```
QilongTang, line 559: Same here

QilongTang (commented):
Some Comments then LGTM

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -24,6 +26,29 @@ public class PreferencesViewModel : ViewModelBase
...
+            }
+        }
+
+        private DynamoPythonScriptEditorTextOptions editTextOptions;
+        public DynamoPythonScriptEditorTextOptions PythonScriptEditorTextOptions
+        {
+            get { return editTextOptions; }
+            set
+            {
+                editTextOptions = value;
```
QilongTang, line 46: Do we need to raise property change here?

Astul-Betizagasti: I believe not in this case, as this is only used to access other properties that the user can change on the panel, and this properties will rise the event. (Example in line 247)

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
Astul-Betizagasti: Just added this methods back and added the obsolete tag, I had to also bring back `internal void HideUnhideNamespace(bool hide, string library, string namespc)` in DynamoModel.cs as it is used by one of them

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
mjkkirschner: does it need to be public then?

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -303,12 +365,15 @@ private void AddPythonEnginesOptions()
         /// <summary>
         /// The PreferencesViewModel constructor basically initialize all the ItemsSource for the corresponding ComboBox in the View (PreferencesView.xaml)
         /// </summary>
-        public PreferencesViewModel()
+        public PreferencesViewModel(PreferenceSettings preferenceSettings, DynamoPythonScriptEditorTextOptions editTextOptions)
```
mjkkirschner, line 368: You'll need to add another constructor and leave this default one, you can mark it obsolete if you think it should not longer be used.

QilongTang: @mjkkirschnner `PreferencesViewModel` is only introduced in 2.12.0, I thought we could touch it this way?

mjkkirschner: sorry guys! You're right, hard for me to keep track!

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
Astul-Betizagasti: good catch, going over that piece of code again I notice it can be simplified, I will update on the next commit

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
Astul-Betizagasti: Good, will update on the next commit.
Just out of curiosity, given that the constructor with no parameters will make the class obsolete and probably even crash if trying to change the properties that make us of things like the PreferenceSettings, why do we need to have a constructor with no parameters?

QilongTang: @Astul-Betizagasti It only make sense to exit for a short while when we have to be API compatible, but in this case we can touch it.

Astul-Betizagasti: Oh, my browser hadn't updated and I posted that message only seeing Michael's first comment. I will leave it as it was then. Thanks for the review and comments guys!

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
Astul-Betizagasti: Just updated this, I refactored a little bit that property and one more as it didnt make much sense having a private property to access another private one. Let me know what you think, thanks.

QilongTang (approved):

PR #11621, opened 2021-04-16
Update NodeModel Autocomplete Match Criteria

QilongTang:
Please Note:

### Purpose

Per https://jira.autodesk.com/browse/DYN-3470. 

For NodeModel nodes, output types are defined by user which usually do not contain namespace

example
```
    [IsDesignScriptCompatible]
    [NodeName("Color Range")]
    [NodeCategory("Core.Color.Create")]
    [NodeDescription("ColorRangeDescription",typeof(Resources))]
    [NodeSearchTags("ColorRangeSearchTags", typeof(Resources))]

    [InPortNames("colors", "indices", "value")]
    [InPortTypes("DSCore.Color[]", "double[]", "double")]
    [InPortDescriptions(typeof(Resources),
        "ColorRangePortDataColorsToolTip",
        "ColorRangePortDataIndicesToolTip",
        "ColorRangePortDataValueToolTip")]
    [OutPortNames("color")]
    [OutPortTypes("Color")]
    [OutPortDescriptions(typeof(Resources),
        "ColorRangePortDataResultToolTip")]
    [AlsoKnownAs("DSCoreNodesUI.ColorRange")]
```
                
Since the OutPortType(OutputParameter) is Color instead of DSCore.Color, it was not a exact match for node autocompete. While we can define output type with namespace there, in order to be consistent on simplified PortType name display in library, we make it tolerant here.

Result:
NodeModel nodes like `Color.Range` and `Color.Pallette` now appear in NodeAutoComplete Suggestions
image

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
@@ -166,8 +166,6 @@ namespace Dynamo.ViewModels
             {
                 //for now, remove rank from descriptors
                 var returnTypeName = ztSearchElement.Descriptor.ReturnType.Name;
-
-                var descriptor = ztSearchElement.Descriptor;
                 if ((returnTypeName == inputPortType)
                     || DerivesFrom(inputPortType, returnTypeName, core))
                 {
@@ -175,10 +173,14 @@ namespace Dynamo.ViewModels
                 }
             }
 
-            // NodeModel nodes, match any output return type to inputport type name
+            // NodeModel nodes, match any output return type to InputPortType name
             foreach (var element in Model.SearchEntries.OfType<NodeModelSearchElement>())
             {
-                if (element.OutputParameters.Any(op => op == inputPortType))
+                // For NodeModel nodes, output types are defined by user which usually do not contain namespace
+                // e.g. For ColorRange, the OutPortType(OutputParameter) is Color instead of DSCore.Color
+                // While we can define output type with namespace there, in order to be consistent on simplified  
+                // PortType name display in library, we make it tolerant here
+                if (element.OutputParameters.Any(op => inputPortType.EndsWith(op)))
                 {
                     elements.Add(element);
                 }


src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -166,19 +166,21 @@ internal IEnumerable<NodeSearchElement> GetMatchingSearchElements()
...
-            // NodeModel nodes, match any output return type to inputport type name
+            // NodeModel nodes, match any output return type to InputPortType name
             foreach (var element in Model.SearchEntries.OfType<NodeModelSearchElement>())
             {
-                if (element.OutputParameters.Any(op => op == inputPortType))
+                // For NodeModel nodes, output types are defined by user which usually do not contain namespace
+                // e.g. For ColorRange, the OutPortType(OutputParameter) is Color instead of DSCore.Color
+                // While we can define output type with namespace there, in order to be consistent on simplified  
+                // PortType name display in library, we make it tolerant here
+                if (element.OutputParameters.Any(op => inputPortType.EndsWith(op)))
```
aparajit-pratap, line 183: This could result in incorrect matches. Shouldn't it be an exact match? I think the only way to achieve that is to impose full namespaces for both input and output port types. For example, this will give a false match if the output port type is `Color` from DSCore namespace for one node and another node expects `PackageX.Color` or just `Color` but from a different namespace.

QilongTang: @aparajit-pratap Sure, please see the reasoning I mentioned. Our library does not display full type either, e.g. for Color.Red, the out put type displayed in library is just Color instead of DSCore.Color. Right now, Color.Range is following the same. If we make it exact match everywhere, should we use the full type everywhere? The false match case right now arguably is already misleading in library. Even if it is mismatch, it could still be valuable to be listed as potential. If we all prefer strong match, I would say we have some systematic changes to make. I am open for what you guys prefer.

aparajit-pratap: Well, I think it's better to be verbose than incorrect. For a correct match, we must have exact matches with full namespaces.

QilongTang: @aparajit-pratap I interpret it as you are in supportive of full type names everywhere in Dynamo then? Otherwise on book, Color.Red can't be suggested to Color.Blue which is the current behavior as well.

aparajit-pratap: > I interpret it as you are in supportive of full type names everywhere in Dynamo then

Yes

mjkkirschner: This is my feeling as well (full namespaces in NodeModel port types) - many geometry packages share the same type names as LibG, but won't work at all.

QilongTang: @aparajit-pratap @mjkkirschner OK, I am fine with that. The problem with that is it will become a much bigger project to use full type everywhere in Dynamo in terms of display and autocomplete. 
1. For this bug, I will go through all 40+ node model nodes and fix their types. 
2. We will also need to create some public wiki to ask package devs to correct their nodes, otherwise their nodes will not appear in autocomplete at all. That was the downside I was seeing.
3. I will log a diff task to make sure our Library display type as full type instead of shorten version which is confusing.

What else you guys think also needed?

mjkkirschner: 2. instead of a wiki, one alternative is some kind of warning at package publish time or other "compile time" check - not to expand the scope even further? 
Maybe a blog post / change in the api wiki docs would be sufficient.

3. hmm - I think theres a balance there - even though it's a bit confusing - there are some really long namespaces with completely repeated information - it may be a lot of information to view in the library UI for the sake of consistency (which is important) - maybe deserves UX checkin?

aparajit-pratap: Regarding point 3, I too am not so sure if it is necessary to have the library UI display the full type names, I think it can be left as-is, i.e. just `Color` instead of `DSCore.Color`. @Amoursol any thoughts?

QilongTang: @aparajit-pratap @mjkkirschner Well, we do not have a choice I believe, if we touch the OutputType of NodeModelNodes, they will be reflected in Library. As a result, Color.Range will display output as `DSCore.Color` which is different than Color.Red. That was the reason why I started to think of alternatives like this PR.

Amoursol: If there are going to be major functionality changes (i.e. how it displays in the Library) can we set up a quick meeting to discuss? I would also like to have Jingyi present too :)

mjkkirschner: @QilongTang is that correct - for example - 
Screen Shot 2021-04-16 at 5 33 23 PM

This is a ZT node of course, but it does not show the fully qualified type name, do NodeModel nodes act differently? If so that sounds like something to make consistent.

QilongTang: @mjkkirschner See the screenshot I provided above for ColorRange, input is `DSCore.Color[]` and output is `Color`.

mjkkirschner: yes, thats my point @QilongTang - your example is a NodeModel - maybe it should behave like the zero touch node's in the library (my screenshot) - instead. Or vice versa.... though lots of repeated information seems superfluous.

QilongTang:
Closing this PR since a preferred alternative solution has been discussed

PR #11622, opened 2021-04-16
DYN-3572 std.lib should be visible in manage node and package path UI

sm6srw:
### Purpose

[DYN-3572](https://jira.autodesk.com/browse/DYN-3572) std.lib should be visible in manage node and package path UI

Note: This will also fix [DYN-3570](https://jira.autodesk.com/browse/DYN-3570) and partially fix [DYN-3571](https://jira.autodesk.com/browse/DYN-3571) as it will stop Packet manager from downloading packages to the Standard Library directory (but it is till possible to download to the `%ProgramData%` directory).

This pull request does:
* filters out the standard library directory token from the directory list in in `PathManager`.  
* add the standard library directory token as the first folder in `DynamoModel`.
* insert the standard library directory token as the first folder if missing in `DynamoModel`.
* add `Standard Library` to the resources in `PackagePathViewModel` so it can be translated.
* substitute the standard library token with the `Standard Library` resource when the dialog initializes.
* prevent the standard library entry from being deleted in `PackagePathViewModel`.
* prevent the standard library entry from being updated in `PackagePathViewModel`.   
* update the wording in the Package Path dialog.
* show the Standard Library with Italic font type if Standard Library is disabled.
* substitute the translated `Standard Library` resource with the standard library token when changes are committed in the dialog.
* make sure that the `DefaultPackagesDirectory` property points to the first non standard library entry in `PackageLoader`
* add standard library first if missing in `PackageLoader`
* substitute the standard library token with the runtime standard library directory in `PackageLoader`

Normal dialog:
image

Dialog when Standard Library is disabled:
image

 

### Declarations


aparajit-pratap:
image
In this dialog, can we change the description to say that the first path below "Standard Library" is the default save location. Can we also explain that the Standard Library is the location from where packages will be loaded first followed by the rest.

sm6srw:
> image
> In this dialog, can we change the description to say that the first path below "Standard Library" is the default save location. Can we also explain that the Standard Library is the location from where packages will be loaded first followed by the rest of the paths.

Good point. I totally missed that text. I will try to come up with some wording that covers the new behavior.

src/DynamoCoreWpf/ViewModels/PackageManager/PackagePathViewModel.cs
```diff
@@ -184,12 +199,35 @@ private void RemovePathAt(int index)
...
+        }
+
+        private List<string> CommitRootLocations()
+        {
+            var rootLocations = new List<string>(RootLocations);
+            var index = rootLocations.IndexOf(Resources.PackagePathViewModel_Standard_Library);
+
+            if (index != -1)
+            {
+                rootLocations[index] = StandardLibraryToken;
```
aparajit-pratap, line 227: What does it mean for the list of root locations to contain the `StandardLibraryToken` as opposed to `Resources.PackagePathViewModel_Standard_Library` string?

src/DynamoPackages/PackageLoader.cs
```diff
@@ -75,15 +75,17 @@ public IEnumerable<IExtension> RequestedExtensions
 
         /// <summary>
         /// Returns the default package directory where new packages will be installed
-        /// The default package directory is currently the second entry in the list
+        /// This is the first non standard library directory
```
aparajit-pratap, line 78: Won't this always be the second one?

src/DynamoPackages/PackageLoader.cs
```diff
@@ -108,6 +110,10 @@ internal string StandardLibraryDirectory
                 }
             }
         }
+
+        // Token representing the standard library directory
+        internal static readonly string StandardLibraryToken = @"%StandardLibrary%";
```
aparajit-pratap, line 115: Use the one from `DynamoModel`?

src/DynamoPackages/PackageLoader.cs
```diff
@@ -148,10 +155,34 @@ private void InitPackageLoader(IEnumerable<string> packagesDirectories, string s
...
+
+            // Setup Default Package Directory
+            if (standardLibraryIndex == -1)
+            {
+                defaultPackagesDirectoryIndex = this.packagesDirectories.Count > 1 ? 1 : -1;
+            }
+            else
+            {
+                var safeIndex = this.packagesDirectories.Count > 1 ? 1 : -1;
+                defaultPackagesDirectoryIndex = standardLibraryIndex == 1 ? 0 : safeIndex;
```
aparajit-pratap, line 183: When will the `standardLibraryIndex `be something other than 0?

mjkkirschner:
> > image
> > In this dialog, can we change the description to say that the first path below "Standard Library" is the default save location. Can we also explain that the Standard Library is the location from where packages will be loaded first followed by the rest of the paths.
> 
> Good point. I totally missed that text. I will try to come up with some wording that covers the new behavior.


this is actually covered in:
https://jira.autodesk.com/browse/DYN-3571

but fixing it now is even better.

I think it may ultimately change, but it's good to make it as correct as possible IMO.

mjkkirschner:
@sm6srw - theres one AC that was included in the task that I don't see in your description:
`if std.lib is disabled, the entry in the ui should appear disabled` - what do you think of this one?

src/DynamoPackages/PackageLoader.cs
mjkkirschner: No, because the package search paths can be manually sorted by users to configure load order of packages.

src/DynamoPackages/PackageLoader.cs
mjkkirschner: as above, when a user moves it down using the manage node and package path UI.

mjkkirschner:
This looks pretty solid to me - one thought that occurs to me is that depending on our final strategy for dealing with conflicts during package install (as opposed to Dynamo Startup) - that we may reevaluate std.lib's location (first or last), additionally feedback from UX may lead us to rethink this anyway.

Given that, do you think the position is difficult to change after this PR or it will be pretty straightforward? - I was wondering if the logic could be abstracted into a position field somewhere we could set.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -724,7 +727,12 @@ protected DynamoModel(IStartConfiguration config)
             // is no additional location specified. Otherwise, update pathManager.PackageDirectories to include
             // PackageFolders
             if (PreferenceSettings.CustomPackageFolders.Count == 0)
-                PreferenceSettings.CustomPackageFolders = new List<string> { pathManager.UserDataDirectory };
+                PreferenceSettings.CustomPackageFolders = new List<string> { StandardLibraryToken, pathManager.UserDataDirectory };
+
+            if (!PreferenceSettings.CustomPackageFolders.Contains(StandardLibraryToken))
```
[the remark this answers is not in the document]
sm6srw: When a settings file already exists that does not contain the standard library token.

src/DynamoCoreWpf/ViewModels/PackageManager/PackagePathViewModel.cs
sm6srw: The token is what we store in the settings file. It stays the same regardless of what language you run Dynamo in. The resource string is what we display and will say Standard Library in the current language.

src/DynamoPackages/PackageLoader.cs
sm6srw: Depends on where in the list standard library currently are.

src/DynamoPackages/PackageLoader.cs
sm6srw: > @sm6srw - theres one AC that was included in the task that I don't see in your description:
> `if std.lib is disabled, the entry in the ui should appear disabled` - what do you think of this one? We could also leave this for later depending on if we expose those options to the UI as well.

Thanks, I missed this one. I will take a look.

sm6srw:
> @sm6srw - theres one AC that was included in the task that I don't see in your description:
> `if std.lib is disabled, the entry in the ui should appear disabled` - what do you think of this one? We could also leave this for later depending on if we expose those options to the UI as well.

Thanks, I missed this one. I will take a look.

sm6srw:
> This looks pretty solid to me - one thought that occurs to me is that depending on our final strategy for dealing with conflicts during package install (as opposed to Dynamo startup) - that we may reevaluate std.lib's location (first or last), additionally feedback from UX may lead us to rethink this anyway.
> 
> Given that, do you think the position is difficult to change after this PR or it will be pretty straightforward? - I was wondering if the logic could be abstracted into a position field somewhere we could set.

It is a small change. It happens in two places right now. But I have been thinking that it might be possible to remove that logic all together now from the `PackageLoader` with these changes. I think I will take a look at that.

src/DynamoCoreWpf/Properties/Resources.en-US.resx
```diff
@@ -2357,4 +2353,12 @@ Uninstall the following packages: {0}?</value>
   <data name="PackagePathViewModel_Standard_Library" xml:space="preserve">
     <value>Standard Library</value>
   </data>
-</root>
+  <data name="PackagePathViewSummary2" xml:space="preserve">
+    <value>The default save location is the first path that is not the Standard Library location</value>
+    <comment>Package path management dialog content that describes about the first non standard libary entry being the default save location.</comment>
+  </data>
+  <data name="PackagePathViewSummary3" xml:space="preserve">
+    <value>Locations in italic are disabled</value>
```
mjkkirschner, line 2361: italic(s) - I think.

src/DynamoCoreWpf/Properties/Resources.en-US.resx
```diff
@@ -2357,4 +2353,12 @@ Uninstall the following packages: {0}?</value>
   <data name="PackagePathViewModel_Standard_Library" xml:space="preserve">
     <value>Standard Library</value>
   </data>
-</root>
+  <data name="PackagePathViewSummary2" xml:space="preserve">
+    <value>The default save location is the first path that is not the Standard Library location</value>
```
mjkkirschner, line 2357: what do you think about default `install` location instead of `save`?

sm6srw: Sounds good to me.

src/DynamoCoreWpf/Properties/Resources.en-US.resx
sm6srw: Thanks!

src/DynamoCoreWpf/Views/PackageManager/PackagePathView.xaml
```diff
@@ -17,14 +17,14 @@
         d:DesignWidth="300"
         Background="#515151"
         Style="{DynamicResource DynamoWindowStyle}"
-        d:DataContext="{d:DesignInstance ViewModels:InstalledPackagesViewModel, IsDesignTimeCreatable=False}">
+        d:DataContext="{d:DesignInstance ViewModels:PackagePathViewModel, IsDesignTimeCreatable=False}">
```
aparajit-pratap, line 20: Was this a bug before?

src/DynamoCoreWpf/Properties/Resources.en-US.resx
```diff
@@ -2354,4 +2350,15 @@ Uninstall the following packages: {0}?</value>
   <data name="MessageSamePackageDiffVersInStdLib" xml:space="preserve">
     <value>A different version of the package {1} is already installed as part of the Standard Library. {0} cannot override packages in the Standard Library location.</value>
   </data>
-</root>
+  <data name="PackagePathViewModel_Standard_Library" xml:space="preserve">
+    <value>Standard Library</value>
+  </data>
+  <data name="PackagePathViewSummary2" xml:space="preserve">
+    <value>The default install location is the first path that is not the Standard Library location.</value>
+    <comment>Package path management dialog content that describes about the first non standard libary entry being the default save location.</comment>
```
aparajit-pratap, line 2358: How about saying `Package path management dialog content that displays the first non-standard library entry being the default save location.`

src/DynamoCoreWpf/Properties/Resources.resx
```diff
@@ -2354,4 +2354,11 @@ Uninstall the following packages: {0}?</value>
   <data name="MessageSamePackageDiffVersInStdLib" xml:space="preserve">
     <value>A different version of the package {1} is already installed as part of the Standard Library. {0} cannot override packages in the Standard Library location.</value>
   </data>
-</root>
+  <data name="PackagePathViewModel_Standard_Library" xml:space="preserve">
+    <value>Standard Library</value>
+  </data>
+  <data name="PackagePathViewSummary3" xml:space="preserve">
+    <value>Locations in italic(s) are disabled.</value>
+    <comment>Italic locations are disabled. This can currently only be the standard library.</comment>
```
aparajit-pratap, line 2362: Italicized locations are disabled.

src/DynamoCoreWpf/Views/PackageManager/PackagePathView.xaml
sm6srw: Yes, copy/paste error. Looks like it didn't matter but I was seeing compile time warnings.

mjkkirschner: so is the datacontext being reset correctly somewhere else?

mjkkirschner (approved):

src/DynamoCoreWpf/Properties/Resources.resx
sm6srw: Thanks!

PR #11634, opened 2021-04-20
Linter ViewExtension

SHKnudsen:
### Purpose

This PR does:
- Adds new view extension that:
    - Enables users to select between loaded linter extension
    - Displays all Node- and GraphRule issues from the selected linter

#### Linter ViewExtension
LinterView_v2_x175

### Declarations

### Reviewers

@saintentropy @nate-peters @BogdanZavu

### FYIs

@mjkkirschner


src/LintingViewExtension/LinterView.xaml
```diff
@@ -0,0 +1,176 @@
...
+                          Style="{StaticResource SComboBox}"
+                          Height="30"
+                          Margin="0 15" />
+
+                <fa:ImageAwesome Icon="InfoCircle"
+                                 Grid.Column="1"
+                                 Foreground="#808080"
+                                 Width="20"
+                                 Height="20"
+                                 ToolTip="Some Tooltip"/>
```
SHKnudsen, line 145: @nate-peters i forgot what we agreed should go into this `info` icon?

Right now its simply an icon with a tooltip that displays when hovering it.

image

src/LintingViewExtension/Controls/GraphRuleIssueDto.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System.Collections.ObjectModel;
+using Dynamo.Linting.Rules;
+
+namespace Dynamo.LintingViewExtension.Controls
+{
+    public class GraphRuleIssueDto
```
mjkkirschner, line 6: if this really needs to be public can you add a summary, does this really need to be public ? and can you add a summary anyway because I don't know if everyone will be familiar with a dto. Usually I am familiar with them in context of a server - is there an analogous design here?

src/LintingViewExtension/LintingViewExtension.cs
```diff
@@ -0,0 +1,109 @@
...
+using Dynamo.Linting;
+using Dynamo.LintingViewExtension.Properties;
+using Dynamo.ViewModels;
+using Dynamo.Wpf.Extensions;
+
+namespace Dynamo.LintingViewExtension
+{
+    public class LintingViewExtension : ViewExtensionBase, IExtensionStorageAccess
+    {
+        private const string EXTENSION_NAME = "Dynamo Linter";
```
mjkkirschner, line 16: If they have not already, it would probably be good to consult with UX on naming for this extension.

src/LintingViewExtension/LintingViewExtension.csproj
```diff
@@ -0,0 +1,130 @@
...
+    <AssemblyName>LintingViewExtension</AssemblyName>
+    <TargetFrameworkVersion>v4.8</TargetFrameworkVersion>
+    <FileAlignment>512</FileAlignment>
+    <Deterministic>true</Deterministic>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+    <DebugSymbols>true</DebugSymbols>
+    <DebugType>full</DebugType>
+    <Optimize>false</Optimize>
+    <OutputPath>..\..\bin\AnyCPU\Debug\</OutputPath>
```
mjkkirschner, line 21: I think these projects should probably be setup with the imported outputs and assembly shared info generator, unless you want this assembly to have a different version number purposefully.

src/LintingViewExtension/Controls/GraphRuleIssueDto.cs
SHKnudsen: This does not need to be public no.

In terms of the name, I think I just hit the big old "Don't know what to call this" wall!
I'll revisit it and try to make the name more obvious.

src/LintingViewExtension/LintingViewExtension.cs
SHKnudsen: Yea mostly a placeholder name for now, was going to ask @nate-peters about this.

src/LintingViewExtension/Controls/GraphRuleIssueDto.cs
SHKnudsen: Might have been a bit too quick with the not needing to be public thing..

As I'm binding a collection of these issues in the view that property needs to be public and therefore it can't be a collection of internal types. I made all Rule issues share a common interface `IRuleIssue` and make the public collections of type `IRuleIssue`.

This means that both `GraphRuleIssue` and `NodeRuleIssue` can be internal. The `IRuleIssue` will still need to be public though.

src/LintingViewExtension/Controls/IRuleIssue.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using Dynamo.Linting.Rules;
+
+namespace Dynamo.LintingViewExtension
+{
+    public interface IRuleIssue
```
BogdanZavu, line 7: I taught it will be enough to rely solely on rule evaluation results that are being provided by the Linter manager. We can enhance those results if needed. I was hoping the view would only need to be aware of the linter manager and its results to do its job. Do we still need IRuleIssue when we have IRuleEvaluationResults ?

SHKnudsen: It is still only aware of the LinterManager and its results. The RuleIssues are essentially wrapper models that make it easier for me to create the UI bindings. In the UI i need to display multiple results under the same issue group, so every time a result comes in I check if the result belongs to an issue section that is already created, if it does i simply add the affected nodes to that section and if it doesn't i create a new one. 

Im not sure how easy it would be to rely on the EvaluationResults for the UI, in my mind there would need to be something to summarize them all and make sure they are grouped together correctly in the UI.

The interface is simple to hide the two concrete implementations `NodeRuleIssue` and `GraphRuleIssue`, so instead of making both of them public i can use the public interface (As im binding the UI to a colleciton of these they will need to be public, otherwise binding dosent work).

SHKnudsen:
Made some updates to the UI, FYI @nate-peters 
image

src/LintingViewExtension/LinterView.xaml
nate-peters: The idea was that icon would link to some documentation about Linting and Graph Rules in the new Dynamo Documentation Browser - @mjkkirschner and I talked about this yesterday and I can make a separate PR that adds the documentation in the expected places. Lets just leave the button here for now and I'll hook it up to the docs later. The tooltip should say "What is a Graph Type?"

src/LintingViewExtension/LintingViewExtension.cs
nate-peters: Yeah, we need to make a final call on this... I still think calling it a "Linter" in a user-facing way feels wrong. Let me talk to a few of the designers and I'll get back to you by the end of the day.

src/LintingViewExtension/LinterView.xaml
SHKnudsen: @nate-peters see this comment https://github.com/DynamoDS/Dynamo/pull/11634#issuecomment-828378420. For now i added a simple placeholder html file that will open when the button is pressed.

src/LintingViewExtension/LintingViewExtension.cs
nate-peters: Let's call this "Graph Status", as opposed to "Dynamo Linter". Elizabeth and I looked at the design for a bit this morning and agreed that we'll probably end up adding more functionality to this extension over time, so we should keep the name broad for now. And we can leave "Graph Type" as-is.

src/LintingViewExtension/Docs/LinterViewHelpDoc.html
```diff
@@ -0,0 +1,185 @@
+﻿<!--
```
nate-peters, line 1: Let's follow the pattern introduced in the Documentation Browser PR for how we handle our `.html` doc files: 
https://github.com/DynamoDS/Dynamo/pull/10343


So this file should move to `src/DocumentationBrowserViewExtension/Docs/LinterViewHelpDoc.html`

... and then embed the file in the DocumentationBrowserExtension.dll by adding a reference here: https://github.com/DynamoDS/Dynamo/blob/e736f30095cc2b554008f6b1e1fa762244c63e33/src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.csproj#L159

I can come back and replace the lorem ipsum in a follow-up PR.

BogdanZavu:
@SHKnudsen since we merged the linter manager branch let's rebase this one

src/LintingViewExtension/Docs/LinterViewHelpDoc.html
SHKnudsen: Technically there shouldnt be any need to put them into the DocumentationBrowsers `doc` folder, think that is mostly for all the expanded error messages and some generic content (at least it was like that back when we did the doc browser). 

IMO it makes sense to keep the help html's alongside what they are intended to describe or help with. We did the same for the python 3 stuff https://github.com/DynamoDS/Dynamo/blob/master/src/Libraries/PythonNodeModels/Docs/PythonMigrationWarning.html

Can easily move it if you prefer, just think its nicer to have it like this.

src/DynamoCore/Linting/LinterManager.cs
```diff
@@ -131,7 +131,8 @@ private void OnRuleEvaluated(IRuleEvaluationResult result)
             }
         }
 
-        private LinterExtensionBase GetLinterExtension(LinterExtensionDescriptor activeLinter)
+
+        internal LinterExtensionBase GetLinterExtension(LinterExtensionDescriptor activeLinter)
         {
             return this.extensionManager.
```
BogdanZavu, line 137: activeLinter might be null

BogdanZavu:
I would say to account for no active linter descriptor as a valid use case.
This will be the initial state I believe once we go live but also users might decide (most likely temporarily) to not have any linting even if they have a few linters available.

src/LintingViewExtension/LinterView.xaml
```diff
@@ -0,0 +1,210 @@
...
+
+        <HeaderedContentControl Header="Graph Type"
+                                HeaderTemplate="{StaticResource HeaderItemControlHeaderTemplate}"
+                                Margin="16 10">
+            <Grid>
+                <Grid.ColumnDefinitions>
+                    <ColumnDefinition Width="*" />
+                    <ColumnDefinition Width="50" />
+                </Grid.ColumnDefinitions>
+                <ComboBox Grid.Column="0"
```
BogdanZavu, line 149: We should have "None" as an option

src/LintingViewExtension/LintingViewExtension.csproj
SHKnudsen: @mjkkirschner hm, should be setup with the Assembly Shared info, but not sure i know about the `imported output paths` where are they coming from?

mjkkirschner: the output paths should be coming from:
https://github.com/DynamoDS/Dynamo/blob/master/src/Config/CS.props#L7
if you are importing this - I think if you define them again here you'll overwite that import if it appears further up in the csproj file.

SHKnudsen: Yea i was importing it, not sure how i ended up overwriting it.. Should be fixed now.

BogdanZavu:
@SHKnudsen based on our last discussion I think there only a few things left before we can merge :
have a guid for None descriptor
remove debugger.break statements and TestLinterExtension project
Now or in the PR from AGD-2060 to master, let's limit api exposure until this becomes available as public api (when we refactor our rules in the extension) by making LinterExtensionBase internal and preventing LinterViewExtension from showing up in the UI.
If I missed anything please add

SHKnudsen:
@BogdanZavu is there a mechanism to prevent the extension to show up?

mjkkirschner:
Aaron implemented a way to disable extensions - I can’t remember where the config is loaded from though

SHKnudsen:
@QilongTang would you be able to help with Mikes comment above? 👆

QilongTang:
> Aaron implemented a way to disable extensions - I can’t remember where the config is loaded from though
> […](#)
> On May 6, 2021, at 2:16 PM, Bogdan Zavu ***@***.***> wrote: ﻿ @SHKnudsen I would say to interfere when all the ViewExtensions are being loaded and skip our extension from the beginning based on its UniquiId. Cant' think of anything better. https://github.com/SHKnudsen/Dynamo/blob/c7e3c19e8b7feacecd8b2b1d9159ad7cc4b446af/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs#L160 — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or unsubscribe.



> @QilongTang would you be able to help with Mikes comment above? 👆

Please refer to the following view extension manifest file: https://github.com/DynamoDS/Dynamo/blob/master/src/NodeAutoCompleteViewExtension/NodeAutoComplete_ViewExtensionDefinition.xml#L4. If you implement this and include the `IsEnabled` field, it can act as a switch to turn on loading this view extension or not. If you do not include this field, the view extension will be treated as always loaded. Happy to answer any more question.

SHKnudsen:
Ah, so easy! thanks, @QilongTang .

@BogdanZavu this should be ready now unless there are additional comments?

BogdanZavu (approved):

PR #11638, opened 2021-04-22
Enable Node Autocompletion for output ports of Dynamo nodes. 

reddyashish:
### Purpose

This PR is to address the task: https://jira.autodesk.com/browse/DYN-3344

This PR will enable the node autocompletion functionality on the output ports of the nodes. The node autocompletion model will find all the matching nodes which has atleast one input parameter matching the output port type of the selected node. When any of the search elements are clicked, it will create the node and connect the output port of the original node to the first available matching input port of the new node. 

autocompletion output ports

**Things to discuss**
1) When no matching search elements are found, do we want to return an empty list? Because the default nodes shown for the input nodes do not have any InPorts so we cannot show them for the output port. 

### Declarations

### Reviewers

@QilongTang @zeusongit @mjkkirschner @aparajit-pratap


mjkkirschner:
@reddyashish maybe when nothing matches we show a list with codeblock, watch node, or python?

src/DynamoCore/Graph/Nodes/PortModel.cs
```diff
@@ -440,6 +440,40 @@ internal string GetInputPortType()
...
+        /// returns a valid type for only Zero Touch, Builtin and NodeModel nodes,
+        /// and returns null otherwise. The string representation of the type also
+        /// contains the rank information of the type, e.g. Point[], or var[]..[]. 
+        /// </summary>
+        /// <returns>output port type</returns>
+        internal string GetOutPortType()
+        {
+            if (PortType == PortType.Input) return null;
+
+            var ztNode = Owner as DSFunction;
```
mjkkirschner, line 456: instead of doing type checking like this using as and var you can probably switch to:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/is#pattern-matching-with-is

maybe a little cleaner.

src/DynamoCore/Graph/Nodes/PortModel.cs
```diff
@@ -440,6 +440,40 @@ internal string GetInputPortType()
...
+            }
+
+            var nmNode = Owner as NodeModel;
+            if (nmNode != null)
+            {
+                var classType = nmNode.GetType();
+
+                var outPortAttribute = classType.GetCustomAttributes().OfType<OutPortTypesAttribute>().FirstOrDefault();
+
+                return outPortAttribute?.PortTypes.ElementAt(Index);
```
mjkkirschner, line 473: you can't guarantee that just because you have the `outputAtrribute` that your index won't be out of bounds. I think `ElementAt` might throw - you may want to bounds check first, or use a try catch.

src/DynamoCore/Search/SearchElements/NodeSearchElement.cs
```diff
@@ -29,6 +29,7 @@ private void OnVisibilityChanged()
             if (VisibilityChanged != null)
                 VisibilityChanged();
         }
+        internal int PortToConnect { get; set; }
```
mjkkirschner, line 32: this seems like an inappropriate place for this data

src/DynamoCore/Graph/Nodes/PortModel.cs
QilongTang: Ah I can see you are following the same approach as `GetInPortType`, do you mind correct both in your PR?
```
 If (ztNode is DSFunction)
```

src/DynamoCore/Graph/Nodes/PortModel.cs
QilongTang: yeah, 3rd party dev or even us could forgot to match the number of `OutPortTypesAttribute`, I have fixed one instance in my other PR

src/DynamoCoreWpf/ViewModels/Core/PortViewModel.cs
```diff
@@ -418,7 +418,7 @@ private bool CanAutoComplete(object parameter)
         {
             DynamoViewModel dynamoViewModel = _node.DynamoViewModel;
             // If the feature is enabled from Dynamo experiment setting and if user interaction is on input port.
-            return dynamoViewModel.EnableNodeAutoComplete && this.PortType == PortType.Input;
+            return dynamoViewModel.EnableNodeAutoComplete;
```
QilongTang, line 421: 👍🏻

src/DynamoCore/Search/SearchElements/NodeSearchElement.cs
QilongTang: I kinda agree, if needed, we may want to inherit this class. I think in the future we may add more properties if need, e.g. favorite, weight, confidence level (after ML kicks in).

QilongTang:
> @reddyashish maybe when nothing matches we show a list with codeblock, watch node, or python?

Due to the convenience of creating CBN, I would maybe just add watch, watch 3D and Python, even watch3D may not be universal. My 2 cents.

src/DynamoCore/Graph/Nodes/PortModel.cs
reddyashish: Changed this to use the "is" operator

src/DynamoCore/Graph/Nodes/PortModel.cs
reddyashish: Added a try-catch block.

src/DynamoCore/Search/SearchElements/NodeSearchElement.cs
reddyashish: Since this data will be specific to each search element in the auto-completion window, do we want to create a new class for these node elements and that would inherit the NodeSearchElement class?

QilongTang: @reddyashish Yeah, I would be supportive

reddyashish:
@mjkkirschner @QilongTang Addressed the comments. 
For the output ports, I am modified it to only show the watch, watch 3D and Python nodes as the CBN node does not have any input parameters.
To store the data for Auto-completion node elements, I created a new class where we can add more attributes in the future. The NodeSearchElement is extending this class as the auto-completion API returns a list of NodeSearchElement and we would need this information in the NodeSearchElementViewModel to connect the port. What do you think about it?

src/DynamoCore/Search/SearchElements/AutoCompletionNodeSearchElement.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Linq;
+using Dynamo.Configuration;
+using Dynamo.Graph.Nodes;
+
+namespace Dynamo.Search.SearchElements
+{
+    /// <summary>
+    ///      Class for all Auto-Completion Node search elements.
+    /// </summary>
+    public abstract class AutoCompletionNodeSearchElement
```
QilongTang, line 12: Can you re-iterate the purpose of this class by comments here? We cant inherit the NodeSearchElement class because auto-completion API returns a list of NodeSearchElement and we would need this information in the NodeSearchElementViewModel to connect the port? I am fine with this, @mjkkirschner @aparajit-pratap what do you think? I think we plan to keep expanding here because we will have more properties coming in specific to AutoComplete function, e.g. user's rank, whether favored, ML rank

mjkkirschner (approved):

PR #11643, opened 2021-04-22
localized packages tests

mjkkirschner:
### Purpose

Even though we don't (well) support publishing packages with localized content currently - we do load it correctly if it exists in the package's bin folder under culture specific folders. 

This PR is intended to add tests so that it doesn't break without us knowing.

- [x] copies the DynamoSamples resources binaries from that repo to the Dynamo test folder.
- [x] will add tests that assert correct resources are loaded based on thread ui culture for dynamo sample package node.
- [x] stub out a std.lib document in the integration docs folder and add info on package localization to that stub.

### Declarations


mjkkirschner:
I've opened this PR as WIP to see if there are any issues with the existing tests now that I've added the resource binaries.

sm6srw (approved):
LGTM!

aparajit-pratap (commented):
Does the resources dll need to be here `test/pkgs/Dynamo Samples/bin/SampleLibraryUI.resources.dll` as well as in the culture folders?

mjkkirschner:
@aparajit-pratap good catch- I'm not sure how that binary got that, I'll delete it.

mjkkirschner:
There are also some failing tests here - I'm looking into them.

mjkkirschner:
ugh - if I'm reading https://github.com/DynamoDS/Dynamo/pull/9306 right - this version of the SampleLibraryUI.dll that I replaced had been modified in that PR and had extra attributes for performing migrations ... 

I can think of two solutions
1. decompile the old dll and merge in my new class?
2. rename the old package to clearly illustrate its not the original dynamo samples package.

I'm most concerned about package collisions during testing time.

mjkkirschner:
I've tried some rename hackery to keep the old modified package around... if tests don't pass this round, maybe it's safer to just add a completely new package for the localization test so nothing collides.

running tests here as well to check all in less processes: https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/

mjkkirschner:
okay, undid the hackery - I decompiled old dll - moved it here: https://github.com/mjkkirschner/DynamoSamples/tree/migrationTest (eventually will make a new branch DynamoSamples) and have this branch compile to a new package with a new namespace `MigrationTesting`

Then I updated old tests that used those weird types to instead use the regular DynamoSamples types.

running tests now to see what we get.

mjkkirschner:
still some failures, looking at them tomorrow.

aparajit-pratap (approved):

mjkkirschner:
2 tests needed to be updated since contents of updated DynamoSamples repo has changed a bit in 2 years - should be good now, but waiting for another test run.

PR #11645, opened 2021-04-27
[DYN-3579] Preferences General tab functionality

Astul-Betizagasti:
### Purpose

Add functionality to the controls on the Preferences window's General tab

### Declarations

### Reviewers

Aaron Tang (@QilongTang )

### FYIs


src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -382,7 +382,10 @@ public string DefaultPythonEngine
         /// If enabled user's custom package locations will not be loaded.
         /// </summary>
         public bool DisableCustomPackageLocations { get; set; }
-
+        /// <summary>
+        /// If true the run mode will be set to Automatically, if false to Manual.
+        /// </summary>
+        public bool RunTypeAutomatic { get; set; }
```
QilongTang, line 388: Can we serialize this info using string so we can just borrow `enum RunType { Manual, Automatic, Periodic }`? What would be the trade from your point of view?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -2591,6 +2593,12 @@ public void ClearCurrentWorkspace()
             OnWorkspaceClearing();
 
             CurrentWorkspace.Clear();
+            if (CurrentWorkspace is HomeWorkspaceModel)
+            {
+                _ = PreferenceSettings.RunTypeAutomatic
```
QilongTang, line 2598: can you add comments for this block?

src/DynamoCoreWpf/UI/Converters.cs
```diff
@@ -1840,18 +1840,59 @@ public object ConvertBack(object value, Type targetType, object parameter, Cultu
...
+                case "f1":
+                    return Resources.DynamoViewSettingMenuNumber00;
+                case "f2":
+                    return Resources.DynamoViewSettingMenuNumber000;
+                case "f3":
+                    return Resources.DynamoViewSettingMenuNumber0000;
+                case "f4":
+                    return Resources.DynamoViewSettingMenuNumber00000;
+                default:
+                    return null;
```
QilongTang, line 1875: I feel we should also have a bunch of legacy code we can obsolete/ delete here? The old number format comes with a dialog, let's delete it?

src/DynamoCore/Models/DynamoModel.cs
Astul-Betizagasti: sure, will do on next update

src/DynamoCore/Models/RunSettings.cs
```diff
@@ -106,7 +106,6 @@ public RunSettings(RunType runType, int period)
         internal void Reset()
         {
             RunEnabled = true;
-            RunType = RunType.Automatic;
```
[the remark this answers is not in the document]
Astul-Betizagasti: this function is called every time a "new" home workspace is open (actually it is the same workspace that gets reset every time) , and it was overriding the user's default run setting.

src/DynamoCoreWpf/UI/Converters.cs
Astul-Betizagasti: Hello Aaron, can you explain further? Previously the node format was chosen from a sub-menu in the Settings Dynamo menu which we removed already. This preference is saved on the PreferencesSettings with the codes "f1", "f2", etc but shown to the user as 0.0, 0.00, etc, so this converter is used to convert back and forth between the two formats.

src/DynamoCore/Configuration/PreferenceSettings.cs
Astul-Betizagasti: This variable is the one that saves the user's Default run setting, and being that the only options the user have are Manual or Automatic it doesn't feel right having the option to save Periodic on it too. On the other hand I believe using the enum as you suggested would make it more manageable in terms of code. I suppose the best option here depends on what we value the most, what I can say for sure is that the name of the property should be more representative of what it saves so I will change it on the next update to something more fitting

QilongTang: @Astul-Betizagasti we can use `IsDefaultRuntypeAuto` etc, but that still looks less intuitive to me and less expandable. I prefer we serialize the actual type, even it is only going to either `Automatic` or `Manual` for now, but this allows us to expand in the future, e.g. when Dynamo become a web service, user could have time based run type etc.

src/DynamoCoreWpf/UI/Converters.cs
QilongTang: ah OK, my bad then that I did not notice we are already re-using the resources

src/DynamoCore/Configuration/PreferenceSettings.cs
Astul-Betizagasti: Make sense to me, I will update that

src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -382,7 +383,10 @@ public string DefaultPythonEngine
         /// If enabled user's custom package locations will not be loaded.
         /// </summary>
         public bool DisableCustomPackageLocations { get; set; }
-
+        /// <summary>
+        /// If true the run mode will be set to Automatically, if false to Manual.
```
QilongTang, line 387: Please update the summary as well

Astul-Betizagasti: good catch, updated!

QilongTang (approved):
LGTM

QilongTang:
@Astul-Betizagasti Do you mind adding unit test for `RunSetting`, you can search for tests for preference setting which cover serialization and deserialization.

Astul-Betizagasti:
@QilongTang, I just made a commit addressing the changes to the tests as you asked. The new property was added as part of the checks made by the existing test related to serializing the preferences settings

QilongTang:
@Astul-Betizagasti Thanks!

PR #11653, opened 2021-04-28
Fix node autocomplete suggetions on custom nodes

zeusongit:
### Purpose

The node-autocomplete feature was unable to identify the input type for custom nodes, thereby displaying default suggestions for all custom node inputs. That was fixed by casting node model to appropriate custom node class and extracting the type info of the input port to be used by the node-autocomplete search. 

cusac2

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCore/Graph/Nodes/PortModel.cs
+++ b/src/DynamoCore/Graph/Nodes/PortModel.cs
@@ -429,9 +429,8 @@ namespace Dynamo.Graph.Nodes
                 }
                 return type;
             }
-
-            var cusNode = Owner as CustomNodes.Function;
-            if (cusNode != null)
+          
+            if (Owner is CustomNodes.Function cusNode)
             {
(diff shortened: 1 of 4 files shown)


src/DynamoCore/Graph/Nodes/PortModel.cs
```diff
@@ -430,6 +430,16 @@ internal string GetInputPortType()
                 return type;
             }
 
+            var cusNode = Owner as CustomNodes.Function;
+            if (cusNode != null)
```
QilongTang, line 434: I think @mjkkirschner suggested to use a `is` check on @reddyashish 's PR, you may want to give it a try to simplify the code as well here

QilongTang:
@zeusongit Would you take a look as to what is failing on self serve? Otherwise, looks good to me

QilongTang (approved):

reddyashish (approved):

PR #11658, opened 2021-04-30
Update Helix Toolkit from 2.11 to 2.15

saintentropy:
### Purpose

This PR is to update the Helix Toolkit (and by association the Cyotek.Drawing.BitmapFont dependency) from version 2.11 to version 2.15. The purpose of this PR is to address a memory optimization related to the https://github.com/DynamoDS/Dynamo/pull/11395 PR (Generating a single render package per node).  Specifically the https://github.com/helix-toolkit/helix-toolkit/pull/1494 PR modifies the behavior of the `VertexArrayBuffer` which was being retained in memory as a sparse array based on the largest size mesh to be rendered in the session.  In 2.15 this buffer is not pinned to the largest allocation size in session.    

As a point of interest.  This Helix PR fix was triggered due to the issue filed by @aparajit-pratap.  https://github.com/helix-toolkit/helix-toolkit/issues/1449

### Declarations

### Reviewers

@aparajit-pratap @mmisol 

### FYIs

@mjkkirschner


mjkkirschner:
I must be misunderstanding that PR - because it still looks like the default is to create a buffer(vertexCount) ?

Also since the default for `isTransient` is false, we're not taking advantage of the buffer be cleared - and if we were, then hit testing would be broken it seems.

saintentropy:
@mjkkirschner for our purposes the new `ThreadBufferManager` is what affects us. Here are the relevant changes to the `MeshBufferModel` and `LineBufferModel`:

image

and
image

The original local `vertexArrayBuffer` was always retained and would just keeping getting larger based on the largest array of the session.  Now per their documentation -> per their notes _/// If requested size is larger than this value, buffer will be temporary instead of being retained for reuse._

saintentropy:
The `isTransient` optimization can be looked at in the future but not related to the memory issue we were addressing

mjkkirschner (approved):

mjkkirschner:
It looks good to me, but it's probably good to run the view comparison tests.

saintentropy:
Image comparisons are pass @mjkkirschner @aparajit-pratap

PR #11666, opened 2021-05-05
Return data instead of void for export node

QilongTang:
Please Note:

### Purpose

Return bool instead of void for node `OpenXMLExportExcel` to indicate of the node has successfully written data into spreadsheet

image

### Declarations

### Reviewers

@mmisol 

### FYIs

### DIFF
--- a/src/Libraries/DSOffice/Excel.cs
+++ b/src/Libraries/DSOffice/Excel.cs
@@ -902,9 +902,10 @@ namespace DSOffice
         /// <param name="overWrite"> Toggle to clear spreadsheet before writing</param>
         /// <param name="writeAsString">Toggle to switch between writing cell values as strings</param>
         /// <search>office,excel,spreadsheet</search>
-        public static void OpenXMLExportExcel(string filePath, string sheetName, object[][] data, int startRow = 0, int startColumn = 0, bool overWrite = false, bool writeAsString = false)
+        /// <returns>Cell values to write to the spreadsheet</returns>
+        public static object[][] OpenXMLExportExcel(string filePath, string sheetName, object[][] data, int startRow = 0, int startColumn = 0, bool overWrite = false, bool writeAsString = false)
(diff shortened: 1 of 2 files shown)


mmisol (changes requested):
Hi @QilongTang . I'm afraid this approach is not compatible with what the `Data.ExportToExcel` node does. That was my reason for returning `void` in the first place. Allow me to illustrate with an example.

Starting with this spreadsheet:
Excel_Sheet

And using this graph to only write the "C3" cell:
Graph

You can see in the watch node that the `Data.ExportToExcel` node returns all of the spreadsheet data. However, if I tried this with the `Data.OpenXmlExportExcel` node I would only get the passed in data, that is `20`.

Given the differences in the implementations of the nodes, I don't think there is an easy way to do this, other than doing what the `Data.OpenXmlImportExcel` node does, but that might be overkill considering the spreadsheet data might no be needed after written and also users can always use the `Data.OpenXmlImportExcel` node itself if they do need it.

mjkkirschner:
What about writing a success or failure bool. Void seems like a bad pattern in a data flow language’s library.

QilongTang:
@mmisol @mjkkirschner I can change the implementation so that we always follow up the write request with a read function call so that this node could return all the data. This may not be 100% ideal performance wise but will be purely done for consistency, what do you think?

mjkkirschner:
@QilongTang can we characterize what the performance cost is?

mmisol:
@mjkkirschner That would depend on the size of the sheet. If we are dealing with one that is very large, it could take a few extra seconds, plus a chunk of memory. I think returning something like a `bool` should work too, so if it's more idiomatic for Dynamo I'm fine with that.

mmisol (approved):
Thanks @QilongTang . This looks good to me

QilongTang:
Merging since all checks passing

PR #11668, opened 2021-05-05
Fix the crash when Autocompletion is triggered on output port of watch node.

reddyashish:
### Purpose

This PR is to fix the crash that was happening when Autocompletion was triggered on output port of watch node. The reason for the crash was the output port type returned for the Watch node is something like "Node Output:var". When this was passed into the parser, it would throw an exception. So we will just be stripping that unnecessary string from the port type. Aabishkar has found that this was happening only for this particular node. I tried looking any other similar nodes but couldn't find any. 

For now, we will be showing the default output suggestions for the Watch node.

### Declarations

### Reviewers

@QilongTang @zeusongit @mjkkirschner

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
@@ -157,8 +157,6 @@ namespace Dynamo.ViewModels
 
             var core = dynamoViewModel.Model.LibraryServices.LibraryManagementCore;
 
-            portType = StripUnwantedStringFromOutputType(portType);
-
             //if inputPortType is an array, use just the typename
             var parseResult = ParserUtils.ParseWithCore($"dummyName:{ portType};", core);
             var ast = parseResult.CodeBlockNode.Children().FirstOrDefault() as IdentifierNode;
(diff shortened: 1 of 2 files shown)


josh2kand8:
Good

QilongTang:
hi @reddyashish , I think we can also try trim the output port type at https://github.com/DynamoDS/Dynamo/blob/master/src/Libraries/CoreNodeModels/Watch.cs#L19 to be just `var` so you dont have to trim it for every node again using these changes. Can you give that a try?

reddyashish:
Yes that would be much better. Do you think changing the output port type of the watch node to just "var" cause any behavior change. I wasn't completely sure about it.

QilongTang:
> Yes that would be much better. @QilongTang Do you think changing the output port type of the watch node to just "var" cause any behavior change. I wasn't completely sure about it.

I think it should be fine. We can open a different PR for that and use self serve to validate

reddyashish:
I have made the changes in this PR only and running the self-serve. If anything fails, I will revert it.

QilongTang (approved):

QilongTang:
> I have made the changes in this PR only and running the self-serve. If anything fails, I will revert it.

Thanks, if everything pass, feel free to merge

reddyashish:
The self-serve validation has passed and tested it manually. Merging this.

PR #11671, opened 2021-05-07
[DYN-3657] Save changes label

Astul-Betizagasti:
### Purpose

Change the current Save Changes button on the Preferences modal into a label that will display information so the user can know when changes where saved

Includes a change to make the first expander on the Features tab be expanded by default when opening the Preferences modal
savechanges

### Declarations

### Reviewers

Aaron Tang (@QilongTang )

### FYIs


src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -580,6 +630,16 @@ public PreferencesViewModel(DynamoViewModel dynamoViewModel)
                                                                                               scaleRanges[GeometryScaleSize.Large].Item3));
             optionsGeometryScal.DescriptionScaleRange.Add(string.Format(Res.ChangeScaleFactorPromptDescriptionContent, scaleRanges[GeometryScaleSize.ExtraLarge].Item2,
                                                                                               scaleRanges[GeometryScaleSize.ExtraLarge].Item3));
+
+            SavedChangesLabel = string.Empty;
+            SavedChangesTooltip = string.Empty;
+        }
+
+        internal void UpdateSavedChangesLabel()
```
QilongTang, line 638: Let's add some comments to this function

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -87,6 +122,7 @@ public string SelectedLanguage
             {
                 selectedLanguage = value;
                 RaisePropertyChanged(nameof(SelectedLanguage));
+                UpdateSavedChangesLabel();
```
QilongTang, line 125: This call is repeated too many times. We can add a property change listener so we can listen to changes and call this automatically. That way we can get rid of all these single calls.

QilongTang (changes requested):
LGTM in general, it would be good that we can address the so many calls of `UpdateSavedChangesLabel()`

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
QilongTang: Something like here: https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs#L636. Single handler can handle all property change I think, then we can filter out properties we need and call `UpdateSavedChangesLabel()` there

QilongTang:
To avoid confusion on customer demo today, I am merging this in order to get rid of the save button. But @Astul-Betizagasti please address the PR comments using a new one.

PR #11675, opened 2021-05-11
Graph properties view extension

SHKnudsen:
### Purpose

This PR adds a new extension that enables setting graph properties from a UI, the extension currently supports these properties:
- `Description`
- `Author`
- `GraphDocumentationURL`
- `Thumbnail`

in addition to the above graph properties, the extension also allows storing CustomProperties in the graphs ExtensionWorkspaceData. The UI adds simple controls to add new custom key/value properties.

### UI
GraphPropertiesExt_x250

### Opening graph with properties set
GraphPropertiesExt_Open

### json sample after setting properties
```javascript
{
  "Uuid": "bd2ba3e8-f006-409b-a34e-98aaeea3e5fe",
  "IsCustomNode": false,
  "Description": "Some description about the graph.......",
  "Name": "CustompropertyTest",
  "ElementResolver": {
    "ResolutionMap": {}
  },
  "Inputs": [],
  "Outputs": [],
  "Nodes": [],
  "Connectors": [],
  "Dependencies": [],
  "NodeLibraryDependencies": [],
  "Thumbnail": "/9j/4AAQSkZJRgABAQEASABIAAD/2wBDAAMCAgMCAgMDAwMEAwMEBQgFBQQEBQoHBwYIDAoMDAsKCwsNDhIQDQ4RDgsLEBYQERMUFRUVDA8XGBYUGBIUFRTUDQsNFBQ.....
  "GraphDocumentationURL": "https://dynamobim.org/",
  "ExtensionWorkspaceData": [
    {
      "ExtensionGuid": "28992e1d-abb9-417f-8b1b-05e053bee670",
      "Name": "Properties",
      "Version": "1.0",
      "Data": {
        "My prop 1": "My value 1",
        "My prop 2": "My Value 2",
        "Custom Property 3": ""
      }
    }
  ],
  "Author": "Name of author",
  "Bindings": [],
  "View": {
    "Dynamo": {
      "ScaleFactor": 1.0,
      "HasRunWithoutCrash": true,
      "IsVisibleInDynamoLibrary": true,
      "Version": "2.12.0.4955",
      "RunType": "Automatic",
      "RunPeriod": "1000"
    },
    "Camera": {
      "Name": "Background Preview",
      "EyeX": -17.0,
      "EyeY": 24.0,
      "EyeZ": 50.0,
      "LookX": 12.0,
      "LookY": -13.0,
      "LookZ": -58.0,
      "UpX": 0.0,
      "UpY": 1.0,
      "UpZ": 0.0
    },
    "NodeViews": [],
    "Annotations": [],
    "X": 0.0,
    "Y": 0.0,
    "Zoom": 1.0
  }
}
```

### Declarations

### Reviewers

@BogdanZavu @mjkkirschner 

### FYIs

@saintentropy @nate-peters


BogdanZavu:
Is there a design agreement already or it is still under discussion ?

SHKnudsen:
> Is there a design agreement already or it is still under discussion ?

Ill leave it to @nate-peters to answer this, but here is the mockups its based on https://www.figma.com/file/u0q3lZqxa0CxbDkZDABwhI/Dynamo-Graph-Documentation-Tools?node-id=221%3A1057

src/GraphMetadataViewExtension/ValidationRules/UriValidationRule.cs
```diff
@@ -0,0 +1,30 @@
...
+        public override ValidationResult Validate(object value, System.Globalization.CultureInfo cultureInfo)
+        {
+            string input = value as string;
+
+            if (String.IsNullOrEmpty(input)) // Valid input, converts to null.
+            {
+                return new ValidationResult(true, null);
+            }
+
+            if (Uri.TryCreate(input, UriKind.Absolute, out Uri outUri))
```
BogdanZavu, line 21: I think se should probably allow www.adress.com. Currently we enforce http://www

src/GraphMetadataViewExtension/GraphMetadataView.xaml
```diff
@@ -0,0 +1,194 @@
...
+    </UserControl.Resources>
+    
+    <Grid Margin="10px 10px">
+        <Grid.RowDefinitions>
+            <RowDefinition Height="Auto" />
+            <RowDefinition Height="*" />
+        </Grid.RowDefinitions>
+        <StackPanel Orientation="Vertical"
+                    Grid.Row="0">
+            <GroupBox Header="Description"
```
BogdanZavu, line 118: I think this should start from a single line and be auto resizable like we have it in generative design tools

src/GraphMetadataViewExtension/GraphMetadataView.xaml
```diff
@@ -0,0 +1,194 @@
...
+                <RowDefinition Height="Auto" />
+                <RowDefinition Height="*" />
+            </Grid.RowDefinitions>
+            <Grid Grid.Row="0"
+                  Margin="10px, 5px">
+                <Grid.ColumnDefinitions>
+                    <ColumnDefinition Width="*" />
+                    <ColumnDefinition Width="20" />
+                </Grid.ColumnDefinitions>
+                <TextBlock Grid.Column="0"
```
BogdanZavu, line 169: Can we activate tabbing ? It would be nice to have when dealing with multiple properties

src/GraphMetadataViewExtension/GraphMetadataViewExtension.cs
```diff
@@ -0,0 +1,92 @@
...
+                extensionData[p.PropertyName] = p.PropertyValue;
+            }
+        }
+
+
+        #endregion
+
+        protected virtual void Dispose(bool disposing)
+        {
+            viewModel.Dispose();
```
BogdanZavu, line 80: Why are we disposing the view model ? In theory a view extension could get unloaded and the viewmodel still stands.
And in the shutdown sequence I imagine viewmodel might still be in use for a while after our extension was disposed or ?

BogdanZavu:
We will need to take over a few fields that are currently part of info.json like description and transfer them to the new slot or ?

src/GraphMetadataViewExtension/ValidationRules/UriValidationRule.cs
SHKnudsen: okay, i think the validation becomes kinda pointless then. As it seems `UriKind.RelativeOrAbsolute` will literally accept anything.

src/GraphMetadataViewExtension/GraphMetadataViewExtension.cs
SHKnudsen: > In theory a view extension could get unloaded and the viewmodel still stands.

I dont think this can actually happen? AFAIK the only way to unload an extension is to uninstall the package it comes from and uninstalling a package needs a restart of Dynamo to happen. And if this actually can happen, why would we want the ViewModel of that extension to kept around, i would assume that if you remove an extension everything belonging to that extension should go/

> And in the shutdown sequence I imagine viewmodel might still be in use for a while after our extension was disposed or ?

hm, not sure i see what the viewModel would be in use for? also the dispose on the viewModel only unsubscribe events so in theory everything else is still available.

BogdanZavu: Aaa yes, sorry, I made a confusion, I don't know why I had DynamoViewModel in mind

src/GraphMetadataViewExtension/GraphMetadataViewModel.cs
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        public Uri HelpLink
+        {
+            get { return currentWorkspace.GraphDocumentationURL; }
+            set { currentWorkspace.GraphDocumentationURL = value; RaisePropertyChanged(nameof(HelpLink)); }
+        }
+
+        public BitmapImage Thumbnail
+        {
```
BogdanZavu, line 39: For these image convertors function should we try to guard them better ? And if there is an exception we can show something in the console but allow to program to run.

src/GraphMetadataViewExtension/GraphMetadataViewModel.cs
```diff
@@ -0,0 +1,177 @@
...
+            // CurrentWorkspaceCleared will trigger everytime a graph is closed which allows us to reset the properties. 
+            this.viewLoadedParams.CurrentWorkspaceCleared += OnCurrentWorkspaceChanged;
+
+            CustomProperties = new ObservableCollection<CustomPropertyControl>();
+            InitializeCommands();
+        }
+
+        private void OnCurrentWorkspaceChanged(Graph.Workspaces.IWorkspaceModel obj)
+        {
+            if (!(obj is HomeWorkspaceModel hwm)) return;
```
BogdanZavu, line 73: Should we clear properties in this case as well ? How this happen actually ?

BogdanZavu:
@SHKnudsen I think this is ready, @nate-peters can you have another look at UI-styling ?

nate-peters:
@SHKnudsen @BogdanZavu just a few minor UI comments and then I think we're good to go on the design side.
- Let's use thinner icons for the `X` and `+` symbols. I'll send you the files in Slack.
- It looks like the text in the custom properties is slightly misaligned with the section above - can you left-justify all of the labels in that section?
Screen Shot 2021-05-20 at 12 49 07 PM

src/GraphMetadataViewExtension/GraphMetadataViewModel.cs
```diff
@@ -0,0 +1,177 @@
...
+namespace Dynamo.GraphMetadata
+{
+    public class GraphMetadataViewModel : NotificationObject
+    {
+        private readonly ViewLoadedParams viewLoadedParams;
+        private HomeWorkspaceModel currentWorkspace;
+
+        public DelegateCommand AddCustomPropertyCommand { get; set; }
+
+        public string GraphDescription
```
mjkkirschner, line 20: please add summaries to public members.

src/GraphMetadataViewExtension/GraphMetadataViewModel.cs
SHKnudsen: Changed this a bit, so now if the workspace is changed to something that isnt a `HomeWorkspaceModel` (i believe the only other scenario currently is `CustomNodeWorkspace`) it will close the extension. I dont think we would want to clear the custom properties in this scenario as users might just want to jump into a custom node and then back to the main graph, in that case i assuem they would still want all of the Custom props they created.

However currently all of the custom properties are being reset in this scenario, this is because when switching from Custom Node ws to Home ws will trigger workspace changed which clears all custom properties. 

Lets discuss later today what we expect to happen in that scenario

BogdanZavu: Ok

saintentropy:
LGTM.  Thanks for all the fixes

saintentropy:
@mjkkirschner @BogdanZavu are you done with your reviews on this

BogdanZavu:
There were a few more items from our last discussion but we can make a different PR for them.
One small issue : the dirty flag is not set when a custom property value or name is changed.

BogdanZavu (approved):

mjkkirschner:
@BogdanZavu it looks okay to me - one thing - are you all sure you want this called `Graph` Properties? I ask because other extensions that mention a graph/workspace/dyn file seem to call it `Workspace`?

@Jingyi-Wen FYI.

mjkkirschner (approved):

Jingyi-Wen:
> @BogdanZavu it looks okay to me - one thing - are you all sure you want this called `Graph` Properties? I ask because other extensions that mention a graph/workspace/dyn file seem to call it `Workspace`?
> 
> @Jingyi-Wen FYI.

IMO `Graph Properties` does make perfect sense here.

PR #11689, opened 2021-05-17
DYN-3606-Preferences Panel Expanders Functionality

RobertGlobant20:
### Purpose

I added the functionality in the Features tab of having just one Expander expander at one time (this functionality already exits in the Visual Settings expander).

Also I added the functionality of recording the current state of the Expander in a way that every time that the Preference window is closed and re-opened then the Expanders will remember the previous state. Due that this functionality was implemented by session then when the Dynamo app is closed  the previous Expander settings will be lost.
### Declarations

### Reviewers

@QilongTang 

### FYIs

@Astul-Betizagasti


src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -653,6 +658,10 @@ protected DynamoModel(IStartConfiguration config)
                 PreferenceSettings.PropertyChanged += PreferenceSettings_PropertyChanged;
             }
 
+            //Due that the Expanders settings are stored by session (when dynamo is closed, settings are lost) then we read the settings from a json string
+            string jsonExpandersSettings = Resources.PreferencesExpandersSettings;
```
mjkkirschner, line 662: this seems quite odd - usually in (our repos anyway) resx files are intended for localized content - not data like this.

Why use json at all, why not just have the default constructor set these default values?

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
```diff
@@ -37,9 +38,59 @@ public PreferencesView(DynamoViewModel dynamoViewModel)
                 viewModel = viewModelTemp;
             }
 
+            SetAllExpanderBindings();
+
             InitRadioButtonsDescription();
         }
 
+        /// <summary>
+        /// Due that all the settings for the expanders is stored in DynamoModel in a List property, we need to create expanders bindings dynamically
```
mjkkirschner, line 47: I think you can do a few things as alternatives if you want.
* use dataTemplates and don't manually create each expander, instead store data that represents each one, including the binding.
* use a valueConverter that converts the list of expanders to a specific expander depending on the name of the expander - this will let you add the binding into the xaml if you prefer.

just some other ideas.

src/DynamoCore/Models/DynamoModel.cs
RobertGlobant20: @mjkkirschner 
I did the change so instead of using the json entry I'm initializing the List in the Constructor, see commit: https://github.com/DynamoDS/Dynamo/pull/11689/commits/af8e31e76e11ad15e55a50a7526c57c093df3d6e

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
RobertGlobant20: @mjkkirschner 
Related to the option of using dataTemplates, It will take some effort since we have already 6 expanders in 2 different tabs so the way expanders are created (considering each expander has several controls) will change and probably how the data is bind.
Related to the option of using a converter, I was trying to implement the converter but currently is using two way binding, so at the beginning the values are read from the List<ExpanderSettings> and when a expander is expanded the IsExpanded value is stored in the List<ExpanderSettings>, then when implementing the ConverterBack function I didn't find any way to store the IsExpanded value updated, since it doesn't have access to the List.

The considering the function below, the value parameter is the boolean IsExpanded value and the parameter is the Expander Name, so there is no way to access the List<ExpanderSettings> as in the Convert( ) method.

**ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)**

If you have any idea how to make it work and I can implement it.

Thanks

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -95,6 +95,22 @@ public PreferenceSettings PreferenceSettings
             get { return Model.PreferenceSettings; }
         }
 
+        /// <summary>
+        /// This property will be used by the Preferences screen to store and retrieve all the seetings from the expanders
+        /// </summary>
+        public List<ExpanderSettings> ExpandersSettings
```
QilongTang, line 101: Why dont we store this in Preferences View Model since it does not directly affect DynamoViewModel?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -379,6 +379,11 @@ internal virtual string AppVersion
         /// </summary>
         public readonly PreferenceSettings PreferenceSettings;
 
+        /// <summary>
+        /// This list contains the settings (current status) of the expanders located in the Preferences panel
+        /// </summary>
+        public List<ExpanderSettings> ExpandersSettings;
```
QilongTang, line 385: Why dont we store this in Preferences ViewModel since it does not directly affect DynamoModel?

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
QilongTang: It is unclear to me why the expanders stored in DynamoModel. It makes most sense to me to store them in Preferences ViewModel so that in the future those code maintained as part of PreferencesView. Can we revisit the base of this PR?

RobertGlobant20: @QilongTang In this case that we are trying to record expansion state of each expander per session so I cannot use the PreferencesViewModel because is created just when the Preferences window is opened and destroyed when is closed then the only way to restore/record the expansion state of each expander is by storing the instance in a class that is not destroyed, in this case I decided to use DynamoModel (like DebugSettings or PreferencesSettings instances), I can use other one but it's not clear to me the idea of using PreferencesViewModel.

src/DynamoCore/Models/DynamoModel.cs
RobertGlobant20: @QilongTang Same reason than the previous comment, the PreferencesViewModel is created until the Preferences window is opened so we won't be able to store/recover data for the expanders.

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
QilongTang: hi @RobertGlobant20 , let's do the right thing. We can create preference viewmodel early as part of Dynamo ViewModel creation instead of only construct/destroy it runtime. Then we can store these properties there right?

RobertGlobant20: @QilongTang I did the required changes so now the PreferencesViewModel is created in DynamoViewModel.cs and the properties for the expanders were moved to the PreferencesViewModel. Thanks
See the commit:
https://github.com/DynamoDS/Dynamo/pull/11689/commits/353a4988e1b4bc4cd5110521be9688e79c88a47e

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -95,6 +96,19 @@ public PreferenceSettings PreferenceSettings
             get { return Model.PreferenceSettings; }
         }
 
+        public PreferencesViewModel PreferencesViewModel
```
QilongTang, line 99: Let's make it internal for now

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -621,6 +642,17 @@ public PreferencesViewModel(DynamoViewModel dynamoViewModel)
...
+
+            //Expanders settings are stored by session (when dynamo is closed, settings are lost)
+            ExpandersSettings = new List<ExpanderSettings>
+            {
+                new ExpanderSettings{ Name = "PythonExpander", IsExpanded = false, Tab = "Features" },
+                new ExpanderSettings{ Name = "ExperimentalExpander", IsExpanded = false, Tab = "Features" },
+                new ExpanderSettings{ Name = "Styles", IsExpanded = false, Tab = "Visual Settings" },
+                new ExpanderSettings{ Name = "Scale", IsExpanded = false, Tab = "Visual Settings" },
+                new ExpanderSettings{ Name = "Precision", IsExpanded = false, Tab = "Visual Settings" },
+                new ExpanderSettings{ Name = "Display", IsExpanded = false, Tab = "Visual Settings" },
```
QilongTang, line 654: Do you think we can auto populate this list somehow?

RobertGlobant20: @QilongTang I cannot think of any way of auto populate this list but in order to avoid doing this I re-implemented the functionality of recording/loading the current expansion state of each expander in a List<ExpanderSettings>, now instead of recording each one only records the one expanded (only one Expander can be expanded per each tab), so the info for just one is loaded when the Preferences window is opened and just one info is saved when the Expander expansion state is updated (this approach was suggested by @Astul-Betizagasti ).
For implementing this functionality I added a Converter that will be binded in each Expander.IsExpanded as @mjkkirschner  suggested.
This re-implementation allows to do the same functionality for Expanders with less code. Thanks

See commit:
https://github.com/RobertGlobant20/Dynamo/commit/3924722fe2c1c05c43f566ae31d511f41996ee6c

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
RobertGlobant20: @QilongTang fixed to be internal in commit: https://github.com/RobertGlobant20/Dynamo/commit/b49c1903df29c06fb917c4cf0f2a2e5ba10e85eb

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -77,6 +78,22 @@ public class PreferencesViewModel : ViewModelBase, INotifyPropertyChanged
             {GeometryScaleSize.ExtraLarge, new Tuple<string, string, string>("extra large", "1", "100,000,000")}
         };
 
+        /// <summary>
+        /// This property will be used by the Preferences screen to store and retrieve all the seetings from the expanders
```
QilongTang, line 82: settings

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
```diff
@@ -41,10 +44,40 @@ public PreferencesView(DynamoViewModel dynamoViewModel)
             {
                 viewModel = viewModelTemp;
             }
-
+            
+            //Find all the Expanders in the LogicalTree (XAML)
```
QilongTang, line 48: hmm Do you intend a call of `FindLogicalChildren` here?

RobertGlobant20: @QilongTang I forgot to delete that method because is not used any more, then I deleted the comment and the method
Thanks

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
RobertGlobant20: @QilongTang fixed in commit: https://github.com/DynamoDS/Dynamo/pull/11689/commits/44d7b21da63ceb0a40024976ff633617fb0e2e97

Thanks

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -828,4 +850,41 @@ public static int ConvertScaleFactorToUI(int scaleValue)
...
+    public class TabSettings : INotifyPropertyChanged
+    {
+        public event PropertyChangedEventHandler PropertyChanged;
+        private void OnPropertyChanged(string propertyName)
+        {
+            PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
+        }
+
+        /// <summary>
+        /// Tab Name (e.g. Features o Visual Settings)
```
QilongTang, line 866: `or`?

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
```diff
@@ -302,116 +303,117 @@
...
-                                                       Foreground="{StaticResource PreferencesWindowFontColor}"/>
-                                            </StackPanel>
-                                        </Grid>
-                                    </StackPanel>
-                                </Expander>
-                            </Grid>
+                            <Expander x:Name="PythonExpander"
+                                        Grid.Row="0"
+                                        Style="{StaticResource MenuExpanderStyle}" 
+                                        IsExpanded="{Binding PreferencesTabs[0].ExpanderActive, Converter={StaticResource ExpandersBindingConverter}, ConverterParameter=PythonExpander}"
```
QilongTang, line 309: This design is a bit strange that if we will need to cover expander state for `General` tab in the future, we will need to shift the index here..

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
RobertGlobant20: @QilongTang fixed in commit: https://github.com/DynamoDS/Dynamo/pull/11689/commits/5ee62bb8dcb52c07eef3929916511372bdbcd128

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -625,6 +642,11 @@ public PreferencesViewModel(DynamoViewModel dynamoViewModel)
             SavedChangesTooltip = string.Empty;
 
             this.PropertyChanged += Model_PropertyChanged;
+           
+            preferencesTabs = new List<TabSettings>();
+            preferencesTabs.Add(new TabSettings() { Name = "Features", ExpanderActive = string.Empty });
```
QilongTang, line 647: Maybe let's add `General` here so the index stable for a while :)

QilongTang (approved):
LGTM with comments

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
QilongTang: Thank you @RobertGlobant20 @Astul-Betizagasti I am fine with the new approach as it reduces some complexity

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
RobertGlobant20: @QilongTang I changed the data structure to use a Dictionary, so in the XAML we can use something more describing like "PreferencesTabs[Features]" also I added the General tab entry as an item in the Dictionary. Thanks
see commit: https://github.com/DynamoDS/Dynamo/pull/11689/commits/4540d66f8358210f8468a3f2ce76482368424622

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
RobertGlobant20: @QilongTang fixed, see commit: https://github.com/DynamoDS/Dynamo/pull/11689/commits/4540d66f8358210f8468a3f2ce76482368424622

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
QilongTang: @RobertGlobant20 Thanks, this looks way cleaner

PR #11698, opened 2021-05-20
add package states under dbg mode

pinzart90:
### Purpose
This PR adds package states to the Installed Packages UI.
The UI changes are not final, they UX is still under review.
I also changed around some of the wording used.

UI changes preview
image

### Declarations

### Reviewers

### FYIs


src/DynamoPackages/Package.cs
```diff
@@ -113,6 +114,60 @@ public bool MarkedForUninstall
             internal set { markedForUninstall = value; RaisePropertyChanged("MarkedForUninstall"); }
         }
 
+        public bool EnableOldMarkedForUnistallState
```
pinzart90, line 117: seems that wpf can only bind to public properties...:(
If we want to have it under a debug mode first....we could just tag these temporary public methods...that they should not be used ..?

mjkkirschner: yeah, one thing we've done in the past is to mark with obsolete immediately - it's confusing, but it let's people know to be wary.

mjkkirschner: I don't see any harm to removing an obsolete method that we've explicitly called out as DO NOT USE.

QilongTang: > I don't see any harm to removing an obsolete method that we've explicitly called out as DO NOT USE.

Yes, that one I agree. I am curious about the case that we obsoleted something but decided that we want to keep it in the end, do we simply remove the obsolete tag?

mjkkirschner: yep :) - Again, what cause concerns you? if no one is using the member for fear it might be removed, then what break could keeping it cause?

pinzart90:
Not sure about the HIG standards (yet) but this is how I chose to handle long texts:
The package state tag/label will have a max width of 150 pixels
The text will be wrapped after max width is exceeded.
The tooltips have more information : The full text form the label + details.

image
image

src/DynamoCore/Properties/Resources.en-US.resx
```diff
@@ -747,4 +747,32 @@ Parameter name: {0}</value>
...
+This package has not been installed due to an unexpected error.</value>
+  </data>
+    <data name="PackageStateLoadedTooltip" xml:space="preserve">
+    <value>Loaded.
+This package is loaded and ready to be used.</value>
+  </data>
+    <data name="PackageStateUnloadedTooltip" xml:space="preserve">
+    <value>Skipped.
+This package has not been loaded because another conflicting package was loaded before it.</value>
+  </data>
```
pinzart90, line 777: @mjkkirschner @Amoursol @Jingyi-Wen
It has been brought to my attention that the terms used are not really consistent (like loaded vs installed)
What about using the following terms:
1. **Installed** - package is loaded into Dynamo and ready to be used.
2. **Scheduled for uninstall** - package will be uninstalled after next Dynamo restart.
3. **Skipped** - package is available on disk but was not loaded into Dynamo due to conflict with another similar package.
4. **Error** - package was not loaded into Dynamo due to an unexpected error

Also what do you think about scheduled vs marked ....for uninstall ?

If you have any alternatives....please tag them with the numbers above so that we know which one is which :)

src/DynamoCore/Properties/Resources.Designer.cs
```diff
@@ -1163,6 +1163,82 @@ public class Resources {
...
+        /// </summary>
+        public static string PackageStateError {
+            get {
+                return ResourceManager.GetString("PackageStateError", resourceCulture);
+            }
+        }
+        
+        /// <summary>
+        ///   Looks up a localized string similar to Error.
+        ///This package has not been installed due to an unexpected error..
```
aparajit-pratap, line 1177: Does _installed_ mean downloaded and loaded here?

src/DynamoCore/Properties/Resources.Designer.cs
```diff
@@ -1163,6 +1163,82 @@ public class Resources {
...
+        ///This package is loaded and ready to be used..
+        /// </summary>
+        public static string PackageStateLoadedTooltip {
+            get {
+                return ResourceManager.GetString("PackageStateLoadedTooltip", resourceCulture);
+            }
+        }
+        
+        /// <summary>
+        ///   Looks up a localized string similar to Scheduled for uninstall.
```
aparajit-pratap, line 1205: Does _Uninstall_ mean completely deleted here or just unloaded?

src/DynamoCore/Properties/Resources.en-US.resx
aparajit-pratap: 1. I would use **Loaded** vs. Installed since it implies that the package needs to be downloaded as well as loaded in one word.
2. I wish to clarify what it means to say **Scheduled for uninstall** - does it mean scheduled for unload or scheduled to be nuked? If not clear, we should probably change it to say **Scheduled for deletion** or **Scheduled for unloading**
3. Instead of skipped why don't we just use the term **Unloaded**?

pinzart90: The main concern, I guess, is about using **Loaded**/**Unloaded** vs Scheduled for **uninstall**.
What @aparajit-pratap is suggesting is consistent and I would say they seem fine to me

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
```diff
@@ -777,6 +778,10 @@ internal void SetPackageState(PackageDownloadHandle packageDownloadHandle, strin
             Package dynPkg;
             if (packageDownloadHandle.Extract(DynamoViewModel.Model, downloadPath, out dynPkg))
             {
+                if (DebugModes.IsEnabled("DynamoPackageStates"))
+                {
+                    dynPkg.PackageState = Package.PackageStates.Loaded;
```
aparajit-pratap, line 783: Shouldn't we set this only after the call to `LoadPackages`?

src/DynamoCoreWpf/Views/PackageManager/InstalledPackagesView.xaml
```diff
@@ -89,12 +89,30 @@
                                                    FontSize="11"
                                                    Foreground="#CCC"></TextBlock>
                                         <TextBlock Text="{x:Static p:Resources.InstalledPackageViewPendingInstallButton}"
-                                                   Visibility="{Binding Path=Model.MarkedForUninstall, Converter={StaticResource BooleanToVisibilityCollapsedConverter}}"
+                                                   Visibility="{Binding Path=Model.EnableOldMarkedForUnistallState, Converter={StaticResource BooleanToVisibilityCollapsedConverter}}"
```
aparajit-pratap, line 92: Little confused here, why couldn't you simply change the implementation of the `MarkedForUninstall` property instead of introducing a new one?

src/DynamoCore/Properties/Resources.en-US.resx
Jingyi-Wen: 1. I agree with @aparajit-pratap on using **Loaded**, since we call the window "Installed packages". And generally I think **install and uninstall** sounds more like user behaviors while **load and unload** sounds more like Dynamo behaviors
2.  Agree on **Scheduled for unloaded** or **deletion**. Same reason above
3.  **Not loaded** or **Unloaded** makes sense to me

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
pinzart90: hmm...I did not attempt to properly set all the states. 
I only focused on the UI/UX side of it.

It may make sense to actually set the states in the `PackageManagerExtension.PackageLoader` ...since we may have more visibility into error states there. Or maybe some other notification system...Not sure yet

I can add a comment that this specific code is temporary...and should be moved somewhere else...And I can move it below as well

src/DynamoCoreWpf/Views/PackageManager/InstalledPackagesView.xaml
pinzart90: I introduced another property because `MarkedForUninstall` is used in other parts of Dynamo, not only used in the UI...and I did not want to break that existing functionality by introducing the DebugMode flag into the equation.
Basically, if I had done : `bool MarkedForUninstall { get {return DebugModes.IsActive() && isMarkedForUninstall; } }` then, while having the DebugMode active, the other parts of the Dynamo that use `MarkedForUninstall` would not work anymore.

pinzart90: MarkedForUninstall...should probably be replaced by the PackageStates ...when the implementation is ready

src/DynamoCore/Properties/Resources.en-US.resx
```diff
@@ -747,4 +747,32 @@ Parameter name: {0}</value>
...
+  <data name="PackageStateUnloaded" xml:space="preserve">
+    <value>Unloaded</value>
+  </data>
+  <data name="PackageStatePendingUnloadTooltip" xml:space="preserve">
+    <value>Scheduled to be unloaded.
+This package will be unloaded after the next Dynamo restart.</value>
+  </data>
+    <data name="PackageStateErrorTooltip" xml:space="preserve">
+    <value>Unloaded.
+This package has not been loaded due to an unexpected error.</value>
```
pinzart90, line 768: Maybe we can be more specific about the errors...once we know where/how to set the package states in the Dynamo code

src/DynamoCore/Properties/Resources.en-US.resx
```diff
@@ -747,4 +747,32 @@ Parameter name: {0}</value>
...
+  </data>
+    <data name="PackageStateLoadedTooltip" xml:space="preserve">
+    <value>Loaded.
+This package is loaded and ready to be used.</value>
+  </data>
+    <data name="PackageStateUnloadedTooltip" xml:space="preserve">
+    <value>Unloaded.
+This package has not been loaded because another conflicting package was loaded before it.</value>
```
pinzart90, line 776: Same as above, we may be able to specify which package caused the current one to get into an Unloaded state

pinzart90:
Latest screenshot
image

src/DynamoPackages/Package.cs
sm6srw: 👍

sm6srw (approved):
Nice, LGTM!

src/DynamoCore/Properties/Resources.en-US.resx
Amoursol: 1. I'm also in favour of **Loaded** as it uses stronger language (Installed could still theoretically fail to load)
2. _Scheduled for Uninstall/Unload/Deletion_ is still the one that feels most ambiguous, but I don't have a better suggestion. In lieu of that I would lean towards **Scheduled for Deletion/Unload** as above.
3. **Unloaded** makes most sense to me here to match the inverse of the _Loaded_ state.
4. **Error** is perfect here - can the user see the error message and do something about it in this state @pinzart?

pinzart90: @Amoursol Regarding the error state. Hopefully we can display something meaningful to the user....but that work will probably be a part of this jira task https://jira.autodesk.com/browse/DYN-3670

PR #11701, opened 2021-05-21
Users should be able to type into the Auto completion search window.

reddyashish:
### Purpose

Currently we have a regression on the master branch. The issue is that users are not able to type text into the Auto completion search window. After the new UI changes to the Autocompletion window, the window will not be closed/hidden when the user has clicked anywhere on the workspace as we now have a close button to close the window. We won't be needing the flag to monitor the behavior of the window.

NodeAutocompletion search box

### Declarations

@QilongTang @zeusongit  @mmisol

### DIFF
--- a/src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml
+++ b/src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml
@@ -81,8 +81,8 @@
                             </Image.Style>
                         </Image>
                         </Button>
-                        <Button Name="CloseButton" Margin="0,7,5,7" BorderThickness="0"  Background="{StaticResource autocompletionWindow}">
-                            <Image Width="14" Height="14" HorizontalAlignment="Right" VerticalAlignment="Center">
+                        <Button Name="CloseButton" Margin="0,7,5,7" Height="14" Width="14" BorderThickness="0" Click="CloseAutocompletionWindow" Background="{StaticResource autocompletionWindow}">
+                                <Image Width="14" Height="14" HorizontalAlignment="Right" VerticalAlignment="Center">
                                 <Image.Style>
                                     <Style TargetType="{x:Type Image}">
                                         <Style.Triggers>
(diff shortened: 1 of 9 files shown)


src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml
```diff
@@ -81,8 +81,8 @@
                             </Image.Style>
                         </Image>
                         </Button>
-                        <Button Name="CloseButton" Margin="0,7,5,7" BorderThickness="0"  Background="{StaticResource autocompletionWindow}">
-                            <Image Width="14" Height="14" HorizontalAlignment="Right" VerticalAlignment="Center">
+                        <Button Name="CloseButton" Margin="0,7,5,7" Height="14" Width="14" BorderThickness="0" Click="CloseAutocompletionWindow" Background="{StaticResource autocompletionWindow}">
```
mmisol, line 84: It's surprising that this button already existed :)

mmisol (commented):
Hi @reddyashish . I think this is a good approach, however you might have to make some changes to the tests. If I remember correctly, in my original PR I had to change a test to make it close the window twice.

Also, does it work so that if you select a node, the dialog still closes automatically. That would be important for usability, not sure if you could add a test for it.

src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml
reddyashish: This button was added later.

QilongTang (approved):

reddyashish:
A couple of issues around the node autocompletion search are also fixed in this PR
1) When the user enters some text in the search bar and clears it, the window should show the initial list of all the autocompletion elements for that port
2) The tag dictionary that is being used in the Search API contains a list of nodes that share a specific tag. So when the SubsetDictionary was being built, if any of the nodes from dictionary match the subset list, the whole list of nodes that share a specific tag was returned. This would generate extra results(which were not a match) when some particular text was entered in the search bar. To fix this, we match the subset list with the returned search results in the end. I will add a gif to show the difference in the results.

reddyashish:
> Hi @reddyashish . I think this is a good approach, however you might have to make some changes to the tests. If I remember correctly, in my original PR I had to change a test to make it close the window twice.
> 
> Also, does it work so that if you select a node, the dialog still closes automatically. That would be important for usability, not sure if you could add a test for it.

Yes changed the test accordingly. Regarding the second question, when you select the node, the node closes automatically. I think we will need a UI automation test for that.

zeusongit (approved):

reddyashish:
Merging this.

PR #11713, opened 2021-05-27
Linting Manager bug fixes (DYN-3732, DYN-3733)

saintentropy:
### Purpose

This PR address two issues with the LintingManager API.  Specifically the following:

* Dynamo Graph type should be reset to "None" in the DropDown choice when switching to a graph that has no type set.  This can occur when a user 1) opens a new Graph by clicking "New", 2) opens a Dynamo file that does not have a Linter type set, or 3) opens a Dynamo file with a linter set that is not installed in the current Dynamo session.  This UX is accomplished by creating a `DefaultLinter` within the `LinterManager` and reseting the `ActiveLinter`to the default in the scenarios described.

* The linter Manager utilizes ObservableCollections in for the Results, `NodeIssues`, `GraphIssues` and `AffectedNodes`. All these collections are typically modified via the UI thread but not always.  For code patches that can be updated by threads other than the UI thread, the modifications to the collection must be Dispatched to the UI thread. 

### Declarations

### Reviewers

@BogdanZavu @SHKnudsen 

### FYIs

@mjkkirschner @nate-peters


src/DynamoCore/Linting/LinterManager.cs
```diff
@@ -20,8 +20,10 @@ namespace Dynamo.Linting
     public class LinterManager : NotificationObject, IDisposable
     {
         #region Private fields
+        private const string NONE_DESCRIPTOR_GUID = "7b75fb44-43fd-4631-a878-29f4d5d8399a";
         private readonly IExtensionManager extensionManager;
         private LinterExtensionDescriptor activeLinter;
+        internal LinterExtensionDescriptor DefaultDescriptor => new LinterExtensionDescriptor(NONE_DESCRIPTOR_GUID, Properties.Resources.NoneLinterDescriptorName);
```
BogdanZavu, line 26: Maybe definition of default descriptor should be in LinterExtensionDescriptor class as a static ?

src/DynamoCore/Linting/LinterManager.cs
```diff
@@ -52,12 +54,14 @@ public LinterExtensionDescriptor ActiveLinter
...
 
+                activeLinter = value;
+
                 if (TryGetLinterExtension(value, out linterExtension))
                 {
                     linterExtension.Activate();
                 }
 
-                activeLinter = value;
+                RaisePropertyChanged("ActiveLinter");
```
BogdanZavu, line 64: I think we can do the same thing for "AvailableLinters" - for the scenario when a new linter extension is part of a package install.

src/LintingViewExtension/LinterViewModel.cs
```diff
@@ -18,14 +19,12 @@ namespace Dynamo.LintingViewExtension
 {
     public class LinterViewModel : NotificationObject
     {
-        private const string NONE_DESCRIPTOR_GUID = "7b75fb44-43fd-4631-a878-29f4d5d8399a";
-
         #region Private Fields
         private static Uri linterViewHelpLink = new Uri("LintingViewExtension;LinterViewHelpDoc.html", UriKind.Relative);
         private LinterExtensionDescriptor activeLinter;
```
BogdanZavu, line 24: I think we can remove this member and rely solely on LinterManager for the ActiveLinter property

src/LintingViewExtension/LinterViewModel.cs
```diff
@@ -144,7 +134,7 @@ private void AddNewNodeIssue(string issueNodeId, string ruleId)
             var newIssue = new NodeRuleIssue(ruleId, GetLinterRule(ruleId) as NodeLinterRule);
             newIssue.AddAffectedNodes(new List<NodeModel> { issueNode });
 
-            NodeIssues.Add(newIssue);
+            this.dispatcher.Invoke(() => { NodeIssues.Add(newIssue); });
```
BogdanZavu, line 137: Should this.GraphIssues.Clear() & this.NodeIssues.Clear() be on dispatcher as well ?

src/LintingViewExtension/LintingViewExtension.cs
```diff
@@ -39,6 +39,8 @@ public override void Loaded(ViewLoadedParams viewLoadedParams)
             this.linterViewModel = new LinterViewModel(linterManager, viewLoadedParamsReference);
             this.linterView = new LinterView() { DataContext = linterViewModel };
 
+            linterViewModel.ActiveLinter = linterManager.ActiveLinter;
```
BogdanZavu, line 42: we can remove this if we rely on LinterManager direct access for ActiveLinter property

src/DynamoCore/Linting/LinterExtensionDescriptor.cs
```diff
@@ -20,5 +20,39 @@ public LinterExtensionDescriptor(string id, string name)
...
+        /// </summary>
+        /// <param name="obj"></param>
+        /// <returns></returns>
+        public override bool Equals(object obj)
+        {
+            if (obj == null)
+            {
+                return false;
+            }
+            if (!(obj is LinterExtensionDescriptor))
```
mjkkirschner, line 36: I think you could simplify this with `(if obj is LinterExtensionDescriptor other)`
https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/pattern-matching#null-checks

src/DynamoCore/Linting/LinterManager.cs
mjkkirschner: nameof?

src/DynamoCore/Linting/LinterManager.cs
```diff
@@ -122,14 +134,18 @@ private void OnRuleEvaluated(IRuleEvaluationResult result)
                         if (storingResult.NodeIds != gRuleResult.NodeIds) 
                         {
                             // remove original result and replace with new one
-                            RuleEvaluationResults.Remove(storingResult);
-                            RuleEvaluationResults.Add(result);
+                            DynamoModel.OnRequestDispatcherInvoke(() =>
```
mjkkirschner, line 137: what would ever call this not from the UI thread? And if it's not the UI thread, could it be multiple other threads -  do you need to add concurrency controls around your collection changes?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -891,6 +891,11 @@ protected DynamoModel(IStartConfiguration config)
                             {
                                 if (loadedExtension is IExtension)
                                 {
+                                    if (loadedExtension is LinterExtensionBase loadedLinter)
+                                    {
+                                        loadedLinter.InitializeBase(this.LinterManager);
```
mjkkirschner, line 896: what is this?

src/LintingViewExtension/LinterViewModel.cs
```diff
@@ -74,30 +73,21 @@ public List<LinterExtensionDescriptor> AvailableLinters
...
-        /// </summary>
-        public IList<LinterExtensionDescriptor> DefaultDescriptor { get => new List<LinterExtensionDescriptor> { defaultDescriptor }; }
-
         #endregion
 
         public LinterViewModel(LinterManager linterManager, ViewLoadedParams viewLoadedParams)
         {
             this.linterManager = linterManager ?? throw new ArgumentNullException(nameof(linterManager));
             this.viewLoadedParams = viewLoadedParams;
+            this.dispatcher = viewLoadedParams.DynamoWindow.Dispatcher;
```
mjkkirschner, line 82: this might be null... I think you should add `?` to all these calls or wrap them in a helper.

src/LintingViewExtension/LinterViewModel.cs
```diff
@@ -257,6 +247,12 @@ private void RuleEvaluationResultsCollectionChanged(object sender, NotifyCollect
             }
         }
 
+        private void OnLinterManagerPropertyChange(object sender, System.ComponentModel.PropertyChangedEventArgs e )
+        {
+            if (e.PropertyName == "ActiveLinter")
```
mjkkirschner, line 252: nameof?

mjkkirschner: and {} please

src/LintingViewExtension/LinterViewModel.cs
```diff
@@ -272,9 +268,10 @@ private NodeModel NodeFromId(string nodeId)
             return node;
         }
 
-        private void CreateDefaultDummyLinterDescriptor()
+        public void Dispose()
```
mjkkirschner, line 271: IDisposable maybe?

src/DynamoCore/Linting/LinterManager.cs
saintentropy: The scenario we ran into so far was a rule that was watching `CachedValue` property change on a node.  This could be triggered from the SchedulerThread if a user had the graph in manual and pressed Run.

saintentropy: At this time it does not seem like the concurrency guards are needed but @BogdanZavu and I still discussing that scenario

src/DynamoCore/Models/DynamoModel.cs
saintentropy: This was added to mirror the init function for the LinterExtensionBase that is a few lines above.  The above lines handle internal extensions are `LinterExtensionBase`.  This init is for those that are loaded via the package manager

src/LintingViewExtension/LinterViewModel.cs
saintentropy: Yes will do

src/DynamoCore/Linting/LinterManager.cs
BogdanZavu: Yes, at this point we only have main and scheduler thread involved and it's wither one or the other.
There is a chance in the future we might allow an api developer to have more control over events that trigger linter evaluation and at that point we will have to deal with concurrency.

src/DynamoCore/Extensions/LinterExtensionBase.cs
```diff
@@ -245,7 +246,7 @@ private void UnlinkFromCurrentWorkspace()
             if (currentWorkspace != null)
             {
                 UnsubscribeGraphEvents(currentWorkspace);
-                linterManager.RuleEvaluationResults.Clear();
+                DynamoModel.OnRequestDispatcherInvoke(() => { this.linterManager.RuleEvaluationResults.Clear(); });
```
BogdanZavu, line 249: I believe these are not needed and we only need the dispatcherInvoke for the changes in the binded collection in the view model.

BogdanZavu (approved):

PR #11714, opened 2021-05-27
add mechanism to reload Cpython modules and mark graph dirty

mjkkirschner:
### Purpose
https://jira.autodesk.com/browse/DYN-3514
https://jira.autodesk.com/browse/DYN-2941

### Update:

To help dynamo users who write their own python modules in cpython deal with the difference in module import/reload between cpython and ironPython- this PR adds a button to Reset CPython. What it does is to reload python modules that are backed by a .py file and are not named `__main__` and a few other checks inspired by `autoreload`
https://ipython.org/ipython-doc/3/config/extensions/autoreload.html

The implementation is a subset of what autoreload does:

- modules that have .py files are reloaded - modules that are reloaded, are logged to the console:
ie:
```
2021-05-28 22:18:42Z : Python Script: considering reload_test3
2021-05-28 22:18:42Z : Python Script: reloading reload_test3
2021-05-28 22:18:42Z : Python Script: failed to reload reload_test3 spec not found for the module 'reload_test3'
2021-05-28 22:18:42Z : Python Script: considering importlib._bootstrap
2021-05-28 22:18:42Z : Python Script: considering importlib._bootstrap_external
2021-05-28 22:18:42Z : Python Script: considering importlib
2021-05-28 22:18:42Z : Python Script: considering importlib.machinery
2021-05-28 22:18:42Z : Python Script: considering importlib.abc
2021-05-28 22:18:42Z : Python Script: considering contextlib
2021-05-28 22:18:42Z : Python Script: considering importlib.util
2021-05-28 22:18:42Z : Python Script: considering pprint
```

Currently, after the modules are reloaded, the Dynamo engine is reset and all nodes are marked dirty - this is very likely overkill. I think we would be safe simply marking the nodes dirty in topological order or disabling execution, marking all nodes dirty, and then re-enabling execution. 
The goal is to have the DS GC collect the old objects created with out dated modules, the easiest way to do this is to rerun the graph.
 
also:

* I had to change the use of python thread macros and lock code we started using in this PR: https://github.com/DynamoDS/Dynamo/pull/10631/ - I think it was a bit confused (passing a pointer from the GIL mutex to a macro that expected a pointer to a different mutex). 
    * I more closely followed the examples in this doc:
https://github.com/pythonnet/pythonnet/wiki/Threading
    * I got rid of the `acquire` and `release` lock calls as I believe using the `GIL()` call is analogous to that.

* I ended up using static internal events to avoid having any references on DSCPython or PythonNet in DynamoCore/DynamoCoreWPF - I guess I could also use interfaces and a singleton or something. Feedback welcome there.
* Modified the style for `flatButton` so that its center aligned @Jingyi-Wen  fyi. (that affects the group add stye button)

Screen Shot 2021-06-01 at 4 30 19 PM
Screen Shot 2021-06-01 at 3 53 08 PM

Screen Shot 2021-06-02 at 1 29 02 PM

TODO:
- [x] address todos
- [x] more tests
- [x] tooltip/docs for when to use this.
- [x] outside scope of this PR - but will create docs for addressing this problem in various ways (import.reload, reloadr, autoreload, this button) - stubbed out here: https://github.com/DynamoDS/Dynamo/wiki/Work-in-progress-to-improve-Python-3-support#import-module-works-differently-in-ironpython-and-cpython-wip
- [x] update license and rtf files to note reference of autoreloader from Ipython project.

### Declarations


Jingyi-Wen:
@mjkkirschner The location of the restart button makes sense! For the button style, could we not have the outline? 
In HIG, the default state of the flat button is without outline (so should look the same as the "Add style" button).

mjkkirschner:
> @mjkkirschner The location of the restart button makes sense! For the button style, could we not have the outline?
> In HIG, the default state of the flat button is without outline (so should look the same as the "Add style" button).

Hey @Jingyi-Wen yep it looks just like the `Add Style` button - I used the same style, I was just hovering over it when I took the screenshot.

Jingyi-Wen:
> > @mjkkirschner The location of the restart button makes sense! For the button style, could we not have the outline?
> > In HIG, the default state of the flat button is without outline (so should look the same as the "Add style" button).
> 
> Hey @Jingyi-Wen yep it looks just like the `Add Style` button - I used the same style, I was just hovering over it when I took the screenshot.

Got it. Looks good then!

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -119,125 +119,6 @@ public partial class DynamoModel : IDynamoModel, IDisposable, IEngineControllerM
 
         #endregion
 
-        #region events
```
mjkkirschner, line 122: moved all DynamoModel events to this events .cs file.

src/DynamoCoreWpf/Properties/Resources.en-US.resx
```diff
@@ -2390,4 +2390,13 @@ Uninstall the following packages: {0}?</value>
   <data name="PreferencesViewEnableNodeAutoCompleteTooltipText" xml:space="preserve">
     <value>Learn more about Node Autocomplete feature.</value>
   </data>
+  <data name="AddStyleButton" xml:space="preserve">
+    <value>Add Style</value>
```
mjkkirschner, line 2394: this was missing a resx property.

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -151,10 +151,74 @@ public static class CPythonEvaluator
         private const string NodeName = "__dynamonodename__";
         static PyScope globalScope;
         internal static readonly string globalScopeName = "global";
+        private static DynamoLogger dynamoLogger;
+        private static IntPtr ts = new IntPtr();
```
mjkkirschner, line 155: because we're no longer calling endAllowThreads should be able to make this local variable again.

src/DynamoCore/Models/DynamoModelEvents.cs
```diff
@@ -357,6 +358,137 @@ internal virtual void OnComputeModelDeserialized()
...
+        /// </summary>
+        internal static event Action<string> RequestPythonReset;
+        internal void OnRequestPythonReset(string pythonEngine)
+        {
+            //only reset if current workspace is a homeworkspace
+            //can't guarantee which workspace to mark dirty otherwise
+            if (CurrentWorkspace is HomeWorkspaceModel hmwsm)
+            {
+                RequestPythonReset?.Invoke(pythonEngine);
+                ResetEngine(true);
```
mjkkirschner, line 487: I think this is probably overkill, but this function is only triggered manually for now.

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -178,16 +242,13 @@ static CPythonEvaluator()
             }
 
             InstallPython();
-
             if (!PythonEngine.IsInitialized)
             {
+                InitializeEncoders();
```
mjkkirschner, line 247: this could safely be moved back to the constructor now.

src/DynamoCore/Models/DynamoModelEvents.cs
pinzart90: by manually you mean via new button in preferences ?
When/if we move to something more reactive/automated....we will probably want to reset only the nodes that are affected by the .py reloads right ?

src/DynamoCoreWpf/Properties/Resources.resx
```diff
@@ -2390,4 +2390,13 @@ Uninstall the following packages: {0}?</value>
...
     <value>Learn more about Node Autocomplete feature.</value>
   </data>
+  <data name="AddStyleButton" xml:space="preserve">
+    <value>Add Style</value>
+  </data>
+  <data name="ResetCPythonButtonText" xml:space="preserve">
+    <value>Reset CPython</value>
+  </data>
+  <data name="ResetCPythonButtonToolTip" xml:space="preserve">
+    <value>Resets CPython environment by reloading modules.</value>
```
pinzart90, line 2400: does this need to be more descriptive ?
all modules ? or some ? do we need to be specific I wonder ...

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -178,16 +241,12 @@ static CPythonEvaluator()
...
             if (!PythonEngine.IsInitialized)
             {
                 PythonEngine.Initialize();
                 PythonEngine.BeginAllowThreads();
+                
             }
-
-            IntPtr gs = PythonEngine.AcquireLock();
-            try
-            {
```
pinzart90, line 190: this lock stuff is not needed anymore ?

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -151,10 +150,74 @@ public static class CPythonEvaluator
...
+                //the following is inspired by:
+                //TODO add attribution to aboutbox/license. clause3 BSD3 (preapproved)
+                //https://github.com/ipython/ipython/blob/master/IPython/extensions/autoreload.py
+                    var global = PyScopeManager.Global.Get(CPythonEvaluator.globalScopeName);
+                    global?.Exec(@"import sys
+import importlib
+import importlib.util
+import os
+def getInfoFile(module):
+    if not hasattr(module, '__file__') or module.__file__ is None:
```
pinzart90, line 183: maybe add comments to all these filters/conditions so that we know what the intention was (in case something goes wrong)

src/DynamoCoreWpf/Properties/Resources.resx
mjkkirschner: thanks @pinzart90 - I was thinking to leave it a bit intentionally vague at this point so we can adjust it easily, but we could specify modules that have file attributes?

src/Libraries/DSCPython/CPythonEvaluator.cs
mjkkirschner: nope, I believe this is the same as using Py.GIL() - see the description and threading wiki that says as much.

src/Libraries/DSCPython/CPythonEvaluator.cs
mjkkirschner: good idea, will add some comments

pinzart90: where did you find documentation for these module attributes ?

src/DynamoCore/Models/DynamoModelEvents.cs
mjkkirschner: Yes, that would be the ideal - BUT identifying those nodes without creating a circular reference is kind of a pain - we can use reflection or string matching if necessary.

src/Libraries/DSCPython/CPythonEvaluator.cs
pinzart90: Maybe we should add a comment (like in the importlib.reload documentation)
```Reloading sys, __main__, builtins and other key modules is not recommended```
https://docs.python.org/3.4/library/importlib.html#importlib.reload

File: src/DynamoCoreWpf/Properties/Resources.resx (hunk 6 above)
Comment (pinzart90) [reply src/DynamoCoreWpf/Properties/Resources.resx:2400]: not sure....maybe vague here and more specific in some dynamoDS wiki ?

File: src/Libraries/DSCPython/CPythonEvaluator.cs [hunk 9]
```diff
@@ -151,10 +149,78 @@ public static class CPythonEvaluator
...
+        py_filename = filename
+    else:
+        try:
+            py_filename = importlib.util.source_from_cache(filename)
+        except ValueError:
+            return None
+    return py_filename
+
+
+for modname,mod in sys.modules.copy().items():
```
pinzart90, line 206: any idea why we need to make the copy() call ?

src/DynamoCore/Models/DynamoModelEvents.cs
pinzart90: so by re executing all the nodes we are also taking care (refreshing) of potentially dangling objects that might still reference the old python files right ?

src/DynamoCoreWpf/Properties/Resources.resx
mjkkirschner: I've added https://github.com/DynamoDS/Dynamo/wiki/Work-in-progress-to-improve-Python-3-support for docs, we can expand on specifics there and easily update it.

src/Libraries/DSCPython/CPythonEvaluator.cs
mjkkirschner: will do - these attributes (and this function) were modified from the autoreload.py script which is why I added attribution to this library in the license and about boxes.

src/Libraries/DSCPython/CPythonEvaluator.cs
mjkkirschner: yes because iterating over sys.modules and reloading modules directly modifies the keys in this dictionary during iteration.

Making a copy avoids modifying the dictionary we're iterating over.

mjkkirschner:
@pinzart90 PTAL - I've added comments, please check if they are adequate. When this is merged, I will update the wiki with specifics and note that the change has been merged.

pinzart90 (approved):

mjkkirschner:
@pinzart90 this passed on the parallel job here:
https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/2562/
and serial one here:
https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/797/

though there were some flaky test failures we've seen before that occurred on the previous parallel run. I'm going to merge this, but if we see them get more frequent we can revert.
FYI @QilongTang @jasonstratton

PR #11724, opened 2021-06-01
Align the control titles in Preferences panel.

reddyashish:
### Purpose

Align the titles in Preferences panel. 
Place **Default Run Settings** and **Show Run Preview** near each other. 

Screen Shot 2021-06-01 at 4 51 47 PM
Screen Shot 2021-06-01 at 4 51 55 PM

### Declarations

### Reviewers

@QilongTang @Amoursol

### DIFF
--- a/src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
+++ b/src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
@@ -177,7 +177,7 @@
 
                                 <!--The first Grid Row will contain the Language label and combobox added in a StackPanel vertical-->
                                 <Grid Grid.Row="0" 
-                                  Margin="0,0,0,20">
+                                  Margin="4,0,0,20">
                                     <StackPanel Orientation="Vertical">
                                         <Label  Content="{x:Static p:Resources.PreferencesViewLanguageLabel}" 
                                             Margin="-3,0,10,0"
@@ -197,7 +197,7 @@
 
                                 <!--This Grid contains controls for selecting Run Settings options-->
                                 <Grid Grid.Row="1"
-                                  Margin="0,0,0,20">
+                                  Margin="4,0,0,20">
                                     <Grid.RowDefinitions>
                                         <RowDefinition Height="Auto" />
                                         <RowDefinition Height="*" />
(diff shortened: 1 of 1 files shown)


reddyashish:
@Jingyi-Wen Let me know if you think of any more changes to the Preferences panel.

Amoursol:
@reddyashish can we have the same amount of white-space (Grey area) at the front of the Geometry Scaling background bar that we have at the end? 

image

QilongTang:
> @reddyashish can we have the same amount of white-space (Grey area) at the front of the Geometry Scaling background bar that we have at the end? The back looks better space wise IMHO 😊
> 
> image

This may be a hard one, the space you were mentioning were due to the scroll bar. Without the scroll bar, the expanders leave no space on the edges. Do you mean adjusting spacing for all the expanders with or without scroll bar on the right?

Jingyi-Wen:
Looks better! Just a minor thing - can we make description tooltip box for geometry scaling wider and increase the spacing between the box and the button slightly? It seems a bit packed now comparing to the original design.  
Screen Shot 2021-06-02 at 9 24 33 AM

QilongTang:
@reddyashish Would you update the screenshots for another review?

reddyashish:
@Amoursol Like Aaron said that space is created due to the scroll bar. The constraint with current design is that the "Geometry Scaling" text has a fixed left-margin inside the header bar. So if we move the header bar little to the right, then the header text will move to the right and other titles should be moved to the right accordingly. What do you think?

@Jingyi-Wen @QilongTang here is the updated shot:
Screen Shot 2021-06-02 at 10 29 19 AM

Screen Shot 2021-06-02 at 10 29 13 AM

Jingyi-Wen:
I'm fine with moving the expander & other titles to the right to give it more space (will have to do the same for other tabs, might be out of the scope of this task?) 
Other than that, looking really good!

Amoursol:
> @Amoursol Like Aaron said that space is created due to the scroll bar. The constraint with current design is that the "Geometry Scaling" text has a fixed left-margin inside the header bar. So if we move the header bar little to the right, then the header text will move to the right and other titles should be moved to the right accordingly. What do you think?
> 
> @Jingyi-Wen @QilongTang here is the updated shot:
> Screen Shot 2021-06-02 at 10 29 19 AM
> 
> Screen Shot 2021-06-02 at 10 29 13 AM

Understood @reddyashish - in that case looks good to me!

QilongTang (approved):

PR #11726, opened 2021-06-02
DYN-3738: enhance readyparams api

BogdanZavu:
### Purpose
Enhance existing ReadyParams API so that we can have better control over graph opening and closing operations as part of an extension. We will use this in LintingViewExtension where we want to do some cleanup when a graph is closed and allow an extension developer to perform cleanup operations on their listing rules.

### Declarations

### DIFF
--- a/src/DynamoCore/Extensions/ReadyParams.cs
+++ b/src/DynamoCore/Extensions/ReadyParams.cs
@@ -23,7 +23,10 @@ namespace Dynamo.Extensions
         {
             dynamoModel = dynamoM;
             dynamoModel.PropertyChanged += OnDynamoModelPropertyChanged;
+            dynamoModel.WorkspaceOpened += OnCurrentWorkspaceModelOpened;
+            dynamoModel.WorkspaceClearing += OnCurrentWorkspaceModelClearing;
             dynamoModel.WorkspaceCleared += OnCurrentWorkspaceModelCleared;
+            dynamoModel.WorkspaceRemoveStarted += OnCurrentWorkspaceRemoveStarted;
             dynamoM.Logger.NotificationLogged += OnNotificationRecieved;
             startupParams = new StartupParams(dynamoModel.AuthenticationManager.AuthProvider,
                 dynamoModel.PathManager, new ExtensionLibraryLoader(dynamoModel), dynamoModel.CustomNodeManager,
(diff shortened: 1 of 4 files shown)


src/DynamoCore/Models/DynamoModelEvents.cs
```diff
@@ -90,15 +90,16 @@ public virtual void OnRequestLayoutUpdate(object sender, EventArgs e)
...
         /// </summary>
-        public event Action WorkspaceClearing;
+        public event Action<WorkspaceModel> WorkspaceClearing;
 
         /// <summary>
         /// Triggers WorkspaceClearing event
         /// </summary>
-        public virtual void OnWorkspaceClearing()
+        /// <param name="workspace">Workspace about to be cleared</param>
+        public virtual void OnWorkspaceClearing(WorkspaceModel workspace)
```
BogdanZavu, line 99: This is an API change. I believe this is ok as long as we document it [here](https://github.com/DynamoDS/Dynamo/wiki/API-Changes-in-Dynamo-2.12) ?

mjkkirschner: That is incorrect. The public API should not be changed unless the major version number is incremented. Instead obsolete the old method and add a new one if possible.

mjkkirschner:
Hi @BogdanZavu, thank you for addressing the comments.
I wanted to clarify what I stated above a bit.

There are times we do break the API and document it like you originally had done. But this should only be done when theres no good alternative solution that does not break the API -  or there is good enough reason to intentionally break an API.

For example - we broke (and probably will again) the Helix 3d graphics APIs within a minor release:
* external types had leaked in the public Dynamo API.
* use was very low.
* most importantly, the experience for regular dynamo users was severely degraded in the older version of helix, and updating meant they no longer needed to install directX10.

In general I think principles like this should be used to judge API breaks.

* What API is being broken/what is the benefit?
* how many users will it effect negatively, how many positively?
* can we do it some other way and keep both apis working for a bit longer?
* Is the API popular? Is it an API we actively advertise (extensions) or just something that is public because of historical reasons.(helix)

src/DynamoCore/Extensions/ReadyParams.cs
```diff
@@ -104,6 +107,36 @@ private void OnCurrentWorkspaceModelCleared(IWorkspaceModel ws)
                 CurrentWorkspaceCleared(ws);
         }
 
+        /// <summary>
+        /// Occurs when current workspace is clearing
+        /// </summary>
+        public event Action<IWorkspaceModel> CurrentWorkspaceClearingStarted;
+        private void OnCurrentWorkspaceModelClearingStarted(IWorkspaceModel ws)
+        {
+            if (CurrentWorkspaceClearingStarted != null)
```
mjkkirschner, line 116: in newer c# versions we can now do this:
`CurrentWorkspaceClearingStarted?.Invoke(ws);`

mjkkirschner (approved):

BogdanZavu:
I was thinking of this more as a fix rather than an API change. Whoever subscribes to the clearing event most likely will get the workspace from somewhere else.

PR #11721, opened 2021-05-28
Add tests for the new IF node.

reddyashish:
### Purpose

This PR is to add tests around Lacing, ListAtLevel and Partial function features for the new IF node

There are some tests in "IfTest" fixture that are testing the old "IF" node. Do we want to keep them around to support the old node or do we plan to replace the tests with the new node?

### Declarations

Check these if you believe they are true

- [ ] The codebase is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.
- [ ] This PR modifies some build requirements and the readme is updated

### Reviewers

@aparajit-pratap @QilongTang

PR #11728, opened 2021-06-02
Merge if node changes

reddyashish:
### Purpose

This PR is to merge the "IF" node implementation changes from feature branch to master. 

https://github.com/DynamoDS/Dynamo/pull/11481
https://github.com/DynamoDS/Dynamo/pull/11598
https://github.com/DynamoDS/Dynamo/pull/11721

### Declarations

### Reviewers

@QilongTang @zeusongit @aparajit-pratap @mjkkirschner


src/Libraries/CoreNodeModels/Logic/If.cs
```diff
@@ -72,4 +78,78 @@ public override IEnumerable<AssociativeNode> BuildOutputAst(List<AssociativeNode
...
         }
     }
+
+    [NodeName("If")]
+    [NodeCategory(BuiltinNodeCategories.LOGIC)]
+    [NodeDescription("ScopeIfDescription", typeof(Resources))]
+    [OutPortTypes("Function")]
+    [IsDesignScriptCompatible]
+    [AlsoKnownAs("DSCoreNodesUI.Logic.If")]
+    public class RefactoredIf : NodeModel
```
reddyashish, line 88: Since this is going to be a public class, is "RefactoredIf" name fine with everyone? Are we going to obsolete the old node in 3.0 and then rename this class name to "IF"?

test/DynamoCoreTests/Nodes/IfTest.cs
```diff
@@ -76,5 +80,102 @@ public void TestScopeIfForPreview()
...
+            AssertPreviewValue("886a464b-9b2b-4e66-a033-c18d0753c2cf", true);
+            AssertPreviewValue("f945529e-62e3-4c7a-b07a-0b18d7449f9b", new object[] { "a1", "b1", "c1" });
+            // The output preview for partially connected "IF" node should return the function itself irrespective of test value.
+            AssertPreviewValue("e1630959-8338-4b1a-9832-248246aef8a0", "Function");
+
+            boolSelectorNode.Value = false;
+            AssertPreviewValue("886a464b-9b2b-4e66-a033-c18d0753c2cf", false);
+            AssertPreviewValue("f945529e-62e3-4c7a-b07a-0b18d7449f9b", new object[] { new object[] { "a2" }, "b2" });
+            // The output preview for partially connected "IF" node should return the function itself irrespective of test value.
+            AssertPreviewValue("e1630959-8338-4b1a-9832-248246aef8a0", "Function");
```
aparajit-pratap, line 102: This is fine but more importantly, we need a test where the `If` node is used as a partially applied function input to another node and ensure that the downstream node works as expected.

reddyashish: Modified the test accordingly.

src/Libraries/CoreNodeModels/Logic/If.cs
mjkkirschner: Maybe a more descriptive name and summary? Is there any reason not to obsolete the old node right now?

test/DynamoCoreTests/Nodes/IfTest.cs
aparajit-pratap: I'm looking for an example like this where the partially applied `If` node acts like a `function`, which when applied to a higher order node accepting a `function` returns a valid result. In this example, the `If` "function" is called twice, once for `true`, and secondly for `false` in the other list:
image

reddyashish: Got it. Updated the test with the above example.

src/Libraries/CoreNodeModels/Logic/If.cs
reddyashish: Sorry I meant, if we are going to remove this old node completely in 3.0. Currently the old one is obsoleted and it has a summary under BuildOutputAst(). Do you want me to add any other information?

aparajit-pratap (changes requested):
@reddyashish there is still work to be done to apply the existing If node icon to this new node. I tried your branch locally and the icon doesn't work for the new node.

reddyashish:
@aparajit-pratap looking into it.

aparajit-pratap:
@reddyashish can you also change the names of the small and large png files from the old class name to the new class name. You should find them in `src/Resources/CoreNodeModels/Large(Small)Icons` if I'm not wrong.

aparajit-pratap:
You might also need to update this file for the icons accordingly for the library tests to pass: `test/ViewExtensionLibraryTests/resources/libraryItems.json`

src/Libraries/CoreNodeModels/CoreNodeModelsImages.resx
```diff
@@ -768,6 +768,25 @@
         L0B4qfoOUL6IiIiIiIiIiIiIiIiIiIiIiIiIyG73B7BuavcidkHUAAAAAElFTkSuQmCC
 </value>
   </data>
+  <data name="CoreNodeModels.Logic.RefactoredIf.Large" type="System.Drawing.Bitmap, System.Drawing" mimetype="application/x-microsoft.net.object.bytearray.base64">
```
aparajit-pratap, line 771: I don't think you need to do this (i.e. add new entries here). If I'm not wrong, you can simply change the name of the existing `If` node class bitmap to the new class and rename the png files too.

reddyashish: Yeah sounds good, forgot that we won't be showing the old node in the library anymore.

reddyashish:
@aparajit-pratap changed the png file name in both the locations and updated libraryItems.json. Let me know if those changes look good to you.

aparajit-pratap (approved):

reddyashish:
Addressed the failing test, but somehow the Self-serve job here is still showing it is unstable. 
The job on master-15 passed all the tests: https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/800/
Merging this.

PR #11731, opened 2021-06-03
Add Python dependency info to packages and display it in DPM UI

zeusongit:
### Purpose
https://jira.autodesk.com/browse/DYN-3643

- Adds a method to scan incoming packages for python nodes and dependent packages
- Adds python engine dependency to package metadata.
- Adds UI component to display dependencies by version
- Updated UI to display latest version's dependencies

PS: One time process to scan and update all existing packages is under development.

SS:
Screen Shot 2021-06-03 at 12 38 40 AM

### Declarations

### Reviewers

@DynamoDS/dynamo


src/DynamoCoreWpf/UI/Converters.cs
```diff
@@ -77,6 +77,65 @@ public object ConvertBack(object value, Type targetType, object parameter, Cultu
...
+            return null;
+        }
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            return null;
+        }
+    }
+
+    public class EmptyDepStringToCollapsedConverter : IValueConverter
```
QilongTang, line 112: Let's add some comments for public class

src/DynamoCoreWpf/ViewModels/PackageManager/PublishPackageViewModel.cs
```diff
@@ -1349,14 +1362,23 @@ private IEnumerable<string> BuildPackage()
...
                 Package.Dependencies.Clear();
                 GetAllDependencies().ToList().ForEach(Package.Dependencies.Add);
 
+                Package.HostDependencies = Enumerable.Empty<string>();
+                Package.HostDependencies = SelectedHosts;
+                foreach (var py in GetPythonDependency()) 
+                {
+                    if (!Package.HostDependencies.Contains(py)) 
+                    {
+                        Package.HostDependencies = Package.HostDependencies.Concat(new[] { py });
```
QilongTang, line 1378: Would you remind me does this `HostDependencies` get saved into DB?

zeusongit: yes

src/DynamoCoreWpf/DynamoCoreWpf.csproj
```diff
@@ -1255,6 +1255,10 @@
       <Project>{6e0a079e-85f1-45a1-ad5b-9855e4344809}</Project>
       <Name>Units</Name>
     </ProjectReference>
+    <ProjectReference Include="..\Libraries\PythonNodeModels\PythonNodeModels.csproj">
+      <Project>{8872CA17-C10D-43B9-8393-5C5A57065EB0}</Project>
+      <Name>PythonNodeModels</Name>
```
mjkkirschner, line 1260: While I don't *think* this causes a circular dependency, it is a bit odd to have DynamoCoreWPF depend on a node library.

 If we could figure out a way to solve this with reflection or some other mechanism - for example an interface in DynamoServiecs that might better decouple the core from our libraries and make it clear what the required interfaces actually are.

mjkkirschner: On the other hand, one could make the argument that this is now a core requirement for a UI feature and as such we need to have this reference.

You may want to spend more time generalizing this if we still intend to allow for fully dynamic engine loading in the future - if not, then maybe defining those interfaces is not worth it at the moment.

QilongTang: @mjkkirschner I feel the problem can be traced back to why package manager UI is part of DynamoCoreWpf. To me it seems PM should be more a view extension but not the current way that UI in CoreWpf but data model as extension.. On that note, if the original design is that PM should be part of Core then I feel this dependency is accurate.. Or we could move Python loading to DynamoCore but that also seems not necessary.

mjkkirschner:
@QilongTang - there is actually a PackageManagerUIViewExtension - we just never finished moving the UI out of core and into the view extension :) - more tech tebt.

src/DynamoCoreWpf/DynamoCoreWpf.csproj
mjkkirschner: @QilongTang - there is actually a PackageManagerUIViewExtension - we just never finished moving the UI out of core and into the view extension :) - more tech debt.

QilongTang: Ah.. @mjkkirschner Then are you fine of this temp dependency and we shall get to it. Our team have Package Manager UI refresh this quarter and it seems the tech debt can be address in that epic?

mjkkirschner: sounds good - maybe we just make a jira task for this (python references) and another for the packageManager UI refactor tech debt to track both of them?

QilongTang: Will do @mjkkirschner

QilongTang (approved):

QilongTang (approved):

src/DynamoCoreWpf/DynamoCoreWpf.csproj
QilongTang: Logged as https://jira.autodesk.com/browse/DYN-3765

mjkkirschner:
@zeusongit @QilongTang - I see the tests are failing - my guess is because some python assemblies are being copied to multiple folders now because of the new references.

mark it private false or add a cleanup target to one of the scripts we have to delete those binaries.

PR #11747, opened 2021-06-10
Add dropdown to package preferences dialog for package paths for download

aparajit-pratap:
### Purpose

DYN-3723: Add a dropdown to list all package paths where packages can be downloaded.

WIP:
- [x] Avoid hardcoding Dynamo version for AppData path
- [x] Filter out paths to file names from custom package paths to display in dropdown
- [x] Add property changed event for changing selected path for analytics and in order to update label and tooltip in preferences dialog
- [x] Use selected package path to tell package manager client to download packages to that path 
- [x] Add tooltip displaying full path for long paths cut off from dropdown display
- [x] Add unit test to verify downloading package to path chosen by simulating selection from dropdown
- [ ] UI Automation tests

image

image

### Declarations

- [ ] The level of testing this PR includes is appropriate **(WIP)**

### FYIs

@QilongTang 
@Amoursol 
@Jingyi-Wen


src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -359,6 +359,28 @@ public string DefaultPythonEngine
...
+        /// Currently selected package path where all packages downloaded from the Package Manager
+        /// will be installed.
+        /// </summary>
+        public string SelectedPackagePathForInstall {
+            get
+            {
+                if (string.IsNullOrEmpty(selectedPackagePathForInstall))
+                {
+                    var folder = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
+                    selectedPackagePathForInstall = Path.Combine(Path.Combine(folder, "Dynamo", "Dynamo Core"), "2.12");
```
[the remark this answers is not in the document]
mjkkirschner: this must have been done somewhere already using the version of the assembly.

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
```diff
@@ -711,14 +711,20 @@
                                 <RowDefinition Height="Auto" />
                             </Grid.RowDefinitions>
 
-                            <!--This Grid row contains the Package Paths section-->
+                            <!--This Grid row contains the Node and Package Paths section-->
```
mjkkirschner, line 714: 👍

src/DynamoCore/Configuration/PreferenceSettings.cs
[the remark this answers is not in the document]
aparajit-pratap: It's done in `PathManager` but I don't have access to it from `PreferenceSettings`.

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -236,6 +239,49 @@ public ObservableCollection<string> LanguagesList
...
+        /// </summary>
+        public ObservableCollection<string> PackagePathsForInstall
+        {
+            get
+            {
+                if (packagePathsForInstall == null || !packagePathsForInstall.Any())
+                {
+                    packagePathsForInstall = new ObservableCollection<string>();
+                    packagePathsForInstall.AddRange(preferenceSettings.CustomPackageFolders.Where(
+                        x => x != DynamoModel.StandardLibraryToken));
```
aparajit-pratap, line 254: WIP: I need to filter out paths that include filenames like `...\myAssembly.dll`

mjkkirschner: nice catch.

aparajit-pratap:
> Screen Shot 2021-06-10 at 4 02 51 PM
> 
> @aparajit-pratap relative to the design - it looks like we're missing this label.

Yes, I had seen this but I didn't quite understand why there was one label underneath another unless the `Node and Package Paths` label is the title for the overall page. @Jingyi-Wen?

mjkkirschner:
@aparajit-pratap node and package paths is the expander/title for the section. It includes both the dropdown to set a specific download path, as well as the search paths - each of those sub sections have a label.

Then there is another expander/section for the installed packages.

aparajit-pratap:
> @aparajit-pratap `node and package paths` is the expander/title for the section. It includes both the dropdown to set a specific download path, as well as the search paths - each of those sub sections have a label.
> 
> Then there is another expander/section for the installed packages.

Ok, I got it now. That clears my confusion. I was wondering where the `Package/Library Search Paths` section disappeared in the mockups. I thought that was an expander by itself, didn't realize it was part of the `Node and Package Paths` expander.

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -693,6 +751,9 @@ private void Model_PropertyChanged(object sender, PropertyChangedEventArgs e)
                 case nameof(SelectedNumberFormat):
                     description = Res.DynamoViewSettingMenuNumberFormat;
                     goto default;
+                case nameof(SelectedPackagePathForInstall):
+                    description = Res.PreferencesViewSelectedPackagePathForDownload;
```
QilongTang, line 755: Thanks!

Jingyi-Wen:
> @Jingyi-Wen for the `New Package Download Directory` label, I've used the same style as the `Default Python Engine` label, however, I notice that the font size of these labels is bigger than the respective expander labels. I feel that the expander label being a top-level label should have a higher font size.
> image

@aparajit-pratap Great catch! We were using the same font size but different thickness(bold vs semi-bold vs regular) to visually indicate levels in the design.  We might need to adjust the thickness for item labels and expander labels

mjkkirschner:
@Jingyi-Wen @aparajit-pratap - another thought here - when a really long package download path is selected it appears like it's getting cut off - which makes sense since the dialog width is fixed. 

Can we do something like if you hover over the path the tooltip shows the full path?

I think addressing @QilongTang wanted to address making the dialog resizable later after 2.12.

aparajit-pratap:
> @Jingyi-Wen @aparajit-pratap - another thought here - when a really long package download path is selected it appears like it's getting cut off - which makes sense since the dialog width is fixed.
> 
> Can we do something like if you hover over the path the tooltip shows the full path?
> 
> I think addressing @QilongTang wanted to address making the dialog resizable later after 2.12.

Yeah, I was thinking the same thing about having the tooltip. I'll try to add it.

Jingyi-Wen:
> > @Jingyi-Wen @aparajit-pratap - another thought here - when a really long package download path is selected it appears like it's getting cut off - which makes sense since the dialog width is fixed.
> > Can we do something like if you hover over the path the tooltip shows the full path?
> > I think addressing @QilongTang wanted to address making the dialog resizable later after 2.12.
> 
> Yeah, I was thinking the same thing about having the tooltip. I'll try to add it.

Thank you! That sounds good. I thought about this but missed it in mockup, adding it now

src/DynamoPackages/PackageLoader.cs
```diff
@@ -85,6 +85,11 @@ public string DefaultPackagesDirectory
             get { return defaultPackagesDirectoryIndex != -1 ? packagesDirectories[defaultPackagesDirectoryIndex] : null; }
         }
 
+        internal void SetPackagesDownloadDirectory(string downloadDirectory)
+        {
+            defaultPackagesDirectoryIndex = packagesDirectories.IndexOf(downloadDirectory);
```
aparajit-pratap, line 90: This does not work right now as I discovered that for some reason there is a mismatch in the `packagesDirectories` list maintained by the `PackageLoader` and the `CustomPackageFolders` held in the `PreferenceSettings`. For example, the former has the entry `C:\Users\pratapa\AppData\Roaming\Dynamo\Dynamo Core\2.12\packages`, while the latter has `C:\Users\pratapa\AppData\Roaming\Dynamo\Dynamo Core\2.12`, without the `packages` subdirectory.

mjkkirschner: @aparajit-pratap this reminds me of an issue that RSA devs reported to us. I can't find the slack thread at the moment but I think we were both confused. Seems like a bug rather than something intentional.

aparajit-pratap: @mjkkirschner I think this is the one you're referring to: https://autodesk.slack.com/archives/CTACCQ91R/p1596438758024600

src/DynamoCore/Configuration/PathManager.cs
```diff
@@ -150,7 +150,14 @@ public string LogDirectory
 
         public string DefaultPackagesDirectory
         {
-            get { return TransformPath(RootDirectories.First(), PackagesDirectoryName); }
+            get
+            {
+                if (Preferences is PreferenceSettings preferences)
```
mjkkirschner, line 155: I'm just curious about this, are these two different overloads of TransformPath?

mjkkirschner: oh, I see now, Preferences is an IPreferences and you added the `SelectedPackagePathForInstall` to the concrete class.
👍

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -236,6 +241,60 @@ public ObservableCollection<string> LanguagesList
...
+        /// PackagePathsForInstall contains the list of all package paths where
+        /// packages can be installed.
+        /// </summary>
+        public ObservableCollection<string> PackagePathsForInstall
+        {
+            get
+            {
+                if (packagePathsForInstall == null || !packagePathsForInstall.Any())
+                {
+                    var programDataPath = new Uri(Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData) + "\\");
```
mjkkirschner, line 254: I'm not sure - but I think using `Path` instead of URI might be better because it will be cross platform, can you check on that? (ie `\\` should be `//` on linux and mac.

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -236,6 +241,60 @@ public ObservableCollection<string> LanguagesList
...
+                    // Filter Std lib and ProgramData paths from list of paths for download
```
mjkkirschner, line 255: thank you for the comments.

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
```diff
@@ -711,14 +711,25 @@
...
+                                    <Label  Content="{x:Static p:Resources.PreferencesViewPackageDownloadDirectory}" 
+                                            Padding="0,5,5,5"
+                                            FontSize="13"
+                                            Foreground="{StaticResource PreferencesWindowFontColor}"/>
+                                    <ComboBox Grid.Row="1" 
+                                              Width="Auto"
+                                              HorizontalAlignment="Left"
+                                              ItemsSource="{Binding Path=PackagePathsForInstall}"
+                                              SelectedItem="{Binding Path=SelectedPackagePathForInstall}"
+                                              Style="{StaticResource NoBordersComboBoxStyle}"
```
mjkkirschner, line 730: what do you think about modifying this style so it has a dark background?
I added (or will add) this color to the modern dictionary:
https://github.com/DynamoDS/Dynamo/pull/11762/files#diff-094c8df98e6724d704bb015c259cdf5352917a2076c57a4b4ffccaa74d003c4dR2269

mjkkirschner:
@aparajit-pratap - this looks really solid, just a few comments besides the ones inline.
* looks like some conflict from a recent pr.
* unit tests for the new property getters, and preference setting would be good I think.

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -236,6 +241,60 @@ public ObservableCollection<string> LanguagesList
...
+        }
+
+        /// <summary>
+        /// Currently selected package path where new packages will be downloaded.
+        /// </summary>
+        public string SelectedPackagePathForInstall
+        {
+            get
+            {
+                return selectedPackagePathForInstall;
```
pinzart90, line 285: couldn't we just use preferenceSettings.selectedPackagePathForInstall instead of the private selectedPackagePathForInstall ?

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
```diff
@@ -711,14 +711,25 @@
...
                             <Expander x:Name="PackagePathsExpander"
                                         Grid.Row="0"
                                         Style="{StaticResource MenuExpanderStyle}" 
                                         IsExpanded="{Binding PreferencesTabs[Features].ExpanderActive, Converter={StaticResource ExpandersBindingConverter}, ConverterParameter=PackagePathsExpander}"
                                         Header="{x:Static p:Resources.PackagePathsExpanderName}">
                                 <StackPanel Orientation="Vertical" Margin="0,6,0,0">
-                                    <Label>SOME CONTENT</Label>
+                                    <Label  Content="{x:Static p:Resources.PreferencesViewPackageDownloadDirectory}" 
+                                            Padding="0,5,5,5"
+                                            FontSize="13"
```
pinzart90, line 723: If this font is supposed to be re-used in other parts of the UI, maybe we can store it in the styles xaml ?

src/DynamoCore/Configuration/PathManager.cs
pinzart90: do we have other types of implementations of IPreferences ?
If not...I wonder why `SelectedPackagePathForInstall` is not part of the Interface...

aparajit-pratap: We cannot change the interface right now. In the future yes, we should be adding it to the interface.

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
aparajit-pratap: You mean simply to have the public property and remove the private backing field? How would we then implement the getter and setter?

mjkkirschner: I think @pinzart90 means to use the `preferenceSettings.selectedPackagePathForInstall` as the backing field.

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
aparajit-pratap: I tried using this but there was no difference. Perhaps I'm using it incorrectly. I tried using it here by adding:
```
Background="{StaticResource PreferencesWindowItemDarkerBackgroundColor}"
```
and to the `NoBordersComboBoxStyle` in DynamoModern.xaml but neither worked. Any idea how I can use it for the combobox?

mjkkirschner: Did you also add the new color? My PR has not been merged yet so it won't exist unless you also add it.

aparajit-pratap: Yes, I did add the color.

pinzart90 (approved):
LGTM

src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -359,6 +359,34 @@ public string DefaultPythonEngine
         /// </summary>
         private static string defaultPythonEngine;
 
+        internal event Func<string> RequestUserDataFolder;
```
mjkkirschner, line 362: What would happen if there was more than one subscriber to this event?

src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -359,6 +359,34 @@ public string DefaultPythonEngine
...
+        /// Currently selected package path where all packages downloaded from the Package Manager
+        /// will be installed. The default package path for install is the user data directory
+        /// currently used by the Dynamo environment.
+        /// </summary>
+        public string SelectedPackagePathForInstall {
+            get
+            {
+                if (string.IsNullOrEmpty(selectedPackagePathForInstall))
+                {
+                    selectedPackagePathForInstall = OnRequestUserDataFolder();
```
mjkkirschner, line 380: is it safe if this is null (lets so there is no handler)

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -31,11 +32,12 @@ public class PreferencesViewModel : ViewModelBase, INotifyPropertyChanged
         private string savedChangesLabel;
         private string savedChangesTooltip;
         private ObservableCollection<string> languagesList;
+        private ObservableCollection<string> packagePathsForInstall;
         private ObservableCollection<string> fontSizeList;
         private ObservableCollection<string> numberFormatList;
         private ObservableCollection<StyleItem> styleItemsList;
         private StyleItem addStyleControl;
-        private ObservableCollection<string> _pythonEngineList;
+        private ObservableCollection<string> pythonEngineList;
```
mjkkirschner, line 40: 👍

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -259,6 +261,61 @@ public ObservableCollection<string> LanguagesList
...
+                        x => x != DynamoModel.StandardLibraryToken && !x.StartsWith(programDataPath));
+
+                    packagePathsForInstall = new ObservableCollection<string>();
+                    foreach (var path in customPaths)
+                    {
+                        var attr = File.GetAttributes(path);
+
+                        // Add only directory paths
+                        if ((attr & FileAttributes.Directory) == FileAttributes.Directory)
+                            packagePathsForInstall.Add(path);
```
mjkkirschner, line 286: do you mind using {} here?

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
mjkkirschner: I have added a style for this now, we can clean it up later though after these all get merged.

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
mjkkirschner: I'm okay figuring this out later.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -85,7 +86,28 @@ public string DefaultPackagesDirectory
             get { return defaultPackagesDirectoryIndex != -1 ? packagesDirectories[defaultPackagesDirectoryIndex] : null; }
         }
 
-        private int defaultPackagesDirectoryIndex = -1;
+        private static string TransformPath(string root, string userDataFolder, string extension)
```
mjkkirschner, line 89: could we add a summary for this even though it's internal? I'd like to know what type of transformation it's doing without reading the entire function.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -85,7 +86,28 @@ public string DefaultPackagesDirectory
...
+            catch (IOException) { }
+            catch (ArgumentException) { }
+            catch (UnauthorizedAccessException) { }
+
+            return root;
+        }
+
+        internal void SetPackagesDownloadDirectory(string downloadDirectory, string userDataFolder)
+        {
+            defaultPackagesDirectoryIndex = packagesDirectories.IndexOf(
```
mjkkirschner, line 108: what if this returns -1, is that acceptable?

test/DynamoCoreWpfTests/DynamoViewModelUnitTest.cs
```diff
@@ -36,6 +37,16 @@ protected override DynamoModel GetModel()
             return ViewModel.Model;
         }
 
+        protected static string GetAppDataFolder()
+        {
+            var folder = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
+            var dynamoVersion = FileVersionInfo.GetVersionInfo(Assembly.GetExecutingAssembly().Location);
+            var appDataFolder = Path.Combine(Path.Combine(folder, "Dynamo", "Dynamo Core"),
```
mjkkirschner, line 44: are these strings possibly already defined somewhere?

test/Libraries/PackageManagerTests/PackageLoaderTests.cs
```diff
@@ -30,6 +32,16 @@ protected override void GetLibrariesToPreload(List<string> libraries)
             base.GetLibrariesToPreload(libraries);
         }
 
+        private static string GetAppDataFolder()
+        {
+            var folder = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
```
mjkkirschner, line 37: I guess this and the above function could be in utils somewhere?

test/Libraries/PackageManagerTests/PackageManagerTests.csproj
```diff
@@ -61,6 +61,9 @@
     <Reference Include="System.Xml" />
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="..\..\..\src\AssemblySharedInfoGenerator\AssemblySharedInfo.cs">
+      <Link>Properties\AssemblySharedInfo.cs</Link>
```
mjkkirschner, line 65: good catch.

mjkkirschner (approved):
a few small comments - probably not blockers.

src/DynamoCore/Configuration/PreferenceSettings.cs
aparajit-pratap: The reason I added an event because this class doesn't have access to the PathManager and I couldn't think of another way to do this. Any ideas are welcome.

src/DynamoCore/Configuration/PreferenceSettings.cs
aparajit-pratap: Same comment as above. To answer your questions about null and other subscribers, there might be unknown issues but since there is at least one subscriber, I don't see a case where it will be null.

src/DynamoPackages/PackageLoader.cs
aparajit-pratap: It was being assigned -1 even before, the core logic is the same as before so I didn't think much about it.

src/DynamoCore/Configuration/PreferenceSettings.cs
aparajit-pratap: If the event approach is not ideal, then I can set the path in preference settings by calling a method on it from PreferencesViewModel and passing it the path manager - something like:
```
preferenceSettings.SetSelectedPackagePathForInstall(pathManager);
```

mjkkirschner: I'm okay with either approach, since handler doesn't modify any data and just returns the path I think it's safe even if someone added more subscribers.

test/DynamoCoreWpfTests/DynamoViewModelUnitTest.cs
aparajit-pratap: These probably should be defined as constants somewhere as they are hardcoded in several places throughout the codebase. I'm not doing it here as it could clutter this PR.

PR #11749, opened 2021-06-14
Fix the crash when opening Preferences dialog.

reddyashish:
### Purpose

Task: https://jira.autodesk.com/browse/DYN-3746

The crash was happening as the viewModel object was null. Moved the UpdateSavedChangesLabel() call into a "If" block that checks for the viewModel not being null. 

### Declarations

### Reviewers

@QilongTang @zeusongit


QilongTang:
hi @reddyashish , can you talk a bit why the preferences view model could be null in this case? Just trying to connect the changes here to the crash call stack in Jira

reddyashish:
@QilongTang When the preferences window is being opened, the constructor of the PreferencesView is called and the Geometry_Scaling_Checked() is also called because of the radio button being checked. These 2 functions are called asynchronously, so before the PreferencesView constructor call gets completed, the Geometry_Scaling_Checked() is called and the Preferences view model will be null in that case. The UpdateSavedChangesLabel() is only needed when there are new changes made.

QilongTang (approved):
LGTM

PR #11757, opened 2021-06-15
Save warning due unresolved linter issues

SHKnudsen:
### Purpose

This PR adds a new dialog when a user saves the graph while there are unresolved issues in the graph. The Dialog gives two options either continue to save the graph with the issues or cancel the save and show the Graph Status extension.

For unit test purposes a new event `SaveWarningOnUnresolvedIssuesShows` has been added, this is only meant to be used for the tests to verify that this dialog actually shows. Because the `GenericTaskDialog` is not owned by the DynamoWindow we cant verify that its opened that way, hence the need for adding this event.

GraphTypeIssueWarningDialog

SaveWarnings_v2

### Declarations

### Reviewers

@saintentropy @BogdanZavu @nate-peters 

### FYIs

@mjkkirschner


mjkkirschner:
I imagine this will break existing UIAutomation and other serial tests if not normally disabled during testmode - is it?

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -1903,6 +1921,47 @@ public void ShowSaveDialogAndSaveResult(object parameter)
             }
         }
 
+        private bool ShowSaveWarningDialog()
```
mjkkirschner, line 1924: this seems like a super general name for a very specific dialog.

mjkkirschner:
Are there plans to add this idea of "graph type" to primer? @Amoursol @saintentropy @Jingyi-Wen - If I was not aware of GDIR I doubt I would have any idea what this was message was referring to.

SHKnudsen:
> I imagine this will break existing UIAutomation and other serial tests if not normally disabled during testmode - is it?

@mjkkirschner if in testmode the dialog is showed as a modeless dialog, will that be enough to not break the UIAutomation tests?
https://github.com/DynamoDS/Dynamo/pull/11757/files#diff-66aa776eedd00c3ed606e067aa057cee77527430d1ab1ab8a749c42b896a58e5R1943

Amoursol:
> Are there plans to add this idea of "graph type" to primer? @Amoursol @saintentropy @Jingyi-Wen - If I was not aware of GDIR I doubt I would have any idea what this was message was referring to.

No plans from the Morpheus/Turbo side as it stands, albeit we do have plans to update the Primer it it's entirety after the Visual Refresh project lands. @saintentropy - any scope on this side from your end in addition to the Repo / Wiki?

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -1903,6 +1921,47 @@ public void ShowSaveDialogAndSaveResult(object parameter)
...
+                WpfResources.GraphIssuesOnSave_Title,
+                WpfResources.GraphIssuesOnSave_Summary, 
+                WpfResources.GraphIssuesOnSave_Description);
+
+            args.AddRightAlignedButton((int)DynamoModel.ButtonId.Proceed, WpfResources.GraphIssuesOnSave_ProceedBtn);
+            args.AddRightAlignedButton((int)DynamoModel.ButtonId.Cancel, WpfResources.GraphIssuesOnSave_CancelBtn);
+
+            var dialog = new GenericTaskDialog(args);
+
+            if (DynamoModel.IsTestMode)
```
BogdanZavu, line 1943: Overall I think this looks good. 
For the test I'm ok with it as it is but I'm not sure how much value we get since we're not going through the same route when user is at the helm and we're adding some code to make that happen.
How about we just test if the LinterView has opened when there are issues and cancel was pressed ?
I'm thinking to do this : we trigger an event for the cancel action for which we have a handler that will check issues and will open linter view if it has to ( this event will be triggered from both user mode and test mode ).
And this would be the testable part, we will just raise this event from the test, no pressing on Cancel button since we don't have access/can't control it.

saintentropy (approved):

saintentropy:
@mjkkirschner UI tests pass in the self service.  Shouldn't be affected.

PR #11761, opened 2021-06-15
DYN-3780 : fix modified flag for default home workspace

BogdanZavu:
### Purpose
Don't set the modified flag for the default home workspace.

### Declarations

### DIFF
--- a/src/GraphMetadataViewExtension/GraphMetadataViewModel.cs
+++ b/src/GraphMetadataViewExtension/GraphMetadataViewModel.cs
@@ -29,13 +29,12 @@ namespace Dynamo.GraphMetadata
             get { return currentWorkspace.Description; }
             set
             {
-                if (this.currentWorkspace != null && GraphDescription != value)
+                if (currentWorkspace != null && GraphDescription != value)
                 {
-                    this.currentWorkspace.HasUnsavedChanges = true;
+                    MarkCurrentWorkspaceModified();
+                    currentWorkspace.Description = value;
+                    RaisePropertyChanged(nameof(GraphDescription));
                 }
-
-                currentWorkspace.Description = value;
-                RaisePropertyChanged(nameof(GraphDescription));
             }
         }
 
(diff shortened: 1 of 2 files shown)


saintentropy:
LGTM @BogdanZavu.  I think the only question is if eventually we need to come back to the Clear method for the workspace.  I don't think there is an environment or scenario at this time where the workspace Clear() method is called when this ViewExtension is not installed.

mjkkirschner:
Seems like this deserves a regression test.

mjkkirschner (approved):

saintentropy (approved):

src/GraphMetadataViewExtension/GraphMetadataViewModel.cs
```diff
@@ -29,13 +29,12 @@ public string GraphDescription
             get { return currentWorkspace.Description; }
             set
             {
-                if (this.currentWorkspace != null && GraphDescription != value)
+                if (currentWorkspace != null && GraphDescription != value)
                 {
-                    this.currentWorkspace.HasUnsavedChanges = true;
+                    MarkCurrentWorkspaceModified();
```
QilongTang, line 34: This become repeated code in a few places, I would recommend you listen for Model property changed event since you RaisePropertyChanged() any way and you can MarkCurrentWorkspaceModified() in the handler.

BogdanZavu: There is indeed a repetitive pattern across all these properties and we will look to address it as part of the next PRs. Not sure if we want to base that on property changed since we don't want to mark graph as modified every time the underlying property is changed - we want to leave that decision to the caller, in this case, the extension.

QilongTang: @BogdanZavu Sure, on the handler you can check the property name and choose what to do. Just seems a bit cleaner and more scalable in the future to me but I will leave the decision to you. example here: https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs#L692

PR #11765, opened 2021-06-17
Show Run Preview preference setting should persist across Dynamo sessions

reddyashish:
### Purpose

Task: https://jira.autodesk.com/browse/DYN-3749

Currently the Show Run Preview control(in preferences tab) is not serialized in the DynamoSettings.xml, so it would reset back to false whenever Dynamo is restarted. We want to serialize this setting so that the user can save their preference across Dynamo sessions. This preference setting is now saved in DynamoSettings.xml.
Note that this control will only be editable when the current workspace is in Manual run mode. Added the same as a tooltip for that control.

Show Run Preview control

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCoreWpf/Properties/Resources.resx
+++ b/src/DynamoCoreWpf/Properties/Resources.resx
@@ -2405,4 +2405,7 @@ Uninstall the following packages: {0}?</value>
   <data name="PreferencesPackageManagerSettingsTab" xml:space="preserve">
     <value>Package Manager</value>
   </data>
+  <data name="PreferencesViewShowRunPreviewTooltip" xml:space="preserve">
+    <value>Editable only when the current workspace is in Manual run mode.</value>
+  </data>
 </root>
--- a/src/DynamoCoreWpf/Properties/Resources.resx
+++ b/src/DynamoCoreWpf/Properties/Resources.resx
@@ -2406,6 +2406,6 @@ Uninstall the following packages: {0}?</value>
     <value>Package Manager</value>
   </data>
   <data name="PreferencesViewShowRunPreviewTooltip" xml:space="preserve">
-    <value>Editable only when the current workspace is in Manual run mode.</value>
+    <value>Switchable only when the current workspace is in Manual run mode.</value>
   </data>
 </root>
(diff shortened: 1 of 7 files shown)


src/DynamoCoreWpf/Properties/Resources.resx
```diff
@@ -2405,4 +2405,7 @@ Uninstall the following packages: {0}?</value>
   <data name="PreferencesPackageManagerSettingsTab" xml:space="preserve">
     <value>Package Manager</value>
   </data>
+  <data name="PreferencesViewShowRunPreviewTooltip" xml:space="preserve">
+    <value>Editable only when the current workspace is in Manual run mode.</value>
```
QilongTang, line 2409: maybe `Switchable`?

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -211,10 +211,11 @@ public bool RunPreviewIsChecked
         {
             get
             {
-                return dynamoViewModel.ShowRunPreview;
+                return preferenceSettings.ShowRunPreview;
```
QilongTang, line 214: Does the orange node outline still show up in manual mode without setting this?

src/DynamoCoreWpf/Properties/Resources.resx
reddyashish: Yeah will change it.

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
reddyashish: The orange node outline is showing up for the newly modified nodes but when the workspace is reopened and the toggle button is switched on, it is not picking up the setting. Looking into it.

reddyashish: @QilongTang Previously whenever the current workspace was closed, the RunType on DynamoViewModel was being reset to Automatic and that would reset the ShowRunPreview flag to 'false'. I have changed that behavior for ShowRunPreview, so that the ShowRunPreview flag will continue to honor the user preference.

reddyashish: dynamoViewModel.ShowRunPreview is being set in the setter method whenever RunPreviewIsChecked control is changed. So the orange outline functionality would be fine.

QilongTang (approved):

PR #11769, opened 2021-06-18
Select up and downstream Nodes via right click context menu

martinstacey:
### Purpose

The purpose of this PR is to extend the functionality of node selection by using its right-click context menu. 
It enables users to select nodes upstream/downstream / upstream and downstream, as shown in the image below:

upstreamDownstream

This PR is part of a series that improve node selection, note annotations, wires and graphs. 

### Declarations

### Reviewers

@Amoursol 
@QilongTang 

### FYIs

@SHKnudsen 
@M-JULIANI


QilongTang (commented):
Looks solid but can you confirm the current undo/redo framework does not cover the selection of upstream/downstream command? Not that it is important or required given it is easy to get rid of all the selection. Just want to check that

Amoursol:
> Looks solid
> 
> 1. Can you confirm the current undo/redo framework does not cover the selection of upstream/downstream command? Not that it is important or required given it is easy to get rid of all the selection. Just want to check that
> 2. @Amoursol @martinstacey I notice the current behavior does not include any annotations associated with the up/downstream nodes, so selection of up/downstream nodes and followed by movement may still be somewhat limited. I feel it might be worth to include groups as well but curious what you think? Does that disable any node generic batch operation?
> 3. Can you add a unit test for `SelectDownSteamdAndUpStream...()` call to increase code coverage on our end, really appreciate it!

Very good point on annotation Aaron. @martinstacey can you please look into whether or not you can select groups as well in this flow?

martinstacey:
> * Can you confirm the current undo/redo framework does not cover the selection of upstream/downstream command? Not that it is important or required given it is easy to get rid of all the selection. Just want to check that
> * @Amoursol @martinstacey I notice the current behavior does not include any annotations associated with the up/downstream nodes, so selection of up/downstream nodes and followed by movement may still be somewhat limited. I feel it might be worth to include groups as well but curious what you think? Does that disable any node generic batch operation?
> * Can you add a unit test for `SelectDownSteamdAndUpStream...()` call to increase code coverage on our end, really appreciate it!

1. Yeah this is confirmed, selection is not included on the undo/redo framework
crltznode

2. Just to confirm on this point. This would mean that nodes that are not upstream of the node, but in the same group are also selected. This can be really useful for moving nodes around but may be a little bit confusing when using these methods for reviewing workflows.

3. Ok, tests added

Amoursol:
> 2. Just to confirm on this point. This would mean that nodes that are not upstream of the node, but in the same group are also selected. This can be really useful for moving nodes around but may be a little bit confusing when using these methods for reviewing workflows.

@martinstacey - Yes please. Select all upstream nodes and the Group it is in which will enable the user to move that whole section. Please confirm that this also moves the other, non-selected, nodes in that group too when you move the group. 

If we don't do this, the graph could get very messy and the feature not be so useful :)

martinstacey:
@Amoursol @QilongTang
ok added group selection to the upstream-downstream-both commands 

dynamoGroupUpstreamDownstreamSelection

martinstacey:
@Amoursol @QilongTang let me know if you have any more comments on this PR.

QilongTang (approved):
LGTM. Will merge once master bumped to 2.13

PR #11775, opened 2021-06-22
Adjust the node position when a Node-Autocompletion element is connected.

reddyashish:
### Purpose

This PR is to adjust the position of the new nodes that are created from Node-Autocompletion functionality. 

In the current algorithm, the X-axis position of the new node is relative to position of the port and that looks fine. But the Y-axis position of the new node was not being set relative to parent node so that is set accordingly in these changes. If the input node is a slider, we want to adjust the X-axis position(move left) to compensate for the extra width of the slider node. For all other nodes, the current algorithm works fine. 

For the output port, we need a solution to handle placement of multiple nodes as it will be hard to compute how many nodes are already connected to that port and based on that we will need to set the location of the new node. @zeusongit are you covering this part in your task?

**Before:**

https://user-images.githubusercontent.com/43763136/122878553-e24e1f80-d305-11eb-94ee-eae927662abb.mov

**After:**
NodeAutocompletion node alignment

 
### Declarations

### Reviewers

@QilongTang @zeusongit @Amoursol

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
@@ -286,7 +286,7 @@ namespace Dynamo.Wpf.ViewModels
                 // Placing the new node to the left of initial node
                 adjustedX -= initialNode.Width + 50;
 
-                // If the new node is a input slider node, adjust the position on X-axis to compensate for higher width of the slider node.
+                // If the new node is a slider input node, adjust the position on X-axis to compensate for higher width of the slider node.
                 if (Model.CreationName.Contains("Slider")) 
                 {
                     adjustedX -= 450;
--- a/src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
@@ -268,6 +268,7 @@ namespace Dynamo.Wpf.ViewModels
             var id = Guid.NewGuid();
 
             var adjustedX = initialNodeVm.X;
+            var adjustedY = initialNodeVm.Y;
 
             var createAsDownStreamNode = portModel.PortType == PortType.Output;
             // Placing the new node based on which port it is connecting to.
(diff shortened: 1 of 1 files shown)


src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
```diff
@@ -278,16 +279,22 @@ protected virtual void CreateAndConnectToPort(object parameter)
...
             }
             else
             {
                 // Placing the new node to the left of initial node
                 adjustedX -= initialNode.Width + 50;
 
+                // If the new node is a slider input node, adjust the position on X-axis to compensate for higher width of the slider node.
+                if (Model.CreationName.Contains("Slider")) 
+                {
+                    adjustedX -= 450;
```
QilongTang, line 292: Let's define local variable for the 450 and 50 above :)

src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
```diff
@@ -278,16 +279,22 @@ protected virtual void CreateAndConnectToPort(object parameter)
 
                 // Create a new node based on node creation name and connection ports
                 dynamoViewModel.ExecuteCommand(new DynamoModel.CreateAndConnectNodeCommand(id, initialNode.GUID,
-                    Model.CreationName, 0, Model.AutoCompletionNodeElementInfo.PortToConnect, adjustedX, 0, createAsDownStreamNode, false, true));
+                    Model.CreationName, 0, Model.AutoCompletionNodeElementInfo.PortToConnect, adjustedX, adjustedY, createAsDownStreamNode, false, true));
```
QilongTang, line 282: I thought we made the adjusted Y 0 on purpose, @aparajit-pratap @zeusongit can you confirm? Maybe we thought the auto layout algorithm will simply solve that? Anyway, the solution here looks simple enough to be consistent

QilongTang (approved):
Some small comments, the end results looks quite appealing

Amoursol:
@reddyashish looks much better! 😊  Can you please check in the zoomed out way on the new PR to see if it still behaves correctly?

src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
zeusongit: There was some back-and-forth on this I remember. But I dont recall the problem. Anyway it seems to work fine.

src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
```diff
@@ -278,16 +279,22 @@ protected virtual void CreateAndConnectToPort(object parameter)
...
             }
             else
             {
                 // Placing the new node to the left of initial node
                 adjustedX -= initialNode.Width + 50;
 
+                // If the new node is a slider input node, adjust the position on X-axis to compensate for higher width of the slider node.
+                if (Model.CreationName.Contains("Slider"))
```
zeusongit, line 290: I think this will restrict this improvement to only those wide nodes that have `Slider` in its name, can we have a more generic approach here that may cover all wide nodes, such as getting the width of the incoming node and then calculating the `adjustedX`.

reddyashish: The issue is during this function call, the new node is not yet created and we don't know the width of the node. I couldn't think of any other nodes that are so wide. Do you recall any?

zeusongit: Ah..yes I remember the issue now, thanks.

reddyashish:
@Amoursol yes verified that the new node is created relative to the node sizes in the workspace(zoomed in/out).

reddyashish:
Merging this. Will have to verify if these changes cause of any of UI automation tests to fail.

PR #11782, opened 2021-06-23
Properly align nodes placed after using autocomplete on output ports

zeusongit:
### Purpose

Nodes placed after using autocomplete feature on output ports were placed on top of each other.
With this PR it is fixed and notw output nodes will be included in the auto layout selection in order to be displayed properly aligned.

outa

### Declarations

### Reviewers

@DynamoDS/dynamo


reddyashish (commented):
Looks good to me.

src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
```diff
@@ -297,14 +308,11 @@ protected virtual void CreateAndConnectToPort(object parameter)
...
-
-            var inputNodes = initialNode.InputNodes.Values.Where(x => x != null).Select(y => y.Item2);
 
-            // Clear current selections and select all input nodes as we need to perform Auto layout on only the input nodes.
-            DynamoSelection.Instance.ClearSelection();
-            DynamoSelection.Instance.Selection.AddRange(inputNodes);
-            DynamoSelection.Instance.Selection.Add(initialNode);
+                //Select all input nodes as we need to perform Auto layout on only the input nodes
+                var inputNodes = initialNode.InputNodes.Values.Where(x => x != null).Select(y => y.Item2);
+                DynamoSelection.Instance.Selection.AddRange(inputNodes);
```
QilongTang, line 314: I thought these lines should be in else block? Can you test if triggering node autocomplete on out put port would affect input nodes position with these code?

reddyashish: Yes I thought so too. I tested it locally and the input nodes position doesn't change. @zeusongit can verify it as well.

QilongTang: hmm not sure why that was not going to affect behavior, but I suggest we move the code to `else` block for code clarity?

QilongTang:
Other than one comment, looks good to me

src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
zeusongit: @QilongTang @reddyashish It is already in the else block and only runs for input nodes. `createAsDownStreamNode ` is true when autocomplete is triggered on output ports.

zeusongit: Input nodes stay as they are when triggered on outports and vice versa

zeusongit:
If everything looks good can anyone of you approve this? Thanks
@QilongTang @reddyashish

src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
reddyashish: @zeusongit Aaron was saying that lines 312-314 should be inside the else block of createAsDownStreamNode(219-310). I guess, since the nodes are already aligned by auto layout, doing autolayout on the input nodes again wouldn't cause any difference but would be better for the reason Aaron mentioned.

reddyashish:
@zeusongit One minor change. Looks great to me otherwise.

src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
QilongTang: @reddyashish @zeusongit Thanks for the clarification!

QilongTang (approved):

src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
zeusongit: Oh sorry I misunderstood, regarding this, it adds the original parent node on whose port node autocomplete is triggered, so it has to be added in both scenarios and auto layout takes care of the relative position.

PR #11783, opened 2021-06-23
Fix Publish Locally button in publish package dialog box

zeusongit:
### Purpose

Fixed with of the `Publish Locally` button on Publish Package dialog box that was cutting off before.

Screen Shot 2021-06-25 at 1 02 09 PM

### Reviewers

@DynamoDS/dynamo


QilongTang:
hi @zeusongit I am worried that simply reducing the size may cause side effects to localization. Do you think we can keep the min width but reduce the padding between buttons?

zeusongit:
Updated margin instead

QilongTang (approved):

QilongTang:
The failures in self serve do not seem related, merging

PR #11794, opened 2021-06-25
fix crash when entire line of python code is removed during migration

mjkkirschner:
### Purpose

fix crash reported in:
https://github.com/DynamoDS/Dynamo/issues/11712

Screen Shot 2021-06-25 at 5 08 33 PM

see last line of migrator is completely removed generating an `imaginary line change`

### Declarations

Check these if you believe they are true

- [ ] The codebase is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.
- [ ] This PR modifies some build requirements and the readme is updated

PR #11795, opened 2021-06-28
py migration test for imaginary change type

mjkkirschner:
@aparajit-pratap I've verified this test case fails before the change in https://github.com/DynamoDS/Dynamo/pull/11794

PTAL


aparajit-pratap (approved):
I wasn't expecting a UI test case for this but this looks good. Is it also possible to test this using the backend API's or was it a purely UI related crash?

mjkkirschner:
The method we need to test is on a `dependecenyObject` `RichTextBoxHandler` - which is created when the migration window is opened - we could try refactoring it out or mocking the other wpf views needed to test that method - but I'm not sure it's worth the effort at the moment.

PR #11800, opened 2021-06-29
Dyn 3598

mjkkirschner:
### Purpose

Because GD will continue to use `%programData%/Dynamo` for the time being - we want to address continuing package issues that users have encountered when this directory somehow becomes the user's default download location, or is somehow repointed or removed. 
Issues include:
* new packages failing to load because they do not have certificates
* new upload packages failing to load because they do not have certificates.
* GD failing to load

To accomplish this the PR does the following:
* UI edit and delete buttons in package search view are disabled if the path is the commonData/ProgramData path.
* obsoletes `PackageLoader.DefaultPackagesDirectory` and tries to replace its use with `PathManager.DefaultPackagesDirectory` - so there is a single source of truth. I've noticed these can be made out of sync with pathologically bad test cases, and I think it would be good to give `PackageLoader` access to `PathManager` and centralize all paths in `PathManager ( (⚠️ look out of potential refactor followup).
*  Adds some todos hinting at that refactor
* adds some tests, including one that deliberately asserts the wrong thing - I've marked it techDebt and not failure because it should start to fail after `PackageLoader.DefaultPackagesDirectory` is made to be consistent with `PathManager` - and should be updated then.

Screen Shot 2021-06-24 at 1 31 14 PM

### Declarations


src/DynamoCoreWpf/ViewModels/PackageManager/PackagePathViewModel.cs
```diff
@@ -150,7 +153,19 @@ private bool CanMoveDown(int param)
...
+            return RootLocations.IndexOf(Resources.PackagePathViewModel_Standard_Library) != param &&
+                programDataPackagePathIndex != param;
+        }
+
+        private int GetIndexOfProgramDataPackagePath()
+        {
+            var programDataPath = Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData);
+            var programDataPackagePath = RootLocations.Where(x => x.StartsWith(programDataPath)).FirstOrDefault();
+            var programDataPackagePathIndex = RootLocations.IndexOf(programDataPackagePath);
+            return programDataPackagePathIndex;
```
sm6srw, line 168: We are only tracking the first program data directory found and I guess a user could add an additional program data location themselves? That could mess things up big time (even without the changes in this PR) but is not very likely to happen I think. And we are hopefully removing program data directory soon.

sm6srw (approved):
Looks good and things are moving in the right direction.

mjkkirschner:
@sm6srw thanks for review - there are 6 test failures on this branch - I will look into them ASAP and if they require big changes I will ask for another review:

```
Dynamo.PackageManager.Tests.PackageLoaderCustomTest.LoadFromMultipleLocationTest
Dynamo.PackageManager.Tests.PackageLoaderTests.HasValidStandardLibraryAndDefaultPackagesPath
Dynamo.PackageManager.Tests.PackageLoaderTests.HasValidStandardLibraryAndDefaultPackagesPathWhenStandardLibraryTokenIsAddedFirst
Dynamo.PackageManager.Tests.PackageLoaderTests.HasValidStandardLibraryAndDefaultPackagesPathWhenStandardLibraryTokenIsAddedLast
DynamoCoreWpfTests.PythonNodeCustomizationTests.OnTextAreaTextEnteringEventTest
DynamoCoreWpfTests.PythonNodeCustomizationTests.TabWithSpacesMatchesEngine
```

src/DynamoCoreWpf/ViewModels/PackageManager/PackagePathViewModel.cs
mjkkirschner: hey @sm6srw are you suggesting this method should return a list of ints?

mjkkirschner:
Self serve and parallel tests are passing now, not sure what was up with Python tests, could be previous failures somehow effected cleanup...

aparajit-pratap (approved):

PR #11816, opened 2021-07-06
DYN-3625: call shutdown on extensions

BogdanZavu:
### Purpose
Call shutdown for extensions and avoid calling Dispose for existing Shutdown implementations.
Until now Shutdown was not called we assume Dispose was called from somewhere else already.

### Declarations

### Reviewers

@mjkkirschner @saintentropy

### DIFF
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -365,6 +365,18 @@ namespace Dynamo.Models
 
             OnShutdownStarted(); // Notify possible event handlers.
 
+            foreach (var ext in ExtensionManager.Extensions)
+            {
+                try
+                {
+                    ext.Shutdown();
+                }
+                catch (Exception exc)
+                {
+                    Logger.Log($"{ext.Name} :  {exc.Message} during shutdown");
+                }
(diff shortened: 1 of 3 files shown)


QilongTang:
@BogdanZavu Thank you. I did not notice Shutdown() was only called for all view extensions before. This is a good step forward. But can you talk a bit more about the reason we should assume Dispose() called somewhere else?

BogdanZavu:
@QilongTang It is more like a safety measure requested by @mjkkirschner . Previously shutdown was not called and the assumption was that it is possible to have some calls to dispose somewhere that are hard to track.

QilongTang (approved):

QilongTang:
@BogdanZavu Sorry that I cut the release branch too early :) Would you cherry-pick this to the RC branch?

mjkkirschner:
It's not that they are just hard to track, it's that we already call dispose!
https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Extensions/ExtensionManager.cs#L142

BogdanZavu:
@QilongTang https://github.com/DynamoDS/Dynamo/pull/11820

PR #11825, opened 2021-07-07
Metadata Extension Changes

OliverEGreen:
### Purpose

This PR adds a Metadata tab to the Preferences window, and through it the ability to create 'Required Properties'. These properties extend the functionality of the GraphMetadataViewExtension, by adding in user-defined text fields to the UI. The values for these properties may be set per-graph, but also overridden at a 'global' level with values being consistent between graphs. 

Note: Sorry there's some oddities in some of the file diffs (e.g. Resources). Have tried to manually fix these, without luck!

jqhYIQTvvV

### Impact assessment

- [ ] Changes the way 3rd party nodes work (Add ikeough to the reviewer list)
- [ ] Includes migrations (Add ikeough to the reviewer list)
- [ ] Adds or changes a public interface (Add lukechurch to the reviewer list)
- [ ] API change is added to the [API Changes document](https://github.com/DynamoDS/Dynamo/wiki/API-Changes).
- [ ] Affects which components of the system depends on which other components (Add pboyer to the reviewer list)
- [ ] Introduces new components (Add lukechurch to the reviewer list)
- [ ] Has a performance impact (Add keyu to the reviewer list)
- [ ] Changes the installer (Add sharadkjaiswal to the reviewer list)

#### References

None

#### Testing strategy

Added in several Unit tests to DynamoCoreWpfTests:

- CanAddRequiredProperty
- SetRequiredPropertyGlobalValue
- PreferenceSettingsRequiredPropertiesEqualGraphMetadataViewExtensionRequiredProperties
- RequiredPropertyAddedToPreferenceSettingsAppearsInViewExtension
- SetRequiredPropertyGraphValue
- CheckRequiredPropertyGlobalValue

### Review staging

- [ ] Architectural Design Review
- [X] UI Review

### Reviewers

@saintEntropy
@BogdanZavu
@nate-peters

### FYIs

@mjkkirschner


src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -421,6 +422,22 @@ public string SelectedPackagePathForInstall
...
+        /// <summary>
+        /// Collection of RequiredProperties, as set in the DynamoSettings file
+        /// </summary>
+        public ObservableCollection<RequiredProperty> RequiredProperties
+        {
+            get => requiredProperties;
+            set
+            {
+                requiredProperties = value;
+                RaisePropertyChanged(nameof(RequiredProperties));
```
mjkkirschner, line 437: you generally shouldn't need to raise property changes for observable collections...

Amoursol:
@OliverEGreen - Can we please have a think about a more consumable title for the preferences panel than "Metadata"? While common in technical realms, this may be harder to pick up for newer users. 

Is the graph metadata set on a per-graph, or per-install basis?

src/DynamoCoreWpf/Extensions/ViewLoadedParams.cs
```diff
@@ -76,6 +77,11 @@ public Window DynamoWindow
             }
         }
 
+        /// <summary>
+        /// A reference the PreferenceSettings as set by the Dynamo > Preferences window
+        /// </summary>
+        public PreferenceSettings PreferenceSettings { get; }
```
mjkkirschner, line 83: this is already available on startup params, which are available here - `loadedParams.StartupParams. Preferences`
https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Extensions/StartupParams.cs#L25

which returns an interface - it can be cast to a PreferenceSettings concrete type to get the new property you've added.

`IPreferenceSettings` cannot be changed at this time unfortunately for backward compatibility reasons.

Theres a few options here like:
* obsoleting Preferences and replacing it with PreferenceSettings
* creating a new Interface like IPreferenceSettingsRequiredProps and asking users to cast to that

Theres probably others - but we shouldn't just add a new property pointing to the same object.

Jingyi-Wen:
> @OliverEGreen - Can we please have a think about a more consumable title for the preferences panel than "Metadata"? While common in technical realms, this may be harder to pick up for newer users.
> 
> Is the graph metadata set on a per-graph, or per-install basis?
> 
> @Jingyi-Wen thoughts here?

"Graph Properties" or just "Properties" would make more sense for users

OliverEGreen:
Updates made: 

- 'Metadata' tab in Preferences window is changed to 'Properties', 
- PreferenceSettings has been moved to the ViewExtension
- The ObservableCollection no longer calls RaisePropertyChanged

@Amoursol: 'Is the graph metadata set on a per-graph, or per-install basis?'
Required Property values can now be set per graph, or optionally also hard coded into the DynamoSettings.xml file. This is for non-changing values such as 'Company Name'.

Amoursol:
> @Amoursol: 'Is the graph metadata set on a per-graph, or per-install basis?'
> Required Property values can now be set per graph, or optionally also hard coded into the DynamoSettings.xml file. This is for non-changing values such as 'Company Name'.

Perfect! Thanks muchly @OliverEGreen.

QilongTang:
@saintentropy @OliverEGreen Can any of you add an example of what would be added to `PreferenceSetting.xml`? This should be able to help our discussion of best way to serialize it if we need to make any adjustment?

OliverEGreen:
> @saintentropy @OliverEGreen Can any of you add an example of what would be added to `PreferenceSetting.xml`? This should be able to help our discussion of best way to serialize it if we need to make any adjustment?

Sure, so the idea is that some Required Properties should be able to have a 'global' value set for them, because they're not going to change. The key example I'm thinking of is 'Company Name', where you'd want the ability to hard-code your company name, rather than having users manually set its value per-graph.

Here is an example of the serialized data when a Required Property has a global value set:

```  
<RequiredProperty>
      <UniqueId>aa1db1a6-7699-4cf5-9aeb-96691d23991b</UniqueId>
      <Key>Company Name</Key>
      <GlobalValue>Autodesk</GlobalValue>
      <ValueIsGlobal>true</ValueIsGlobal>
</RequiredProperty>
```

And here is an example of the original intended functionality, where a Required Property's name is set, but its value can only be defined per-graph.

```
<RequiredProperty>
      <UniqueId>ab8c237a-a818-4e3c-9d20-657fd0b3849f</UniqueId>
      <Key>Graph Category</Key>
      <ValueIsGlobal>false</ValueIsGlobal>
</RequiredProperty>
```

These live in a collection, so are wrapped between: \<RequiredProperties> ... etc ... \</RequiredProperties>

In the UI, one currently switches between defining globally/locally set Required Property values using the globe icon:

image

OliverEGreen:
> Thanks @OliverEGreen for the example
> 
> 1. I feel `<GlobalValue>Autodesk</GlobalValue>` and `<ValueIsGlobal>true</ValueIsGlobal>` a bit duplicate, how about simply `<value>xxx</value>` and `<IsGlobalValue>xxx</IsGlobalValue>`?
> 2. I think the original design of PreferencesSetting xml is to hold user specific data except for any PII, to @mjkkirschner 's point, we may want to wrap `<RequiredProperty>` to one additional layer like `MetaData` or some block to indicate it is essential part of Preferences but rather than Required. This is still to be discussed FYI: @saintentropy @jasonstratton

I've just remembered why it's `GlobalValue` instead of `value`: because `value` is a key word in setters, so I'd wind up with:

```csharp
set { 
    value = value;
    RaisePropertyChanged(nameof(Value))
}

PR #11826, opened 2021-07-07
Don't migrate SelectedPackagePathForInstall preference

mjkkirschner:
### Purpose
Found that my 2.13 settings file contained a path pointing to 2.12 - we discovered this is because during first time setup all prefs except custom package paths were migrated - now we also exclude this preference from migration.

### Declarations


test/DynamoCoreTests/UserDataMigrationTests.cs
```diff
@@ -357,5 +359,39 @@ public void TestMigration()
...
+            var targetMigrator = DynamoMigratorBase.MigrateBetweenDynamoVersions(
+                mockPathManager.Object);
+
+            var sourcePrefs = PreferenceSettings.Load(settingsFilePath);
+            Assert.AreEqual(sourceVersionDir, sourcePrefs.SelectedPackagePathForInstall);
+
+            // Assert that new SelectedPackagePath is not equal to the old path.
+            Assert.AreNotEqual(sourcePrefs.SelectedPackagePathForInstall,
+                targetMigrator.PreferenceSettings.SelectedPackagePathForInstall);
+        }
```
aparajit-pratap, line 395: Can we assert that `targetMigrator.PreferenceSettings.SelectedPackagePathForInstall` is equal to `currentVersionDir`?

mjkkirschner: @aparajit-pratap I couldn't get this working because in the setter for `PreferenceSettings.SelectionPackagePathForInstall` - if `selectionPackagePathForInstall is null or empty string` then it tries to get the UserDataFolder from the PathManager - but in this test harness DynamoModel has not been created and the event is not hooked up.

I couldn't find a nice way to get access to the internal `SourceMigrator.PreferenceSettings` to hookup the event, but maybe I can mock it. I just didn't think it was worth it for 2.12 as opposed to moving onto the next bug.

Let me know.

aparajit-pratap: That's fine. LGTM!

aparajit-pratap (approved):

PR #11830, opened 2021-07-07
DYN-3818-GuidedTours-MenuEntries

RobertGlobant20:
### Purpose

I added the "Get Started" submenu and the InteractiveGuide MenuItem, so we can show the Get Started popups.
Also in the DynamoView.xaml.cs I added the code needed for creating most of the popups for the GetStarted tour (the content of the popup is read from the Resources file, that's why I added several entries in the resources).

### Declarations

### Reviewers

@QilongTang

### DIFF
--- a/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
+++ b/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
@@ -38,6 +38,7 @@ using Dynamo.Wpf;
 using Dynamo.Wpf.Authentication;
 using Dynamo.Wpf.Controls;
 using Dynamo.Wpf.Extensions;
+using Dynamo.Wpf.UI.GuidedTour;
 using Dynamo.Wpf.Utilities;
 using Dynamo.Wpf.ViewModels.Core;
 using Dynamo.Wpf.Views;
@@ -47,6 +48,7 @@ using Dynamo.Wpf.Views.PackageManager;
 using Dynamo.Wpf.Windows;
 using HelixToolkit.Wpf.SharpDX;
 using ResourceNames = Dynamo.Wpf.Interfaces.ResourceNames;
+using Res = Dynamo.Wpf.Properties.Resources;
 using String = System.String;
 
 namespace Dynamo.Controls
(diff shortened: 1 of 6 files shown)


src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -2322,6 +2324,143 @@ private void Window_PreviewMouseLeftButtonUp(object sender, MouseButtonEventArgs
...
+        }
+
+        /// <summary>
+        /// This method probably will be modified or deleted in the future when the GuideManager and Guide class are created
+        /// For now will be used just for testing/demo purposes since the popups will be created probably in the Guide class.
+        /// </summary>
+        private void ShowGetStartedGuidedTour()
+        {
+
+            Step.TotalSteps = 6;
```
QilongTang, line 2339: Can we define this 6 as const somewhere?

QilongTang:
@RobertGlobant20 Positions of all popups are currently fixed, are we going to work on APIs to get the dynamically?

RobertGlobant20:
> @RobertGlobant20 Positions of all popups are currently fixed, are we going to work on APIs to get the dynamically?

@QilongTang Could you give me more details about what you mean by fixed?
If you are talking about the creation of the six popups in the DynamoView.xaml.cs, yes, in future, we will need to implement the functionality for Guide class and GuidesManager class, then I think the popups creation will be done in the Guide class, my idea was to use a xml where the "guides creator" will define all the guides and steps and then read all the xaml and the popups should be created automatically with all the settings coming from the xml but I still need to work on that design in order to decide if all the information will be deserialized or not.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
RobertGlobant20: @QilongTang This method (private void ShowGetStartedGuidedTour()) was added just for testing purposes, so Aabishkar will be able to validate all the PopupWindow UI implemented (when the SurveyWindow UI is finished also will be added in this method), then I'm not sure if is worth to create a constant for the value 6.
Thanks

QilongTang: I see, then dont worry about it

QilongTang (approved):
LGTM

PR #11831, opened 2021-07-08
DYN-3814 & DYN-3822: Add/remove path does not update default download combo

sm6srw:
### Purpose

This pull request does:
* add a new path to the default download directory combo when a new custom package directory is added.
* update the path in the default download directory combo when a custom directory path is updated
* remove a path in the default download directory combo when a custom directory path is removed.
* decouple the `SelectedPackagePathForInstall` view model property from the settings property.
* initialize `SelectedPackagePathForInstall` and `PackagePathsForInstall` view model properties from the settings every time the preference dialog is launched.
* commit the `SelectedPackagePathForInstall` view model property to the settings every time the preference dialog exits.
* start to threat the default app data folder as a built in folder.

Notes:
* Decouple the selection property from the settings property solved a lot of the issues that we where seeing. WPF can set the selection property to null when the list is manipulated and having a property that immediately reverts back to a string value when set to null really messed things up.
* I added the default app data folder to the list of built in paths. The fact that we add the path right back at the next start if missing points in that direction. Lets discuss at standup.

### Declarations

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
@@ -771,6 +771,7 @@ namespace Dynamo.ViewModels
 
         protected virtual void UnsubscribeAllEvents()
         {
+            preferencesViewModel.UnsubscribeAllEvents();
             UnsubscribeDispatcherEvents();
             UnsubscribeModelUiEvents();
             UnsubscribeModelChangedEvents();
--- a/src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
@@ -278,7 +278,7 @@ namespace Dynamo.ViewModels
         {
             get
             {
-                if (packagePathsForInstall == null)
+                if (packagePathsForInstall == null || !packagePathsForInstall.Any())
                 {
                     var programDataPath = Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData);
                     // Filter Builtin Packages and ProgramData paths from list of paths for download
--- a/src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
@@ -8,6 +8,7 @@ using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.ComponentModel;
+using System.Diagnostics;
 using System.Drawing;
 using System.IO;
 using System.Linq;
@@ -791,14 +792,30 @@ namespace Dynamo.ViewModels
         /// </summary>
         private void PackagePathsViewModel_PackagePathOperations(object sender, PackagePathOperationsEventArgs e)
         {
-            if (e.AddOrRemove)
+            switch (e.Operation)
             {
-                packagePathsForInstall.Add(e.Path);
+                case PackagePathOperationsEventArgs.PackagePathOperations.Add:
+                    packagePathsForInstall.Add(e.Path);
+                    break;
+                case PackagePathOperationsEventArgs.PackagePathOperations.Remove:
+                    if (SelectedPackagePathForInstall == e.Path)
+                    {
+                        SelectedPackagePathForInstall =
+                            packagePathsForInstall.Count > 1 ? packagePathsForInstall[0] : "";
+                    }
+                    packagePathsForInstall.Remove(e.Path);
+                    break;
+                case PackagePathOperationsEventArgs.PackagePathOperations.Update:
+                    if (SelectedPackagePathForInstall == e.OldPath)
+                    {
+                        SelectedPackagePathForInstall = e.Path;
+                    }
+                    var index = packagePathsForInstall.IndexOf(e.OldPath);
+                    if (index != -1)
+                        packagePathsForInstall[index] = e.Path;
+                    break;
             }
-            else
-            {
-                packagePathsForInstall.Remove(e.Path);
-            }
+
         }
 
         /// <summary>
--- a/src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
@@ -7,6 +7,7 @@ using Dynamo.Wpf.ViewModels.Core.Converters;
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.Collections.Specialized;
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Drawing;
@@ -781,7 +782,7 @@ namespace Dynamo.ViewModels
             var packageLoader = dynamoViewModel.Model.GetPackageManagerExtension()?.PackageLoader;            
             PackagePathsViewModel = new PackagePathViewModel(packageLoader, loadPackagesParams, customNodeManager);
 
-            PackagePathsViewModel.PackagePathOperationsEvent += PackagePathsViewModel_PackagePathOperations;
+            PackagePathsViewModel.RootLocations.CollectionChanged += PackagePathsViewModel_RootLocations_CollectionChanged;
 
             this.PropertyChanged += Model_PropertyChanged;
         }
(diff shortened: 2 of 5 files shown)


mjkkirschner:
@sm6srw I'm confused now - I didn't notice before that `PackagePathsForInstall` was only ever updated when it was null or empty - how did it ever get updated? Or was that the problem...

sm6srw:
> @sm6srw I'm confused now - I didn't notice before that `PackagePathsForInstall` was only ever updated when it was null or empty - how did it ever get updated? Or was that the problem...

yes, that is part of the problem.

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -277,7 +278,7 @@ public ObservableCollection<string> PackagePathsForInstall
         {
             get
             {
-                if (packagePathsForInstall == null || !packagePathsForInstall.Any())
```
mjkkirschner, line 280: why make this change?

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -782,9 +781,67 @@ public PreferencesViewModel(DynamoViewModel dynamoViewModel)
             var packageLoader = dynamoViewModel.Model.GetPackageManagerExtension()?.PackageLoader;            
             PackagePathsViewModel = new PackagePathViewModel(packageLoader, loadPackagesParams, customNodeManager);
 
+            PackagePathsViewModel.PackagePathOperationsEvent += PackagePathsViewModel_PackagePathOperations;
```
mjkkirschner, line 784: I would say it would be good to clear this subscription if we can - it's always hard for me to guess which ones will cause memory leaks during testing.

src/DynamoCoreWpf/ViewModels/PackageManager/PackagePathViewModel.cs
```diff
@@ -52,6 +52,36 @@ public class PackagePathEventArgs : EventArgs
         public string Path { get; set; }
     }
 
+    public class PackagePathOperationsEventArgs : EventArgs
```
mjkkirschner, line 55: can this be internal?

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -782,9 +781,67 @@ public PreferencesViewModel(DynamoViewModel dynamoViewModel)
...
+
             this.PropertyChanged += Model_PropertyChanged;
         }
 
+        /// <summary>
+        /// Listen for changes to the custom package paths and update package paths for install accordingly
+        /// </summary>
+        private void PackagePathsViewModel_PackagePathOperations(object sender, PackagePathOperationsEventArgs e)
```
mjkkirschner, line 792: Actually - instead of creating this new class, new event type etc - why not just watch `RootLocations` directly, it's already public and it's an observable collection, it should have events like this already I think.

src/DynamoCoreWpf/ViewModels/PackageManager/PackagePathViewModel.cs
```diff
@@ -52,6 +52,36 @@ public class PackagePathEventArgs : EventArgs
...
+    {
+
+        /// <summary>
+        /// Add -> Add path to custom folder
+        /// Remove -> Remove path to custom folder
+        /// Update -> Update path to custom folder
+        /// </summary>
+        public enum PackagePathOperations
```
mjkkirschner, line 63: see above if ObservableCollection change events will suffice.

src/DynamoCoreWpf/ViewModels/PackageManager/PackagePathViewModel.cs
```diff
@@ -132,8 +172,9 @@ private void RaiseCanExecuteChanged()
         private bool CanDelete(int param)
         {
             var programDataPackagePathIndex = GetIndexOfProgramDataPackagePath();
+            var appDataPackagePathIndex = GetIndexOfDefaultAppDataPackagePath();
             if (RootLocations.IndexOf(Resources.PackagePathViewModel_BuiltInPackages) == param ||
-                    programDataPackagePathIndex == param)
+                    programDataPackagePathIndex == param || appDataPackagePathIndex == param)
```
mjkkirschner, line 177: yeah, while it's funny that nothing can be deleted now, it also makes sense...

src/DynamoCoreWpf/ViewModels/PackageManager/PackagePathViewModel.cs
```diff
@@ -154,10 +195,11 @@ private bool CanMoveDown(int param)
         private bool CanUpdate(int param)
         {
             var programDataPackagePathIndex = GetIndexOfProgramDataPackagePath();
+            var appDataPackagePathIndex = GetIndexOfDefaultAppDataPackagePath();
 
             //editing builtin packages or programData package paths is not allowed.
             return RootLocations.IndexOf(Resources.PackagePathViewModel_BuiltInPackages) != param &&
-                programDataPackagePathIndex != param;
+                programDataPackagePathIndex != param && appDataPackagePathIndex != param;
```
mjkkirschner, line 202: this one is more arguable I guess - if you repoint this path, on the next restart, you actually now have two paths - right?

I'm fine with this though.

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
```diff
@@ -63,6 +63,7 @@ private void SetupPreferencesViewModel(DynamoViewModel dynamoViewModel)
             //Clear the Saved Changes label and its corresponding tooltip when the Preferences Modal is opened
             dynamoViewModel.PreferencesViewModel.SavedChangesLabel = string.Empty;
             dynamoViewModel.PreferencesViewModel.SavedChangesTooltip = string.Empty;
+            dynamoViewModel.PreferencesViewModel.InitPackagePathsForInstall();
```
mjkkirschner, line 66: maybe we can also change the summary of this function.

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
```diff
@@ -92,6 +93,7 @@ private void CloseButton_Click(object sender, RoutedEventArgs e)
                 Actions.Close,
                 Categories.Preferences);
             viewModel.PackagePathsViewModel.SaveSettingCommand.Execute(null);
+            viewModel.CommitPackagePathsForInstall();
```
mjkkirschner, line 96: it's a bit weird to me that some of this logic lives in `PackagePathsViewModel` and some on `PreferencesViewModel` directly, guess it would be nice if it all lived in PackagePathViewModel and was one method we could call.

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
sm6srw: It was needed for a while but not any more. I will revert.

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
sm6srw: Thanks, I missed that.

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
sm6srw: Ok, I will give it a try.

mjkkirschner:
@sm6srw this looks good. I'm still hoping we can find a way to unsubscribe from the root collection collection changed event though?

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
sm6srw: I leave that for a future refactoring.

sm6srw:
> @sm6srw this looks good. I'm still hoping we can find a way to unsubscribe from the root collection collection changed event though?

 I will look at that next.

sm6srw:
Serial tests: https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/814/

src/DynamoCoreWpf/ViewModels/PackageManager/PackagePathViewModel.cs
aparajit-pratap: @sm6srw why are you preventing appdata from being deleted?

aparajit-pratap: Is it because it's the default download path?

aparajit-pratap (approved):
One question but other than that LGTM!

mjkkirschner:
@sm6srw - sorry I think I caused a conflict for your PR.

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -771,6 +771,7 @@ internal void OnNodeViewReady(object nodeView)
 
         protected virtual void UnsubscribeAllEvents()
         {
+            preferencesViewModel.UnsubscribeAllEvents();
```
mjkkirschner, line 774: 👍

src/DynamoCoreWpf/ViewModels/PackageManager/PackagePathViewModel.cs
mjkkirschner: I think - yes - because if a user removes it, it comes right back on restart.

mjkkirschner (approved):

src/DynamoCoreWpf/ViewModels/PackageManager/PackagePathViewModel.cs
sm6srw: Yes, this is the path that we always add at startup if missing. A lot of strange edges cases never materializes if we prevent this path from being removed/edited. I can't see any value in the previous behavior. It is just confusing.

PR #11832, opened 2021-07-08
don't filter missing paths out  & update package prefs ui on prefs open.

mjkkirschner:
### Purpose

PR addresses two issues:
* https://jira.autodesk.com/browse/DYN-3823 - package download path is blank after directory is removed from disk
*https://jira.autodesk.com/browse/DYN-3827-  new package path is not added to package paths when installing package to a specific directory.
    * this also fixes a related issue where importing a .ds or .dll did not add it to the package paths so it was not imported on next reload (IF the preferences UI was opened before closing Dynamo).
    * updates some error messages and notifications with new location of package path setting.

PR does the following:
* does not filter out paths for the dropdown if they do not exist, it tries to filter out paths which end in specific file extensions only - as these are the ones anticipated to be added by the filter defined here: https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs#L2504 - When a path is deleted on disk, is will remain in the dropdown and will be recreated when a package is downloaded to that path. Tested with meshtoolkit.

* each time the preferences dialog is opened, `PackagePathViewModel` updates its `RootLocations` based on the latest `PreferencesSettings.CustomNodePackagePaths` - this was not happening previously, so when the UI was opened and then closed, it would overwrite any changes made directly to the PreferenceSettings model. I am not very happy with this solution - if anyone has other suggestions please let me know. I'm also worried this is a general problem with the architecture of the new UI and the preferences Model. 🤔 .

see below that the .dll path is filtered from the dropdown.
Screen Shot 2021-07-08 at 11 17 42 AM

TODO:
- [x] Tests

### Declarations


aparajit-pratap:
@mjkkirschner is there any benefit to adding the dll file path to the list of package search paths or is it simply to align it with the old dialog? How about adding only the directory containing the dll to this list of search paths?

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -277,23 +277,20 @@ public ObservableCollection<string> PackagePathsForInstall
...
+                    var directoryPaths = customPaths.Where(path => !(Path.HasExtension(path) && allowedFileExtensions.Contains(Path.GetExtension(path).ToLower())));
 
                     packagePathsForInstall = new ObservableCollection<string>();
-                    foreach (var path in customPaths)
+                    foreach (var path in directoryPaths)
                     {
-                        var attr = File.GetAttributes(path);
-
-                        // Add only directory paths
-                        if ((attr & FileAttributes.Directory) == FileAttributes.Directory)
```
aparajit-pratap, line 293: Did this check not filter out the file paths? Shouldn't we be filtering out all file paths, not just .dll and .ds?

mjkkirschner: > @mjkkirschner is there any benefit to adding the dll file path to the list of package search paths or is it simply to align it with the old dialog? How about adding only the directory _containing_ the dll to this list of search paths?

We need to add the .dll path directly to the search paths - it's for loading that binary directly, like we loaded it using `ImportLibrary`


> Did this check not filter out the file paths? Shouldn't we be filtering out all file paths, not just .dll and .ds, as this list is supposed to only contain directories?

This check relies on the path actually existing, the AC we decided on yesterday during planning was not to remove paths which no longer exist from the dropdown - instead we'll just recreate them. So we can't check if a path is a directory using this lookup because the path may not exist. As I explained yesterday - `Path.HasPath` will return true for a path like `.... Dynamo Core/ 2.12` - we don't want to filter that. Because we limit the paths that can be selected to be directories, .dlls, or .ds files this should be safe. It *won't* be safe if a user manually edits the custom package paths to include some other file type.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -1784,6 +1784,10 @@ private void OnPreferencesWindowClick(object sender, RoutedEventArgs e)
         {
             var preferencesWindow = new PreferencesView(this);
             preferencesWindow.WindowStartupLocation = WindowStartupLocation.CenterOwner;
+            // when the preferences window is opened update the package paths UI based on most up to date preferences model.
+            // TODO refactor and generalize this solution after 2.12 release - what about other preferences
+            // that can be updated on the model (from an extension for example) and need to be updated in the UI?
+            (preferencesWindow.PackagePathView.DataContext as PackagePathViewModel)?.InitializeRootLocations();
```
aparajit-pratap, line 1790: Can we raise an event here that is handled by `PackagePathViewModel`? That way `DynamoView `does not need to know about `PackagePathViewModel`.

mjkkirschner: @aparajit-pratap totally, let me give that a shot.

mjkkirschner: @aparajit-pratap ran into some trouble doing using events, came up with another 1 liner that I think is cleaner than this - if you think an event is a better solution, let's talk, can show you the issues I'm running into, -mostly about cleaning up the event subscription.

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
```diff
@@ -49,10 +48,17 @@ public PreferencesView(DynamoView dynamoView)
...
+
+            SyncPreferenceSettingsAndViewModels();
             InitRadioButtonsDescription();
         }
 
+        // solution 1 - preferencesView is reconstructed each time the prefs window is opened
+        // update any state we need to here on any child viewModels.
+        private void SyncPreferenceSettingsAndViewModels()
+        {
+            viewModel?.PackagePathsViewModel?.InitializeRootLocations();
```
sm6srw, line 60: I used a slightly different way in my PR: `PreferencesView::SetupPreferencesViewModel()` is called everytime the dialog is opened.

mjkkirschner: @sm6srw - yeah that existing method could work as well, I can just call initialize from there - I'm assuming we're eventually going to find other preferences (besides package prefs) that will now be out of sync with the preferences dialog.

It's not usual for these to be set from anywhere but the UI - EXCEPT, extensions probably set them all over the place - so I think it's good to have a central location to sync them. I'm okay with moving over to this method if you agree?

sm6srw: Yes, I do. Lets reuse the existing method.

mjkkirschner:
@aparajit-pratap @sm6srw I've added tests and used the same method as @sm6srw to do the reset upon opening of the preferences view - PTAL. Running serial tests now.
https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/813/

mjkkirschner:
All tests have passed here.

aparajit-pratap (approved):

PR #11835, opened 2021-07-10
DYN-3815 - Prevent reloading of packages unless there are new packages in list of package paths

aparajit-pratap:
### Purpose

DYN-3815:
Prevent reloading of packages unless there are new packages in the list of package paths. This fixes the following issues:

- Avoid library load failed exception notifications each time packages are reloaded upon closing preferences dialog
- Fix issue of duplicate nodes appearing when existing packages are reloaded

Pending issues:
- [x] ~~Loose custom nodes are not loaded unless a path points explicitly to the `definitions` folder. Since packages are loaded simply by pointing to the top-level folder that contains, `packages`, `Logs`, and `definitions`, one would expect this behavior to be consistent for loose custom nodes (dyf files) as well. Note, however, that this is not a regression and it existed earlier.~~ 

- When a path is removed, the packages/custom nodes in that path continue to remain in the library. This _might_ be expected and is _not_ a regression, however, when the path is re-added, the nodes are duplicated in the library. This too is not a regression. The duplication happens in older versions as well. I propose an improvement task to mark these packages whose paths have been removed, as `pending unload` (or equivalent) so that there is some visual feedback to the user that the paths they have removed will cause some future impact to the nodes they are loaded in the library. FYI: @jasonstratton 
- I also think that it is possible to unload a package (from the VM) including loose custom nodes right after deleting a path and updating the library accordingly. Note, that in the case of binaries, they will still be loaded in the process but at least we can prevent the VM from loading those whose paths have been removed.

### Declarations

### FYIs

@Amoursol


src/DynamoCoreWpf/ViewModels/PackageManager/PackagePathViewModel.cs
```diff
@@ -230,10 +230,14 @@ private void CommitChanges(object param)
             //if paths are modified, reload packages and update prefs.
             if (!setting.CustomPackageFolders.SequenceEqual(newpaths))
             {
+                if(newpaths.Count > setting.CustomPackageFolders.Count)
```
mjkkirschner, line 233: what if a user deletes an existing path and then adds a different one won't this miss the new path?

src/DynamoPackages/PackageLoader.cs
```diff
@@ -94,12 +100,8 @@ public string DefaultPackagesDirectory
         /// <param name="userDataFolder"></param>
         /// <param name="extension">subdirectory or subpath</param>
         /// <returns>combined root and extension path</returns>
-        private static string TransformPath(string root, string userDataFolder, string extension)
+        private static string TransformPath(string root, string extension)
```
mjkkirschner, line 103: can you update the summary - I have a hard time understanding what this function does - and hows its changed.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -386,9 +365,7 @@ public void LoadAll(LoadPackageParams loadPackageParams)
         /// <param name="packages"></param>
         public void LoadPackages(IEnumerable<Package> packages)
         {
-            var packageList = packages.ToList();
```
mjkkirschner, line 389: I vaguely remember this was done to avoid some concurrency bug - can you check the history?

src/DynamoPackages/PackageLoader.cs
```diff
@@ -397,29 +374,61 @@ public void LoadPackages(IEnumerable<Package> packages)
...
+                LoadPackages(LocalPackages);
+            }
+        }
+
+        public void LoadCustomNodesAndPackages(LoadPackageParams loadPackageParams, CustomNodeManager customNodeManager)
+        {
+            var packages = new List<Package>(localPackages);
+            foreach(var package in packages)
+            {
+                localPackages.Remove(package);
```
pinzart90, line 416: what is this doing ?
Your creating a copy of the localPackages.
Then you are removing all the copies from the original localPackages ?
What about a localPackages.clear() or something similar ?

src/DynamoPackages/PackageLoader.cs
```diff
@@ -397,29 +374,61 @@ public void LoadPackages(IEnumerable<Package> packages)
...
+                        pathManager.AddResolutionPath(pkg.BinaryDirectory);
+                    }
+
                 }
             }
-            LoadAll(loadPackageParams);
+
+            if (LocalPackages.Any())
+            {
+                LoadPackages(LocalPackages);
```
pinzart90, line 407: Isn't the LocalPackages array supposed to get populated by the LoadPackages call ? And now we are passing the class member LoadPackages as an argument to the sam class's method ?
or has something changed...?

src/DynamoPackages/PackageLoader.cs
```diff
@@ -397,29 +374,61 @@ public void LoadPackages(IEnumerable<Package> packages)
...
 
-        public void LoadCustomNodesAndPackages(LoadPackageParams loadPackageParams, CustomNodeManager customNodeManager)
+        private void LoadNewPackages(LoadPackageParams loadPackageParams)
         {
-            foreach (var path in loadPackageParams.Preferences.CustomPackageFolders)
+            foreach (var path in loadPackageParams.NewPaths)
             {
-                customNodeManager.AddUninitializedCustomNodesInPath(path, false, false);
-                if (!packagesDirectories.Contains(path))
+                var packagesDirectory = pathManager.PackagesDirectories.Where(x => x.StartsWith(path)).FirstOrDefault();
```
mjkkirschner, line 389: can you explain the why "startsWith" is used here? What case are you trying to cover that is not covered by contains?

src/DynamoPackages/PackageLoader.cs
```diff
@@ -397,29 +374,61 @@ public void LoadPackages(IEnumerable<Package> packages)
...
             }
-            LoadAll(loadPackageParams);
+
+            if (LocalPackages.Any())
+            {
+                LoadPackages(LocalPackages);
+            }
+        }
+
+        public void LoadCustomNodesAndPackages(LoadPackageParams loadPackageParams, CustomNodeManager customNodeManager)
```
mjkkirschner, line 411: can you add a summary tag to this method, also do you think external parties would need to use this from an extension or integration? If not can it be made internal?

src/DynamoPackages/PackageLoader.cs
aparajit-pratap: Yes, I'll update it. This is what it does: 
```
Combines the extension with the root path and returns it if the path exists. 
         If not, the root path is returned unchanged.
```

mjkkirschner:
@aparajit-pratap could you give an overview of not just the effect of your changes but the implementation as well?

src/DynamoPackages/PackageLoader.cs
aparajit-pratap: This is the commit that added the `ToList`: https://github.com/DynamoDS/Dynamo/commit/8d9805e0b615bdb6a6d370723816c5293158b5a0#diff-7ae21a0e381958737cc115ada84cbd9f96efe3315928cbc44b367967923cf55b
I still don't see a reason to do that.

src/DynamoPackages/PackageLoader.cs
aparajit-pratap: The new path added by the user can be a subdirectory of an already existing package directory or a path to a DLL that is a subpath of an existing package directory.

src/DynamoPackages/PackageLoader.cs
aparajit-pratap: `LocalPackages` can be populated by both `ScanPackageDirectory `and `LoadPackages` call to `TryLoadPackageIntoLibrary`. If a package is already added to the list of `LocalPackages` in `ScanPackageDirectory`, it will not be added again in the subsequent call to `TryLoadPackageIntoLibrary`.

src/DynamoPackages/PackageLoader.cs
mjkkirschner: okay, and in those cases we don't want to readd that package - correct?

src/DynamoPackages/PackageLoader.cs
aparajit-pratap: LoadPackages is being called in several places, sometimes with a subset of packages, not necessarily with `LocalPackages` in all places. In this instance, we need to call it with the same class member but that is not always the case.

src/DynamoPackages/PackageLoader.cs
aparajit-pratap: I think this is an existing public method. I'm not sure why it appears as new code in this diff.

mjkkirschner: yes, I see that now - apologies for the noise.

src/DynamoPackages/PackageLoader.cs
aparajit-pratap: Yes, I think I can use clear(). I need to make a copy in `packages` as I need to remember this list so as to add it back in line 426.

src/DynamoCore/Library/LibraryServices.cs
```diff
@@ -627,7 +627,10 @@ internal bool LoadNodeLibrary(string library, bool isExplicitlyImportedLib)
                     throwOnFailure: !isExplicitlyImportedLib));
                 return false;
             }
-            importedLibraries.Add(library);
+            if (!importedLibraries.Contains(library))
```
mjkkirschner, line 630: so are these 2 if statements mostly related to stoping duplicate nodes from being added to the library?

src/DynamoPackages/PackageLoader.cs
mjkkirschner: yeah seems to deserve a comment

src/DynamoCore/Library/LibraryServices.cs
aparajit-pratap: Not really, duplicates are not added to the library anymore because I'm not reloading existing packages unless necessary. I saw that these checks were still needed as there were cases (I can't remember now) where duplicate entries were being added to these lists inadvertently, which were harmless but confusing nonetheless.

src/DynamoPackages/PackageLoader.cs
mjkkirschner: because the diff is hard to read this is the previous code:
```
 public void LoadCustomNodesAndPackages(LoadPackageParams loadPackageParams, CustomNodeManager customNodeManager)
        {
            foreach (var path in loadPackageParams.Preferences.CustomPackageFolders)
            {
                customNodeManager.AddUninitializedCustomNodesInPath(path, false, false);
                if (!packagesDirectories.Contains(path))
                {
                    packagesDirectories.Add(path);
                }
            }
            LoadAll(loadPackageParams);
        }
```

test/Libraries/PackageManagerTests/PackageLoaderTests.cs
```diff
@@ -107,6 +107,91 @@ public void PackageLoaderRequestsExtensionsBeLoaded()
...
+        public void NoPackageNodeDuplicatesOnAddingNewPackagePath()
+        {
+            var pathManager = new Mock<Dynamo.Interfaces.IPathManager>();
+            pathManager.SetupGet(x => x.PackagesDirectories).Returns(
+                () => new List<string> { PackagesDirectory });
+
+            var loader = new PackageLoader(pathManager.Object);
+            var libraryLoader = new ExtensionLibraryLoader(CurrentDynamoModel);
+
+            loader.PackagesLoaded += libraryLoader.LoadPackages;
```
mjkkirschner, line 162: I don't think these will cause issues - but I've been surprised before and it's hard to predict when this could cause a memory leak during testing - if it's easy to do can we either
1. check this branch on self serve on master-15
or 
2. unsubscribe at the end of the test?

src/DynamoPackages/PackageLoader.cs
aparajit-pratap: Correct, I don't want to process the child directory again, just the parent directory, which should take care of all child paths.

src/DynamoPackages/PackageLoader.cs
pinzart90: Ok, I get it now. Thanks

src/DynamoPackages/PackageLoader.cs
aparajit-pratap: Will add a summary.

src/DynamoPackages/PackageLoader.cs
mjkkirschner: but - whats the point? To avoid duplicates?

test/Libraries/PackageManagerTests/PackageLoaderTests.cs
```diff
@@ -905,7 +1029,11 @@ public void LocalizedPackageLocalizedCorrectly()
             Thread.CurrentThread.CurrentCulture = esculture;
             Thread.CurrentThread.CurrentUICulture = esculture;
 
-            var loader = new PackageLoader(new[] { PackagesDirectory }, new[] {string.Empty});
+            var pathManager = new Mock<Dynamo.Interfaces.IPathManager>();
+            pathManager.SetupGet(x => x.PackagesDirectories).Returns(() => new[] { PackagesDirectory });
```
mjkkirschner, line 1033: was this a necessary change for some reason or just a nice cleanup unifying tests?

mjkkirschner (approved):
It looks good to me - in my brief testing it seems to work - just a few questions.

src/DynamoPackages/PackageLoader.cs
aparajit-pratap: Yes, I'm first removing packages from `localPackages` to avoid adding duplicates. Then I'm adding them back to keep this list current as at all times it need to maintain the currently loaded list of packages.

mjkkirschner:
regarding loose custom nodes - is there anything that you think should be updated in the threat model?

test/Libraries/PackageManagerTests/PackageLoaderTests.cs
aparajit-pratap: Yes, this is a cleanup to unify tests. I do not want to continue using the old `PackageLoader` constructors and wish to use the new ones that take in a `PathManager` instance instead so as to set a good example of writing new tests. This will also make it clearer as to which old constructors can be deprecated in the code.

test/Libraries/PackageManagerTests/PackageLoaderTests.cs
aparajit-pratap: @mjkkirschner I've unsubscribed at the end of all `packageloader `tests including existing tests.

aparajit-pratap:
> @aparajit-pratap could you give an overview of not just the effect of your changes but the implementation as well?

Ok, will try. Merging first.

sm6srw (commented):
LGTM!

PR #11840, opened 2021-07-12
Pin Note to Node

martinstacey:
### Purpose

The purpose of this PR is to add functionality to pin a note to a node. It allows the user to tether a note to a node and have them move in tandem.

image

It also makes sure that the note is correctly positioned if the node has warnings

nodeWithWarning

### Declarations

### Reviewers

@Amoursol 
@QilongTang  

### FYIs

@SHKnudsen


Amoursol:
This looks awesome @martinstacey! Perfect behavior 💪 

One other thing I noted (Not for this PR), but we probably want more empty space (buffering) around the pin icon. @Jingyi-Wen thoughts on this?

mjkkirschner:
What happens if you try to copy a note that is pinned to a node? Are two pinned notes created that overlap each other?

Jingyi-Wen:
Agreed with @Amoursol that more spacing especially between the text and icon would be better. 
(Could be something like - if it's pinned, the height of the note extends slightly at the bottom)

martinstacey:
> What happens if you try to copy/paste a note that is pinned to a node? Are two pinned notes created that overlap each other?

The pin behaivior works through selection, so if you select either note or pinned node the other one will be selected, this means that if you copy/paste the note the pinned node will also be selected and copied and pasted as well.

Amoursol:
> > What happens if you try to copy/paste a note that is pinned to a node? Are two pinned notes created that overlap each other?
> 
> @mjkkirschner
> The pin behavior works through selection, so if you select either note or pinned node the other one will be selected, this means that if you copy/paste the note the pinned node will also be selected and copied and pasted as well.

A couple of scenarios at play here I think:
1. What happens when you select a node with a pinned note and copy and paste it?
2. What happens when you select a note that is pinned to a node and copy and paste it?

Best behavior in this would be to copy and paste both, and place them at the screen center (As currently copy/paste of a node works). Is this what you are seeing @martinstacey ?

martinstacey:
> > > What happens if you try to copy/paste a note that is pinned to a node? Are two pinned notes created that overlap each other?
> > 
> > 
> > @mjkkirschner
> > The pin behavior works through selection, so if you select either note or pinned node the other one will be selected, this means that if you copy/paste the note the pinned node will also be selected and copied and pasted as well.
> 
> A couple of scenarios at play here I think:
> 
> 1. What happens when you select a node with a pinned note and copy and paste it?
> 2. What happens when you select a note that is pinned to a node and copy and paste it?
> 
> Best behavior in this would be to copy and paste both, and place them at the screen center (As currently copy/paste of a node works). Is this what you are seeing @martinstacey ?

yes, copy and paste applies to node and note when either is selected, the only issue is that pin relation is lost on the newly created node and note.

copypaste

Amoursol:
> > > > What happens if you try to copy/paste a note that is pinned to a node? Are two pinned notes created that overlap each other?
> > > 
> > > 
> > > @mjkkirschner
> > > The pin behavior works through selection, so if you select either note or pinned node the other one will be selected, this means that if you copy/paste the note the pinned node will also be selected and copied and pasted as well.
> > 
> > 
> > A couple of scenarios at play here I think:
> > 
> > 1. What happens when you select a node with a pinned note and copy and paste it?
> > 2. What happens when you select a note that is pinned to a node and copy and paste it?
> > 
> > Best behavior in this would be to copy and paste both, and place them at the screen center (As currently copy/paste of a node works). Is this what you are seeing @martinstacey ?
> 
> yes, copy and paste applies to node and note when either is selected, the only issue is that pin relation is lost on the newly created node and note.
> 
> copypaste

Ok cool - would it be possible to retain the pinned status during this action?

mjkkirschner:
retaining pin in copy paste will mean adding pin to the xml deserialize / serialize core methods for note I think.
https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Graph/Notes/NoteModel.cs#L61

martinstacey:
> Agreed with @Amoursol that more spacing especially between the text and icon would be better.
> (Could be something like - if it's pinned, the height of the note extends slightly at the bottom)

@Amoursol , @Jingyi-Wen 
Ok, I just added some spacing that is only present if the pin icon is visible. On this pr: ea63f7d
image

Jingyi-Wen:
> > Agreed with @Amoursol that more spacing especially between the text and icon would be better.
> > (Could be something like - if it's pinned, the height of the note extends slightly at the bottom)
> 
> @Amoursol , @Jingyi-Wen
> Ok, I just added some spacing that is only present if the pin icon is visible. On this pr: [ea63f7d](https://github.com/DynamoDS/Dynamo/commit/ea63f7db1fbd2e217751a6cb03bf554f7f1affcf)
> image

Looks good to me, thank you @martinstacey

Amoursol:
> retaining pin in copy paste will mean adding pin to the xml deserialize / serialize core methods for note I think.
> https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Graph/Notes/NoteModel.cs#L61

@mjkkirschner @QilongTang - Thoughts on this? Would be a much better user experience.

Amoursol:
> > Agreed with @Amoursol that more spacing especially between the text and icon would be better.
> > (Could be something like - if it's pinned, the height of the note extends slightly at the bottom)
> 
> @Amoursol , @Jingyi-Wen
> Ok, I just added some spacing that is only present if the pin icon is visible. On this pr: [ea63f7d](https://github.com/DynamoDS/Dynamo/commit/ea63f7db1fbd2e217751a6cb03bf554f7f1affcf)
> image

In the interest of being maybe a little too OCD - is it possible to go half-way between what was, and the new version? There is a balance between legibility and not using up too much screen real-estate we are striving for 😁 

Something like the attached image below.

image

src/DynamoCoreWpf/Views/Core/NoteView.xaml.cs
```diff
@@ -86,9 +88,7 @@ void OnViewModelRequestsSelection(object sender, EventArgs e)
 
         void OnNoteTextPreviewMouseDown(object sender, MouseButtonEventArgs e)
         {
-            System.Guid noteGuid = this.ViewModel.Model.GUID;
-            ViewModel.WorkspaceViewModel.DynamoViewModel.ExecuteCommand(
-                new DynCmd.SelectModelCommand(noteGuid, Keyboard.Modifiers.AsDynamoType()));
+            ViewModel.SelectNoteAndPinnedNode();
```
QilongTang, line 91: If I remember correctly, all the command will be recorded into xml if Dynamo crashes so the code here does make a difference, but probably not matter too much since it is not likely crash would happen immediately after this particular action

src/DynamoCore/Graph/Notes/NoteModel.cs
```diff
@@ -24,6 +26,25 @@ public string Text
...
+        private NodeModel pinnedNode;
+
+        /// <summary>
+        /// NodeModel which this Note is pinned to
+        /// When using the pin to node command  
+        /// note and node become entangled so that 
+        /// if you select and move one the other one 
+        /// moves as well. 
+        /// </summary>
+        public NodeModel PinnedNode
```
QilongTang, line 38: hi @martinstacey  can you specify the reason to keep a reference on the Model level?

martinstacey:
Oops yeah I think I exaggerated a little bit on the spacing. 😂
Just to show you the finished result on both spacing and copying and pasting behavior:

pinToNode

these 2 things are addressed in this pr: 8005f5f

src/DynamoCore/Graph/Notes/NoteModel.cs
martinstacey: This is needed as serialization both for saving the file and copying/pasting use model references.

src/DynamoCoreWpf/Views/Core/NoteView.xaml.cs
martinstacey: @QilongTang I can move the code on the `SelectNoteAndPinnedNode` so that it is on the command

martinstacey: @QilongTang , ok just moved selection to command on PR: c6bf701

src/DynamoCoreWpf/Views/Core/NoteView.xaml.cs
```diff
@@ -87,10 +89,20 @@ void OnViewModelRequestsSelection(object sender, EventArgs e)
...
+            {
+                ViewModel.WorkspaceViewModel.DynamoViewModel.ExecuteCommand(
+                    new DynCmd.SelectModelCommand(noteGuid, Keyboard.Modifiers.AsDynamoType()));
+            }
+            else
+            {
+                var nodeGuid = this.ViewModel.Model.PinnedNode.GUID;
+                var selectionGuids = new List<Guid> { noteGuid, nodeGuid };
+                ViewModel.WorkspaceViewModel.DynamoViewModel.ExecuteCommand(
+                    new DynCmd.SelectModelCommand(selectionGuids, Keyboard.Modifiers.AsDynamoType()));
```
QilongTang, line 102: Thanks!

src/DynamoCoreWpf/UI/Converters.cs
```diff
@@ -1769,6 +1769,19 @@ public object ConvertBack(object value, Type targetType, object parameter, Cultu
         }
     }
 
+    public sealed class NullToPinWidthConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            return value == null ? 0 : 4;
```
QilongTang, line 1776: Can you define `4` as const pin width here? Thanks!

QilongTang (approved):
One comment then LGTM

QilongTang:
Can you fix the following too unit tests?
```
DynamoCoreWpfTests.NoteViewTests.NoteIsPlacedAbovePinnedNode
DynamoCoreWpfTests.NoteViewTests.NoteIsPlacedAbovePinnedNodeInErrorState
```

src/DynamoCoreWpf/UI/Converters.cs
martinstacey: Addressed in 5c96555

martinstacey:
> Can you fix the following too unit tests?
> 
> ```
> DynamoCoreWpfTests.NoteViewTests.NoteIsPlacedAbovePinnedNode
> DynamoCoreWpfTests.NoteViewTests.NoteIsPlacedAbovePinnedNodeInErrorState
> ```

Ok fixed on commit: 7bfe089

PR #11841, opened 2021-07-12
Add ways to pass Analytics Host Entitlement info for out of process integrators

QilongTang:
Please Note:

### Purpose

- [x] Add analytics host entitlement info including parent id and session id. See linked PR: https://git.autodesk.com/Dynamo/Analytics.NET/pull/34
- [x] Add ability to pass in these two field using CLI or CLIWPF
- [x] Added unit tests 

These are done so that out of progress integrators like FormIt or Alias could pass in analytics info.

### Declarations

### Reviewers

@DynamoDS/dynamo 

### FYIs


src/DynamoCore/Updates/UpdateManager.cs
```diff
@@ -164,6 +165,11 @@ public interface IUpdateManager
         /// Get the current name of the Host
         /// </summary>
         String HostName { get; set; }
+
+        /// <summary>
+        /// Analytics info of the host
+        /// </summary>
+        HostContextInfo HostInfo { get; set; }
```
QilongTang, line 172: I am exposing this type defined in Analytics.Net here as we did for ProductInfo and CustomerInfo. The only difference is that `HostContextInfo` cant be constructed run time but rely on integrators to pass in, in that case, we will need this type in a couple of other places where I had to add reference.

QilongTang: On another though.. This seems problematic in a way that integrators will now depend on Analytics.Net where this type is defined.. @mjkkirschner @pinzart This seems both API breaking and problematic.. Maybe we only expose two strings `ParentId` and `SessionId` and ask DynamoCore to construct them into `HostContextInfo`. What do you think?

mjkkirschner: yeah, there seems to be 3 problems
* why is this property being added to IUpdateManager - what does this have to do with updates?
* this is breaking a public interface
* new dependency on analytics.net 


I'm fine with your proposed solution I think - not sure exactly what it would look like though - where would DynamoCore get those strings?

QilongTang: @mjkkirschner @pinzart I decided to use the alternative approach on this. By inspecting https://git.autodesk.com/Alias/dynamo-plugin/blob/5324734fa78953c7575715733e59f1952b379d7b/src/DynamoIPC/UpdateManager.cs#L87, I think introducing this type at this level will not only break integration but also introduce extra dependency there

QilongTang: @mjkkirschner These strings will be empty for Sandbox but passed in by integrators using CLI or when they define update manager. I think another good place might be `StartConfiguration` but given `HostName` is also in UpdateManager, for now I would like they all live in the same place

src/DynamoCore/Logging/DynamoAnalyticsClient.cs
```diff
@@ -178,9 +180,10 @@ public DynamoAnalyticsClient(DynamoModel dynamoModel)
             
             var hostName = string.IsNullOrEmpty(dynamoModel.HostName) ? "Dynamo" : dynamoModel.HostName;
 
-            string buildId = "", releaseId = "";
-            Version version;
-            if (Version.TryParse(dynamoModel.Version, out version))
+            hostInfo = new HostContextInfo() { ParentId = dynamoModel.UpdateManager.ParentId, SessionId = dynamoModel.UpdateManager.SessionId };
```
QilongTang, line 183: We only reference `HostContextInfo` here so Analytics.Net refrence is kept only for DynamoCore

src/DynamoCore/Logging/DynamoAnalyticsClient.cs
```diff
@@ -232,7 +235,7 @@ public void Start()
                     RegisterADPTracker(service);
 
                 //If not ReportingAnalytics, then set the idle time as infinite so idle state is not recorded.
-                Service.StartUp(product, new UserInfo(Session.UserId), TimeSpan.FromMinutes(30));
+                Service.StartUp(product, new UserInfo(Session.UserId), hostInfo, TimeSpan.FromMinutes(30));
```
QilongTang, line 238: Arguably we can add hostInfo as 4th param and make it optional but since only Dynamo will be consuming this API, I think this should be acceptable

QilongTang:
@DynamoDS/dynamo This PR is now ready for another look.

src/DynamoApplications/StartupUtils.cs
```diff
@@ -220,7 +233,21 @@ public static DynamoModel MakeModel(bool CLImode, string asmPath = "", string ho
...
+        /// <summary>
+        /// Use this overload to construct a DynamoModel when the location of ASM to use is known and host name is known.
+        /// </summary>
+        /// <param name="CLImode">CLI mode starts the model in test mode and uses a separate path resolver.</param>
+        /// <param name="asmPath">Path to directory containing geometry library binaries</param>
+        /// <param name="hostName">Dynamo variation identified by host.</param>
+        /// <param name="parentId">Dynamo host parent id for analytics purpose</param>
+        /// <param name="sessionId">Dynamo host session id for analytics purpose</param>
+        /// <returns></returns>
+        public static DynamoModel MakeModel(bool CLImode, string asmPath = "", string hostName = "", string parentId = "", string sessionId = "")
```
pinzart90, line 247: maybe its time to put these analytics related properties in a separate struct ? might.. make it easier to pass around..

src/DynamoApplications/StartupUtils.cs
```diff
@@ -197,13 +206,17 @@ public static void PreloadShapeManager(ref string geometryFactoryPath, ref strin
         /// which if we are running on mac os or *nix will use different logic then SandboxLookup
         /// </summary>
         /// <param name="hostName">Dynamo variation identified by host.</param>
+        /// <param name="parentId"></param>
+        /// <param name="sessionId"></param>
         /// <returns></returns>
-        private static IUpdateManager InitializeUpdateManager(string hostName = "")
+        private static IUpdateManager InitializeUpdateManager(string hostName = "", String parentId = "", String sessionId = "")
         {
```
pinzart90, line 213: yeah, weird place to put them, UpdateManager
Can you add a separate jira task to refactor this (move them out) ?
Or maybe we can do it as part of this PR ?

src/DynamoApplications/StartupUtils.cs
mjkkirschner: yeah, if this is easy @QilongTang seems like a good idea to wrap them up into some options bag, we won't need to keep adding new constructors for MakeModel or worrying about binary compatibility with overloads.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -542,8 +542,7 @@ protected DynamoModel(IStartConfiguration config)
             geometryFactoryPath = config.GeometryFactoryPath;
 
             IPreferences preferences = CreateOrLoadPreferences(config.Preferences);
-            var settings = preferences as PreferenceSettings;
-            if (settings != null)
+            if (preferences is PreferenceSettings settings)
```
pinzart90, line 545: I think this declaration pattern already checks for null value.
From msdn:
`Beginning with C# 7.0, a declaration pattern with type T matches an expression when an expression result is non-null and any of the following conditions are true:...`
Example:
https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/pattern-matching#null-checks

src/DynamoApplications/StartupUtils.cs
QilongTang: I have wrapped the strings into a struct as part of this PR

src/DynamoApplications/StartupUtils.cs
QilongTang: Will give that a try

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -431,6 +449,10 @@ public interface IStartConfiguration
             /// No update checks or analytics collection should be done.
             /// </summary>
             bool IsHeadless { get; set; }
+            /// <summary>
+            /// Host analytics info
+            /// </summary>
+            HostAnalyticsInfo HostAnalyticsInfo { get; set; }
```
QilongTang, line 455: This is now part of `IStartConfiguration` so integrators can set it there

QilongTang:
@mjkkirschner @pinzart I cleaned up the solution and used a `HostAnalyticsInfo` struct in DynamoModel to hold pieces of info so this is scalable. Also it has been added to `IStartConfiguration` so integrators can set use the config or CLI. Let me know what you think

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -183,6 +196,11 @@ public string Version
         /// </summary>
         public string HostName { get; set; }
```
pinzart90, line 197: maybe obsolete HostName ?

src/DynamoApplications/StartupUtils.cs
```diff
@@ -196,14 +203,14 @@ public static void PreloadShapeManager(ref string geometryFactoryPath, ref strin
         /// for now, building an updatemanager instance requires finding Dynamo install location
         /// which if we are running on mac os or *nix will use different logic then SandboxLookup
         /// </summary>
-        /// <param name="hostName">Dynamo variation identified by host.</param>
+        /// <param name="info">Dynamo host analytics info.</param>
         /// <returns></returns>
-        private static IUpdateManager InitializeUpdateManager(string hostName = "")
+        private static IUpdateManager InitializeUpdateManager(HostAnalyticsInfo info)
```
pinzart90, line 208: Do we still need InitializeUpdateManager(HostAnalyticsInfo) ?
or can we obsolete it and only used COnfig + DynamoModel ?

src/DynamoCore/Models/DynamoModel.cs
mjkkirschner: This is a breaking change - you cannot change a public interface like this - it means that anyone that implemented the interface will no longer conform to it. Instead add the new property to the concrete class DefaultStartConfiguration and if you prefer you can add a new interface.

mjkkirschner (changes requested):

src/DynamoCore/Models/DynamoModel.cs
QilongTang: Will do

src/DynamoApplications/StartupUtils.cs
QilongTang: We still need to initialize it but, we can remove the param. Since this is a private function, no need to obsolete

src/DynamoCore/Models/DynamoModel.cs
QilongTang: Thanks @mjkkirschner . Addressed

mjkkirschner (approved):

QilongTang:
Hi @pinzart @mjkkirschner Addressed all comments. I think FormIt should be covered with the CLI changes but still curious how Alias need to update their code base, if they stick to `IStartConfig` how do they set the `ParentId`?

PR #6967, opened 2016-08-01
Added ScaleFactor property to WorkspaceModel and corresponding WorkspaceEvent

aparajit-pratap:
### Purpose

This PR is for [MAGN-10468] (http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-10468) to subscribe to workspace events in Dynamo from external libraries like ProtoGeometry and pass on the scaling factor set in the Dynamo UI. 

This is necessary for the [UI task] (http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-10476) to prompt the user for the geometry scaling factor and setting it in the DYN file .

### Declarations

Check these if you believe they are true

- [ ] The code base is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.

### Reviewers

@ellensi

PR #11845, opened 2021-07-12
DYN-3834-GeoScaling-GraphRun

RobertGlobant20:
### Purpose

There was an issue that when the Geometry Scaling was updated in a very large graph the Preferences panel was blocked not allowing the user to close it. I think this behavior was because when the Preference panel was opened immediately a graph run was executed (without changing anything in the Preferences panel).
I was NOT able to reproduce the Preferences panel freezing behavior but we noticed that the graph execution was always happening when the Preferences panel was opened (when nothing was changed), then I did a fix in a way that the selected radio button (in geometry scaling section) is updated every time a new workspace is created or a dyn file is loaded (as is happening with the WorkspaceModel.ScaleFactor), so the graph run will be executed just when the ScaleFactor is changed.

### Declarations

### Reviewers

@QilongTang 
@mjkkirschner 

### FYIs

@aparajit-pratap

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
@@ -1,4 +1,5 @@
 ﻿using Dynamo.Configuration;
+using Dynamo.Events;
 using Dynamo.Graph.Workspaces;
 using Dynamo.Logging;
 using Dynamo.Models;
@@ -747,8 +748,7 @@ namespace Dynamo.ViewModels
             //This piece of code will populate all the description text for the RadioButtons in the Geometry Scaling section.
             optionsGeometryScale = new GeometryScalingOptions();
 
-            //This will set the default option for the Geometry Scaling Radio Buttons, the value is comming from the DynamoViewModel
-            optionsGeometryScale.EnumProperty = (GeometryScaleSize)GeometryScalingOptions.ConvertScaleFactorToUI(dynamoViewModel.ScaleFactorLog);
+            UpdateGeoScaleRadioButtonSelected(dynamoViewModel.ScaleFactorLog);
 
             optionsGeometryScale.DescriptionScaleRange = new ObservableCollection<string>();
             optionsGeometryScale.DescriptionScaleRange.Add(string.Format(Res.ChangeScaleFactorPromptDescriptionContent, scaleRanges[GeometryScaleSize.Small].Item2,
@@ -781,7 +781,28 @@ namespace Dynamo.ViewModels
 
             PackagePathsViewModel.RootLocations.CollectionChanged += PackagePathsViewModel_RootLocations_CollectionChanged;
 
+            WorkspaceEvents.WorkspaceSettingsChanged += PreferencesViewModel_WorkspaceSettingsChanged;
+
             PropertyChanged += Model_PropertyChanged;
+        }
+
+        /// <summary>
+        /// This method will be executed everytime the WorkspaceModel.ScaleFactor value is updated.
+        /// </summary>
+        /// <param name="args"></param>
+        private void PreferencesViewModel_WorkspaceSettingsChanged(WorkspacesSettingsChangedEventArgs args)
+        {
+            //The WorkspaceSettingsChanged event is refering to the double ScaleFactor, then we need to make the conversion to Logarithm scale factor ScaleFactorLog       
+            UpdateGeoScaleRadioButtonSelected(Convert.ToInt32(Math.Log10(args.ScaleFactor)));
+        }
+
+        /// <summary>
+        /// This method will update the currently selected Radio Button in the Geometry Scaling section.
+        /// </summary>
+        /// <param name="scaleFactor"></param>
+        public void UpdateGeoScaleRadioButtonSelected(int scaleFactor)
+        {
+            optionsGeometryScale.EnumProperty = (GeometryScaleSize)GeometryScalingOptions.ConvertScaleFactorToUI(scaleFactor);
         }
 
         /// <summary>
--- a/src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
@@ -798,7 +798,7 @@ namespace Dynamo.ViewModels
         /// This method will update the currently selected Radio Button in the Geometry Scaling section.
         /// </summary>
         /// <param name="scaleFactor"></param>
-        public void UpdateGeoScaleRadioButtonSelected(int scaleFactor)
+        private void UpdateGeoScaleRadioButtonSelected(int scaleFactor)
         {
             optionsGeometryScale.EnumProperty = (GeometryScaleSize)GeometryScalingOptions.ConvertScaleFactorToUI(scaleFactor);
         }
--- a/src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
@@ -812,6 +812,7 @@ namespace Dynamo.ViewModels
         {
             PropertyChanged -= Model_PropertyChanged;
             PackagePathsViewModel.RootLocations.CollectionChanged -= PackagePathsViewModel_RootLocations_CollectionChanged;
+            WorkspaceEvents.WorkspaceSettingsChanged -= PreferencesViewModel_WorkspaceSettingsChanged;
         }
 
 
--- a/src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
+++ b/src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
@@ -130,7 +130,8 @@ namespace Dynamo.Wpf.Views
                         Res.PreferencesViewVisualSettingsGeoScaling);
                 }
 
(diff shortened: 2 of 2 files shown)


mjkkirschner:
@sm6srw curious for your thoughts on this as well, it seems like kind of the same problem we just fixed, but fixed in a different way. I'm wondering if theres a way to make this consistent.

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -781,9 +781,30 @@ public PreferencesViewModel(DynamoViewModel dynamoViewModel)
 
             PackagePathsViewModel.RootLocations.CollectionChanged += PackagePathsViewModel_RootLocations_CollectionChanged;
 
+            WorkspaceEvents.WorkspaceSettingsChanged += PreferencesViewModel_WorkspaceSettingsChanged;
```
mjkkirschner, line 784: if we do end up fixing it this way we should clean this up on shutdown, I think @sm6srw just added a spot where the other eventHandlers here are unsubscribed.

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -781,9 +781,30 @@ public PreferencesViewModel(DynamoViewModel dynamoViewModel)
...
+        {
+            //The WorkspaceSettingsChanged event is refering to the double ScaleFactor, then we need to make the conversion to Logarithm scale factor ScaleFactorLog       
+            UpdateGeoScaleRadioButtonSelected(Convert.ToInt32(Math.Log10(args.ScaleFactor)));
+        }
+
+        /// <summary>
+        /// This method will update the currently selected Radio Button in the Geometry Scaling section.
+        /// </summary>
+        /// <param name="scaleFactor"></param>
+        public void UpdateGeoScaleRadioButtonSelected(int scaleFactor)
```
mjkkirschner, line 803: seems this can be internal or private.

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
QilongTang: Agreed

sm6srw: Yes, PreferencesViewModel:UnsubscribeAllEvents() can be used for that purpose.

sm6srw:
> @sm6srw curious for your thoughts on this as well, it seems like kind of the same problem we just fixed, but fixed in a different way. I'm wondering if theres a way to make this consistent.

Our approach is to read the properties we care about when the dialog opens (happens in `PreferencesView:SetupPreferencesViewModel()`) and then write those properties back when the dialog closes (happens in  `PreferencesView:CloseButton_Click()`. 

That might not work in this case if the expectation is that the graph should execute at the time when the property change.

Just curious but is `WorkspaceSettingsChangedEvent` new?

mjkkirschner:
@sm6srw no, I think `WorkspaceSettingsChangedEvent` is old and is used by LibG to actually set the scale factor on the hostfactory - 
https://git.autodesk.com/Dynamo/LibG/blob/3e848bb66fe1a40d80bacc4b82a6a2059060677f/ProtoGeometry/Core/HostFactory.cs#L166

mjkkirschner:
@RobertGlobant20 theres a conflict on this branch.

RobertGlobant20:
> > @sm6srw curious for your thoughts on this as well, it seems like kind of the same problem we just fixed, but fixed in a different way. I'm wondering if theres a way to make this consistent.
> 
> Our approach is to read the properties we care about when the dialog opens (happens in `PreferencesView:SetupPreferencesViewModel()`) and then write those properties back when the dialog closes (happens in `PreferencesView:CloseButton_Click()`.
> 
> That might not work in this case if the expectation is that the graph should execute at the time when the property change.
> 
> Just curious but is `WorkspaceSettingsChangedEvent` new?

I think you are right the graph should be executed when the ScaleFactor changes, the difficult part is that the property is in WorkspaceModel.ScaleFactor and it doesn't have a RaisePropertyChange instead is calling OnWorkspaceSettingsChanged, then I can see that the only way to find if the property changed is by subscribing to the static event WorkspaceSettingsChanged.

Also I can see that the WorkspaceSettingsChanged event was added on 2016, here is the pull request:
https://github.com/DynamoDS/Dynamo/pull/6967

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
RobertGlobant20: @QilongTang @mjkkirschner 
unsubscribing the event in the next commit: https://github.com/DynamoDS/Dynamo/pull/11845/commits/e4e1a7e52bfd42bf592c6be18222c1ae41caad8c

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
RobertGlobant20: @mjkkirschner this method was changed to private in the next commit:
https://github.com/DynamoDS/Dynamo/pull/11845/commits/f3acc0bfdba9db78dfd2e2bd2ab53e2f098fa7c2

RobertGlobant20:
> @RobertGlobant20 theres a conflict on this branch.

@mjkkirschner the conflict was fixed

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
```diff
@@ -96,9 +101,39 @@ private void CloseButton_Click(object sender, RoutedEventArgs e)
...
+                    Log(String.Format("Geometry working range changed to {0} ({1}, {2})",
+                    viewModel.ScaleRange.Item1, viewModel.ScaleRange.Item2, viewModel.ScaleRange.Item3));
+                    viewModel.UpdateSavedChangesLabel();
+                    Dynamo.Logging.Analytics.TrackEvent(
+                        Actions.Switch,
+                        Categories.Preferences,
+                        Res.PreferencesViewVisualSettingsGeoScaling);
+                }
+
+                dynViewModel.ExecuteCommand(new DynamoModel.ForceRunCancelCommand(false, false));
```
aparajit-pratap, line 133: As discussed on slack please make the change back to:
```
var allNodes = dynViewModel.HomeSpace.Nodes;
                dynViewModel.HomeSpace.MarkNodesAsModifiedAndRequestRun(allNodes, forceExecute: true);
```
in place of `ForceRunCancelCommand`.

aparajit-pratap (commented):
@RobertGlobant20 can you add tests simulating the UI setting changes?

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
RobertGlobant20: > @RobertGlobant20 can you add tests simulating the UI setting changes?

@aparajit-pratap right now there is a UI Automation test validating that the Geometry Scaling functionality works (adds a node with a large number that sends a warning due that the Geometry Scaling is medium, then change the Geometry Scaling and the warning disappear).
then should I be adding a new UI Automation test that validates that if the RunMode = Manual and the Geometry Scaling is changed the graph is not executed and also if the the RunMode = Automatic and the Geometry Scaling is changed the graph is executed automatically, right?

aparajit-pratap: I think this test can be added in Dynamo as well, no?

RobertGlobant20: Ok, I will start the test implementation in Dynamo project.

RobertGlobant20: @aparajit-pratap I'm not sure if is possible to test this functionality using UnitTests in Dynamo project, because I was checking that the only way to execute the MarkNodesAsModifiedAndRequestRun() method (in RunMode Manual or Automatic) is by pressing the close button in the Preferences dialog and for doing that I will need to open the Preferences dialog (I not sure if this is possible using Unit Testing).
Also setting the Property (in ViewModel) that binds the Geometry Scaling Radio Buttons will not execute the Geometry_Scaling_Checked() method that sets the ScaleFactor variable unless you do it manually clicking the radio buttons.

Please let me know your ideas for knowing if this is possible otherwise I can create an UI Automation test for this behavior. Thanks

QilongTang (approved):

QilongTang:
@RobertGlobant20 @aparajit-pratap Seems UIAutomation here is more useful by reading the comments from @RobertGlobant20 . I will merge this for now

mjkkirschner:
@RobertGlobant20 here is an example test that opens and closes the prefs dialog.
https://github.com/DynamoDS/Dynamo/blob/master/test/DynamoCoreWpfTests/PackagePathTests.cs#L289

I see that @QilongTang has already merged this - which is fine - but can you please add tests to master.

QilongTang:
> @RobertGlobant20 here is an example test that opens and closes the prefs dialog.
> https://github.com/DynamoDS/Dynamo/blob/master/test/DynamoCoreWpfTests/PackagePathTests.cs#L289
> 
> I see that @QilongTang has already merged this - which is fine - but can you please add tests to master.

Thank you!

PR #11861, opened 2021-07-15
[DYN-3773] Fix right sidebar extension bug

zeusongit:
### Purpose

Fix the bug with right sidebar extension that collapsed the extensions opened after we resize the first extension.

Before:
ex-be

After:
ex-af

### Declarations

### Reviewers

@DynamoDS/dynamo


reddyashish (approved):
Looks good to me.

PR #11867, opened 2021-07-16
DYN-3834-GeoScaling-UnitTest

RobertGlobant20:
### Purpose

I added one unit tests with two parameters, one is for validating that when RunMode=Manual and the MarkNodesAsModifiedAndRequestRun() method is called the graph execution is NOT kicked off.
The other when the RunMode=Automatic and the MarkNodesAsModifiedAndRequestRun() method is called the graph execution is kicked off.

### Declarations

### Reviewers

@QilongTang 
@mjkkirschner

### DIFF
--- a/test/DynamoCoreWpfTests/PreferencesGeometryScalingTests.cs
+++ b/test/DynamoCoreWpfTests/PreferencesGeometryScalingTests.cs
@@ -1,11 +1,9 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.IO;
+using System.Collections.Generic;
 using System.Windows;
 using System.Windows.Controls.Primitives;
 using Dynamo.Graph.Nodes;
-using Dynamo.Tests;
-using Dynamo.ViewModels;
 using Dynamo.Wpf.Views;
 using DynamoCoreWpfTests.Utility;
 using NUnit.Framework;
@@ -13,7 +11,7 @@ using NUnit.Framework;
 namespace DynamoCoreWpfTests
 {
     [TestFixture]
-    class PreferencesGeometryScalingTests : DynamoTestUIBase
+    public class PreferencesGeometryScalingTests : DynamoTestUIBase
     {
         public override void Open(string path)
         {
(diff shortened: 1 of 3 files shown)


test/DynamoCoreWpfTests/PreferencesGeometryScalingTests.cs
```diff
@@ -0,0 +1,72 @@
...
+            base.GetLibrariesToPreload(libraries);
+        }
+
+        /// <summary>
+        /// The test validates that when RunType=Automatic and the MarkNodesAsModifiedAndRequestRun() method is called the graph execution is kicked off
+        /// The test validates that when RunType=Manual and the MarkNodesAsModifiedAndRequestRun() method is called the graph execution is NOT kicked off
+        /// </summary>
+        [Test]
+        [TestCase("Automatic", ElementState.Warning)]
+        [TestCase("Manual", ElementState.Active)]
```
QilongTang, line 38: Nice, but do you think we can these two string from DynamoCoreWPF resources just so in the future if we change those the test still pass?

QilongTang (approved):
One comment, then LGTM

test/DynamoCoreWpfTests/PreferencesGeometryScalingTests.cs
RobertGlobant20: @QilongTang I had to split the tests in two (one for Automatic and other for Manual) due that the Resources strings can not be used as a parameters in nUnit Tests parameters (since parameters expected should be constant values).
https://github.com/DynamoDS/Dynamo/pull/11867/commits/699036291fb12bdd05c06883b1ccf67a5f9d6b2e

mjkkirschner (approved):

test/DynamoCoreWpfTests/PreferencesGeometryScalingTests.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            //Close the Preferences Dialog and due that the ScaleFactor was updated the MarkNodesAsModifiedAndRequestRun() method will be called
+            preferencesWindow.CloseButton.RaiseEvent(new RoutedEventArgs(ButtonBase.ClickEvent));
+            DispatcherUtil.DoEvents();
+
+            //This will get the Circle.ByCenterPointRadius node (connected to the CodeBlock)
+            var nodeView = NodeViewWithGuid("bcf1c893-5311-4dff-9a04-203ffb9d426c");
+
+            //When RunType = Automatic when the graph is executed the ByCenterPointRadius node change status to Warning due that ScaleFactor = Small
+            Assert.AreEqual(nodeView.ViewModel.State, ElementState.Warning);
```
aparajit-pratap, line 59: Can we also assert that this node is not in a warning state before changing the scale factor?

test/DynamoCoreWpfTests/PreferencesGeometryScalingTests.cs
```diff
@@ -0,0 +1,91 @@
...
+            DispatcherUtil.DoEvents();
+
+            //Close the Preferences Dialog and due that the ScaleFactor was updated the MarkNodesAsModifiedAndRequestRun() method will be called
+            preferencesWindow.CloseButton.RaiseEvent(new RoutedEventArgs(ButtonBase.ClickEvent));
+            DispatcherUtil.DoEvents();
+
+            //This will get the Circle.ByCenterPointRadius node (connected to the CodeBlock)
+            var nodeView = NodeViewWithGuid("bcf1c893-5311-4dff-9a04-203ffb9d426c");
+
+            //When RunType = Automatic when the graph is executed the ByCenterPointRadius node change status to Warning due that ScaleFactor = Small
```
aparajit-pratap, line 87: This comment doesn't apply here. Needs to be updated for this case.

QilongTang:
Nice aspects @aparajit-pratap !

test/DynamoCoreWpfTests/PreferencesGeometryScalingTests.cs
RobertGlobant20: @aparajit-pratap the comment was updated in commit: https://github.com/DynamoDS/Dynamo/pull/11867/commits/688b825e39d3edfe446131556757a59360040a5b

test/DynamoCoreWpfTests/PreferencesGeometryScalingTests.cs
RobertGlobant20: @aparajit-pratap an extra assert was added before changing the scale factor.

aparajit-pratap (approved):

PR #11871, opened 2021-07-19
add dump bytecode debug mode

pinzart90:
Moved DebugModes code in DynamoUtilities (so that they can be referenced in low level assemblies)
Added a new Debug mode to trigger creation of bytecode logs on each Graph execution
Logs will be in the Current Working directory/ByteCodeLogs/somelog.log


src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -1462,26 +1512,29 @@ private void SynchronizeInternal(GraphSyncData syncData)
         {
             runnerCore.Options.IsDeltaCompile = true;
 
-            if (syncData == null)
+            using (DebugModes.IsEnabled("DumpByteCode") ? new DebugByteCodeMode(runnerCore) : null)
```
mjkkirschner, line 1515: does `using null` do anything?

pinzart90: MSDN says that the compiler translates the using statement into a try catch finally with a null check in the finally block
`finally
    {
        reader?.Dispose();
    }`
    
    https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement

pinzart90: Also `If a null resource is acquired, then no call to Dispose is made, and no exception is thrown`
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/statements#the-using-statement

src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -1028,6 +1031,52 @@ public interface ILiveRunner
...
+            private ExecutionMode oldExecutionMode;
+
+            private Core core;
+
+            internal DebugByteCodeMode(Core core)
+            {
+                // Setup the log file location
+                var logFolder = Path.Combine(Directory.GetCurrentDirectory(), "ByteCodeLogs");
+                Directory.CreateDirectory(logFolder);
+                var newWriter = File.CreateText(Path.Combine(logFolder, DateTime.Now.ToString("yyyy-MM-ddTHH-mm-ss-fff") + ".log"));
```
mjkkirschner, line 1048: should this be wrapped in a try catch?

mjkkirschner: so to enable or disable logging you need to turn this flag on(or off) and then reboot Dynamo?

mjkkirschner (approved):
It looks fine to me @pinzart90 - two questions though.

src/Engine/ProtoScript/Runners/LiveRunner.cs
pinzart90: no....you can start dynamo and enable the debug mode to start generating logs.
no need for a reboot.

Not sure a try catch is worth it...this is going to be used by devs to debug

mjkkirschner: ok so `SynchronizeInternal` gets called a lot - I guess this means we might have log files per each run - that seems good. 👍

PR #11882, opened 2021-07-23
Dyn 2693 solution 1 - implicit behavior for loading NodeLibraries that contain INodeViewCustomizations

mjkkirschner:
### Purpose

This PR fixes https://jira.autodesk.com/browse/DYN-2693 - I will be sending another PR for another proposed solution that introduces a new explicit flag for this assembly type.

PR does the following:
* uses reflection to identify if assemblies marked as `NodeLibraries` in the package header contain `INodeViewCustomization`s. If the assembly has any customization implementations then the assembly is loaded as if it was an assembly containing `NodeModel` types, this will end up applying the customizations within. Other types are not imported into the VM/Library.
* reflection is used because `INodeViewCustomization` is defined in DynamoCoreWpf and references WPF types.

Consequences:
1. If an assembly contains `INodeViewCustomization` implementations, other types in the assembly (not including `NodeModels` - those are still loaded of course) are not loaded - I believe this is consistent with the current behavior of assemblies marked NodeLib that contain `NodeModel` types.
2. It *could* break packages where a user had dormant and broken `INodeViewCustomization` somewhere in an assembly that was marked `NodeLib` and was being imported. I think this would be rare and easily fixed by removing the previously broken customization from the assembly.

It's also worth noting the TODO I have flagged in the packageLoader where ALL binaries are treated as "nodeLib" - this only occurs if the nodeLib entry is missing from the package header - I assume for handbuilt and legacy packages from before the property existed.

Screen Shot 2021-07-23 at 2 26 41 PM

As discussed I've updated the docs on publishing to add a bit more information on `node_library` - the final statement here could change when address https://jira.autodesk.com/browse/DYN-3895

Screen Shot 2021-08-04 at 6 11 06 PM

TODO:
- [x] test reflection method works so type name change will not break method
- [x] test package loading with UI only assembly correct customizes
- [x] test package loading with UI only assembly is not imported into VM, even if it contains other types.

### Declarations


src/DynamoCore/Models/DynamoModelEvents.cs
```diff
@@ -490,6 +491,8 @@ private void OnShutdownCompleted()
         /// </summary>
         public event Action<ReadyParams> DynamoReady;
         private bool dynamoReady;
+
+
```
aparajit-pratap, line 495: Minor: I think we should revert this file as it's basically unchanged.

src/DynamoCore/Properties/AssemblyInfo.cs
```diff
@@ -39,4 +39,4 @@
 [assembly: InternalsVisibleTo("GenerativeDesign.Dynamo.ViewExtension")]
 [assembly: InternalsVisibleTo("DSCPython")]
 [assembly: InternalsVisibleTo("DynamoPythonTests")]
-[assembly: InternalsVisibleTo("GraphMetadataViewExtension")]
+[assembly: InternalsVisibleTo("GraphMetadataViewExtension")]
```
aparajit-pratap, line 42: I think we can revert this file too.

src/DynamoCoreWpf/NodeViewCustomization/NodeViewCustomizationLoader.cs
```diff
@@ -48,6 +48,14 @@ public static INodeViewCustomizations LoadCustomizations(Assembly assem, ILogger
             return new NodeViewCustomizations(types);
         }
 
+        private static IEnumerable<Type> GetCustomizationTypes(Assembly assem)
```
aparajit-pratap, line 51: Isn't this method the same as `GetCustomizationTypesUsingReflection` in `NodeModelAssemblyLoader`? Can we just have one function?

src/DynamoCoreWpf/NodeViewCustomization/NodeViewCustomizationLoader.cs
```diff
@@ -100,21 +108,11 @@ public static Type MostDerivedCommonBase(IEnumerable<Type> types)
...
-                    x.IsGenericType &&
-                        x.GetGenericTypeDefinition() == generic);
-
-            if (generic == cur || isGenInterf)
-            {
-                return true;
-            }
-
-            return false;
+        internal static bool ContainsNodeViewCustomizationType(Assembly assem)
```
aparajit-pratap, line 113: Same question here - can we reuse the method of the same name defined in `NodeModelAssemblyLoader`?

src/DynamoCoreWpf/Properties/AssemblyInfo.cs
```diff
@@ -43,4 +43,4 @@
 [assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")]
 [assembly: InternalsVisibleTo("PythonNodeModelsWpf")]
 [assembly: InternalsVisibleTo("LintingViewExtension")]
-[assembly: InternalsVisibleTo("GraphMetadataViewExtension")]
+[assembly: InternalsVisibleTo("GraphMetadataViewExtension")]
```
aparajit-pratap, line 46: Revert file - I don't see any real changes.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -676,6 +676,8 @@ private void SubscribeNodeViewCustomizationEvents()
             dynamoViewModel.NodeViewReady += ApplyNodeViewCustomization;
         }
 
+       
+
```
aparajit-pratap, line 680: Revert unchanged file.

src/DynamoPackages/Package.cs
```diff
@@ -325,10 +325,12 @@ internal void AddAssemblies(IEnumerable<PackageAssembly> assems)
...
 
         /// <summary>
-        ///     Enumerates all assemblies in the package
+        ///     Enumerates all assemblies in the package. This method currently has the side effect that it will
+        ///     load all binaries in the package bin folder unless the package is loaded from a special package path
+        ///     I.E. Builtin Packages.
         /// </summary>
         /// <returns>The list of all node library assemblies</returns>
-        internal IEnumerable<PackageAssembly> EnumerateAssembliesInBinDirectory()
+        internal IEnumerable<PackageAssembly> EnumerateAndLoadAssembliesInBinDirectory()
```
aparajit-pratap, line 333: 👍

src/DynamoCoreWpf/NodeViewCustomization/NodeViewCustomizationLoader.cs
mjkkirschner: @aparajit-pratap this function is the same except it doesn't use reflection - if you're okay relying on reflection everywhere I can use that one and put it somewhere both classes can access - what do you think?

src/DynamoCore/Core/NodeModelAssemblyLoader.cs
```diff
@@ -198,6 +199,22 @@ internal static bool ContainsNodeModelSubType(Assembly assem)
...
         }
+        internal static bool ContainsNodeViewCustomizationType(Assembly assem)
+        {
+            return GetCustomizationTypesUsingReflection(assem).Any();
+        }
+
+        internal static IEnumerable<Type> GetCustomizationTypesUsingReflection(Assembly assem)
+        {
+
+            var customizerType = Type.GetType("Dynamo.Wpf.INodeViewCustomization`1,DynamoCoreWpf");
```
pinzart90, line 210: shouldn't we use the fully qualified name here ?
Any chance of ambiguity when running in headless mode (no wpf ) ? Or we always load the wpf assemblies

pinzart90:
I am in favor of this PR (as opposed to https://github.com/DynamoDS/Dynamo/pull/11891)
Less changes in this PR
I guess we would only need to add some comments in the Dynamo wiki, explaining the new behavior

src/DynamoCore/Core/NodeModelAssemblyLoader.cs
mjkkirschner: Hey @pinzart90 - what else would be included in the fully qualified name? this is `namespace.type ,assembly`

In headless mode this type will not be loaded, but theres no reason to load or apply view customizations then anyway since WPF types will not be loaded either.

mjkkirschner: let me give it some thought, but would you agree with the following points:
1. If dynamo is running headless on a system where `DynamoCoreWpf` can actually be loaded, then loading it should not cause issues - because the DynamoView will still never be instantiated even if the assembly is loaded, and so customizations will not be applied (this occurs when node views are created)

2. If dynamo is running headless on a system where `DynamoCoreWpf` cannot be loaded, then this call should return null and we won't try to load any view customizations which we couldn't apply anyway.

I do like your suggestions, though they may be overkill - hard to tell I guess. 

Checking if the assembly has already been loaded may work (I'll have to see if that works based on when this function gets called... it's normally before the view is created...)
ReflectionOnlyLoad context may be the way to go since we only care about checking for the existence of types.

src/DynamoCoreWpf/NodeViewCustomization/NodeViewCustomizationLoader.cs
aparajit-pratap: I'm not sure what you mean? Isn't doing `Assembly.GetTypes` reflection too?

mjkkirschner: yes, you're right of course - sorry, the distinction I was trying to make was in how the two functions find the `INodeViewCustomization` type - one uses a direct reference to the loaded type - the other has to use Type.GetTypes() to search for the type or load it using a string.

aparajit-pratap: Ah, ok, gotcha

aparajit-pratap (approved):
Other than the white space only and `using` only file changes that I think can be reverted, LGTM!

src/DynamoCore/Core/NodeModelAssemblyLoader.cs
pinzart90: 1. I guess the only issue will be that we're loading DynamoCoreWpf + all direct dependencies when there is no need to.
2. It might throw an exception...we should double check the docs.

ReflectionOnlyLoad will not load any of the dependencies (afaik). So trying to get a specific type (or a list of all types) might fail.

If we do introduce a breaking change in the interface and an older dynamocorewpf assembly is loaded, I guess the TypeExtensions.ImplementsGeneric(customizerType, t) will simply fail....
Should not be a common occurrence though. Not sure how we could guard against such a situation without development difficult (as you mentioned)...

pinzart90: If `Checking if the assembly has already been loaded` does not seem reliable enough, then I am fine with the current state of things

pinzart90 (approved):

sm6srw (approved):

mjkkirschner:
thanks for reviews all - just working on tests now.

src/DynamoCore/Core/NodeModelAssemblyLoader.cs
```diff
@@ -198,6 +201,34 @@ internal static bool ContainsNodeModelSubType(Assembly assem)
...
+        internal static bool ContainsNodeViewCustomizationType(Assembly assem)
+        {
+            return GetCustomizationTypesUsingReflection(assem).Any();
+        }
+
+        internal static IEnumerable<Type> GetCustomizationTypesUsingReflection(Assembly assem)
+        {
+            IEnumerable<Type> output = new Type[] { };
+            //to avoid changing when we load DynamoCoreWpf bail out if it's not yet loaded.
+            if (AppDomain.CurrentDomain.GetAssemblies().All(x => !x.FullName.StartsWith("DynamoCoreWpf"))){
```
mjkkirschner, line 213: @pinzart90 this seems to work fine - I'm going to test in Revit context as well just to be sure.

pinzart90 (approved):

mjkkirschner:
3 test failures, looking at them now.

src/DynamoCoreWpf/ViewModels/PackageManager/Docs/PublishPackageDocumentation.html
```diff
@@ -217,10 +209,19 @@ <h4>Package Name:</h4>
     </p>
     <p>
         <b>Select Node Libraries:</b><br />
-        Select the node libraries/DLLs from the list of added files.
+    <p>
+        Specify which of the files you added to the package are the node libraries. Node libraries contain the types and methods that Dynamo will import as nodes.
+    </p>
+    <p>
+        It's also valid to mark an assembly which only contains UI NodeCustomization classes as a Node Library.
```
aparajit-pratap, line 216: Should we say it is `valid` or should we say it is `necessary` to mark such assemblies as Node libraries?

aparajit-pratap: Should we also list all other assemblies that should be marked as node libraries or is that stated elsewhere already?

mjkkirschner: This is the only documentation I could find on `node_libraries` within Dynamo - I couldn't find anything in the dev docs or the primer.

I've tried to describe the valid use cases above (ZT OR Nodemodel&NodeViewCustomization OR  Nodemodel OR NodeViewCustomization )

can you suggest a specific improvement?

pinzart90 (approved):

mjkkirschner:
this passed here: https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/2999/ merging.

PR #11886, opened 2021-07-26
gc crash fix

pinzart90:
How GC works and why:

When a certain memory threshold is reached (or at the end of a graph execution), GC will start up.
Initially all heap elements are marked as white (potentially good for cleanup).
GC works incrementally so that the VM execution does not have long periods of wait time. 
When GC starts up, it will first take a snapshot of the stack -> get all the elements on the stack that are of reference type (pointer, heap, string).
These active stack elements will be marked as gray (meaning that they should not be cleaned up and that they have not yet been fully processed).
After this step, the async process of tracing all references to heap elements begins. During this time the VM execution can startup and modify the stack and the heap (that is the async part).
Tracing heap references means that all the captured, active stack elements are parsed for references to heap elements (ex. an array on the stack that has heap elements as items) and marked as Black (not to be touched by the current cycle of GC).
Once all tracing is done, the remaining heap elements (still marked as white) will be cleaned up.
One important note: The GC algorithm preserves (should preserve) an important invariant – no black objects reference white objects

Animation_of_tri-color_garbage_collection
https://en.wikipedia.org/wiki/Tracing_garbage_collection#Tri-color_marking

The problem:
Strings have a special status in Dynamo. New stack elements (black) can reference white heap strings (marked for cleanup by current GC cycle) because strings can be referenced even if they are out of scope (not on the stack anymore)

As an example:
```
def generateString() {
  return 1+"@" +2;
};

vv = [Imperative]{
  generateString();
// "1@2" will be on the heap
// current stack has no reference type elements
//  --------------------------------GC kicks in due to mem threshold----------------------------
// "1@2" on the heap will be marked as white - since no stack elements have a reference to it
//
// Async VM execution computes a new generateString() fn Call
  cc = generateString();
//
// cc is pushed on the stack and holds a reference to the "1@2" heap element
//
// GC starts the sweep process in which it deletes "1@2" heap element
// ----------------------------------GC is finished---------------------------------------------
  return cc;
};
//
// vv is still on the stack with a reference to a now null heap element
```
Note: I could not reproduce a similar scenario for pointers and arrays.

The proposed fix:
During a GC cycle, if a new string heap allocation is tried and if we match it with an already existing (White) string on the heap, then we mark the existing heap element as black so that it is not collected by the current GC cycle.

### DIFF
--- a/src/Engine/ProtoCore/DSASM/Heap.cs
+++ b/src/Engine/ProtoCore/DSASM/Heap.cs
@@ -301,6 +301,8 @@ namespace ProtoCore.DSASM
         private readonly List<HeapElement> heapElements = new List<HeapElement>();
         private HashSet<int> fixedHeapElements = new HashSet<int>(); 
         private readonly StringTable stringTable = new StringTable();
+        // The expected lifetime of the sweepSet is start of GC to end of GC
+        private HashSet<int> sweepSet = new HashSet<int>();
 
         private const int GC_THRESHOLD = 1024*1024;
         // Totaly allocated StackValues
@@ -309,7 +311,6 @@ namespace ProtoCore.DSASM
         private int gcDebt = GC_THRESHOLD;
 
         private LinkedList<StackValue> grayList;
-        private HashSet<int> sweepSet;
         private List<StackValue> roots;
         private Executive executive;
         private bool isDisposing = false;
@@ -429,10 +430,8 @@ namespace ProtoCore.DSASM
                 if (isDuringGCPropagateCycle && isValidHeapIndex)
                 {
                     var he = heapElements[index];
-                    if (he == null)
-                    {
-                        throw new NullReferenceException($"Null heap element found at index {index} during AllocateStringInternal");
-                    }
+                    Validity.Assert(he != null, $"Null heap element found at index {index} during AllocateStringInternal");
+
                     bool isHeapElementMarkedAsWhite = he.Mark == GCMark.White;// Either not processed by Propagate step yet or processed and found as garbage.
                     bool isHeapElementConsideredForCleanup = sweepSet.Contains(index);
                     if (isHeapElementMarkedAsWhite && isHeapElementConsideredForCleanup)
@@ -682,10 +681,7 @@ namespace ProtoCore.DSASM
                 StackValue value = ptrs.Dequeue();
                 int rawPtr = (int)value.RawData;
                 var hp = heapElements[rawPtr];
-                if (hp == null)
-                {
-                    throw new NullReferenceException($"Null heap element found at index {rawPtr} during garbage collection");
-                }
+                Validity.Assert(hp != null, $"Null heap element found at index {rawPtr} during garbage collection");
 
                 if (hp.Mark == GCMark.Black)
                     continue;
@@ -718,7 +714,10 @@ namespace ProtoCore.DSASM
         /// </summary>
         private void StartCollection()
         {
-            sweepSet = new HashSet<int>(Enumerable.Range(0, heapElements.Count));
+            // We start from a clean sweepSet
+            sweepSet.Clear();
+
+            sweepSet.Concat(Enumerable.Range(0, heapElements.Count));
             sweepSet.ExceptWith(freeList);
             sweepSet.ExceptWith(fixedHeapElements);
 
@@ -777,10 +776,7 @@ namespace ProtoCore.DSASM
             foreach (var ptr in sweepSet)
             {
                 var hp = heapElements[ptr];
-                if (hp == null)
-                {
-                    throw new NullReferenceException($"Null heap element found at index {ptr} during garbage collection");
-                }
+                Validity.Assert(hp != null, $"Null heap element found at index {ptr} during garbage collection");
 
                 if (hp.Mark != GCMark.White)
                     continue;
--- a/src/Engine/ProtoCore/DSASM/Heap.cs
+++ b/src/Engine/ProtoCore/DSASM/Heap.cs
@@ -282,7 +282,7 @@ namespace ProtoCore.DSASM
     // SatB: GC uses a snapshot of the stack and heap when deciding what heap elements need to be collected.
     public class Heap
     {
-        private enum GCState
+        internal enum GCState
         {
             Pause,
             WaitingForRoots,
@@ -296,6 +296,8 @@ namespace ProtoCore.DSASM
             Gray, // Root objects that haven't had all their references traced yet.
             Black // Objects that are in use (are not candidates for garbage collection) and have had all their references traced.
         }
+
+        private Dictionary<GCState, (Action, Action)> testNotifications;
 
         private readonly List<int> freeList = new List<int>();
         private readonly List<HeapElement> heapElements = new List<HeapElement>();
@@ -425,16 +427,16 @@ namespace ProtoCore.DSASM
             if (stringTable.TryGetPointer(str, out index)) {
                 // Any existing heap elements, marked as white, that are in the sweepSet and that are referenced during the sweep cycle will be marked as Black.
                 // This will ensure that no reachable data is mistakenly cleaned up.
-                bool isDuringGCPropagateCycle = gcState == GCState.Propagate;
+                bool isDuringGCCriticalAsyncCycle = gcState != GCState.Pause;// Between the time the GC takes a snapshot of the stack and heap untill GC cycle is over.
                 bool isValidHeapIndex = index >= 0 && index < heapElements.Count;
-                if (isDuringGCPropagateCycle && isValidHeapIndex)
+                if (isDuringGCCriticalAsyncCycle && isValidHeapIndex)
                 {
                     var he = heapElements[index];
                     Validity.Assert(he != null, $"Null heap element found at index {index} during AllocateStringInternal");
 
-                    bool isHeapElementMarkedAsWhite = he.Mark == GCMark.White;// Either not processed by Propagate step yet or processed and found as garbage.
-                    bool isHeapElementConsideredForCleanup = sweepSet.Contains(index);
-                    if (isHeapElementMarkedAsWhite && isHeapElementConsideredForCleanup)
+                    // If heap element is marked as white then it is either not processed by Propagate step yet or processed and found as garbage.
+                    // If the sweepSet does not contain the heap element's index then there is no need to mark it black (since cleanup will not even be tried)
+                    if (he.Mark == GCMark.White && sweepSet.Contains(index))
                     {
                         // Set the heap element's Mark as Black so that it will not get cleaned up.
                         // No need to do a recursive mark on the it since it is just a string and thus cannot have references to other heap elements.
@@ -717,7 +719,7 @@ namespace ProtoCore.DSASM
             // We start from a clean sweepSet
             sweepSet.Clear();
 
-            sweepSet.Concat(Enumerable.Range(0, heapElements.Count));
+            sweepSet.UnionWith(Enumerable.Range(0, heapElements.Count));
             sweepSet.ExceptWith(freeList);
             sweepSet.ExceptWith(fixedHeapElements);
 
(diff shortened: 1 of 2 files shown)


src/Engine/ProtoCore/DSASM/Heap.cs
```diff
@@ -767,6 +803,7 @@ private void Sweep()
             }
 
             gcDebt = totalAllocated > GC_THRESHOLD ? totalAllocated : GC_THRESHOLD;
+            sweepSet.Clear();
```
mjkkirschner, line 806: I didn't see this change mentioned in your PR?

src/Engine/ProtoCore/DSASM/Heap.cs
```diff
@@ -416,7 +421,29 @@ public StackValue AllocatePointer(int size, MetaData metadata)
...
+                bool isDuringGCPropagateCycle = gcState == GCState.Propagate;
+                bool isValidHeapIndex = index >= 0 && index < heapElements.Count;
+                if (isDuringGCPropagateCycle && isValidHeapIndex)
+                {
+                    var he = heapElements[index];
+                    if (he == null)
+                    {
+                        throw new NullReferenceException($"Null heap element found at index {index} during AllocateStringInternal");
+                    }
+                    bool isHeapElementMarkedAsWhite = he.Mark == GCMark.White;// Either not processed by Propage step yet or processed and found as garbage.
```
mjkkirschner, line 436: spelling (propagate step)

src/Engine/ProtoCore/DSASM/Heap.cs
pinzart90: the `SweepSet` is used to capture the state of the heap when a GC cycle starts up. It basically holds the heap indexes that are to be considered for cleanup by the current GC cycle.
It used to be overridden only when another GC cycle started up (with no reason). Since I am using it in the new string allocation logic I need to clean it at the end of every GC cycle.

mjkkirschner: do you need to also get rid of the reassignment call - or thats fine?

mjkkirschner (approved):
LGTM - but please add a test - it seems like the one in your description will work well.

src/Engine/ProtoCore/DSASM/Heap.cs
```diff
@@ -416,7 +421,29 @@ public StackValue AllocatePointer(int size, MetaData metadata)
...
+                // Any existing heap elements, marked as white, that are in the sweepSet and that are referenced during the sweep cycle will be marked as Black.
+                // This will ensure that no reachable data is mistakenly cleaned up.
+                bool isDuringGCPropagateCycle = gcState == GCState.Propagate;
+                bool isValidHeapIndex = index >= 0 && index < heapElements.Count;
+                if (isDuringGCPropagateCycle && isValidHeapIndex)
+                {
+                    var he = heapElements[index];
+                    if (he == null)
+                    {
+                        throw new NullReferenceException($"Null heap element found at index {index} during AllocateStringInternal");
```
aparajit-pratap, line 434: What happens when this exception is thrown? Is it a crash or is it caught somewhere? If it isn't expected in general, we add `Validity.Assert()` statements:
```
Validity.Assert(he != null, "Heap element at given index cannot be null");
```

src/Engine/ProtoCore/DSASM/Heap.cs
```diff
@@ -416,7 +421,29 @@ public StackValue AllocatePointer(int size, MetaData metadata)
...
             int index;
-            if (!stringTable.TryGetPointer(str, out index))
+            if (stringTable.TryGetPointer(str, out index)) {
+                // Any existing heap elements, marked as white, that are in the sweepSet and that are referenced during the sweep cycle will be marked as Black.
+                // This will ensure that no reachable data is mistakenly cleaned up.
+                bool isDuringGCPropagateCycle = gcState == GCState.Propagate;
+                bool isValidHeapIndex = index >= 0 && index < heapElements.Count;
+                if (isDuringGCPropagateCycle && isValidHeapIndex)
+                {
+                    var he = heapElements[index];
```
aparajit-pratap, line 431: Does the index found in the `stringTable` match that in the list of `heapElements`? Where do we ensure that the index in the `stringTable` points to the same element at the same index in the `heapElements` list?

src/Engine/ProtoCore/DSASM/Heap.cs
```diff
@@ -656,6 +682,11 @@ private int RecursiveMark(StackValue root)
                 StackValue value = ptrs.Dequeue();
                 int rawPtr = (int)value.RawData;
                 var hp = heapElements[rawPtr];
+                if (hp == null)
+                {
+                    throw new NullReferenceException($"Null heap element found at index {rawPtr} during garbage collection");
+                }
```
aparajit-pratap, line 688: Can add another `Validity.Assert `here.

src/Engine/ProtoCore/DSASM/Heap.cs
```diff
@@ -746,6 +777,11 @@ private void Sweep()
             foreach (var ptr in sweepSet)
             {
                 var hp = heapElements[ptr];
+                if (hp == null)
+                {
+                    throw new NullReferenceException($"Null heap element found at index {ptr} during garbage collection");
+                }
```
aparajit-pratap, line 783: and here.

pinzart90:
Adding a reliable test is difficult. A simple graph will not suffice since we also have to satisfy some async execution conditions.
I added a more fine grained unit test

src/Engine/ProtoCore/DSASM/Heap.cs
pinzart90: The index in the stringTable and the heap are kept in sync by always cleaning both at the same time.
When a heap element is cleaned up, the stringTable also cleans the matching data.

src/Engine/ProtoCore/DSASM/Heap.cs
pinzart90: I refactored the code to not do reassignment all the time

src/Engine/ProtoCore/DSASM/Heap.cs
```diff
@@ -687,7 +717,10 @@ private int RecursiveMark(StackValue root)
         /// </summary>
         private void StartCollection()
         {
-            sweepSet = new HashSet<int>(Enumerable.Range(0, heapElements.Count));
+            // We start from a clean sweepSet
+            sweepSet.Clear();
+
+            sweepSet.UnionWith(Enumerable.Range(0, heapElements.Count));
```
pinzart90, line 723: UnionWith seems to work fine for this (concat does not work for some reason)

src/Engine/ProtoCore/DSASM/Heap.cs
```diff
@@ -416,7 +424,28 @@ public StackValue AllocatePointer(int size, MetaData metadata)
...
+                bool isDuringGCCriticalAsyncCycle = gcState != GCState.Pause;// Between the time the GC takes a snapshot of the stack and heap untill GC cycle is over.
+                bool isValidHeapIndex = index >= 0 && index < heapElements.Count;
+                if (isDuringGCCriticalAsyncCycle && isValidHeapIndex)
+                {
+                    var he = heapElements[index];
+                    Validity.Assert(he != null, $"Null heap element found at index {index} during AllocateStringInternal");
+
+                    // If heap element is marked as white then it is either not processed by Propagate step yet or processed and found as garbage.
+                    // If the sweepSet does not contain the heap element's index then there is no need to mark it black (since cleanup will not even be tried)
+                    if (he.Mark == GCMark.White && sweepSet.Contains(index))
```
aparajit-pratap, line 440: Does this imply that a heap element will be cleaned up if both are true, i.e., it is marked white and it is contained in the sweep set? If it is contained in the sweep set, does it mean that it is processed and found as garbage?

src/Engine/ProtoCore/DSASM/Heap.cs
aparajit-pratap: Is this simply pre-allocating space in the hash set?

src/Engine/ProtoCore/DSASM/Heap.cs
pinzart90: `Does this imply that a heap element will be cleaned up only if both are true, i.e., it is marked white and it is contained in the sweep set?`
Yes, both need to be true for a heap element to be cleaned up by the GC.

`If it is contained in the sweep set, does it mean that it is processed and found as garbage?`
No. The sweepSet represents the heap elements that are to be considered for cleaning. Not all heap elements are allowed to be cleaned up (ex. constant strings - **fixedHeapElements** and heap elements that have already been cleaned up - **freeList** - and not yet re-used)

there are 2 ways to **not clean up** a heap element. Keep it out of the SweepSet or mark it as gray/black during GC cycle.  Both mechanism are used because of different lifecycles (colors will be reset after each gc cycle....but **freeList** and **fixedHeapElements** persist)

src/Engine/ProtoCore/DSASM/Heap.cs
pinzart90: no, it is actually adding all indexes from 0 to HeapElements.Count
basically takes a snapshot of the heap (using indexes)

mjkkirschner: I think `concat` may not work because it returns another IEnumerable - which means that it may use deferred execution - so the snapshot is not taken when you intended it.

With `UnionWith`, this returns a Set and should execute the `Enumerable.Range` call.

src/Engine/ProtoCore/DSASM/Heap.cs
```diff
@@ -707,17 +740,30 @@ private void StartCollection()
         /// </summary>
         private void SingleStep(bool forceGC)
         {
+            // Testing purposes only.
```
mjkkirschner, line 743: do you think these if statements and construction of a new function within each singleStep could affect performance?

what about using `#if DEBUG`?

src/Engine/ProtoCore/DSASM/Heap.cs
```diff
@@ -416,7 +424,28 @@ public StackValue AllocatePointer(int size, MetaData metadata)
...
+            if (stringTable.TryGetPointer(str, out index))
+            {
+                // Any existing heap elements, marked as white, that are in the sweepSet and that are referenced during the sweep cycle will be marked as Black.
+                // This will ensure that no reachable data is mistakenly cleaned up.
+                bool isDuringGCCriticalAsyncCycle = gcState != GCState.Pause;// Between the time the GC takes a snapshot of the stack and heap untill GC cycle is over.
+                bool isValidHeapIndex = index >= 0 && index < heapElements.Count;
+                if (isDuringGCCriticalAsyncCycle && isValidHeapIndex)
+                {
+                    var he = heapElements[index];
+                    Validity.Assert(he != null, $"Null heap element found at index {index} during AllocateStringInternal");
```
aparajit-pratap, line 436: Sorry for nitpicking: here we are asserting that something cannot be null so I think it will be appropriate to negate the above message and make it sound something like: `Heap element found at index {index} during Allocate cannot be null`

src/Engine/ProtoCore/DSASM/Heap.cs
aparajit-pratap: I agree, that way it is also clearer that these statements are needed for debugging purposes.

pinzart90: are cicd tests run on debug builds ?

mjkkirschner: no, they use release builds.

src/Engine/ProtoCore/DSASM/Heap.cs
pinzart90: Sounds correct

src/Engine/ProtoCore/DSASM/Heap.cs
pinzart90: Just so that we can run it in release builds, I could duplicate the FullGC method for testing purposes ...

aparajit-pratap (approved):

src/Engine/ProtoCore/DSASM/Heap.cs
```diff
@@ -872,6 +906,38 @@ public void FullGC(IEnumerable<StackValue> gcroots, Executive exe)
...
+            while (gcState != GCState.Pause)
+            {
+                (Action start, Action end) fns = (() => { }, () => { });
+                var testCB = testNotifications?.TryGetValue(gcState, out fns) ?? false;
+
+                if (testCB) fns.start();
+                SingleStep(true);
+                if (testCB) fns.end();
+            }
+        }
```
pinzart90, line 939: Maybe this can be a good compromise. 
Upsides:
No performance penalty for the live code.
Can run in cicd test suite
Downside
Code duplication and maintainability

PR #11891, opened 2021-07-27
Dyn 2693 solution 2 - explicit behavior for marking/loading only INodeViewCustomizations from assemblies.

mjkkirschner:
### Purpose


This PR fixes https://jira.autodesk.com/browse/DYN-2693 - a PR showing a simpler alternative approach is here:
https://github.com/DynamoDS/Dynamo/pull/11882

<img width="1384" alt="Screen Shot 2021-07-27 at 5 48 38 PM" src="https://user-images.githubusercontent.com/508936/127232055-8413822a-fcba-42f8-bf87-4f7ab58341da.png">

This PR does the following:
* introduces a new `node_view_libraries` property to the package header and upload package requests in `Greg.dll` (package manager client) - this is why the build is broken.
* Greg.dll references are temporarily updated to a local build with the new property.
* when the package loader encounters an assembly marked with this `node_view_library` entry point it records it, but does not try to import anything into the VM. ⚠️ The binary is still loaded because all binaries are loaded from non `built-in packages` until we modify this line: https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoPackages/Package.cs#L346 ( I believe this is currently true for backwards compatibility with legacy packages)  
* Later, when the DynamoView loads `NodeViewCustomizations` the `packageManger.PackageLoader.LocalPackages` is scanned for `NodeViewOnlyLibraries` and these assemblies are added to the `CustomizationLibrary` to be used when NodeViews are drawn.
* if an assembly is in both the `node_libraries` and `node_view_libraries` lists - the package will not be loaded.


### comparison with https://github.com/DynamoDS/Dynamo/pull/11882
|                                       | https://github.com/DynamoDS/Dynamo/pull/11882/                                                                                                                                                                                                                                                                         | This PR (node_view_library)                                                                                                                                                                                                                                                                                                                                                                           |
|---------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| package header support                | `NodeViewCustomizations` are loaded from binaries marked as `nodelib` -  pkg header already supports this list.                                                                                                                                                                                                        | currently in this PR - the author must manually edit the generated package header `pkg.json`  to include the `node_view_libraries` property, this is not yet populated at publish time.                                                                                                                                                                                                               |
| package publish support               | Because we reuse `nodelib` - package authors can mark their nodeview only libraries as `nodelib` in the Dynamo package publish UI. we *could* improve the experience by  adding some information to explain that `Node Library`s can contain NodeCustomizations&NodeModels OR ZeroTouch nodes. (Not both - see below)  | currently in this PR - there is no way to set an assembly during package publish to  `node view library`  - I assume it requires discussion with @Jingyi-Wen  if we decide to go with the approach in this PR. It might also be possible to inspect the types  and use this as an implementation detail only. IE NodeLibrary in UI -> node_view_library if the assembly only contains Customizations. |
| What can be marked as `node_library`? | There are 4 sets of valid assemblies that can be marked as `node_library` and will load "correctly" 1. ZeroTouch assemblies 2. NodeModel & NodeViewCustomizations 3. NodeModel 4. NodeViewCustomizations   BUT - note that if ZT types are present in case 2 -4 - they will not be loaded into the VM.                 | There are 3 sets of valid assemblies that can be marked as  `node_library` and will load "correctly" 1. ZeroTouch assemblies 2. NodeModel & NodeViewCustomizations 3. NodeModel   BUT - note that if ZT types are present in case 2 -3 - they will not be loaded into the VM.  also note that case 4. is now covered by `node_view_library`.                                                          |
| pros                                  | * simple and small - only changes in Dynamo.                                                                                                                                                                                                                                                                           | * `node_view_library` is simple to reason about - this assembly ONLY contains NodeViewCustomizations and nothing else in it gets imported into the VM.                                                                                                                                                                                                                                                |
| cons                                  | * `node_library` label is overloaded - can be used for 4 different cases and loading behavior depends on label and what combination of types are actually in the assembly.                                                                                                                                             | * changes span multiple repos and projects (package manager eventually, package manager client and Dynamo now) * theres no support for this property in package publishing - might require UI UX design. *                                                                                                                                                                                            |

### What this PR does not address:
* generating packages with `node_view_library` set correctly, this would need to be manually generated for the packages we produce - since this is for ATF/meshtoolkit thats probably fine at the moment.
* viewing `node_view_library` state in package publish view for installed packages - basically same issue as above.

### TODO 
- [ ] test loading package at startup applies customization
- [ ] test loading package late applies customization
- [ ] test loading node_view_library with other types does not import them into VM (including nodeModels)
- [ ] test package where nodeLib and nodeViewLib contains same assembly throws exception


### Declarations


aparajit-pratap:
| if an assembly is in both the node_libraries and node_view_libraries lists - the package will not be loaded.

If this is true in this PR, isn't this a breaking change? Currently, we have the ability to load both `nodemodel`s and `nodeviewcustomization`s if both are defined in the same assembly. In fact, the only way `nodemodel `package authors can have their custom views loaded right now is if they have their `nodeviewcustomization`s defined in the same assembly as `nodemodel`.

src/DynamoCore/Core/NodeModelAssemblyLoader.cs
```diff
@@ -7,6 +7,7 @@
 using Dynamo.Graph.Nodes;
 using Dynamo.Logging;
 using Dynamo.Migration;
+using Dynamo.Utilities;
```
aparajit-pratap, line 10: Can we revert this file?

src/DynamoCore/Models/DynamoModelEvents.cs
```diff
@@ -8,6 +8,7 @@
 using System.Collections.Generic;
 using Dynamo.Graph;
 using Dynamo.Extensions;
+using System.Reflection;
```
aparajit-pratap, line 11: I think we can revert this file too?

src/DynamoCore/Properties/AssemblyInfo.cs
```diff
@@ -39,4 +39,4 @@
 [assembly: InternalsVisibleTo("GenerativeDesign.Dynamo.ViewExtension")]
 [assembly: InternalsVisibleTo("DSCPython")]
 [assembly: InternalsVisibleTo("DynamoPythonTests")]
-[assembly: InternalsVisibleTo("GraphMetadataViewExtension")]
+[assembly: InternalsVisibleTo("GraphMetadataViewExtension")]
```
aparajit-pratap, line 42: ... and this one?

src/DynamoCoreWpf/Properties/AssemblyInfo.cs
```diff
@@ -43,4 +43,4 @@
 [assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")]
 [assembly: InternalsVisibleTo("PythonNodeModelsWpf")]
 [assembly: InternalsVisibleTo("LintingViewExtension")]
-[assembly: InternalsVisibleTo("GraphMetadataViewExtension")]
+[assembly: InternalsVisibleTo("GraphMetadataViewExtension")]
```
aparajit-pratap, line 46: revert?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -667,19 +667,37 @@ private void LoadNodeViewCustomizations()
...
         }
 
         private void SubscribeNodeViewCustomizationEvents()
         {
+            dynamoViewModel.Model.GetPackageManagerExtension().PackageLoader.PackgeLoaded += PackageLoader_PackgeLoaded;
             dynamoViewModel.Model.Loader.AssemblyLoaded += LoaderOnAssemblyLoaded;
             dynamoViewModel.NodeViewReady += ApplyNodeViewCustomization;
         }
 
+        private void PackageLoader_PackgeLoaded(Package loaddedPackage)
```
aparajit-pratap, line 688: typo: `loadedPackage`

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -667,19 +667,37 @@ private void LoadNodeViewCustomizations()
         {
             nodeViewCustomizationLibrary.Add(new CoreNodeViewCustomizations());
             foreach (var assem in dynamoViewModel.Model.Loader.LoadedAssemblies)
+            {
                 nodeViewCustomizationLibrary.Add(new AssemblyNodeViewCustomizations(assem));
+               
+            }
+            //add customizations from each NodeViewLibrary in all loaded packages.
+            foreach(var assem in dynamoViewModel.Model.GetPackageManagerExtension().PackageLoader.LocalPackages.SelectMany(x => x.NodeViewLibraries))
```
aparajit-pratap, line 675: How will the node view customization libraries get added to `LocalPackages `in the first place?

src/DynamoPackages/Package.cs
```diff
@@ -338,15 +353,32 @@ internal IEnumerable<PackageAssembly> EnumerateAssembliesInBinDirectory()
...
+            nodeViewLibraries = nodeViewLibraries == null ? new string[] { } : nodeViewLibraries;
+
+            if (nodeLibraries != null)
+            {
+                //if an assembly is somhehow listed as both a node_view_library and a node_library - then treat this as a malformed package.
+                foreach (var nvl in nodeViewLibraries)
+                {
+                    if (nodeLibraries.Contains(nvl))
+                    {
+                        throw new LibraryLoadFailedException(BinaryDirectory, $"package header specified {nvl} as both a node_library, and node_view_library");
```
aparajit-pratap, line 367: Why have this restriction?

mjkkirschner:
> | if an assembly is in both the node_libraries and node_view_libraries lists - the package will not be loaded.
> 
> I see, both can be loaded even if they belong to the same assembly if the assembly is tagged as a `node_library`, which is great! However, I would still like to understand why if the same assembly is marked as `node_library` and `node_view_library`, why limit from loading it?


So my thought process was as follows:
* `node_view_library` essentially means this assembly **only** contains `NodeViewCustomizations` - and `node_library` means the assembly contains `nodeModel` or zero touch implementations (but not both, see the chart above).
*  If somehow an assembly was listed in both lists, it seemed to not make sense - one attribute means only load the `NodeViewCustomizations` within the assembly, the other means only load the Implementations.

I guess one approach is to combine the fix from the previous solution, treat the assembly as a `node_library`, and then import both types from it.

src/DynamoPackages/Package.cs
mjkkirschner: please see my comment below https://github.com/DynamoDS/Dynamo/pull/11891#issuecomment-889391358

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
mjkkirschner: The binaries are loaded when the package is loaded, and are set as NodeViewLibraries here:
`IsNodeViewOnlyLibrary = IsNodeViewLibrary(nodeViewLibraries, assem.GetName(), warnings)`
https://github.com/DynamoDS/Dynamo/pull/11891/files#diff-eb9c2a35d407323409fa2fc260e277955b068bc335e05165dc6156bf6284f9a7R397

aparajit-pratap:
> > | if an assembly is in both the node_libraries and node_view_libraries lists - the package will not be loaded.
> 
> > 
> 
> > I see, both can be loaded even if they belong to the same assembly if the assembly is tagged as a `node_library`, which is great! However, I would still like to understand why if the same assembly is marked as `node_library` and `node_view_library`, why limit from loading it?
> 
> 
> 
> 
> 
> So my thought process was as follows:
> 
> * `node_view_library` essentially means this assembly **only** contains `NodeViewCustomizations` - and `node_library` means the assembly contains `nodeModel` or zero touch implementations (but not both, see the chart above).
> 
> *  If somehow an assembly was listed in both lists, it seemed to not make sense - one attribute means only load the `NodeViewCustomizations` within the assembly, the other means only load the Implementations.
> 
> 
> 
> I guess one approach is to combine the fix from the previous solution, treat the assembly as a `node_library`, and then import both types from it.
> 
> 
> 
> I'm not sure this is very convincing 😉 .
> 
> 

My thought process was that if I were the package author and had an assembly that defined both a nodemodel and a nodeviewcustomization, I might tag it as both a node_library as well as a node_view_library unless it was clear that the nodeviewlibrary tag was for assemblies that _only_ contained customizations while the nodelibrary tag was for both. Basically it could be confusing as the rules don't apply equally to both tags. Still a minor point as compared to the decision of which solution to go with, your first PR or this one.

PR #11896, opened 2021-07-29
Dyn-3640 - Recover execution after clearing cyclic dependency from graph

aparajit-pratap:
### Purpose

DYN-3640
This fixes a bug where if there are cycles in a graph, and the cycles are fixed (cleared) on a subsequent run, one or more nodes continue to show warnings and do not execute (they continue to return null). 

The solution implemented is to try to clear the `isCyclic` flag from the graph nodes involved in the delta execution and clear node warnings corresponding to those graph nodes before each delta execution of the graph. 

TODO:
- [x] Clear cyclic dependency warning from nodes after clearing cycle
- [x] Add unit tests

### Declarations


src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -1267,6 +1267,9 @@ internal void AddAndRegisterNode(NodeModel node, bool centered = false)
 
             HasUnsavedChanges = true;
 
+            if (node is CodeBlockNodeModel cbn
+                && string.IsNullOrEmpty(cbn.Code)) return;
```
aparajit-pratap, line 1271: Unrelated - this just prevents an unnecessary graph run when double-clicking on canvas to create an empty CBN.

src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -254,6 +256,36 @@ private void ApplyChangeSetForceExecute(ChangeSetData changeSet)
...
         }
 
+        private void ReActivateGraphNodesInCycle(List<AssociativeNode> nodeList)
+        {
+            if (nodeList == null || !nodeList.Any()) return;
+
+            var assocGraph = core.DSExecutable.instrStreamList[0].dependencyGraph;
+            var graphNodes = assocGraph.GetGraphNodesAtScope(Constants.kInvalidIndex, Constants.kInvalidIndex);
+
+            foreach (var node in nodeList)
```
mjkkirschner, line 266: will nodeList usually be quite small?

src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -254,6 +256,36 @@ private void ApplyChangeSetForceExecute(ChangeSetData changeSet)
...
+            {
+                var bNode = node as BinaryExpressionNode;
+
+                var identifier = bNode?.LeftNode as IdentifierNode;
+                if (identifier == null) continue;
+
+                GraphNode rootNode = null;
+                foreach(var gNode in graphNodes)
```
mjkkirschner, line 274: is this all graphNodes?

src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -254,6 +256,36 @@ private void ApplyChangeSetForceExecute(ChangeSetData changeSet)
...
+                {
+                    if(identifier.Value == gNode.updateNodeRefList[0].nodeList[0].symbol.name)
+                    {
+                        rootNode = gNode;
```
mjkkirschner, line 278: isn't it possible that a graph has two cycles that have been fixed? Won't this only clear the first?

test/DynamoCoreTests/Engine/EngineControllerTest.cs
```diff
@@ -82,6 +83,7 @@ public void EngineControllerReconcileTraceDataAndNotify()
             //Act
             var codeBlockNodeOne = CreateCodeBlockNode();
             CurrentDynamoModel.AddNodeToCurrentWorkspace(codeBlockNodeOne, true);
+            codeBlockNodeOne.UpdateValue(new UpdateValueParams("Code", "10;"));
```
mjkkirschner, line 86: I guess this was triggering a run before your changes in `WorkspaceModel.cs`?

src/Engine/ProtoCore/AssociativeGraph.cs
```diff
@@ -961,6 +961,36 @@ public void PushChildNode(GraphNode child)
             child.ParentNodes.Add(this);
         }
 
+        internal void ClearCycles(IEnumerable<GraphNode> graphNodes)
+        {
+            Stack<GraphNode> stack = new Stack<GraphNode>();
```
mjkkirschner, line 966: this looks like DFS - is there any benefit to this traversal algorithm rather than BFS or it should not matter what order traversal happens in as long as we iterate over all downstream nodes?

test/DynamoCoreTests/Engine/EngineControllerTest.cs
aparajit-pratap: Yes

src/Engine/ProtoCore/AssociativeGraph.cs
aparajit-pratap: In general each graphnode has just one child node so I guess it shouldn't benefit much if we use BFS plus as you said we need to touch each child node once including the ones that are the deepest. In most cases the dependency graph is actually just a chain of dependencies so the most optimal way would have been a simple linear search but since I need to account for multiple dependents in the most general case I chose DFS.

src/Engine/ProtoScript/Runners/LiveRunner.cs
aparajit-pratap: Thanks for pointing this out. I will check on this.

src/Engine/ProtoScript/Runners/LiveRunner.cs
aparajit-pratap: No, graphNodes are only those that are in global scope (class index and function index = -1 or invalid index).

aparajit-pratap: I can't think of a case where we can have a cycle in a single code block node scope or function definition scope as we disallow variable redefinitions in code block nodes 🤔

src/Engine/ProtoScript/Runners/LiveRunner.cs
aparajit-pratap: These are only those nodes that are modified in a single delta execution run. I guess it could be a long list if the user is in manual mode and has made a large number of edits to their graph before hitting `Run`.

aparajit-pratap: It might be worth profiling this change or monitoring the perf benchmark tests after this goes in to see if there is a significant slowdown as a result of this additional check on each delta execution.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
pinzart90: don;t know if this is more readable but it is a bit shorter
```
if (string.IsNullOrEmpty((node as CodeBlockNodeModel)?.Code))
```

src/Engine/ProtoCore/AssociativeGraph.cs
pinzart90: Could be less code if we make it recursive , but I remember you are not a fan :)

src/Engine/ProtoScript/Runners/LiveRunner.cs
sm6srw: 👍

aparajit-pratap:
I reran DynamoSelfServe and it passed here: https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/3044/

mjkkirschner (approved):

src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -254,6 +257,44 @@ private void ApplyChangeSetForceExecute(ChangeSetData changeSet)
...
+                var bNode = node as BinaryExpressionNode;
+
+                var identifier = bNode?.LeftNode as IdentifierNode;
+                if (identifier == null) continue;
+
+                var rootNodes = new List<GraphNode>();
+                foreach(var gNode in graphNodes)
+                {
+                    if(identifier.Value == gNode.updateNodeRefList[0].nodeList[0].symbol.name)
+                    {
```
pinzart90, line 278: are these always valid ?
`gNode.updateNodeRefList[0].nodeList[0]`

src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -254,6 +257,44 @@ private void ApplyChangeSetForceExecute(ChangeSetData changeSet)
...
+                        rootNodes.Add(gNode);
+                    }
+                }
+                if (!rootNodes.Any()) return;
```
pinzart90, line 282: kind of a redundant check

pinzart90 (approved):
2 small observations
LGTM

src/Engine/ProtoScript/Runners/LiveRunner.cs
aparajit-pratap: Yes

PR #11909, opened 2021-08-04
[DYN-3816] Overlay to be shown during guides

Astul-Betizagasti:
### Purpose

Adds an overlay that will become active during the new guides, restricting the area where the user can click.
With the current implementation the "hole" in the background is static and will not update to fit the corresponding controller for each step.
shadow

### Declarations

### Reviewers

Aaron Tang (@QilongTang )

### FYIs


src/DynamoCoreWpf/Views/GuidedTour/GuideBackground.xaml.cs
```diff
@@ -0,0 +1,68 @@
...
+{
+    /// <summary>
+    /// Interaction logic for GuideBackground.xaml
+    /// </summary>
+    public partial class GuideBackground : UserControl
+    {
+        public GuideBackground()
+        {
+            InitializeComponent();
+            DataContext = this;
```
RobertGlobant20, line 27: @Astul-Betizagasti  I'm not sure if is a good idea of having the Model and ViewModel inside the View (GuideBackground), thinking that in the future the HoleRect  will be dynamic (that could be a Rect or UIElement) and probably will be read (deserialized) from a json file. I guess we should create a GuideBackgroundViewModel class and define all the properties used inside that ViewModel, please let me know what you think. Thanks

src/DynamoCoreWpf/Views/GuidedTour/GuideBackground.xaml.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// Height of the background
+        /// </summary>
+        public double CanvasHeight
+        {
+            get { return (double)GetValue(CanvasHeightProperty); }
+            set { SetValue(CanvasHeightProperty, value); }
+        }
+
+        /// <summary>
+        /// Rect with the sieze of the Dynamo Window
```
RobertGlobant20, line 49: size misspelled?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -2335,8 +2336,10 @@ private void GetStartedMenuItem_Click(object sender, RoutedEventArgs e)
         /// </summary>
         private void ShowGetStartedGuidedTour()
         {
+            GuideBackground guideBackground = this.FindName("GuidesBackground") as GuideBackground;
```
RobertGlobant20, line 2339: The whole DynamoView is passed to the GuideManager then also you could find the "GuidesBackground" inside the GuidesManager class (using Guide.FindChild) without the need of modifying the GuidesManager constructor and passing it as parameter.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml
```diff
@@ -1828,6 +1829,13 @@
         </Grid>
 
         <Grid Name="FocusableGrid" Width="0" Height="0" Focusable="True"/>
+        <guides:GuideBackground x:Name ="GuidesBackground"
```
QilongTang, line 1832: Would you add comments here?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
QilongTang: @RobertGlobant20 Indeed, worth a try

src/DynamoCoreWpf/Views/GuidedTour/GuideBackground.xaml.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Threading.Tasks;
+using System.Windows;
+using System.Windows.Controls;
+using System.Windows.Data;
+using System.Windows.Documents;
+using System.Windows.Input;
+using System.Windows.Media;
+using System.Windows.Media.Imaging;
+using System.Windows.Navigation;
+using System.Windows.Shapes;
```
QilongTang, line 15: Are all of these valid usings?

QilongTang (commented):
Some comments, then LGTM

src/DynamoCoreWpf/Views/GuidedTour/GuideBackground.xaml.cs
Astul-Betizagasti: Correct, will update this on next commit

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
Astul-Betizagasti: That's right, will update on next commit

src/DynamoCoreWpf/Views/Core/DynamoView.xaml
Astul-Betizagasti: Will do

src/DynamoCoreWpf/Views/GuidedTour/GuideBackground.xaml.cs
Astul-Betizagasti: Will check and update if necessary, thanks

src/DynamoCoreWpf/Views/GuidedTour/GuideBackground.xaml.cs
Astul-Betizagasti: It was my understanding that the common practice for UserControls is not to have a viewmodel (I could be wrong here). Although I do agree that it could be a viable option. @QilongTang do you have any preferences?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
Astul-Betizagasti: I am having a bit of trouble with this, the DynamoView is actually passed as a UIElement so to get the overlay element I have to first cast it to DynamoView, which is a bit sketchy to do here in my opinion. Also I couldn't find a FindChild method available here.
image
Another solution I can think of would be to save the guideBackground on the DynamoViewModel beforehand to have it accessible here.

@QilongTang, @RobertGlobant20 what are your thoughts on this?

QilongTang: hmm, the `guideBackground` is a UI control right? In that way, I wont save it to the view model, maybe we can save it under DynamoView so you dont have to always search for it runtime?

QilongTang (approved):

PR #11911, opened 2021-08-04
Local Definitions in workspace references

reddyashish:
### Purpose

Task: https://jira.autodesk.com/browse/DYN-3718

This PR is to add the Local Definitions section in the Workspace References view extension. 

The local definitions data is serialized into the DYN file and is displayed in the workspace references view extension under Local Definitions tab item. This tab will display the references pointing to local custom nodes and local zero touch nodes. 

Local definitions

The following example shows how the local definitions data is serialized.

Screen Shot 2021-08-26 at 6 33 22 AM

### Declarations

### Reviewers

@QilongTang @zeusongit @Amoursol @Jingyi-Wen @DynamoDS/dynamo


reddyashish:
Committed new changes related to serialization and the information thats needs to be shown in the extension.

@Jingyi-Wen @Amoursol Do we have the relevant icons needed for the new section? Also regarding the size, do we want to display the file size? I am not sure how important that would be.

Amoursol:
> Committed new changes related to serialization and the information thats needs to be displayed in the extension.
> 
> @Jingyi-Wen @Amoursol Do we have the relevant icons needed for the new section? Also regarding the size, do we want to display the file size? I am not sure how important that would be.

@Jingyi-Wen are you able to please provide icons for this? 

@reddyashish yes we do want to show file size. This gives a user a better understanding of how much content needs to ship with Dynamo. Some firms are very limited in what they can send, or the mechanisms in which they can send things so an indication matters here.

reddyashish:
Does it matter if we want to serialize the size as well or do we want to get that at runtime?

QilongTang:
Hi @Amoursol @Jingyi-Wen @reddyashish I prefer the size of these new references read and display runtime, and we skip the serialization of them. That indicates that when they are missing on graph consumer's computer, the size would be displayed as `N/A`. This not only made our dev process simpler but also make sure we do not serialize info that we dont need to, file size is already part of file info on disk :) What do you think?

reddyashish:
Sounds good to me @QilongTang. Will make those changes.

Amoursol:
> Hi @Amoursol @Jingyi-Wen @reddyashish I prefer the size of these new references read and display runtime, and we skip the serialization of them. That indicates that when they are missing on graph consumer's computer, the size would be displayed as `N/A`. This not only made our dev process simpler but also make sure we do not serialize info that we dont need to, file size is already part of file info on disk :) What do you think?

Agreed 💯  - No need to serialize.

mjkkirschner:
I'm a little curious about the choice to have a new top level property in the .dyn `NodeLocalDefinitions` - why not reuse the existing one `NodeLibraryDependencies` and modify the referenceType - is it because `NodeLibraryDependencies` is too specific as a name?

reddyashish:
@mjkkirschner Yes I thought the naming might cause some confusion, so I had added a new section. The structure of the objects are also different, so I am using 2 different fields: NodeLocalDefinitions and NodeLibraryDependencies, in WorkspaceModel. Do you think it is better to add this data to the same NodeLibraryDependencies property and get rid of NodeLocalDefinitions property in DYN?

mjkkirschner:
@reddyashish - what properties did the other objects need - it seems they actually have less information, not more? I'm also curious if these serialized objects will also support other references or we'll need another object type to support references like arbitrary files?

Jingyi-Wen:
Here're the icons
dfy_icon
dll_icon

reddyashish:
> @reddyashish - what properties did the other objects need - it seems they actually have less information, not more? I'm also curious if these serialized objects will also support other references or we'll need another object type to support references like arbitrary files?
@mjkkirschner When you look at the serialized data, the local definitions information is less. Are you referring to only the serialized data in DYN?

mjkkirschner:
@reddyashish - I was thinking version could be left blank - yes I was referring to serialized data schema, not in memory objects - they don't need to be the same.

My other question was about other reference types, like arbitrary files.

reddyashish:
Yes got it. We can add it the local definitions data to node library dependencies schema. I will make those changes. 

I am also working now on the arbitrary file references which will be in a new PR. I think the same serialized object can be used for those too.

src/WorkspaceDependencyViewExtension/Properties/Resources.en-US.resx
```diff
@@ -117,6 +117,10 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <assembly alias="System.Windows.Forms" name="System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" />
```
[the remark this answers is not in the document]
reddyashish: I think this was added when I was trying to add the image resources. Removed it.

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
```diff
@@ -1056,25 +1066,39 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
...
+                case ReferenceType.DYFFILE:
+                    depInfo = new LocalDefinitionInfo(name, ReferenceType.DYFFILE);
+
+                    break;
+                case ReferenceType.ZeroTouch:
+                    depInfo = new LocalDefinitionInfo(name, ReferenceType.ZeroTouch);
+                    break;
                 default:
-                    depInfo = new PackageDependencyInfo(name, version);
+                    depInfo = new LocalDefinitionInfo(name);
```
QilongTang, line 1098: Can you specify the reason for this change?

src/WorkspaceDependencyViewExtension/Properties/Resources.en-US.resx
```diff
@@ -117,6 +117,10 @@
...
+  <data name="CustomNodeReferenceIcon" type="System.Resources.ResXFileRef, System.Windows.Forms">
+    <value>..\Resources\CustomNodeReferenceIcon.png;System.Drawing.Bitmap, System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a</value>
```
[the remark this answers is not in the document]
QilongTang: Can you point us where that dynamic code is?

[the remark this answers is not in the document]
reddyashish: Here: based on this field, the icon will be set. https://github.com/DynamoDS/Dynamo/pull/11911/files#diff-6689e4eac4bfa3a7075e6613dc64a2096bb6b1126184bb0d38a0480cb5c6cb80R485

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -126,6 +134,37 @@ internal void DependencyRegen(WorkspaceModel ws)
...
+                        if (customNodeInfo != null)
+                        {
+                            info.Path = customNodeInfo.Path;
+                        }
+                    }
+
+                    if (info.Path != null)
+                    {
+                        FileInfo localDefinitionFileInfo = new FileInfo(info.Path);
+                        long size = localDefinitionFileInfo.Length / 1024;
```
QilongTang, line 158: should we make 1024 constant? And would this always be in kb unit?

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
```diff
@@ -101,6 +101,131 @@ public override string ToString()
...
+        public string Size { get; internal set; }
+
+        /// <summary>
+        /// ReferenceType of the LocalDefinition
+        /// </summary>
+        public ReferenceType ReferenceType { get; internal set; }
+
+        private HashSet<Guid> nodes;
+
+        public HashSet<Guid> Nodes
```
QilongTang, line 131: I dont see this field got leveraged? Did I miss anything?

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
reddyashish: We used to have only PackageDependencyInfo before so it was in the default field. Now we have added cases for DYFFILE and ZeroTouch along with Package reference types. So I was defaulting to LocalDefinitionInfo instead. We can remove this once all cases are addressed.

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
reddyashish: It gets called in the AddDependent method where the node is added to the list.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
reddyashish: Yes because the Length always returns in bytes, the result will be in kb.

src/DynamoCore/Graph/Workspaces/SerializationConverters.cs
QilongTang: Thanks!

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
QilongTang: Thanks!

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
```diff
@@ -101,6 +101,131 @@ public override string ToString()
...
+        {
+            get { return nodes; }
+        }
+
+        public Version Version { get; internal set; }
+
+        /// <summary>
+        /// Indicates whether this LocalDefinition is loaded in the current session
+        /// </summary>
+        [Obsolete("This property is obsolete, use PackageDependencyState property instead", false)]
```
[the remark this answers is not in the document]
reddyashish: PackageDependencyState is a public Enum so it would cause a API break if we change the field name. Do you want me to add this as a TODO for 2.13?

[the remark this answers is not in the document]
QilongTang: So it seems to me we have a tech debt to update the interface structure here

src/WorkspaceDependencyViewExtension/Properties/Resources.en-US.resx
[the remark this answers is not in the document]
QilongTang: @reddyashish Understood, I guess this way we could also embed into the binary as well? Like our node icons?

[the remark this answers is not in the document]
QilongTang: image

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -685,22 +686,91 @@ internal List<INodeLibraryDependencyInfo> NodeLibraryDependencies
...
                         }
                     }
-                   
                 }
 
                 RaisePropertyChanged(nameof(NodeLibraryDependencies));
             }
         }
 
+        internal List<INodeLibraryDependencyInfo> NodeLocalDefinitions
```
QilongTang, line 703: Would this be updated when there are new dyf filed imported into current workspace?

reddyashish: When the dyf is placed and the user saves the workspace, this will be updated.

src/WorkspaceDependencyViewExtension/Properties/Resources.en-US.resx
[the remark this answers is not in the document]
reddyashish: Changed the Icon properties to "Embedded in .resx". Not storing the image path in resx file anymore.

src/WorkspaceDependencyViewExtension/Properties/Resources.en-US.resx
```diff
@@ -383,7 +390,13 @@
...
   </data>
-  <data name="VersionHeaderText" xml:space="preserve">
+  <data name="SizeHeader" xml:space="preserve">
+    <value>Size</value>
+  </data>
+  <data name="VersionHeader" xml:space="preserve">
     <value>Version</value>
   </data>
+  <data name="ZeroTouchNodeReferenceIcon" type="System.Resources.ResXFileRef, System.Windows.Forms">
+    <value>..\Resources\ZeroTouchNodeReferenceIcon.png;System.Drawing.Bitmap, System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a</value>
```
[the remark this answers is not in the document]
reddyashish: Addressed this.

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
```diff
@@ -101,6 +101,131 @@ public override string ToString()
         }
     }
 
+    /// <summary>
+    /// Class containing info about a LocalDefinition
+    /// </summary>
+    internal class LocalDefinitionInfo: INodeLibraryDependencyInfo
```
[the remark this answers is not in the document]
QilongTang: After discussion with @mjkkirschner , we think this implementation is good to proceed

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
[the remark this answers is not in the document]
reddyashish: I will create Tech Debt tasks once this is merged.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -126,6 +135,37 @@ internal void DependencyRegen(WorkspaceModel ws)
         {
             RestartBanner.Visibility = Visibility.Hidden;
             var packageDependencies = ws.NodeLibraryDependencies.Where(d => d is PackageDependencyInfo).ToList();
+            var localDefinitions = ws.NodeLocalDefinitions.Where(d => d is LocalDefinitionInfo).ToList();
+
+            foreach (LocalDefinitionInfo info in localDefinitions)
+            {
+                try
+                {
+                    if (info.ReferenceType == ReferenceType.DYFFILE)
```
QilongTang, line 144: I understand this is defined previously.. Comparing to  `DSFile`, can we rename this to `DyfFile` or `DYFFile` to be consistent?

QilongTang (approved):
With one last comment, then LGTM

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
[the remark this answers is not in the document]
QilongTang: Thanks!

QilongTang:
Thank you @reddyashish This is looking good! One more step forward. Would you also update the gif when finalizing the PR?

QilongTang:
> Can we expand this work to include DS file references?

Sorry @aparajit-pratap I am a bit hesitate to expand the original scope given so much going on recently. Would you log a task with an example for our team to schedule in the future? Appreciate it

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
reddyashish: Updated it.

reddyashish:
Updated the gif in here.

src/DynamoCore/Graph/Workspaces/PackageDependencyInfo.cs
```diff
@@ -101,14 +101,139 @@ public override string ToString()
...
+            if (obj == null)
+            {
+                return false;
+            }
+            if (!(obj is LocalDefinitionInfo))
+            {
+                return false;
+            }
+
+            var other = obj as LocalDefinitionInfo;
```
aparajit-pratap, line 197:
```
var other = obj as LocalDefinitionInfo;
if(other == null) return false;

if (other.Name == this.Name && other.Path == this.Path) 
{ ... }
```

aparajit-pratap: How about also checking for the reference type to be equal?

reddyashish: I will be creating another PR for external references and will make this change in that PR.

PR #11912, opened 2021-08-04
Group improvements to the graph view

SHKnudsen:
### Purpose
This PR does:

- Adds a header section to groups which contains a Name textbox and Description textbox, it also adds buttons to collapse the group and open the group's context menu
- Functionality to collapse groups. Collapsing groups will hide all wires, nodes, notes and groups belonging to the group. A new property has been added to the ViewModelBase `IsCollapsed` which can be used to control the visibility of controls.
- When a group is collapsed it will build input and output ports on the group. All inputs that are unconnected or connected to something outside the group will show up on the group, same for output ports.
- Groups can now be added to other groups (only 1 nested layer).
- Drag and drop nodes into groups
- Groups can manually be resized by dragging the bottom right corner.

#### Drag and drop nodes to groups
Groups_dragndrop

#### Collapse groups
Groups_collapse

#### Resize groups
resize_new

#### Grouped groups
grouped_groups

### Known limitations
Currently, the input/output ports that are added to the group are simply a view placeholder, connectors wil still snap to the original portmodel location

### Declarations

### Reviewers

@QilongTang 

### FYIs

@Amoursol 
@mjkkirschner


QilongTang:
I may also rebase this PR, need to confirm with team about release plan

src/DynamoCore/Graph/Annotations/AnnotationModel.cs
```diff
@@ -66,14 +66,17 @@ public override double Width
...
                 width = value;
                 RaisePropertyChanged("Width");
             }
         }
 
         private double height;
         /// <summary>
-        /// Returns height of the group
+        /// Returns the full height of the group.
+        /// That means ModelAreaHeight + TextBlockHeight
```
QilongTang, line 79: 👍🏻

src/DynamoCore/Graph/Annotations/AnnotationModel.cs
```diff
@@ -141,15 +176,28 @@ public IEnumerable<ModelBase> Nodes
...
-                nodes = value.ToList(); ;
+                // Before setting this property
+                // we need to reset all previous values.
+                if (nodes != null)
+                {
+                    foreach (var node in nodes)
+                    {
+                        node.BelongsToGroup = false;
+                    }
+                }
```
QilongTang, line 187: For this kind of loop with only one action for each, we usually leverage `ForEach()`
```
nodes = value.ToList().ForEach(x => x.BelongsToGroup = false);
```

src/DynamoCore/Graph/Annotations/AnnotationModel.cs
```diff
@@ -243,33 +291,94 @@ public NodeModel PinnedNode
             }
         }
 
+        private double widthAdjustment;
+        /// <summary>
+        /// Adjustment set to the size of the groups
```
QilongTang, line 296: do you mean `group's`?

src/DynamoCore/Graph/Annotations/AnnotationModel.cs
```diff
@@ -243,33 +291,94 @@ public NodeModel PinnedNode
...
+            {
+                if (value == widthAdjustment) return;
+                widthAdjustment = value;
+                UpdateBoundaryFromSelection();
+            }
+        }
+
+        private double userHeightAdjustment;
+        /// <summary>
+        /// Adjustment set to the size of the groups
```
QilongTang, line 312: same comment here

src/DynamoCore/Graph/Annotations/AnnotationModel.cs
```diff
@@ -243,33 +291,94 @@ public NodeModel PinnedNode
...
         /// Initializes a new instance of the <see cref="AnnotationModel"/> class.
         /// </summary>
         /// <param name="nodes">The nodes.</param>
         /// <param name="notes">The notes.</param>
-        public AnnotationModel(IEnumerable<NodeModel> nodes, IEnumerable<NoteModel> notes)
-        {                                 
-            var nodeModels = nodes as NodeModel[] ?? nodes.ToArray();           
+        public AnnotationModel(IEnumerable<NodeModel> nodes, IEnumerable<NoteModel> notes) : this(nodes, notes, new List<AnnotationModel>()) { }
+
+        public AnnotationModel(IEnumerable<NodeModel> nodes, IEnumerable<NoteModel> notes, IEnumerable<AnnotationModel> groups)
```
QilongTang, line 354: Would you add comments for this new constructor? Usually for constructors, it's must-have

src/DynamoCore/Graph/ModelBase.cs
```diff
@@ -211,6 +212,21 @@ public bool IsSelected
             }
         }
 
+        /// <summary>
+        /// If this model belongs to an AnnotationModel
+        /// </summary>
+        [JsonIgnore]
+        public bool BelongsToGroup
```
QilongTang, line 219: @mjkkirschner @aparajit-pratap What do you feel about adding this property to `ModelBase`?

src/DynamoCore/Graph/Nodes/PortModel.cs
```diff
@@ -505,6 +505,31 @@ internal string GetOutPortType()
...
 
+    /// <summary>
+    /// This currently doesn't do anything other than copy the portmodel it receives.
+    /// Keeping it here for now as we might add to it when a decision about how to 
+    /// move forward has been made.
+    /// </summary>
+    internal class ProxyPortModel : PortModel
+    {
+        //private int index;
+        //private Point2D center;
```
QilongTang, line 516: hmm do we need to clean up the commented code?

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -1446,6 +1455,28 @@ internal AnnotationModel AddAnnotation(string text, Guid id)
...
+            if (!CheckIfModelExistsInSomeGroup(selectedNodes, selectedNotes))
+            {
+                var annotationModel = new AnnotationModel(selectedNodes, selectedNotes)
+                {
+                    GUID = id,
+                    AnnotationDescriptionText = titleText,
+                    AnnotationText = text
+                };
+                annotationModel.ModelBaseRequested += annotationModel_GetModelBase;
+                annotationModel.Disposed += (_) => annotationModel.ModelBaseRequested -= annotationModel_GetModelBase;
```
QilongTang, line 1472: Do we need to unsubscribe these somewhere?

src/DynamoCore/Graph/ModelBase.cs
mjkkirschner: It seems a bit odd - groups already have a list of models that they contain, and not ALL models can belong to a group - ie connectors.

SHKnudsen: Fair enough, the reason i felt this was needed was that I have to do this check a lot for the drag and drop functionality. Essentially i need to check if the dragged model belongs to a group every time its dragged over a group, it seems very inefficient to have to go over all Groups and check its nodes collection everytime this happens.

I get that there are some Models that does not have any functionality that enables it to be added to a group, but at the same time the `Nodes` collection is of type `IEnumerable<ModelBase>` which does indicate that in theory, it can take all `ModelBase`s.

Happy to revert this and simply do the `AnnotaitonModel.Nodes.Contains(modelBase)` check every time, my worry was just on performance as this has to be done a lot.

https://github.com/DynamoDS/Dynamo/pull/11912/files#diff-32e4994e829fb0abe599845f0677a235d127cf860b8a85eb31667dfef50db6c0R822

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
SHKnudsen: Should already be unsubscribed on Disposed?
`annotationModel.Disposed += (_) => annotationModel.ModelBaseRequested -= annotationModel_GetModelBase;`

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -112,6 +153,192 @@ public PreviewState PreviewState
...
+        public int InbetweenNodesCount
+        {
+            get => Nodes
+                .Except(InputNodes
+                    .Select(x => x.NodeModel)
+                    .Union(OutputNodes.Select(x => x.NodeModel)))
+                .Count();
+        }
+
+        public System.Windows.Rect ModelAreaRect
```
QilongTang, line 318: Would you add comments?

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -112,6 +153,192 @@ public PreviewState PreviewState
...
+
+
+        private GeometryCollection groupedGroupsGeometry;
+        /// <summary>
+        /// Collection of rectangles based on AnnotationModels
+        /// that belongs to this group.
+        /// This is used to make a cutout in this groups background
+        /// where another group is placed so there wont be an overlay.
+        /// </summary>
+        public GeometryCollection GroupedGroupsGeometry
```
QilongTang, line 334: Is there a better name for this? A bit hard at first :)

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -160,30 +422,87 @@ private void AddToGroup(object obj)
...
+            if (annotationModel.IsSelected)
             {
-                annotationModel.FontSize = value;                
+                var selectedModels = DynamoSelection.Instance.Selection
+                    .OfType<AnnotationModel>()
+                    .Where(x => x.GUID != this.AnnotationModel.GUID && !x.BelongsToGroup);
+
+                foreach (var model in selectedModels)
+                {
+                    WorkspaceViewModel.DynamoViewModel.AddGroupToGroupModelCommand.Execute(this.AnnotationModel.GUID);
```
QilongTang, line 463: I saw you have a constructor to take a collection, I guess not using here because we have follow-ups below to do in addition?

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -193,7 +512,149 @@ public AnnotationViewModel(WorkspaceViewModel workspaceViewModel, AnnotationMode
...
+                .ToList()
+                .ForEach(x => AddToCutGeometryDictionary(x));
+        }
+
+        /// <summary>
+        /// Creates input ports for the group based on its Nodes.
+        /// Input ports that either is connected to a Node outside of the
+        /// group, or has a port that is not connected will be used for the group.
+        /// </summary>
+        internal void SetGroupInputPorts()
```
QilongTang, line 538: Would you add some comments in this function body? I think it is quite readable but I want to make sure other team member know what is the process

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -193,7 +512,149 @@ public AnnotationViewModel(WorkspaceViewModel workspaceViewModel, AnnotationMode
...
+            var originalPortViewModels = WorkspaceViewModel.Nodes
+                .SelectMany(x => x.InPorts.Concat(x.OutPorts))
+                .Where(x => groupPortModels.Contains(x.PortModel))
+                .ToList();
+
+            var newPortViewModels = new List<PortViewModel>();
+            for (int i = 0; i < proxyModels.Count(); i++)
+            {
+                var proxyModel = proxyModels[i];
+                //proxyModel.SetCenterPoint(new Point2D(0, 0));
```
QilongTang, line 653: Is this still needed? I think from our discussion the position or the proxy does not work yet since the center point for port model is inaccurate?

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -203,9 +664,95 @@ internal void ClearSelection()
...
-        private bool CanChangeFontSize(object obj)
+        /// <summary>
+        /// Collapse all nodes in this group
+        /// by settings its IsCollapsed property
+        /// to true.
+        /// Only the Parent group should handle collapsing
+        /// the connector that needs it.
+        /// </summary>
+        /// <param name="collapseConnectors"></param>
+        private void CollapseGroupNodes(bool collapseConnectors)
```
QilongTang, line 675: Considering we are collapsing all content under group, can we rename this to `CollapseGroupContent` and also the function below from `ShowGroupNodes` to `ShowGroupContent`?

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -219,6 +766,47 @@ private void UpdateFontSize(object parameter)
             WorkspaceViewModel.DynamoViewModel.RaiseCanExecuteUndoRedo();
         }
 
+        /// <summary>
+        /// Selects this group and models within it.
+        /// </summary>
+        internal void Select()
```
QilongTang, line 772: Can we name this `SelectAll`?

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -193,7 +512,149 @@ public AnnotationViewModel(WorkspaceViewModel workspaceViewModel, AnnotationMode
...
             }
-            
+
+            InPorts = new ObservableCollection<PortViewModel>();
+            OutPorts = new ObservableCollection<PortViewModel>();
+            InputNodes = new ObservableCollection<NodeViewModel>();
+            OutputNodes = new ObservableCollection<NodeViewModel>();
+
+            InPorts.CollectionChanged += InPorts_CollectionChanged;
+            OutPorts.CollectionChanged += OutPorts_CollectionChanged;
```
QilongTang, line 522: Do we need to unsubscribe somewhere?

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -160,30 +422,87 @@ private void AddToGroup(object obj)
...
-       
+        #endregion
+
         public AnnotationViewModel(WorkspaceViewModel workspaceViewModel, AnnotationModel model)
-        {             
+        {
             annotationModel = model;           
             this.WorkspaceViewModel = workspaceViewModel;
             model.PropertyChanged += model_PropertyChanged;
+            DynamoSelection.Instance.Selection.CollectionChanged += SelectionOnCollectionChanged;
```
QilongTang, line 504: Do we need to unsubscribe somewhere?

src/DynamoCoreWpf/ViewModels/Core/Converters/SerializationConverters.cs
```diff
@@ -107,13 +109,28 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
...
+            writer.WritePropertyName(nameof(ExtraAnnotationViewInfo.HeightAdjustment));
+            writer.WriteValue(anno.HeightAdjustment);
             writer.WritePropertyName("Nodes");
             writer.WriteStartArray();
             foreach (var m in anno.Nodes)
             {
                 writer.WriteValue(m.GUID.ToString("N"));
             }
             writer.WriteEndArray();
+            writer.WritePropertyName("Groups");
```
QilongTang, line 127: @mjkkirschner @aparajit-pratap @pinzart @sm6srw What are your thoughts on this? I feel we are no longer strict on CoGS format so this could be fine? Or we can maybe add the nesting info to view layer but that may cause deserialization problem? I am not sure.

QilongTang (commented):
Some comments. I have 15 more files to finish tomorrow

src/DynamoCore/Graph/ModelBase.cs
mjkkirschner: @SHKnudsen if you are worried about the performance one option is replace the implementation in AnnotationModel with a hashset of nodes. Should there ever be duplicates in the group?

src/DynamoCoreWpf/ViewModels/Core/Converters/SerializationConverters.cs
mjkkirschner: I think this is fine, it makes sense to add this serialization in the annotation, but aren't we going to end up serializing these ids twice ⚠️ ?
Once in the nodes property and once in the groups property? Maybe we only need the NodesProperty and it really means "models"?

src/DynamoCore/Graph/ModelBase.cs
SHKnudsen: No there won't ever be duplicates, my worry was more that I need to go through every group in the graph here to check if the dragged node belongs to any of them, so if there was a large number of groups in the graph it seemed inefficient to go through them all every time i need to do this check.

the workflow that has to happen when a node is repositioned: 
- check if its over a group (we have to use the mouse cursor position here)
    - because we are dragging another element we cant use `MouseOver` event on the group
- then check if the dragged node already belongs to a group (this is where it was convenient to simply do `draggedNode.BelongsToGroup`)
 - if we are hovering over a group and the node does not already belong to a group then set the group in a `NodeHovering state` so the group ui gets a grey border around it indicating that we can drop the node in it

src/DynamoCoreWpf/DynamoCoreWpf.csproj
```diff
@@ -1338,6 +1338,10 @@
     <Resource Include="UI\Images\profile_normal.png" />
   </ItemGroup>
   <ItemGroup>
+    <Resource Include="UI\Images\menu_grey_48px.png" />
+    <Resource Include="UI\Images\caret_down_grey_48px.png" />
+    <Resource Include="UI\Images\caret_up_grey_48px.png" />
+    <Resource Include="UI\Images\NodeIcon_placeholder.png" />
```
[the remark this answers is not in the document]
SHKnudsen: this a temporary placeholder for the node icons. The node refresh work has added functionality to get the icon from the nodeViewModel which we can use for this, so this was just added until i can access that
image

src/DynamoCore/Graph/ModelBase.cs
mjkkirschner: okay thanks for the detailed description @SHKnudsen - 

1. definitely sounds like it could get slow at the moment you need to check what group you are over if you need to iterate every group, and every node in that group using a list of nodes (how it's implemented today). My suggestion was to modify the `AnnotationModel.Nodes` internal datastructure to be a hashSet or dictionary so that the lookup time is constant for each group, then you just need to iterate over all groups - a much smaller collection.

2. Checking the mouse position is fraught - it will need to be tested with different DPI scaling settings applied.
3. If you feel your solution is better (personally I don't like trying to keep state like that correct) - then what about an interface that NodeModel and AnnotationModel can implement. `ICanBelongToGroup` ?

pinzart90: I agree with Mike that iterating over existing groups only when MouseUp is triggered should not cause that much of a performance penalty. 
One other solution could be a new static (or Singleton ) class that can store a map of NodeId to GroupId, This data can be populated when Dynamo starts up and when any group is altered.

src/DynamoCoreWpf/DynamoCoreWpf.csproj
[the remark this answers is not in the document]
QilongTang: Thanks @SHKnudsen

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
QilongTang: Yup, my bad missed this

test/DynamoCoreWpfTests/SerializationTests.cs
```diff
@@ -464,8 +465,13 @@ public void JSONisSameBeforeAndAfterSaveWithDummyNodes()
             // Ignoring the ExtensionWorkspaceData property as this is added after the re-save,
             // this will cause a difference between jobject1 and jobject2 if it is not ignored.
             // Same thing goes for the Linting property...
+            // We also need to ignore the new IsCollapsed property on ViewModelBase
```
QilongTang, line 468: hmm can we add JsonIgnore on `ViewModelBase`?

src/DynamoCoreWpf/Views/Core/AnnotationView.xaml
```diff
@@ -101,76 +112,326 @@
                     </ControlTemplate>
                 </Setter.Value>
             </Setter>
-        </Style>    
+        </Style>
+
+        <ControlTemplate x:Key="GroupExpanderTemplate"
```
QilongTang, line 117: Can you add some comments to major blocks in xaml definition?

src/DynamoCoreWpf/Views/Core/AnnotationView.xaml.cs
```diff
@@ -32,7 +34,25 @@ public AnnotationView()
...
             InitializeComponent();
             Loaded += AnnotationView_Loaded;
+            this.DataContextChanged += AnnotationView_DataContextChanged;
             this.GroupTextBlock.SizeChanged += GroupTextBlock_SizeChanged;
+
+            // Because the size of the CollapsedAnnotationRectangle doesn't necessarily change 
+            // when going from Visible to collapse (and other way around), we need to also listen
+            // to IsVisibleChanged. Both of these handlers will set the ModelAreaHeight on the ViewModel
+            this.CollapsedAnnotationRectangle.SizeChanged += CollapsedAnnotationRectangle_SizeChanged;
+            this.CollapsedAnnotationRectangle.IsVisibleChanged += CollapsedAnnotationRectangle_IsVisibleChanged;
```
QilongTang, line 44: Do we need to unsubscribe these somewhere?

QilongTang (commented):
Some comments then LGTM

src/DynamoCoreWpf/ViewModels/Core/Converters/SerializationConverters.cs
QilongTang: Yeah we already have a `Annotation` block in view, originally I was thinking if we can just leverage that block. Here is an example:
```
 "Annotations": [
      {
        "Id": "9150975dc1b045ab9214c57533c519f3",
        "Title": "Test Group",
        "Nodes": [
          "5b151b86d7c04a2a9874fda51ae526c3",
          "6434e70dd8004aca9c26cbcf1a3dca64",
          "d58a319a377d40a3b80ffefbaaaff267",
          "621a399415ba495d9924ee39138bfb74"
        ],
        "Left": 181.19999999999987,
        "Top": 48.40000000000002,
        "Width": 578.40000000000009,
        "Height": 316.2,
        "FontSize": 36.0,
        "InitialTop": 101.60000000000002,
        "InitialHeight": 293.0,
        "TextblockHeight": 43.2,
        "Background": "#FFC1D676"
      }
    ]
```

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
SHKnudsen: yea the makes more sense. The only reason i stuck with Nodes was that the "content" property is called `Nodes` which might also be a bad name as it can contain both `NoteModels`, `NodeModels` and now `AnnotationModels`.

Will use content for these 👍

src/DynamoCoreWpf/ViewModels/Core/Converters/SerializationConverters.cs
SHKnudsen: yea good points.. i originally added the `Groups` prop because i was running into issues when deserializing annotationModels that had nested groups. The issue being that we need to create any AnnotationModels that belong to other groups first, otherwise we cannot add them to the group we are creating.

I changed this now so that we serialize the `HasNestedGroups` property, and when deserializing i use a queue to build all nested groups first
```csharp
private void LoadAnnotations(IEnumerable<ExtraAnnotationViewInfo> annotationViews)
{
    if (annotationViews == null) return;

    var annotationQueue = new Queue<ExtraAnnotationViewInfo>(annotationViews);
    while (annotationQueue.Any())
    {
        var annotationViewInfo = annotationQueue.Dequeue();
        // Before creating this group we need to create
        // any group belonging to this group.
        if (annotationViewInfo.HasNestedGroups &&
            !annotationQueue.All(x=>x.HasNestedGroups))
        {
            annotationQueue.Enqueue(annotationViewInfo);
            continue;
        }

        LoadAnnotation(annotationViewInfo);
    }
}
```
https://github.com/DynamoDS/Dynamo/pull/11912/files#diff-fa18da253bafe5cc21047083856d6bc7b2d2b7f49d1441c589b274d405b947edR2172

SHKnudsen:
```json
{
  "Id": "7fbc02d54b664e38b37a639e6c9d9663",
  "Title": "GroupWithGroupedGroup",
  "DescriptionText": "<Click here to edit the group description>",
  "IsExpanded": true,
  "WidthAdjustment": 0.0,
  "HeightAdjustment": 0.0,
  "Nodes": [
    "1ffbaf75bf12424b8f6f55fdcea208dc",
    "3a8039e80d984ae19f1a4dfd08a49289",
    "1997f3c83f764260a3f2f10cc73f802d"
  ],
  "HasNestedGroups": true,
  "Left": 185.70075726305083,
  "Top": 1104.2283106882585,
  "Width": 623.505028610443,
  "Height": 392.58280439753753,
  "FontSize": 36.0,
  "InitialTop": 1177.5616440215917,
  "InitialHeight": 387.46579475350063,
  "TextblockHeight": 63.333333333333336,
  "Background": "#FFC1D676"
}

test/DynamoCoreWpfTests/SerializationTests.cs
SHKnudsen: will still need this property serialized, but as this property isn't there when we open the graph but gets applied after save the test will fail

PR #11917, opened 2021-08-05
built-in package states UI

pinzart90:
DYN-3670
Builtin package states UI

1. Refactored `LoadCustomNodesAndPackages` so that it can handle showing 2 packages with the same name (one built-in and the other non-built-in) in the installed packages UI
2. Added 1 new Commands :
   Load 
   
   Loading an unloaded built-in package will try to delete any conflicting packages and then either ask for a restart or go ahead with an immediate Load.


src/DynamoCoreWpf/Controls/InstalledPackagesControl.xaml
```diff
@@ -111,6 +111,18 @@
                                                                               Header="{Binding Path=PackageViewContextMenuUnmarkUninstallText}"
                                                                               ToolTip="{Binding Path=PackageViewContextMenuUnmarkUninstallTooltip}">
                                                             </MenuItem>
+                                                            <MenuItem Name="InstallButton"
+                                                                              Visibility="{Binding Path=CanInstall}"
```
sm6srw, line 115: Looks like we have a lot of overlap with my PR for DYN-3726 (my PR gets smaller and smaller). I solved the visibility aspect of this in a slightly different way for the uninstall commands using a style that hides menu items that are disabled. Lets pick one of these solutions in the end.

src/DynamoCoreWpf/Controls/InstalledPackagesControl.xaml
```diff
@@ -264,6 +276,11 @@
                                 </Grid>
                             </Expander>
                         </Border>
+                        <DataTemplate.Triggers>
+                            <DataTrigger Binding="{Binding Path=Unloaded}" Value="true">
+                                <Setter TargetName="PackageGrid" Property="Opacity" Value="0.5" />
```
sm6srw, line 281: What is this for? Was this in the UI Spec?

src/DynamoCoreWpf/Controls/InstalledPackagesControl.xaml
pinzart90: Sure....your solution sounds better
I can remove my changes

src/DynamoCoreWpf/Controls/InstalledPackagesControl.xaml
pinzart90: Ugh..I do not remember why I thought this was a good idea.
Basically makes Unloaded packages to appear greyed out

I will attach a snapshot. Maybe @Jingyi-Wen can tell us if it is useful or not

pinzart90: image

Jingyi-Wen: I can see this being helpful in terms of making the state more distinguishable. However in this case it may also feel like the context menu is disabled, which could be confusing

src/DynamoCoreWpf/Controls/InstalledPackagesControl.xaml
sm6srw: You can leave it if you want. I can make that change in my PR as soon as this is merged.

src/DynamoCoreWpf/DynamoCoreWpf.csproj
```diff
@@ -170,8 +170,8 @@
     <Compile Include="Interfaces\ILibraryViewCustomization.cs" />
     <Compile Include="Interfaces\LayoutSpecification.cs" />
     <Compile Include="Properties\Resources.Designer.cs">
-      <AutoGen>True</AutoGen>
       <DesignTime>True</DesignTime>
+      <AutoGen>True</AutoGen>
```
sm6srw, line 174: Is this change useful in any way? Looks like this change can be reverted.

sm6srw (commented):
I am little confused why we need these new install/uninstall commands. Can you give some background?

src/DynamoCoreWpf/DynamoCoreWpf.csproj
pinzart90: it was moved here automatically ....I think
I can move it back

sm6srw:
> > I am little confused why we need these new install/uninstall commands. Can you give some background?
> 
> I only added 2 commands:
> InstallCommand and UnmarkForInstallCommand (I chose Install vs Load because it fits better with the other public APIs...)
> We can change them all to Load/Delete ....
> 
> InstallCommand corresponds to the user context menu item : Load
> UnmarkForInstallCommand corresponds to the user context menu item: Cancel Load
> 
> Installing (loading) is available only for unloaded built-in packages
> Cancel install (load) is available only for built-in packages that have been scheduled for install (load) and still waiting for other conflicting packages to get deleted+ a dynamo restart
> 
> Installing(loading) an unloaded built-in package will try to delete any conflicting packages (pending a user confirmation)
> If conflicting packages cannot be deleted immediately then we get in scheduled states for the conflicting packages and the built in package

Ok, I think I get it. Thanks.

src/DynamoCoreWpf/Controls/InstalledPackagesControl.xaml
```diff
@@ -111,6 +111,16 @@
...
+                                                            <MenuItem Name="InstallButton"
+                                                                              Command="{Binding Path=InstallCommand}"
+                                                                              Header="{x:Static p:Resources.PackageViewContextMenuInstallText}"
+                                                                              ToolTip="{x:Static p:Resources.PackageViewContextMenuInstallTooltip}">
+                                                            </MenuItem>
+                                                            <MenuItem Name="UnmarkForInstallButton"
+                                                                              Command="{Binding Path=UnmarkForInstallCommand}"
+                                                                              Header="{x:Static p:Resources.PackageViewContextMenuUnmarkForInstallText}"
+                                                                              ToolTip="{x:Static p:Resources.PackageViewContextMenuUnmarkForInstallTooltip}">
+                                                            </MenuItem>
```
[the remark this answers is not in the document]
pinzart90: Actually we still need to bind to public properties for the others.
The new commands (Install and UnmarkInstall) are designed only for BuiltInPackages. That's why we can use static resources.
The rest still depend on the boolean `BuiltInPackage`

pinzart90:
The new Load and Cancel Load context menu items (and related messaging and states) look like this:
1. Trying to Load an unloaded built-in package that has a conflict with another package (with the same name) that has no loaded assemblies
image
2. Trying to Load an unloaded built-in package that has a conflict with another package (with the same name) that has loaded assemblies (requires restart)
image
image

src/DynamoPackages/PackageLoader.cs
```diff
@@ -386,16 +389,28 @@ public void LoadPackages(IEnumerable<Package> packages)
...
 
         /// <summary>
-        /// Scans and compiles any new packages contained in new package paths that are added.
-        /// Note: This method will still reload ALL packages in the VM even those that are not new.
+        /// This method is called when custom nodes and packages need to be reloaded if there are new package paths.
         /// </summary>
         /// <param name="loadPackageParams"></param>
-        private void LoadNewPackages(LoadPackageParams loadPackageParams)
+        /// <param name="customNodeManager"></param>
+        public void LoadCustomNodesAndPackages(LoadPackageParams loadPackageParams, CustomNodeManager customNodeManager)
```
sm6srw, line 396: @pinzart90 This looks OK to me.

sm6srw (approved):
LGTM!

mjkkirschner:
I'm going to continue this review I first I do not understand this:
```
Refactored LoadCustomNodesAndPackages so that it can handle loading 2 packages with the same name (one built-in and the other non-built-in)
```

how could this work?

src/DynamoCore/Search/SearchDictionary.cs
```diff
@@ -42,6 +42,15 @@ public IEnumerable<V> SearchEntries
             get { return entryDictionary.Keys; }
         }
 
+        /// <summary>
+        ///     Cleans up all recorded search entries
+        ///     Used for writing unit tests.
+        /// </summary>
+        internal void ClearAllEntries()
```
mjkkirschner, line 49: why is this needed, SearchDictonary shouldn't be static, are there tests where you want to clear search in the middle of a test?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageViewModel.cs
```diff
@@ -31,6 +32,14 @@ public bool HasAdditionalAssemblies
             get { return Model.LoadedAssemblies.Any(x => !x.IsNodeLibrary); }
         }
 
+        public bool Unloaded
```
mjkkirschner, line 35: please add /// summary to public fields.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageViewModel.cs
```diff
@@ -233,24 +256,125 @@ private void Uninstall()
...
             return false;
         }
 
-        private void GoToRootDirectory()
+        // Loads a built-in package that was previously set as Unloaded
+        private void Load()
+        {
+            var dynModel = dynamoViewModel.Model;
+            var packageLoader = dynModel.GetPackageManagerExtension().PackageLoader;
+            var conflicts = packageLoader.LocalPackages.ToList().Where(x => x.Name == Model.Name && x != Model);
```
mjkkirschner, line 280: does `package` override equals correctly or do you wish to use reference equality, if so, maybe use ReferenceEquals explicitly?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageViewModel.cs
```diff
@@ -233,24 +256,125 @@ private void Uninstall()
...
+                var dialogResult = MessageBoxService.Show(message,
+                    Resources.CannotDownloadPackageMessageBoxTitle,
+                        MessageBoxButton.OKCancel,
+                        MessageBoxImage.Exclamation);
+
+                if (dialogResult == MessageBoxResult.Cancel) return;
+            }
+
+            try
+            {
```
mjkkirschner, line 303: It's hard for me to tell, does anything in this try catch ever result in an error state?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageViewModel.cs
```diff
@@ -233,24 +256,125 @@ private void Uninstall()
...
+                {
+                    package.NotifyLoadStatePropertyChanged();
+                }
+            }
+        }
+
+        // Cancels the ScheduledForLoad status on a built-in package
+        private void UnmarkForLoad()
+        {
+            var dynModel = dynamoViewModel.Model;
```
mjkkirschner, line 343: this seems like it could be factored out into a private helper

src/DynamoPackages/PackageLoader.cs
```diff
@@ -386,16 +389,28 @@ public void LoadPackages(IEnumerable<Package> packages)
...
         {
+            foreach (var path in loadPackageParams.Preferences.CustomPackageFolders)
+            {
+                // Append the definitions subdirectory for custom nodes.
+                var dir = path == DynamoModel.BuiltInPackagesToken ? PathManager.BuiltinPackagesDirectory : path;
+                dir = TransformPath(dir, PathManager.DefinitionsDirectoryName);
+
+                customNodeManager.AddUninitializedCustomNodesInPath(dir, false, false);
```
mjkkirschner, line 404: did we actually ever discuss loading the /dyf folder from the Built-In Packages folder? what do we need this for?

src/DynamoPackages/PackageLoader.cs
```diff
@@ -386,16 +389,28 @@ public void LoadPackages(IEnumerable<Package> packages)
...
+            }
+
             foreach (var path in loadPackageParams.NewPaths)
             {
-                var packagesDirectory = pathManager.PackagesDirectories.Where(x => x.StartsWith(path)).FirstOrDefault();
-                ScanPackageDirectories(packagesDirectory, loadPackageParams.Preferences);
+                var packageDirs = pathManager.PackagesDirectories.Where(x => x.StartsWith(path));
```
mjkkirschner, line 409: so this has been improved so more than one path can be added?

src/DynamoPackages/PackageLoader.cs
```diff
@@ -503,7 +500,18 @@ private void ScanPackageDirectories(string root, IPreferences preferences)
 
                     var pkg = ScanPackageDirectory(dir, checkCertificates);
                     if (pkg != null && preferences.PackageDirectoriesToUninstall.Contains(dir))
-                        pkg.MarkForUninstall(preferences);
+                    {
+                        if (pkg.BuiltInPackage)
+                        {
+                            // If the discovered package is built-in and marked to be unloaded then we set it directly to Unloaded state.
```
mjkkirschner, line 506: so this comment says unloaded - but it's really if the built-in package's package root dir was contained in the uninstall list in preferences - yes?

src/DynamoPackages/PackageLoader.cs
```diff
@@ -544,7 +552,10 @@ public Package ScanPackageDirectory(string directory, bool checkCertificates)
 
                 var discoveredVersion = CheckAndGetPackageVersion(discoveredPackage.VersionName, discoveredPackage.Name, discoveredPackage.RootDirectory);
 
-                var existingPackage = LocalPackages.FirstOrDefault(package => package.Name == discoveredPackage.Name);
+                var existingPackage = LocalPackages.FirstOrDefault(package => (package.Name == discoveredPackage.Name) && (package.LoadState.State != PackageLoadState.StateTypes.Unloaded));
```
mjkkirschner, line 555: could you make this span multiple lines?

src/DynamoPackages/PackageLoader.cs
```diff
@@ -544,7 +552,10 @@ public Package ScanPackageDirectory(string directory, bool checkCertificates)
...
+
+                // Do we want to notify the user when we encounter unloaded packages ?
```
mjkkirschner, line 557: maybe just log to console for now?

mjkkirschner: hmm - actually, do you think logging this to analytics would be useful?

test/Libraries/PackageManagerTests/PackageLoaderTests.cs
```diff
@@ -277,13 +327,14 @@ public void LoadingBuiltInZTPackageAddsItToLibrary()
                 Preferences = CurrentDynamoModel.PreferenceSettings,
 
             };
+
+            CurrentDynamoModel.SearchModel.ClearAllEntries();
```
mjkkirschner, line 331: the fact that you need to do this has me a bit worried about loading ZT nodes multiple times into search- a bug that keeps coming back - can you explain it?

test/pkgs_signed/SignedPackage2/pkg.json
```diff
@@ -0,0 +1 @@
+{"license":"","file_hash":null,"name":"SignedPackage","version":"1.0.0","description":"original package","group":"","keywords":null,"dependencies":[],"contents":"","engine_version":"2.1.0.7840","engine":"dynamo","engine_metadata":"","site_url":"","repository_url":"","contains_binaries":true,"node_libraries":["SignedPackage2, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null"]}
```
mjkkirschner, line 1: I've already added this package here:
https://github.com/DynamoDS/Dynamo/tree/master/test/builtinpackages%20testdir/Packages/SignedPackage2

perhaps you could make a new signed package so we don't encounter conflicts between the loaded binaries in tests.

pinzart90:
> I'm going to continue this review but first I do not understand this, and it is ringing alarm bells.
> 
> ```
> Refactored LoadCustomNodesAndPackages so that it can handle loading 2 packages with the same name (one built-in and the other non-built-in)
> ```
> 
> how could this work?

By "load" I meant show in the installed packages UI (without actually loading into the VM)
Built-in packages are always shown in the UI (even if they are unloaded). A non-built in package (with the same name as a built-in) will be shown if it is loaded or in error state. @mjkkirschner let me know if you need more clarification on this

src/DynamoCore/Search/SearchDictionary.cs
pinzart90: I will answer in the test...where this is actually used.

test/Libraries/PackageManagerTests/PackageLoaderTests.cs
pinzart90: The problem seems to be only during tests (at least as far as I can tell)
Many tests use new instances of `PackageLoader`...expecting that there will be a clean slate every time.
In other words...many tests load packages into the VM....and do not clean it up (not even possible without a restart ...I think).
I this case...some tests (that ran before the `LoadingBuiltInZTPackageAddsItToLibrary` had loaded a bunch of packages). Now when the current test checks the SearchModel...it actually sees packages from previous tests, that have not been cleaned up.
My proposal, using `CurrentDynamoModel.SearchModel.ClearAllEntries`, is just a band aid for the immediate issue.

let me know if you can think of a better way of doing it ...or if you think this (or something similar)might happen when clients use Dynamo.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageViewModel.cs
pinzart90: From my understanding, by default, the referenceEquals is used by equality operators on objects 
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/equality-operators#reference-types-equality
The Package class does not override or overload anything related to equality comparers.
But I agree that it is a bit unclear...and always triggers an alarm. We can always override the Equality methods in the Package class to make things clearer...

src/DynamoCoreWpf/ViewModels/PackageManager/PackageViewModel.cs
pinzart90: I do not think it does.
I think `UninstallCore` can throw if the files on disk are write protected..
Not sure if we should continue or exit the "Load" command immediately in that case...

src/DynamoCoreWpf/ViewModels/PackageManager/PackageViewModel.cs
pinzart90: good idea

src/DynamoPackages/PackageLoader.cs
pinzart90: I think that logic came out of this commit https://github.com/DynamoDS/Dynamo/commit/a9b67336d76d9a94e7736305572a7d2620fd3f52
Not sure about that @mjkkirschner 
@aparajit-pratap any ideas ?

src/DynamoPackages/PackageLoader.cs
pinzart90: No, not really. We still only scan one path.
I really don't remember what I was trying to solve by removing the FIrstOrDefault.
I'll put that back

src/DynamoPackages/PackageLoader.cs
pinzart90: Correct. I can replace it (makes it clearer)

test/pkgs_signed/SignedPackage2/pkg.json
pinzart90: is the package that you added supposed to be the same as the built-in package ? (only not in a built in location ?)

src/DynamoPackages/PackageLoader.cs
pinzart90: Not sure if this is the best place for analytics (seeing that this is automatically done by Dynamo...and could be many...). Maybe we can add analytics on Specific user commands (Load, Unload, etc) ?

test/pkgs_signed/SignedPackage2/pkg.json
mjkkirschner: The tests I added which used this package override the builtin packages location to be `test/builtinpackages testdir/`

sm6srw:
> > I'm going to continue this review but first I do not understand this, and it is ringing alarm bells.
> > ```
> > Refactored LoadCustomNodesAndPackages so that it can handle loading 2 packages with the same name (one built-in and the other non-built-in)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > how could this work?
> 
> By "load" I meant show in the installed packages UI (without actually loading into the VM)
> Built-in packages are always shown in the UI (even if they are unloaded). A non-built in package (with the same name as a built-in) will be shown if it is loaded or in error state. @mjkkirschner let me know if you need more clarification on this

This was one of the things I had a hard time to understand until I realized that additional packages with the same name are simply discarded with a warning and will never show up in the UI. Would changing the comment to `can show 2 packages with the same name` and `regardless if they are loaded or not` make it clearer?

test/pkgs_signed/SignedPackage2/pkg.json
pinzart90: A conflict is what I am trying to test here.
I want the same package to be in both a Built-in localtion and a non built in location.
I want to test that both packages will be shown in the UI but only one will be loaded in the VM

pinzart90: Are you suggesting that I should move the duplicate, non-built in package (`SignedPackage2`) to a new location...so that other existing tests do not get conflicts ?

src/DynamoCore/Search/SearchDictionary.cs
pinzart90: Removed

test/Libraries/PackageManagerTests/PackageLoaderTests.cs
pinzart90: OK..I forgot that the DynamoModel is torn down after every test.
I updated the code with a proper fix

pinzart90:
@mjkkirschner @aparajit-pratap do you have any other thoughts on this PR ?

src/DynamoPackages/Package.cs
```diff
@@ -528,15 +528,45 @@ internal void MarkForUninstall(IPreferences prefs)
...
+        internal void UnmarkForLoad(IPreferences prefs)
+        {
+            if (BuiltInPackage && 
+                LoadState.ScheduledState == PackageLoadState.ScheduledTypes.ScheduledForLoad && 
+                LoadState.State == PackageLoadState.StateTypes.Unloaded)
+            {
+                LoadState.ResetScheduledState();
+                if (!prefs.PackageDirectoriesToUninstall.Contains(RootDirectory))
+                {
+                    prefs.PackageDirectoriesToUninstall.Add(RootDirectory);
```
mjkkirschner, line 565: I remember we had some trouble with the new package path views in preferences dialog being updated when this model was changed - I think thats fixed now, right @sm6srw ?

src/DynamoPackages/PackageLoader.cs
```diff
@@ -254,6 +254,9 @@ internal void TryLoadPackageIntoLibrary(Package package)
             // Prevent duplicate loads
             if (package.LoadState.State == PackageLoadState.StateTypes.Loaded) return;
```
mjkkirschner, line 255: so we only load packages with load state of none?

test/pkgs_signed/SignedPackage2/pkg.json
mjkkirschner: ah, I didn't realize you wanted a conflict...

test/DynamoCoreTests/DynamoModelTestBase.cs
```diff
@@ -37,6 +39,8 @@ protected override DynamoModel GetModel()
         public override void Setup()
         {
             base.Setup();
+
+            originalBuiltinPackagesDirectory = PathManager.BuiltinPackagesDirectory;
```
mjkkirschner, line 43: @pinzart90 I guess this is better... definitely more magical :) Maybe update the comment on `BuiltinPackagesDirectory` to note it's reset only on `DynamoModelTestBase` - one unfortunate issue we have lots of test base classes.

mjkkirschner (approved):
looks solid!

src/DynamoPackages/PackageLoader.cs
```diff
@@ -503,7 +500,19 @@ private void ScanPackageDirectories(string root, IPreferences preferences)
 
                     var pkg = ScanPackageDirectory(dir, checkCertificates);
                     if (pkg != null && preferences.PackageDirectoriesToUninstall.Contains(dir))
```
sm6srw, line 502: Something is strange here. `PackageDirectoriesToUninstall` is always empty in my testing. Even when there is a package path in the settings file. Any ideas?

sm6srw: FYI @pinzart90

mjkkirschner:
okay - I do have one more question:
```
Loading an unloaded built-in package will try to delete any conflicting packages (pending a user confirmation)
If conflicting packages cannot be deleted immediately then we get in scheduled states for the conflicting packages and the built in package
```

Theres another solution to this problem I think - in the case where the packages cannot be deleted -  have you considered that instead of deleting the other conflicting packages, we just don't do anything - in most cases the `Built-In Packages` directory will be first in the search paths, so on next restart the other packages will be skipped since the builtin package will load first. (unless the user reorganized their search paths)

I'm not sure this is better - it's not consistent with how package install works today, but it could takes advantage of the new states - ie don't force a delete thats not necessary.

src/DynamoPackages/Package.cs
sm6srw: Yes

src/DynamoPackages/PackageLoader.cs
pinzart90: You mean the value of `PackageDirectoriesToUninstall ` when debugging ?
Or the UI combobox that corresponds to that data ?

src/DynamoPackages/PackageLoader.cs
pinzart90: Well I think at the moment we allow Error states to be loaded as well
maybe we could only allow None ...
Error is pretty ambiguous at the moment (could be partially loaded in the VM...I guess)

pinzart90:
> okay - I do have one more question:
> 
> ```
> Loading an unloaded built-in package will try to delete any conflicting packages (pending a user confirmation)
> If conflicting packages cannot be deleted immediately then we get in scheduled states for the conflicting packages and the built in package
> ```
> 
> Theres another solution to this problem I think - in the case where the packages cannot be deleted - have you considered that instead of deleting the other conflicting packages, we just don't do anything - in most cases the `Built-In Packages` directory will be first in the search paths, so on next restart the other packages will be skipped since the builtin package will load first. (unless the user reorganized their search paths)
> 
> I'm not sure this is better - it's not consistent with how package install works today, but it could takes advantage of the new states - ie don't force a delete thats not necessary.
> 
> I may be misinterpreting what happens when the delete cannot be done immediately, ie which states do the other packages get set to? (schedule for delete? or something else)

Well I guess I could just add an optimization to check against the path load order. If the conflicts can be resolved without deletion...we could just ask for a restart. But that does not guard against the scenario where the user changes the Paths after trying to Load a built-in Package.

src/DynamoPackages/PackageLoader.cs
sm6srw: When debugging. I am trying to get a single built-in package set to `Unloaded` after restart. It always show up as `Loaded` for me after restart (and the path are removed from the settings file).

sm6srw: Looks like the directories are removed here: https://github.com/DynamoDS/Dynamo/pull/11917/files#diff-7ae21a0e381958737cc115ada84cbd9f96efe3315928cbc44b367967923cf55bR804 before we get to this spot.

pinzart90: found the issue...I was resetting the `PackageDirectoriesToUninstall` for built in packages.
SHould work now...but there is still a bit of UX to do I guess...there is no clear indication when a built in package is manually unloaded (will be unloaded until explicitly Loaded) or automatically unloaded due to a conflict (which will be unloaded only as long as the conflict persists)

src/DynamoPackages/PackageLoader.cs
mjkkirschner: sounds okay to me - just curious.

mjkkirschner:
> I guess it sounds a bit safer to not depend on the load order....

well - you're assuming that if a user marks a builtin package to load - that Dynamo should try it's damndest to make it load on the next restart - I think thats a pretty good assumption - it's just not the simplest thing we could do to unblock users.

src/DynamoPackages/PackageLoader.cs
sm6srw: Thanks, I will give it a try. For the UX part, I have not found it that confusing. I will think about it when I play around with it today.

sm6srw: @pinzart90 Works fine now!

src/DynamoCoreWpf/ViewModels/PackageManager/PackageViewModel.cs
```diff
@@ -204,53 +235,154 @@ private void Uninstall()
...
+                foreach (var package in conflicts)
+                {
+                    package.UninstallCore(dynamoModel.CustomNodeManager, packageLoader, dynamoModel.PreferenceSettings);
+                }
+
+                if (!hasConflictsWithLoadedAssemblies)
+                {
+                    try
+                    {
+                        packageLoader.TryLoadPackageIntoLibrary(Model);
```
sm6srw, line 322: @pinzart90 New problem. Here is what I did with a single singed package installed at the built-in location:
1. Started Dynamo and marked the package for `Unload`.
2. Restarted Dynamo
3. The package is now labeled as `Unloaded`. I mark it for `Load`. 
The package is now labeled as `Loaded` but is not actually loaded (It does not show up in the library). The call to `TryLoadPackageIntoLibrary` does nothing.
And finally, the package end up as `Unloaded` again if I restart `Dynamo` as this point.

pinzart90: Nice catch. Should be fixed now.
Will try to add test coverage for this

sm6srw: Works fine now!

src/DynamoCoreWpf/Properties/Resources.en-US.resx
```diff
@@ -2565,6 +2569,32 @@ This package will be deleted after the next Dynamo restart.</value>
   <data name="PackageContextMenuUnmarkUnloadPackageTooltip" xml:space="preserve">
     <value>Remove the scheduled unload status</value>
   </data>
+  <data name="PackageViewContextMenuLoadText">
+    <value>Load</value>
+  </data>
+  <data name="PackageViewContextMenuLoadTooltip">
+    <value>Load this package into Dynamo. Other packages with the same name will be automatically deleted</value>
```
aparajit-pratap, line 2576: I don't understand this. Loading a package will automatically delete other packages, isn't that bad?

aparajit-pratap: Does this mean if you try to load a built-in package, non-built-in packages with the same name will automatically be marked for deletion? And I think the built-in package will remain unloaded until you restart, correct?

aparajit-pratap: So in this case, the non-built-in package will have the `marked for deletion` state and the built-in will have the `scheduled for load` state?

src/DynamoCoreWpf/Properties/Resources.en-US.resx
```diff
@@ -2565,6 +2569,32 @@ This package will be deleted after the next Dynamo restart.</value>
...
+  <data name="MessageLoadBuiltInPackageWithRestart" xml:space="preserve">
+    <value>The built in package {1} cannot be loaded.
+Loading it would conflict with one or more packages that are already loaded. 
+To load {1}, Dynamo needs to first delete the conflicting packages. 
+Restart Dynamo to complete the install.
+
+Uninstall the following packages: {0}?</value>
+  </data>
+  <data name="MessageLoadBuiltInPackage" xml:space="preserve">
+    <value>Loading the built-in package {1} conflicts with one or more packages that are already loadeded.
```
aparajit-pratap, line 2593: typo: loaded

aparajit-pratap: is this for non-assembly containing packages that are conflicting?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageViewModel.cs
```diff
@@ -70,7 +85,9 @@ public string PackageLoadStateTooltip
 
                 switch (Model.LoadState.State)
                 {
-                    case PackageLoadState.StateTypes.Unloaded: return Resources.PackageStateUnloadedTooltip;
+                    case PackageLoadState.StateTypes.Unloaded:
+                        return dynamoModel.PreferenceSettings.PackageDirectoriesToUninstall.Contains(Model.RootDirectory) ?
+                           Resources.PackageStateManuallyUnloadedTooltip : Resources.PackageStateAutomaticallyUnloadedTooltip;
```
aparajit-pratap, line 90: Could you explain what this i doing? I find the descriptions automatically and manually unloaded confusing.

aparajit-pratap (commented):
~~Still reviewing ...~~

~~As always super confused ...~~

src/DynamoCoreWpf/Properties/Resources.en-US.resx
pinzart90: > So in this case, the non-built-in package will have the `marked for deletion` state and the built-in will have the `scheduled for load` state?

Correct.

In regards to the tooltip wording. I am not 100% happy with it. Not even sure we should make a distinction between Loading an automatic Unloaded package or a manually unloaded package.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -386,16 +389,28 @@ public void LoadPackages(IEnumerable<Package> packages)
         }
 
         /// <summary>
-        /// Scans and compiles any new packages contained in new package paths that are added.
-        /// Note: This method will still reload ALL packages in the VM even those that are not new.
+        /// This method is called when custom nodes and packages need to be reloaded if there are new package paths.
         /// </summary>
         /// <param name="loadPackageParams"></param>
-        private void LoadNewPackages(LoadPackageParams loadPackageParams)
```
aparajit-pratap, line 393: Why has this been removed? This method was added for a reason as part of a recent fix. It was specifically done as part of https://github.com/DynamoDS/Dynamo/pull/11835/files for a bug fix to prevent reloading of packages ⚠️

src/DynamoPackages/PackageLoader.cs
aparajit-pratap: I think this is the discussion we had related to this: https://autodesk.slack.com/archives/C01C8GGT9EY/p1625175309318800

sm6srw:
@pinzart90 I think I have covered all combinations in my WIP UI Automation PR. Looks like everything works as expected. This is ready to go from my standpoint.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -553,6 +564,14 @@ public Package ScanPackageDirectory(string directory, bool checkCertificates)
                     return discoveredPackage; // success
                 }
 
+                // Conflict invloving a built-in package
+                if (discoveredPackage.BuiltInPackage || existingPackage.BuiltInPackage)
+                {
+                    // We show both packages but we mark the new one as unloaded.
+                    discoveredPackage.LoadState.SetAsUnloaded();
+                    Add(discoveredPackage);
```
aparajit-pratap, line 572: Is this `Add` call so that we can show it in the UI?

aparajit-pratap:
Just a couple of questions. I don't have any suggestions for changes in implementation as I don't understand it fully but maybe the tooltips can be improved to make them clearer as to what exactly they mean.

pinzart90:
> @pinzart90 I think I have covered all combinations in my WIP UI Automation PR. Looks like everything works as expected. This is ready to go from my standpoint.

I just found a case when we do not provide the best UX for the user. 
A built-in package Xb, can be in an **Unloaded** state and at the same time `PreferenceSettings.PackageDirectoriesToUninstall` can contain Xb.RootDir or not.
IF package Xb was set as Unloaded directly by the user, then Xb.RootDir will be added to `PreferenceSettings.PackageDirectoriesToUninstall` - to save the user's choice
If the package Xb was set as Unloaded by the PackageLOader, because of conflicts with a non-builtin Package Xnb, then then Xb.RootDir will not be added to `PreferenceSettings.PackageDirectoriesToUninstall`

The problem is with calling the command "LOad" and then "UnmarkForLoad" when you have conflicts.
Calling Load will always remove the built in package Xb.RootDir from `PreferenceSettings.PackageDirectoriesToUninstall`
Now calling "UnmarkForLoad" has no way of knowing if Xb.RootDir should be re-added or not to `PreferenceSettings.PackageDirectoriesToUninstall`

src/DynamoPackages/Package.cs
```diff
@@ -528,15 +529,52 @@ internal void MarkForUninstall(IPreferences prefs)
...
+                    LoadState.SetScheduledForLoad();
+                }
+                else
+                {
+                    LoadState.SetAsLoaded();
+                }
+
+                if (prefs.PackageDirectoriesToUninstall.Contains(RootDirectory))
+                {
+                    prefs.PackageDirectoriesToUninstall.Remove(RootDirectory);
```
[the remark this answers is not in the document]
pinzart90: When a user wants to Unload a built-in package (form the UI), the built-in package's Root directory will be added to prefs.PackageDirectoriesToUninstall

Now If the user wants to Load that package, we remove the package;'s rootDIr from prefs.PackageDirectoriesToUninstall so that it will not be set as Unloaded on the next dynamo restart.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -247,13 +247,16 @@ private IEnumerable<CustomNodeInfo> OnRequestLoadCustomNodeDirectory(string dire
...
-        internal void TryLoadPackageIntoLibrary(Package package)
+        private void TryLoadPackageIntoLibrary(Package package)
         {
             Add(package);
 
             // Prevent duplicate loads
             if (package.LoadState.State == PackageLoadState.StateTypes.Loaded) return;
 
+            // Prevent loading packages that have been specifically marked as unloaded
+            if (package.LoadState.State == PackageLoadState.StateTypes.Unloaded) return;
```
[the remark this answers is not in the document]
pinzart90: well the `Add(package)` call only adds the package to the LocalPackages array
The 2 checks are supposed to filter out loading into the VM (which is done bellow those lines)

src/DynamoPackages/PackageLoader.cs
pinzart90: yes. The `Add` call will add the package to the LocalPackages array....but with the state set to Unloaded (which will not allow loading into the VM)

src/DynamoCoreWpf/ViewModels/PackageManager/PackageViewModel.cs
```diff
@@ -204,53 +236,157 @@ private void Uninstall()
...
+
+                if (!hasConflictsWithLoadedAssemblies)
+                {
+                    try
+                    {
+                        // Set the package load state as None so that we can try to load it again
+                        Model.LoadState.ResetState();
+
+                        packageLoader.LoadPackages(new List<Package>() { Model });
+                        Model.MarkForLoad(dynamoModel.PreferenceSettings, false);
```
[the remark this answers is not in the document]
pinzart90: Yes, MarkForLoad will do a bit of redundant work, 
The thing we actually need is to remove the package from the PreferenceSettings.PackageDirectoriesToUninstall
I could replace it with just that call actually....

aparajit-pratap:
> If the package Xb was set as Unloaded by the PackageLOader, because of conflicts with a non-builtin Package Xnb
How/when will this case happen?

pinzart90:
> > If the package Xb was set as Unloaded by the PackageLOader, because of conflicts with a non-builtin Package Xnb
> 
> How/when will this case happen?

This would happen when you have the same package in a Builtin load location and in a Non-Builtin load location.
If the Non-builtin location takes precedence (ffrom the Path manager) then the built in package will be set as Unloaded (but shown in the isntalled packages UI)

src/DynamoPackages/PackageLoader.cs
aparajit-pratap: But won't the `discoveredPackage `already be added to the array some time earlier?

src/DynamoCoreWpf/Properties/Resources.en-US.resx
aparajit-pratap: How about changing this to:
```
Load this package into Dynamo. Upon restart, other packages with the same name will be automatically deleted if they are custom packages or will be unloaded if they are built-in packages.
```

src/DynamoPackages/PackageLoader.cs
pinzart90: discoveredPackage is the new package that we are trying to Load.
existingPackage is the package with the same name that was already loaded.

At line code 570, the discoveredPackage  has not yet been added to the LocalPackages list

src/DynamoCoreWpf/Properties/Resources.en-US.resx
pinzart90: YEs...which pretty much translates to "Loading that does not require a restart".

src/DynamoCoreWpf/ViewModels/PackageManager/PackageViewModel.cs
pinzart90: I removed this...IT was an attempt to distinguish between implicit and explicit Unload at UI level

src/DynamoCoreWpf/Properties/Resources.en-US.resx
pinzart90: Well, Load only applies to built in packages and conflicting packages can only be non-built in.
THan means that conflicting packages can only be deleted (never just unloaded)

src/DynamoPackages/Package.cs
```diff
@@ -550,8 +588,21 @@ internal void UninstallCore(CustomNodeManager customNodeManager, PackageLoader p
...
-                packageLoader.Remove(this);
-                Directory.Delete(RootDirectory, true);
+                if (BuiltInPackage)
+                {
+                    LoadState.SetAsUnloaded();
+                    RaisePropertyChanged(nameof(LoadState));
+
+                    if (!prefs.PackageDirectoriesToUninstall.Contains(RootDirectory))
+                    {
+                        prefs.PackageDirectoriesToUninstall.Add(RootDirectory);
```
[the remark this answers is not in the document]
pinzart90: UninstallCore handle both built in and non built in.
For non built in ...pretty much unchanged.
For built in - if it has assemblies loaded - goes to markForUninstall
                  - if it does not have any assemblies loaded - sets state as UNloaded
                                                                                        - removes LoadedCustomNodes from customNOdeManager
                                                                                        - adds rootDir to PackageDirectoriesToUninstall (explicit unload)

mjkkirschner:
@pinzart90 where do these latest change leave us? It appears the `scheduleForLoad` state still exists - but not accessible via the UI? Is that correct?

sm6srw (commented):
I have tested the new workflow and it looks like it works fine. Two restarts are now needed to get to the end and finally be able to Load the package.

aparajit-pratap:
> I have tested the new workflow and it looks like it works fine. Two restarts are now needed to get to the end and finally be able to Load the package.

One restart to unload and delete the existing package, and what is the second one for? Can't the user directly load the built-in package upon restart since the conflicting package should've been deleted by then?

sm6srw:
> > I have tested the new workflow and it looks like it works fine. Two restarts are now needed to get to the end and finally be able to Load the package.
> 
> One restart to unload and delete the existing package, and what is the second one for? Can't the user directly load the built-in package upon restart since the conflicting package should've been deleted by then?

Correct, it is only one restart if you put it that way. I should have been clearer about that. I have been playing around with a little more complicated case that before required one restart (and it did actually not work in the end because of this issue) and it now works with one additional restart. That was what I was trying to say.

pinzart90:
> @pinzart90 where do these latest change leave us? It appears the `scheduleForLoad` state still exists - but not accessible via the UI? Is that correct?

Yes..I was thinking it might be useful to leave it in there in case we have to revisit this part later...not convinced though...

sm6srw:
> Yes..I was thinking it might be useful to leave it in there in case we have to revisit this part later...not convinced though...
I think best practice is to remove code like this. We can always add it back if we need it. But it will come back and confuse someone in the future otherwise.

test/Libraries/SystemTestServices/SystemTestBase.cs
```diff
@@ -35,6 +35,8 @@ public abstract class SystemTestBase
         protected string workingDirectory;
         private Preloader preloader;
         private AssemblyResolver assemblyResolver;
+        // Some tests override the static property PathManager.BuiltinPackagesDirectory, so we need a way to reset it after each test.
```
sm6srw, line 38: 👍

sm6srw (approved):

PR #11919, opened 2021-08-06
Callsite ExecWithZeroRI optimizations to improve Cross Product replication performance 

saintentropy:
### Purpose

This PR is follow up to some previous optimizations within the Callsite class, specifically to the ExecWithZeroRI method related to Cross Product replication.  A majority of these changes reduce temporary object allocations and associated GC time.  The larger the data set the larger the impact of the optimization.  This task covers:

- Removing the `AreParametersHomogeneousTypes` method check as we can more efficiently depend on the FunctionEndPoint count.  If the count is equal to 1 then we can directly resolve the `finalFunctionEndPoint` and cache it.  The optimization was already in place but based on the `AreParametersHomogeneousTypes` which appears to be extraneous. 
- Rename `FunctionEndPoint` list variable name to plural for clarity
- Adjust allocation of temporary parameter arrays outside of replication loop to save allocation and GC time.
- Use `DSArray` directly to access parameters vs copying items to temporary arrays.
- Set list capacity directly vs resizing dynamically after initialization.

todo => validate test
 
### Declarations

### FYIs

@jasonstratton @QilongTang

### DIFF
--- a/src/Engine/ProtoCore/Lang/CallSite.cs
+++ b/src/Engine/ProtoCore/Lang/CallSite.cs
@@ -1530,7 +1530,7 @@ namespace ProtoCore
 
 
         private StackValue Execute(
-            List<FunctionEndPoint> functionEndPoint,
+            List<FunctionEndPoint> functionEndPoints,
             Context c,
             List<StackValue> formalParameters,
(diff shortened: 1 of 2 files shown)


saintentropy:
Wrong branch...  closing

PR #11920, opened 2021-08-07
DYN-4229 GraphicDataProvider optimization for repeated calls to get similar object types for tessellation

saintentropy:
### Purpose

https://jira.autodesk.com/browse/DYN-4229

This optimization is to short-circuit the repeated search for and allocation of numerous `CLRObjectMarshaller` objects during `UpdateRenderPackageAsyncTask` execution.  Right now, within `UpdateRenderPackageAsyncTask`, we call `mirrorData.GetData()` for every object we are attempting to tessellate.  Within that call, a new instance of an ObjectMarshaller allocated based on the objects type information.  This optimization caches the ObjectMarshaller and reuses it if the object type is the same for repeated calls.

### Declarations

### Reviewers

TBD

### FYIs

@jasonstratton @QilongTang


saintentropy:
In a test graph with a lot of repeated calls to Circle.ByPointRadius this call reduced the time spent in this method from .5s to .07s and also reduced temporary memory allocation by 90%.

sm6srw (approved):
Great! This looks solid to me. I did spend some time making sure Clear gets called when it should be called and it looks good.

src/Engine/ProtoCore/Reflection/GraphicDataProvider.cs
```diff
@@ -141,8 +141,17 @@ internal List<IGraphicItem> GetGraphicItems(DSASM.StackValue svData, RuntimeCore
             return null;
         }
 
+        //Store marshaller for repeated calls to GetCLRObject.  Used for short-circuit of re-allocation of marshaller / interpreter object.
+        private static Interpreter interpreter;
+        private static ProtoFFI.FFIObjectMarshaler marshaler;
+        
         internal object GetCLRObject(StackValue svData, RuntimeCore runtimeCore)
         {
+            if (marshaler != null)
```
aparajit-pratap, line 150: Do we also need to do a null check for the interpreter here?

src/Engine/ProtoCore/Reflection/GraphicDataProvider.cs
```diff
@@ -141,8 +141,17 @@ internal List<IGraphicItem> GetGraphicItems(DSASM.StackValue svData, RuntimeCore
             return null;
         }
 
+        //Store marshaller for repeated calls to GetCLRObject.  Used for short-circuit of re-allocation of marshaller / interpreter object.
+        private static Interpreter interpreter;
+        private static ProtoFFI.FFIObjectMarshaler marshaler;
```
aparajit-pratap, line 146: Why do these need to be static?

src/Engine/ProtoCore/Reflection/GraphicDataProvider.cs
saintentropy: I don't think so....  The marshaler and interpreter are assigned at the same time.  But maybe good just to be sure.

src/Engine/ProtoCore/Reflection/GraphicDataProvider.cs
saintentropy: They are static because you only need one Interpreter and FFIObjectMarshaller for the `GraphicDataProvider` class until the runtime is reset.  If you made the instances you would allocate permanently an interpreter and marshaller for every mirror data object and that would be worse than temporary allocation

aparajit-pratap: I see that the `graphicdataprovider `member for `mirrordata` is static too so it looks like there will only be one instance of `graphicdataprovider ` ever but I guess it doesn't hurt to make these static here as well.

aparajit-pratap (approved):

PR #11921, opened 2021-08-07
DYN-4230 Callsite ExecWithRISlowPath optimizations to improve Cross Product replication performance

saintentropy:
### Purpose

https://jira.autodesk.com/browse/DYN-4230

This PR is follow up to some previous optimizations within the Callsite class, specifically to the ExecWithZeroRI method related to Cross Product replication.  A majority of these changes reduce temporary object allocations and associated GC time.  The larger the data set the larger the impact of the optimization.  This task covers:

- Removing the `AreParametersHomogeneousTypes` method check as we can more efficiently depend on the FunctionEndPoint count.  If the count is equal to 1 then we can directly resolve the `finalFunctionEndPoint` and cache it.  The optimization was already in place but based on the `AreParametersHomogeneousTypes` which appears to be extraneous. 
- Rename `FunctionEndPoint` list variable name to plural for clarity
- Adjust allocation of temporary parameter arrays outside of replication loop to save allocation and GC time.
- Use `DSArray` directly to access parameters vs copying items to temporary arrays.
- Set list capacity directly vs resizing dynamically after initialization.

todo => validate test
 
### Declarations

### FYIs

@jasonstratton @QilongTang


sm6srw:
@saintentropy Hey, can you merge master and make this PR current when you have a moment?

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -1615,7 +1615,12 @@ internal StackValue DispatchDispose(StackValue stackValue, RuntimeCore runtimeCo
             //Recursion base case
             if (replicationInstructions.Count == 0)
             {
-                return ExecWithZeroRI(functionEndPoint, c, formalParameters, stackFrame, runtimeCore, previousTraceData, newTraceData, finalFunctionEndPoint);
+                return ExecWithZeroRI(functionEndPoints, c, formalParameters, stackFrame, runtimeCore, previousTraceData, newTraceData, finalFunctionEndPoint);
+            }
+
+            if (finalFunctionEndPoint == null && functionEndPoints.Count == 1)
+            {
+                finalFunctionEndPoint = SelectFinalFep(c, functionEndPoints, formalParameters, stackFrame, runtimeCore);
```
aparajit-pratap, line 1623: What's the logic of determining the final function endpoint on the basis of `functionEndPoints` count vs. homogeneous parameters? I don't quite understand how one check can be swapped for another. The two seem to be different things, don't they?

aparajit-pratap: Ah, does this mean that previously you determined how many overloads of a function needed to be processed depending on the types of parameters whereas now you just check for the number of overloads?

aparajit-pratap: If there's just one `functionEndPoint`, can't we just say `finalFunctionEndPoint = functionEndPoints[0]`?

aparajit-pratap: Never mind, I guess it could be `null`.

sm6srw:
@saintentropy Looks good. I have not seen any regressions. I have seen execution times drop up to 50% but I have not seen huge drops in use of temporary memory. But the heap is in much better shape after this changes.

sm6srw:
I will wait with approving/merging these changes until we are done with our releases.

src/Engine/ProtoCore/Lang/CallSite.cs
saintentropy: @aparajit-pratap Yes you second comment is correct.  `AreParametersHomogeneousTypes` didn't check how may overloads functions were required.  It just check if all the types were the same and then inferred no overloads were required.   What I realized in testing is that the functionsEndPoints.Count was already determining the same information and a better test of the same result.  The `AreParametersHomogeneousTypes` check is now superfluous.  If you only have one option return from `ComputeFeps()` which means `functionsEndpoints.Count == 1` then it is safe to assign the functionEndPoint once for the life of the `Callsite` replication passes.

aparajit-pratap (approved):

sm6srw (approved):

sm6srw:
Tests are passing. I will merge this now.

PR #11923, opened 2021-08-09
DYN-4231 Fast Path for sweep during FullGC

saintentropy:
### Purpose

https://jira.autodesk.com/browse/DYN-4231

The purpose of this PR is to optimize the Sweep pass for the FullGC pass that occurs after any change to the graph that causes re-execution.   Specifically this optimization does the rolling:

*  introduces a specific Call method to the Executive (`CallDispose`) tailored specifically for dispatching Dispose calls vs allowing these calls from the GC to pass through the generic `Callr`.  Additionally it introduces a specific Dispatch method to the Callsite ('DispatchDispose`) also tailored to Dispatch Dispose calls vs `Dispatch`.  In both case the normal overhead of the `Callr` and `Dispatch` is much higher and .NET memory/GC intensive than the actual `Dispose` calls.
* Adds caching based on object type where possible as the GC is often collecting similar types repeatedly

In testing a graph that had ~150000 items to collect in the sweep pass the time was reduced from 2s to .5s and the .NET memory allocation from 800mb to 16mb.

Todo is to validate with testing include DS object dispose.

### Declarations

### Reviewers

TBD

### FYIs

@jasonstratton @QilongTang


src/Engine/ProtoCore/DSASM/Executive.cs
```diff
@@ -817,6 +817,36 @@ private List<StackValue> PopArgumentsFromStack(int argumentCount)
             return sv;
         }
 
+        // cached values for repeated calls of the same object type and dispose function.
+        private int previousClassIndex;
+        private string previousProcedureName;
+        private CallSite callsite;
+
+        public StackValue CallDipose(ProcedureNode fNode,
```
sm6srw, line 825: Spelling `CallDispose`

src/Engine/ProtoCore/DSASM/Executive.cs
```diff
@@ -817,6 +817,36 @@ private List<StackValue> PopArgumentsFromStack(int argumentCount)
...
+            {
+                previousClassIndex = classIndex;
+                previousProcedureName = fNode.Name;
+                callsite = new CallSite(classIndex, fNode.Name, exe.FunctionTable, runtimeCore.Options.ExecutionMode);
+            }
+
+            Validity.Assert(null != callsite);
+
+            StackValue sv = StackValue.Null;
+            sv = callsite.DispatchDispose(svThisPtr, runtimeCore);
```
aparajit-pratap, line 845: We can simply combine these statements into one:
```
StackValue sv = callsite.DispatchDispose(svThisPtr, runtimeCore);
```

src/Engine/ProtoCore/DSASM/Heap.cs
```diff
@@ -624,40 +624,45 @@ private bool TryFindFreeIndex(out int index)
             }
         }
 
+        //cache ClassIndex and ProcdureNode for repeated calls of the same type object.
+        private int previousClassIndex;
+        private ProcedureNode pn;
```
aparajit-pratap, line 629: I think we should rename this to something more descriptive for clarity.

src/Engine/ProtoCore/DSASM/Heap.cs
```diff
@@ -624,40 +624,45 @@ private bool TryFindFreeIndex(out int index)
...
                 }
             }
 
             if (pn != null)
             {
-                // TODO Jun/Jiong: Use build pointer utilities 
-                exe.rmem.Push(StackValue.BuildArrayDimension(0));
-                exe.rmem.Push(StackValue.BuildPointer(svPtr.Pointer, svPtr.metaData));
-                exe.rmem.Push(StackValue.BuildInt(1));
-
```
aparajit-pratap, line 653: Why have these lines been removed?

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -1528,6 +1528,27 @@ private FunctionGroup FirstFunctionGroupInInheritanceChain(RuntimeCore runtimeCo
             return ret;
         }
 
+        //Pre-initialize array for repeated calls.  The StackValue is inserted vs making a new array for every call.
+        private static List<StackValue> disposeArguments = new List<StackValue>() { StackValue.Null };
+        //Cache final function endpoint for repeated calls;
+        private FunctionEndPoint finalFep;
+
+        public StackValue DispatchDispose(
```
aparajit-pratap, line 1536: Why is this public?

src/Engine/ProtoCore/DSASM/Executive.cs
aparajit-pratap: +1

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -1528,6 +1528,27 @@ private FunctionGroup FirstFunctionGroupInInheritanceChain(RuntimeCore runtimeCo
...
+            StackValue stackValue, RuntimeCore runtimeCore)
```
aparajit-pratap, line 1537: Can we have this line on the same line as 1536?

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -1528,6 +1528,27 @@ private FunctionGroup FirstFunctionGroupInInheritanceChain(RuntimeCore runtimeCo
...
+        {
+            if (finalFep == null)
+            {
+                var funcGroup = FirstFunctionGroupInInheritanceChain(runtimeCore, classScope);
+                finalFep = funcGroup.FunctionEndPoints[0];
```
aparajit-pratap, line 1542: Why is the Dispose method always the first fep?

src/Engine/ProtoCore/Lang/FFIFunctionEndPoint.cs
```diff
@@ -84,9 +84,7 @@ private void Init(RuntimeCore runtimeCore)
         }
 
         public override StackValue Execute(Runtime.Context c, List<StackValue> formalParameters, StackFrame stackFrame, RuntimeCore runtimeCore)
-        {   
-            StackValue svThisPtr = stackFrame.ThisPtr;
```
aparajit-pratap, line 88: Was the stackFrame null in some cases? If so, was this crashing before, is this why you added the null check in line 98?

src/Engine/ProtoCore/DSASM/Heap.cs
```diff
@@ -624,40 +624,45 @@ private bool TryFindFreeIndex(out int index)
...
-                
                 ++exe.RuntimeCore.FunctionCallDepth;
 
                 // TODO: Need to move IsExplicitCall to DebugProps and come up with a more elegant solution for this
                 // fix for IDE-963 - pratapa
                 bool explicitCall = exe.IsExplicitCall;
                 bool tempFlag = explicitCall;
-                exe.Callr(pn.RuntimeIndex, pn.ID, classIndex, ref explicitCall);
+                //exe.Callr(pn.RuntimeIndex, pn.ID, classIndex, ref explicitCall);
+                exe.CallDipose(pn, svPtr, classIndex, ref explicitCall);
```
aparajit-pratap, line 665: `explicitCall` is not being touched inside of `CallDispose()`? Can we check this? As far as I remember this had some implications in the case of Dispose calls. Is it no longer relevant?

src/Engine/ProtoCore/Lang/CallSite.cs
aparajit-pratap: Has this been tested with DesignScript objects? Is the Dispose method defined on a DS class called after these changes?

sm6srw (commented):
Many thanks @aparajit-pratap for your insights. I have been trying to break this in various ways, unsuccessfully so far. It works well. In many cases where it currently takes 10-15 times longer to dispose objects than it takes to create them it now takes about the same time to create and dispose.  I am also seeing about 20% less memory being used.

src/Engine/ProtoCore/DSASM/Executive.cs
saintentropy: Spelling fixed and set to internal

src/Engine/ProtoCore/DSASM/Executive.cs
saintentropy: Good call... fixed

src/Engine/ProtoCore/DSASM/Heap.cs
saintentropy: I renamed it to `disposeProcedureNode`

src/Engine/ProtoCore/DSASM/Heap.cs
```diff
@@ -624,40 +624,45 @@ private bool TryFindFreeIndex(out int index)
...
-                exe.rmem.Push(StackValue.BuildInt(1));
-                
                 ++exe.RuntimeCore.FunctionCallDepth;
 
                 // TODO: Need to move IsExplicitCall to DebugProps and come up with a more elegant solution for this
                 // fix for IDE-963 - pratapa
                 bool explicitCall = exe.IsExplicitCall;
                 bool tempFlag = explicitCall;
-                exe.Callr(pn.RuntimeIndex, pn.ID, classIndex, ref explicitCall);
+                //exe.Callr(pn.RuntimeIndex, pn.ID, classIndex, ref explicitCall);
```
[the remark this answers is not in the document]
saintentropy: deleted

src/Engine/ProtoCore/DSASM/Heap.cs
saintentropy: This was not necessary with the new route and would be a memory leak.  Also FYI in Dynamo today it does seem to have a memory leak regardless.  We always add 3 items but only ever Pop two.

src/Engine/ProtoCore/Lang/CallSite.cs
saintentropy: moved to internal

src/Engine/ProtoCore/Lang/CallSite.cs
saintentropy: Fixed

src/Engine/ProtoCore/Lang/CallSite.cs
saintentropy: I pulled some test back out of the past and it did fail.  Not sure how much this case needs to be covered but I updated the PR to handle legacy DS object dispose with the original callr

src/Engine/ProtoCore/Lang/FFIFunctionEndPoint.cs
saintentropy: This check is new.  The `stackFrame` is not passed here when being invoked from the `DisposeDispatch`.  This is now a check for that case so it skips the processing necessary for a generic call and jumps to Execute.

src/Engine/ProtoCore/DSASM/Heap.cs
saintentropy: cleaned up

sm6srw:
@saintentropy You have 4 test failures now.

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -1528,6 +1528,25 @@ private FunctionGroup FirstFunctionGroupInInheritanceChain(RuntimeCore runtimeCo
...
+            }
+            
+            disposeArguments[0] = stackValue;
+
+            //EXECUTE
+            return finalFep.Execute(null, disposeArguments, null, runtimeCore);
```
aparajit-pratap, line 1548: Why is the context (first arg) passed as null?

saintentropy: Only because for this execution path it was not referenced.

aparajit-pratap (approved):

sm6srw (approved):

PR #11940, opened 2021-08-17
perform static cyclic dependency detection before each graph run

aparajit-pratap:
### Purpose
This adds the capability to perform a static cyclic dependency detection at the Dynamo graph level before each graph run (delta execution). If a cycle is detected, other than reporting node warnings on the nodes that are participating in the cycle, it prevents the graph from executing in the first place and getting into a warning state thus saving on some execution cycles (pun unintended).

TODO:

- [ ] Check if cyclic dep checks can be removed from codegen
- [ ] Check if fixes made in PR #11896 are still necessary
- [ ] Profile existing cyclic dep checks with checks in this PR

### Declarations

### FYIs

### DIFF
--- a/src/DynamoCore/Scheduler/UpdateGraphAsyncTask.cs
+++ b/src/DynamoCore/Scheduler/UpdateGraphAsyncTask.cs
@@ -6,6 +6,7 @@ using ProtoCore.BuildData;
 using ProtoScript.Runners;
 using Dynamo.Graph.Nodes;
 using Dynamo.Graph.Workspaces;
+using DynamoServices;
 
 namespace Dynamo.Scheduler
 {
@@ -36,6 +37,32 @@ namespace Dynamo.Scheduler
             this.verboseLogging = verboseLogging1;
         }
 
+        private bool IsGraphCyclic(NodeModel node, HashSet<NodeModel> visitTracker, 
+            HashSet<NodeModel> recursionTracker, List<NodeModel> cyclicNodes)
+        {
+            if(recursionTracker.Contains(node))
+            {
+                cyclicNodes.Add(node);
+                return true;
+            }
+            if(visitTracker.Contains(node))
+            {
+                return false;
+            }
+            recursionTracker.Add(node);
+            visitTracker.Add(node);
+
+            var outputNodes = node.OutputNodes.Values.SelectMany(set => set.Select(t => t.Item2)).Distinct();
+            foreach (var child in outputNodes)
+            {
+                if (IsGraphCyclic(child, visitTracker, recursionTracker, cyclicNodes))
+                    return true;
+            }
+            recursionTracker.Remove(node);
+
+            return false;
+        }
+
         /// <summary>
         /// This method is called by code that intends to start a graph update.
         /// This method is called on the main thread where node collection in a 
@@ -59,6 +86,29 @@ namespace Dynamo.Scheduler
                 TargetedWorkspace = workspace;
 
                 ModifiedNodes = ComputeModifiedNodes(workspace);
+
+                // Detect cyclic dependencies in graph
+                var cyclicNodes = new List<NodeModel>();
+
+                foreach (var node in ModifiedNodes)
+                {
+                    var visited = new HashSet<NodeModel>();
+                    var recursed = new HashSet<NodeModel>();
+
+                    if(!IsGraphCyclic(node, visited, recursed, cyclicNodes))
+                    {
+                        node.ClearErrorsAndWarnings();
+                    }
+                }
+                if (cyclicNodes.Any())
+                {
+                    foreach (var node in cyclicNodes)
+                    {
+                        // Log cyclic dependency warning on node
+                        node.Warning("This node has a cyclic dependency", true);
+                    }
+                    return false;
+                }
                 graphSyncData = engineController.ComputeSyncData(workspace.Nodes, ModifiedNodes, verboseLogging);
                 if (graphSyncData == null)
                     return false;


src/DynamoCore/Scheduler/UpdateGraphAsyncTask.cs
```diff
@@ -59,6 +94,31 @@ internal bool Initialize(EngineController controller, WorkspaceModel workspace)
                 TargetedWorkspace = workspace;
 
                 ModifiedNodes = ComputeModifiedNodes(workspace);
+
+                // Detect cyclic dependencies in graph
+                var cyclicNodes = new List<NodeModel>();
+
+                foreach (var node in ModifiedNodes)
```
mjkkirschner, line 101: hmm, so this appears that it will start multiple DFS searches from each modified node - on opening a new graph all nodes will be modified so this has the possibility of recomputing the same information multiple times -and could have a performance like `O(n(n+v))` if I've got that right.

can't each invocation of DFS share the visited data structure? If one DFS has already computed this subtree of the graph whats the reason to recompute it?

aparajit-pratap: @mjkkirschner, you're right. I made a mistake earlier. I was initializing the `visited` and `recursion` lists for each modified node. I have now moved their initialization outside the `for` loop so it's done only once. That way if a node is already visited it will be skipped. The overall complexity of this algo is O(v+e). The reason why I'm running this on all nodes as there could be disconnected portions of the graph and I am trying to collect all nodes involved in a cycle and report warnings on those. 

The good thing is this will run on all nodes only the first time a graph is opened and executed but subsequently, it will only run on the modified nodes for each delta execution.

src/DynamoCore/Scheduler/UpdateGraphAsyncTask.cs
```diff
@@ -36,6 +37,40 @@ internal UpdateGraphAsyncTask(IScheduler scheduler, bool verboseLogging1) : base
...
 
+        /// <summary>
+        /// DFS algorithm to check cyclic dependency in Dynamo graph.
+        /// </summary>
+        /// <param name="node"></param>
+        /// <param name="visitTracker"></param>
+        /// <param name="recursionTracker"></param>
+        /// <param name="cyclicNodes">list of all nodes participating in a cycle</param>
+        /// <returns>true if graph is cyclic, false otherwise</returns>
+        private bool IsGraphCyclic(NodeModel node, HashSet<NodeModel> visitTracker,
```
pinzart90, line 48: is there no other similar method already implemented that we could reuse (optionally enhance)?

aparajit-pratap: There is a better algorithm used called `StronglyConnectedComponent` but it is used at a lower, `graphnode` level in the engine. We could probably use it for this too at the dynamo `nodemodel` level but we'll need to generalize (templatize) it to work for any graph vertex, which will be tricky as getting the adjacent nodes for each type is very different. It would be easier to rewrite it to make it work specifically with `NodeModel` type.

aparajit-pratap: My first goal here was to establish that this idea could work in general, we can always refine the exact algo later and make it more efficient, which at the moment, I can think of the `stronglyconnectedcomponent` algo.

pinzart90: Ok
Is this supposed to detect all cycles or just the first one ?

aparajit-pratap: I think this will detect only one cycle for a given node but the outer method call will record at most one cycle for **all** nodes and report warnings on all of them.

src/DynamoCore/Scheduler/UpdateGraphAsyncTask.cs
```diff
@@ -59,6 +94,40 @@ internal bool Initialize(EngineController controller, WorkspaceModel workspace)
...
+                {
+                    if(!IsGraphCyclic(node, visited, recursed, cyclicNodes))
+                    {
+                        var warnings = controller.LiveRunnerCore.BuildStatus.Warnings.ToList();
+                        foreach(var warning in warnings)
+                        {
+                            if(warning.ID == WarningID.InvalidStaticCyclicDependency && 
+                                node.GUID == warning.GraphNodeGuid)
+                            {
+                                node.ClearErrorsAndWarnings();
```
sm6srw, line 113: So these warnings are from a previous execution when the graph was in a cyclic state?

aparajit-pratap: Yes, if a cyclic warning is present for this node from a previous execution and the graph is found to be acyclic for the node in the current execution, the warning will be cleared for that node.

pinzart90: `ClearErrorsAndWarnings` will clear everything (all previous persistent warnings or errors)
Do you think there might be a case where older (persistent) errors should not be cleared ?

aparajit-pratap: Yes, it will clear everything but I'm reducing the surface area of it clearing warnings only when the node in question had a cyclic warning before, but you're right, this will still clear any and all other warnings that the node previously had as well. However, I can't think of a scenario where that would do any harm as I'm relying on the fact that the same warnings should reappear if the node is re-executed.

pinzart90: Most of the internal (not created by API users) are indeed mostly regenerated when re executing. We do have example of non execution related warnings (like the Gizmo warning)
I guess the issue would be if API users create warnings that do not get recreated by re execution
Not a big issue though...warning infrastructure does not support it yet

pinzart90 (approved):
I have not looked at the cycle detection algorithm too much (because we said we're going to look into stronglyconnectedcomponent later on)
Asides from that, LGTM

PR #11941, opened 2021-08-17
Node visual refresh updates

OliverEGreen:
This PR responds to recent comments made on the visual refresh, and addresses the vast majority of the failing regression tests.

The main changes to get tests working was: 

- Updating the PortViewModel.RefreshPortColors method to use hard-coded color values, rather than those from a ResourceDictionary. Conversion using an IValueConverter or IMultiValueConverter, while preferable, was not possible.
- I've added a [RequiresSTA] attribute to the failing OpenXML tests, which fixed them. 

Other changes:
- The location of the Artifakt font and its license files has also been updated.
- The rest of the change are minor styling updates in response to comments received from the team.


src/DynamoCoreWpf/ViewModels/Core/PortViewModel.cs
```diff
@@ -641,19 +641,19 @@ private void RefreshPortColors()
...
-                    PortBorderBrushColor = (SolidColorBrush)SharedDictionaryManager.DynamoColorsAndBrushesDictionary["PortIsConnectedBorderBrush"];
+                    PortBackgroundColor = new SolidColorBrush(Color.FromRgb(70, 90, 99));
+                    PortBorderBrushColor = new SolidColorBrush(Color.FromRgb(106, 192, 231));
                 }
                 else
                 {
-                    PortBackgroundColor = (SolidColorBrush)SharedDictionaryManager.DynamoColorsAndBrushesDictionary["PortDefaultBackground"];
-                    PortBorderBrushColor = (SolidColorBrush)SharedDictionaryManager.DynamoColorsAndBrushesDictionary["PortDefaultBorderBrush"];
+                    PortBackgroundColor = new SolidColorBrush(Colors.Transparent);
+                    PortBorderBrushColor = new SolidColorBrush(Color.FromRgb(204, 204, 204));
```
aparajit-pratap, line 650: Ugh 😨

mjkkirschner: yeah... agreed, we should figure out why this does not work during testing instead of making the code less maintainable.

If you're blocked @OliverEGreen I think @QilongTang can file a task to have the teams investigate - but a minimal repro case will be needed.

QilongTang: Filed as https://jira.autodesk.com/browse/DYN-3919

PR #7623, opened 2017-02-20
Recover from geometry exceptions thrown during manipulator creation

aparajit-pratap:
### Purpose

This fixes:
1. [bug ](https://jira.autodesk.com/browse/DYN-541) when very small values (to the order of -4) are used in the creation of the point manipulator. In such cases the geometry creation for the manipulator gizmo can fail, and an exception can be thrown from LibG that could cause the system to crash. This can be reproduced by setting the Geometry scaling setting to the largest value (since this setting forces all values to be scaled down by a factor of 10,000 under the hood and geometry could fail with such small values). 

The fix is to catch any such exceptions thrown from the geometry library and display a warning on the manipulator node to the effect that direct manipulation has failed due to reasons mentioned in exception messages handed back from LibG.

image

Note that the node in warning state continues to run, just that its manipulator fails to generate. If inputs change and the node re-evaluates so that it can successfully generate the manipulator, the warning disappears as expected. If the node is unselected, then too the warning state disappears.

2. [Bug](https://jira.autodesk.com/browse/DYN-536) when the graph is re-executed upon changing the scale setting. Previously we called the `ForceRunCancel` command to reset the engine and rerun the graph. This caused a crash when running in Revit. The fix is to instead mark all nodes in the graph dirty and force re-execute without resetting the engine. Take a look at `DynamoViewModelChangeScaleFactor` in `DynamoView.xaml.cs`.

### Declarations

Check these if you believe they are true

- [X] The code base is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [X] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [X] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions), and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.

### Reviewers

@Benglin  

### FYIs

@monikaprabhu @riteshchandawar

PR #11910, opened 2021-08-04
Node informational states (#4)

OliverEGreen:
### Purpose

This PR introduces a new design to the InfoBubble UserControl, which is used to report node warnings and errors to the user. 

There are now 3 kinds of user-facing messages a node can display. In order of ascending severity, these are Info, Warning and Message. 

The InfoBubble is now capable of displaying multiple info messages, warnings and errors simultaneously, and gives the user the ability to dismiss both Warnings and Info Messages. Errors cannot be dismissed. 

Dismissed messages may be retrieved by the node's context menu.

The following GIFs demonstrate the new functionality added in this PR:

DismissingMessages

1. User-facing messages may now be dismissed by the user, and undismissed using the node's context menu. When messages are dismissed, a small counter glyph appears at the base of the node whose messages were dismissed. 

StackingMessages

2. Messages may be stacked according to their severity level. Errors (if any) will always appear at the base of the stack, with warnings above and info messages at the highest level. Each level of the stack may be expanded and contracted by clicking on the message body.

The message body may take 1 of 3 states:

- A simple icon, with no messages displayed. This is the most compact representation.
- An icon alongside the first message at the corresponding severity level, e.g. Errors. Where there are multiple messages at this level, an iterating counter appears beside each message.
- Where multiple messages are present at the same severity level, clicking 'show all' will expand all messages at this level. An interating counter will be displayed to help users keep track of the messages.

ScrollingMessages

3. A maximum height has been set per message-level, in case warning messages are very long (such as a long Python traceback message). In such cases, a scrollbar appears to allow for convenient traversal and review of any messages. 

### Impact assessment

- [ ] Changes the way 3rd party nodes work (Add ikeough to the reviewer list)
- [ ] Includes migrations (Add ikeough to the reviewer list)
- [ ] Adds or changes a public interface (Add lukechurch to the reviewer list)
- [ ] API change is added to the [API Changes document](https://github.com/DynamoDS/Dynamo/wiki/API-Changes).
- [ ] Affects which components of the system depends on which other components (Add pboyer to the reviewer list)
- [ ] Introduces new components (Add lukechurch to the reviewer list)
- [ ] Has a performance impact (Add keyu to the reviewer list)
- [ ] Changes the installer (Add sharadkjaiswal to the reviewer list)
- [X] User facing strings, if any, are extracted into `*.resx` files
- [X] All tests pass using the self-service CI.
- [X] Snapshot of UI changes, if any.

#### References (FILL ME IN)

[...]

PR #11947, opened 2021-08-19
Fix nodeManipulator persistent warnings issue

pinzart90:
NodeManipulator can add a persistent warning for the Node it corresponds to.
https://github.com/DynamoDS/Dynamo/pull/7623

In this PR we are enhancing the Warning framework so that we can better manage transient warnings
0. No longer clean up persistent warnings when adding a transient warning.
1. Added 1 new cleanup method: CleanTransientWarning
2. Now we will store the transient warning so that we can later find it and only clean it if it matches some value.
3. Manipulator will only clean the transient warning if it matches the the one it added.

As a side note:
Looks to me like some of the Warning infrastructure behavior is not well defined. I tried to refactor the whole thing...but quickly got at a stage where we would probably need UX input. I have since simplified my changes to the current state.

### DIFF
--- a/src/AssemblySharedInfoGenerator/AssemblySharedInfo.cs
+++ b/src/AssemblySharedInfoGenerator/AssemblySharedInfo.cs
@@ -45,7 +45,7 @@ using System.Runtime.InteropServices;
 // to distinguish one build from another. AssemblyFileVersion is specified
 // in AssemblyVersionInfo.cs so that it can be easily incremented by the
 // automated build process.
-[assembly: AssemblyVersion("2.13.0.2305")]
+[assembly: AssemblyVersion("2.13.0.2104")]
 
 
 // By default, the "Product version" shown in the file properties window is
@@ -64,4 +64,4 @@ using System.Runtime.InteropServices;
 // You can specify all the values or you can default the Build and Revision Numbers 
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyFileVersion("2.13.0.2305")]
+[assembly: AssemblyFileVersion("2.13.0.2104")]
--- a/src/AssemblySharedInfoGenerator/AssemblySharedInfo.cs
+++ b/src/AssemblySharedInfoGenerator/AssemblySharedInfo.cs
@@ -45,7 +45,7 @@ using System.Runtime.InteropServices;
 // to distinguish one build from another. AssemblyFileVersion is specified
 // in AssemblyVersionInfo.cs so that it can be easily incremented by the
 // automated build process.
-[assembly: AssemblyVersion("2.13.0.2104")]
+[assembly: AssemblyVersion("2.13.0.2305")]
 
 
 // By default, the "Product version" shown in the file properties window is
@@ -64,4 +64,4 @@ using System.Runtime.InteropServices;
 // You can specify all the values or you can default the Build and Revision Numbers 
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyFileVersion("2.13.0.2104")]
+[assembly: AssemblyFileVersion("2.13.0.2305")]
--- a/src/DynamoCore/Graph/Nodes/NodeModel.cs
+++ b/src/DynamoCore/Graph/Nodes/NodeModel.cs
@@ -1595,6 +1595,7 @@ namespace Dynamo.Graph.Nodes
                     if (string.Compare(t, transientWarning) == 0)
                     {
                         transientWarning = string.Empty;
+                        SetNodeStateBasedOnConnectionAndDefaults();
                         ToolTipText = persistentWarning;
                     }
                     // if no match is found, we do nothing.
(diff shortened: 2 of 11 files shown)


src/DynamoManipulation/NodeManipulator.cs
```diff
@@ -547,7 +548,11 @@ public void Dispose()
         {
             Dispose(true);
 
-            Node.ClearErrorsAndWarnings();
+            if (!string.IsNullOrEmpty(persistentWarning))
```
mjkkirschner, line 551: do you think this deserves a comment why the disposal of a manipulator is clearing warnings?

mjkkirschner:
I doubt they intersect, but might be worth looking at https://github.com/DynamoDS/Dynamo/pull/11910

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -1561,24 +1565,51 @@ private void ClearTooltipText()
...
+                SetNodeStateBasedOnConnectionAndDefaults();
+
+                if (State == ElementState.PersistentWarning)
+                {
+                    ToolTipText = persistentWarning;
+                }
+                else
+                {
+                    ClearTooltipText();
+                }
```
aparajit-pratap, line 1609: Why have you added all of this new logic?

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -1667,17 +1698,17 @@ public void Warning(string p, bool isPersistent = false)
             if (isPersistent)
             {
                 State = ElementState.PersistentWarning;
-                if (!string.Equals(persistentWarning, p))
+                if (!persistentWarnings.Contains(p))
```
aparajit-pratap, line 1701: Since `persistentWarnings` is a hashset, this check may not be required, right?

pinzart90: uhmm..I think you are right about that 
THanks

src/DynamoCore/Graph/Nodes/NodeModel.cs
pinzart90: Because the NodeManipulator called ClearErrorsAndWarnings() before these changes.
ClearErrorsAndWarnings had a similar logic.

What I wanted to do is:
1. More than 1 persistent warning exists (i.e after clear is called other warnings remain)
    Keep the same state as before...only update the persistentWarning string and the Tooltip string
2. Single persistent warning (i.e after clear is called no other warnings remain)
   Try to find a valid state in which to put the Node's warning system.

aparajit-pratap: But all existing `ClearPersistentWarning` calls have _also_ been replaced with this new method and the old method didn't have this logic.

pinzart90: Yep, a lot of redundant work.
I'll try a refactor

aparajit-pratap (approved):

pinzart90:
Still testing this. Will try to figure out a test for this...

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -1667,17 +1722,14 @@ public void Warning(string p, bool isPersistent = false)
...
-                    persistentWarning += p;
-                }
+                persistentWarnings.Add(p);
                 ToolTipText = persistentWarning;
             }
             else
             {
                 State = ElementState.Warning;
-                ToolTipText = string.IsNullOrEmpty(persistentWarning) ? p : string.Format("{0}\n{1}", persistentWarning, p);
-                ClearPersistentWarning();
```
pinzart90, line 1680: This would be the biggest change in the Warning logic.
We will not clear persisting warnings when adding a transient warning. (not sure why we ever did).

test/DynamoCoreTests/SerializationTests.cs
```diff
@@ -1014,7 +1014,7 @@ public void SerializationNonGuidIdsTest(string filePath)
                                     "of the graph will not execute; skipping test ...");
             }
 
-            if (((HomeWorkspaceModel)ws1).RunSettings.RunType == Models.RunType.Manual)
+            if (((HomeWorkspaceModel)ws1).RunSettings.RunType == Dynamo.Models.RunType.Manual)
```
pinzart90, line 1017: build fails without this

src/AssemblySharedInfoGenerator/AssemblySharedInfo.cs
```diff
@@ -45,7 +45,7 @@
 // to distinguish one build from another. AssemblyFileVersion is specified
 // in AssemblyVersionInfo.cs so that it can be easily incremented by the
 // automated build process.
-[assembly: AssemblyVersion("2.13.0.2104")]
+[assembly: AssemblyVersion("2.13.0.2305")]
```
sm6srw, line 48: Correct me if I am wrong but I don't think we normally submit changes to these version numbers in a typical PR.

sm6srw (approved):
LGTM!

src/AssemblySharedInfoGenerator/AssemblySharedInfo.cs
pinzart90: I do not think anyone is actually using the build number for anything.
It gets incremented on every Build automatically and I sometimes commit it by mistake.

I reverted it.

PR #11725, opened 2021-06-02
Node markdown generator tool

SHKnudsen:
### Purpose

This pr adds a new NodeDocumentationGenerator tool to automatically generate node documentation markdown files, the tool creates markdown files which (if specified) will have content from the DynamoDictionary added. The tool has two different verbs one that creates documentation from dlls in a directory, this can be used to create docs for Dynamo Core and Dynamo hosts, and another that generates docs from a package.

The PR also adds new APIs to `ProtoCore.ProtoFFI.CLRDLLModule` to enable scanning dlls in reflection mode, along with this new reflection extensions have been added to `ProtoCore.Reflection`.

The tool does not require Dynamo to be running but it uses the same APIs to scan dlls, main difference is that it dosent load anything into Dynamo but simply gives us back a list of all node types in the assembly being scanned

Dynamo loads documentation from Core and hosts by looking in a `fallback_docs` folder in the root directory of core or the host (TBD). By default when `help` is pressed on a node the `DocumentationBrowser` will look for the help doc associated with that node in the `PackageDocumentationManager` if no match is found it will first try and find a match in the host `fallback_docs` and secondly in Cores `fallback_docs`.

Besides that i made a slight modification to how the `PackageDocumentationManager` loads doc files, essentially i came across an issue where two packages was using the same namespace, this will cause an error on the manager. To address this i added the package name to the key of `nodeDocumentationFileLookup`.

#### Verbs
| Verb            | Description                                                                                                                                                                                                 |
| --------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `fromdirectory` | Generates documentation for all dlls (or only those specified in the filter) and all dyfs (if specified by the  `--includedyfs` flag) in the input directory                                |
| `frompackage`   | Looks up node libraries in the pkg.json file and creates documentation for those dlls and creates documentation for all dyfs in the `/dyf` folder. All generated files gets saved in `path/to/package/docs` |

#### fromdirectory - flags
| Short name | Long name           | Optional | Description                                                                            |
| ---------- | ------------------- | :------: | -------------------------------------------------------------------------------------- |
| `-i`       | `--input`           |          | Input binary file, containing nodes that documentaion should be generated for.         |
| `-o`       | `--output`          |          | Location where markdown files should be generated                                      |
| `-l`       | `--logger-path`     |          | Directory path to where logger file should be saved                                    |
[...]

PR #11950, opened 2021-08-19
node docs generator tool without MetaDataLoadContext

mjkkirschner:
### Purpose

Please see @SHKnudsen's PR for a great summary:
https://github.com/DynamoDS/Dynamo/pull/11725

differences between the PRs:

* `MetaDataLoadContext` support has been removed for the current time, and we'll explore its necessity when we deploy this tool. This removes ~ 700 LOC.
    * This means that in general full dependencies are required (ie all managed Revit API dlls)    
    * As a consequence the tool now generates 100 additional nodes it was missing before - see the readme for details.
* removes MetaDataLoadContext nuget packages.
* animated gif compression has been added. `-g`
* adds verbose `-v` mode-  I found helpful to track down some issues.
* adds ImageMagick .net and native to the license and about box.
* adds a readme - mostly stolen from @SHKnudsen's summary, but also with some of our use cases described - hopefully to help devs understand what aspects are important to maintain.
* moves layout spec with a type forward attribute so the exe no longer depends on DynamoCoreWPF.

TODO:
- [ ]  after this is merged, I will file the known issues as tasks / spikes for us to pick up later.
- [ ] after this is merged I will work with Sylvester on this PR: https://github.com/SHKnudsen/Dynamo/pull/34 to add unit and integration tests.

### Declarations


src/DocumentationBrowserViewExtension/PackageDocumentationManager.cs
```diff
@@ -39,17 +46,64 @@ public static PackageDocumentationManager Instance
...
+
+            if (!string.IsNullOrEmpty(pathManager.DynamoCoreDirectory))
+            {
+                var coreDir = new DirectoryInfo(Path.Combine(pathManager.DynamoCoreDirectory, FALLBACK_DOC_DIRECTORY_NAME));
+                dynamoCoreFallbackDocPath = coreDir.Exists ? coreDir : null;
+            }
+
+            if (!string.IsNullOrEmpty(pathManager.HostApplicationDirectory))
+            {
+                var hostDir = new DirectoryInfo(Path.Combine(pathManager.HostApplicationDirectory, FALLBACK_DOC_DIRECTORY_NAME));
```
mjkkirschner, line 70: @sm6srw to answer your question about how useful this is - it actually uses the `DynamoHostPath` property as the root - which can be set using a startupconfiguration - for example, DynamoRevit sets this to the location of the `DynamoRevit.dll` entry point.

src/Tools/NodeDocumentationMarkdownGenerator/readme.md
```diff
@@ -0,0 +1,107 @@
...
+There are two different high level commands:
+1. Create Documentation From Directory
+2. Create Documentation From Package
+## Use Cases
+
+1. One time dynamo dictionary migration to .md files. The resulting files will be stored in the Dynamo and host specific repos. (DynamoRevit)
+
+2. Continual creation of new docs stubs for use on dictionary and in Dynamo docs browser. DynamoDictionary site will consume .md files.
+
+3. user by package authors to stub out documentation for nodes.
```
sm6srw, line 15: user?

sm6srw (commented):
I have made a first pass but I need to come back and take another pass on all new files. Looks good. Just a few questions.

src/DocumentationBrowserViewExtension/PackageDocumentationManager.cs
```diff
@@ -39,17 +46,64 @@ public static PackageDocumentationManager Instance
...
+
         private PackageDocumentationManager() { }
 
         /// <summary>
         /// Retrieves the markdown node documentation file associated with the input node namespace
         /// </summary>
         /// <param name="nodeNamespace">Namespace of the node to lookup documentation for</param>
         /// <returns></returns>
-        public string GetAnnotationDoc(string nodeNamespace)
+        public string GetAnnotationDoc(string nodeNamespace, string packageName)
```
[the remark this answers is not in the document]
mjkkirschner: yes @sm6srw, I had the same reaction initially, https://github.com/DynamoDS/Dynamo/pull/11725/files#r646888243 - but it's actually inside an extension, I think it's okay to break this with chance of risk and if possible it would be nice to codify extensions being part of the API boundary or not but it will take a cross team effort / discussion I think.

in this case I'm okay with the change unless there are objections.

src/DynamoCoreWpf/Interfaces/ILibraryViewCustomization.cs
```diff
@@ -1,4 +1,5 @@
-﻿using System;
```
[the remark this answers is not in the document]
mjkkirschner: ah ha! good catch, actually I think I should apply the a type forward to this interface as well, move them all to DynamoCore now as a single unit - theres no WPF dependencies in this interface AFAIK.

then, in 3.0 we can get rid of the type forward. Sound good?

src/DynamoCore/Interfaces/LayoutSpecification.cs
```diff
@@ -4,7 +4,7 @@
 using Newtonsoft.Json;
 using Newtonsoft.Json.Converters;
 
-namespace Dynamo.Wpf.Interfaces
+namespace Dynamo.Interfaces
```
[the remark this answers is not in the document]
mjkkirschner: responded below.

src/DynamoCoreWpf/Properties/AssemblyInfo.cs
```diff
@@ -44,3 +44,9 @@
 [assembly: InternalsVisibleTo("PythonNodeModelsWpf")]
 [assembly: InternalsVisibleTo("LintingViewExtension")]
 [assembly: InternalsVisibleTo("GraphMetadataViewExtension")]
+
+[assembly: TypeForwardedTo(typeof(Dynamo.Interfaces.LayoutSpecification))]
```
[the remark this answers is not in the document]
mjkkirschner: yes, responded above.

src/Tools/NodeDocumentationMarkdownGenerator/readme.md
mjkkirschner: should be `used` - will fix.

src/DynamoCoreWpf/Interfaces/ILibraryViewCustomization.cs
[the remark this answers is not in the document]
sm6srw: Yep, make much more sense.

pinzart90:
so the what was the reason for the move from Wpf to Core ?
```
   <Compile Include="Interfaces\ILibraryViewCustomization.cs" />
    <Compile Include="Interfaces\LayoutSpecification.cs" />
    ```
    
 Was it because new projects that needed those interfaces had to depend on the dyn*wpf assembly ?
 And it was really no need to, since all the data in those 2 interfaces is pretty bare metal (no wpf) ?
 
 Also what is the main purpose of those interfaces ? TO provide a simple and uniform way for CustomViews to arrange their data  ?

File: src/Tools/NodeDocumentationMarkdownGenerator/Commands/FromDirectoryCommand.cs [hunk 7]
```diff
@@ -0,0 +1,92 @@
...
+                if (fileInfo is null) continue;
+
+                fileInfos.Add(fileInfo);
+            }
+            return fileInfos;
+        }
+
+        private static List<MdFileInfo> ScanFolderForAssemblies(string inputFolderPath, IEnumerable<string> filter, IEnumerable<string> referencePaths, SearchOption searchOption)
+        {
+            var allAssembliesFromInputFolder = Directory.GetFiles(inputFolderPath, "*.*", searchOption)
```
pinzart90, line 45: Not sure how many filles we expect to parse here.
`EnumerateFiles` might have some benefits...
`
The EnumerateFiles and GetFiles methods differ as follows: When you use EnumerateFiles, you can start enumerating the collection of names before the whole collection is returned; when you use GetFiles, you must wait for the whole array of names to be returned before you can access the array. Therefore, when you are working with many files and directories, EnumerateFiles can be more efficient.
`

src/Tools/NodeDocumentationMarkdownGenerator/Commands/FromDirectoryCommand.cs
```diff
@@ -0,0 +1,92 @@
...
+                .Where(x => x.EndsWith(".dll") || x.EndsWith(".ds"))
+                .Select(x => new FileInfo(x))
+                .GroupBy(x => x.Name)
+                .Select(x => x.FirstOrDefault())
+                .ToList();
```
pinzart90, line 50: Any reason to force all the IEnumerables into Lists this early ?

src/Tools/NodeDocumentationMarkdownGenerator/DynamoDictionaryEntry.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    /// <summary>
+    /// Class used to serialize an entry in the Dynamo Dictionary
+    /// </summary>
+    internal class DynamoDictionaryEntry
+    {
+        [JsonProperty("Name")]
+        public string Name { get; set; }
+
+        [JsonProperty("imageFile")]
```
pinzart90, line 15: Why LowerCase first letter ?
Does this need to match with some other properties ?

src/Tools/NodeDocumentationMarkdownGenerator/NodeDocumentationMarkdownGenerator.csproj
```diff
@@ -0,0 +1,154 @@
...
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{0DD3E4ED-66AA-47FC-9EB9-2A0E6A0EA01B}</ProjectGuid>
+    <OutputType>Exe</OutputType>
+    <RootNamespace>NodeDocumentationMarkdownGenerator</RootNamespace>
+    <AssemblyName>NodeDocumentationMarkdownGenerator</AssemblyName>
+    <TargetFrameworkVersion>v4.8</TargetFrameworkVersion>
```
pinzart90, line 11: Looking at the PR and at the Readme (`The NodeDocumentationGenerator is a CLI tool to generate node documentation markdown stubs.`) it seems that there is no UI being done in this project.
What are the reasons for making this project use the full dotnet framework (instead of core or dotnet standard)

pinzart90:
How are we going to test the new functionality ? Unit tests (with mocked input files....?) ?

src/Tools/NodeDocumentationMarkdownGenerator/DynamoDictionaryEntry.cs
mjkkirschner: yes, this needs to match the already existing dynamo dictionary json files.

src/Tools/NodeDocumentationMarkdownGenerator/NodeDocumentationMarkdownGenerator.csproj
mjkkirschner: It's not clear to me that it's so simple to have a .net standard library depend on assemblies targeting .net framework (they are different targets after all) and this PR is big enough dissuade me from trying to also include multi-targeting our core dynamo assemblies for .net standard.

If I can simply change the target to .net standard and everything works I'm fine with that 😉  - otherwise I think it's probably best to include that improvement later.

mjkkirschner:
> so the what was the reason for the move from Wpf to Core ?
> 
> ```
>    <Compile Include="Interfaces\ILibraryViewCustomization.cs" />
>     <Compile Include="Interfaces\LayoutSpecification.cs" />
>     ```
>     
>  Was it because new projects that needed those interfaces had to depend on the dyn*wpf assembly ?
>  And it was really no need to, since all the data in those 2 interfaces is pretty bare metal (no wpf) ?
>  
>  Also what is the main purpose of those interfaces ? TO provide a simple and uniform way for CustomViews to arrange their data  ?
> ```

I moved them because I wanted to 
1. avoid referencing DynamoCoreWpf in this tool to optionally move it to mono,dotstandard etc.
2.The types incldue no references to WPF so I changed the namespace to reflect what they depend on.

src/Tools/NodeDocumentationMarkdownGenerator/NodeDocumentationMarkdownGenerator.csproj
mjkkirschner: from stack overflow
```
No, .NET Standard projects cannot reference framework projects.

.NET Standard projects need to be usable across platforms, forcing a dependency on the .NET framework by referencing an assembly targeting it makes this impossible.

Note that with some of the magic Microsoft is doing with .NET Standard 2.0 this is less true but the overall idea still stands.
```

mjkkirschner: this answer may be old, so I'll try it.

mjkkirschner:
> How are we going to test the new functionality ? Unit tests (with mocked input files....?) ?

I am working on a test PR here:
https://github.com/mjkkirschner/Dynamo/commits/nmgtunittestsmjk
based on Sylvester's work

src/Tools/NodeDocumentationMarkdownGenerator/NodeDocumentationMarkdownGenerator.csproj
pinzart90: Yeah..dotnet standard probably cannot reference dotnet framework....
Unless we take on the gigantic task of braking apart the dependencies into standard vs non standard chunks
Fine with this even if it stays as it is

src/Tools/NodeDocumentationMarkdownGenerator/Commands/FromDirectoryCommand.cs
mjkkirschner: I'd rather stick with the simplicity of `GetFiles()` - this is not going to affect users, and image compression time completely dwarfs the time it takes to read these files.

I'm thinking back to the last time I ran into trouble groking deferred execution and `EnumerateFiles`.

src/Tools/NodeDocumentationMarkdownGenerator/Commands/FromDirectoryCommand.cs
mjkkirschner: let me look at this, it does seem a bit odd since we can just do it once at the end of the method.

pinzart90 (approved):
LGTM

sm6srw (approved):

PR #11958, opened 2021-08-23
AGD-2199 - custom dropdown node

LongNguyenP:
### Purpose

 This PR addresses JIRA task [AGD-2199](https://jira.autodesk.com/browse/AGD-2199): A node that allows user to pre-define a dropdown menu with an arbitrary number of customizable menu items

image

### Declarations

### Reviewers

### FYIs


src/Libraries/CoreNodeModels/Properties/Resources.en-US.resx
```diff
@@ -617,4 +617,10 @@ Default value: {0}</value>
   <data name="IFNodeWarningMessage" xml:space="preserve">
     <value>This node has been updated and will be removed in a future version of Dynamo. Existing behavior is retained, but a new version now supports Empty Lists, Null values and inputs of varying length. Please replace this node if you wish to use this improved behavior.</value>
   </data>
+  <data name="CustomSelectionSearchTags" xml:space="preserve">
+      <value>[Tags Missing]</value>
+  </data>
+  <data name="CustomSelectionNodeDescription" xml:space="preserve">
+      <value>[Description Missing]</value>
```
saintentropy, line 624: Maybe dropdown;custom;custom dropdown;enum;custom enum

QilongTang:
@LongNguyenP @saintentropy There is a regression `NodeDocumentationMarkdownGeneratorTests.MarkdownGeneratorCommandTests.ProducesCorrectOutputFromCoreDirectory`

LongNguyenP:
Merge conflicts with the current master has been solved. LGTM

src/Libraries/CoreNodeModels/CustomSelectionItem.cs
```diff
@@ -0,0 +1,165 @@
...
+using Newtonsoft.Json;
+
+namespace CoreNodeModels
+{
+    /// <summary>
+    /// Represents an Enum item
+    /// </summary>
+    public class CustomSelectionItem
+    {
+        public string Name { get; set; }
```
mjkkirschner, line 13: public properties should have `///` summary tags. These will generate compiler warnings otherwise. Also part of our standard https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards - mostly just follow MSDN standards for c#.

src/Libraries/CoreNodeModels/CustomSelectionItem.cs
```diff
@@ -0,0 +1,165 @@
...
+using System.ComponentModel;
+using System.Windows.Input;
+using Newtonsoft.Json;
+
+namespace CoreNodeModels
+{
+    /// <summary>
+    /// Represents an Enum item
+    /// </summary>
+    public class CustomSelectionItem
```
mjkkirschner, line 11: why is this public? when possible we should use the *most* conservative access modifier possible to avoid exposing types and methods and having to maintain them for years.

src/Libraries/CoreNodeModels/CustomSelectionItem.cs
```diff
@@ -0,0 +1,165 @@
...
+    {
+        public string Name { get; set; }
+
+        public string Value { get; set; }
+
+        public CustomSelectionItem()
```
mjkkirschner, line 17: why do you need a parameterless constructor? If there is a reason add a comment.

src/Libraries/CoreNodeModels/CustomSelectionItem.cs
```diff
@@ -0,0 +1,165 @@
...
+        {
+
+        }
+    }
+
+
+    public class CustomSelectionItemViewModel : INotifyPropertyChanged, IDataErrorInfo
```
mjkkirschner, line 24: I'm not sure this is the right place for this class, what about in CoreNodeModels WPF? Though I realize it does not have any WPF deps... judgement call.

src/Libraries/CoreNodeModels/CustomSelectionItem.cs
```diff
@@ -0,0 +1,165 @@
...
+        {
+            get
+            {
+                return customSelectionItem.Name;
+            }
+            set
+            {
+                customSelectionItem.Name = value;
+
+                OnPropertyChanged("Name");
```
mjkkirschner, line 51: use `nameof`

src/Libraries/CoreNodeModels/CustomSelectionItem.cs
```diff
@@ -0,0 +1,165 @@
...
+
+    public class CustomSelectionItemViewModel : INotifyPropertyChanged, IDataErrorInfo
+    {
+        private readonly CustomSelectionItem customSelectionItem;
+
+        public event PropertyChangedEventHandler PropertyChanged;
+
+        #region Properties
+
+        [JsonIgnore]
```
mjkkirschner, line 32: is this type serialized somewhere? That is surprising to me - and if it's not serialized, why do you need the the json attributes?

src/Libraries/CoreNodeModels/CustomSelectionItem.cs
```diff
@@ -0,0 +1,165 @@
...
+        public Func<CustomSelectionItem, bool> IsUnique { get; set; }
+
```
mjkkirschner, line 34: definitely document these public properties, and try to reduce their access if possible.

src/Libraries/CoreNodeModels/CustomSelectionItem.cs
```diff
@@ -0,0 +1,165 @@
...
+        [JsonIgnore]
+        public string this[string columnName]
+        {
+            get
+            {
+                if (columnName.Equals("Name"))
+                {
+                    if (string.IsNullOrWhiteSpace(Name))
+                        return "Name cannot be empty";
```
mjkkirschner, line 111: this text needs to be stored in a resx so it can be localized. See examples in other nodes like:
https://github.com/DynamoDS/Dynamo/blob/102bd97db0ccc33ada9ff346178ae6da6de34607/src/Libraries/UnitsUI/UnitsUI.cs#L128

src/Libraries/CoreNodeModels/Input/CustomSelection.cs
```diff
@@ -0,0 +1,291 @@
...
+
+
+namespace CoreNodeModels.Input
+{
+    [NodeName("Custom Dropdown")]
+    [NodeCategory(BuiltinNodeCategories.CORE_INPUT)]
+    [NodeDescription("CustomSelectionNodeDescription", typeof(Properties.Resources))]
+    [NodeSearchTags("CustomSelectionSearchTags", typeof(Properties.Resources))]
+    [OutPortNames("value")]
+    [OutPortTypes("object")]
```
mjkkirschner, line 23: I think `var` or `var[]..[]` is more appropriate depending on what this node does.

src/Libraries/CoreNodeModels/CoreNodeModels.csproj
```diff
@@ -1,105 +1,105 @@
...
-    </GetReferenceAssemblyPaths>
-    <!-- Get assembly -->
-    <GetAssemblyIdentity AssemblyFiles="$(OutDir)$(TargetName).dll">
-      <Output TaskParameter="Assemblies" ItemName="AssemblyInfo" />
-    </GetAssemblyIdentity>
-    <!-- Generate customization dll -->
-    <GenerateResource SdkToolsPath="$(TargetFrameworkSDKToolsDirectory)" UseSourcePath="true" Sources="$(ProjectDir)CoreNodeModelsImages.resx" OutputResources="$(ProjectDir)CoreNodeModelsImages.resources" References="$(FrameworkAssembliesPath)System.Drawing.dll" />
-    <AL SdkToolsPath="$(TargetFrameworkSDKToolsDirectory)" TargetType="library" EmbedResources="$(ProjectDir)CoreNodeModelsImages.resources" OutputAssembly="$(OutDir)CoreNodeModels.customization.dll" Version="%(AssemblyInfo.Version)" />
-   </Target>
+    <ImportGroup Label="PropertySheets">
```
mjkkirschner, line 2: it would be good if you could revert these white space changes, they make reviewing the history more difficult.

src/Libraries/CoreNodeModels/RelayCommand.cs
```diff
@@ -0,0 +1,33 @@
+using System;
+using System.Windows.Input;
+
+namespace CoreNodeModels
+{
+    public class RelayCommand : ICommand
```
mjkkirschner, line 6: ///summary - I'll not repeat this comment everywhere, but please address.

src/Libraries/CoreNodeModels/RelayCommand.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        private readonly Action<object> execute;
+        private readonly Func<object, bool> canExecute;
+
+        public event EventHandler CanExecuteChanged;
+        // {
```
mjkkirschner, line 12: what is this?

src/Libraries/CoreNodeModelsWpf/NodeViewCustomizations/CustomSelectionNodeModelNodeViewCustomization.cs
```diff
@@ -0,0 +1,35 @@
...
+        public void CustomizeView(CustomSelectionNodeModel model, NodeView nodeView)
+        {
+            nodeView.inputGrid.Children.Add(
+                new CustomSelectionControl()
+                {
+                    DataContext = model
+                });
+        }
+
+        public void Dispose()
```
mjkkirschner, line 24: this standard pattern for disposal does not really make sense when we're not actually cleaning up any native resources. Theres no reason to suppress the GC for a fully managed type. I'd just leave it empty for now.

test/Tools/docGeneratorTestFiles/sampledictionarycontent/Dynamo_Nodes_Documentation.json
```diff
@@ -2992,6 +2992,12 @@
     "dynFile": ["Output"],
     "folderPath": "Input/Basic/Action",
     "inDepth": "Output will create an output port for a custom node. The syntax for an output node is simply its name, along with an optional custom comment. In the example below, an output named Percentage is created with a custom comment. It is important to note that this node is only available while creating a custom node."
+}, {
+    "Name": "Custom Dropdown",
+    "imageFile": ["Custom Dropdown"],
+    "dynFile": ["Custom Dropdown"],
+    "folderPath": "Input/Basic/Action",
+    "inDepth": "This node allows you to pre-define a dropdown menu with with an arbitrary number of custom items. Each item consists of a label and a value. The label is always a string, while the value can be integer, number of string"
```
mjkkirschner, line 3000: type (`or` string?) - Does this allow return types to be arbitrary objects or only int,double,string?

mjkkirschner:
Hi Long, I added some change requests, in addition please also the following:
1. tests for the new node and new classes if possible. 
2. add an image or gif of the node in action to this PR

mjkkirschner (changes requested):
see comments inline.

mjkkirschner:
Also curious, @Jingyi-Wen @LongNguyenP Is there a new icon planned for this node?

src/Libraries/CoreNodeModels/CustomSelectionItem.cs
LongNguyenP: OnPropertyChanged(nameof(customSelectionItem.Name))

You mean this?

mjkkirschner: yes, leverage the compiler to avoid typos and name changes.

src/Libraries/CoreNodeModels/RelayCommand.cs
LongNguyenP: I have removed RelayCommand and created AddMenuItemCommand and RemoveMenuItemCommand classes explicitly. These are nested inside CustomSelectionNodeModel and CustomSelectionItemViewModel classes respectively to limit their usage scope

src/Libraries/CoreNodeModels/CustomSelectionItem.cs
LongNguyenP: This assessor and other things related to IDataErrorInfo are not really necessary and will be removed

src/.idea/.idea.Dynamo.All/.idea/.gitignore
```diff
@@ -0,0 +1,13 @@
+# Default ignored files
+/shelf/
```
mjkkirschner, line 2: @LongNguyenP please revert these committed files in the ./idea folder.

src/Libraries/CoreNodeModelsWpf/Controls/CustomSelectionControl.xaml
```diff
@@ -0,0 +1,170 @@
...
+                <Setter Property="Source">
+                    <Setter.Value>
+                        <DrawingImage>
+                            <DrawingImage.Drawing>
+                                <GeometryDrawing Brush="{Binding ElementName=RemoveBtn, Path=Foreground}">
+                                    <GeometryDrawing.Pen>
+                                        <Pen Brush="{Binding ElementName=RemoveBtn, Path=Foreground}" Thickness="5"/>
+                                    </GeometryDrawing.Pen>
+                                    <GeometryDrawing.Geometry>
+                                        <PathGeometry>
```
mjkkirschner, line 79: does this align with other "remove"/"cancel" buttons in Dynamo - I would think this should look like the `remove` button on variable input nodes.?
FYI @Jingyi-Wen 
- I would also be a bit curious about any performance impact of drawing all these paths, if for example a graph contains dozens of these nodes with many entries each - maybe we can just replace this with an image which will only need to be loaded once - we have lots of "x" images floating around.

mjkkirschner: ⚠️ I also believe these styles should be moved the shared resource dictionary, embedding styles within the resource entry of a particular view means that for every view you're going to be instantiating this style again and again, increasing memory use for each node view, if instead you put this in the shared dictionary, you avoid that cost - please move it.

you are already importing one of the shared dictionaries above.

see:
https://docs.microsoft.com/en-us/windows/uwp/debug-test-perf/optimize-xaml-loading#resourcedictionary-in-a-usercontrol

src/Libraries/CoreNodeModelsWpf/Controls/CustomSelectionControl.xaml.cs
```diff
@@ -0,0 +1,74 @@
...
+            catch (Exception ex)
+            {
+                var viewModel = DataContext as CustomSelectionNodeModel;
+                if (viewModel != null)
+                    viewModel.Log(ex);
+            }
+        }
+
+
+        private static void CollectChildrenOfType<T>(DependencyObject dependencyObj, List<T> children) where T : DependencyObject
```
mjkkirschner, line 57: this helper already exists I believe:
https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/Utilities/WpfUtilities.cs#L62

please don't duplicate it if possible to avoid... but can you explain why you need to do this, it's going to be slow, and it's brittle if the xaml tree is changed by another dev in the future.

mjkkirschner (changes requested):
Added some more comments, please address when you can.
Let's talk if you need some examples for tests.

Jingyi-Wen:
Two ideas for the icon (grey background is just for reviewing visibility)
custom dropdown icon ideas

src/Libraries/CoreNodeModelsWpf/Controls/CustomSelectionControl.xaml
mjkkirschner: the relevant quote is:
```
A ResourceDictionary defined inside of a [UserControl](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.usercontrol) carries a penalty. The platform creates a copy of such a ResourceDictionary for every instance of the UserControl. If you have a UserControl that is used a lot, then move the ResourceDictionary out of the UserControl and put it at the page level.
```

src/Libraries/CoreNodeModels/Input/CustomSelection.cs
```diff
@@ -0,0 +1,313 @@
...
+using System.Windows.Input;
+using System.Xml;
+
+
+namespace CoreNodeModels.Input
+{
+    /// <summary>
+    /// This node allow the user to create a dropdown menu with an with an arbitrary number of customization items
+    /// </summary>
+    [NodeName("Custom Dropdown Menu")]
```
mjkkirschner, line 19: @LongNguyenP @saintentropy - not sure if this is worth doing, but perhaps this node should actually serialize as the `cogs` `Selection` input type by overriding `InputData` ?

LongNguyenP:
@mjkkirschner @saintentropy Michael. Craig and I are thinking of making another PR for the tests so that this PR can go ahead without getting too big. Would this be a reasonable way forward?

LongNguyenP:
@mjkkirschner @saintentropy. I fixed a regression bug, added a couple of tests to ensure the saved menu items and selected items are saved and opened correctly from the dyn file. Also change quite a few public methods and to private to be as conservative as possible.

LongNguyenP:
Add @BogdanZavu as an reviewer

mjkkirschner:
@LongNguyenP will review when I am able - but there are 3 test failures:
https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/5401/

src/Libraries/CoreNodeModelsWpf/Controls/CustomSelectionControl.xaml
saintentropy: @mjkkirschner Also going to cover in follow up task

saintentropy:
@mjkkirschner I think this is ready to go :-) with the follow tasks mentioned

mjkkirschner:
@saintentropy @LongNguyenP can you link the jira tickets of the followup tasks?

src/AssemblySharedInfoGenerator/AssemblySharedInfo.cs
```diff
@@ -45,7 +45,7 @@
 // to distinguish one build from another. AssemblyFileVersion is specified
 // in AssemblyVersionInfo.cs so that it can be easily incremented by the
 // automated build process.
-[assembly: AssemblyVersion("2.14.0.3986")]
+[assembly: AssemblyVersion("2.14.0.4484")]
```
mjkkirschner, line 48: in general this file should not be commited.

src/Libraries/CoreNodeModels/CoreNodeModels.csproj
```diff
@@ -1,105 +1,108 @@
...
+            <Generator>PublicResXFileCodeGenerator</Generator>
+            <LastGenOutput>Resources.Designer.cs</LastGenOutput>
+        </EmbeddedResource>
+        <EmbeddedResource Update="Properties\Resources.en-US.resx">
+            <Generator>PublicResXFileCodeGenerator</Generator>
+            <LastGenOutput>Resources.en-US.Designer.cs</LastGenOutput>
+        </EmbeddedResource>
+    </ItemGroup>
+    <ItemGroup>
+      <Folder Include="Resources\" />
```
mjkkirschner, line 90: @LongNguyenP what is this change for?

src/Libraries/CoreNodeModels/Input/CustomSelection.cs
```diff
@@ -0,0 +1,337 @@
...
+
+            return new List<AssociativeNode> { AstFactory.BuildAssignment(GetAstIdentifierForOutputIndex(0), associativeNode) };
+        }
+
+        /// <summary>
+        /// Log an exception
+        /// </summary>
+        /// <param name="ex"></param>
+        [IsVisibleInDynamoLibrary(false)]
+        public void Log(Exception ex)
```
mjkkirschner, line 140: why is this public?

mjkkirschner (approved):
it looks okay to me, a few question remaining though - and please link follow up tasks to this PR.

LongNguyenP (commented):
Response

src/Libraries/CoreNodeModels/CoreNodeModels.csproj
```diff
@@ -70,11 +70,6 @@
       <DesignTime>True</DesignTime>
       <DependentUpon>Resources.resx</DependentUpon>
     </Compile>
-    <Compile Update="Properties\Resources.en-US.Designer.cs">
```
mjkkirschner, line 73: I don't **think** this should be removed from the csproj...

test/DynamoCoreTests/DynamoCoreTests.csproj
```diff
@@ -72,6 +72,7 @@
       <Project>{a4794476-7d0e-41c0-ad83-4ab929c0a46c}</Project>
       <Name>ProtoScript</Name>
     </ProjectReference>
+    <ProjectReference Include="..\..\src\Libraries\CoreNodeModelsWpf\CoreNodeModelsWpf.csproj" />
```
mjkkirschner, line 75: we should find a way to avoid adding this dependency, there is a strict boundary between the core/model binaries and the wpf/view binaries - so that Dynamo can be built and tested cross platform.

twastvedt: Oops. Looks like I no longer needed that reference anyway. Thanks!

src/DynamoCoreWpf/UI/Themes/Modern/DynamoModern.xaml
```diff
@@ -1446,29 +1451,20 @@
...
-                            <Setter TargetName="roundedBorder" Property="CornerRadius" Value="4,0,0,4" />
-                        </DataTrigger>
-                        <DataTrigger Binding="{Binding ElementName=text, Path=Text}" Value="-">
-                            <Setter TargetName="text" Property="FontSize" Value="28px" />
-                            <Setter TargetName="text" Property="Margin" Value="0,3,1,0" />
-                            <Setter TargetName="text" Property="Foreground" Value="#999999" />
-                            <Setter TargetName="container" Property="Height" Value="22px" />
-                            <Setter TargetName="container" Property="Width" Value="22px" />
-                            <Setter TargetName="roundedBorder" Property="CornerRadius" Value="0,4,4,0" />
-                        </DataTrigger>
```
twastvedt, line 1465: As far as I can tell this is left-over. The +/- buttons in list join use the AddRemoveButton below. It would be nice if that could be merged into this style, but that's separate. Removing these here allows us to use SNodeTextButton to display stand-alone +/- buttons.

src/DynamoCoreWpf/UI/Themes/Modern/DynamoModern.xaml
```diff
@@ -1446,29 +1451,20 @@
...
                     </ControlTemplate.Triggers>
                 </ControlTemplate>
             </Setter.Value>
         </Setter>
     </Style>
 
+    <Style TargetType="{x:Type nodes:DynamoNodeButton}" BasedOn="{StaticResource SNodeTextButton}" />
```
twastvedt, line 1460: A default style for DynamoNodeButton, an alternative to the style set in the constructor, to allow the style to be overridden.

src/Libraries/CoreNodeModels/Input/CustomSelection.cs
```diff
@@ -0,0 +1,96 @@
...
+    [IsDesignScriptCompatible]
+    public class CustomSelection : DSDropDownBase
+    {
+        /// <summary>
+        /// Construct a new Custom Dropdown Menu node
+        /// </summary>
+        public CustomSelection() : base("Value")
+        {
+            // TODO: This isn't done in RevitDropDown. Necessary?
+            ArgumentLacing = LacingStrategy.Disabled;
```
twastvedt, line 31: I don't know enough about what this affects to know whether it should be here.

src/Libraries/CoreNodeModelsWpf/Controls/CustomSelectionControl.xaml
```diff
@@ -0,0 +1,109 @@
+<UserControl x:Class="CoreNodeModelsWpf.Controls.CustomSelectionControl"
```
twastvedt, line 1: This is pretty different from before. Based on Michael's feedback about using existing resources, I revised this to use the +/- buttons we already have for the List Join component instead of the "Add" and "X" buttons. 🤷

src/Libraries/CoreNodeModelsWpf/NodeViewCustomizations/CustomSelection.cs
```diff
@@ -0,0 +1,38 @@
...
+            // Add the dropdown.
+            base.CustomizeView(model, nodeView);
+
+            var dropdown = (ComboBox)nodeView.inputGrid.Children[1];
+
+            formControl.BaseComboBox = dropdown;
+
+            // Add margin to the dropdown to show the expander.
+            dropdown.Margin = new Thickness(40, 0, 0, 0);
+            dropdown.VerticalAlignment = VerticalAlignment.Top;
```
twastvedt, line 35: @saintentropy Turns out I didn't need the inherited dropdown view to be in xaml after all. I think this sort of manipulation should work for your unit node as well.

src/Libraries/CoreNodeModels/CoreNodeModelsImages.resx
```diff
@@ -1343,6 +1343,39 @@
         PIw7vteQPuhLVMkA479rRIQ9hegBi3gVlTPYvr1bMcEHxT5Ginf+R2dm53gHaDPpF5En0WZeR3VH6qT1
         trlz75fwK2w6VIVFBQ7AJvsESZk6BO/iW/E+9u9A57xW7BNsVoaQqfvVwBky/FPkch8Yc7bTvUnhewAA
         AABJRU5ErkJggg==
+</value>
+  </data>
+  <data name="CoreNodeModels.Input.CustomSelectionNodeModel.Large" type="System.Drawing.Bitmap, System.Drawing" mimetype="application/x-microsoft.net.object.bytearray.base64">
```
saintentropy, line 1348: I think this needs to be `CoreNodeModels.Input.CustomSelection` vs `CoreNodeModels.Input.CustomSelectionNodeModel`.  The icons are not loading

src/Libraries/CoreNodeModels/Input/CustomSelection.cs
twastvedt: Now I do. Looks like all components without inputs have lacing disabled, I think just to avoid an empty for loop. This should probably be set on the Revit dropdowns as well?

src/DynamoCoreWpf/UI/Themes/Modern/DynamoModern.xaml
```diff
@@ -3162,11 +3154,11 @@
                     <ControlTemplate.Triggers>
                         <Trigger Property="IsMouseOver" Value="True">
                             <Setter TargetName="border" Property="Opacity" Value="0.9" />
-                            <Setter TargetName="mouseOverBorder" Property="Visibility " Value="Visible" />
```
twastvedt, line 3165: I believe these are just typos. I was getting warnings in VS because of the extra empty space.

src/DynamoCoreWpf/UI/Themes/Modern/DynamoModern.xaml
```diff
@@ -5352,4 +5344,104 @@
             </Trigger>
         </Style.Triggers>
     </Style>
+
```
twastvedt, line 5347: Moving the expander style here from the slider component so we can use it in the dropdown as well.

src/DynamoCoreWpf/Controls/DynamoNodeButton.cs
```diff
@@ -30,7 +30,6 @@ private DynamoViewModel DynamoViewModel
 
         public DynamoNodeButton()
         {
-            Style = (Style)SharedDictionaryManager.DynamoModernDictionary["SNodeTextButton"];
```
twastvedt, line 33: Setting the style in the constructor was preventing me from overriding the style in XAML. I moved this to a default style in DynamoModern which applies to all `DynamoNodeButton`s but can be overridden in XAML.

src/Libraries/CoreNodeModelsWpf/Controls/DynamoSlider.xaml
```diff
@@ -94,104 +94,6 @@
                     </Setter.Value>
                 </Setter>
             </Style>
-            <ControlTemplate x:Key="ExpanderTemplate" TargetType="Expander">
```
twastvedt, line 97: Extracted to DynamoModern to share with the dropdown.

twastvedt:
@QilongTang Craig suggested I bring this PR to your attention. I've made some modifications to make better use of existing UI and styles. There are a few changes outside of this component, mostly to enable better sharing and style inheritance, namely in [DynamoModern.xaml](https://github.com/DynamoDS/Dynamo/pull/11958/files#diff-094c8df98e6724d704bb015c259cdf5352917a2076c57a4b4ffccaa74d003c4d). See my comments there. Happy to discuss!

twastvedt:
There is a bug when removing a list item that is fixed in #13217.

QilongTang:
@twastvedt @saintentropy I'm less worried about the style changes. While this is a new node I see, I highly encourage you to add an example of how it works as part of our node help docs. There are plenty examples at https://github.com/DynamoDS/Dynamo/tree/master/doc/distrib/NodeHelpFiles. 

And consider the real effort is to 
1. add some description
2. add an image of a small block of an example graph illustrating how it works
3. name the files following how other files are named, so that the context menu of nodes would locate your doc OOTB

doc/distrib/NodeHelpFiles/CoreNodeModels.Input.DoubleInput.md
```diff
@@ -1,5 +1,5 @@
 ## In Depth
-Room numbers are returned as strings.
+The Number node allows a user to enter a static number. This node is useful for things like constants that don't change values. Numbers can also be created by using a Code Block node. In the example below, the number is paired with a code block to create a list sequence with a variable step size.
```
mjkkirschner, line 2: thanks for fixing this @twastvedt - we should really fix it in the node markdown docs generation tool which is finding the wrong information from the dictionary - I see that it also happens for a bunch of a other nodes with similar names in different namespaces.

twastvedt: Jira issue filed

doc/distrib/NodeHelpFiles/CoreNodeModels.Input.CustomSelection.md
```diff
@@ -0,0 +1,6 @@
+## In Depth
+The Custom Dropdown node allows a user to create a dropdown selection input with custom labels and values. If all values are numbers, the output will be a double, and if all values are integers the output will be an integer. In the example below, "Two" is selected in the Custom Dropdown Menu node, making the output of that node the integer `2`.
```
mjkkirschner, line 2: what happens if the types are mixed? Can you use strings? What are valid value types etc?

Is this dynamic behavior important - I ask this because as we're working on the MSIL compiler - it's very hard to determine the output types of nodes that have dynamic behavior like this - not super important but I imagine that if we move forward with that work outside of a POC we're going to need some principles around not introducing dynamic behavior without a really good reason.

src/AssemblySharedInfoGenerator/AssemblySharedInfo.tt
```diff
@@ -73,4 +73,4 @@ int BuildNumber = 0;
 // The datetime baseline we choose using this algorithm will affect build number and all nuget packages uploaded
 // Please only change when major or minor version got incremented
 int RevisionNumber = ((int)(DateTime.UtcNow - new DateTime(2022,1,1)).TotalDays)*10+((int)DateTime.UtcNow.Hour)/3;
-#>
+#>
```
mjkkirschner, line 76: I think you should revert this change.

src/DynamoCoreWpf/UI/Themes/Modern/DynamoModern.xaml
```diff
@@ -1,11 +1,11 @@
-﻿<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
                     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
                     xmlns:avalonedit="clr-namespace:ICSharpCode.AvalonEdit;assembly=ICSharpCode.AvalonEdit"
                     xmlns:clr="clr-namespace:System;assembly=mscorlib"
                     xmlns:controls="clr-namespace:Dynamo.Views;assembly=DynamoCoreWpf"
+                    xmlns:nodes="clr-namespace:Dynamo.Nodes;assembly=DynamoCoreWpf"
                     xmlns:dynui="clr-namespace:Dynamo.UI.Controls;assembly=DynamoCoreWpf"
                     xmlns:fa="clr-namespace:FontAwesome.WPF;assembly=FontAwesome.WPF"
-                    xmlns:nodes="clr-namespace:Dynamo.Nodes;assembly=DynamoCore"
```
mjkkirschner, line 8: so was this just incorrect?

src/DynamoCoreWpf/UI/Themes/Modern/DynamoModern.xaml
```diff
@@ -1414,61 +1414,53 @@
     </Style>
 
     <Style x:Key="SNodeTextButton" TargetType="{x:Type Button}">
+        <Setter Property="FontSize" Value="17px"/>
+        <Setter Property="Foreground" Value="{StaticResource Blue300Brush}" />
+        <Setter Property="Background" Value="#4D4D4D" />
```
mjkkirschner, line 1419: is this color reused in a lot of places? Can it be pulled out?

src/DynamoCoreWpf/UI/Themes/Modern/DynamoModern.xaml
```diff
@@ -1414,61 +1414,53 @@
...
                                        FontWeight="Bold"
                                        Style="{StaticResource SZoomFadeText}"
                                        Text="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=Content}" />
                         </Border>
                     </Grid>
                     <ControlTemplate.Triggers>
                         <Trigger Property="Button.IsMouseOver" Value="true">
-                            <Setter TargetName="roundedBorder" Property="Background" Value="#535353" />
-                            <Setter TargetName="text" Property="Foreground" Value="{StaticResource Blue300Brush}" />
+                            <Setter Property="Background" Value="#535353" />
```
mjkkirschner, line 1446: same color question

src/DynamoCoreWpf/UI/Themes/Modern/DynamoModern.xaml
```diff
@@ -1414,61 +1414,53 @@
...
                     </ControlTemplate.Triggers>
                 </ControlTemplate>
             </Setter.Value>
         </Setter>
     </Style>
 
+    <Style TargetType="{x:Type nodes:DynamoNodeButton}" BasedOn="{StaticResource SNodeTextButton}" />
+
+    <Style x:Key="SingleCharButton" TargetType="{x:Type Button}" BasedOn="{StaticResource SNodeTextButton}">
+        <Setter Property="Foreground" Value="#999999" />
```
mjkkirschner, line 1458: color reuse question

src/Libraries/CoreNodeModels/Input/CustomSelection.cs
```diff
@@ -0,0 +1,161 @@
...
+using Dynamo.Graph.Nodes;
+
+using Newtonsoft.Json;
+
+using ProtoCore.AST.AssociativeAST;
+
+namespace CoreNodeModels.Input
+{
+    /// <summary>
+    /// This node allow the user to create a dropdown menu with an with an arbitrary number of customization items
```
mjkkirschner, line 19: allow(s)

src/Libraries/CoreNodeModels/Input/CustomSelection.cs
```diff
@@ -0,0 +1,161 @@
...
+    /// </summary>
+    [NodeName("Custom Dropdown Menu")]
+    [NodeCategory(BuiltinNodeCategories.CORE_INPUT)]
+    [NodeDescription("CustomSelectionNodeDescription", typeof(Properties.Resources))]
+    [NodeSearchTags("CustomSelectionSearchTags", typeof(Properties.Resources))]
+    [OutPortNames("value")]
+    [OutPortTypes("var")]
+    [OutPortDescriptions("Selected value")]
```
mjkkirschner, line 27: shouldn't this also use a localized resource?

src/Libraries/CoreNodeModels/Input/CustomSelection.cs
```diff
@@ -0,0 +1,161 @@
...
+    [IsDesignScriptCompatible]
+    public class CustomSelection : DSDropDownBase
+    {
+        private List<DynamoDropDownItem> serializedItems;
+
+        public List<DynamoDropDownItem> SerializedItems
```
mjkkirschner, line 33: public properties should have /// summary tags.

src/Libraries/CoreNodeModels/Input/CustomSelection.cs
```diff
@@ -0,0 +1,161 @@
...
+
+            return selectedItem?.Item;
+        }
+
+        protected override SelectionState PopulateItemsCore(string currentSelection)
+        {
+            return SelectionState.Restore;
+        }
+
+        [OnSerializing]
```
mjkkirschner, line 124: what does this attribute do?

src/Libraries/CoreNodeModelsWpf/Controls/CustomSelectionControl.xaml.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.Windows.Input;
+
+using CoreNodeModels.Input;
+
+using Dynamo.Utilities;
+
+namespace CoreNodeModelsWpf.Controls
+{
+    /// <summary>
+    /// Interaction logic for DropDownControl.xaml
```
mjkkirschner, line 13: this comment seems to be out of date.

src/Libraries/CoreNodeModelsWpf/Controls/CustomSelectionControl.xaml.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// </summary>
+        public CustomSelectionControl(CustomSelectionViewModel viewModel)
+        {
+            DataContext = viewModel;
+            model = viewModel.Model;
+
+            InitializeComponent();
+        }
+
+        public ComboBox BaseComboBox { get; set; }
```
mjkkirschner, line 30: why is this public, is it bound somewhere?

src/Libraries/CoreNodeModelsWpf/Controls/CustomSelectionControl.xaml.cs
```diff
@@ -0,0 +1,80 @@
...
+        private void EnumItemsListbox_PreviewKeyDown(object sender, KeyEventArgs e)
+        {
+            if (!( sender is ListBox listBox ))
+            {
+                return;
+            }
+
+            if (e.Key == Key.Tab)
+            {
+                int offset = e.KeyboardDevice.Modifiers == System.Windows.Input.ModifierKeys.Shift ? 0 : 1;
```
mjkkirschner, line 41: I am usually not a stickler for style stuff besides `if(){ }` - but the whitespace seems a bit confusing to the eye.

src/Libraries/CoreNodeModelsWpf/CustomSelectionViewModel.cs
```diff
@@ -0,0 +1,48 @@
+using CoreNodeModels;
+using CoreNodeModels.Input;
+
+using Dynamo.Core;
+using Dynamo.UI.Commands;
+
+namespace CoreNodeModelsWpf
+{
+    public class CustomSelectionViewModel : NotificationObject
```
mjkkirschner, line 9: please add /// summary tags to public members, especially when something has such a generic name such as custom selection.

src/Libraries/CoreNodeModelsWpf/CustomSelectionViewModel.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        public CustomSelection Model { get; }
+
+        /// <summary>
+        /// This command is bound to the Add button in the GUI
+        /// </summary>
+        public DelegateCommand AddCommand { get; }
+
+        public DelegateCommand RemoveCommand { get; }
```
mjkkirschner, line 18: ///

src/Libraries/CoreNodeModelsWpf/CustomSelectionViewModel.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            Model.Items.Add(new DynamoDropDownItem(string.Empty, string.Empty));
+        }
+
+        private void RemoveItem(object parameter)
+        {
+            if (parameter is DynamoDropDownItem item)
+            {
+                Model.Items.Remove(item);
+
```
mjkkirschner, line 30: funny whitespace - alright, thats my last comment about it.

src/Libraries/CoreNodeModelsWpf/NodeViewCustomizations/CustomSelection.cs
```diff
@@ -0,0 +1,38 @@
...
+using CoreNodeModels.Input;
+using CoreNodeModelsWpf.Controls;
+using System.Windows.Controls;
+using System.Windows;
+using Dynamo.Wpf;
+
+namespace CoreNodeModelsWpf.Nodes
+{
+    /// <summary>
+    /// View customizer for DropDown node model.
```
mjkkirschner, line 11: seems out of date.

src/Libraries/CoreNodeModelsWpf/NodeViewCustomizations/CustomSelection.cs
```diff
@@ -0,0 +1,38 @@
...
+        public void CustomizeView(CustomSelection model, NodeView nodeView)
+        {
+            var formControl = new CustomSelectionControl(new CustomSelectionViewModel(model));
+
+            nodeView.inputGrid.Children.Add(formControl);
+
+            // Add the dropdown.
+            base.CustomizeView(model, nodeView);
+
+            var dropdown = (ComboBox)nodeView.inputGrid.Children[1];
```
mjkkirschner, line 29: this seems brittle, maybe make it a bit more defensive, index checking etc.

src/LibraryViewExtensionMSWebBrowser/web/library/layoutSpecs.json
```diff
@@ -443,6 +443,9 @@
                 },
                 {
                   "path": "Core.Input.Output"
+                },
+                {
+                  "path": "Core.Input.Custom Dropdown Menu"
```
mjkkirschner, line 448: I won't die on this hill, but is this really a dropdown **menu**?

test/Tools/docGeneratorTestFiles/sampledictionarycontent/Dynamo_Nodes_Documentation.json
```diff
@@ -2992,6 +2992,12 @@
     "dynFile": ["Output"],
     "folderPath": "Input/Basic/Action",
     "inDepth": "Output will create an output port for a custom node. The syntax for an output node is simply its name, along with an optional custom comment. In the example below, an output named Percentage is created with a custom comment. It is important to note that this node is only available while creating a custom node."
+}, {
+    "Name": "Custom Dropdown Menu",
```
mjkkirschner, line 2996: so, this shouldn't be necessary, it's been a long time since I wrote these tests though so maybe I am wrong - BUT it seems a bit backwards that you need to do this as we're hopefully getting rid of this wacky json and instead using the generated .md files and the layoutspec to construct the dictionary one day, no need to change this now though if you found it was required.

mjkkirschner (commented):
a few questions / comments

mjkkirschner:
also kicked off: https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/1083/ on the serial tests just to see if anything unexpected happens there.

doc/distrib/NodeHelpFiles/CoreNodeModels.Input.CustomSelection.md
twastvedt: Immediate answers: values default to strings unless the criteria are met. So, anything's valid, it will just be a string unless we can make everything numbers. I'll add that, or whatever we decide here, to the description.

I'm definitely open to alternatives, though I think this behavior is pretty convenient, intuiting a user's expectations in most common scenarios. I'm not sure what the alternative to dynamic behavior here is. Separate String custom selection and Number custom selection nodes sounds awkward for the user. I imagine the original inspiration for this part may have been the experience in Grasshopper. Though there it's less a feature of the custom selection node itself and more of the global type coercion.

src/DynamoCoreWpf/UI/Themes/Modern/DynamoModern.xaml
twastvedt: Looks like it's used in 5 places in DynamoModern. It already exists in ColorsAndBrushes, but for more specific uses. I wasn't sure how far to take a style revision, so I defaulted to just making the changes necessary to use the styles for the new component. From my interaction so far it looks like there are a _lot_ of optimization opportunities like this: unnecessary styles, duplicated sections, hard-coded colors.

test/Tools/docGeneratorTestFiles/sampledictionarycontent/Dynamo_Nodes_Documentation.json
twastvedt: Yeah, `MarkdownGeneratorCommandTests.DictionaryContentIsFoundCorrectlyForCoreNodes` fails if I don't have this.

twastvedt (commented):
Thanks for the review! I've fixed most things (I think!), a few might be worth further discussion.

doc/distrib/NodeHelpFiles/CoreNodeModels.Input.CustomSelection.md
mjkkirschner: let's put this on the back burner.

src/Libraries/CoreNodeModelsWpf/CustomSelectionViewModel.cs
mjkkirschner: yes space between lines - 
I think you're on the right track - I usually just go by what ms says:
https://docs.microsoft.com/en-us/learn/modules/csharp-readable-code/4-exercise-use-whitespace

src/Libraries/CoreNodeModelsWpf/NodeViewCustomizations/CustomSelection.cs
mjkkirschner: that sounds good, as long as there is no crash.

mjkkirschner (approved):
Thanks @twastvedt looks good to me. :shipit:

PR #11972, opened 2021-08-25
add inner exceptions to crash report

pinzart90:
https://jira.autodesk.com/browse/DYN-3945

Wpf control initialization crash usually wraps Exceptions.
In this PR I added inner exceptions in the crash prompt

### DIFF
--- a/src/DynamoCore/Core/CrashPromptArgs.cs
+++ b/src/DynamoCore/Core/CrashPromptArgs.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 
 namespace Dynamo.Core
 {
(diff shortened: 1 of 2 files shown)


src/DynamoCore/Core/CrashPromptArgs.cs
```diff
@@ -63,6 +64,28 @@ public CrashPromptArgs(string details, string overridingText = null, string file
...
+            {
+                var eDetails = new List<string>();
+                GetDetailsFromException(e, ref eDetails);
+
+                Details = string.Join($"\n\n-------- {nameof(Exception.InnerException)} --------\n\n", eDetails);
+                Options |= DisplayOptions.HasDetails;
+            }
+        }
+
+        private void GetDetailsFromException(Exception e, ref List<string> details)
```
aparajit-pratap, line 79: does the second param need to be a `ref` param, I don't think so?

PR #11992, opened 2021-08-30
add logs to method crashing on test machine

mjkkirschner:
DNM


QilongTang:
Hey @mjkkirschner How would you handle this moving forward? I think group branch still suffering with this problem?

mjkkirschner:
@SHKnudsen @QilongTang the next step is to add more logs to try to figure out what test is actually crashing - might give us hints as to why AnnotationText is null.

PR #11994, opened 2021-08-30
DYN-3923: Package List Filtering UI

sm6srw:
### Purpose

This pull request does:
* add filtering capabilities to the list of installed packages.

DYN-3923

### Declarations


pinzart90:
What happens if the new filter buttons do not all fit ? Do they continue down ?

sm6srw:
> What happens if the new filter buttons do not all fit ? Do they continue down ?

Yes

src/DynamoCoreWpf/ViewModels/PackageManager/InstalledPackagesViewModel.cs
```diff
@@ -1,51 +1,264 @@
...
+                        break;
+                }
+
+                switch (LoadState.State)
+                {
+                    case PackageLoadState.StateTypes.Unloaded: return Resources.PackageStateUnloaded;
+                    case PackageLoadState.StateTypes.Loaded: return Resources.PackageStateLoaded;
+                    case PackageLoadState.StateTypes.Error: return Resources.PackageStateError;
+                    default:
+                        return Resources.PackageFilter_Name_All;
```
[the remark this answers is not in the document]
pinzart90: Or you could just store the Name (as a string) instead of the packageViewModel or the LoadState

sm6srw:
> > What happens if the new filter buttons do not all fit ? Do they continue down ?
> 
> Yes

I take that back. I broke it along the way.

src/DynamoCoreWpf/ViewModels/PackageManager/InstalledPackagesViewModel.cs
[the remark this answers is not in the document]
sm6srw: Yes, I have been thinking about using the plain string instead. It makes it a little harder to do the actual filtering but might simplify things overall. I will give it a try.

sm6srw:
> I take that back. I broke it along the way.

Should be fixed now.

src/DynamoCoreWpf/ViewModels/PackageManager/InstalledPackagesViewModel.cs
```diff
@@ -1,51 +1,257 @@
...
-        internal void GoToWebsite()
+        /// <summary>
+        /// Populate the list of available filters based on the packages currently known by the dynamo view model
+        /// </summary>
+        internal void PopulateFilters()
         {
-            dynamoViewModel.PackageManagerClientViewModel.GoToWebsite();
+            var currentSelection = CurrentFilterSelection;
+            Filters.Clear();
+            var filterNames = Model.LocalPackages.Select(pkg => new PackageViewModel(dynamoViewModel, pkg))
```
pinzart90, line 142: Why don;t you use the `public ObservableCollection<PackageViewModel> LocalPackages` here, instead of calling `Model.LocalPackages.Select(pkg => new PackageViewModel(dynamoViewModel, pkg))` ?

src/DynamoCoreWpf/ViewModels/PackageManager/InstalledPackagesViewModel.cs
```diff
@@ -1,51 +1,257 @@
...
+        private string CurrentFilterSelection => Filters.FirstOrDefault(f => f.IsChecked)?.Name;
+
+        private bool NoActiveFilterSelected
         {
-            foreach (var pkg in Model.LocalPackages)
+            get
             {
-                LocalPackages.Add(new PackageViewModel(this.dynamoViewModel, pkg));
+                var currentFilter = CurrentFilterSelection;
+                return currentFilter == null || currentFilter == Resources.PackageFilter_Name_All;
```
pinzart90, line 163: I think you can just use `return CurrentFilterSelection== Resources.PackageFilter_Name_All` 
if CurrentFilterSelection is null then it is still a valid check (i.e null == Resources.PackageFilter_Name_All)

src/DynamoCoreWpf/ViewModels/PackageManager/InstalledPackagesViewModel.cs
sm6srw: LocalPackages may be filtered when we get here and only contain a subset of all packages.

src/DynamoCoreWpf/ViewModels/PackageManager/InstalledPackagesViewModel.cs
sm6srw: Yes, but I want  the no filter (currentFilter is null) to be the same as a the All Filter.

pinzart90 (approved):

src/DynamoCoreWpf/Controls/InstalledPackagesControl.xaml
```diff
@@ -30,7 +30,42 @@
...
+                <DataTemplate>
+                    <RadioButton Content="{Binding Path=Name, Mode=OneWay}" 
+                                 GroupName="PackageFilters"
+                                 IsChecked="{Binding Path=IsChecked, Mode=TwoWay}"
+                                 Style="{StaticResource {x:Type ToggleButton}}"
+                                 FontSize="10"
+                                 FontFamily="Artifakt Element"
+                                 HorizontalAlignment="Center"
+                                 BorderThickness="1,1,1,1"
+                                 Background="#CCCCCC"
```
mjkkirschner, line 53: are these colors we can get from the style dictionary?

src/DynamoCoreWpf/ViewModels/PackageManager/InstalledPackagesViewModel.cs
```diff
@@ -1,51 +1,261 @@
 ﻿using System.Collections.ObjectModel;
+using System.ComponentModel;
 using System.Linq;
-
+using System.Security.Cryptography;
```
mjkkirschner, line 4: whats this using about?

src/DynamoCoreWpf/ViewModels/PackageManager/InstalledPackagesViewModel.cs
```diff
@@ -1,51 +1,261 @@
...
+
+            LocalPackages.Clear();
+        }
+
+        private void InitializeLocalPackages()
+        {
+            ClearPackagesViewModel();
+            LocalPackages.AddRange(Model.LocalPackages.Select(NewPackageViewModel));
+
+            Model.PackageAdded += (pkg) =>
```
mjkkirschner, line 205: I think we should try to clean these handlers up if possible... packageLoader may live longer than the `InstalledPackagesViewModel` - maybe - I can't remember if it lives for the application lifetime or just when the prefs view is opened... Also need to consider test contexts.

src/DynamoCoreWpf/ViewModels/PackageManager/InstalledPackagesViewModel.cs
sm6srw: I don't know how it got there. I will remove it.

src/DynamoCoreWpf/ViewModels/PackageManager/InstalledPackagesViewModel.cs
sm6srw: It is for the application lifetime as it stands right now.

src/DynamoCoreWpf/Controls/InstalledPackagesControl.xaml
sm6srw: Yes

pinzart90:
@mjkkirschner anything else that you think needs to be addressed ?
I would like to get this merged

mjkkirschner:
@pinzart90 - it looks fine to me, but I would like to confirm that the new event handlers which are not cleaned up do not cause issues when running in serial:
I've kicked off:
https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/866/

mjkkirschner (approved):

mjkkirschner:
@pinzart90 self serve serial passed.

PR #12002, opened 2021-09-01
Dyn 3477 - Avoid unexpected DS function-not-found errors in CBNs, invalidate function definitions on opening new workspace

aparajit-pratap:
### Purpose

This PR fixes a bug where when a graph containing more than one CBN is opened, where one of those CBNs contains a function definition, while the other contains a function call, there can be a `function not found` warning if the CBN with the function call is pre-compiled first. This is obviously unexpected for the user as the function definition actually exists in another CBN. 

This is fixed by splitting the pre-compilation of CBNs into two passes. The first pass parses and compiles the CBN like normal except that in the first pass any `function not found` warnings are skipped. In the second pass of pre-compilation, the CBN code (except for function definitions) is compiled again turning back on those warnings as in the second pass it needs to take into account any function definitions that may have already been added to `library management core` in the first pass. Only after the second pass any `function not found` warnings will be _real_ as by this time all function definitions should have been taken into account in the first pass.

Note that this two-stage pre-compilation of CBNs _only_ needs to happen on opening a workspace, _not_ in other cases of CBN creation.

Additionally, any function definitions in CBNs from a previous workspace are invalidated (`procedureNode `is marked inactive) upon opening a new workspace so that they no longer continue to show up in autocomplete, and `function not found` warnings appear if an attempt is made to call them.

### Declarations

- [x] All tests pass using the self-service CI. - ~~all tests pass with the exception of `Dynamo.Tests.CoreTests.TestHasUnsavedChangesOnNodeRemovalEdgeCase`, a fix is in progress ...~~


src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1817,9 +1818,10 @@ private void SetPeriodicEvaluation(WorkspaceModel ws)
           bool forceManualExecutionMode,
           out WorkspaceModel workspace)
         {
-            CustomNodeManager.AddUninitializedCustomNodesInPath(
-                Path.GetDirectoryName(filePath),
-                IsTestMode);
+            CustomNodeManager.AddUninitializedCustomNodesInPath(Path.GetDirectoryName(filePath), IsTestMode);
+
+            var currentHomeSpace = Workspaces.OfType<HomeWorkspaceModel>().FirstOrDefault();
+            currentHomeSpace.UndefineCBNFunctionDefinitions();
```
pinzart90, line 1824: why do we need to undefine func definitions before we ReCompileCodeBlockNodesForFunctionDefinitions ?

aparajit-pratap: We need to undefine the function definitions if any in the previous workspace before we open a new workspace.

aparajit-pratap: ReCompileCodeBlockForFunctionDefinitions will be called _after_ new function definitions have been compiled once the new workspace has been parsed.

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -565,12 +565,23 @@ private void NodeOnRequestSilenceNodeModifiedEvents(NodeModel node, bool value)
         public override void Clear()
         {
             WorkspaceClosed?.Invoke();
+            UndefineCBNFunctionDefinitions();
```
sm6srw, line 568: So we have never undefined function definitions before loading a new workspace until now? Where they still around after a new home workspace was loaded?

sm6srw (approved):
Nice, one question then LGTM!

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
aparajit-pratap: Yes, that is correct. There are two cores in the VM - one static and the other the runtime core. The runtime core was being reset every time a new workspace was opened but not the static core. I realized that if we reset the static core, we would also need to reload all libraries and packages again into the core, so I decided to simply undefine the function definitions from the static core instead.

aparajit-pratap:
> So regarding
> `Additionally, any function definitions in CBNs from a previous workspace are invalidated (procedureNode is marked inactive) upon opening a new workspace so that they no longer continue to show up in autocomplete, and function not found warnings appear if an attempt is made to call them`
> I wonder if this was useful in certain cases...Maybe having multiple workspaces opened at the same time...so you can share function definitions ...
> @mjkkirschner @aparajit-pratap ?

We don't have multiple workspace support right now or a proposed design for how it should work w.r.t. function definitions.

mjkkirschner:
> > So regarding
> > `Additionally, any function definitions in CBNs from a previous workspace are invalidated (procedureNode is marked inactive) upon opening a new workspace so that they no longer continue to show up in autocomplete, and function not found warnings appear if an attempt is made to call them`
> > I wonder if this was useful in certain cases...Maybe having multiple workspaces opened at the same time...so you can share function definitions ...
> > @mjkkirschner @aparajit-pratap ?
> 
> We don't have multiple workspace support right now or a proposed design for how it should work w.r.t. function definitions.
> With the way things stand currently, it is confusing to still be able to access a function that was defined in a previously closed workspace.

I agree - because it creates a situation where users can build broken graphs.

src/Engine/ProtoCore/Utils/ParserUtils.cs
```diff
@@ -137,11 +137,12 @@ public static List<AST.AssociativeAST.AssociativeNode> GetAstNodes(AST.Associati
 
         private static Core CodeBlockParserCore()
         {
-            var core = new Core(new Options());
-            core.Options.ExecutionMode = ExecutionMode.Serial;
```
mjkkirschner, line 141: is ExecutionMode.Serial the default?

src/DynamoCore/Engine/EngineController.cs
```diff
@@ -609,14 +609,20 @@ public CompilationServices(LibraryServices libraryServices)
...
         /// </summary>
         /// <param name="parseParams">Container for compilation related parameters</param>
         /// <returns>true if code compilation succeeds, false otherwise</returns>
         public bool PreCompileCodeBlock(ref ParseParam parseParams)
         {
-            return CompilerUtils.PreCompileCodeBlock(compilationCore, ref parseParams, priorNames);
+            return CompilerUtils.PreCompileCodeBlock(compilationCore, parseParams, priorNames);
+        }
+
+        internal bool PreCompileCodeBlock(ParseParam parseParams)
```
mjkkirschner, line 623: could we copy the summary stuff over if still applicable, helps to understand the method imo.

src/DynamoCore/Graph/Nodes/CodeBlockNode.cs
```diff
@@ -45,6 +45,7 @@ public class CodeBlockNodeModel : NodeModel
         private readonly LibraryServices libraryServices;
 
         private bool shouldFocus = true;
+        internal ParseParam ParseParam { get; private set; }
```
mjkkirschner, line 48: so this is a cache for keeping around results from previous compilations?

src/DynamoCore/Graph/Nodes/CodeBlockNode.cs
```diff
@@ -604,12 +618,14 @@ public override ProtoCore.Type GetTypeHintForOutput(int index)
 
         #region Private Methods
 
-        internal void ProcessCodeDirect()
+        internal delegate void ProcessCodeDelegate(ref string errorMessage, ref string warningMessage);
```
mjkkirschner, line 621: should this actually use `out` to be explicit about the purpose of these parameters? Is there any time that these params are meaningful before the handler is invoked? (ie do we ever pass anything useful into the handler?)

src/DynamoCore/Graph/Nodes/CodeBlockNode.cs
```diff
@@ -627,30 +643,122 @@ internal void ProcessCodeDirect()
...
+
+                core.IsParsingCodeBlockNode = parsingCbnFlag;
+                core.IsParsingPreloadedAssembly = parsingPreloadFlag;
+
+                ParseParam.AppendErrors(buildStatus.Errors);
+                ParseParam.AppendWarnings(buildStatus.Warnings);
+            }
+            catch (Exception)
+            {
+                buildStatus = null;
```
mjkkirschner, line 702: should we log something?

src/DynamoCore/Graph/Nodes/CodeBlockNode.cs
```diff
@@ -627,30 +643,122 @@ internal void ProcessCodeDirect()
...
+                ProcessError();
+                return;
+            }
+            if (ParseParam.Warnings != null)
+            {
+                // Unbound identifiers in CBN will have input slots.
+                // 
+                // To check function redefinition, we need to check other
+                // CBN to find out if it has been defined yet. Now just
+                // skip this warning.
```
mjkkirschner, line 717: should this be a todo/followup jira?

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
mjkkirschner: is this called both when opening a new workspace and creating a new workspace?

src/DynamoCore/Graph/Nodes/CodeBlockNode.cs
aparajit-pratap: Yes

src/DynamoCore/Graph/Workspaces/NodesToCodeExtensions.cs
```diff
@@ -104,7 +104,6 @@ internal static class NodesToCodeExtensions
 
                 var codeBlockNode = new CodeBlockNodeModel(
                     code,
-                    System.Guid.NewGuid(),
```
mjkkirschner, line 107: ??

src/Engine/ProtoCore/Utils/ParserUtils.cs
aparajit-pratap: Yes, parallel was never implemented.

src/Engine/ProtoCore/Core.cs
```diff
@@ -315,53 +324,59 @@ public void SetFunctionInactive(FunctionDefinitionNode functionDef)
...
-
-                // Make a copy of all symbols defined in this function
-                var localSymbols = block.symbolTable.symbolList.Values
-                                        .Where(n => 
-                                                n.classScope == Constants.kGlobalScope 
-                                             && n.functionIndex == index)
-                                        .ToList();
-
-                foreach (var symbol in localSymbols)
+                procNodes = block.procedureTable.Procedures.Where(p => p.HashID == hash).ToList();
```
mjkkirschner, line 332: so what situation causes a procedure in the procedureTable to somehow have the same hashID, isn't it a hash table? How can their be more than one?

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
aparajit-pratap: When opening an existing workspace, this function is called in another place - when reading the JSON. There is another call to the same method in this PR.

src/Engine/ProtoCore/Core.cs
```diff
@@ -315,53 +324,59 @@ public void SetFunctionInactive(FunctionDefinitionNode functionDef)
...
+
+                        foreach (CodeBlock cb in CodeBlockList)
+                        {
+                            cb.children.RemoveAll(x => x.codeBlockId == cbID);
+                        }
                     }
                 }
             }
 
+            if (DSExecutable == null) return;
```
mjkkirschner, line 379: when does this happen?

src/DynamoCore/Graph/Nodes/CodeBlockNode.cs
aparajit-pratap: Yes, I think it makes sense for them to be `out` instead of `ref`. Will make the change.

src/DynamoCore/Graph/Workspaces/NodesToCodeExtensions.cs
aparajit-pratap: This happens inside the other constructor.

src/Engine/ProtoCore/Core.cs
aparajit-pratap: When the method is invoked on the static core. There is no `DSExecutable` for this core.

mjkkirschner:
@aparajit-pratap it looks good, I've got a few more questions in line - 
but on the subject of performance, it seems we need to compile most of the AST twice on open - depending on the graph I guess this could be a little costly?

I guess we need to recompile only things that refer to those functions? ie, function calls, and function objects?

src/Engine/ProtoCore/Core.cs
aparajit-pratap: `Procedures` is a list, not a hashtable.

aparajit-pratap: As long as the function name and the argument types and strings are the same, the same hash ID is generated. So if a function is deleted, then re-created, say if you delete a CBN and undo the deletion, a new entry for the same function with the same hash ID will be added, after the previous one has been marked inactive. The new one will be active.

mjkkirschner: 1. It's surprising that it's not a table at all... any idea why it's implemented with a list?
2. Does it make sense to use some query method on the procedureTable instead of accessing the list direclty?
3. Did you see this case come up or it's just defensive coding?

mjkkirschner (approved):
lgtm

src/Engine/ProtoCore/Core.cs
mjkkirschner: oh interesting.. so the procedure table will fill up with inactive functions - should it be cleaned up at some point? If we don't already do that?

aparajit-pratap: @pinzart90 had the same question. I don't have an answer to why a list was chosen and not a dictionary/hash table. I just think it was a poor choice of data structure. We do this even for the instruction stream where that is another list that keeps growing. Deleting instead of maintaining an ever-expanding list (and marking stale items inactive) is definitely the way to go but it's a bigger change and probably outside the scope of this PR. There needs to be a separate spike/epic to optimize these data structures.

aparajit-pratap: The reason its name contains the word "table" is probably because it is aimed at representing that concept of a function table in traditional compiler design.

aparajit-pratap:
> I guess we need to recompile only things that refer to those functions? ie, function calls, and function objects? Maybe this is an optimization we can skip for now, just wanted to start the discussion depending on if you have any performance concerns.

Yes, code now needs to be recompiled twice, which is a downside of this PR, however, for the most part compilation is relatively inexpensive as compared to actual execution (except for detecting cycles). I can distil the AST's down to only function call ASTs and recompile only those in the second pass as you suggested but I'm not sure whether that will make much of a difference as I would first need to walk each AST to check for the function calls on the rhs. Also, most of these function calls will most likely be FFI calls just because ZT methods are more common than DS function definitions. Definitely, something we can explore further.

mjkkirschner:
@aparajit-pratap - I'm happy with the state for now, we can try to optimize it further if we need, we could watch the perf CI test to see if opening graphs is affected by a significant margin.

src/DynamoCore/Graph/Nodes/CodeBlockNode.cs
aparajit-pratap: https://jira.autodesk.com/browse/DYN-4104

src/DynamoCore/Graph/Nodes/CodeBlockNode.cs
aparajit-pratap: I'm not sure if we can log any useful information here that would let the user debug the node. For example, how would we identify this code block node for the user? We are catching a similar exception [here](https://github.com/DynamoDS/Dynamo/blob/3df0eede2b94dc967354535b35703d9d13c0dd0f/src/Engine/ProtoCore/Utils/CompilerUtils.cs#L371) and not logging anything there either.

Any compilation warnings should appear on the code block node in any case.

PR #12012, opened 2021-09-02
Fix duplicated nodes when re adding package paths

pinzart90:
https://jira.autodesk.com/browse/DYN-3852

Prevent nodes from being duplicated if the package paths are re-added to the node and package paths list.


src/DynamoPackages/PackageLoader.cs
```diff
@@ -297,6 +294,7 @@ private void TryLoadPackageIntoLibrary(Package package)
 
                 package.LoadState.SetAsLoaded();
                 PackgeLoaded?.Invoke(package);
+                PackagesLoaded?.Invoke(loadedAssemblies);
```
aparajit-pratap, line 297: Can you explain how moving this call into `TryLoadPackageIntoLibrary` fixes the bug?

mjkkirschner: agree - I'm not clear on how this works.

pinzart90: There are conditions in the `TryLoadPackageIntoLibrary` that (if evaluated as true) will skip loading the package.
Ex: If the package was already loaded or if it was marked as unloaded, or if an exception is thrown...

In the case where we do not load a package then we should not invoke the `PackagesLoaded` event for that package.

In the unit test I wrote (for example) the `PackagesLoaded` event is not called if the package was already loaded

sm6srw (approved):
LGTM!

aparajit-pratap (approved):

PR #12020, opened 2021-09-07
fix string format arguments

pinzart90:
https://jira.autodesk.com/browse/DYN-3856
String.Format exception thrown

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
@@ -510,7 +510,7 @@ namespace Dynamo.ViewModels
                     catch
                     {
                         MessageBoxService.Show(
-                            String.Format(Resources.MessageFailedToDownloadPackageVersion, dep._id, depVersion),
+                            String.Format(Resources.MessageFailedToDownloadPackageVersion, depVersion, dep._id),
                             Resources.PackageDownloadErrorMessageBoxTitle,
                             MessageBoxButton.OK, MessageBoxImage.Error);
                         return null;
--- a/src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
@@ -498,7 +498,7 @@ namespace Dynamo.ViewModels
             var pmExt = DynamoViewModel.Model.GetPackageManagerExtension();
             if (result == MessageBoxResult.OK)
             {
-                package.full_dependency_ids.Count == package.full_dependency_versions.Count
+                System.Diagnostics.Debug.Assert(package.full_dependency_ids.Count == package.full_dependency_versions.Count);
                 // get all of the dependency version headers
                 var dependencyVersionHeaders = package.full_dependency_ids.Select((dep, i) =>
                 {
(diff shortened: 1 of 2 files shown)


src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
```diff
@@ -498,18 +498,19 @@ internal void ExecutePackageDownload(string name, PackageVersion package, string
             var pmExt = DynamoViewModel.Model.GetPackageManagerExtension();
             if (result == MessageBoxResult.OK)
             {
+                package.full_dependency_ids.Count == package.full_dependency_versions.Count
```
pinzart90, line 501: I wonder what the best assert utility would be to check that these 2 arrays are in sync...
@aparajit-pratap any ideas ?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
```diff
@@ -498,18 +498,19 @@ internal void ExecutePackageDownload(string name, PackageVersion package, string
...
                     try
                     {
-                        var depVersion = package.full_dependency_versions[i];
                         return Model.GetPackageVersionHeader(dep._id, depVersion);
                     }
                     catch
                     {
                         MessageBoxService.Show(
-                            String.Format(Resources.MessageFailedToDownloadPackageVersion, dep._id),
+                            String.Format(Resources.MessageFailedToDownloadPackageVersion, dep._id, depVersion),
```
pinzart90, line 513: THe message looks like this
```
Failed to download version {0} of package with id: {1}.  Please try again and report the package if you continue to have problems.
```

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
aparajit-pratap: `Debug.Assert`?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
```diff
@@ -498,18 +498,19 @@ internal void ExecutePackageDownload(string name, PackageVersion package, string
...
                     try
                     {
-                        var depVersion = package.full_dependency_versions[i];
                         return Model.GetPackageVersionHeader(dep._id, depVersion);
                     }
                     catch
                     {
                         MessageBoxService.Show(
-                            String.Format(Resources.MessageFailedToDownloadPackageVersion, dep._id),
+                            String.Format(Resources.MessageFailedToDownloadPackageVersion, depVersion, dep._id),
```
pinzart90, line 513: ```Failed to download version {0} of package with id: {1}.  Please try again and report the package if you continue to have problems.``

aparajit-pratap (commented):
How do these changes avert the crash?

pinzart90:
Calling String.Format with the wrong number of arguments will cause an exception to be thrown.

Resources.MessageFailedToDownloadPackageVersion expects 2 arguments

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
aparajit-pratap: Did it crash here at `dep._id` before? Was `dep` `null` in that case?

pinzart90: Calling `String.Format` with the wrong number of arguments will cause an exception to be thrown. Resources.MessageFailedToDownloadPackageVersion expects 2 args and we were only passing one (dep._id)

aparajit-pratap (approved):

PR #12033, opened 2021-09-13
Package Manager Refresh - Consumer Experience

OliverEGreen:
### Purpose

This PR restyles Dynamo's Package Manager Search window.
It contains mostly xaml changes and many icons used in the new design.

snLp4mMUaQ

### Declarations

### Reviewers

@QilongTang 

### FYIs

@SHKnudsen 
@mjkkirschner


src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
```diff
@@ -280,6 +278,17 @@ public bool HasNoResults
...
+        /// <summary>
+        /// Checks if the user already has a package by the given name.
+        /// </summary>
+        /// <param name="packageName"></param>
+        /// <returns></returns>
+        internal bool UserAlreadyHasPackageInstalled(string packageName)
+        {
+            PackageManagerExtension packageManagerExtension = PackageManagerClientViewModel.PackageManagerExtension;
+            List<string> localPackageNames = packageManagerExtension.PackageLoader.LocalPackages.Select(x => x.Name).ToList();
+            return localPackageNames.Contains(packageName) || Downloads.Select(x => x.Name).Contains(packageName);
```
mjkkirschner, line 290: I think you should also be checking package state - fyi @pinzart90 @sm6srw  thoughts?

src/Tools/DynamoShapeManager/Utilities.cs
```diff
@@ -278,6 +278,8 @@ public static Version GetInstalledAsmVersion2(IEnumerable<Version> versions, ref
                 // Fallback mechanism, look inside libg folders if any of them contains ASM dlls.
                 foreach (var v in versions)
                 {
+                    if (v == null) continue;
```
mjkkirschner, line 281: when did this happen?

mjkkirschner:
The title *requirements* and showing all the different hosts seems confusing to me - does that mean I need all the different host products to use IronPython? It seems like it should say supported on?

src/Tools/DynamoShapeManager/Utilities.cs
OliverEGreen: About a week ago I was seeing some nulls in here when using the package manager search as normal.

OliverEGreen:
> The title _requirements_ and showing all the different hosts seems confusing to me - does that mean I need all the different host products to use IronPython? It seems like it should say supported on? (commenting on your gif)

It's a good question - I went according to the existing functionality and this mockup. Perhaps we could ping the UI/UX team?

image

mjkkirschner:
@Jingyi-Wen 🏓

src/Tools/DynamoShapeManager/Utilities.cs
mjkkirschner: that's odd - can you get a stacktrace please? I can't think of a good reason that the package manager search is looking for installations of ASM - that seems like a mistake if so.

Jingyi-Wen:
> @Jingyi-Wen 🏓
@mjkkirschner This is a good point! The reason we decided to use "Requirements" was that we what to cover everything including host dependencies, outside package dependencies (currently we have those two as "Full download dependencies" in the Package manager website), plus Python version. Would "Supported on" makes sense for additional packages?

src/DynamoCoreWpf/UI/Converters.cs
```diff
@@ -312,6 +302,29 @@ public class NonEmptyStringToCollapsedConverter : IValueConverter
...
+    public class InstalledButtonTextConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter,
+            CultureInfo culture)
+        {
+            if (!(value is bool booleanValue)) return null;
+            
+            return booleanValue
+                ? Properties.Resources.PackageManagerInstalled
+                : Properties.Resources.PackageManagerInstall;
```
QilongTang, line 318: use `Wpf.Properties.Resources` so you dont have to define these in DynamoCore projects at all :)

src/DynamoCore/Properties/Resources.resx
```diff
@@ -745,6 +745,23 @@ Parameter name: {0}</value>
...
+    <value>Install</value>
+    <comment>Used on the package manager search result card. If the package is not installed, the button will say 'Install'.</comment>
+  </data>
+  <data name="PackageManagerInstalled" xml:space="preserve">
+    <value>Installed</value>
+    <comment>Used on the package manager search result card. If the package is already installed, the button will say 'Installed'.</comment>
+  </data>
+  <data name="PackageManagerPackageHasNoDependencies" xml:space="preserve">
+    <value>No dependencies.</value>
+  </data>
```
QilongTang, line 764: These shouldn't appear here, please remove, see my comments below why

src/DynamoCore/Properties/Resources.resx
```diff
@@ -745,6 +745,23 @@ Parameter name: {0}</value>
...
+  <data name="PackageManagerPackageNew" xml:space="preserve">
+    <value>New</value>
+  </data>
+  <data name="PackageManagerInstall" xml:space="preserve">
+    <value>Install</value>
+    <comment>Used on the package manager search result card. If the package is not installed, the button will say 'Install'.</comment>
+  </data>
+  <data name="PackageManagerInstalled" xml:space="preserve">
+    <value>Installed</value>
```
QilongTang, line 759: Duplicate with `PackageDownloadStateInstalling`

src/DynamoCoreWpf/UI/Themes/Modern/DynamoConverters.xaml
```diff
@@ -63,6 +63,7 @@
     <controls:PackageSearchStateToStringConverter x:Key="PackageSearchStateToStringConverter" />
     <controls:EmptyStringToCollapsedConverter x:Key="EmptyStringToCollapsedConverter" />
     <controls:NonEmptyStringToCollapsedConverter x:Key="NonEmptyStringToCollapsedConverter" />
+    <controls:InstalledButtonTextConverter x:Key="InstalledButtonTextConverter" />
```
QilongTang, line 66: How about the converters around dates, do we need to add them here?

src/DynamoCoreWpf/UI/Converters.cs
OliverEGreen: Done!

src/DynamoCore/Properties/Resources.resx
OliverEGreen: Removed duplicate

src/DynamoCoreWpf/UI/Themes/Modern/DynamoConverters.xaml
OliverEGreen: As in the PrettyDateConverter? That one existed already, I'm seeing it on line 85.

src/DynamoCore/Properties/Resources.en-US.resx
```diff
@@ -810,4 +810,19 @@ This package has not been loaded because another conflicting package was loaded
   <data name="NodeUnhandledMsg" xml:space="preserve">
     <value>Unhandled 'DummyNode.NodeNature' value: {0}</value>
   </data>
+  <data name="HideWiresPopupMenuItem" xml:space="preserve">
+    <value>Hide Wires</value>
+  </data>
+  <data name="UnhideWiresPopupMenuItem" xml:space="preserve">
+    <value>Unhide Wires</value>
+  </data>
```
QilongTang, line 818: I think I intentionally removed these two

src/DynamoCoreWpf/Views/PackageManager/PackageManagerSearchView.xaml.cs
```diff
@@ -123,5 +129,53 @@ private void OnRequestShowFileDialog(object sender, PackagePathEventArgs e)
...
+        /// <summary>
+        /// When the use clicks close on this window, closes the window.
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="e"></param>
+        private void CloseButton_Click(object sender, RoutedEventArgs e)
+        {
+            Dynamo.Logging.Analytics.TrackEvent(
+                Actions.Close,
+                Categories.PackageManagerOperations);
```
QilongTang, line 153: Man, appreciate this..

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
QilongTang: @mjkkirschner @pinzart @sm6srw When you say states, do you mean if the package is scheduled for uninstall etc?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
```diff
@@ -451,6 +481,42 @@ public void SetSortingDirection(object sortingDir)
...
+        {
+            // to be implemented in future PR   
+        }
+
+        /// <summary>
+        /// Clears the search text box and resets the search.
+        /// </summary>
+        public void ClearSearchTextBox(object obj)
+        {
+            SearchText = "";
```
QilongTang, line 498: Let's use `String.Empty` if we can

src/DynamoCore/Properties/Resources.en-US.resx
```diff
@@ -810,4 +810,19 @@ This package has not been loaded because another conflicting package was loaded
...
+  <data name="PackageManagerPackageHasNoDependencies" xml:space="preserve">
+    <value>No dependencies.</value>
+  </data>
```
QilongTang, line 821: I think this one can also be removed

src/DynamoCore/Properties/Resources.en-US.resx
```diff
@@ -810,4 +810,19 @@ This package has not been loaded because another conflicting package was loaded
...
+  <data name="PackageManagerPackageUpdated" xml:space="preserve">
+    <value>Updated</value>
+  </data>
+  <data name="PackageManagerPackageNew" xml:space="preserve">
+    <value>New</value>
+  </data>
```
QilongTang, line 827: Can you indicate where are these two used?

OliverEGreen: On the bubble next to the package in the results view. Appears if it's New or recently Updated

QilongTang: Then I would also move them to `DynamoCoreWpf` because they are utilized there but not in DynamoCore project?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchElementViewModel.cs
```diff
@@ -41,6 +50,46 @@ public PackageManagerSearchElementViewModel(PackageManagerSearchElement element,
...
+        /// <param name="element">A PackageManagerSearchElement</param>
+        /// <param name="canLogin">A Boolean used for access control to certain internal packages.</param>
+        /// <param name="isPackageInstalledFunction">A function that we pass to this constructor which takes a string and returns a Boolean.</param>
+        public PackageManagerSearchElementViewModel(PackageManagerSearchElement element, bool canLogin, Func<string,bool> isPackageInstalledFunction) : this(element, canLogin)
+        {
+            this.isPackageInstalled = isPackageInstalledFunction;
+            this.IsInstalled = isPackageInstalled(this.Model.Name);
+        }
+
+        private bool isInstalled;
```
pinzart90, line 70: I think we should use the existing `PackageLoadState` APIs to tell if a user has already "installed" a package or not.

When a Package is successfully downloaded and Loaded into Dynamo it gets assigned a "LoadState.State == ackageLoadState.StateTypes.Loaded"
If a package is downloaded but Loading into Dynamo fails...it gets assigned "LoadState.State == ackageLoadState.StateTypes.Error"
If a package is downloaded but intentionally not Loaded into dynamo it gets assigned "LoadState.State == ackageLoadState.StateTypes.Unloaded"
Another thing to keep in mind is that a Package can be Built-in and in one of the states above, but I guess a user should be able to install a non-built-in package with the same name

@OliverEGreen @mjkkirschner @QilongTang 
When should the user be able to install/reinstall a package xxx with version yyy?
1. The package xxx@yyy does not exist (i.e not in the LocalPackages) - Yes
2. The package xxx@yyy is in an error state - Yes
3. The package xxx@yyy is loaded but built-in - Yes
4. The package xxx@yyy is unloaded - No
5. The pacakge xxx@yyy has a state == None (i.e downloaded but not processed by Dynamo yet) - Not sure

QilongTang (commented):
Some comments, more comments in `PackageManagerSearchView.xaml` to help the review would be appreciated. Looks solid

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
```diff
@@ -734,6 +733,21 @@ public static string FormatPackageVersionList(IEnumerable<Tuple<PackageHeader, P
         /// <param name="packageDownloadHandle">package download handle</param>
         internal virtual Task<(PackageDownloadHandle handle, string downloadPath)> Download(PackageDownloadHandle packageDownloadHandle)
         {
+            // We only want to display the last 3 downloaded packages to the user
+            // in the form of toast notifications.
+            if (Downloads.Count >= 3)
+            {
+                // Iterates backwards through the downloads list and removes items
```
pinzart90, line 740: you should be able to use `Downloads.RemoveRange(index: 2, count: Downloads.Count - 2);` instead of this whole for loop

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
pinzart90: PackageLoadState.StateTypes hold the current load states
PackageLoadState.ScheduledTypes hold the scheduled types (ex. scheduled for deletion)
You should only have to do with PackageLoadState.StateTypes here...

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchElementViewModel.cs
```diff
@@ -41,6 +50,46 @@ public PackageManagerSearchElementViewModel(PackageManagerSearchElement element,
...
+                isInstalled = value;
+                RaisePropertyChanged(nameof(IsInstalled));
+            }
+        }
+
+        /// <summary>
+        /// Checks whether the user has already installed a package with the given name.
+        /// </summary>
+        /// <returns>A Boolean indicating whether the user already has a package installed with the given name.</returns>
+        internal bool IsPackageInstalled()
```
pinzart90, line 89: I do not see the point of this method.
You could just use the public method IsInstalled instead.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
```diff
@@ -451,6 +481,42 @@ public void SetSortingDirection(object sortingDir)
...
+        /// appears at the base of the window. This command fires when the user clicks to dismiss
+        /// one of these toast notifications.
+        /// </summary>
+        /// <param name="obj"></param>
+        public void ClearDownloadToastNotification(object obj)
+        {
+            if (!(obj is PackageDownloadHandle packageDownloadHandle)) return;
+
+            PackageDownloadHandle packageDownloadHandleToRemove = PackageManagerClientViewModel.Downloads
+                .FirstOrDefault(x => x.Id == packageDownloadHandle.Id);
```
pinzart90, line 512: x is always valid ?

src/Tools/DynamoShapeManager/Utilities.cs
OliverEGreen: No luck getting that stacktrace I'm afraid. Should it happen again, I'll send you a screenshot immediately :)

src/DynamoCore/Properties/Resources.en-US.resx
OliverEGreen: Removed for now, will get these once #12050 is merged :)

src/DynamoCore/Properties/Resources.en-US.resx
OliverEGreen: Removed from DynamoCore resources

src/DynamoCore/Properties/Resources.en-US.resx
OliverEGreen: Moved

src/DynamoCoreWpf/Views/PackageManager/PackageManagerSearchView.xaml
```diff
@@ -1,559 +1,961 @@
...
-                <Setter Property="BorderBrush" Value="{x:Null}" />
+            <ResourceDictionary.MergedDictionaries>
+                <ui:SharedResourceDictionary Source="{x:Static ui:SharedDictionaryManager.DynamoModernDictionaryUri}" />
+                <ui:SharedResourceDictionary Source="{x:Static ui:SharedDictionaryManager.DynamoConvertersDictionaryUri}" />
+                <ui:SharedResourceDictionary Source="{x:Static ui:SharedDictionaryManager.DynamoColorsAndBrushesDictionaryUri}" />
+            </ResourceDictionary.MergedDictionaries>
+        </ResourceDictionary>
+    </Window.Resources>
 
+    <Grid Name="mainGrid">
```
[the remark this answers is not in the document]
OliverEGreen: Comments incoming!

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
OliverEGreen: Have just tested and this doesn't quite do what we want - this only replaces the last item if there's more than 3 in the list. The UI here displays the list like a stack, so the most recent downloads appear at the bottom, above that the 2nd most recent, above that the 3rd etc. The oldest item on the stack is 'popped' off when a new download happens.

OliverEGreen: Sorry for the slightly long GIF, but this is what happens:

xLrqrxNX9S

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchElementViewModel.cs
OliverEGreen: I'll agree this was a difficult one to get working, because there's not great visibility between the PackageManagerSearchElementViewModel and the PackageManagerSearchViewModel.

If there's a straightforward way for the PackageManagerSearchElement to determine whether it's installed, we could clear out a lot of code here.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
OliverEGreen: I think so: if the button is visible for the user to click (and hence fire this method), then there should always be at least one item in the PackageManagerClientViewModel.Downloads collection.
We then check against null in the following statement.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
OliverEGreen: @pinzart90 Have modified the code to include a check against LoadState.State. 

https://github.com/DynamoDS/Dynamo/blob/0b9424dd795c84d47e793fc9338728d80184ae47/src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs#L286-L303

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
pinzart90: Oh, I misunderstood how that list in interpreted by the UI
Then you should be able to use `Downloads.RemoveRange(index: 0, count: Downloads.Count - 2);`
That should basically remove from the beginning of the list (i.e index 0) until only 2 elements remain in the list

OliverEGreen: Works great, thanks :)

QilongTang (approved):

QilongTang:
I noticed there are a few follow up improvements we can take on as suggested by @pinzart and I suspect we shall have some user feedback after people trying this new UI out. After some discussion with @OliverEGreen We agreed to take further code improvements and any bug fixing in follow up PRs.

PR #12039, opened 2021-09-15
PR - Jira Issues: [DYN-4095] Cleanup layout rearranges pin and node

M-JULIANI:
### Description
Fixes a previous bug visible when running `Cleanup Layout`. When selecting a single node and running the command multiple
times, any connector pin in the graph moved upwards.

Fixed behaviour: 
WireIssueFix-CleanLayoutPin

This PR also adds a `HasUnsavedChanges` flag when calling `BreakConnection` from a port.

Fixed behaviour: 
WireIssueFix-UndoFlagAddedToBreakConnection-FromPort

### Declarations

Check these if you believe they are true

- [ ] The codebase is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.
- [ ] This PR modifies some build requirements and the readme is updated

### Reviewers

(FILL ME IN) Reviewer 1  (If possible, assign the Reviewer for the PR)

(FILL ME IN, optional) Any additional notes to reviewers or testers.

### FYIs

(FILL ME IN, Optional) Names of anyone else you wish to be notified of

PR #12040, opened 2021-09-15
PR - Jira Issues: [DYN-4091] Dynamo crashes when dragging a pin w/ ctrl key pressed.

M-JULIANI:
*@QilongTang  please merge the following PR to master first: https://github.com/DynamoDS/Dynamo/pull/12039.

### Description
Fixes the crash issue by:
Adding gates in `StateMachine`-> `PasteSelection` and in `DynamoModel`->`Paste` to ensure that ConnectorPins aren't pasted.

Fix outcome: 
WireIssueFix-PastePinCrash

### Declarations

### Reviewers
@QilongTang 

### FYIs
@SHKnudsen 
@Amoursol

### DIFF
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -2385,7 +2385,7 @@ namespace Dynamo.Models
         public void Paste(Point2D targetPoint, bool useOffset = true)
         {
             //When called from somewhere other than StateMachine and only ConnectorPins are selected.
-            if (!ClipBoard.Where(m => !(m is ConnectorPinModel)).Select(m => m).Any())
+            if (ClipBoard.All(m => m is ConnectorPinModel))
             {
                 return;
             }
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -2384,6 +2384,11 @@ namespace Dynamo.Models
         /// directly in this point. </param>
         public void Paste(Point2D targetPoint, bool useOffset = true)
         {
+            //When called from somewhere other than StateMachine and only ConnectorPins are selected.
+            if (!ClipBoard.Where(m => !(m is ConnectorPinModel)).Select(m => m).Any())
+            {
+                return;
+            }
             if (useOffset)
             {
                 // Provide a small offset when pasting so duplicate pastes aren't directly on top of each other
--- a/src/DynamoCoreWpf/ViewModels/Core/StateMachine.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/StateMachine.cs
@@ -948,7 +948,7 @@ namespace Dynamo.ViewModels
 
                 model.Copy();
                 // Prevents Paste from being called when only ConnectorPins are selected.
-                if (model.ClipBoard.Where(m => !(m is ConnectorPinModel)).Select(m => m).Any())
+                if (!model.ClipBoard.All(m => m is ConnectorPinModel))
                 {
                     model.Paste(targetPoint, false);
                     owningWorkspace.DynamoViewModel.UndoCommand.RaiseCanExecuteChanged();
--- a/src/DynamoCoreWpf/ViewModels/Core/StateMachine.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/StateMachine.cs
@@ -947,7 +947,8 @@ namespace Dynamo.ViewModels
                 var oldClipboardData = model.ClipBoard.ToList();
 
                 model.Copy();
-                if (model.ClipBoard.Any())
+                // Prevents Paste from being called when only ConnectorPins are selected.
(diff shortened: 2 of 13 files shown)


src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -2384,6 +2384,11 @@ public void Paste()
         /// directly in this point. </param>
         public void Paste(Point2D targetPoint, bool useOffset = true)
         {
+            //When called from somewhere other than StateMachine and only ConnectorPins are selected.
+            if (!ClipBoard.Where(m => !(m is ConnectorPinModel)).Select(m => m).Any())
```
QilongTang, line 2388: Maybe use `All()` instead of the more complex way you are using?

src/DynamoCoreWpf/ViewModels/Core/StateMachine.cs
```diff
@@ -947,7 +947,8 @@ internal void PasteSelection(Point2D targetPoint)
                 var oldClipboardData = model.ClipBoard.ToList();
 
                 model.Copy();
-                if (model.ClipBoard.Any())
+                // Prevents Paste from being called when only ConnectorPins are selected.
+                if (model.ClipBoard.Where(m => !(m is ConnectorPinModel)).Select(m => m).Any())
```
QilongTang, line 951: same here, maybe use `All()`?

QilongTang (commented):
LGTM with two comments

src/DynamoCore/Models/DynamoModel.cs
M-JULIANI: Here that makes sense. Fixed.

QilongTang:
Well, please also resolve the merge conflicts.. @M-JULIANI in the future maybe not build your branches on top of each other, otherwise once you address comments in earlier PRs, all of your PRs in a chain are affected this way

src/DynamoCoreWpf/ViewModels/Core/StateMachine.cs
M-JULIANI: Here doing so changes the functionality ever so slightly. If we apply `All()` here, whenever there is a pin in the selection, nothing drags. If we keep the old conditional whatever else that isn't a pin copies over. What do you prefer? @QilongTang

QilongTang: Hey @M-JULIANI I was only reading your comments and think maybe use the following, what do you think?
```
if (!ClipBoard.All(m => m is ConnectorPinModel))
```

M-JULIANI:
```
     if (ClipBoard.All(m => m is ConnectorPinModel))
            {
                return;
            }
```

M-JULIANI: ^that is what we need to avoid having pins pasted, which throws an error.. However using this same logic at `StateMachine` prevents other nodes in the selection form being copied as well. Is that expected behaviour or do you want other things in selection to copy over? If that is the case we need to leave the conditional alone at `StateMachine`. @QilongTang

M-JULIANI: All I am saying is this comment has to be discarded if we want other things to be copied (and not pins) when pins are part of the selection. @QilongTang

QilongTang: Gotcha! @Amoursol  @Jingyi-Wen any preference?

Jingyi-Wen: I also think we should have other things copied as well

Amoursol: Agreed - please copy all the things :)

QilongTang: Thanks! @M-JULIANI Please go ahead with this decision, team will re-test after this PR merged.

M-JULIANI: Sound good, comment resolved! @QilongTang

QilongTang: Hey @M-JULIANI I am still not understanding here.. Why dont we use the simplified version since you only mean as long as the `ClipBoard` does not only contain pins, we should proceed right?
```
if (!ClipBoard.All(m => m is ConnectorPinModel))
{
...
}
```

M-JULIANI: Fair enough, just tested it and it's equivalent :)

M-JULIANI: Was confused because in the other block it was the opposite. Fixes made.

QilongTang (approved):

PR #12041, opened 2021-09-16
PR - Codebase modification: Changing visibility property in ConnectorViewModel to work with `IsCollapsed` from groups

M-JULIANI:
@QilongTang please make sure https://github.com/DynamoDS/Dynamo/pull/12039 and https://github.com/DynamoDS/Dynamo/pull/12040 (in that order) are merged to master prior to this PR being merged.

### Description
This PR changes the `IsVisible` property on the `ConnectorModel`, `ConnectorViewModel`, `ConnectorPinViewModel` and related classes to the inverse boolean `IsCollapsed`. This has repercussions across the codebase as you can see from the number of files changed.

*This PR stops short of ensuring that all pins work seamlessly with groups. Annotation models do _not_ contain pins, and thus pins don't behave as expected other notes and nodes do (with groups) currently. That is pins don't currently move with groups.
This can be tackled in a separate PR:  the thought is the current PR is about bringing the codebase up-to-date with the inverse logic change and ensuring that the change doesn't break anything. (I ran some unit tests but would not be surprised if there are regressions that I didn't see in the unit tests I ran).

### Declarations

### Reviewers
@QilongTang 

### FYIs
@SHKnudsen 
@Amoursol

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Core/ConnectorPinViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/ConnectorPinViewModel.cs
@@ -122,6 +122,8 @@ namespace Dynamo.ViewModels
             get { return model.Y- OneThirdWidth; }
             set
             {
+                //Through trial and error using the OneThirdWidth value to offset the pin location works 
+                //better than using OneHalf.
                 model.Y = value + OneThirdWidth;
                 RaisePropertyChanged(nameof(Top));
             }
--- a/src/DynamoCoreWpf/ViewModels/Core/ConnectorPinViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/ConnectorPinViewModel.cs
@@ -119,10 +119,10 @@ namespace Dynamo.ViewModels
         /// </summary>
         public double Top
         {
-            get { return model.Y; }
+            get { return model.Y- OneThirdWidth; }
             set
             {
-                model.Y = value;
+                model.Y = value + OneThirdWidth;
                 RaisePropertyChanged(nameof(Top));
             }
         }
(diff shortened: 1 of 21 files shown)


src/DynamoCoreWpf/ViewModels/Core/ConnectorPinViewModel.cs
```diff
@@ -138,39 +138,42 @@ public bool IsSelected
...
-                RaisePropertyChanged(nameof(IsVisible));
-                RaisePropertyChanged(nameof(IsPartlyVisible));
+                if (isCollapsed == value)
+                {
+                    return;
+                }
+
+                isCollapsed = value;
+                RaisePropertyChanged(nameof(IsCollapsed));
+                RaisePropertyChanged(nameof(IsTemporarilyVisible));
```
QilongTang, line 155: Can you remind me why we need to raise change here if the value is not touched?

src/DynamoCoreWpf/ViewModels/Core/ConnectorPinViewModel.cs
```diff
@@ -138,39 +138,42 @@ public bool IsSelected
...
+                return isTemporarilyVisible;
             }
             set
             {
-                isPartlyVisible = value;
-                RaisePropertyChanged(nameof(IsPartlyVisible));
-                RaisePropertyChanged(nameof(IsVisible));
+                isTemporarilyVisible = value;
+                RaisePropertyChanged(nameof(IsTemporarilyVisible));
+                RaisePropertyChanged(nameof(IsCollapsed));
```
QilongTang, line 176: Same comment here

QilongTang (commented):
Two comments then LGTM

src/DynamoCoreWpf/ViewModels/Core/ConnectorPinViewModel.cs
M-JULIANI: Fixed- this was a leftover.

src/DynamoCoreWpf/ViewModels/Core/ConnectorPinViewModel.cs
M-JULIANI: Fixed- this was a leftover.

QilongTang (approved):

PR #12042, opened 2021-09-16
Fix crash bug when deleting a node with note pinned will crash Dynamo

zeusongit:
### Purpose

Fix a crash bug when deleting a node with note pinned will crash Dynamo
(Will attach a snapshot soon)

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Core/NoteViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/NoteViewModel.cs
@@ -362,7 +362,7 @@ namespace Dynamo.ViewModels
             if (PinnedNode != null)
             {
                 PinnedNode.PropertyChanged -= PinnedNodeViewModel_PropertyChanged;
-                PinnedNode.RequestsSelection -= PinnedNodeViewModel_OnPinnedNodeSelected;
+                PinnedNode.Selected -= PinnedNodeViewModel_OnPinnedNodeSelected;
                 PinnedNode.Removed -= PinnedNodeViewModel_OnPinnedNodeRemoved;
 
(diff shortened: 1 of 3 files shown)


src/DynamoCoreWpf/ViewModels/Core/NoteViewModel.cs
```diff
@@ -344,10 +344,13 @@ private void SubscribeToPinnedNode()
             Model.PinnedNode.PropertyChanged += PinnedNodeModel_PropertyChanged;
             PinnedNode.PropertyChanged += PinnedNodeViewModel_PropertyChanged;
 
-            // Subscribe to pinnedNode.RequestSelection (fires before node is selected)
+            // Subscribe to PinnedNode.Selected (fires before node is selected)
             // so that this note is added to the selection
             PinnedNode.Selected += PinnedNodeViewModel_OnPinnedNodeSelected;
```
QilongTang, line 349: Maybe outside of this task but can you check if this is unsubscribed correctly?

QilongTang (commented):
LGTM with a comment

src/DynamoCoreWpf/ViewModels/Core/NoteViewModel.cs
QilongTang: When reading the code from unsubscribe function, seems two events about select are misaligned, would you correct this case?

zeusongit: Updated

QilongTang (approved):

PR #12055, opened 2021-09-23
DYN-777 : performance improvements demo

BogdanZavu:
We don't always need to execute UI-related code and render UI components.
So we are going to load UI elements for nodes ( main content plus ports ) on demand when the user sets focus on a specific control.
If we are doing this Dynamo is ~= 50% faster when we are opening a graph and consumes ~= 50% less memory.
UI-less performance should be improved as well but probably not that much.
Note that even if full UI components are not loaded and a node looks empty graph execution works as expected.
Another nice aspect of this is that this work can be done incrementally one control after another.

### DIFF
--- a/src/DynamoCoreWpf/Views/Core/NodeView.xaml.cs
+++ b/src/DynamoCoreWpf/Views/Core/NodeView.xaml.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Diagnostics;
 using System.Windows;
 using System.Windows.Controls;
@@ -28,6 +29,7 @@ namespace Dynamo.Controls
         private NodeViewModel viewModel = null;
         private PreviewControl previewControl = null;
         private const int previewDelay = 1000;
+        private bool inoutportsInitialized = false;
 
         /// <summary>
         /// If false - hides preview control until it will be explicitly shown.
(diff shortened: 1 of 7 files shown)


src/Libraries/CoreNodeModelsWpf/NodeViewCustomizations/DoubleSlider.cs
```diff
@@ -7,18 +7,39 @@ namespace CoreNodeModelsWpf.NodeViewCustomizations
...
         {
-            var slider = new DynamoSlider(model, nodeView)
+            viewNode = nodeView;
+            sliderModel = model;
+            viewNode.MouseEnter += ViewNode_MouseEnter;
+        }
+
+        private void ViewNode_MouseEnter(object sender, System.Windows.Input.MouseEventArgs e)
+        {
+            ConstructView();
```
mjkkirschner, line 22: so this is a little worrying - I think we'd want to add this to some base class to help authors not need to do this themselves, and we'll call ConstructView or whatever for them?

src/DynamoCoreWpf/Views/Core/NodeView.xaml.cs
```diff
@@ -65,6 +67,41 @@ private set
...
+                    inoutportsInitialized = true;
+                    CreateInOutPortsControls("inputPortControl", 0, ViewModel.InPorts);
+                    CreateInOutPortsControls("outputPortControl", 2, ViewModel.OutPorts);
+                }
+            }
+        }
+
+        private void NodeView_MouseEnter(object sender, MouseEventArgs e)
+        {
+            PrepareInOutPortsControls();
```
mjkkirschner, line 102: what about destroying the ports control as well after zoom out or something to free up memory?

src/Libraries/CoreNodeModelsWpf/NodeViewCustomizations/DoubleSlider.cs
BogdanZavu: something like that yes, we'll need a better mechanism to handle this different than mouse enter event

src/DynamoCoreWpf/Views/Core/NodeView.xaml.cs
BogdanZavu: that too yes

mjkkirschner:
@BogdanZavu I'm going to mark this draft, it seems to be making the tests run for a long time, if you plan to continue it then unmark.

BogdanZavu:
@mjkkirschner I don't plan on continuing with this, for now. Sorry, I just have made it a draft in the first place.

PR #12056, opened 2021-09-23
Node Search performance improvements

pinzart90:
Simplify search to improve performance
Changes:
1. Fix double search event caused by UI events
2. Do not add the whole search text if there is only one word (using whitespace as separator)
3. Limit search to 300 characters
4. use OrdinalIgnoreCase option in IndexOf calls

Note:
Difference between culture sensitive and ordinal serarch:
https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings#string-comparisons-that-use-the-current-culture

I would opt for OrdinalIgnoreCase because we get better performance...and the order of the returned results is not really a deal breaker (I think...)


src/DynamoCoreWpf/Controls/IncanvasSearchControl.xaml
```diff
@@ -105,8 +105,7 @@
                              Text="{Binding Path=SearchText, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}"
                              MinWidth="200"
                              CaretBrush="{StaticResource CommonSidebarTextColor}"
-                             Margin="26,0,0,-1"
-                             TextChanged="OnSearchTextBoxTextChanged" />
+                             Margin="26,0,0,-1"/>
```
pinzart90, line 108: We already bind the Text control to the "SearchText" Path.
When the text is changed, the "SearchText" property's setter will be called...and that is when we will execute the Search command

src/DynamoCore/Search/SearchDictionary.cs
```diff
@@ -13,6 +13,7 @@ namespace Dynamo.Search
     public class SearchDictionary<V>
     {
         private ILogger logger;
+        private static int LIMIT_SEARCH_TAG_SIZE = 100;
```
mjkkirschner, line 16: could this be too small for some localized searches?

src/DynamoCore/Search/SearchDictionary.cs
```diff
@@ -342,7 +343,7 @@ internal void RebuildTagDictionary()
                             tagAndWeight =>
                                 new
                                 {
-                                    Tag = tagAndWeight.Key,
+                                    Tag = tagAndWeight.Key.Substring(0, tagAndWeight.Key.Length > LIMIT_SEARCH_TAG_SIZE ? LIMIT_SEARCH_TAG_SIZE : tagAndWeight.Key.Length),
```
mjkkirschner, line 346: is this actually useful at this point? Does the cost of all these extra conditional undo the potential savings?

src/DynamoCore/Search/SearchDictionary.cs
```diff
@@ -379,12 +380,13 @@ internal IEnumerable<V> Search(string query, int minResultsForTolerantSearch = 0
             query = query.ToLower();
 
             var subPatterns = SplitOnWhiteSpace(query);
-
-            // Add full (unsplit by whitespace) query to subpatterns
-            var subPatternsList = subPatterns.ToList();
-            subPatternsList.Insert(0, query);
-            subPatterns = (subPatternsList).ToArray();
-
+            if (subPatterns.Length > 1)// More than one word
```
mjkkirschner, line 383: 👍

src/DynamoCoreWpf/Controls/IncanvasSearchControl.xaml
mjkkirschner: could this also be fixed by making this a one way binding?

mjkkirschner:
@pinzart90 it looks good after the tests pass, can I ask you to manually check that regular library search still works?

mjkkirschner (approved):

Amoursol:
@pinzart do we have any metrics on how much better search is before and after this 🙏  ?

src/DynamoCore/Search/SearchDictionary.cs
```diff
@@ -379,12 +380,13 @@ internal IEnumerable<V> Search(string query, int minResultsForTolerantSearch = 0
...
+            {
+                // Add full (unsplit by whitespace) query to subpatterns
+                var subPatternsList = subPatterns.ToList();
+                subPatternsList.Insert(0, query);
+                subPatterns = (subPatternsList).ToArray();
```
aparajit-pratap, line 388: Could you explain with an example? If the query is `all elements of category`, the subpatterns list will be `["all", "elements", "of", "category"]`, then the subpatterns list will now be `["all elements of category", "all", "elements", "of", "category"]`?

src/DynamoCore/Search/SearchDictionary.cs
aparajit-pratap: Same q here.

src/DynamoCore/Search/SearchDictionary.cs
pinzart90: Maybe...the value is definitely up for debate. I wonder if it would actually be best to warn the user if this limit is surpassed (I am not sure when or where)
I can take a look at the tags we got from Jostein to get a better idea of how many characters there are in a "normal" situation

src/DynamoCore/Search/SearchDictionary.cs
pinzart90: well...after all search tags have been added/removed  `RebuildTagDictionary` is called only once.
Performance penalty here should be negligible, compared to the constant searches on each text change.

src/DynamoCore/Search/SearchDictionary.cs
pinzart90: that is correct.
I think the logic behind it was to look for the full text first (a match with the full text should have the highest priority).

pinzart90:
@mjkkirschner I am working now on adding a performance test...and I will also check that existing functionality is not broken
@Amoursol  I will get you those numbers

aparajit-pratap (approved):

src/DynamoCoreWpf/Controls/IncanvasSearchControl.xaml
pinzart90: Changing around the binding types broke other parts of the UI (ex. making it one way did not update the Textbox initial value "Search" anymore)

pinzart90:
Here are some numbers @Amoursol 
These have been generated on my machine

With Fix "all elements of category" 5 times:
Searching for: "all elements of category", [Entries:1254, Tags:28110] : 386ms -> 2
Searching for: "all elements of category", [Entries:1254, Tags:28110] : 397ms -> 2
Searching for: "all elements of category", [Entries:1254, Tags:28110] : 368ms -> 2
Searching for: "all elements of category", [Entries:1254, Tags:28110] : 381ms -> 2
Searching for: "all elements of category", [Entries:1254, Tags:28110] : 378ms -> 2

Without Fix "all elements of category" 5 times:
Searching for: "all elements of category", [Entries:1254, Tags:28110] : 854ms -> 2
Searching for: "all elements of category", [Entries:1254, Tags:28110] : 822ms -> 2
Searching for: "all elements of category", [Entries:1254, Tags:28110] : 863ms -> 2
Searching for: "all elements of category", [Entries:1254, Tags:28110] : 936ms -> 2
Searching for: "all elements of category", [Entries:1254, Tags:28110] : 869ms -> 2

With Fix "category" 5 times:
Searching for: "category", [Entries:1254, Tags:28110] : 78ms -> 20
Searching for: "category", [Entries:1254, Tags:28110] : 80ms -> 20
Searching for: "category", [Entries:1254, Tags:28110] : 71ms -> 20
Searching for: "category", [Entries:1254, Tags:28110] : 66ms -> 20
Searching for: "category", [Entries:1254, Tags:28110] : 71ms -> 20

Without Fix "category" 5 times:
Searching for: "category", [Entries:1254, Tags:28110] : 309ms -> 20
Searching for: "category", [Entries:1254, Tags:28110] : 348ms -> 20
Searching for: "category", [Entries:1254, Tags:28110] : 290ms -> 20
Searching for: "category", [Entries:1254, Tags:28110] : 291ms -> 20
Searching for: "category", [Entries:1254, Tags:28110] : 282ms -> 20

test/DynamoCoreTests/Search/SearchDictionaryTest.cs
```diff
@@ -57,6 +59,43 @@ public void TestAddItems()
...
+            searchDictionary.Search(query4);
+            var query = "az";
+            var results = searchDictionary.Search(query);
+
+            stopwatch.Stop();
+
+            Assert.AreEqual(results.Count(), 20);
+
+            int timeLimit = 260;//ms
+            Assert.IsTrue(Math.Abs(stopwatch.ElapsedMilliseconds - timeLimit) < 0.2 * timeLimit, $"Search time should be within a range of +/- 20% of {timeLimit}ms but we got {stopwatch.ElapsedMilliseconds}ms");
```
pinzart90, line 96: Still might be a bit flaky ...

src/DynamoCore/Search/SearchDictionary.cs
```diff
@@ -13,6 +13,7 @@ namespace Dynamo.Search
     public class SearchDictionary<V>
     {
         private ILogger logger;
+        private static int LIMIT_SEARCH_TAG_SIZE = 300;
```
pinzart90, line 16: Do we all agree that 300 characters is good enough ?
Should we increase ?

Looking through the search tags we got from Jostein, I found the following:
1. The longest search tag is 899 characters
2. There are 34 search tags (out of 6525) that are over 300 characters

pinzart90: ALso this limit will be used in the Package Search window as well..(using package descriptions)

pinzart90: FYI @mjkkirschner @aparajit-pratap

src/DynamoCore/Search/SearchDictionary.cs
pinzart90: Increased to 300

Amoursol:
> Here are some numbers @Amoursol
> These have been generated on my machine...

Ok awesome, so that looks to be roughly a:

- **56%** performance boost on _All Elements of Category_
- **76%** performance boost on _Category_

That's pretty epic! We can discuss further works after 2.13 drops :) Thanks Tibi. 

image

src/DynamoCoreWpf/ViewModels/Search/SearchViewModel.cs
```diff
@@ -214,8 +214,6 @@ internal void Filter()
                 }
                 strBuilder.Append(", ");
             }
-
-            Analytics.LogPiiInfo("Filter-categories", strBuilder.ToString().Trim());
```
aparajit-pratap, line 212: When is this triggered?

src/DynamoCore/Search/SearchDictionary.cs
aparajit-pratap: This is fine by me but I'm still confused - these search tags from Jostein look like node descriptions. I'm curious if the entire description is used as a search tag, then what's the benefit of the search tags included in the `<search></search>` XML?

aparajit-pratap (approved):

src/DynamoCore/Search/SearchDictionary.cs
mjkkirschner: @aparajit-pratap it's to inject extra search terms without adding them to the description.

So I can add `box` to the search terms for Cuboid - without needing to add `box` to the description.

mjkkirschner: @pinzart90 what do you think about making this a hidden preference? I know it's more work, and kind of a pain, but it let's users tweak it if they run into trouble.

mjkkirschner (approved):

src/DynamoCoreWpf/ViewModels/Search/SearchViewModel.cs
pinzart90: Analytics.LogPiiInfo is deprecated and actually does not log anything anymore. I just cleaned up the SearchViewModel class ...I do not think performance is affected by this though..

src/DynamoCore/Search/SearchDictionary.cs
pinzart90: @mjkkirschner I added a new assembly config for it

src/DynamoCore/Search/SearchDictionary.cs
```diff
@@ -15,6 +17,22 @@ public class SearchDictionary<V>
         private ILogger logger;
         private static int LIMIT_SEARCH_TAG_SIZE = 300;
 
+        static SearchDictionary()
+        {
+            try
+            {
+                // Look up search tag limit in the assembly configuration
+                var assemblyConfig = ConfigurationManager.OpenExeConfiguration(Assembly.GetExecutingAssembly().Location);
```
aparajit-pratap, line 25: Is this the `DynamoCore.dll.config` file?

pinzart90: yes

aparajit-pratap (approved):

src/DynamoCore/Search/SearchDictionary.cs
mjkkirschner: oh, sorry @pinzart90 - I was not clear, by hidden preference -I just meant a preference in the dynamosettings.xml with no analog in the UI to control it. This is fine as well, users can still modify this if we want them to test something.

pinzart90: @mjkkirschner if you think there is good enough reason to move to Preferences, then I can do that

aparajit-pratap: I think it would be a good idea to move it to the preference settings file so we have all preferences in one place. That is usually the one file where all such user preferences go.

PR #12069, opened 2021-09-27
Crash Fix.  Add visibility trigger to empty tooltip strings.

saintentropy:
### Purpose

This PR fixes a crash bug when hovering wires after opening the graph in `Manual` mode.  Specifically it hides the tooltips for the hover menu.

### Declarations

### Reviewers

@marimano @QilongTang

### DIFF
--- a/src/DynamoCoreWpf/UI/Converters.cs
+++ b/src/DynamoCoreWpf/UI/Converters.cs
@@ -36,7 +36,7 @@ namespace Dynamo.Controls
     {
         public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
         {
-            if (value == null) return "";
+            if (value == null) return string.Empty;
             string incomingString = value as string;
             return incomingString.Split(new[] { '\r', '\n' }, 2)[0].Trim();
         }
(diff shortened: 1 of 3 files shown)


src/DynamoCoreWpf/ViewModels/Core/ConnectorViewModel.cs
```diff
@@ -520,7 +520,7 @@ private void UpdateConnectorDataToolTip()
                 var portValue = model.Start.Owner.GetValue(model.Start.Index, workspaceViewModel.DynamoViewModel.EngineController);
                 if (portValue is null)
                 {
-                    ConnectorDataTooltip = "N/A";
```
saintentropy, line 523: This can also go back to "N/A" but seems like we should just not show a tooltip if there is nothing to show

src/DynamoCoreWpf/UI/Converters.cs
```diff
@@ -36,6 +36,7 @@ public class ToolTipFirstLineOnly : IValueConverter
     {
         public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
         {
+            if (value == null) return "";
```
QilongTang, line 39: Make sense, would you update to use `String.Empty` where applies, thanks!

QilongTang (commented):
One comment, then LGTM

QilongTang (approved):

PR #12074, opened 2021-09-28
Lazy Load Node Context Menu

OliverEGreen:
### Purpose

This PR improves the way in which all nodes' context menus are loaded, so that they are **constructed when required by the user**, and **disposed immediately after closing**. This lazy-loaded approach is designed to improve graph opening speed and reduce Dynamo's memory usage. 

The new approach is handled by a new static class: `NodeContextMenuBuilder.cs`. This flexible class can be adapted in the future, should new MenuItems be required.

Any MenuItems created during a node's `NodeViewCustomization process` are stashed per-node, and dynamically injected into the ContextMenu at a consistent location when the menu is built.

Please Note: 

### Declarations

### Reviewers
@saintentropy 
@mjkkirschner 
@QilongTang 

### FYIs
@Amoursol

### DIFF
--- a/src/DynamoCoreWpf/Properties/Resources.en-US.resx
+++ b/src/DynamoCoreWpf/Properties/Resources.en-US.resx
@@ -2627,6 +2627,9 @@ Delete the following packages: {2}?
   <data name="DynamoViewViewMenuConnectorShowTooltip" xml:space="preserve">
     <value>_Show Tooltip</value>
   </data>
+  <data name="NodeInformationalStateDismissedAlerts" xml:space="preserve">
+    <value>Dismissed Alerts</value>
+  </data>
   <data name="ContextMenu" xml:space="preserve">
     <value>Node Options</value>
     <comment>Node context menu</comment>
--- a/src/DynamoCoreWpf/Properties/Resources.en-US.resx
+++ b/src/DynamoCoreWpf/Properties/Resources.en-US.resx
@@ -821,6 +821,14 @@ You will get a chance to save your work.</value>
     <value>Selected package path for download</value>
     <comment>Preferences | Package Manager | Node and Package Paths | New Package Download Directory | Selected package path for download</comment>
   </data>
+  <data name="PreferencesViewDisableBuiltInPackages" xml:space="preserve">
+    <value>Disable Loading Built-In Packages</value>
+    <comment>Preferences | Package Manager | Node and Package Paths | Disable Loading Built-In Packages</comment>
+  </data>
+  <data name="PreferencesViewDisableCustomPackages" xml:space="preserve">
+    <value>Disable Loading Custom Packages</value>
+    <comment>Preferences | Package Manager | Node and Package Paths | Disable Loading Custom Packages</comment>
+  </data>
   <data name="DynamoViewSettingMenuSquareCentimeter" xml:space="preserve">
     <value>Square Centimeter</value>
     <comment>Setting menu | Square centimeter</comment>
@@ -2517,8 +2525,8 @@ Uninstall the following packages: {0}?</value>
   <data name="HideWiresPopupMenuItem" xml:space="preserve">
     <value>Hide Wires</value>
   </data>
-  <data name="UnhideWiresPopupMenuItem" xml:space="preserve">
-    <value>Unhide Wires</value>
+  <data name="ShowWiresPopupMenuItem" xml:space="preserve">
+    <value>Show Wires</value>
   </data>
   <data name="ExitTourWindowContent" xml:space="preserve">
     <value>You can return to this guide later from the Help menu.</value>
@@ -2640,5 +2648,8 @@ Delete the following packages: {2}?
   <data name="PreferencesViewShowPreviewBubbles" xml:space="preserve">
     <value>Show Preview Bubbles</value>
     <comment>Preferences | Visual Settings | Display Settings | Show Preview Bubbles</comment>
+  </data>
+  <data name="NodeAutoCompleteNotAvailableForCollapsedGroups" xml:space="preserve">
+    <value>Node AutoComplete is not available on ports belonging to collapsed groups.</value>
   </data>
 </root>
--- a/src/DynamoCoreWpf/Properties/Resources.en-US.resx
+++ b/src/DynamoCoreWpf/Properties/Resources.en-US.resx
@@ -254,7 +254,7 @@
     <comment>Context menu item</comment>
   </data>
   <data name="ContextMenuDelete" xml:space="preserve">
-    <value>Remove</value>
+    <value>Delete</value>
     <comment>Context menu for selected node - delete selected node</comment>
   </data>
   <data name="ContextMenuEditCustomNode" xml:space="preserve">
--- a/src/DynamoCoreWpf/Properties/Resources.en-US.resx
+++ b/src/DynamoCoreWpf/Properties/Resources.en-US.resx
@@ -2627,9 +2627,6 @@ Delete the following packages: {2}?
   <data name="DynamoViewViewMenuConnectorShowTooltip" xml:space="preserve">
     <value>_Show Tooltip</value>
   </data>
-  <data name="NodeInformationalStateDismissedAlerts" xml:space="preserve">
-    <value>Dismissed Alerts</value>
-  </data>
   <data name="ContextMenu" xml:space="preserve">
     <value>Node Options</value>
     <comment>Node context menu</comment>
--- a/src/DynamoCoreWpf/Utilities/NodeContextMenuBuilder.cs
+++ b/src/DynamoCoreWpf/Utilities/NodeContextMenuBuilder.cs
@@ -1,0 +1,451 @@
+﻿using System.Collections.Generic;
+using System.Windows;
+using System.Windows.Controls;
+using System.Windows.Data;
+using Dynamo.Controls;
+using Dynamo.Graph.Nodes;
+using Dynamo.ViewModels;
+
+namespace Dynamo.Wpf.Utilities
+{
+    /// <summary>
+    /// A static class containing methods for dynamically building a Node's Context Menu.
+    /// </summary>
+    internal static class NodeContextMenuBuilder
+    {
+        /// <summary>
+        /// A reference to the NodeViewModel whose ContextMenu is being built.
+        /// </summary>
+        private static NodeViewModel NodeViewModel { get; set; }
+
+        /// <summary>
+        /// A reference to the ContextMenu to which items are being added. 
+        /// </summary>
+        private static ContextMenu ContextMenu { get; set; }
+
+        // Builds the node's Context Menu, including re-adding any injected MenuItems from
+        // the NodeViewCustomization process.
+        internal static void Build(ContextMenu contextMenu, NodeViewModel nodeViewModel, Dictionary<string, object> nodeViewCustomizationMenuItems)
+        {
+            ContextMenu = contextMenu;
+            NodeViewModel = nodeViewModel;
+            
+            AddContextMenuItem(BuildDeleteMenuItem());
+            AddContextMenuItem(BuildGroupsMenuItem());
+            if (nodeViewModel.ShowsVisibilityToggles) AddContextMenuItem(BuildPreviewMenuItem());
+            AddContextMenuItem(BuildFreezeMenuItem());
+            AddContextMenuItem(BuildShowLabelsMenuItem());
+            AddContextMenuItem(BuildRenameMenuItem());
+            if (nodeViewModel.ArgumentLacing != LacingStrategy.Disabled) AddContextMenuItem(BuildLacingMenuItem());
+            
+            //AddContextMenuItem(BuildDismissedAlertsMenuItem());
+            
+            if (nodeViewCustomizationMenuItems.Count > 0)
+            {
+                AddContextMenuSeparator();
+                AddInjectedNodeViewCustomizationMenuItems(nodeViewCustomizationMenuItems);
+            }
+            
+            if(NodeViewModel.IsInput || NodeViewModel.IsOutput) AddContextMenuSeparator();
+            if(NodeViewModel.IsInput) AddContextMenuItem(BuildIsInputMenuItem());
+            if(NodeViewModel.IsOutput) AddContextMenuItem(BuildIsOutputMenuItem());
+
+            AddContextMenuSeparator();
+            AddContextMenuItem(BuildHelpMenuItem());
+        }
+        
+        /// <summary>
+        /// Adds items to a context menu.
+        /// </summary>
+        /// <param name="menuItem"></param>
+        private static void AddContextMenuItem(MenuItem menuItem) => ContextMenu.Items.Add(menuItem);
+
+        /// <summary>
+        /// Adds a new separator a context menu.
+        /// </summary>
+        private static void AddContextMenuSeparator() => ContextMenu.Items.Add(new Separator());
+
+        /// <summary>
+        /// Creates a new MenuItem in the node's Context Menu.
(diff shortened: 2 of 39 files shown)


Amoursol:
@OliverEGreen is it possible to include the benchmarking numbers you have in the description here? Before and After 😄

OliverEGreen:
> @OliverEGreen is it possible to include the benchmarking numbers you have in the description here? Before and After 😄

image

src/DynamoCoreWpf/Properties/Resources.en-US.resx
```diff
@@ -254,7 +254,7 @@
     <comment>Context menu item</comment>
   </data>
   <data name="ContextMenuDelete" xml:space="preserve">
-    <value>Remove</value>
+    <value>Delete</value>
```
mjkkirschner, line 257: In general - please don't combine different tasks/improvements into one PR.

src/DynamoCoreWpf/Views/Core/NodeView.xaml
```diff
@@ -47,7 +47,10 @@
 
         <!--  CONTEXT MENU  -->
         <Grid.ContextMenu>
-            <ContextMenu Name="MainContextMenu" x:FieldModifier="public">
+            <ContextMenu Name="MainContextMenu"
+                         x:FieldModifier="public"
+                         Closed="MainContextMenu_OnClosed"
+                         Style="{StaticResource NodeContextMenuStyle}">
```
mjkkirschner, line 53: I have two questions here:

1. I think including this style for menuItems in contextMenu.Resources can be optimized - can it also be moved to the sharedResource dictionary and only applied in the nodeContextMenuBuilder and applied there?

2. Why does this style include references to the dismissed badge, is that in the context menu at certain times?

src/DynamoCoreWpf/Views/Core/NodeView.xaml.cs
```diff
@@ -687,26 +689,64 @@ internal void TogglePreviewControlAllowance()
 
         #endregion
 
-        private void OptionsButton_Click(object sender, RoutedEventArgs e)
+        /// <summary>
+        /// A dictionary of MenuItems which are added during certain nodes' NodeViewCustomization process.
+        /// </summary>
+        private Dictionary<string, object> NodeViewCustomizationMenuItems { get; } = new Dictionary<string, object>();
```
mjkkirschner, line 695: why are the values of type `object` and not `MenuItem`?

mjkkirschner: @OliverEGreen a few other issues with this...


Do you need the order of keys / values in this dictionary to stay stable? If that is the case you need to use an ordered dictionary like:
https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.ordereddictionary?view=netframework-4.8

also I was curious why you did not just use a hashset - it looks like MenuItem's base class ItemControl implements `Equals`
https://referencesource.microsoft.com/#PresentationFramework/src/Framework/System/windows/Controls/ItemsControl.cs,3808

though - if you need order, go with the ordered dictionary, hashset does not preserve insertion order either.

src/DynamoCoreWpf/Views/Core/NodeView.xaml.cs
```diff
@@ -687,26 +689,64 @@ internal void TogglePreviewControlAllowance()
...
+        /// </summary>
+        private void DisplayNodeContextMenu(object sender, RoutedEventArgs e)
         {
             Guid nodeGuid = ViewModel.NodeModel.GUID;
             ViewModel.DynamoViewModel.ExecuteCommand(
                 new DynCmd.SelectModelCommand(nodeGuid, Keyboard.Modifiers.AsDynamoType()));
 
-            grid.ContextMenu.DataContext = viewModel;
-            grid.ContextMenu.IsOpen = true;
+            ContextMenu contextMenu = grid.ContextMenu;
```
mjkkirschner, line 729: var

src/DynamoCoreWpf/Views/Core/NodeView.xaml.cs
```diff
@@ -687,26 +689,64 @@ internal void TogglePreviewControlAllowance()
...
             
+            contextMenu.DataContext = viewModel;
+            contextMenu.IsOpen = true;
+
+            e.Handled = true;
+        }
+
+        private void MainContextMenu_OnClosed(object sender, RoutedEventArgs e)
+        {
+            grid.ContextMenu.Items.Clear();
```
mjkkirschner, line 749: if you've already gone this far, can't we get rid of the ContextMenu as well, why instantiate one contextMenu per nodeView when we only need 1 at a time.

src/DynamoCoreWpf/Utilities/NodeContextMenuBuilder.cs
```diff
@@ -0,0 +1,451 @@
...
+        internal static readonly List<string> NodeContextMenuDefaultItemNames = new List<string>
+        {
+            Properties.Resources.ContextMenuDelete,
+            Properties.Resources.ContextMenuGroups,
+            Properties.Resources.NodeContextMenuPreview,
+            Properties.Resources.NodesRunStatus,
+            Properties.Resources.NodeContextMenuShowLabels,
+            Properties.Resources.NodeContextMenuRenameNode,
+            Properties.Resources.ContextMenuLacing,
+            //Wpf.Properties.Resources.NodeInformationalStateDismissedAlerts,
```
mjkkirschner, line 445: ??

src/DynamoCoreWpf/Utilities/NodeContextMenuBuilder.cs
```diff
@@ -0,0 +1,451 @@
...
+        private static MenuItem BuildHelpMenuItem()
+        {
+            return CreateMenuItem
+            (
+                name: "help_cm",
+                header: Properties.Resources.NodeContextMenuHelp,
+                command: new Binding
+                {
+                    Source = NodeViewModel,
+                    Path = new PropertyPath("ShowHelpCommand")
```
mjkkirschner, line 428: can you use nameof?

src/DynamoCoreWpf/Utilities/NodeContextMenuBuilder.cs
```diff
@@ -0,0 +1,451 @@
...
+        {
+            return CreateMenuItem
+            (
+                name: "isOutput_cm",
+                header: Properties.Resources.NodeContextMenuIsOutput,
+                isCheckable: true,
+                isChecked: new Binding
+                {
+                    Source = NodeViewModel,
+                    Path = new PropertyPath("IsSetAsOutput"),
```
mjkkirschner, line 413: nameof?

src/DynamoCoreWpf/Utilities/NodeContextMenuBuilder.cs
```diff
@@ -0,0 +1,451 @@
...
+            //dismissedAlertsMenuItem.ItemTemplate = itemTemplate;
+
+            //AddContextMenuItem(dismissedAlertsMenuItem, insertionPoint++);
+
+            return null;
+        }
+
+        private static void AddInjectedNodeViewCustomizationMenuItems(Dictionary<string, object> nodeViewCustomizationMenuItems)
+        {
+            foreach (KeyValuePair<string, object> keyValuePair in nodeViewCustomizationMenuItems)
```
mjkkirschner, line 381: var

src/DynamoCoreWpf/Utilities/NodeContextMenuBuilder.cs
```diff
@@ -0,0 +1,451 @@
...
+using Dynamo.Controls;
+using Dynamo.Graph.Nodes;
+using Dynamo.ViewModels;
+
+namespace Dynamo.Wpf.Utilities
+{
+    /// <summary>
+    /// A static class containing methods for dynamically building a Node's Context Menu.
+    /// </summary>
+    internal static class NodeContextMenuBuilder
```
mjkkirschner, line 14: 1. please use nameof where you can, when these strings change, everything will break.
2. There is a lot of work in here to construct the views and bindings with code - thats fine - but I think the xaml could also have been reused by extracting it to a resource and instantiating it - perhaps that would not play well with the dynamic menu items - but I think if we can avoid re-writing all of our xaml we should.

mjkkirschner:
This PR should have tests for the new builder class if you decide to go that way.

src/DynamoCoreWpf/Views/Core/NodeView.xaml
OliverEGreen: 1. As far as I can tell this is a dumb WPF thing: 

We can style the `ContextMenu` from a ResourceDictionary fine, but styling the items is a little harder. `ContextMenus` can contain both `MenuItems` and `Separators` (we have both of these in the mockup). 
Telling the `ContextMenu` that its ItemContainerStyle should be a style specifically targeting `MenuItems` won't work and this results in an exception. The `Separators` don't like being given a style that's not for them.

Therefore, I put both the style for Separators and MenuItems within a local resources section, so they get picked up as is appropriate and with no exceptions.

2. Yes it is! It's there at all times.

src/DynamoCoreWpf/Views/Core/NodeView.xaml.cs
OliverEGreen: "why are the values of type object and not MenuItem?"

As above, they can be both Separators and MenuItems.

OliverEGreen: I'd say order stability does matter yeah, will switch out to an OrderedDictionary! 

I considered HashSet, but I wasn't sure about using Equals to compare one MenuItem to another - would that work? Hence, I'm using the 'Header' values to indicate uniqueness.

src/DynamoCoreWpf/Views/Core/NodeView.xaml.cs
OliverEGreen: Changed to var

src/DynamoCoreWpf/Views/Core/NodeView.xaml
OliverEGreen: Am investigating whether we can use an ItemContainerStyleSelector to get past #2.

src/DynamoCoreWpf/Utilities/NodeContextMenuBuilder.cs
OliverEGreen: This is intentionally commented out, to be tied together with the node info states once that work has been merged.

src/DynamoCoreWpf/Views/Core/NodeView.xaml.cs
OliverEGreen: If it's possible to style this centrally then it should be possible. However, we're relying on a lot of locally-defined XAML styles currently.

src/DynamoCoreWpf/Utilities/NodeContextMenuBuilder.cs
OliverEGreen: Have switched all PropertyPath("X") out for PropertyPath(nameof(NodeViewModel.X))

src/DynamoCoreWpf/Utilities/NodeContextMenuBuilder.cs
OliverEGreen: var doesn't let me grab the Key or Value, which is needed here. 
I've swapped for:

`foreach (DictionaryEntry keyValuePair in nodeViewCustomizationMenuItems)
            {
                ContextMenu.Items.Add(keyValuePair.Value);
            }`

src/DynamoCoreWpf/Utilities/NodeContextMenuBuilder.cs
OliverEGreen: 1. I've inserted nameof everywhere. The only hard-coded strings in the class are now equivalents of those that were previously hard-coded in XAML.
2. I'm afraid I don't understand this approach. Please feel free to reach out on Slack/Zoom if you want to go over this.

src/DynamoCoreWpf/Properties/Resources.en-US.resx
OliverEGreen: Ok, will separate this out into its own PR.

OliverEGreen: https://github.com/DynamoDS/Dynamo/pull/12088

src/DynamoCoreWpf/Views/Core/NodeView.xaml
OliverEGreen: I've had no luck with implementing the ItemContainerStyleSelector. It seems like a sensible approach, but I've not seen it being used inside of a ResourceDictionary.

OliverEGreen:
> This PR should have tests for the new builder class if you decide to go that way.

Added 5 tests to cover the new functionality.

src/DynamoCoreWpf/Utilities/NodeContextMenuBuilder.cs
mjkkirschner: `var` doesn't change the type, the type is already inferred from the type of `nodeViewCustomizationMenuItems` - it's a KeyValuePair no matter what.

OliverEGreen: I don't think it will let me access the Value property if it's 'var' though:

image

QilongTang:
The regressions reported for this PR should be unrelated
```
Test Result (3 failures / +3)
DynamoCoreWpfTests.PackageManagerUITests.CannotCreateDuplicatePackageSearchDialogs
DynamoCoreWpfTests.PackageManagerUITests.CanOpenPackageSearchDialogAndWindowIsOwned
DynamoCoreWpfTests.PackageManagerUITests.PackageSearchDialogClosesWithDynamo
```

mjkkirschner:
just do var objectname - there’s no reason have have two type names - the syntax in your example does not look correct

src/DynamoCoreWpf/Utilities/NodeContextMenuBuilder.cs
mjkkirschner: try compiling that - I'm guessing resharper is just getting confused- for example:
https://github.com/DynamoDS/Dynamo/blob/b17e9843a6ce0c3568c4a20fb960ff631cf103e8/test/DynamoCoreTests/SerializationTests.cs#L238


and an example:
https://dotnetfiddle.net/nekXuD

src/DynamoCoreWpf/Utilities/NodeContextMenuBuilder.cs
```diff
@@ -0,0 +1,453 @@
...
+            return null;
+        }
+
+        internal static void AddInjectedNodeViewCustomizationMenuItems(OrderedDictionary nodeViewCustomizationMenuItems)
+        {
+            foreach (DictionaryEntry keyValuePair in nodeViewCustomizationMenuItems)
+            {
+                ContextMenu.Items.Add(keyValuePair.Value);
+            }
+        }
```
OliverEGreen, line 387: @mjkkirschner I think in the latest version of the files there's a DictionaryEntry in this foreach loop. Not entirely sure why the PR isn't showing this on the main page! Must be outdated.

reddyashish (approved):

mjkkirschner:
@OliverEGreen your PR still changes `ContextMenuStyle` to `NodeContextMenuStyle` causing test failures because that style is used in other places. I believe that change was just reverted by @QilongTang - is there a reason you are bringing it back in?

PR #12075, opened 2021-09-28
External Files info in the workspace dependency.

reddyashish:
### Purpose

External Files info in Workspace dependency.

Will be adding tests and snapshots. 

### Declarations

### Reviewers

### FYIs

### DIFF
--- a/src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
+++ b/src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
@@ -728,7 +728,7 @@ namespace Dynamo.Graph.Workspaces
         {
             get
             {
-               var nodeLocalDefinitions = new Dictionary<object, LocalDefinitionInfo>();
+               var nodeLocalDefinitions = new Dictionary<object, DependencyInfo>();
 
                 foreach (var node in Nodes)
                 {
@@ -744,7 +744,7 @@ namespace Dynamo.Graph.Workspaces
 
                             if (!nodeLocalDefinitions.ContainsKey(localDefinitionName)) 
                             {
-                                nodeLocalDefinitions[localDefinitionName] = new LocalDefinitionInfo(localDefinitionName);
+                                nodeLocalDefinitions[localDefinitionName] = new DependencyInfo(localDefinitionName);
                             }
 
                             nodeLocalDefinitions[localDefinitionName].AddDependent(node.GUID);
@@ -757,7 +757,7 @@ namespace Dynamo.Graph.Workspaces
 
                             if (!nodeLocalDefinitions.ContainsKey(localDefinitionName))
                             {
-                                nodeLocalDefinitions[localDefinitionName] = new LocalDefinitionInfo(localDefinitionName, assemblyPath);
+                                nodeLocalDefinitions[localDefinitionName] = new DependencyInfo(localDefinitionName, assemblyPath);
                             }
 
                             nodeLocalDefinitions[localDefinitionName].AddDependent(node.GUID);
@@ -784,7 +784,7 @@ namespace Dynamo.Graph.Workspaces
                     {
                         foreach (var node in dependency.Nodes)
                         {
-                            localDefinitionsDictionary[node] = dependency as LocalDefinitionInfo;
+                            localDefinitionsDictionary[node] = dependency as DependencyInfo;
                         }
                     }
                 }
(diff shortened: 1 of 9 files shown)


src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -793,8 +793,62 @@ internal List<INodeLibraryDependencyInfo> NodeLocalDefinitions
             }
         }
 
+        internal List<INodeLibraryDependencyInfo> ExternalFiles
+        {
+            get
+            {
+                var externalFiles = new Dictionary<object, DependencyInfo>();
```
mjkkirschner, line 800: just from a reusability perspective, I think it would be better to have the getter be very small and extract this to a method.

mjkkirschner:
Hey @reddyashish I'm confused by this - this only seems to handle hard coded file paths inside of codeblocks, filepaths could come from anywhere, they could be computed as part of a python script and thats why gathering them after an execution using an AST walker made sense.

If you are going to limit this feature to hard coded file paths then you should also check string nodes or other types of input nodes that could return strings.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
reddyashish: Yes, I will refactor this and extract it into a method.

reddyashish:
Added the case to handle the nodes that output string values, like in the example below
Screen Shot 2021-09-29 at 8 44 41 AM

The 2 cases, still not handled, are:
1) like dynamically creating the file path from multiple strings(like the example @mjkkirschner pointed out)
2) Assigning the path string to a variable in the CBN line.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -793,8 +793,86 @@ internal List<INodeLibraryDependencyInfo> NodeLocalDefinitions
...
+                                    externalFiles[externalFileName] = new DependencyInfo(externalFileName, path);
+                                }
+
+                                externalFiles[externalFileName].AddDependent(node.GUID);
+                                externalFiles[externalFileName].ReferenceType = ReferenceType.External;
+                            }
+                        }
+                    }
+
+                    if(node.CachedValue != null)
```
aparajit-pratap, line 834: Does this cover nodes like string node, file path node etc.?

reddyashish: Yes, it covers these nodes.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -333,14 +347,31 @@ public void Dispose()
...
 
         private void Refresh_MouseLeftButtonDown(object sender, System.Windows.Input.MouseButtonEventArgs e)
         {
             DependencyRegen(currentWorkspace);
         }
+
+        /// <summary>
+        /// Computes the file size from the path.
+        /// </summary>
+        private string GetFileSize(string path)
```
QilongTang, line 364: This can probably be moved to `DynamoUtilities` project -> File Path helper

reddyashish: Yes, will move this in the next commit.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -753,15 +753,21 @@ internal List<INodeLibraryDependencyInfo> NodeLocalDefinitions
                         else if (node is DSFunctionBase functionNode)
                         {
                             string assemblyPath = functionNode.Controller.Definition.Assembly;
-                            localDefinitionName = Path.GetFileName(assemblyPath);
+                            var directoryName = Path.GetDirectoryName(assemblyPath);
 
-                            if (!nodeLocalDefinitions.ContainsKey(localDefinitionName))
+                            // For the local definition reference, the assembly directory exists on disc.
+                            if (!directoryName.Equals("") && Directory.Exists(directoryName))
```
QilongTang, line 759: Let's use String.Empty when it applies

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -455,19 +472,19 @@ public ImageSource Icon
     }
 
     /// <summary>
-    /// Represents information about a local dependency as a row in the dependency table
+    /// Represents information about a dependency as a row in the dependency table
     /// </summary>
-    public class LocalDefinitionRow
+    public class DependencyRow
```
QilongTang, line 477: I think this is API breaking but should be fine since we did ask folks to not depending on our UI and extension APIs? @reddyashish @mjkkirschner

reddyashish: This class was added after 2.12 release. It wouldn't affect users using the daily builds, right?

QilongTang: @reddyashish I see, then it should be fine. Thanks!

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -793,8 +799,83 @@ internal List<INodeLibraryDependencyInfo> NodeLocalDefinitions
...
+            }
+        }
+
+        private List<INodeLibraryDependencyInfo> GetExternalFiles()
+        {
+            var externalFiles = new Dictionary<object, DependencyInfo>();
+
+            // Computes the external file references if the Workspace Model is a HomeWorkspaceModel.
+            // The workspace should be executed for the external references to be computed because the node output port values are needed.
+            if (this is HomeWorkspaceModel homeWorkspaceModel)
```
QilongTang, line 831: Can you work with @Jingyi-Wen for a way to show indication about this limitation? Unless it is already implemented

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -793,8 +799,83 @@ internal List<INodeLibraryDependencyInfo> NodeLocalDefinitions
...
+                    foreach (var port in node.OutPorts)
+                    {
+                        var id = node.GetAstIdentifierForOutputIndex(port.Index).Name;
+                        var mirror = homeWorkspaceModel.EngineController.GetMirror(id);
+                        var data = mirror?.GetData().Data;
+
+                        if (data is string dataString)
+                        {
+                            // If the value exists on disk
+                            if (File.Exists(data.ToString()))
```
QilongTang, line 847: Can you check if there is a quicker way to check for path string before making system calls? I suspect this may introduce performance issue with large graph. If there is no good way, should we check for string including `\` as a first path?

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -793,8 +799,83 @@ internal List<INodeLibraryDependencyInfo> NodeLocalDefinitions
...
+            // Computes the external file references if the Workspace Model is a HomeWorkspaceModel.
+            // The workspace should be executed for the external references to be computed because the node output port values are needed.
+            if (this is HomeWorkspaceModel homeWorkspaceModel)
+            {
+                foreach (var node in nodes)
+                {
+                    externalFilesDictionary.TryGetValue(node.GUID, out var serializedDependencyInfo);
+
+                    // Check for the file path string value at each of the output ports of all nodes in the workspace. 
+                    foreach (var port in node.OutPorts)
```
QilongTang, line 838: Thought you mention this implementation come with some limitation on CBN, is that still the case?

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
reddyashish: Yes, will look into it. Also @Jingyi-Wen Do we have any icons for the external file references?

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
mjkkirschner: have you looked at FileInfo constructor instead? 
https://stackoverflow.com/questions/3067479/determine-via-c-sharp-whether-a-string-is-a-valid-file-path

reddyashish:
@QilongTang We can do that in a followup PR once it we have a final confirmation.

PR #12083, opened 2021-09-28
a couple of fixes for the Package Search window redesign

pinzart90:
What changed in this PR:

1. Swapped IsInstalled flag to CanInstall (reversing its meaning)
2. Handler managements fixes (properly add and remove them from the PMSearchViewModel)
3. Removed a SetAsLoaded call when a package is finished trying to load.
4. CanInstall will be updated in the UI for all corresponding DownloadHandles (when the download handle's DownloadState property is changed)
5. Fixed test

A package will have the Install button enabled/disabled with the following logic:
```
 case PackageDownloadHandle.State.Uninitialized:
 case PackageDownloadHandle.State.Error:
                    return true;// Allowed if Download/Install not yet begun or if in Error state.
 case PackageDownloadHandle.State.Downloaded:
 case PackageDownloadHandle.State.Downloading:
 case PackageDownloadHandle.State.Installing:
     return false; // Not allowed if Download or Install is in progress
 case PackageDownloadHandle.State.Installed:
    // If Install finished...then false if a non built-in package with the same name already exists. True otherwise
    return false == PackageLoader.LocalPackages.Any(x => (x.Name == name) && !x.BuiltInPackage);
```


src/DynamoCoreWpf/Views/PackageManager/PackageManagerSearchView.xaml
```diff
@@ -630,7 +630,7 @@
                                                             <DataTrigger Binding="{Binding Model.IsDeprecated}" Value="False">
                                                                 <Setter Property="TextBlock.Foreground" Value="#84D7CE" />
                                                             </DataTrigger>
-                                                            <DataTrigger Binding="{Binding IsInstalled}" Value="True">
+                                                            <DataTrigger Binding="{Binding CanInstall}" Value="False">
```
pinzart90, line 633: I reversed the meaning of the boolean

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
```diff
@@ -913,5 +916,12 @@ public void ExecuteSelected()
             SearchResults[SelectedIndex].Model.Execute();
         }
 
+        private void OnPackageAddedOrRemoved(Package _)
+        {
+            foreach (var sr in SearchResults)
+            {
+                sr.CanInstall = CanInstallPackage(sr.Model.Name);
```
pinzart90, line 923: O every added/removed package we will update the CanInstall flag for each searchElement

src/DynamoCoreWpf/DynamoCoreWpf.csproj
```diff
@@ -1386,28 +1386,31 @@
...
     <Resource Include="UI\Images\Profile.png" />
+    <Resource Include="UI\Images\Progress circle.png" />
+    <Resource Include="UI\Images\Requirements - disabled.png" />
     <Resource Include="UI\Images\Requirements.png" />
-    <Resource Include="UI\Images\Sort.png" />
+    <Resource Include="UI\Images\View details - disabled.png" />
     <Resource Include="UI\Images\View details.png" />
+    <Resource Include="UI\Images\watch_select_48_48.png" />
+    <Resource Include="UI\Images\Sort.png" />
+    <Resource Include="UI\Images\package_error.png" />
```
pinzart90, line 1413: Added a couple of missed icons

QilongTang:
FYI: @OliverEGreen

test/DynamoCoreWpfTests/PackageManager/PackageManagerSearchElementViewModelTests.cs
```diff
@@ -18,20 +22,45 @@ class PackageManagerSearchElementViewModelTests : SystemTestBase
...
 
-            PackageManagerSearchElementViewModel packageManagerSearchElementViewModel = new PackageManagerSearchElementViewModel
-            (
-                packageManagerSearchElements.First(),
-                false,
-                packageManagerSearchViewModel.UserAlreadyHasPackageInstalled
-            );
 
-            packageManagerSearchElementViewModel.RequestDownload += packageManagerSearchViewModel.PackageOnExecuted;
+            newSE.RequestDownload += packageManagerSearchViewModel.PackageOnExecuted;
```
mjkkirschner, line 63: @pinzart90 can we unsub this at the end of the test?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
```diff
@@ -793,6 +795,10 @@ internal virtual void InstallPackage(PackageDownloadHandle packageDownloadHandle
                 {
                     packageDownloadHandle.Error(e.Message);
                 }
+                finally
+                {
+                    PackageInstallFinished?.Invoke(packageDownloadHandle);
```
mjkkirschner, line 800: is this actually triggered at the right time? BeginInvoke is async isn't it?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
```diff
@@ -808,8 +807,6 @@ internal void SetPackageState(PackageDownloadHandle packageDownloadHandle, strin
             {
                 PackageManagerExtension.PackageLoader.LoadPackages(new List<Package> { dynPkg });
                 packageDownloadHandle.DownloadState = PackageDownloadHandle.State.Installed;
-
-                dynPkg.LoadState.SetAsLoaded();
```
pinzart90, line 812: This was wrong.
LoadState is set inside the PackageManagerExtension.PackageLoader.LoadPackages method (which can be different from Loaded)

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchElementViewModel.cs
```diff
@@ -21,24 +21,25 @@ public class PackageManagerSearchElementViewModel : BrowserItemViewModel, IEquat
...
-        public PackageManagerSearchElementViewModel(PackageManagerSearchElement element, bool canLogin) : base(element)
+        /// <summary>
+        /// Alternative constructor which takes a Function object in order to
+        /// assist communication between the PackageManagerSearchViewModel
+        /// and the PackageManagerSearchElementViewModel.
+        /// </summary>
+        /// <param name="element">A PackageManagerSearchElement</param>
+        /// <param name="canLogin">A Boolean used for access control to certain internal packages.</param>
+        /// <param name="canInstall">Whether a package can be installed.</param>
+        internal PackageManagerSearchElementViewModel(PackageManagerSearchElement element, bool canLogin, bool canInstall) : base(element)
```
pinzart90, line 34: Made this internal. I hope we can do that in between releases...

pinzart90: @mjkkirschner ?

QilongTang: hmm I think constructors are usually public, any reason to force `internal`?

mjkkirschner: @pinzart90 - no sorry, if this was actually part of a public class, and it was public, and released, then making it internal is like removing the method.

mjkkirschner: I guess I'm confused by what you mean by "between releases"

pinzart90: @QilongTang correction
The public CanInstall property has an internal setter.
Do you guys still think I should make the Constructor public ?

mjkkirschner:
@pinzart90 - in your final case, what if the package name and version match - is there any reason to allow the install button to be enabled to install the same package and version that is already installed?

aparajit-pratap:
Can you explain this case:
```
case PackageDownloadHandle.State.Installed:
    // If Install finished...then false if a non built-in package with the same name already exists. True otherwise
    return false == PackageLoader.LocalPackages.Any(x => (x.Name == name) && !x.BuiltInPackage);
```
I don't understand why return false for only non-built-in packages?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
aparajit-pratap: So it's possible for package download handle state to say "installed" but the package `loadstate` to be unloaded or error?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchElementViewModel.cs
```diff
@@ -21,24 +21,25 @@ public class PackageManagerSearchElementViewModel : BrowserItemViewModel, IEquat
...
             this.Model = element;
-
-            this.ToggleIsExpandedCommand = new DelegateCommand(() => this.Model.IsExpanded = !this.Model.IsExpanded );
+            
+            this.ToggleIsExpandedCommand = new DelegateCommand(() => this.Model.IsExpanded = !this.Model.IsExpanded);
 
             this.DownloadLatestCommand = new DelegateCommand(
                 () => OnRequestDownload(Model.Header.versions.Last(), false),
-                () => !Model.IsDeprecated && !IsInstalled);
+                () => !Model.IsDeprecated && CanInstall);
```
aparajit-pratap, line 42: Should this be `canInstall` or `CanInstall` after assigning it to `canInstall`?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
pinzart90: Yes, it can. 
The PackageDownloadHandle's downloadState does not handle "load into library" failures that well
We can definitely make it keep count of  "loadState" but that may be something that needs some UX thought ...

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
```diff
@@ -279,27 +279,36 @@ public bool HasNoResults
...
-
-            List<string> loadedPackageNames = packageManagerExtension.PackageLoader.LocalPackages
-                .Where(x => x.LoadState.State == PackageLoadState.StateTypes.Loaded)
-                .Select(x => x.Name)
-                .ToList();
-
-            List<string> recentlyDownloadedPackageNames = Downloads
-                .Select(x => x.Name)
-                .ToList();
+            // Return true if there are no matching non built-in packages
```
aparajit-pratap, line 288: Why check for non-built-in only? Don't we need to check for no matching package among all packages?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
```diff
@@ -279,27 +279,36 @@ public bool HasNoResults
...
+            return false == PackageManagerClientViewModel.PackageManagerExtension.PackageLoader.LocalPackages
```
aparajit-pratap, line 289: how about `return !PackageManagerClientViewModel.PackageManagerExtension.PackageLoader.LocalPackages...`

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
```diff
@@ -279,27 +279,36 @@ public bool HasNoResults
...
+                .Any(x => (x.Name == name) && !x.BuiltInPackage);
+        }
 
-            return loadedPackageNames.Contains(packageName) || recentlyDownloadedPackageNames.Contains(packageName);
+        /// <summary>
+        /// Checks if the package corresponding to the PackageDownloadHandle can be installed.
+        /// </summary>
+        /// <param name="packageName"></param>
```
aparajit-pratap, line 296: change to `dh`

test/DynamoCoreWpfTests/PackageManager/PackageManagerSearchElementViewModelTests.cs
```diff
@@ -1,50 +1,97 @@
...
+            Assert.AreEqual(false, newSE2.CanInstall);
+
+            dHandle1.DownloadState = PackageDownloadHandle.State.Installing;
+            dHandle2.DownloadState = PackageDownloadHandle.State.Downloaded;
+            Assert.AreEqual(false, newSE1.CanInstall);
+            Assert.AreEqual(false, newSE2.CanInstall);
+
+            dHandle1.DownloadState = PackageDownloadHandle.State.Installed;
+            dHandle2.DownloadState = PackageDownloadHandle.State.Installed;
+            Assert.AreEqual(true, newSE1.CanInstall);
```
aparajit-pratap, line 91: How come this is true, the package is installed, right?

test/DynamoCoreWpfTests/PackageManager/PackageManagerSearchElementViewModelTests.cs
```diff
@@ -1,50 +1,97 @@
...
+            var newSE2 = new PackageManagerSearchElementViewModel(new PackageManagerSearchElement(new PackageHeader()
+            {
+                name = name2
+            }), false);
+
+            packageManagerSearchViewModel.AddToSearchResults(newSE1);
+            packageManagerSearchViewModel.AddToSearchResults(newSE2);
+
+            // Default CanInstall should be true
+            Assert.AreEqual(true, newSE1.CanInstall);
```
aparajit-pratap, line 49: How is this true? It is already added to packageloader, right?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
```diff
@@ -701,17 +706,27 @@ private void DownloadsOnCollectionChanged(object sender, NotifyCollectionChanged
...
+                    handle.PropertyChanged += (o, eventArgs) => {
+                        // Hooking into propertyChanged works only if each download handle is added to
+                        // the Downloads collection before Download/Install begins.
+                        if (eventArgs.PropertyName == nameof(PackageDownloadHandle.DownloadState)) {
+                            PackageManagerSearchElementViewModel sr = SearchResults.FirstOrDefault(x => x.Model.Name == handle.Name);
+                            if (sr == null) return;
+
+                            sr.CanInstall = CanInstallPackage(handle);
+                        }
+                    };
```
aparajit-pratap, line 720: how do we unsubscribe these anonymous handlers for each handle later?

test/DynamoCoreWpfTests/PackageManager/PackageManagerSearchElementViewModelTests.cs
```diff
@@ -1,50 +1,97 @@
...
+                name = name1
+            }), false);
+
+            var newSE2 = new PackageManagerSearchElementViewModel(new PackageManagerSearchElement(new PackageHeader()
+            {
+                name = name2
+            }), false);
+
+            packageManagerSearchViewModel.AddToSearchResults(newSE1);
+            packageManagerSearchViewModel.AddToSearchResults(newSE2);
```
aparajit-pratap, line 46: How is this true? It is already added to packageloader in line 30, right?

test/DynamoCoreWpfTests/PackageManager/PackageManagerSearchElementViewModelTests.cs
```diff
@@ -1,50 +1,97 @@
...
+
+            // Default CanInstall should be true
+            Assert.AreEqual(true, newSE1.CanInstall);
+            Assert.AreEqual(true, newSE2.CanInstall);
```
aparajit-pratap, line 50: How is this true? It is already added to packageloader in line 30, right?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchElementViewModel.cs
pinzart90: CanInstall is a wrapper over canInstall.
Not sure if there is a benefit to using the private property here ...Can you explain a little bit more what you mean ?

src/DynamoCoreWpf/UI/Converters.cs
```diff
@@ -316,8 +316,8 @@ public class InstalledButtonTextConverter : IValueConverter
             if (!(value is bool booleanValue)) return null;
             
             return booleanValue
-                ? Wpf.Properties.Resources.PackageDownloadStateInstalled
-                : Wpf.Properties.Resources.PackageManagerInstall;
+                ? Wpf.Properties.Resources.PackageManagerInstall
+                : Wpf.Properties.Resources.PackageDownloadStateInstalled;
```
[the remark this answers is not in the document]
pinzart90: Correct

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchElementViewModel.cs
```diff
@@ -51,45 +52,28 @@ public PackageManagerSearchElementViewModel(PackageManagerSearchElement element,
...
-        /// <param name="canLogin">A Boolean used for access control to certain internal packages.</param>
-        /// <param name="isPackageInstalledFunction">A function that we pass to this constructor which takes a string and returns a Boolean.</param>
-        public PackageManagerSearchElementViewModel(PackageManagerSearchElement element, bool canLogin, Func<string,bool> isPackageInstalledFunction) : this(element, canLogin)
-        {
-            this.isPackageInstalled = isPackageInstalledFunction;
-            this.IsInstalled = isPackageInstalled(this.Model.Name);
-        }
-
-        private bool isInstalled;
+        public PackageManagerSearchElementViewModel(PackageManagerSearchElement element, bool canLogin) : this(element, canLogin, true)
```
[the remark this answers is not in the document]
pinzart90: I can add some. There weren't any  ...

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
pinzart90: My thinking was that we should allow a user to install a package that already exists as a built-in. How else would the user be able to override an existing built-in package ?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
pinzart90: why not...I wanted it to be very explicit just because it can be a bit confusing...

test/DynamoCoreWpfTests/PackageManager/PackageManagerSearchElementViewModelTests.cs
pinzart90: There is no special meaning to it. It just checks that the default value set by the PackageManagerSearchElementViewModel constructor is true.

test/DynamoCoreWpfTests/PackageManager/PackageManagerSearchElementViewModelTests.cs
pinzart90: In this test scenario (and I can add some more descriptions) there are 2 packages being downloaded:
            var name1 = "non-duplicate"; (downloadHandle newSE1)
            var name2 = "duplicate"; (downloadHandle newSE2)
newSE1 - has no conflicts with existing localPackages
newSE2 - has a conflict with localPackcages  
          
Moreover newSE1 never gets added to localPackages (that's just what I chose to do in the test). Because of this, after newSE1 gets an  Installed downloadState, it can still be installed. I guess this could simulate an exception being thrown before the newSE1 package can get loaded in localPackages

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
pinzart90: uhm...good catch
I missed that

pinzart90:
> @pinzart90 - in your final case, what if the package name and version match - is there any reason to allow the install button to be enabled to install the same package and version that is already installed?

Not sure I understand your scenario @mjkkirschner 
If a package is already installed (non-built in) then a package with the same name should not have the install button enabled

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchElementViewModel.cs
aparajit-pratap: You're using the old value of `CanInstall`, not the value of `canInstall` that is passed in, why?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
aparajit-pratap: Right, but why not allow the user to also override a custom package?

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchElementViewModel.cs
aparajit-pratap: Also, there is the public property, `CanInstall`, the private backing field, `canInstall`, and a function parameter, also called `canInstall`. Perhaps you might want to change the parameter name to something else to avoid confusion.

test/DynamoCoreWpfTests/PackageManager/PackageManagerSearchElementViewModelTests.cs
aparajit-pratap: So it looks like this can never be a real-life scenario. If so, then why have a test for it?

pinzart90: I added the newSE1 to local packages in the end (simulating a successful load by the packageLoader).
Now both packages cannot be installed

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
pinzart90: If by custom package you mean non-built in package, then the answer is in the original PR https://github.com/DynamoDS/Dynamo/pull/12033
Looks like the original requirement was to disable the Install button if a package is already installed

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchElementViewModel.cs
pinzart90: got it. Changed

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
aparajit-pratap: Yes by custom package I mean non-built-in package. We are using custom package as the term to describe these types of packages, which is what you'll see in the toggles.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchElementViewModel.cs
pinzart90: I made it public

aparajit-pratap (approved):

PR #12087, opened 2021-09-28
Fix unintended serialization of the SelectedIndex for Input nodes

saintentropy:
### Purpose

This sets the SelectedIndex to nullable so that it does not serialize with every input type.  This key / value should only be serialized when the input type is a DropDown.

### Declarations

### Reviewers

@mjkkirschner  @nate-peters 

### FYIs

@BogdanZavu


src/DynamoCore/Graph/Nodes/NodeInputData.cs
```diff
@@ -92,7 +92,7 @@ public class NodeInputData
         /// The index of the selected item.
         /// </summary>
         [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
-        public int SelectedIndex { get; set; }
+        public int? SelectedIndex { get; set; }
```
mjkkirschner, line 95: it's likely this is an API break.
I don't know if we care - but I think it fits in


❌ DISALLOWED: Changing the type of a member

The return value of a method or the type of a property or field cannot be modified. For example, the signature of a method that returns an Object cannot be changed to return a String, or vice versa.

this may cause binary compatibility issues with any code that referenced this member, can you check?

saintentropy: I think you are right.  That was my reading of it as well.  I do think we should probably make the change though.  I will double check.

mjkkirschner: Hey - finally time to do this?

QilongTang: Marked as 3.0

saintentropy: Yep let's do this.  Just updated the branch @mjkkirschner @QilongTang

mjkkirschner (approved):

PR #12112, opened 2021-10-05
add hidden preference for tag size limit

pinzart90:
Add hidden preference for the SearchDictionary tag size limit

### DIFF
--- a/src/DynamoCore/Configuration/PreferenceSettings.cs
+++ b/src/DynamoCore/Configuration/PreferenceSettings.cs
@@ -445,6 +445,22 @@ namespace Dynamo.Configuration
         /// Show Run Preview flag.
         /// </summary>
         public bool ShowRunPreview { get; set; }
+
+        /// <summary>
+        /// Limits the size of the tags used by the SearchDictionary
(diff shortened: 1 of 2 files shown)


src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -445,6 +445,22 @@ public bool DisableCustomPackageLocations
         /// Show Run Preview flag.
         /// </summary>
         public bool ShowRunPreview { get; set; }
+
+        /// <summary>
+        /// Limits the size of the tags used by the SearchDictionary
+        /// This static property is not serialized and is assigned NodeSearchTagSizeLimit's value 
+        /// if found at deserialize time.
+        /// </summary>
+        internal static int NodeSearchTagSizeLimitValue = 300;
```
pinzart90, line 454: this static property is just for easy access from the SearchDictionary.
If I remove it ...I would have to pass along the property `NodeSearchTagSizeLimit` to all instances of SearchDictionary (which means more changes...)

mjkkirschner (approved):

PR #12113, opened 2021-10-05
Connector/ ConnectorPins Collapsable with Groups Fixes

M-JULIANI:
### Description

This PR adds mostly focuses on getting connectors and connector pins to work correctly with groups. They now collapse/expand with groups. An additional flag is added to the ConnectorModel(`IsHidden`) to work in tandem with `IsCollapsed` (which now lives exclusively in the ConnectorViewModel). `IsCollapsed` comes from groups and thus doesn't require serialization whereas `IsHidden` is required for serialization as it tells the graph how to render any and all connectors when a graph is re-opened.

This PR also fixes the following:

- Connectors that are hidden/collapsed are no longer right-clickable.
- Correct styling for ConnectorPinView context menu for unpinning pin.
ConnectorPins-GroupCollapsable-05 10 2021

### Declarations

### Reviewers
@QilongTang 
@saintentropy 

### FYIs
@SHKnudsen 
@Amoursol

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Core/ConnectorPinViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/ConnectorPinViewModel.cs
@@ -135,7 +135,7 @@ namespace Dynamo.ViewModels
             get { return model.IsSelected; }
         }
 
-        private bool isCollapsed;
+        private bool isCollapsed = false;
         [JsonIgnore]
         public override bool IsCollapsed
         {
@@ -151,6 +151,23 @@ namespace Dynamo.ViewModels
                 RaisePropertyChanged(nameof(IsCollapsed));
             }
         }
+
+        private bool isHidden;
+        public bool IsHidden
+        {
+            get => isHidden;
+            set
+            {
+                if (isHidden == value)
+                {
+                    return;
+                }
+
+                isHidden = value;
+                RaisePropertyChanged(nameof(IsHidden));
+            }
+        }
+
 
         private bool isTemporarilyVisible;
         /// <summary>
--- a/src/DynamoCoreWpf/ViewModels/Core/ConnectorPinViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/ConnectorPinViewModel.cs
@@ -126,6 +126,11 @@ namespace Dynamo.ViewModels
             }
         }
         public bool isHoveredOver = false;
+        /// <summary>
+        /// This flag let's the ConnectorViewModel when it can and cannot run a ConnectorContextMenu. It CANNOT
+        /// do so when IsHoveredOver for any pin is set to true, as in that case we want only that ConnectorPins 
+        /// ContextMenu to be enabled on right click.
+        /// </summary>
         public bool IsHoveredOver
         {
             get => isHoveredOver;
--- a/src/DynamoCoreWpf/ViewModels/Core/ConnectorPinViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/ConnectorPinViewModel.cs
@@ -125,6 +125,21 @@ namespace Dynamo.ViewModels
                 RaisePropertyChanged(nameof(Top));
             }
         }
+        public bool isHoveredOver = false;
+        public bool IsHoveredOver
+        {
+            get => isHoveredOver;
+            set
+            {
+                if (isHoveredOver == value)
+                {
+                    return;
+                }
+
+                isHoveredOver = value;
+                RaisePropertyChanged(nameof(IsHoveredOver));
+            }
+        }
 
         /// <summary>
         /// Provides the ViewModel (this) with the selected state of the ConnectorPinModel.
--- a/src/DynamoCoreWpf/ViewModels/Core/ConnectorViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/ConnectorViewModel.cs
@@ -175,6 +175,7 @@ namespace Dynamo.ViewModels
 
                 isCollapsed = value;
                 RaisePropertyChanged(nameof(IsCollapsed));
+                SetCollapseOfPins(IsCollapsed);
             }
         }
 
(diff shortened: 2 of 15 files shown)


src/DynamoCoreWpf/ViewModels/Core/ConnectorViewModel.cs
```diff
@@ -1032,10 +1040,14 @@ private void PinViewModelPropertyChanged(object sender, PropertyChangedEventArgs
...
             switch (e.PropertyName)
             {
-                case nameof(ConnectorPinModel.IsSelected):
+                case nameof(ConnectorPinViewModel.IsSelected):
                     var vm = sender as ConnectorPinViewModel;
                     AnyPinSelected = vm.IsSelected;
                     break;
+                case nameof(ConnectorPinViewModel.IsHoveredOver):
+                    vm = sender as ConnectorPinViewModel;
+                    AnyPinHoveredOver = vm.IsHoveredOver;
```
SHKnudsen, line 1050: Do you actually need to set this property like this?
Seems like you could just have

```csharp
internal bool IsAnyPinHovered => ConnectorPinViewCollection.Any(p => p.IsHoveredOver);
```

Or even just do the linq statement when you needed it, i cant really figure out how often this property is used

SHKnudsen: Seems like the same could be done on the `AnyPinSelected`

M-JULIANI: You would still need to wrap it in a handler right? Need to handle it every time those properties are updated.

src/DynamoCoreWpf/ViewModels/Core/ConnectorViewModel.cs
```diff
@@ -1059,6 +1071,11 @@ private void HandleConnectorPinViewModelRemove(object sender, EventArgs e)
             workspaceViewModel.Model.RecordAndDeleteModels(
                 new List<ModelBase>() { viewModelSender.Model });
             ConnectorModel.ConnectorPinModels.Remove(viewModelSender.Model);
+
+            if(ConnectorContextMenuViewModel!= null)
+            {
+                ConnectorContextMenuViewModel.DisposeViewModel();
```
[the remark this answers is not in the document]
M-JULIANI: Only reason is not to confuse it with the built-in `dispose` on modelbase

src/DynamoCoreWpf/ViewModels/Core/ConnectorViewModel.cs
M-JULIANI: After looking at it seems like overkill to iterate thru all the pins in this scenario, I guess you only need the pin whose property changed.. no?

SHKnudsen: Why do you need to handle it every time the property is updated? It looks to me like you only ever use this property in the 'CanInstantiateContextMenu' so don't see the need to constantly keeping track of it. Instead you can easily make that check only when you need it.

src/DynamoCoreWpf/ViewModels/Core/ConnectorViewModel.cs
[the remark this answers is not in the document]
SHKnudsen: But isn't this exactly what dispose is for? Or is this different, it seems a bit confusing. 

You can just override the ViewModelBase Dispose.

src/DynamoCoreWpf/ViewModels/Core/ConnectorViewModel.cs
M-JULIANI: ^That goes for `AnyPinSelected`. Similar logic for `AnyPinHoveredOver`. However this flag tells us whether we can instantiate a ContextMenu for the wire. If `AnyPinHoveredOver` is true, then we cannot fire that command.

src/DynamoCoreWpf/ViewModels/Core/ConnectorViewModel.cs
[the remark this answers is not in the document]
M-JULIANI: Yeah that's fair. Was inheriting from `NotificationObject` when I should have been inheriting from `ViewModelBase`. Done.

src/DynamoCoreWpf/ViewModels/Core/ConnectorViewModel.cs
SHKnudsen: Right, so what i am asking is, do you really need to constantly update this state, or could you not just on-demand check if any pins is hovered or selected, as in the Linq statement above?

I havent looked into the AnyPinSelected state so maybe thats different, but AnyPinHovered is used once in `CanInstantiateContextMenu` your current implementation does not require that state to be dynamically updated, you might as well call `ConnectorPinViewCollection.Any(p => p.IsHoveredOver);` there instead, that way you do not need to worry about these events.

src/DynamoCoreWpf/ViewModels/Core/ConnectorPinViewModel.cs
```diff
@@ -125,12 +125,13 @@ public double Top
                 RaisePropertyChanged(nameof(Top));
             }
         }
-        public bool isHoveredOver = false;
+        private bool isHoveredOver = false;
```
SHKnudsen, line 128: is this property actually needed anymore?

src/DynamoCoreWpf/ViewModels/Core/ConnectorViewModel.cs
```diff
@@ -505,7 +496,23 @@ public bool AnyPinSelected
                 RaisePropertyChanged(nameof(AnyPinSelected));
             }
         }
-
+        /// <summary>
+        /// This flag let's the ConnectorViewModel when it can and cannot run a ConnectorContextMenu. It CANNOT
+        /// do so when IsHoveredOver for any pin is set to true, as in that case we want only that ConnectorPins 
+        /// ContextMenu to be enabled on right click.
+        /// </summary>
+        public bool AnyPinHoveredOver
```
SHKnudsen, line 504: Should be able to remove this one now, right?

QilongTang:
Would you also address the merge conflicts before merge, it's related to the port splits PR

PR #12116, opened 2021-10-06
cef scanning issue

pinzart90:
https://jira.autodesk.com/browse/DYN-4051

Removed LibraryViewExtension projects from the build. Still included in the solution...
(cef will not be downloaded with nuget restore anymore)

Later on we can add a separate build target for the libraryViewExtension projects...so that we can optionally build it in cicd...


pinzart90:
Alternatively I can completely remove the LibraryViewExtension projects from the solution

mjkkirschner:
@QilongTang for some context - we discussed this morning removing the cef based library UI from the build .sln - this way we keep the code around, it does not build, and does not trigger scan alerts - if anyone actually needs it, then we'll turn it back on, deal with the consequences etc - but for now seems it's not in use by an integrator.

mjkkirschner (approved):

aparajit-pratap:
So did you just do an `unload project` for LibraryViewExtension?

PR #12148, opened 2021-10-14
Remove IronPython related code

pinzart90:
Changes in this PR
1. DSIronPython, IronPythonExtension  have been disabled from the build process (still in the solution though)
2. Changed how Dynamo manages optional existence of IronPython (src code changes)
3. Simplified DynamoPythonTests to only handle CPython3
4. Moved all IronPython related tests to a new test Project: IronPythonTests (disabled from build)
5. Tweaked some of the Samples so that they do not reference IronPython

TODO: will skip IronPython packages in the harmony scans.

### DIFF
--- a/src/DisabledProjects.md
+++ b/src/DisabledProjects.md
@@ -8,7 +8,7 @@
 3. IronPythonTests
 
 ### Reason
-IronPython is no longer maintained and has many security issues
+Python2 and by extension IronPython2 has reached end of life and has many security issues
 
 ### Resolution
 IronPython has been deactivated and replaced with Python.NET (pythonnet)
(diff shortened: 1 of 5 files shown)


aparajit-pratap:
> @pinzart90 - would you mind explaining some of your approach in a sentence or two? For example, I'm a bit confused about the changes to find the available engines - which previously used reflection and now appear to use a resolve handler?

Yes, it would be good to explain this approach/flow. It looks like this sequence is triggered at startup and each time a new assembly is loaded?
```
ScanPythonEngine => OnPythonEngineScan (AddPythonEnginesOptions) => GetPythonEngineOptions => GetEnabledEngines => (check for IsIronPythonEnabled and IsCPythonEnabled) which have been set in ScanPythonEngine 
```

PR #12225, opened 2021-11-02
Fix the connectors and ports, when the group is collapsed.

reddyashish:
### Purpose

This PR is to fix the issue related to connectors and ports when the group is collapsed. 

Task: https://jira.autodesk.com/browse/DYN-3983

group collapsing

### Declarations

### Release Notes

(FILL ME IN) Brief description of the fix / enhacement. **Mandatory section** 

### Reviewers

@QilongTang @SHKnudsen

### DIFF
--- a/src/DynamoCore/Graph/Nodes/PortModel.cs
+++ b/src/DynamoCore/Graph/Nodes/PortModel.cs
@@ -153,6 +153,7 @@ namespace Dynamo.Graph.Nodes
             }
         }
 
+        private Point2D center = new Point2D();
         /// <summary>
         /// Center is used by connected connectors to update their shape
         /// The "center" of a port is derived from the type of port and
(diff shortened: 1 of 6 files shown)


src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -649,6 +649,39 @@ internal IEnumerable<PortModel> GetGroupOutPorts(IEnumerable<NodeModel> ownerNod
...
+            int index = portModel.LineIndex == -1 ? portModel.Index : portModel.LineIndex;
+
+            //If the port was not found, then it should have just been deleted. Return from function
+            if (index == -1)
+                return verticalOffset;
+
+            // calculate the vertical offset for the port index.
+            int portVerticalMidPoint = 17;
+            double portHeight = portModel.Height;
+            return verticalOffset + (index * portHeight) + portVerticalMidPoint;
```
SHKnudsen, line 665: Will this work if there are Ports from multiple nodes on the group?

The index is relative to the owner node and not the group, right?

reddyashish: Yes, you are right. Modified it to use the port index on the group.

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -649,6 +649,39 @@ internal IEnumerable<PortModel> GetGroupOutPorts(IEnumerable<NodeModel> ownerNod
                 );
         }
 
+        private double GetPortVerticalOffset(PortModel portModel, int proxyPortIndex)
+        {
+            // vertical offset accounts for the port margins
+            double verticalOffset = 20;
```
QilongTang, line 655: can you define the constants used as const in class?

src/DynamoCore/Graph/Nodes/PortModel.cs
```diff
@@ -309,8 +323,11 @@ public bool IsConnected
                 return Connectors.Any() || (UsingDefaultValue && DefaultValue != null);
             }
         }
+
         #endregion
-        
+
+        internal bool IsProxyPort { get; set; } = false;
```
QilongTang, line 329: If we add this property here, should we remove the one in PortViewModel?

QilongTang (commented):
Some comments

src/DynamoCore/Graph/Nodes/PortModel.cs
reddyashish: yes, that is already removed.

QilongTang: Great, just saw it

QilongTang:
Thank you @reddyashish , once the one remaining comment addressed, this LGTM

QilongTang (commented):
LGTM, do you see any possibility to add unit tests to guard the position when opening a file with collapsed group?

reddyashish:
Discussed with @QilongTang regarding the tests and decided that the test wouldn't be of much use in this case as the proxy ports position will be calculated internally based on the position of the annotation model. Merging this.

QilongTang (approved):

PR #12235, opened 2021-11-03
Refactor PythonEvaluator classes

pinzart90:
Reason for this PR:
As a side effect of the removal of IronPython from the Dynamo deliverables ...existing API was broken (see `The old API` below)
This PR aims to provide new APIs so cover what was broken

What changed/will be changed in this PR:
Added new Events to the PythonEvaluator classes (with a common signature)
Added a new PythonEngine base class common to all PythonEvaluator classes,
Renamed PythonEngineSelector to PythonEngineManger and moved it to DynamoServices

What was not done because of large number of changes:
Dynamo VM still calls the PythonEvaluator classes via static methods (not using the new common base class)

The old API
```
IronPythonEvaluator.OutputMarshaler.RegisterMarshaler(
                (Element element) => element.ToDSType(true));

            // Turn off element binding during iron python script execution
            IronPythonEvaluator.EvaluationBegin +=
                (a, b, c, d, e) => ElementBinder.IsEnabled = false;
            IronPythonEvaluator.EvaluationEnd += (a, b, c, d, e) => ElementBinder.IsEnabled = true;
```
The new API would looke something like this
```
            foreach (PythonEngine engine in PythonEngineManager.Instance.AvailableEngines)
            {
                engine.InputDataMarshaler.RegisterMarshaler((object o) => o);
                engine.EvaluationStarted += ((a, b, scopeSet) => { scopeSet("d", 4); });
                engine.EvaluationFinished += ((a, b) => { });
            }

            PythonEngineManager.Instance.AvailableEngines.CollectionChanged += (object sender, NotifyCollectionChangedEventArgs e) =>
            {
                if (e.Action == NotifyCollectionChangedAction.Add)
                {
                    foreach (var item in e.NewItems)
                    {
                        PythonEngine engine = (PythonEngine)item;
                        engine.InputDataMarshaler.RegisterMarshaler((object o) => o);
                        engine.EvaluationStarted += ((a, b, scopeSet) => { scopeSet("d", 4); });
                        engine.EvaluationFinished += ((a, b) => { });
                    }
                }
            }
```


src/NodeServices/PythonServices.cs
```diff
@@ -0,0 +1,272 @@
...
+        Unspecified,
+        IronPython2,
+        CPython3
+    }
+
+    /// <summary>
+    /// Singleton class that other class can access and use for query loaded Python Engine info.
+    /// TODO: Make the Singleton class public when Dynamic loading is also ready.
+    /// </summary>
+    public sealed class PythonEngineManager
```
pinzart90, line 27: this class is/will be the old PythonEngineSelector (which will be removed)

src/Libraries/PythonNodeModels/PythonEngineSelector .cs
```diff
@@ -1,22 +1,11 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using Dynamo.PythonServices;
```
pinzart90, line 1: Will remove this whole .cs file

src/Libraries/PythonNodeModels/PythonEngineSelector .cs
```diff
@@ -1,22 +1,11 @@
...
 using System.Linq;
 using System.Reflection;
 
 namespace PythonNodeModels
 {
-    /// <summary>
-    /// Enum of possible values of python engine versions.
-    /// TODO: Remove when dynamic loading logic is there then we no longer need a hard copy of the options
-    /// </summary>
-    public enum PythonEngineVersion
```
mjkkirschner, line 13: hmm, I think you can safely move this to a new assembly using a type forward if you keep the namespace and typename exactly the same.
https://docs.microsoft.com/en-us/dotnet/standard/assembly/type-forwarding

src/NodeServices/PythonServices.cs
mjkkirschner: whats the reasoning for making this public now?

pinzart90: I will be needed for the new API (PythonEngineManager.Instance.AvailableEngines).

src/NodeServices/DynamoServices.csproj
```diff
@@ -56,10 +56,17 @@
     <Compile Include="Interfaces.cs" />
     <Compile Include="MessageEvents.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
+    <Compile Include="PythonServices.cs" />
     <Compile Include="TraceSupport.cs" />
     <Compile Include="Validity.cs" />
     <Compile Include="WorkspaceEvents.cs" />
   </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\DynamoUtilities\DynamoUtilities.csproj">
```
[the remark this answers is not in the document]
pinzart90: DynamoUtilities depends on a couple of 3d party packages like NewtonSoft and AvalonEdit...but nothin from Dynamo

src/Libraries/PythonNodeModels/PythonEngineSelector .cs
pinzart90: I thought I could forward the type to a new assembly and not have to add a reference to it for all projects that use the type.
This is what I am seeing for a project that uses PythonEngineVersion and only has a reference to PythonNodeModels (the old assembly location)
image

pinzart90: Maybe I misunderstood the documentation...looks like I need to add a reference to the new assembly location to all the projects that use the PythonEngineVersion type

src/PackageDetailsViewExtension/PackageDetailsViewExtension.csproj
```diff
@@ -93,6 +93,10 @@
       <Name>PythonNodeModels</Name>
       <Private>False</Private>
     </ProjectReference>
+    <ProjectReference Include="..\NodeServices\DynamoServices.csproj">
```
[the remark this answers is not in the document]
pinzart90: In order to be able to use the PythonEngineVersion, looks like I need to add the new assembly as a reference

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -65,12 +66,12 @@ public PythonEngineVersion Engine
...
         /// </summary>
         [Obsolete(@"This method will be removed in future versions of Dynamo.
-        Please use PythonEngineSelector.Instance.AvailableEngines instead")]
+        Please use PythonEngineManager.Instance.AvailableEngines instead")]
         public static ObservableCollection<PythonEngineVersion> AvailableEngines
         {
             get
             {
-                return PythonEngineSelector.Instance.AvailableEngines;
+                return new ObservableCollection<PythonEngineVersion>(PythonEngineManager.Instance.AvailableEngines.Select(x=> x.Version));
```
pinzart90, line 74: returning a new Observable on getter will loose old subscribers...but since this method is obsolete...it should not matter that much

src/Libraries/DSIronPython/IronPythonEvaluator.cs
```diff
@@ -272,13 +272,25 @@ public static DataMarshaler OutputMarshaler
...
+        ///     Emitted immediately before execution begins
+        /// </summary>
+        [SupressImportIntoVM]
+        public static event Action<string, IList, Action<string, object>> EvaluationStarted;
+
         /// <summary>
         ///     Emitted immediately after execution ends or fails
         /// </summary>
         [SupressImportIntoVM]
         [Obsolete("Do not use! This will be subject to changes in a future version of Dynamo")]
```
mjkkirschner, line 285: should we change these messages, to note these events are not even raised any longer?

src/NodeServices/PythonServices.cs
```diff
@@ -0,0 +1,326 @@
...
+                {
+                    AvailableEngines.Add(engine);
+                }
+            }
+        }
+    }
+
+    /// <summary>
+    /// Proxy class for any Python Engine that can be loaded by Dynamo
+    /// Provides basic APIs to interact with the Python Engine
```
mjkkirschner, line 195: I was confusing this class with the dummy evaluator - but I guess this class is really just a wrapper, maybe you could modify the summary to mention that this class is intended to wrap or act as a proxy for many different python engine versions?

src/NodeServices/PythonServices.cs
```diff
@@ -0,0 +1,326 @@
...
+    /// </summary>
+    [SupressImportIntoVM]
+    public class PythonEngineProxy : IDisposable
+    {
+        /// <summary>
+        /// The version of the Python Engine connected to this proxy instance
+        /// </summary>
+        public readonly PythonEngineVersion Version;
```
mjkkirschner, line 203: I think it's advised by MSDN to use a property instead of a field in most cases for public members.

src/NodeServices/PythonServices.cs
```diff
@@ -0,0 +1,326 @@
...
+        {
+            EngineType = eType;
+            Version = version;
+        }
+
+        public void Dispose()
+        {
+            try
+            {
+                var events = typeof(PythonEngineProxy).GetEvents();
```
mjkkirschner, line 219: hmm, never seen this before - why use it?

src/NodeServices/PythonServices.cs
```diff
@@ -0,0 +1,326 @@
...
+                foreach(var eventInfo in events)
+                {
+                    var eventHandlerName = String.Empty;
+                    if (eventInfo.Name.Equals("EvaluationBegin"))
```
mjkkirschner, line 223: can we use nameof?

src/NodeServices/PythonServices.cs
```diff
@@ -0,0 +1,326 @@
...
+            OnBeginEvaluation?.Invoke(code, bindingValues, scopeFn);
+        }
+
+        private void HandleEvaluationEnd(string code, IList bindingValues)
+        {
+            OnEndEvaluation?.Invoke(code, bindingValues);
+        }
+
+
+        private void AddEventHandler(string targetEventName, string handlerName)
```
mjkkirschner, line 309: this acts dynamically because the underlying type doesn't have an interface we can use (or shared base class)?

test/Libraries/IronPythonTests/IronPythonTests.csproj
```diff
@@ -127,9 +127,15 @@
...
     </ProjectReference>
+    <ProjectReference Include="..\..\..\src\NodeServices\DynamoServices.csproj">
+      <Project>{ef879a10-041d-4c68-83e7-3192685f1bae}</Project>
+      <Name>DynamoServices</Name>
+      <Private>False</Private>
+    </ProjectReference>
     <ProjectReference Include="..\..\..\src\PythonMigrationViewExtension\PythonMigrationViewExtension.csproj">
       <Project>{10af430d-0d3a-49ce-a63d-848912959745}</Project>
       <Name>PythonMigrationViewExtension</Name>
+      <Private>False</Private>
```
mjkkirschner, line 138: good catch.

mjkkirschner:
@pinzart90 it's looking solid though I would like to understand why you went with reflection.

pinzart90:
> @pinzart90 it's looking solid though I would like to understand why you went with reflection.

I used reflection because it seemed to me like less code refactor than implementing a common interface/base class for all Evaluators (ex. Evaluators are public static classes now...refactoring them to have a common base ....would be tricky)

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -632,10 +648,17 @@ private static bool IsMarkedToSkipConversion(PyObject pyObj)
                                             string code,
                                             IList bindingValues)
         {
-            if (EvaluationEnd != null)
+            // Call deprecated events until they are completely removed.
```
mjkkirschner, line 651: 👍

test/Libraries/DynamoPythonTests/PythonEngineSelectorTests.cs
```diff
@@ -69,5 +73,45 @@ public void CanCopydAndPasteAndUndoPythonEngine()
...
+            var cPython3Eng = PythonEngineManager.Instance.CreateEngineProxy(
+                Assembly.GetAssembly(typeof(DSCPython.CPythonEvaluator)), 
+                PythonEngineVersion.CPython3);
+
+            Assert.IsNotNull(cPython3Eng);
+
+            cPython3Eng.OnEvaluationBegin((code, bindings, scopeSet) => { scopeSet("IN", new ArrayList { " ", "  " }); });
+
+            int counter = 0;
+            cPython3Eng.OnEvaluationEnd((state, code, bindings, scopeGet) => { counter++; });
```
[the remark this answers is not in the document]
mjkkirschner: oh hmm - getting confused between the proxyengine and explicit engine versions of this method... can you clarify for me?

src/NodeServices/PythonServices.cs
```diff
@@ -0,0 +1,338 @@
...
+            AddEventHandler(EvaluationStartedCallback.GetType(), nameof(HandleEvaluationStarted));
+        }
+
+        /// <summary>
+        /// Add an event handler after the Python evaluation ends
+        /// </summary>
+        /// <param name="callback"></param>
+        public void OnEvaluationEnd(EvaluationFinishedEventHandler callback)
+        {
+            EvaluationFinishedCallback += callback;
```
mjkkirschner, line 240: hmm - so if you call this multiple times will the handler be registered over and over to be called multiple times?

mjkkirschner: do these get cleaned up in the dispose call that uses reflection? If so I think it deserves some comments.

mjkkirschner (changes requested):
Hey @pinzart90 it looks good but I have some questions I would like clarified before we go ahead.

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -611,15 +647,15 @@ private static bool IsMarkedToSkipConversion(PyObject pyObj)
...
+        public override event EvaluationFinishedEventHandler EvaluationFinished;
 
         /// <summary>
         /// Called immediately before evaluation starts
         /// </summary>
         /// <param name="scope">The scope in which the code is executed</param>
         /// <param name="code">The code to be evaluated</param>
         /// <param name="bindingValues">The binding values - these are already added to the scope when called</param>
-        private static void OnEvaluationBegin(PyScope scope,
+        private void OnEvaluationBegin(PyScope scope,
```
mjkkirschner, line 658: do these static events still exist in the base class?

mjkkirschner:
the failing test https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/4839/testReport/junit/NodeDocumentationMarkdownGeneratorTests/MarkdownGeneratorCommandTests/ProducesCorrectOutputFromCoreDirectory_preloadedbinaries/ - makes it seem like something is being imported that was not previously.

src/Libraries/DSCPython/CPythonEvaluator.cs
pinzart90: No...they do not exist anymore
SInce they are private and are note accessed by reflection anywhere....we do not need to keep it static

src/NodeServices/PythonServices.cs
```diff
@@ -0,0 +1,338 @@
...
+            AddEventHandler(EvaluationFinishedCallback.GetType(), nameof(HandleEvaluationFinished));
+        }
+
+        internal PythonEngineProxy(Type eType, PythonEngineVersion version)
+        {
+            EngineType = eType;
+            Version = version;
+        }
+
+        public void Dispose()
```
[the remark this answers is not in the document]
pinzart90: No more Dispose

src/NodeServices/PythonServices.cs
```diff
@@ -0,0 +1,338 @@
...
+        {
+            try
+            {
+                if (EvaluationStartedCallback != null)
+                {
+                    MethodInfo handlerInfo = typeof(PythonEngineProxy).GetMethod(nameof(HandleEvaluationStarted), BindingFlags.NonPublic | BindingFlags.Instance);
```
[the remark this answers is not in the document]
pinzart90: No more reflection

src/NodeServices/PythonServices.cs
```diff
@@ -0,0 +1,338 @@
...
+            }
+            catch
+            { }
+        }
+
+        /// <summary>
+        /// Gets the input Marshaler from the target Python Engine
+        /// </summary>
+        /// <returns>DataMarshaler as object</returns>
+        public object GetInputMarshaler()
```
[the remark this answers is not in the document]
pinzart90: I used generic object because if I bring in the datatype DataMarshaler then I either have to add a new dependency on DYnamoUtilities or I have to move the Marshaler class to DynamoServices (and it seems like a good fit for the  Marshaler to be in DynamoUtilities)

src/NodeServices/PythonServices.cs
```diff
@@ -0,0 +1,338 @@
...
+            {
+                return null;
+            }
+        }
+
+        /// <summary>
+        /// Gets the output Marshaler from the target Python Engine
+        /// </summary>
+        /// <returns>DataMarshaler as object</returns>
+        public object GetOutputMarshaler()
```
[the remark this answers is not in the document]
pinzart90: See comment above

test/Libraries/DynamoPythonTests/PythonEngineSelectorTests.cs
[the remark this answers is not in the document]
pinzart90: Has been replaced with base class implementation

src/NodeServices/PythonServices.cs
pinzart90: not used anymore

test/Tools/NodeDocumentationMarkdownGeneratorTests/MarkdownGeneratorCommandTests.cs
```diff
@@ -163,7 +163,7 @@ public void ProducesCorrectOutputFromCoreDirectory_preloadedbinaries()
             var generatedFileNames = tempDirectory.GetFiles().Select(x => x.Name);
             //assert count is correct.
             //TODO this should be 653 - but 2 tsplines nodes have such long signatures the paths are too long for windows.
-            Assert.AreEqual(652, generatedFileNames.Count());
+            Assert.AreEqual(660, generatedFileNames.Count());
```
mjkkirschner, line 166: @pinzart I see you updated this test- but why? Should your changes be importing 4 more nodes? I would not expect that? My guess is you need to mark some public methods as [IsVisibleInDynamoLibrary(false)]

src/NodeServices/PythonServices.cs
```diff
@@ -0,0 +1,241 @@
...
+        private void ScanPythonEngine(Assembly assembly)
+        {
+            if (assembly == null)
+            {
+                return;
+            }
+
+            PythonEngine engine = null;
+            // Currently we are using try-catch to validate loaded assembly and evaluation method exist
+            // but we can optimize by checking all loaded types against evaluators interface later
```
mjkkirschner, line 225: do you mean interface or abstract base class?

mjkkirschner (changes requested):
one thing I don't understand regarding the markdown test

src/NodeServices/PythonServices.cs
pinzart90: abstract base class..will change

test/Tools/NodeDocumentationMarkdownGeneratorTests/MarkdownGeneratorCommandTests.cs
pinzart90: Ok, the new PythonEngine public methods are the extra ones. I will mark those as IsVisibleInDynamoLibrary false and ImportIntoVM false

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -131,24 +132,25 @@ public void Dispose()
...
-    [Obsolete("Do not use! This will be subject to changes in a future version of Dynamo")]
+    [Obsolete("Deprecated. Please use evaluation handlers from Dynamo.PythonServices instead.")]
     public delegate void EvaluationEventHandler(EvaluationState state, PyScope scope, string code, IList bindingValues);
 
     /// <summary>
     ///     Evaluates a Python script in the Dynamo context.
     /// </summary>
     [IsVisibleInDynamoLibrary(false)]
-    public static class CPythonEvaluator
+    public class CPythonEvaluator : Dynamo.PythonServices.PythonEngine
```
aparajit-pratap, line 146: So this is an API break, right?

src/Libraries/DSIronPython/IronPythonEvaluator.cs
```diff
@@ -32,7 +33,7 @@ public enum EvaluationState { Begin, Success, Failed }
     ///     Evaluates a Python script in the Dynamo context.
     /// </summary>
     [IsVisibleInDynamoLibrary(false)]
-    public static class IronPythonEvaluator
+    public class IronPythonEvaluator : Dynamo.PythonServices.PythonEngine
```
aparajit-pratap, line 36: API break?

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -131,24 +132,25 @@ public void Dispose()
...
     {
         private const string DynamoSkipAttributeName = "__dynamoskipconversion__";
         private const string DynamoPrintFuncName = "__dynamoprint__";
         private const string NodeName = "__dynamonodename__";
-        static PyScope globalScope;
         internal static readonly string globalScopeName = "global";
+
+        private static PyScope globalScope;
         private static DynamoLogger dynamoLogger;
         internal static DynamoLogger DynamoLogger {
```
aparajit-pratap, line 155: So these also need not be static anymore, do they?

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -704,9 +704,9 @@ internal void HideUnhideNamespace(bool hide, string library, string namespc)
         private void AddPythonEnginesOptions()
         {
             var options = new ObservableCollection<string> { Res.DefaultPythonEngineNone };
-            foreach (var item in PythonNodeModels.PythonEngineSelector.Instance.AvailableEngines)
+            foreach (var item in PythonEngineManager.Instance.AvailableEngines)
```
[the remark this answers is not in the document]
pinzart90: I seemed like a better fit
Also the old class was internal so not breaking changes

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -471,11 +495,17 @@ public static DataMarshaler InputMarshaler
...
+        /// </summary>
+        [SupressImportIntoVM]
+        public static DataMarshaler InputMarshaler => Instance.InputDataMarshaler as DataMarshaler;
+
         /// <summary>
         ///     Data Marshaler for all data coming out of a Python node.
         /// </summary>
         [SupressImportIntoVM]
-        public static DataMarshaler OutputMarshaler
+        public override object OutputDataMarshaler
```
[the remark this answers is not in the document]
pinzart90: It needs to replace an existing public API
PythonEngine.OutputDataMarshaler will be used instead of the already existing public API CPythonEvaluator.OutputMarshaler

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -432,7 +456,7 @@ private static string GetTraceBack(Exception e)
         ///     Data Marshaler for all data coming into a Python node.
         /// </summary>
         [SupressImportIntoVM]
-        public static DataMarshaler InputMarshaler
+        public override object InputDataMarshaler
```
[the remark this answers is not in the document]
pinzart90: It needs to replace an existing public API
PythonEngine.InputDataMarshaler will be used instead of the already existing public API CPythonEvaluator.InputMarshaler

src/Libraries/DSCPython/CPythonEvaluator.cs
pinzart90: Should not be an API break...since all the public methods were static already

src/Libraries/DSIronPython/IronPythonEvaluator.cs
pinzart90: Should not be an API break...since all the public methods were static already

src/Libraries/DSCPython/CPythonEvaluator.cs
pinzart90: They can still be static..or converted to singleton members
Should be no real difference for now...
I chose to leave them as they were to avoid too many changes

src/Libraries/PythonNodeModels/Properties/AssemblyInfo.cs
```diff
@@ -21,4 +21,5 @@
 [assembly: InternalsVisibleTo("PythonNodeModelsWpf")]
 [assembly: InternalsVisibleTo("PythonMigrationViewExtension")]
 [assembly: InternalsVisibleTo("DynamoCoreWpf")]
+[assembly: TypeForwardedTo(typeof(PythonNodeModels.PythonEngineVersion))]
```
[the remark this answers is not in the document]
pinzart90: Had to move PythonEngineVersion to a lower DynamoServices so that it can be easily accessed in the rest of the projects (without bringing in the PythonNodeModel's dependencies)

src/Libraries/PythonNodeModelsWpf/PythonNode.cs
```diff
@@ -176,5 +193,45 @@ private void UpdateToPython3Engine(object sender, EventArgs e)
...
+        private void PythonEnginesChanged(object sender, NotifyCollectionChangedEventArgs e)
+        {
+            if (e.Action == NotifyCollectionChangedAction.Add)
+            {
+                foreach (var item in e.NewItems)
+                {
+                    PythonEngine engine = item as PythonEngine;
+                    if (engine != null && engine.Version.Equals(PythonEngineVersion.IronPython2))
+                    {// Handle only IronPython2 for now
+                        AddIronPython2MenuItem();
```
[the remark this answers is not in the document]
pinzart90: I did not refactor the VM execution or the UI to support generic python engines yet
That would have meant a lot more changes. I will file a separate PR for that

aparajit-pratap (approved):

mjkkirschner (approved):

PR #12251, opened 2021-11-04
Fix exit modal decline terms

filipeotero:
### Purpose

This PR fixes the exit modal when the user tries to decline the terms

declineTerms

### Declarations

### Release Notes

Fix the exit guide when decline the terms of user

### Reviewers

@QilongTang 

### FYIs

@RobertGlobant20


src/DynamoCoreWpf/UI/GuidedTour/GuidesValidationMethods.cs
```diff
@@ -83,6 +121,23 @@ private static void AcceptButton_Click(object sender, RoutedEventArgs e)
...
+        /// <summary>
+        /// This method will be executed when the Decline button is pressed in the TermsOfUseView Window and creates the exit tour modal
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="e"></param>
+        internal void DeclineButton_Click(object sender, RoutedEventArgs e)
+        {
+            var exitGuide = new ExitGuide();
+            exitGuide.Height = 273;
+            exitGuide.Width = 480;
```
QilongTang, line 133: Can we set these as constants and why do we need to do these here?

filipeotero: Yes, this is because we usually read from json here. I've updated setting those values to const variables.

QilongTang: @filipeotero Can we read the setting again or get it from some other class? Usually, I prefer a single source of truth

filipeotero: Oh, I see. Now is getting from JSON file.

src/DynamoCoreWpf/UI/GuidedTour/dynamo_guides.json
```diff
@@ -269,6 +269,12 @@
 				"TooltipPointerDirection": "top_left",
 				"Width": 480,
 				"Height": 260,
+				"ExitGuide": {
+					"Height": 273,
+					"Width": 480,
+					"Title": "PackagesGuideExitTitle",
+					"FormattedText": "PackagesGuideExitAcceptTerms"
```
QilongTang, line 276: Are these two strings displayed as part of the popup @RobertGlobant20 @filipeotero if there are strings need to be localized, then lets not put them here. In general, we have been doing a good job here to only put control name, handler name in this Json. But for display content, let's not. I am not sure so just asking 😉

src/DynamoCoreWpf/UI/GuidedTour/StepUIAutomation.cs
```diff
@@ -63,5 +64,15 @@ public enum UIControlType { MENUITEM, BUTTON, FUNCTION };
         /// </summary>
         public string WindowName { get; set; }
 
+        public List<AutomaticHandlers> AutomaticHandlers { get; set; }
+    }
+
+    public class AutomaticHandlers
```
QilongTang, line 70: Please add comments for all public members if they indeed need to be/ expected to be public

QilongTang (commented):
One more comment before we move forward

src/DynamoCoreWpf/UI/GuidedTour/StepUIAutomation.cs
filipeotero: Sorry, comments added.

QilongTang (approved):

PR #12267, opened 2021-11-09
Don't fail to parse urls from warning messages that include repeated warnings.

mjkkirschner:
### Purpose

This PR fixes two issues with the documentation browser:
1. DereferencingANonePointer exception was mis spelled.

2. Sometimes during delta execution warnings will have different expressionIDs assigned for certain nodes that are composed of multiple expressions - like a multi-output node.

The end result is that the warning text is composed of multiple warnings like:

``` Dereferencing null pointer - DereferencingNonePointer.html/n Dereferencing null pointer - DereferencingNonePointer.html/n Dereferencing null pointer - DereferencingNonePointer.html ```

The code that parses url's expected that all warning strings would only be composed of a single single url - instead we make the splitting a bit smarter, by also using the same delimiter that is used to join the warnings. 

This code returns the first url found.

###
TODO 
- [x] tests

### Declarations

### Release Notes

Fixed a bug where documentation for errors could not be found, or reported illegal characters in the documentation browser.


sm6srw (approved):
LGTM!

pinzart90:
@mjkkirschner do you think that other parts of dynamo could misbehave because of the `Delta execution warnings will have different expressionIDs` issue ?

pinzart90 (approved):
LGTM

mjkkirschner:
@pinzart90 - I'm not sure, I actually think that the delta execution behavior is correct, and the first execution only containing a single expressionID is incorrect - it just happened that this masks the documentation issue when testing simple interactions.

PR #12273, opened 2021-11-10
autocomplete for CPython3

pinzart90:
Autocomplete for CPython3 Evaluator

* Most of the CPython3 autocomplete functionality was merged from https://github.com/DynamoDS/Dynamo/tree/Python3AutoComplete
* Moved the common functionality/utilities in the DynamoServices assembly
Note: most of the utilities in the new `PythonCodeCompletionProviderCommon` class were moved from the IronPythonAutoCompleteProvider

### DIFF
--- a/src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
+++ b/src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
@@ -69,15 +69,15 @@ clr.setPreload(True)
         }
 
         /// <summary>
-        /// List all of the members in a PythonModule
+        /// List all of the members in a PythonObject
         /// </summary>
-        /// <param name="module">A reference to the module</param>
+        /// <param name="pyObject">A reference to the module</param>
         /// <param name="name">The name of the module</param>
         /// <returns>A list of completion data for the module</returns>
-        private IEnumerable<Tuple<string, string, bool, ExternalCodeCompletionType>> EnumerateMembersForPyObj(object module, string name)
+        private IEnumerable<Tuple<string, string, bool, ExternalCodeCompletionType>> EnumerateMembersForPyObj(object pyObject, string name)
         {
             var items = new List<Tuple<string, string, bool, ExternalCodeCompletionType>>();
-            var d = module as PyObject;
+            var d = pyObject as PyObject;
             if (d == null)
             {
                 return items;
@@ -250,40 +250,7 @@ clr.setPreload(True)
         /// </summary>
         /// <param name="dynamoCorePath"></param>
         public override void Initialize(string dynamoCorePath)
-        {
-            var pythonLibDir = string.Empty;
-            var executionPath = Assembly.GetExecutingAssembly().Location;
-            // Determine if the Python Standard Library is available in the DynamoCore path
-            if (!String.IsNullOrEmpty(dynamoCorePath))
-            {
-                // set pythonLibDir path.
-            }
-
-            // TODO: check if any python libraries for CPython engine has to be imported. 
-            if (!Directory.Exists(pythonLibDir))
-            {
-                // import them here if needed. 
-            }
-
-            if (!String.IsNullOrEmpty(pythonLibDir))
-            {
-                // Try to load Python Standard Library Type for autocomplete
-                try
-                {
-                    var pyLibImports = String.Format("import sys\nsys.path.append(r'{0}')\n", pythonLibDir);
-                    // engine.CreateScriptSourceFromString(pyLibImports, SourceCodeKind.Statements).Execute(scope);
-                }
-                catch (Exception e)
-                {
-                    Log(e.ToString());
-                    Log("Failed to register IronPython's native library. Python autocomplete will not see standard modules.");
-                }
-            }
-            else
-            {
-                Log("Valid IronPython Standard Library not found. Python autocomplete will not see native modules.");
-            }
-        }
+        {}
         #endregion
 
         #region PythonCodeCompletionProviderCommon protected methods implementations
--- a/src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
+++ b/src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
@@ -11,7 +11,7 @@ using System.Text.RegularExpressions;
 
 namespace DSCPython
 {
-    internal class DSCPythonCodeCompletionProviderCore : PythonCodeCompletionProviderCommon, ILogSource
+    internal class DSCPythonCodeCompletionProviderCore : PythonCodeCompletionProviderCommon, ILogSource, IDisposable
     {
         #region Private members
 
@@ -74,10 +74,14 @@ clr.setPreload(True)
         /// <param name="module">A reference to the module</param>
         /// <param name="name">The name of the module</param>
         /// <returns>A list of completion data for the module</returns>
-        private IEnumerable<Tuple<string, string, bool, ExternalCodeCompletionType>> EnumerateMembersForModule(object module, string name)
+        private IEnumerable<Tuple<string, string, bool, ExternalCodeCompletionType>> EnumerateMembersForPyObj(object module, string name)
         {
             var items = new List<Tuple<string, string, bool, ExternalCodeCompletionType>>();
             var d = module as PyObject;
+            if (d == null)
+            {
+                return items;
+            }
 
             foreach (var member in d.Dir())
             {
@@ -178,16 +182,16 @@ clr.setPreload(True)
                             // Variable type
                             else if (VariableTypes.TryGetValue(name, out type))
                             {
-                                items = EnumerateMembers(type, name).Select(x => new PythonCodeCompletionDataCore(x.Item1, x.Item2, x.Item3, x.Item4, this));
+                                items = EnumerateMembersForPyObj(PyObject.FromManagedObject(type), name).Select(x => new PythonCodeCompletionDataCore(x.Item1, x.Item2, x.Item3, x.Item4, this));
                             }
                             else
                             {
                                 var mem = LookupMember(name);
-                                var pythonModule = mem as PyObject;
+                                var pythonObject = mem as PyObject;
                                 // Python Module type
-                                if (pythonModule != null)
+                                if (pythonObject != null)
                                 {
-                                    items = EnumerateMembersForModule(pythonModule, name).Select(x => new PythonCodeCompletionDataCore(x.Item1, x.Item2, x.Item3, x.Item4, this));
+                                    items = EnumerateMembersForPyObj(pythonObject, name).Select(x => new PythonCodeCompletionDataCore(x.Item1, x.Item2, x.Item3, x.Item4, this));
                                 }
                             }
                         }
@@ -321,9 +325,9 @@ clr.setPreload(True)
 
             BasicVariableTypes = new List<Tuple<Regex, Type>>();
 
-            BasicVariableTypes.Add(Tuple.Create(STRING_VARIABLE, typeof(string)));
-            BasicVariableTypes.Add(Tuple.Create(DOUBLE_VARIABLE, typeof(double)));
-            BasicVariableTypes.Add(Tuple.Create(INT_VARIABLE, typeof(int)));
+            BasicVariableTypes.Add(Tuple.Create(STRING_VARIABLE, typeof(PyString)));
+            BasicVariableTypes.Add(Tuple.Create(DOUBLE_VARIABLE, typeof(PyFloat)));
+            BasicVariableTypes.Add(Tuple.Create(INT_VARIABLE, typeof(PyInt)));
             BasicVariableTypes.Add(Tuple.Create(LIST_VARIABLE, typeof(PyList)));
             BasicVariableTypes.Add(Tuple.Create(DICT_VARIABLE, typeof(PyDict)));
 
@@ -399,7 +403,7 @@ clr.setPreload(True)
             MessageLogged?.Invoke(LogMessage.Info(message));
         }
 
-        ~DSCPythonCodeCompletionProviderCore()
+        public void Dispose()
         {
             if (Scope != null)
             {
--- a/src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
+++ b/src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
@@ -1,0 +1,806 @@
+﻿using Autodesk.DesignScript.Interfaces;
+using Dynamo.Logging;
+using Dynamo.PythonServices;
+using Python.Runtime;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Text.RegularExpressions;
+
+namespace DSCPython
+{
+    internal class DSCPythonCodeCompletionProviderCore : IExternalCodeCompletionProviderCore, ILogSource, IDisposable
+    {
+        #region Private members
+        /// <summary>
+        /// Maps a basic variable regex to a basic python type.
+        /// </summary>
+        private List<Tuple<Regex, Type>> BasicVariableTypes;
+
+        /// <summary>
+        /// Tracks already referenced CLR modules
+        /// </summary>
+        private HashSet<string> ClrModules { get; set; }
+
+        /// <summary>
+        /// Keeps track of failed statements to avoid poluting the log
+        /// </summary>
+        private Dictionary<string, int> BadStatements { get; set; }
+
+        private PythonEngine engine;
+        /// <summary>
+        /// The engine used for autocompletion.  This essentially keeps
+        /// track of the state of the editor, allowing access to variable types and
+        /// imported symbols.
+        /// </summary>
+        private object Engine
+        {
+            get { return engine; }
+            set { engine = (PythonEngine)value; }
+        }
+        private PyScope scope;
+
+        private string uniquePythonScopeName = "uniqueScope";
+
+        /// <summary>
+        /// The scope used by the engine.  This is where all the loaded symbols
+        /// are stored.  It's essentially an environment dictionary.
+        /// </summary>
+        private PyScope Scope
+        {
+            get { return scope; }
+            set { scope = (PyScope)value; }
+        }
+        #endregion
+
+        #region Private Methods
+        private PyScope CreateUniquePythonScope()
+        {
+            PyScopeManager.Global.TryGet(uniquePythonScopeName, out PyScope Scope);
+
+            if (Scope == null)
+            {
+                Scope = Py.CreateScope(uniquePythonScopeName);
+                Scope.Exec(@"
+import clr
+import sys
+clr.setPreload(True)
+");
+            }
+
+            return Scope;
+        }
+
+        private object ExecutePythonScriptCode(string code)
+        {
+            Python.Included.Installer.SetupPython().Wait();
+
+            if (!PythonEngine.IsInitialized)
+            {
+                PythonEngine.Initialize();
+                PythonEngine.BeginAllowThreads();
+            }
+
+            IntPtr gs = PythonEngine.AcquireLock();
+            try
+            {
+                using (Py.GIL())
+                {
+                    var result = Scope.Eval(code);
+                    return CPythonEvaluator.OutputMarshaler.Marshal(result);
+                }
+            }
+            finally
+            {
+                PythonEngine.ReleaseLock(gs);
+            }
+        }
+
+
+        /// <summary>
+        /// Find all import statements and import into scope.  If the type is already in the scope, this will be skipped.
+        /// </summary>
+        /// <param name="code">The code to discover the import statements.</param>
+        private void UpdateImportedTypes(string code)
+        {
+            // Detect all lib references prior to attempting to import anything
+            if (!PythonEngine.IsInitialized)
+            {
+                PythonEngine.Initialize();
+                PythonEngine.BeginAllowThreads();
+            }
+
+            IntPtr gs = PythonEngine.AcquireLock();
+            try
+            {
+                var refs = PythonCodeCompletionUtils.FindClrReferences(code);
+                foreach (var statement in refs)
+                {
+                    var previousTries = 0;
+                    BadStatements.TryGetValue(statement, out previousTries);
+                    // TODO - Why is this 3?  Should this be a constant? Is it related to knownAssembies.Length?
+                    if (previousTries > 3)
+                    {
+                        continue;
+                    }
+
+                    try
+                    {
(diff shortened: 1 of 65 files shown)


src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,806 @@
...
+                PythonEngine.BeginAllowThreads();
+            }
+
+            IntPtr gs = PythonEngine.AcquireLock();
+            try
+            {
+                using (Py.GIL())
+                {
+                    var result = Scope.Eval(code);
+                    return CPythonEvaluator.OutputMarshaler.Marshal(result);
```
pinzart90, line 92: Using the Evaluator OutputMarshaler might not be the best thing.
I am thinking of either adding a dedicated Marshaler or trying to re use the CPythonEvaluator.Evaluate method directly (probably too much refactor)

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,806 @@
...
+            }
+            finally
+            {
+                PythonEngine.ReleaseLock(gs);
+            }
+            // If unable to find matching results and expand was set to false,
+            // try again using the full namespace (set expand to true)
+            if (!items.Any() && !expand)
+            {
+                return GetCompletionData(code, true);
```
mjkkirschner, line 576: could this end up recursing infinitely?

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,806 @@
...
+            VariableTypes = FindAllVariableAssignments(code);
+        }
+
+        /// <summary>
+        /// Attempts to find all variable assignments in the code. Has basic variable unpacking support.
+        /// We don't need to check the line indices because regex matches are ordered as per the code.
+        /// </summary>
+        /// <param name="code">The code to search</param>
+        /// <returns>A dictionary of variable name and type pairs</returns>
+        private Dictionary<string, Type> FindAllVariableAssignments(string code)
```
mjkkirschner, line 249: not something for today I guess - but it would be really nice to replace all these complex regex matches throughout these classes with https://docs.python.org/3/library/ast.html#ast.parse

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,806 @@
...
+                            }
+                            // Variable type
+                            else if (VariableTypes.TryGetValue(name, out type))
+                            {
+                                items = EnumerateMembers(type, name).Select(x => new PythonCodeCompletionDataCore(x.Item1, x.Item2, x.Item3, x.Item4, this));
+                            }
+                            else
+                            {
+                                var mem = LookupMember(name);
+                                var pythonModule = mem as PyObject;
```
mjkkirschner, line 553: why does this only cover modules? Wouldn't this cover any pyobject?

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,806 @@
...
+            if (foundType != null)
+            {
+                ImportedTypes[name] = foundType;
+            }
+
+            return foundType;
+        }
+
+        /// <summary>
+        /// List all of the members in a PythonModule
```
mjkkirschner, line 352: should this say Python Object, why would this only be used on modules?

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,806 @@
...
+            return items;
+        }
+
+        /// <summary>
+        /// List all of the members in a CLR type
+        /// </summary>
+        /// <param name="type">The type</param>
+        /// <param name="name">The name for the type</param>
+        /// <returns>A list of completion data for the type</returns>
+        private IEnumerable<Tuple<string, string, bool, ExternalCodeCompletionType>> EnumerateMembers(Type type, string name)
```
mjkkirschner, line 391: have you tested that python objects (not clr objects) get auto complete? For example, try creating a list or dictionary directly in Python - does auto complete work?

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,806 @@
...
+                    {
+                        docCommand = "type(" + stub + ")" + "." + item + ".__doc__";
+                    }
+                    else
+                    {
+                        docCommand = stub + "." + item + ".__doc__";
+                    }
+
+                    object value = ExecutePythonScriptCode(docCommand);
+                    //object value = engine.CreateScriptSourceFromString(docCommand, SourceCodeKind.Expression).Execute(scope);
```
mjkkirschner, line 607: todo?

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,806 @@
...
+                // import them here if needed. 
+            }
+
+            if (!String.IsNullOrEmpty(pythonLibDir))
+            {
+                // Try to load Python Standard Library Type for autocomplete
+                try
+                {
+                    var pyLibImports = String.Format("import sys\nsys.path.append(r'{0}')\n", pythonLibDir);
+                    // engine.CreateScriptSourceFromString(pyLibImports, SourceCodeKind.Statements).Execute(scope);
```
mjkkirschner, line 666: why is this commented out?

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,806 @@
...
+            }
+
+            IntPtr gs = PythonEngine.AcquireLock();
+            try
+            {
+                using (Py.GIL())
+                {
+                    if (Scope == null)
+                    {
+                        Scope = CreateUniquePythonScope();
```
mjkkirschner, line 716: is this scope ever disposed?

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,806 @@
...
+                    PythonEngine.Initialize();
+                    PythonEngine.BeginAllowThreads();
+                }
+
+                IntPtr gs = PythonEngine.AcquireLock();
+                try
+                {
+                    using (Py.GIL())
+                    {
+                        Scope.Dispose();
```
mjkkirschner, line 795: 👍  - does Dispose get called somewhere?

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,806 @@
...
+            BadStatements = new Dictionary<string, int>();
+            Guid pythonScopeGUID = Guid.NewGuid();
+            uniquePythonScopeName += pythonScopeGUID.ToString();
+
+            // Special case for python variables defined as null
+            ImportedTypes["None"] = null;
+
+            BasicVariableTypes = new List<Tuple<Regex, Type>>();
+
+            BasicVariableTypes.Add(Tuple.Create(PythonCodeCompletionUtils.STRING_VARIABLE, typeof(string)));
```
mjkkirschner, line 697: not pystring etc?

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
pinzart90: I does not seem to work for python native types. 
Same thing for IronPythonAutoComplete..

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
pinzart90: Only 1 more recursion is possible because of the `expand` flag

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
pinzart90: renamed

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
pinzart90: Yes .. in the Wpf editor windows

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
pinzart90: fixed

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,394 @@
...
+        public event Action<ILogMessage> MessageLogged;
+
+        internal void Log(string message)
+        {
+            MessageLogged?.Invoke(LogMessage.Info(message));
+        }
+
+        public void Dispose()
+        {
+            if (Scope != null)
```
mjkkirschner, line 375: so - does `uniquePythonScopeName` ever get called more than once during the lifetime of this providerCore - if so, then shouldn't dispose somehow keep track of all the guids and dispose all the scopes - not just the last one?

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,394 @@
...
+        {
+            PyScopeManager.Global.TryGet(uniquePythonScopeName, out PyScope Scope);
+
+            if (Scope == null)
+            {
+                Scope = Py.CreateScope(uniquePythonScopeName);
+                Scope.Exec(@"
+import clr
+import sys
+clr.setPreload(True)
```
mjkkirschner, line 47: I'm unclear if this is required here:
https://github.com/DynamoDS/Dynamo/pull/10950 for more info.

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,394 @@
...
+        private IEnumerable<Tuple<string, string, bool, ExternalCodeCompletionType>> EnumerateMembersForPyObj(object pyObject, string name)
+        {
+            var items = new List<Tuple<string, string, bool, ExternalCodeCompletionType>>();
+            var d = pyObject as PyObject;
+            if (d == null)
+            {
+                return items;
+            }
+
+            foreach (var member in d.Dir())
```
mjkkirschner, line 86: I wonder what the difference is between dir and dynamicMembers

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,394 @@
...
+                            // Attempt to get type using naming
+                            Type type = expand ? TryGetTypeFromFullName(name) : TryGetType(name);
+
+                            // CLR type
+                            if (type != null)
+                            {
+                                items = EnumerateMembers(type, name).Select(x => new PythonCodeCompletionDataCore(x.Item1, x.Item2, x.Item3, x.Item4, this));
+                            }
+                            // Variable type
+                            else if (VariableTypes.TryGetValue(name, out type))
```
mjkkirschner, line 183: can you explain what the difference is between case 2 and case 3 here?

src/NodeServices/PythonServices.cs
```diff
@@ -341,4 +344,863 @@ private void RemoveEventHandler(Type handlerType, string handlerName)
...
+        /// Detect all library references given the provided code
+        /// </summary>
+        /// <param name="code">Script code to search for CLR references</param>
+        /// <returns></returns>
+        protected static List<string> FindClrReferences(string code)
+        {
+            var statements = new List<string>();
+            foreach (var line in code.Split(new[] { '\n', ';' }))
+            {
+                if (line.Contains("clr.AddReference"))
```
mjkkirschner, line 650: guess it would be good to pull this out into a const

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,394 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Text.RegularExpressions;
+
+namespace DSCPython
+{
+    internal class DSCPythonCodeCompletionProviderCore : PythonCodeCompletionProviderCommon, ILogSource, IDisposable
```
mjkkirschner, line 14: I think the name of this file is missing an `S` in DS

mjkkirschner (commented):
This looks really solid @pinzart90 - I am curious (since I can't remember) how it affects the existing completion tests (were those moved to stop building?) and if you had considered replacing them with tests of the common completion provider or on the CPython one?

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
pinzart90: uniquePythonScopeName is assigned a unique ID only on DCPythonCodeCompletionProviderCore constructor calls and only one Scope is created during the lifetime of the provider core.
So it makes sense to only cleanup that one Scope

src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs
pinzart90:
```
        /// <summary>
        /// Returns the enumeration of all dynamic member names.
        /// </summary>
        /// <remarks>
        /// This method exists for debugging purposes only.
        /// </remarks>
        /// <returns>A sequence that contains dynamic member names.</returns>
        public override IEnumerable<string> GetDynamicMemberNames()
        {
            foreach (PyObject pyObj in Dir())
            {
                yield return pyObj.ToString();
            }
        }
        ```

File: src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs (hunk 15 above)
Comment (pinzart90) [reply src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs:183]: `else if (VariableTypes.TryGetValue(name, out type))`
Handles the case of variable assignment (ex. a = [] then a.)
` else { var mem = LookupMember(name);`
Handles the case of Python Objects directly from the Scope (ex. modules: import sys then sys. )

**Comment by pinzart90** [comment]
> This looks really solid @pinzart90 - I am curious (since I can't remember) how it affects the existing completion tests (were those moved to stop building?) and if you had considered replacing them with tests of the common completion provider or on the CPython one?

Working on adding the completion tests now (will convert to use CPython)

File: src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs (hunk 13 above)
Comment (pinzart90) [reply src/Libraries/DSCPython/DCPythonCodeCompletionProviderCore.cs]: I created a static global scope which will do the Preload 
I guess the only duplication now is between the Evaluator's preloading and the Autocomplete's preloading

File: src/Libraries/DSIronPython/IronPythonCodeCompletionProviderCore.cs [hunk 18]
```diff
@@ -382,7 +382,7 @@ protected override void LogError(string msg)
             Log(msg);
         }
 
-        protected override bool ScopeHasVariable(string name)
+        internal override bool ScopeHasVariable(string name)
```
mjkkirschner, line 385: perhaps you want to use `protected internal` - this is an OR access modifier:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/protected-internal

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,428 @@
...
+        private IEnumerable<Tuple<string, string, bool, ExternalCodeCompletionType>> EnumerateMembersForPyObj(object pyObject, string name)
+        {
+            var items = new List<Tuple<string, string, bool, ExternalCodeCompletionType>>();
+            var d = pyObject as PyObject;
+            if (d == null)
+            {
+                return items;
+            }
+
+            foreach (var member in d.Dir())
```
mjkkirschner, line 98: so one thing that is a bit unclear to callers of this method is that it needs to be wrapped in the GIL block - I think - comment maybe?

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,428 @@
...
+                {
+                    Log(ex.ToString());
+                }
+            }
+
+            return items;
+        }
+
+        /// <summary>
+        /// Recursively lookup a variable in the _scope
```
mjkkirschner, line 131: this method mentions recursion, but this does not look recursive, and all it seems to do is return false if the string contains a period?

Is it a base case of recursion maybe?

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
```diff
@@ -0,0 +1,428 @@
...
+                            else if (VariableTypes.TryGetValue(name, out type))
+                            {
+                                items = EnumerateMembersForPyObj(PyObject.FromManagedObject(type), name).Select(x => new PythonCodeCompletionDataCore(x.Item1, x.Item2, x.Item3, x.Item4, this));
+                            }
+                            else
+                            {// Python objects from the Scope
+                                var mem = LookupMember(name);
+                                if (mem == null && expand)
+                                {
+                                    // There is an issue with calling Dir() on nested PyObjects (not all members are discovered)
```
mjkkirschner, line 208: perhaps this is related to the preload=true for this scope?

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
```diff
@@ -267,6 +266,7 @@ private void OnEngineChanged(object sender, System.Windows.Controls.SelectionCha
 
         private void OnScriptEditorWindowClosed(object sender, EventArgs e)
         {
+            completionProvider?.Dispose();
```
mjkkirschner, line 269: I know I suggested this - have you had time to test that a new completion provider is created when the script editor window is reopened?

pinzart90: YEs, it is

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
pinzart90: I will look into it

mjkkirschner (approved):
I had a few comments that I think you've seen, but this looks good to me.

test/Libraries/DynamoPythonTests/CodeCompletionTests.cs
```diff
@@ -1,18 +1,444 @@
...
+        [Test]
+        [Category("UnitTests")]
+        public void CanGetCompletionDataForArrayListVariable()
+        {
+            var str = "from System.Collections import ArrayList\na = ArrayList()\na.";
+            var completionProvider = new DSCPythonCodeCompletionProviderCore();
+            var matches = completionProvider.GetCompletionData(str);
+
+            Assert.AreNotEqual(0, matches.Length);
+            //Assert.AreEqual(typeof(IronPython.Runtime.PythonDictionary), matches["a"].Item3);
```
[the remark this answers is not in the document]
pinzart90: refactored

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
pinzart90: Added a comment
Also this is a private method.

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
pinzart90: renamed

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
pinzart90: I tried it with preload=true...and a bunch of other tricks...but nothing works.
Maybe it's a pythonnet bug...(I could not track it down though...)

test/Libraries/DynamoPythonTests/CodeCompletionTests.cs
```diff
@@ -1,18 +1,444 @@
...
+        }
+
```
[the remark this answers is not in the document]
pinzart90: I added tests to cover clr, sys and str builtin types
There are already tests that cover instances of types str, int, list and dictionary

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
```diff
@@ -95,14 +104,24 @@ private object ExecutePythonScriptCode(string code)
                 return items;
             }
 
+            var isCLRType = d.GetPythonType()?.ToString().Contains("CLR") ?? false;
```
mjkkirschner, line 108: @pinzart90 this is kind of ugly, in our wrapper of PythonNet we added `GetManagedObject` does that work?
https://git.autodesk.com/Dynamo/pythonnet/blob/dynamo_master/src/runtime/pyobject.cs#L1444

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
```diff
@@ -284,7 +312,9 @@ protected override object EvaluateScript(string script, PythonScriptType evalTyp
                 switch (evalType)
                 {
                     case PythonScriptType.Expression:
-                        return Scope.Eval(script);
+                        var result = Scope.Eval(script);
+                        return result.GetManagedObject();
+                        //return CPythonEvaluator.OutputMarshaler.Marshal(result);
```
mjkkirschner: comment out code - to be removed?

mjkkirschner (approved):
I had one more suggestion to try instead of the "CLR" string compare, but otherwise LGTM.

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
pinzart90: Sadly GetManagedObject does not work unless we call it on an object evaluateed as clr.GetClrType(type)
I did however refactor the code a bit so that searching for the "CLR" keyword only happens in limited cases (only CLR namespaces ...so far)

PR #12286, opened 2021-11-16
[DYN-4068] Add analytics for group operations

zeusongit:
### Purpose

Added tracking events for:

- Create Group
- Ungroup
- Delete
- Add Node to Group
- Remove Node from Group
- Add Group to Group
- Remove Group from Group
- Collapse Group
- Expand Group

Analytics repo PR: https://git.autodesk.com/Dynamo/Analytics.NET/pull/38

### Declarations

### Reviewers

@DynamoDS/dynamo


src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -1662,6 +1662,9 @@ internal AnnotationModel AddAnnotation(string titleText, string text, Guid id)
             annotationModel.ModelBaseRequested += annotationModel_GetModelBase;
             annotationModel.Disposed += (_) => annotationModel.ModelBaseRequested -= annotationModel_GetModelBase;
             AddNewAnnotation(annotationModel);
+
+            Logging.Analytics.TrackEvent(Actions.Create, Categories.GroupOperations, "New Group Created", nodes.Count());
```
QilongTang, line 1666: I feel for each of the action tracked, the combination of analytics event category and action name is enough to describe what is happen on a high level. We have been sticking to that so we may not need another detailed description of what this event is. I would maybe save it for what nodes are included rather than just the node count, what do you think. I think this applies to all the changes in this PR but I would like your opinion. There are also other examples in Dynamo code we can refer to.

QilongTang (approved):

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -787,6 +791,8 @@ private void CollapseGroupContents(bool collapseConnectors)
             if (!collapseConnectors) return;
 
             CollapseConnectors();
+
+            Analytics.TrackEvent(Actions.Collapsed, Categories.GroupOperations, "Group is collapsed");
```
QilongTang, line 795: There are still some description here missed

QilongTang (changes requested):
Oops, I found there are still some description not removed yet

zeusongit:
Removed remaining descriptions

QilongTang:
There is one regression reported: `DynamoCoreWpfTests.PackageDetailsViewExtensionTests.TestOpenDependencyDetails` can you check if it fails locally?

zeusongit:
> There is one regression reported: `DynamoCoreWpfTests.PackageDetailsViewExtensionTests.TestOpenDependencyDetails` can you check if it fails locally?

The test passes locally, also seems unrelated to this PR.
Screen Shot 2021-11-29 at 10 04 07 AM

QilongTang (approved):

PR #12297, opened 2021-11-17
disable analytics flag

pinzart90:
Changes:
Merged Disable Analytics feature that was already delivered to ALIAS
Added new functionality to start up LibG with disableFlag passed as an argument

Note: The changes in ANalytics.Net are not needed for this functionality (will revert those)


src/NodeServices/Interfaces.cs
```diff
@@ -150,7 +166,7 @@ internal interface IExtensionApplication
         /// <summary>
         /// Called when first time this application is loaded.
         /// </summary>
-        void StartUp();
+        void StartUp(ExtensionStartupParams startParams);
```
pinzart90, line 168: breaking change ....but should be ok since the interface is internal.
Only LibG implements this interface so we can make the necessary changes to that specific implementation class to keep API compatibility

mjkkirschner:
@pinzart90 something strange in the mono build.... were there any big changes made to the new binaries?

src/NodeServices/Interfaces.cs
```diff
@@ -141,6 +141,21 @@ internal interface IExecutionSession
         string SearchFile(string filename);
     }
 
+    /// <summary>
+    /// An options structure used when calling IExtensionApplication.Startup
+    /// </summary>
+    public struct ExtensionStartupParams
```
mjkkirschner, line 147: any problem with making this internal or it needs to be public to be used as a public parameter?

pinzart90: needs to be public to be used as a public parameter?

pinzart90:
> @pinzart90 something strange in the mono build.... were there any big changes made to the new binaries?

Nothing major...same as it was for the previous ALIAS release..
Will investigate

mjkkirschner:
it's also strange that the 18 smoke tests failed in the engOps build.

pinzart90:
> it's also strange that the 18 smoke tests failed in the engOps build.

Those smoke tests are trying to load ProtoGeometry.dll 
It is expected to fail until we fix LibG

src/Engine/ProtoCore/FFI/ExtensionAppLoader.cs
```diff
@@ -165,7 +167,7 @@ private void InitializeExtensionApp(Assembly assembly)
             }
 
             if (null != extesionApp)
```
aparajit-pratap, line 169: `extensionApp`

src/NodeServices/Analytics.cs
```diff
@@ -7,6 +7,11 @@ namespace Dynamo.Logging
     /// </summary>
     public static class Analytics
     {
+        /// <summary>
+        /// Disables all analytics collection (Google, ADP, etc.) for the lifetime of the process.
+        /// </summary>
+        public static bool DisableAnalytics;
```
aparajit-pratap, line 13: Is this using or setting the `AnalyticsUtils.DisableAnalyticsForProcessLifetime` property from Analytics.NET repo?

aparajit-pratap: Where is this being set?

pinzart90: This is the new public API that will enable integrators to completely shutoff Analytics

mjkkirschner: @pinzart90 is it true that this needs to be used before DynamoModel is constructed to be most effective? Can we add that to the comments?

pinzart90: done

pinzart90:
Tests passed https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/4932/

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -507,9 +506,11 @@ public struct DefaultStartConfiguration : IStartConfiguration
             /// Default Python script engine
             /// </summary>
             public string DefaultPythonEngine { get; set; }
+
             /// <summary>
             /// Disables ADP for the entire process for the lifetime of the process.
             /// </summary>
+            [Obsolete("This method will be removed in Dynamo 3.0 - please use Dynamo.Logging.Analytics.DisableAnalytics instead.")]
```
mjkkirschner, line 513: So - this is still here, but it's not used anywhere - is that correct? If so maybe we should mention that it is effectively removed and is not to be used at all.

PR #12298, opened 2021-11-17
DynamoUnits new tests for ForgeUnits nodes

mjkkirschner:
### Purpose

Adds tests for forge unit nodes
* tests covering c# zt nodes
* simple tests covering deserialization of standard and hand modified .dyns to replicate schema versions changing over time.
* tests checking behavior of interaction with new conversion node.

### Declarations

### Release Notes

Added Forge Unit Node extended automated tests.


test/DynamoCoreTests/UnitsOfMeasureTests.cs
```diff
@@ -726,32 +726,348 @@ public void UnitValues()
...
+        {
+            protected override void GetLibrariesToPreload(List<string> libraries)
+            {
+                libraries.Add("FunctionObject.ds");
+                libraries.Add("DynamoUnits.dll");
+                base.GetLibrariesToPreload(libraries);
+            }
+
+            [Test, Category("Failure")]
+            public void CanMapOverUnits()
```
mjkkirschner, line 742: The diff is a bit weird, but this is really a whitespace change I guess.

mjkkirschner:
one failure - in node docs generator looking at it.

sm6srw:
@mjkkirschner This looks good to me but I have one question: There is a lot of direct references to specific schema versions. How do we make sure that we cover new versions of schemas that we care about when we consume a new version of the unit packages? Is this just something we need to remember to check manually? I am not really sure if this is important or not.

mjkkirschner:
> @mjkkirschner This looks good to me but I have one question: There is a lot of direct references to specific schema versions. How do we make sure that we cover new versions of schemas that we care about when we consume a new version of the unit packages? Is this just something we need to remember to check manually? I am not really sure if this is important or not.

Thanks for checking it out @sm6srw - the way @saintentropy wrote the library it will actually work even if the version number is not an exact match - so I think the tests will work until they update their major version number - and that will show up during a nuget update of the ForgeUnitCLR package.

sm6srw (approved):

PR #12303, opened 2021-11-18
Update DynamoVisualProgramming.Core.nuspec

pinzart90:


tools/NuGet/template-nuget/DynamoVisualProgramming.Core.nuspec
```diff
@@ -10,16 +10,15 @@
...
-        10) DSCPython.dll</description>
+        1) DynamoApplications.dll
+        2) DynamoCore.dll
+        3) DynamoInstallDetective.dll
+        4) DynamoShapeManager.dll 
+        5) DynamoUtilities.dll
+        6) ProtoCore.dll
+        7) DesignScriptBuiltin.dll
+        8) VMDataBridge.dll
+        9) DSCPython.dll</description>
```
mjkkirschner, line 21: @pinzart90 to support Civil with minimal changes does it make sense to keep DSCPython here?

mjkkirschner (approved):
one question

mjkkirschner:
this change should go into the API changes docs etc.

tools/NuGet/template-nuget/DynamoVisualProgramming.Core.nuspec
pinzart90: Well..we never set out to get rid of DSCPython for 2.13
Keeping it around will provide at least 1 Pyuthon engine out of the box.
And yes...it will allow CIVIL and ALIAS to work with not changes in API

PR #12320, opened 2021-11-24
[DYN-4311] Group Warning Icons + Warning Bug Fixes

OliverEGreen:
### Purpose

This PR adds warning/error icons to groups when they contain nodes (or nested groups with nodes) in a warning/error state. 
Note: Groups will still display these icons even if the only warnings they contain are dismissed. This is intentionally done so that warnings cannot be hidden inside of groups, making them hard to find for users. 

- This PR also updates the MissingNode.png to a larger image resolution so as to appear smoother.
- This PR also fixes a bug with warnings not appearing for certain nodes (e.g. missing nodes). 

FxeGY7lyP9

### Declarations

### Release Notes

Adds warning/error icons to group headers. 

### Reviewers

@QilongTang
@mjkkirschner 

### FYIs
@Amoursol 
@SHKnudsen


src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -1074,10 +1093,81 @@ private bool BelongsToGroup()
...
+
+            // Subscribing to changes in the nodes' state. If a node's state 
+            // changes (e.g. to warning) we'll want to reflect this.
+            foreach (NodeModel nodeModel in nodeModels)
+            {
+                nodeModel.PropertyChanged += NodeModelOnPropertyChanged;
+            }
+
+            // Adding to this nodes from any nested groups.
+            nodeModels.AddRange
```
mjkkirschner, line 1121: so instead of checking all nested nodes couldn't you just check the `groupState` of the nested group?

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -1074,10 +1093,81 @@ private bool BelongsToGroup()
...
+                    break;
+            }
+        }
+
+
         public override void Dispose()
         {
             annotationModel.PropertyChanged -= model_PropertyChanged;
             annotationModel.RemovedFromGroup -= OnModelRemovedFromGroup;
+            foreach (NodeModel nodeModel in Nodes.OfType<NodeModel>())
```
mjkkirschner, line 1167: is this also done when a node is removed from a group?

mjkkirschner:
@QilongTang is this intended for 2.13?

QilongTang:
> @QilongTang is this intended for 2.13?

Yes, https://jira.autodesk.com/browse/DYN-4311

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
OliverEGreen: Would for sure be nicer - GroupState lives on the `AnnotationViewModel` rather than the `AnnotationModel` we have here.

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
OliverEGreen: Seems like the `AnnotationViewModel` method `HandleNodesCollectionChanges` isn't picking up the `removedFromGroup` and `addedToGroup` nodes correctly. This would be the best place to subscribe/unsubscribe nodes from the event.

QilongTang:
@reddyashish @mjkkirschner Do you mind take another look into this PR?

reddyashish (approved):

reddyashish:
@mjkkirschner Merging this unless you have any other concerns.

mjkkirschner:
@reddyashish yeah, I was still concerned that dispose was not called when the node was removed from a group, because now a bunch of handlers will be added potentially causing memory leaks. I didn't follow @OliverEGreen's comment if he fixed it or if it's an existing problem or what?

reddyashish:
Will wait for @OliverEGreen to address it.

OliverEGreen:
I didn't get it working yet because HandleNodesCollectionChanges doesn't seem to be working as it's meant to for me.

mjkkirschner:
Hi @OliverEGreen - it looks to me like `HandleNodeCollectionChange` only executes when the entire `Nodes` property is assigned or an event manually raised... is that ever done?

These changes look pretty recent - I think if you want to use this method you'll need to raise some property notifications because `Nodes` is not an observable collection.

https://github.com/DynamoDS/Dynamo/blame/72c347fc9f077b03ccfdd9981f07a51b557c93b9/src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs#L994

OliverEGreen:
> Hi @OliverEGreen - it looks to me like `HandleNodeCollectionChange` only executes when the entire `Nodes` property is assigned or an event manually raised... is that ever done?
> 
> These changes look pretty recent - I think if you want to use this method you'll need to raise some property notifications because `Nodes` is not an observable collection.
> 
> https://github.com/DynamoDS/Dynamo/blame/72c347fc9f077b03ccfdd9981f07a51b557c93b9/src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs#L994

Hi @mjkkirschner , the method seems to fire at the appropriate times for me (when adding or removing node(s) from a group). It's that the two collections - `removedFromGroup` and `addedToGroup` seem to receive nothing. 

I believe @QilongTang wrong this method, are you able to help me understand why the collections aren't being populated? I don't understand the GroupIdToCutGeometry.Keys stuff here:

```csharp
var removedFromGroup = GroupIdToCutGeometry.Keys
                .ToList()
                .Except(allGroupedGroups.Select(x => x.GUID.ToString()));
```

SHKnudsen:
> > Hi @OliverEGreen - it looks to me like `HandleNodeCollectionChange` only executes when the entire `Nodes` property is assigned or an event manually raised... is that ever done?
> > These changes look pretty recent - I think if you want to use this method you'll need to raise some property notifications because `Nodes` is not an observable collection.
> > https://github.com/DynamoDS/Dynamo/blame/72c347fc9f077b03ccfdd9981f07a51b557c93b9/src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs#L994
> 
> Hi @mjkkirschner , the method seems to fire at the appropriate times for me (when adding or removing node(s) from a group). The problem I'm facing is that the two collections - `removedFromGroup` and `addedToGroup` don't seem to receive anything.
> 
> I believe @QilongTang wrong this method, are you able to help me understand why the collections aren't being populated? I don't understand the GroupIdToCutGeometry.Keys stuff here:
> 
> ```cs
> var removedFromGroup = GroupIdToCutGeometry.Keys
>                 .ToList()
>                 .Except(allGroupedGroups.Select(x => x.GUID.ToString()));
> ```

The `HandleNodesCollectionChanges` only cares about nested groups, it is used to make masks in a parent group's background where the nested group is positioned. This method currently does not handle NodeModels at all, which i believe is what you are looking for, so you would have to modify it to use it for that.

From what i understand is happening here i would say it would be much easier to have all of the `GroupState` logic on the AnnotationModel, as that already subscribe and unsubscribe to PropertyChanged on every node. 
https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Graph/Annotations/AnnotationModel.cs#L390

I get that you need the groupState on the viewModel because of binding, but you can easily have a property there as well that just mirrors the AnnotationModels GroupState property.

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -254,6 +255,7 @@ public bool IsExpanded
                 WorkspaceViewModel.HasUnsavedChanges = true;
                 AddGroupToGroupCommand.RaiseCanExecuteChanged();
                 RaisePropertyChanged(nameof(IsExpanded));
+                AnnotationModel.UpdateErrorAndWarningIconVisibility();
```
SHKnudsen, line 257: Might make sense to move this call to the AnnotaionModel?
https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Graph/Annotations/AnnotationModel.cs#L327

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -540,6 +542,8 @@ public AnnotationViewModel(WorkspaceViewModel workspaceViewModel, AnnotationMode
                 SetGroupOutPorts();
                 CollapseGroupContents(true);
             }
+
+            AnnotationModel.UpdateErrorAndWarningIconVisibility();
```
SHKnudsen, line 545: Could this be on the AnnotationModel ctor instead?

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
OliverEGreen: Adjusted this method to check the GroupState of the nested group now.

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
OliverEGreen: This is taken care of with existing `model_PropertyChanged` on `AnnotationModel` now.

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -1043,6 +1048,8 @@ private void HandleNodesCollectionChanges()
 
                 AddToCutGeometryDictionary(groupViewModel);
             }
+
+            AnnotationModel.UpdateErrorAndWarningIconVisibility();
```
SHKnudsen, line 1051: Whats the reason for calling this here? could we not just call it when nodes are added or removed from the AnnotationModel, here for example https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Graph/Annotations/AnnotationModel.cs#L607

src/Libraries/CoreNodeModelsWpf/NodeViewCustomizations/DummyNode.cs
```diff
@@ -14,13 +15,18 @@ public void CustomizeView(DummyNode model, Dynamo.Controls.NodeView nodeView)
...
-            Image dummyNodeImage = new Image()
+            Image dummyNodeImage = new Image
             {
-                Stretch = System.Windows.Media.Stretch.None,
+                Width = 66.0,
+                Height = 66.0,
+                Stretch = Stretch.UniformToFill,
                 Source = new BitmapImage(new Uri(src, UriKind.Relative))
             };
+            RenderOptions.SetBitmapScalingMode(dummyNodeImage, BitmapScalingMode.HighQuality);
```
OliverEGreen, line 29: Hmm, odd this is part of a different branch.
This branch is up to date with upstream/master.

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
OliverEGreen: Makes sense, have removed all references from AnnotationViewModel and set to Private.

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
OliverEGreen: Moved

src/DynamoCore/Graph/Annotations/AnnotationModel.cs
```diff
@@ -331,6 +332,23 @@ public bool IsExpanded
...
+            }
+        }
+
+        private ElementState groupState = ElementState.Active;
+
+        /// <summary>
+        /// Indicates whether the group contains nodes that are in an info/warning/error state.
+        /// This includes the state of any nodes that are in nested groups.
+        /// </summary>
+        public ElementState GroupState
```
mjkkirschner, line 345: @OliverEGreen @reddyashish is this supposed to be serialized to the .dyn? Should it have a [JsonIgnore] attribute applied?

OliverEGreen: It doesn't need to be serialised. 
Will add tomorrow

OliverEGreen: Added!

reddyashish:
Thank you @OliverEGreen. @mjkkirschner Verified that the GroupState property is not serialized. Merging this PR.

PR #12332, opened 2021-11-26
Add New Node GetNumber

chuongmep:
### Purpose

Support Get Number In String.This is very helpful, I have come across forum posts within the past year, users need it for object separation, counting and statistics, checking,...
image
Node In Action 
image

### Declarations

- [ ] User facing strings, if any, are extracted into *.resx files

### Reviewers

@QilongTang @mjkkirschner 

### FYIs

@Amoursol


src/Libraries/CoreNodes/String.cs
```diff
@@ -38,6 +38,24 @@ public static object ToNumber(string @string)
             throw new ArgumentException(Properties.Resources.StringToNumberInvalidNumberMessage, "str");
         }
 
+        /// <summary>
+        /// Get Number In String
```
QilongTang, line 42: Would you be more specific about the description? e.g. get first number string from target string, etc or get all of the number string from target string as a list

src/Libraries/CoreNodes/String.cs
```diff
@@ -38,6 +38,24 @@ public static object ToNumber(string @string)
...
+        /// </summary>
+        /// <param name="string">String to be get</param>
```
QilongTang, line 44: `Target string`

QilongTang:
Our team can add node icon later but so far I feel this would be a good addition to the existing node base. Thank you for proposing this.

src/Libraries/CoreNodes/String.cs
chuongmep: I don't know how many problems like this from user, I have had to write an entire article, you can check it out here
https://chuongmep.com/Number-In-String

src/Libraries/CoreNodes/String.cs
chuongmep: Edited.

src/Libraries/CoreNodes/String.cs
chuongmep: Has updated to : "Get all of the number strings from the target string as a string"

src/Libraries/CoreNodes/String.cs
```diff
@@ -38,6 +38,24 @@ public static object ToNumber(string @string)
             throw new ArgumentException(Properties.Resources.StringToNumberInvalidNumberMessage, "str");
         }
 
+        /// <summary>
+        /// Get all of the number strings from the target string as a string
```
QilongTang, line 42: By reading the code, you are actually returning the first number string, right? I guess I was referring to if it's worth to return a list of them if there are multiple numbers

chuongmep: It returns all numbers, including if they have spaces or more letters inside the string, the numbers will be merged.

chuongmep: image

chuongmep: Actually we still need the first or last zero in the string to do a lot of things, so I had to return the string result of the number

Amoursol:
This is awesome @chuongmep - thank you for submitting this and making Dynamo better! 🙌

src/Libraries/CoreNodes/String.cs
```diff
@@ -38,6 +38,25 @@ public static object ToNumber(string @string)
...
+        /// <search>getnumber,tonumber,strtonumber,numberinstring,string2number,stringtonumber,int,double,cast</search> 
+        public static string GetNumber(string @string)
+        {
+            string str = string.Empty;
+            if (string.IsNullOrEmpty(@string)) return str;
+            List<double> result = new List<double>();
+            for (int i = 0; i < @string.Length; i++)
+            {
+                if (char.IsDigit(@string[i]))
+                    str += @string[i];
```
mjkkirschner, line 55: this might be a good case to use a string builder - because this string could be arbitrarily long.
https://michaelscodingspot.com/challenging-the-c-stringbuilder/

QilongTang:
Regression found `DSCoreNodesTests.StringTests.GetNumber`

QilongTang (approved):
LGTM

QilongTang:
Seems Dynamo SelfServe PR check not triggered correctly for this PR. I am running on demand job at https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/917/ and monitoring the results. FYI: @DynamoDS/dynamo

QilongTang:
I see the following regression result, can you take another look?
```
Failed : DSCoreNodesTests.StringTests.GetNumber
  Expected string length 12 but was 5. Strings differ at index 2.
  Expected: "01 Level 200"
  But was:  "01200"
```

chuongmep:
> 
@QilongTang  sorry about that, updated.

PR #12335, opened 2021-11-27
Update to Adp4 and use standard ADP consent dialog.

mjkkirschner:
### Purpose

related: https://git.autodesk.com/Dynamo/Analytics.NET/pull/42

Screen Shot 2021-11-29 at 4 29 45 PM

This PR attempts to update to ADP4 and accommodate the changes to the the API.
* modifies our current usageReportingDialog - ADP consent text is slightly modified, and handed off to the standard ADP dialog which can be brought up with a new button.
* The dialog is dynamic and handles many cases for us, including new legal wording, timeout, syncing, localization, and differing numbers of consent checkboxes per user, and per locale.
* Moves to use `IsOptedIn` with a timeout to handle the case where user prefs are not ready - set to 30 seconds.
* No longer calls `SetsOptIn` - only the consent dialog does this.
* Uses a new API for checking if ADP API is available before enabling consent dialog access.

### Declarations

### Release Notes

Update to ADP4


src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1124,7 +1124,8 @@ private void OnAsyncTaskStateChanged(DynamoScheduler sender, TaskStateChangedEve
                         long end = e.Task.ExecutionEndTime.TickCount;
                         var executionTimeSpan = new TimeSpan(end - start);
 
-                        if (Logging.Analytics.ReportingAnalytics)
+                        //don't attempt to send these events unless GA is active or ADP will actually record these events.
+                        if (Logging.Analytics.ReportingGoogleAnalytics || AnalyticsService.IsADPCollectionEnabled(Actions.Run,Categories.NodeOperations))
```
pinzart90, line 1128: is there a significant performance in checking `IsADPCollectionEnabled` here...rather than checking it directly in Analytics.NET.ADP ?
Deferring the check to Analytics.NET.ADP would simplify the changes in Dynamo...but might be a bit slower...

src/DynamoCore/Logging/AnalyticsService.cs
```diff
@@ -65,9 +68,10 @@ internal static bool IsADPOptedIn
                 {
                     return false;
                 }
-
-                return adpAnalyticsUI.IsOptedIn();
+                //TODO add this overload to IAnalyticsUI.
+                return (adpAnalyticsUI as ADPAnalyticsUI ).IsOptedIn(30000);
```
pinzart90, line 72: if you add a reference to `ADPAnalyticsUI` here...then, while calling  `IsADPOptedIn` the ADP assembly will be loaded even if `DisableAnalytics` will be set to true.
DotNet resolves its types when parsing the function body...even if the type's function/property is never called ...

src/DynamoCore/Logging/DynamoAnalyticsClient.cs
```diff
@@ -127,7 +127,7 @@ public bool ReportingAnalytics
         /// <summary>
         /// Return if Google Analytics Client is allowed to send analytics info
         /// </summary>
-        private bool ReportingGoogleAnalytics
+        public bool ReportingGoogleAnalytics
```
pinzart90, line 130: why public?

src/DynamoCore/Logging/AnalyticsService.cs
```diff
@@ -87,5 +92,23 @@ internal static void ShutDown()
...
+        public static void ShowADPConsetDialog(IntPtr? host)
+        {
+            adpAnalyticsUI.ShowOptInDialog(System.Threading.Thread.CurrentThread.CurrentUICulture.Name, false, host);
+        }
+
+        /// <summary>
+        /// Check if this event will be logged to ADP, or if it will be filtered out
+        /// by optin status or ADP allow listing.
+        /// </summary>
+        public static bool IsADPCollectionEnabled(Actions action, Categories category)
```
pinzart90, line 109: does this need to be public?

src/DynamoCore/Logging/AnalyticsService.cs
```diff
@@ -87,5 +92,23 @@ internal static void ShutDown()
         {
             Analytics.ShutDown();
         }
+
+        /// <summary>
+        /// Show the ADP dynamic consents dialog.
+        /// </summary>
+        /// <param name="host">main window</param>
+        public static void ShowADPConsetDialog(IntPtr? host)
```
pinzart90, line 100: needs to be public?

src/DynamoCoreWpf/UI/Prompts/UsageReportingAgreementPrompt.xaml.cs
```diff
@@ -34,36 +36,18 @@ public UsageReportingAgreementPrompt(IBrandingResourceProvider resourceProvider,
...
 
             AcceptGoogleAnalyticsCheck.IsChecked = UsageReportingManager.Instance.IsAnalyticsReportingApproved;
 
             if (Analytics.DisableAnalytics)
             {
-                AcceptADPAnalyticsCheck.IsChecked = false;
-                AcceptADPAnalyticsTextBlock.IsEnabled = false;
-                AcceptADPAnalyticsCheck.IsEnabled = false;
-
+                //TODO anything to do here?
```
pinzart90, line 45: maybe disable the ADP button ?

src/DynamoCoreWpf/UI/Prompts/UsageReportingAgreementPrompt.xaml.cs
```diff
@@ -103,5 +86,15 @@ protected override void OnClosed(System.EventArgs e)
             base.OnClosed(e);
             viewModel = null;
         }
+
+        private void configure_adp_button_Click(object sender, RoutedEventArgs e)
+        {
+            IntPtr handle = new IntPtr();
```
pinzart90, line 92: do we always need an owner handle ?
should this be initialized to null ?

pinzart90:
Using `IsCollectionEnabled` in Dynamo complicates the PR a bit much.
@mjkkirschner do you see performance gains with using such a granular check ? The alternative would be to check `IsCollectionEnabled` on ANalytics.Net side...on every TrackEvent call

src/DynamoCore/Models/DynamoModel.cs
mjkkirschner: Sorry @pinzart90 I'm not sure what you mean - is what you mean here similar to your suggestion below to check `IsADPCollectionEnabled` in the `TrackEvent` calls?

If so - this is not what the sdk docs suggest. I've looked at the source a bit and I'm not sure what the performance impact could be - maybe we can look it over together.
My worry is that since the api docs say it checks if the user is opted out, and that the call is blocking - that in the worst case this might mean we inadvertently introduce a blocking call that waits on a server into every single track event call.

src/DynamoCore/Logging/AnalyticsService.cs
mjkkirschner: So you are saying by using this `IAnalyticsUI` we can avoid loading the ADP binary when `DisableAnalytics` - so if I complete this TODO everything is fine?

Are there tests that assert when `DisableAnalytics` is true that the binaries are not loaded? It's hard to reason about otherwise.

src/DynamoCore/Logging/AnalyticsService.cs
mjkkirschner: can't recall - will try making internal

src/DynamoCore/Logging/AnalyticsService.cs
mjkkirschner: can't recall - will try making internal

src/DynamoCore/Logging/DynamoAnalyticsClient.cs
mjkkirschner: I needed to make this public to satisfy the changes to `IAnalyticsClient.cs` and to avoid adding new references.

src/DynamoCoreWpf/UI/Prompts/UsageReportingAgreementPrompt.xaml.cs
mjkkirschner: yes, thats a good idea.

src/DynamoCoreWpf/UI/Prompts/UsageReportingAgreementPrompt.xaml.cs
mjkkirschner: We `can` use null, but ADP dev suggested not leaving it null if possible.

src/DynamoCore/Logging/AnalyticsService.cs
pinzart90: `so if I complete this TODO everything is fine` I think so
`Are there tests that assert when DisableAnalytics is true that the binaries are not loaded? ` I'll take a look. I think I can add it if it does not already exist

src/DynamoCore/Models/DynamoModel.cs
pinzart90: ` is what you mean here similar to your suggestion below to check IsADPCollectionEnabled in the TrackEvent calls?`
Yes

src/DynamoCore/Logging/AnalyticsService.cs
```diff
@@ -65,9 +67,9 @@ internal static bool IsADPOptedIn
                 {
                     return false;
                 }
-
-                return adpAnalyticsUI.IsOptedIn();
+                return adpAnalyticsUI.IsOptedIn(150,200);
```
pinzart90, line 70: maybe add a private readonly OPTIN_RETRIES, OPTIN_SLEEP or maybe just a comment on the function call
`adpAnalyticsUI.IsOptedIn(/*retries*/150, /*sleep*/200);

src/NodeServices/Analytics.cs
```diff
@@ -58,6 +58,7 @@ internal static bool IsEnabled
         /// </summary>
         public static bool ReportingAnalytics { get { return client != null && client.ReportingAnalytics; } }
 
+
```
pinzart90, line 61: you can remove this change

src/DynamoCore/Logging/AnalyticsService.cs
```diff
@@ -78,6 +80,12 @@ internal static bool IsADPOptedIn
 
                 adpAnalyticsUI.SetOptedIn(value);
             }
+            
+        }
+
+        internal static bool IsADPAvailable()
+        {
+            return adpAnalyticsUI.IsProviderAvailable();
```
pinzart90, line 88: you should check that `if (!Analytics.DisableAnalytics && adpAnalyticsUI != null)`

pinzart90:
@mjkkirschner don't forget to use the Release build of the ADP binaries 
Also please check that starting DynamoSandbox with `-da` flag (DisableAnalytics) does not load ADP or Google binaries.

pinzart90 (approved):
Couple of comments then LGTM

src/DynamoCore/Logging/AnalyticsService.cs
```diff
@@ -87,5 +95,17 @@ internal static void ShutDown()
         {
             Analytics.ShutDown();
         }
+
+        /// <summary>
+        /// Show the ADP dynamic consents dialog.
+        /// </summary>
+        /// <param name="host">main window</param>
+        internal static void ShowADPConsetDialog(IntPtr? host)
```
aparajit-pratap, line 103: typo: consent

src/DynamoCoreWpf/Properties/Resources.en-US.resx
```diff
@@ -2978,4 +2978,7 @@ To install the latest version of a package, click Install. \n
   <data name="PackageContextMenuDeletePackageCustomNodesInUseTooltip" xml:space="preserve">
     <value>This package contains custom nodes that are in use. These custom nodes need to be deleted or the graph needs to be closed before the package can be deleted.</value>
   </data>
+  <data name="ConfigureADPButtonText" xml:space="preserve">
+    <value>Configure Autodesk Analytics Program</value>
```
aparajit-pratap, line 2982: Does this button bring up the ADP dynamic consent dialog?

mjkkirschner: yes

src/DynamoWPFCLI/Program.cs
```diff
@@ -16,6 +16,11 @@ static internal void Main(string[] args)
             {
                 var cmdLineArgs = StartupUtils.CommandLineArguments.Parse(args);
                 var locale = StartupUtils.SetLocale(cmdLineArgs);
+                if (cmdLineArgs.DisableAnalytics)
+                {
+                    Dynamo.Logging.Analytics.DisableAnalytics = true;
```
aparajit-pratap, line 21: `Dynamo.Logging.Analytics.DisableAnalytics = cmdLineArgs.DisableAnalytics;`?

pinzart90: kind of the same thing
We only set it if `cmdLineArgs.DisableAnalytics` is true

mjkkirschner:
passed tests here: https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/4981/

PR #12338, opened 2021-11-29
Fix default package search filter results.

reddyashish:
### Purpose

This PR is to sort the default package search results in descending order, by default.

Screen Shot 2021-11-29 at 11 29 50 AM

### Declarations

### Release Notes

Fix default package search filter results.

### Reviewers
@QilongTang

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
@@ -796,6 +796,8 @@ namespace Dynamo.PackageManager
                 this.AddToSearchResults(result);
             }
 
+            this.Sort();
+
             SearchState = HasNoResults ? PackageSearchState.NoResults : PackageSearchState.Results;
         }
 
@@ -895,7 +897,13 @@ namespace Dynamo.PackageManager
                         PackageManagerClientViewModel.AuthenticationManager.HasAuthProvider,
                         CanInstallPackage(x.Name), isEnabledForInstall)))
                     .ToList();
+
                 Sort(list, this.SortingKey);
+
+                if (SortingDirection == PackageSortingDirection.Descending)
+                {
+                    list.Reverse();
+                }
             }
 
             foreach (var x in list)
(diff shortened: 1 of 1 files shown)


QilongTang:
hi @reddyashish Were we not sorting the list before when initializing the dialog?

reddyashish:
They were being sorted in ascending order initially for any key. Since the default filter is being changed to descending order of LastUpdate, we had to update order.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
```diff
@@ -895,7 +897,13 @@ internal IEnumerable<PackageManagerSearchElementViewModel> Search(string query)
                         PackageManagerClientViewModel.AuthenticationManager.HasAuthProvider,
                         CanInstallPackage(x.Name), isEnabledForInstall)))
                     .ToList();
+
                 Sort(list, this.SortingKey);
+
+                if (SortingDirection == PackageSortingDirection.Descending)
+                {
+                    list.Reverse();
+                }
```
QilongTang, line 906: Maybe this can be part of the Sort function? So other functions would get the reverse behavior for free?

reddyashish: That is already being done inside the Sort(), which will sort the SearchResults in the View. This case is when the package search window opens and the complete list of packages are showed. We want this list in descending order.

QilongTang: Sorry for the confusion.. I mean do we need to make this part of the function `private static void Sort(List<PackageManagerSearchElementViewModel> results, PackageSortingKey key)`? The one we are calling above. Otherwise, we have two sorting function behaving differently, one respect `PackageSortingDirection` but the other do not

QilongTang: @reddyashish I see. If you prefer that way, I am fine with it. Would you update the function description to reflect that just so ppl know when calling `sort()` in the future?

reddyashish: Sort() is sorting the results by key and then by sorting direction. I have updated the function description.

QilongTang (approved):

PR #12346, opened 2021-11-30
update iron python warnings

pinzart90:
### Purpose

Updated IronPython migration wording
Fixed python node warning bubble disappearance issue

https://jira.autodesk.com/browse/DYN-4362
https://jira.autodesk.com/browse/DYN-4367

### Release Notes

Fixed IronPython messaging and warning bubble

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
@@ -1130,7 +1130,7 @@ namespace Dynamo.ViewModels
             if (DynamoViewModel == null) return;
 
             bool hasErrorOrWarning = NodeModel.IsInErrorState || NodeModel.State == ElementState.Warning || NodeModel.State == ElementState.PersistentWarning;
-            if (!NodeModel.WasInvolvedInExecution || !hasErrorOrWarning) return;
+            if (!NodeModel.WasInvolvedInExecution && !hasErrorOrWarning) return;
 
             if (ErrorBubble == null) BuildErrorBubble();
 
--- a/src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
@@ -1127,8 +1127,13 @@ namespace Dynamo.ViewModels
 
         public void UpdateBubbleContent()
         {
+            if (DynamoViewModel == null) return;
+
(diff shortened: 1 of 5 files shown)


src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
```diff
@@ -1127,8 +1127,13 @@ private void DisposeErrorBubble()
 
         public void UpdateBubbleContent()
         {
+            if (DynamoViewModel == null) return;
+
+            bool hasErrorOrWarning = NodeModel.IsInErrorState || NodeModel.State == ElementState.Warning || NodeModel.State == ElementState.PersistentWarning;
+            if (!NodeModel.WasInvolvedInExecution || !hasErrorOrWarning) return;
```
pinzart90, line 1133: I changed the condition here a bit
Seems to me like if the node WasInvolvedInExecution then there is no need to try and create a bubble

pinzart90: unless there can be warning bubbles that do not have anything to do with execution
@QilongTang FYI

mjkkirschner: yes, @pinzart90 - there are warning methods that clients can call at time directly - for Example
`Warning()` 
https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Graph/Nodes/NodeModel.cs#L1714

pinzart90: I'll just put it back as it was...

QilongTang: I think you mean node was not involved in execution then do not create a bubble? Yes, this is the current state. However, uur team have a design ready to inject other information regardless if the node is involved in execution or not

mjkkirschner: here is another example:
https://github.com/DynamoDS/Dynamo/blob/80e2bb6b738f3ec7beea699d772251b2faeed708/src/DynamoCore/Models/DynamoModel.cs#L1519

QilongTang (approved):

mjkkirschner:
@pinzart90 is this intended for 2.13?

pinzart90:
> @pinzart90 is this intended for 2.13?

Yes

PR #12350, opened 2021-12-01
AGD-2323 : add more control over custom package load options

BogdanZavu:
### Purpose
Add more control and clarity over custom package load options.

### Declarations
Sometimes we want to be able to load packages from specific locations that don't necessarily have an entry in the pathManager. In the previous form of this utility, we had to temporarily manipulate pathManger to allow our packages to load.
On the same note, we also don't want to create a temporary Preferences instance or modify the existing one.

### Reviewers
@mjkkirschner @aparajit-pratap @saintentropy

### DIFF
--- a/src/DynamoPackages/PackageLoader.cs
+++ b/src/DynamoPackages/PackageLoader.cs
@@ -376,13 +376,13 @@ namespace Dynamo.PackageManager
         /// <param name="packagePaths"></param>
         /// <param name="nodePaths"></param>
         /// <param name="customNodeManager"></param>
-        /// <param name="uninstallPackageDir">
+        /// <param name="uninstallPackageDirs">
         /// After the next Dynamo restart, the package will be in an unloaded state (if it is a built-in package) or 
-        /// deleted from package locations if it's path is among uninstallPackeDir.
+        /// deleted from package locations if it's path is among uninstallPackeDirs.
         /// </param>
         /// <param name="checkPathInPathManager">Load new packages only if their path is accounted for in path manager</param>
         public void LoadCustomNodesAndPackages(IEnumerable<string> packagePaths, IEnumerable<string> nodePaths, 
-            CustomNodeManager customNodeManager, IEnumerable<string> uninstallPackageDir = null, bool checkPathInPathManager = true)
+            CustomNodeManager customNodeManager, IEnumerable<string> uninstallPackageDirs = null, bool checkPathInPathManager = true)
         {
             foreach (var path in nodePaths)
             {
@@ -405,7 +405,7 @@ namespace Dynamo.PackageManager
                     }
                 }
 
-                ScanPackageDirectories(packageDirectory, uninstallPackageDir);
+                ScanPackageDirectories(packageDirectory, uninstallPackageDirs);
             }
 
             if (pathManager != null)
--- a/src/DynamoPackages/PackageLoader.cs
+++ b/src/DynamoPackages/PackageLoader.cs
@@ -373,12 +373,13 @@ namespace Dynamo.PackageManager
         /// <summary>
         /// This method is called when custom nodes and packages need to be reloaded if there are new package paths.
         /// </summary>
-        /// <param name="newPaths"></param>
-        /// <param name="preferences"></param>
+        /// <param name="packagePaths"></param>
+        /// <param name="nodePaths"></param>
         /// <param name="customNodeManager"></param>
-        public void LoadCustomNodesAndPackages(IEnumerable<string> newPaths, IPreferences preferences, CustomNodeManager customNodeManager)
+        public void LoadCustomNodesAndPackages(IEnumerable<string> packagePaths, IEnumerable<string> nodePaths, 
+            CustomNodeManager customNodeManager, IEnumerable<string> uninstallPackageDir = null, bool checkPathInPathManager = true)
         {
-            foreach (var path in preferences.CustomPackageFolders)
+            foreach (var path in nodePaths)
             {
                 // Append the definitions subdirectory for custom nodes.
                 var dir = path == DynamoModel.BuiltInPackagesToken ? PathManager.BuiltinPackagesDirectory : path;
@@ -387,13 +388,19 @@ namespace Dynamo.PackageManager
                 customNodeManager.AddUninitializedCustomNodesInPath(dir, false, false);
             }
 
-            foreach (var path in newPaths)
+            foreach (var path in packagePaths)
             {
-                var packageDirectory = pathManager.PackagesDirectories.FirstOrDefault(x => x.StartsWith(path));
-                if (packageDirectory != null)
+                var packageDirectory = path;
+                if (checkPathInPathManager)
                 {
-                    ScanPackageDirectories(packageDirectory, preferences);
+                    packageDirectory = pathManager.PackagesDirectories.FirstOrDefault(x => x.StartsWith(path));
+                    if (packageDirectory == null)
+                    {
+                        continue;
+                    }
                 }
+
+                ScanPackageDirectories(packageDirectory, uninstallPackageDir);
             }
 
             if (pathManager != null)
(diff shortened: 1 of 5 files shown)


src/DynamoPackages/PackageLoader.cs
```diff
@@ -373,12 +373,13 @@ public void LoadPackages(IEnumerable<Package> packages)
         /// <summary>
         /// This method is called when custom nodes and packages need to be reloaded if there are new package paths.
         /// </summary>
-        /// <param name="newPaths"></param>
-        /// <param name="preferences"></param>
+        /// <param name="packagePaths"></param>
+        /// <param name="nodePaths"></param>
         /// <param name="customNodeManager"></param>
-        public void LoadCustomNodesAndPackages(IEnumerable<string> newPaths, IPreferences preferences, CustomNodeManager customNodeManager)
+        public void LoadCustomNodesAndPackages(IEnumerable<string> packagePaths, IEnumerable<string> nodePaths,
```
mjkkirschner, line 379: 1. adding optional parameters is an API break resulting in a missing method exception: see: https://stackoverflow.com/questions/1456785/a-definitive-guide-to-api-breaking-changes-in-net / Adding a parameter with a default value.


2. changing parameter names on a public method is a source level break, code that uses named parameters will no longer compile.


please add a new method instead.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -411,7 +418,7 @@ public void LoadCustomNodesAndPackages(IEnumerable<string> newPaths, IPreference
             if (LocalPackages.Any())
             {
                 // Load only those recently addeed local packages (that are located in any of the new paths)
```
mjkkirschner, line 420: this comment now seems incorrect... I'd like to understand how/if these changes modify existing behavior where this method was used - it's important we don't try to load packages that already are loaded to avoid throwing errors that confuse users.

src/DynamoPackages/PackageLoader.cs
BogdanZavu: This API was recently introduced to replace the other version of LoadCustomNodesAndPackages at line 432 which is now obsolete. However, this is not working for us quite as we expected and I believe we are the only ones that attempted to use it. Since there was no official release with this I thought we would simply be better off by changing it rather than creating another variant. Do you still think we should have a new method?

src/DynamoPackages/PackageLoader.cs
BogdanZavu: There is no change in existing behavior. There was only one call internally to this utility in PackagePathViewModel and that was adapted accordingly.

src/DynamoPackages/PackageLoader.cs
mjkkirschner: No - you are correct @BogdanZavu, thanks for correcting me, I didn't see that @aparajit-pratap just added this method.

mjkkirschner:
@BogdanZavu just a process thing - usually we try to get in master first, review, then cherry pick to the RC branch just to avoid the commits not making it to master.

BogdanZavu:
@mjkkirschner yes I know, I was planning to discuss this within my team first before expanding it to a larger audience but you picked it up right away :) . I will make the PR for master first.

aparajit-pratap:
@BogdanZavu please explain the rationale for these changes in the PR description.

mjkkirschner (approved):

QilongTang:
@BogdanZavu Would you also update if there are any UI differences after your PR?

src/DynamoPackages/PackageLoader.cs
aparajit-pratap: New paths was meant to be specifically `new` paths. I'm not sure I'm okay with us changing this to be any path. ⚠️

aparajit-pratap: `nodePaths` and `packagePaths`? What's the difference? I can see how it can get confusing for API users/deveopers.

aparajit-pratap: unless you're referring to `nodePaths `as custom node paths only, in which case I'd suggest to name it as `customNodePaths` and document it as well.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -387,13 +388,19 @@ public void LoadCustomNodesAndPackages(IEnumerable<string> newPaths, IPreference
                 customNodeManager.AddUninitializedCustomNodesInPath(dir, false, false);
             }
 
-            foreach (var path in newPaths)
+            foreach (var path in packagePaths)
             {
-                var packageDirectory = pathManager.PackagesDirectories.FirstOrDefault(x => x.StartsWith(path));
-                if (packageDirectory != null)
+                var packageDirectory = path;
+                if (checkPathInPathManager)
```
aparajit-pratap, line 394: I see why you don't want this check in your case, but on the other hand, I think this additional flag looks confusing, and hard to understand what its purpose is.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -373,12 +373,13 @@ public void LoadPackages(IEnumerable<Package> packages)
...
         /// This method is called when custom nodes and packages need to be reloaded if there are new package paths.
         /// </summary>
-        /// <param name="newPaths"></param>
-        /// <param name="preferences"></param>
+        /// <param name="packagePaths"></param>
+        /// <param name="nodePaths"></param>
         /// <param name="customNodeManager"></param>
-        public void LoadCustomNodesAndPackages(IEnumerable<string> newPaths, IPreferences preferences, CustomNodeManager customNodeManager)
+        public void LoadCustomNodesAndPackages(IEnumerable<string> packagePaths, IEnumerable<string> nodePaths, 
+            CustomNodeManager customNodeManager, IEnumerable<string> uninstallPackageDir = null, bool checkPathInPathManager = true)
```
aparajit-pratap, line 380: please add xml documentation for `uninstallPackageDir ` and `checkPathInPathManager ` for API users to understand how to call this API

aparajit-pratap: Also, it should be `dirs` (plural) if it can be more than one directory.

aparajit-pratap (changes requested):
Requesting changes as per comments.

src/DynamoPackages/PackageLoader.cs
BogdanZavu: There was no enforcement whatsoever previously for being a "new" path and it was already used as any path.
Regarding the difference between packagePaths and nodePaths the difference is the same as before, nothing changed here, packages are loaded from a given location and custom nodes from another.
Regarding the naming, the utility is named LoadCustomNodesAndPackages which implies the custom nature of nodePaths parameter.

Anyway as an API user/developer I made this PR because the existing API is a little bit confusing and not working as one would expect. Also forcing an API developer to manipulate path manager to use this could have undesired effects if the changes to path manager are not reverted.

src/DynamoPackages/PackageLoader.cs
BogdanZavu: It is better to expose it and document it (as you noted) than perform associated code silently

src/DynamoPackages/PackageLoader.cs
aparajit-pratap: If your assertion is true that this API "doesn't" work, then it would be best to leave this unchanged as an internal method and you create a new one for your purposes.

aparajit-pratap: There is no enforcement of new paths but it is described in the code comments and also implied by the name of the argument.

aparajit-pratap (changes requested):
I think it would be best to write a new public-facing API, even though it probably won't be useful for Dynamo internal purposes. Manipulating the existing one as done in this PR and making it suitable for internal use as well as making it useful for third parties does not seem to make much sense.

BogdanZavu:
Closing this PR since new PR to restore the old API has been created : 
https://github.com/DynamoDS/Dynamo/pull/12444/files

PR #12358, opened 2021-12-01
update conflicting loaded packages  text

mjkkirschner:
### Purpose

https://jira.autodesk.com/browse/DYN-4387

* moves cancel button to far right (following standard messagebox pattern)
* adds new overload to `IMessageBox`, `DefaultMessageBox` and `DynamoMessageBox` to allow overriding of button names - the number of names must match the number of buttons, and they are applied in order. IE the order of the names in the `MessageBoxButton` parameter is used to determine which name applies to which button.
* names and text are overridden for this particular message according to UX/Content design.
* modifies `DynamoMessageBox` so that it sets a `MessageBoxResult` when any button is clicked, so we don't need to try converting from a nullable bool (`bool?`) with 3 states, to a MessageBoxResult with 5 states (none,yes,no,cancel,ok)... which was not sufficient to cover all buttons.
* handle the case where a the messageBoxResult could be MessageBoxResult.None (closing the dialog without clicking a button) - this usually results in the same result as cancel, but now we don't proceed as if `OK` was returned...

### TODO:
- [x] add more unit tests for DynamoMessageBox
- [x] audit use of MessageBoxService to check that we are dealing correctly with the return results now that `cancel`, `no` and `closing` the dialog do convert based on the dialogButtons present in the particular dialog, but instead are set directly to their messageboxResult type. 
- [ ] https://jira.autodesk.com/browse/DYN-4404 ... probably let morpheus solve this holistically later.

Screen Shot 2021-12-01 at 5 27 05 PM

### Declarations

### Release Notes

update text of conflicting loaded packages message, and fix message box button order, and fixed issue with custom dynamo message box buttons not returning correct results.


mjkkirschner:
@aparajit-pratap - unfortunately, while testing this I discovered that because of our wrapped implementation of a dialogResult in a message box - the `cancel` button result is converted to false, and so does the same exact thing as the `No` or `Uninstall loaded packages` button...

So - we can either get rid of the cancel button, or I can look into how to get the working. I'll take a look at it, but I don't want to break any other dialogs at this point.

mjkkirschner:
FYi @OliverEGreen would appreciate your feedback on my changes to DynamoMessageBox.

src/DynamoCoreWpf/UI/Prompts/DynamoMessageBox.xaml
```diff
@@ -110,8 +110,21 @@
                            Grid.Row="2"
                            Margin="0,0,20,20"
                            LastChildFill="False">
+                    
+                    <Button Name="CancelButton"
+                            MinWidth="90"
```
QilongTang, line 115: We had an issue somewhere saying German translation of cancel would take a lot of space making cancel button looks like cut. I suspect you may run into same issue here, our team has no solution yet. I think the task is https://jira.autodesk.com/browse/DYN-4404

src/DynamoCoreWpf/UI/Prompts/DynamoMessageBox.xaml.cs
```diff
@@ -23,6 +26,15 @@ public partial class DynamoMessageBox : INotifyPropertyChanged
 
         #region Public Properties
         
+        [Obsolete("Do not use, instead use CustomDialogResult")]
+        public new bool? DialogResult
```
QilongTang, line 30: You can just delete, this class is introduced in 2.13.0

src/DynamoCoreWpf/UI/Prompts/DynamoMessageBox.xaml
mjkkirschner: I'll take a look, but I've made it so the buttons now grow in width, see the screenshot above.

src/DynamoCoreWpf/UI/Prompts/DynamoMessageBox.xaml.cs
mjkkirschner: thanks @QilongTang I'm actually trying to wrap the base class (window) public property to alert clients and devs not to use this property, we can't remove the underlying property, but I want to note that it should not be used.

Are you okay with me no longer setting it at all, that will make my test pass.

QilongTang: Sure, this is our own new UI class, we can do whatever we wish is clear

src/DynamoCoreWpf/UI/Prompts/DynamoMessageBox.xaml
QilongTang: @mjkkirschner Neat, then we may come back here when tackling the other one

src/DynamoCoreWpf/UI/Prompts/DynamoMessageBox.xaml.cs
mjkkirschner: okay done, DialogResult is no longer set, but I kept the `new` method around to hide the base class version, and added some notes to the class summary about how to use it.

mjkkirschner:
https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/905/

mjkkirschner:
serial tests above passed, still waiting on master-5

src/DynamoCoreWpf/UI/Prompts/DynamoMessageBox.xaml.cs
```diff
@@ -159,7 +175,24 @@ public static MessageBoxResult Show(string messageBoxText, string caption)
                 BodyText = messageBoxText
             };
             dynamoMessageBox.ConfigureButtons(MessageBoxButton.OK);
-            return dynamoMessageBox.ConvertResult(dynamoMessageBox.ShowDialog());
+            dynamoMessageBox.ShowDialog();
+            return dynamoMessageBox.CustomDialogResult;
+        }
+
+        internal static MessageBoxResult Show(string messageBoxText, string caption, MessageBoxButton button, IEnumerable<string> buttonNames,
```
pinzart90, line 182: how are the `buttonNames` assigned to the window buttons ?
Is there an order defined somewhere ?
Maybe it should be a list<tuple<button_enum, name>> or dictionary<button_enum, name>

src/DynamoCoreWpf/Properties/Resources.en-US.resx
```diff
@@ -1378,11 +1378,13 @@ Do you wish to continue installing it while keeping the packages that are alread
 If not, {2} needs to uninstall {1} to continue but cannot as they are in use. Try restarting {2} and download {0} again.</value>
   </data>
   <data name="MessageForceInstallOrUninstallUponRestart" xml:space="preserve">
-    <value>Package {0} has one or more dependencies that conflict with the following packages: {1}. Dependency conflicts could cause unintended behavior to occur.
-    
-Do you wish to continue installing it while keeping the packages that are already installed?
+    <value>Package {0} has one or more dependencies that conflict with the following package(s):
+
+{1}
```
aparajit-pratap, line 1383: Does this formatting handle the case where there is more than one package listed one below the other?

src/DynamoCoreWpf/UI/Prompts/DynamoMessageBox.xaml.cs
mjkkirschner: as described in the summary, they order just matches the order defined in the ENUM - so OKCancel , "1" ,"2" -> OK:1 Cancel:2.

src/DynamoCoreWpf/Properties/Resources.en-US.resx
mjkkirschner: it does not, I will fix that by adding an option to the `JoinPackageNames` method to join with newlines.

src/DynamoCoreWpf/UI/Prompts/DynamoMessageBox.xaml.cs
mjkkirschner: Is that acceptable or do you prefer the more explicit parameter?

src/DynamoCoreWpf/UI/Prompts/DynamoMessageBox.xaml.cs
```diff
@@ -168,56 +201,51 @@ public static MessageBoxResult Show(string messageBoxText, string caption)
         /// Shows or hides buttons on the view as necessary.
         /// </summary>
         /// <param name="messageBoxButton"></param>
-        private void ConfigureButtons(MessageBoxButton messageBoxButton)
+        /// /// <param name="buttonNames">names that will be used to override the standard button names.
```
aparajit-pratap, line 204: double code comments chars

src/DynamoCoreWpf/UI/Prompts/DynamoMessageBox.xaml.cs
```diff
@@ -168,56 +201,51 @@ public static MessageBoxResult Show(string messageBoxText, string caption)
...
+                        OkButton.Content = buttonNames.ElementAt(0);
+                    }
                     OkButton.Visibility = Visibility.Visible;
                     break;
                 case MessageBoxButton.OKCancel:
+                    if (buttonNames != null && buttonNames.Count() == 2)
+                    {
+                        OkButton.Content = buttonNames.ElementAt(0);
+                        CancelButton.Content = buttonNames.ElementAt(1);
+                    }
```
aparajit-pratap, line 222: What if buttonNames count is 1 here, meaning if there is only one custom name, and the other is the default name?

src/DynamoCoreWpf/UI/Prompts/DynamoMessageBox.xaml.cs
```diff
@@ -168,56 +201,51 @@ public static MessageBoxResult Show(string messageBoxText, string caption)
...
                     OkButton.Visibility = Visibility.Visible;
                     CancelButton.Visibility = Visibility.Visible;
                     break;
                 case MessageBoxButton.YesNoCancel:
+                    if (buttonNames != null && buttonNames.Count() == 3)
+                    {
+                        YesButton.Content = buttonNames.ElementAt(0);
+                        NoButton.Content = buttonNames.ElementAt(1);
+                        CancelButton.Content = buttonNames.ElementAt(2);
+                    }
```
aparajit-pratap, line 232: same question here

src/DynamoCoreWpf/UI/Prompts/DynamoMessageBox.xaml.cs
aparajit-pratap: Should we also validate that the number of custom names is equal to the number of buttons and if not, just use the standard names?

mjkkirschner: > Should we also validate that the number of custom names is equal to the number of buttons and if not, just use the standard names?


That is what this does currently, at least thats what I thought I wrote 😉  - if the `buttonNames` count does not equal the expected number of buttons, the content is not modified - (it's set to the original names in the xaml)

src/DynamoCoreWpf/UI/Prompts/DynamoMessageBox.xaml.cs
mjkkirschner: checkout the xaml for `DynamoMessageBox` - each button has standard content like
` Content="{x:Static p:Resources.GenericTaskDialogOptionCancel}"`

src/DynamoCoreWpf/UI/Prompts/DynamoMessageBox.xaml.cs
aparajit-pratap: Ah, sorry, I missed that.

aparajit-pratap (approved):

src/DynamoCoreWpf/UI/Prompts/DynamoMessageBox.xaml.cs
```diff
@@ -23,6 +27,15 @@ public partial class DynamoMessageBox : INotifyPropertyChanged
...
         #region Public Properties
         
+        [Obsolete("Do not use, instead use CustomDialogResult. This property will not be set during ShowDialog()")]
+        public new bool? DialogResult
+        {
+            get => base.DialogResult;
+            set => base.DialogResult = value;
+        }
+
+        internal MessageBoxResult CustomDialogResult { get; set; } = MessageBoxResult.None;
```
pinzart90, line 37: is None equivalent to Cancel?
Do we handle None differently anywhere  ?
If not ...can we just use Cancel instead...might be simpler

src/DynamoCoreWpf/UI/Prompts/DynamoMessageBox.xaml.cs
pinzart90: must of missed the comment, just goes to  show that it is not idiot proof :)
I think it is acceptable for this API (probably not a lot of traffiic)

pinzart90 (approved):
1 comment then LGTM

src/DynamoCoreWpf/UI/Prompts/DynamoMessageBox.xaml.cs
mjkkirschner: `None` occurs when the close button is clicked, Cancel only occurs when the `Cancel` button is clicked. In this PR I have modified our code to usually check for both of them as the same, but that may not always be the intended behavior.

mjkkirschner:
tests passed here, merging this and will send cherry pick.

PR #12380, opened 2021-12-03
Turn Microsoft.Xaml.Behaviors into Dynamo.Xaml.Behaviors!

mjkkirschner:
### Purpose
* In an effort to stop constant conflicts with system.windows.interactivity in the Revit context, I've used `mono.cecil` wrapped in a small CLI tool - to modify all types and assembly name to be `Dynamo.Microsoft.Xaml.Behaviors`. Then removed system.windows.interactivity, and modified the xaml to point to the new namespace and assembly, the code remains unchanged otherwise.
* The CLI tool using mono.cecil is hardcoded to run on a local copy of Microsoft.Xaml.Behaviors.dll - so to generate a new copy a dev needs to copy the dll from nuget, into the local bin directory of this tool.
* The output dll is saved to Dynamo/extern.
* The tool also strips off the strong name of the assembly, as I didn't want our modified version to somehow have the same identity as the original binary.

* Manual testing works, creating nodes, pinning connectors, list@level, snapping etc (various places system.windows.interactivity types were used)
* an empty file (0 bytes) named `system.windows.interactivity` is added to the `legacy_remove_me` folder and copied to bin to satisfy patch installer requirements of not removing files...

Screen Shot 2021-12-03 at 4 34 17 PM
Screen Shot 2021-12-03 at 4 34 41 PM
Screen Shot 2021-12-03 at 4 34 31 PM

### Declarations

### Release Notes

Remove system.windows.interactivity dependency to avoid conflicts with Revit and other addins.


pinzart90:
I am curious if using a strong name assembly for the interactivity dll would solve any potential conflicts.

mjkkirschner:
@pinzart90 thats a great point, in theory yes... - when I'm at my work machine I will verify if the existing `system.windows.interactivity` we ship is strong named or not...

1. if it is not strong named, then let's try just switching to xaml behaviors with a strong name and see what happens. It's unfortunate I don't have an easily reproducible case to test with though I can try to find one from the reported cases.
2. if it already is strong named, then it's obviously not helping and lets embed the types.

If 1. fails to help we can always move to embedding in a future release.

It's also worth pointing out that if xaml behaviors strong names stay constant for multiple versions (like json.net) then this does not help us at all.
https://stackoverflow.com/questions/31608983/newtonsoft-json-net-version-incompatibility-dll-hell

mjkkirschner:
@pinzart90 I've verified that the System.Windows.Interactivity 4.5.0 already had a strong name :(

also worth noting from msdn:
```
For .NET Core and .NET 5+, strong-named assemblies do not provide material benefits. The runtime never validates the strong-name signature, nor does it use the strong-name for assembly binding.
```

pinzart90:
> @pinzart90 I've verified that the System.Windows.Interactivity 4.5.0 already had a strong name so not sure why we were still running into compatibility problems.
> 
> also worth noting from msdn:
> 
> ```
> For .NET Core and .NET 5+, strong-named assemblies do not provide material benefits. The runtime never validates the strong-name signature, nor does it use the strong-name for assembly binding.
> ```

@mjkkirschner Thanks for looking into that. That is definitely surprising.
One other suggestion.,..Can you turn off the new Projects build by default (It does not seem beneficial to build it every time with the main solution)

mjkkirschner:
@pinzart90, yes will look into that, looking at the test failures first.

pinzart90 (approved):
LGTM

QilongTang (approved):

QilongTang:
The regression `DynamoPythonTests.PythonTestsWithLogging.DynamoPrintLogsToConsole` does not seem real to me

src/Tools/AssemblyRenamerCLI/Program.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Threading.Tasks;
+
+namespace AssemblyRenamerCLI
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            var executingPath = Assembly.GetExecutingAssembly().Location;
+            var xamlBehaviorsPath = Path.Combine(Path.GetDirectoryName(executingPath), "Microsoft.Xaml.Behaviors.dll");
```
aparajit-pratap, line 17: Can we make this tool generic by supplying dll and namespace names as command line args?

aparajit-pratap (approved):

src/Tools/AssemblyRenamerCLI/Program.cs
mjkkirschner: sure, that is definitely possible, will look into it.

sm6srw (approved):

mjkkirschner:
@aparajit-pratap please take another look, i've added the CLI flags and a readme as well as handling another case that I missed.

mjkkirschner:
gah, because this will not make it into 2.13 - I think we'll need to keep shipping system.windows.interactivity for another year. I will add it to the removal list for tracking and keep it in this PR, but no longer reference it.

QilongTang:
> gah, because this will not make it into 2.13 - I think we'll need to keep shipping system.windows.interactivity for another year. I will add it to the removal list for tracking and keep it in this PR, but no longer reference it.

What's keeping it from getting released in 2.13?

mjkkirschner:
@QilongTang - the team decided the risk was too large vs the potential benefit because I could not prove that the xaml parse stack trace was caused by a system.windows.interactivity conflict - so we decided to merge to master and test there, and hopefully find a repro case.

mjkkirschner:
tests running https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/5135/

PR #12396, opened 2021-12-06
 use aggregated exception type to log aggregated xaml parse exceptions

mjkkirschner:
Please Note:

### Purpose

create an aggregate exception to send to analytics - this exception type is now handled by the TrackException API.

### Declarations

### Release Notes

log aggregated exceptions to analytics when inoutport panel fails to calculate size during layout.


src/DynamoCoreWpf/UI/InOutPortPanel.cs
```diff
@@ -48,8 +52,23 @@ protected override Size MeasureOverride(Size constraint)
...
+                        //if the inner exeption is not null, gather its inner stack trace
+                        //and create a new exception 
+                    {   if(e.InnerException != null)
+                        {
+                            var aggregatedException = new DynamoUtilities.ExceptionHelpers.DynamoWrappedException
+                                (e,
+                                $"{e.StackTrace} {System.Environment.NewLine} {e.InnerException.StackTrace}" );
+
+                             Analytics.TrackException(aggregatedException,false);
+                            continue;
```
aparajit-pratap, line 67: What is the `continue` for? Did you intend the outer `if` to be a `while(e != null)`? Can't we simply aggregate all innerexceptions into a single exception and track it just once in analytics?

src/DynamoUtilities/ExceptionHelpers.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+
+namespace DynamoUtilities.ExceptionHelpers
+{
+
+    /// <summary>
+    /// Exception class thats useful for sending aggregated exceptions and stacktraces to analytics.
+    /// </summary>
+    [Obsolete("For internal use only")]
```
aparajit-pratap, line 9: Why not just make this internal?

src/DynamoCoreWpf/UI/InOutPortPanel.cs
mjkkirschner: I left the continue from @QilongTang 's original change, but I like your idea better - can try that.

mjkkirschner: whoops, but I forgot to move it 🤦 - see line 72.

src/DynamoUtilities/ExceptionHelpers.cs
mjkkirschner: Because this type needs to be passed to the Analytics library where it will need to call these methods.

src/DynamoCoreWpf/UI/InOutPortPanel.cs
```diff
@@ -48,8 +54,14 @@ protected override Size MeasureOverride(Size constraint)
...
                 }
-                catch(System.Exception)
+                catch(XamlParseException e)
                 {
+                    if(e != null)
+                     //aggregate the stack traces as we'll loop over all children in this panel.
+                    {
+                        aggretatedExceptionStack += $"------------------ {System.Environment.NewLine } { e.StackTrace} {System.Environment.NewLine}" +
+                        e.InnerException != null ? $" inner: {e.InnerException.StackTrace} {System.Environment.NewLine}" : string.Empty;
+                    }
```
pinzart90, line 64: what about 
```
List<XamlParseException> exList = null;

try
{
    // Default behavior of getting each child to measure.
    child.Measure(constraint);
}
catch(XamlParseException e)
{
    exList.Add(e);
    continue;
}

//log to analytics if we recorded any exceptions
if(exList.Count > 0)
{
    Analytics.TrackException(new AggregateException("MeasureOverride XML exception", exList), false);
}
```
Then, In analytics.Core, you can simply handle the case when the Exception is of type aggregate
```
        public ExceptionEvent(Exception ex, bool isFatal = true)
        {
            exception = ex;
            Description = ex.GetType().ToString() + Environment.NewLine;
            if (!string.IsNullOrEmpty(ex.StackTrace))
                Description += ex.StackTrace.ToString().Replace("&", "");

            if (exception is AggregateException aggrEx)
            {
                foreach (var iex in aggrEx.Flatten().InnerExceptions) {
                    Description += "\n Internal exception \n" + iex.StackTrace.ToString().Replace("&", "");
                }
            }
            
            IsFatal = isFatal;
        }
 ```

Comment (pinzart90) [reply src/DynamoCoreWpf/UI/InOutPortPanel.cs:64]: Unless you are trying to avoid modifying the Analytics assembly

File: src/DynamoUtilities/ExceptionHelpers.cs (hunk 2 above)
Comment (aparajit-pratap) [reply src/DynamoUtilities/ExceptionHelpers.cs:9]: internalsvisibleto?

File: src/DynamoCoreWpf/UI/InOutPortPanel.cs (hunk 3 above)
Comment (mjkkirschner) [reply src/DynamoCoreWpf/UI/InOutPortPanel.cs:64]: well yes, I was trying to avoid altering the analytics dll - but more importantly it doesn't really handle @aparajit-pratap 's use case here:
https://github.com/DynamoDS/Dynamo/pull/12395

I guess he could make an AggregateException where the inner exception is just another exception of the new aggregate... but feels like a bunch of incantations just to get a string in the end.

I'm okay with it if y'all feel it's more idiomatic and makes the API easier to understand?

Comment (mjkkirschner) [reply src/DynamoCoreWpf/UI/InOutPortPanel.cs:64]: It does nicely avoid a new exception type though.

Comment (pinzart90) [reply src/DynamoCoreWpf/UI/InOutPortPanel.cs:64]: I think using AggregateException would be simpler
And I think we should enhance the existing Analtics.Core TrackException API to handle aggregates, to use the inner exceptions (if they exist) and to use the Exception message (not just the stackTrace)

File: src/DynamoCoreWpf/UI/InOutPortPanel.cs [hunk 4]
```diff
@@ -37,6 +41,8 @@ protected override Size ArrangeOverride(Size arrangeSize)
 
         protected override Size MeasureOverride(Size constraint)
         {
+            var aggretatedExceptions = new List<Exception>();
```
aparajit-pratap, line 44: typo

mjkkirschner: ty

src/DynamoCoreWpf/UI/InOutPortPanel.cs
```diff
@@ -48,8 +54,13 @@ protected override Size MeasureOverride(Size constraint)
                     // Default behavior of getting each child to measure.
                     child.Measure(constraint);
                 }
-                catch(System.Exception)
+                catch(XamlParseException e)
```
pinzart90, line 57: so we will not catch any other Exception types correct ?

src/DynamoCoreWpf/UI/InOutPortPanel.cs
```diff
@@ -48,8 +54,13 @@ protected override Size MeasureOverride(Size constraint)
...
                 {
+                    if(e != null)
```
aparajit-pratap, line 59: Is this null check really required, won't `e` always be non-null?

src/DynamoCoreWpf/UI/InOutPortPanel.cs
mjkkirschner: correct, those other exceptions are not expected and will crash.

src/DynamoCoreWpf/UI/InOutPortPanel.cs
mjkkirschner: https://stackoverflow.com/questions/5634417/caught-exception-is-null-itself/27902976

mjkkirschner: yeah, I thought I did this for a good reason, but it was a hold over from checking inner exceptions, I will remove unless theres something I'm missing.

pinzart90: I do not think it is useful here

aparajit-pratap: It looks like it can be `null `if there are multiple `catch `blocks using the same variable name from that StackOverflow page.

pinzart90: You can check for null in the Analytics.Core APIs

aparajit-pratap (approved):

pinzart90 (approved):

mjkkirschner:
running tests: https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/5085/

mjkkirschner:
tests passed, merging this when free, and will cherry pick next.

PR #12400, opened 2021-12-07
Adjust the back and next button on guided tour popup and add keyboard…

reddyashish:
### Purpose

This PR is to adjust the click area for back and next buttons on guided tour popup and user can navigate the steps using the keyboard arrow keys. The one thing that needs to be done is that the popup window should be focused to react to keyboard buttons. I am looking into some options for this. 

Task: https://jira.autodesk.com/browse/DYN-3972

### Declarations

Check these if you believe they are true

- [ ] The codebase is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.
- [ ] This PR modifies some build requirements and the readme is updated

### Release Notes

Adjust the back and next button on guided tour popup window.

### Reviewers

@QilongTang @filipeotero @RobertGlobant20

PR #12416, opened 2021-12-08
Set the focus on the guided-tour popup window when the current step is loaded.

reddyashish:
### Purpose

This is just to set the focus on the popup window element so that keyboard events can be handled smoothly. This was missed as a part of this PR: https://github.com/DynamoDS/Dynamo/pull/12400. So creating a new one and will cherry-pick it too.

### Declarations

### Release Notes

Improvement to https://github.com/DynamoDS/Dynamo/pull/12400

### Reviewers
@QilongTang @RobertGlobant20

### DIFF
--- a/src/DynamoCoreWpf/UI/GuidedTour/Step.cs
+++ b/src/DynamoCoreWpf/UI/GuidedTour/Step.cs
@@ -26,6 +26,7 @@ namespace Dynamo.Wpf.UI.GuidedTour
     {
         #region Private Fields
         private static string WindowNamePopup = "PopupWindow";
+        private static readonly string NextButton = "NextButton";
         private static string calculateLibraryFuncName = "CalculateLibraryItemLocation";
         private static string libraryScrollToBottomFuncName = "LibraryScrollToBottom";
         #endregion
@@ -236,7 +237,7 @@ namespace Dynamo.Wpf.UI.GuidedTour
 
             stepUIPopup.IsOpen = true;
 
-            var nextButton = stepUIPopup.FindName("NextButton") as Button;
+            var nextButton = stepUIPopup.FindName(NextButton) as Button;
             if (nextButton != null) 
             {
                 nextButton.Focus();
--- a/src/DynamoCoreWpf/UI/GuidedTour/Step.cs
+++ b/src/DynamoCoreWpf/UI/GuidedTour/Step.cs
@@ -235,6 +235,12 @@ namespace Dynamo.Wpf.UI.GuidedTour
                
 
             stepUIPopup.IsOpen = true;
+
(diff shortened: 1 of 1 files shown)


QilongTang:
When you say keyboard, you mean `enter` key?

reddyashish:
Right and left arrows on keyboard to navigate to prev and next steps.

src/DynamoCoreWpf/UI/GuidedTour/Step.cs
```diff
@@ -235,6 +235,12 @@ where automation.Name.Equals(calculateLibraryFuncName)
                
 
             stepUIPopup.IsOpen = true;
+
+            var nextButton = stepUIPopup.FindName("NextButton") as Button;
```
QilongTang, line 239: anyway we can move this string somewhere?

reddyashish: Yes, moved it to a readonly string.

src/DynamoCoreWpf/UI/GuidedTour/Step.cs
```diff
@@ -235,6 +236,12 @@ where automation.Name.Equals(calculateLibraryFuncName)
                
 
             stepUIPopup.IsOpen = true;
+
+            var nextButton = stepUIPopup.FindName(NextButton) as Button;
+            if (nextButton != null) 
+            {
+                nextButton.Focus();
```
QilongTang, line 243: So I assume once the button gets focus, we can use arrow keys for these?

reddyashish: yeah, that's correct.

src/DynamoCoreWpf/UI/GuidedTour/Step.cs
```diff
@@ -235,6 +236,12 @@ where automation.Name.Equals(calculateLibraryFuncName)
                
 
             stepUIPopup.IsOpen = true;
+
+            var nextButton = stepUIPopup.FindName(NextButton) as Button;
```
QilongTang, line 240: I found this code which doing similar, https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/UI/GuidedTour/GuidesValidationMethods.cs#L368, can you take a look?

reddyashish: changed it to use the FindChild API on Guide class.

QilongTang: Hi @reddyashish Seems we need a `null` similar to the code example above

PR #12482, opened 2021-12-10
dynamic python engines part 2

pinzart90:
Co-Authored-By: pinzart <tiberiu.pinzariu@autodesk.com>

### Purpose

Switch to fully dynamic python engines
Dynamo will try to load python engines from any assembly loaded by PackageLoader.
To add a new Python engine you need to implement the PythonEngine base class and add a static property named `Instance` that returns an instance of the PythonEngine class. The idea behind requiring a static `Instance` property was to enforce some sort of Singleton behavior in derived class code. 

### Declarations

### Release Notes

### Reviewers

### FYIs


src/Libraries/DSIronPython/IronPythonEvaluator.cs
```diff
@@ -54,7 +55,7 @@ public class IronPythonEvaluator : Dynamo.PythonServices.PythonEngine
         /// </summary>
         public const string packageExtraFolderName = @"extra";
 
-        public override string Name => PythonNodeModels.PythonEngineVersion.IronPython2.ToString();
+        public override string Name => PythonEngineManager.IronPython2EngineName;
```
mjkkirschner, line 58: @pinzart90 do you need to make changes to the published ironPython package as well?

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
```diff
@@ -33,12 +33,25 @@ public partial class ScriptEditorWindow : ModelessChildWindow
         public PythonNode nodeModel { get; set; }
         private bool nodeWasModified = false;
         private string originalScript;
-        public PythonEngineVersion CachedEngine { get; set; }
+
+        public string CachedEngineName { get; set; }
```
mjkkirschner, line 37: because this is a view - I would say let's just break these public signatures.

src/Libraries/DSIronPython/IronPythonEvaluator.cs
pinzart90: yes. after this PR

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
pinzart90: well...that would simplify some of the changes. Will do

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -33,20 +33,45 @@ public PythonCodeMigrationEventArgs(string oldCode, string newCode)
 
     public abstract class PythonNodeBase : VariableInputNode
     {
-        private PythonEngineVersion engine = PythonEngineVersion.Unspecified;
+        private string engine = string.Empty;
 
         [JsonConverter(typeof(StringEnumConverter))]
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.Populate)]
-        [DefaultValue(nameof(PythonEngineVersion.IronPython2))]
+        [DefaultValue("IronPython2")]
```
mjkkirschner, line 40: this is because ironPython project is not built?

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -57,7 +82,7 @@ public PythonEngineVersion Engine
                 if (engine != value)
                 {
                     engine = value;
-                    RaisePropertyChanged(nameof(Engine));
+                    RaisePropertyChanged(nameof(EngineName));
```
mjkkirschner, line 85: is a public setter useful?

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
```diff
@@ -33,12 +33,22 @@ public partial class ScriptEditorWindow : ModelessChildWindow
...
-        public PythonEngineVersion CachedEngine { get; set; }
+
+        public string CachedEngine { get; set; }
 
         /// <summary>
         /// Available Python engines.
         /// </summary>
-        public ObservableCollection<PythonEngineVersion> AvailableEngines
+        [Obsolete()]
+        public ObservableCollection<PythonEngineVersion> AvailableEnginesd
```
mjkkirschner, line 43: spelling (d?)

test/core/python/pythonFromString.dyn
```diff
@@ -11,7 +11,6 @@
   "Nodes": [
     {
       "ConcreteType": "PythonNodeModels.PythonStringNode, PythonNodeModels",
-      "Engine": "CPython3",
```
mjkkirschner, line 14: is this a modified .dyn or do we no longer serialize the engine for some reason?

mjkkirschner (commented):
@pinzart90 it looks good to me - but I had one question about serialization I think we should answer before merging - also why did the sonar scan fail?

test/core/python/pythonFromString.dyn
pinzart90: I removed the serialized engine so that the corresponding test can cover the case of the Default serialized value (more coverage in the test)

src/Libraries/PythonNodeModelsWpf/ScriptEditorWindow.xaml.cs
pinzart90: Oh..missed that. I changed the name so I would not miss any references to it while refactoring.

test/core/python/pythonFromString.dyn
mjkkirschner: if you modified it by hand maybe rename the file so that anyone in the future resaving this will aware something is hand modified.

src/Libraries/PythonNodeModels/PythonNode.cs
pinzart90: Because I deprecated PythonEngineVersion and I did not want to use it anymore
For some reason ..using the PackageEngineManager.IronPythonEngineName does not work...compiler complains that it is not a const expression (even though it is a static readonly string...)

test/core/python/pythonFromString.dyn
pinzart90: can do

src/Libraries/PythonNodeModels/PythonNode.cs
pinzart90: I can revert this change...The parent property is obsolete anyway

src/NodeServices/PythonServices.cs
```diff
@@ -164,73 +171,39 @@ private PythonEngineManager()
...
-        {
-            return GetEngine(version.ToString());
+            AppDomain.CurrentDomain.AssemblyLoad += new AssemblyLoadEventHandler((object sender, AssemblyLoadEventArgs args) =>
+            {
+                if (args.LoadedAssembly != null && 
+                    args.LoadedAssembly.GetName().Name == CPythonAssemblyName)
+                {
+                    Instance.LoadPythonEngine(args.LoadedAssembly);
+                }
+            });
```
aparajit-pratap, line 184: Why are you only checking for CPython? What if other assemblies are being loaded that contain other python evaluators?

src/DynamoPackages/PackageLoader.cs
```diff
@@ -287,7 +287,10 @@ private void TryLoadPackageIntoLibrary(Package package)
 
                 package.SetAsLoaded();
                 PackgeLoaded?.Invoke(package);
-                PackagesLoaded?.Invoke(loadedAssemblies);
+                PackagesLoaded?.Invoke(loadedNodeLibs);
+
+                PythonServices.PythonEngineManager.Instance.
+                    LoadPythonEngine(package.LoadedAssemblies.Select(x => x.Assembly));
```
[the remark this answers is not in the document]
aparajit-pratap: Are you relying on the fact that the python assembly will be passed only once?

src/Libraries/PythonNodeModels/PythonNode.cs
pinzart90: We decided to leave IronPython as the DefaultSerialized value...so that older graphs (that did not have an explicit engine serialized) can show the migration warnings

src/DynamoPackages/PackageLoader.cs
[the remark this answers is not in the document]
pinzart90: The idea here is to be able to load PythonEngines from any assembly without performance penalties (like from listening to all assembly loaded events)
I want to check all assemblies from a package. We could also check on extension loading ...not sure which would be best.
Should python engines be loaded only from packages or only from extensions ?

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -136,17 +161,14 @@ protected override string GetInputTooltip(int index)
...
-            // Here we switched to use the AstFactory.BuildFunctionCall version that accept
-            // class name and function name. They will be set by PythonEngineManager by the engine value. 
-            PythonEngineManager.Instance.GetEvaluatorInfo(Engine, out string evaluatorClass, out string evaluationMethod);
-
             return AstFactory.BuildAssignment(
                 GetAstIdentifierForOutputIndex(0),
                 AstFactory.BuildFunctionCall(
-                    evaluatorClass,
-                    evaluationMethod,
+                    "PythonEvaluator",
```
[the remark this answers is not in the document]
pinzart90: We do not have access to the assembly that defines that property (DSCoreNodes)

src/NodeServices/PythonServices.cs
pinzart90: We only check for the Default python engine because that is the only one loaded through static references.
If we were to change it so that CPythonEnigne would be a Built in package or an extension...we could get rid of this

src/Libraries/PythonNodeModelsWpf/PythonNode.cs
```diff
@@ -199,38 +178,25 @@ private void PythonEnginesChanged(object sender, NotifyCollectionChangedEventArg
...
+                }   
             }
         }
 
         /// <summary>
-        /// Adds IronPython2 MenuItem
+        /// Adds python engine to MenuItems
         /// </summary>
-        private void AddIronPython2MenuItem()
+        private void AddPythonEngine2MenuItems(string engineName)
```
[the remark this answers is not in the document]
pinzart90: Yeah...I can change that

src/Libraries/PythonNodeModels/PythonNode.cs
pinzart90: It is bound

pinzart90:
Looks like the sonar failure was a fluke. Running again to confirm

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -33,20 +34,50 @@ public PythonCodeMigrationEventArgs(string oldCode, string newCode)
...
+                return engineVersion;
+            }
+            set
+            {
+                engine = value.ToString();
+                RaisePropertyChanged(nameof(EngineName));
+            }
+        }
+
+        [XmlIgnore]
```
mjkkirschner, line 69: this is needed because we serialize to preferences or...?

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -71,7 +102,8 @@ public static ObservableCollection<PythonEngineVersion> AvailableEngines
         {
             get
             {
-                return new ObservableCollection<PythonEngineVersion>(PythonEngineManager.Instance.AvailableEngines.Select(x=> x.Version));
+                return new ObservableCollection<PythonEngineVersion>(PythonEngineManager.Instance.AvailableEngines.
```
mjkkirschner, line 105: this weirdness again, returning a new observable, I think we may have talked about it in another PR, I know this is not new code, does it matter here?

test/Libraries/DynamoPythonTests/PythonTestsWithLogging.cs
```diff
@@ -18,6 +19,7 @@ public class PythonTestsWithLogging : DynamoModelTestBase
         protected override void GetLibrariesToPreload(List<string> libraries)
         {
             libraries.Add("DSCPython.dll");
+            libraries.Add("DSCoreNodes.dll");
```
mjkkirschner, line 22: why this change?

mjkkirschner (approved):
Thanks @pinzart90 - I think it would be beneficial if the PR description had just a sentence or two about how this works now as opposed to currently.

Otherwise just a few small questions.

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -33,20 +34,50 @@ public PythonCodeMigrationEventArgs(string oldCode, string newCode)
...
+        [Obsolete("This property will be deprecated in Dynamo 3.0. Please use EngineName instead")]
         public PythonEngineVersion Engine
+        {
+            get
+            {
+                if (!Enum.TryParse(engine, out PythonEngineVersion engineVersion) ||
+                    engineVersion == PythonEngineVersion.Unspecified)
+                {
+                    // This is a first-time case for newly created nodes only
+                    SetEngineByDefault();
```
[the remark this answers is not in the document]
pinzart90: Yes

src/Libraries/PythonNodeModels/PythonNode.cs
pinzart90: I figured it is acceptable...since the property is Obsolete and not very useful to access from here.

src/Libraries/PythonNodeModels/PythonNode.cs
pinzart90: The reason is so that we do not searialize 2 properties that represent the same model data.
When EngineName will be removed we can only use Engine for serialization.
See explanation from code at line 44:
```
// When removing this property also replace the serialized property in EngineName
// (i.e remove XmlIgnore and add [JsonProperty("Engine", DefaultValueHandling = DefaultValueHandling.Populate)]
        ```

test/Libraries/DynamoPythonTests/PythonTestsWithLogging.cs
pinzart90: needed because of the new class I added in DSCoreNodes. The new class needs to be loaded so that Evaluation can work

PR #12486, opened 2021-12-12
DYN-4461 Warning Bar color persists warning or error state after resolution if the node is set preview off

saintentropy:
_###_ Purpose

https://jira.autodesk.com/browse/DYN-4461

This PR addresses the issue were once a node enters a warning or error state, the color will never reset back to grey to show preview off state.  This is especially confusing in situations where a null value sets a swath of nodes with preview off to a warning state.  Once the error or warning is actually resolved, the nodes still appear to be in that bad state.  See the example below:

https://user-images.githubusercontent.com/2145751/145697148-62d459e4-f2c1-452f-9733-f034d42466f9.mp4

I have not been able to find any work around that resets the `WarningBarColor` back to grey once this condition has been reached other than closing and reopening the file.  Please let me know if you find another way.

This PR changes how and when we set the color so that the preview color is restored.  The PR changes the logic so that we set the color when the `State` property or `IsVisible` property changes on the node.  The PR also changes the logic of `GetWarningColor`.  It does not need outside data to set the color as existing node properties drive the color.  

~~Note: It looks like the`WarningBarColor` property should not actually have a setter.  The getter should just have the logic found in the `GetWarningColor` function as it is determinant.  I wasn't sure if I could remove the internal setter and keep binary compatibility as 2.13 has been cut.  (Thoughts @mjkkirschner?) If I can remove the setter with binary compatibility we should make that change so that we don't have to remember to set `WarningBarColor` for other situations in the future.~~

[x] Todo update the tests

### Declarations

### Release Notes

Bug fixed with Warning Bar color persisting as warning or error state after resolution if the node is set preview off

### Reviewers

@QilongTang @mjkkirschner  

### FYIs

@Amoursol @OliverEGreen

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
@@ -1015,7 +1015,6 @@ namespace Dynamo.ViewModels
                     break;
                 case "State":
                     RaisePropertyChanged("State");
-                    RaisePropertyChanged(nameof(NodeWarningBarVisible));
                     WarningBarColor = GetWarningColor();
                     break;
                 case "ArgumentLacing":
@@ -1027,7 +1026,6 @@ namespace Dynamo.ViewModels
                     break;
                 case "IsVisible":
                     RaisePropertyChanged("IsVisible");
-                    RaisePropertyChanged(nameof(NodeWarningBarVisible));
                     WarningBarColor = GetWarningColor();
                     break;
                 case "Width":
@@ -1067,6 +1065,21 @@ namespace Dynamo.ViewModels
         }
 
         /// <summary>
+        /// Respond to property changes on the error bubble.
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="e"></param>
+        void ErroBubble_PropertyChanged(object sender, PropertyChangedEventArgs e)
+        {
+            switch (e.PropertyName)
+            {
+                case nameof(ErrorBubble.DoesNodeDisplayMessages):
+                    WarningBarColor = GetWarningColor();
+                    break;
+            }
+        }
+
+        /// <summary>
         /// Updates the width of the node's Warning/Error bubbles, in case the width of the node changes.
         /// </summary>
         private void UpdateErrorBubbleWidth()
@@ -1087,6 +1100,7 @@ namespace Dynamo.ViewModels
             ErrorBubble.NodeInfoToDisplay.CollectionChanged += UpdateOverlays;
             ErrorBubble.NodeWarningsToDisplay.CollectionChanged += UpdateOverlays;
             ErrorBubble.NodeErrorsToDisplay.CollectionChanged += UpdateOverlays;
+            ErrorBubble.PropertyChanged += ErroBubble_PropertyChanged;
             
             if (DynamoViewModel.UIDispatcher != null)
             {
@@ -1129,6 +1143,12 @@ namespace Dynamo.ViewModels
 
             if (NodeModel.State == ElementState.Warning || NodeModel.State == ElementState.PersistentWarning)
             {
+                //Handle the case where the user has dismissed the warning and no warnings are showing.
+                if (ErrorBubble != null && !ErrorBubble.DoesNodeDisplayMessages)
+                {
+                    return noPreviewColor;
+                }
+
                 return warningColor;
             }
 
@@ -1154,6 +1174,7 @@ namespace Dynamo.ViewModels
             ErrorBubble.NodeInfoToDisplay.CollectionChanged -= UpdateOverlays;
             ErrorBubble.NodeWarningsToDisplay.CollectionChanged -= UpdateOverlays;
             ErrorBubble.NodeErrorsToDisplay.CollectionChanged -= UpdateOverlays;
+            ErrorBubble.PropertyChanged -= ErroBubble_PropertyChanged;
 
             ErrorBubble.Dispose();
         }
--- a/src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
@@ -1106,6 +1106,14 @@ namespace Dynamo.ViewModels
             ErrorBubble.DismissedMessages.CollectionChanged += DismissedNodeMessages_CollectionChanged;
         }
 
+        // These colors are duplicated from the DynamoColorsAndBrushesDictionary as it is not assumed that the xaml will be loaded before setting the color
+        // SharedDictionaryManager.DynamoColorsAndBrushesDictionary["NodeErrorColor"];
+        private static SolidColorBrush errorColor = (SolidColorBrush)(new BrushConverter().ConvertFrom("#EB5555"));
+        // SharedDictionaryManager.DynamoColorsAndBrushesDictionary["NodeWarningColor"];
+        private static SolidColorBrush warningColor = (SolidColorBrush)(new BrushConverter().ConvertFrom("#FAA21B"));
+        // SharedDictionaryManager.DynamoColorsAndBrushesDictionary["NodePreviewColor"];
+        private static SolidColorBrush noPreviewColor = (SolidColorBrush)(new BrushConverter().ConvertFrom("#BBBBBB"));
+
         /// <summary>
         /// Sets the color of the warning bar, which informs the user that the node is in
         /// either an error or a warning state.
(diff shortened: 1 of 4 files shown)


mjkkirschner:
If 2.13 has really been cut then we should not remove the setter - if we're okay changing what our final RC version actually is then we should remove the setter and fix the bugs here.

FYI @QilongTang @sm6srw @jasonstratton I think this question above is your call.

I don't understand this sentence:

```
I have not been able to find any work around that resets the WarningBarColor back to grey once this condition has been reached other than closing and reopening the file. Please let me know if you find another way.

```

does your PR fix this issue or not?

saintentropy:
@mjkkirschner I meant no work around in the current UI.  My PR is a fix for the issue.  We can leave the setter.  I will update the comments.

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
```diff
@@ -1111,30 +1112,21 @@ private void BuildErrorBubble()
...
-                    }
-                    break;
-                case InfoBubbleViewModel.Style.Warning:
-                case InfoBubbleViewModel.Style.WarningCondensed:
-                    return (SolidColorBrush)SharedDictionaryManager.DynamoColorsAndBrushesDictionary["NodeWarningColor"];
-                case InfoBubbleViewModel.Style.Error:
-                case InfoBubbleViewModel.Style.ErrorCondensed:
-                    return (SolidColorBrush)SharedDictionaryManager.DynamoColorsAndBrushesDictionary["NodeErrorColor"];
-                case InfoBubbleViewModel.Style.Info:
-                    return (SolidColorBrush)SharedDictionaryManager.DynamoColorsAndBrushesDictionary["NodeInfoColor"];
```
mjkkirschner, line 1131: @saintentropy - after your changes I do not see this color set anywhere.

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
```diff
@@ -1026,8 +1028,7 @@ void logic_PropertyChanged(object sender, PropertyChangedEventArgs e)
                 case "IsVisible":
                     RaisePropertyChanged("IsVisible");
                     RaisePropertyChanged(nameof(NodeWarningBarVisible));
-                    if (ErrorBubble != null) return;
```
mjkkirschner, line 1029: could this cause an issue where the preview off color will override the error color?

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
mjkkirschner: `NodeInfoColor`

mjkkirschner: @Amoursol confirms that this color was planned but the functionality for setting it is not exposed, so I think it's okay to remove it, we'll have to refactor this method or add an info state to the node later - or something like that.

QilongTang:
@mjkkirschner @saintentropy Which setter were you referring to? I believe after the discussion, team decided to include this release as part of 2.13

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
QilongTang: I have the same question, it was there on that purpose

saintentropy: The color was being set previously as a function of the ErrorBubble.  This doesn't actually matter.  The state of the error bubble only controls the visibility of the WarningBar.  The color of the warning bar is a function of the state of the node.  Error -> Red, Warning or Persistent Warning -> Yellow, Default -> Grey.  Even if this event trigger the update color, the logic pin the `GetWarningColor` will never let it be set to Grey.

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
saintentropy: Correct.  I took it out as it wasn't included yet.  We will do as you say when the state is added.

QilongTang: Agree on above, let's leave it out for now

QilongTang (commented):
These LGTM but I am still unsure about the setter discussion

saintentropy:
@Amoursol @mjkkirschner @QilongTang One thing that my PR does not change that I am not sure was intended is that if you have a node that has preview off that does into a warning state, dismissing the warning leaves the warning bar yellow.  
image
I did not change anything in this PR that affects that logic.  Not sure if that is intended UI

mjkkirschner:
@saintentropy in a few minutes of testing this seems pretty good - I guess if we keep the setter the info color is also accessible by someone using the setter manually - I don't know enough about the intended design / use cases to know if thats better or a new state is. All theoretical at this point I guess.

QilongTang:
> @Amoursol @mjkkirschner @QilongTang One thing that my PR does not change that I am not sure was intended is that if you have a node that has preview off that does into a warning state, dismissing the warning leaves the warning bar yellow. image I did not change anything in this PR that affects that logic. Not sure if that is intended UI

are you referring to https://jira.autodesk.com/browse/DYN-4368?

mjkkirschner (approved):
It looks ok to me!

mjkkirschner:
> 
@QilongTang I think thats a different issue.

mjkkirschner:
@Amoursol @QilongTang this seems like another bug, but I think a smaller one than this PR addresses.

QilongTang:
> @Amoursol @QilongTang this seems like another bug, but I think a smaller one than this PR addresses.

I agree

saintentropy:
Also easy to fix...  Just need to know the intended UI.   If the there any dismissed issues than then default back to Grey color assuming that is the intended UI.

mjkkirschner:
running serial tests: 
https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/

QilongTang:
> Also easy to fix... Just need to know the intended UI. If the there any dismissed issues than then default back to Grey color assuming that is the intended UI.

Yes, that should be the case

Jingyi-Wen:
> > Also easy to fix... Just need to know the intended UI. If the there any dismissed issues than then default back to Grey color assuming that is the intended UI.
> 
> Yes, that should be the case. @Jingyi-Wen Can you confirm as well?

That is correct, it should be set back to grey

QilongTang (approved):

QilongTang:
@Jingyi-Wen Thank you for confirming. @saintentropy as we discussed with @Amoursol , let's time-box the bug fix of warning state competing with preview off state. If you find it complex, we can leave that to 2.14

saintentropy:
https://user-images.githubusercontent.com/2145751/145886776-18f5e053-7e76-48f2-9a5e-3b5b62b290e1.mp4

saintentropy:
@mjkkirschner @QilongTang I pushed the changes to respect the dismissed error state.  Let me know how those changes look.  Also good if someone else can test in the UI.  I don't know if I am missing a condition.

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
```diff
@@ -1105,36 +1120,41 @@ private void BuildErrorBubble()
             ErrorBubble.DismissedMessages.CollectionChanged += DismissedNodeMessages_CollectionChanged;
         }
 
+        // These colors are duplicated from the DynamoColorsAndBrushesDictionary as it is not assumed that the xaml will be loaded before setting the color
+        // SharedDictionaryManager.DynamoColorsAndBrushesDictionary["NodeErrorColor"];
+        private static SolidColorBrush errorColor = (SolidColorBrush)(new BrushConverter().ConvertFrom("#EB5555"));
```
mjkkirschner, line 1125: this is kind of a bummer for maintainability, I'm fine with it for 2.13 - but I'm assuming if the color gets set before the view is loaded it's better to just deal with the null or exception and handle it later when the view is actually loaded. For now can you add a TODO / file a task as tech debt to improve this?

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
```diff
@@ -1066,6 +1064,21 @@ void logic_PropertyChanged(object sender, PropertyChangedEventArgs e)
             }
         }
 
+        /// <summary>
+        /// Respond to property changes on the error bubble.
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="e"></param>
+        void ErroBubble_PropertyChanged(object sender, PropertyChangedEventArgs e)
```
mjkkirschner, line 1074: @saintentropy spelling

mjkkirschner:
@saintentropy awesome that you found the culprit, it was as you guessed?

PR #12498, opened 2021-12-15
fix performance issue with error bars by batching dispatcher calls

mjkkirschner:
### Purpose

https://jira.autodesk.com/browse/DYN-4286
batch repeated calls to the dispatcher into a single call so there is less overhead - we avoid switching back and forth between threads, and the dispatcher only needs to schedule a single task, not `n*m` (where n is nodes and m is error message count)

This PR has the downside that we're modify the model on the dispatcher thread asynchronously, this is to avoid switching between the UI and scheduler thread, but it may mean that node messages/ warnings / state are not set to their expected values at exactly the right time. ⚠️ On The CI machines, it appears tests fail for this reason - I cannot reproduce this locally yet but guess it is up to different threading behaviors between nunit console and the vs test runner.

It would be ideal if we could retain these performance improvements while, still modifying the node models on the scheduler thread so that these properties are set before the execution is marked complete.

Using a synchronous call to the dispatcher results in deadlocks in periodic mode, so that does not appear to be a simple option either.

I plan to explore in an alternative PR if we can make the updates to the node models on the scheduler, but temporarily suppress property notifications for specific properties, then when the execution is complete trigger the UI to build the appropriate error bubbles. I think this approach will trade keeping client changes to a minimum for less idiomatic code.

In any case, these changes require testing in Revit and Sandbox integrators, (Formit,Alias etc), especially around error messages, api access to error messages, and shutdown while periodic mode is active and executing runs.

after
Dec-15-2021 12-09-45

### Declarations

### Release Notes

Add new API to WorkspaceModel to dispatch calls to the UI thread.


mjkkirschner:
I'll take a look at the test failures.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -565,6 +565,32 @@ private void OnSyncWithDefinitionEnd(NodeModel nodeModel)
         {
             hasNodeInSyncWithDefinition = false;
         }
+        public void DispatchOnUIThread(Action a, bool blocking = false)
```
QilongTang, line 568: Comments thanks!

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -565,6 +565,32 @@ private void OnSyncWithDefinitionEnd(NodeModel nodeModel)
...
+        {
+            OnDispatchedToUI(this, new UIDispatcherEventArgs(a,blocking));
+        }
+
+        private void OnDispatchedToUI(object sender, UIDispatcherEventArgs e)
```
QilongTang, line 573: This is great, I remember there are some other places where I check if dispatcher is null, if yes, just execute it. This is worth making an APi

QilongTang (commented):
LGTM

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -2764,10 +2764,19 @@ public UIDispatcherEventArgs(Action a)
         {
             ActionToDispatch = a;
         }
+        public UIDispatcherEventArgs(Action a, bool blocking)
```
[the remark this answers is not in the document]
mjkkirschner: yes, will add comments, theres probably something better we can do - ideally the model layer wouldn't need to care about the UI, but I wanted to have this fixed fast if we needed it for a hotfix. 

I'm hoping we can find something more elegant long term in which case I was thinking about making this internal... I can still add comments though.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
aparajit-pratap: Does this need to be public now?

src/DynamoCoreWpf/ViewModels/Core/HomeWorkspaceViewModel.cs
```diff
@@ -69,11 +70,37 @@ public HomeWorkspaceViewModel(HomeWorkspaceModel model, DynamoViewModel dynamoVi
...
+                }
+
+            }
+            else
+            {
+                if (e.Blocking)
+                {
+                    e.ActionToDispatch.Invoke();
+                }
+                e.ActionToDispatch.BeginInvoke(e.ActionToDispatch.EndInvoke,null);
```
aparajit-pratap, line 99: Shouldn't this be inside an `else`?

mjkkirschner: yikes!, thanks @aparajit-pratap!

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
mjkkirschner: Agreed, I will see about making it internal.

aparajit-pratap (approved):

mjkkirschner:
there were still a few failing tests here - they failed intermittently because I moved from setting the warning data on the `NodeModel` from synchronously to asynchronously, these tests expect that the after a run command that the warning data on `NodeModel` will be up to date - I would also expect this, so I've reverted that particular change in behavior.

Performance is still good.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -566,6 +566,39 @@ private void OnSyncWithDefinitionEnd(NodeModel nodeModel)
             hasNodeInSyncWithDefinition = false;
         }
 
+        /// <summary>
+        /// Dispatches an action onto the UI thread. If there are no handlers for the underlying event
+        /// action is executed directly.
+        /// </summary>
+        /// <param name="a">the Action to execute</param>
+        /// <param name="blocking">If this action should run synchronously.</param>
+        internal void DispatchOnUIThread(Action a, bool blocking = false)
```
aparajit-pratap, line 575: If synchronous calls are the more common case, then should we just make the default value to be `true`?

mjkkirschner: I thought about it. Probably. I was hoping it would encourage clients to ponder - can I make this asynchronous somehow, but maybe it will just cause hard to find bugs. I'll invert it.

mjkkirschner:
moving this to draft until I can get back to it - latest changes seem to make tests time out 🙃 ...

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -744,15 +744,19 @@ private void OnUpdateGraphCompleted(AsyncTask task)
...
+                    var node = workspace.Nodes.FirstOrDefault(n => n.GUID == guid);
+                    if (node == null)
+                        continue;
 
-                node.Warning(message.Value); // Update node warning message.
-            }
+                    node.Warning(message.Value); // Update node warning message.
+                }
+            }),false);
+
```
pinzart90, line 759: so before this code...the `node.Warning(message.Value)` call was invoked in the context of the Scheduler thread ?

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
pinzart90: I would have been nice to have this logic in a lower level assembly....like DynamoUtilities

src/DynamoCore/Graph/Nodes/NodeModel.cs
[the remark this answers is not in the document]
pinzart90: Would have been nice to have this in a lower level assembly

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
pinzart90: So this method looks like it can do the following:
1. Execute code in the same thread context
  a. Async
  b. Sync
2. Execute code in another thread context
  a. Async
  b. Sync

pinzart90:
@mjkkirschner  Looks like great work.
Maybe we can create a separate jira task to address this issue more generically (not just for warnings). Or at least an investigation so we can establish some guidelines

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
mjkkirschner: yep

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
mjkkirschner: yes, exactly, and foreach node foreach message update that triggers a dispatch to the UI thread.

mjkkirschner:
Just a note that I personally feel this solution to this particular problem is not acceptable due to the change in API behavior - I could probably get the tests to pass by mess around with DispatcherUtils, but I'm inclined not to, and not sure that is the issue. I will keep it open for discussion though for now.

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
pinzart90: I don;t see any locking mechanisms in the warning logic...
I wonder if there could be issues with that..

PR #12542, opened 2022-01-13
fix performance issue with error bars by suppressing property change notifications.

mjkkirschner:
### Purpose
alternative to:
https://github.com/DynamoDS/Dynamo/pull/12498

This PR is a bit more brute force than the other, but I think it's simpler to reason about. ~(at least until the tests run)~..

A new internal property `SuppressStatePropertyNotifications` is added to `NodeModel` to stop property notifications for `ToolTipText` and `State` from being raised to the View temporarily.

These properties kick off `error bubble` and `warning color` updates which are quite slow when marshaled between threads.
Something I didn't know is that WPF will automatically marshal property change events for properties that are part of a binding to the dispatcher thread. This means that both `WarningBarColor` and `TooltipText` changes on other threads are the real culprits for this slow down, and changes to `State` kick off changes to `WarningBarColor`.

To avoid the cross thread chatter, we disable the change notifications and manually update all node error bubbles that had generated a message to display.

So far, simple testing look ok, will see what the tests come back with.

### Declarations

Check these if you believe they are true

- [x] The codebase is in a better state after this PR
- [x] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [x] User facing strings, if any, are extracted into `*.resx` files
- [x] All tests pass using the self-service CI.
- [x] Snapshot of UI changes, if any.
- [x] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.
- [ ] This PR modifies some build requirements and the readme is updated

### Release Notes

(FILL ME IN) Brief description of the fix / enhancement. **Mandatory section**

PR #12553, opened 2022-01-18
fix performance issue with error bars by suppressing property change notifications.

mjkkirschner:
opening this again - accidentally closed https://github.com/DynamoDS/Dynamo/pull/12542...
 Tests passed previously.

### Purpose
alternative to:
https://github.com/DynamoDS/Dynamo/pull/12498

### UPDATED
I've tried to address the review feedback and make this notion of controlling when `propertyChanged` events are raised more general, I think the solution is okay, and can be used in other spots - at the same time, it's difficult to test/profile if this actually causes any performance hit. My intuition says minimal.... (famous last words)

A new internal property `PropertyChangeManager` is added to `NotificationObject` to control if property notifications are raised for `NotificationObject`s - this class has a list of prop names to check against - `NotificationObject` now checks this list before raising any `PropertyChanged` events - so we can use this functionality from any `NotificationObject` without changing the property get/set accessor code.

I think the downside is that now we need to do a null check for all property change events as we check if the list is null before raising the event, in 99.9% of cases the list is null, and we don't need to search it for any matching property names.

It implements `IDisposable`, so it's easy to set and then clear the list of prop names just for a small timeframe.

One open question I have is if we should lock on the assessors of the prop names incase one attempts to dispose and set the list from different threads - I think this is rather unlikely given this is an internal API, and the issue this is attempting to resolve only manifests when we raise property changes on the model - my feeling is it's an unnecessary complication / potential for deadlock.

------
`State` and `TooltipText` properties kick off `error bubble` and `warning color` updates which are quite slow when marshaled between threads.
Something I didn't know is that WPF will automatically marshal property change events for properties that are part of a binding to the dispatcher thread. This means that both `WarningBarColor` and `TooltipText` changes on other threads are the real culprits for this slow down, and changes to `State` kick off changes to `WarningBarColor`.

To avoid the cross thread chatter, we disable the change notifications and manually update all node error bubbles that had generated a message to display.

So far, simple testing look ok, will see what the tests come back with.

### Declarations

### Release Notes


QilongTang:
Thank you @mjkkirschner

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -68,7 +68,11 @@ public abstract class NodeModel : ModelBase, IRenderPackageSource<NodeModel>, ID
 
         private readonly Dictionary<int, Tuple<int, NodeModel>> inputNodes;
         private readonly Dictionary<int, HashSet<Tuple<int, NodeModel>>> outputNodes;
-
+        /// <summary>
+        /// Can be used to suppress some property notifications related to node warnings/state. This is useful
+        /// to defer UI updates for better performance. //TODO should this be generic for all nodemodel props?
+        /// </summary>
+        internal bool SuppressStatePropertyNotifications { get; set; } = false;
```
QilongTang, line 75: From reading your description, it almost seems that this flag will never be true? Do we want to ask Matterlab to update the design of that feature?

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -767,8 +768,8 @@ private void OnUpdateGraphCompleted(AsyncTask task)
             // Dispatch the failure message display for execution on UI thread.
             // 
             EvaluationCompletedEventArgs e = task.Exception == null || IsTestMode
-                ? new EvaluationCompletedEventArgs(true)
-                : new EvaluationCompletedEventArgs(true, task.Exception);
+                ? new EvaluationCompletedEventArgs(true,messages.Keys.Select(x=>x.ToString()),null)
+                : new EvaluationCompletedEventArgs(true, messages.Keys.Select(x => x.ToString()), task.Exception);
```
QilongTang, line 772: I noticed the keys were converted into string and later converted back to Guids, any reason for not passing the keys as a collection?

src/DynamoCore/Graph/Nodes/NodeModel.cs
mjkkirschner: this flag will be `true` only when when we're en masse setting the `NodeModel.ToolTipText` property or `state` -  at other times it will be set false so that if a client application (for example an extension) sets a single node's state the UI should update automatically.

We're only taking control and making these updates explicit when runs are occurring and it's likely theres going to be many updates happening.

If we want to improve performance for all clients with arbitrarily large change sets we'll need to rethink dynamo's UI architecture.

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
mjkkirschner: yeah, this was just lazy, will make this cleaner.

src/DynamoCore/Graph/Nodes/NodeModel.cs
pinzart90: tihs flag could be moved to the notificationObject class

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -750,8 +750,9 @@ private void OnUpdateGraphCompleted(AsyncTask task)
                 var node = workspace.Nodes.FirstOrDefault(n => n.GUID == guid);
                 if (node == null)
                     continue;
-
+                node.SuppressStatePropertyNotifications = true;
                 node.Warning(message.Value); // Update node warning message.
+                node.SuppressStatePropertyNotifications = false;
```
pinzart90, line 755: if this solution proves to be the best way to solve this problem (generically).
We can create a smart object to make enabling/disabling easier
```
using(new SuppressStatePropertyNotifications(["prop1","prop2"])) {
code...
}
```

aparajit-pratap: RAII: https://www.stubbornprogrammer.com/2020/07/11/raii-in-c-2/

src/DynamoCore/Models/DynamoModelEventArgs.cs
```diff
@@ -172,6 +176,12 @@ public EvaluationCompletedEventArgs(bool evaluationTookPlace, Exception errorMsg
 
             error = errorMsg != null ? Option.Some(errorMsg) : Option.None<Exception>();
         }
+        internal EvaluationCompletedEventArgs(bool evaluationTookPlace, IEnumerable<string> messageKeys, Exception errorMsg = null)
+        {
+            EvaluationTookPlace = evaluationTookPlace;
+            MessageKeys = messageKeys;
+            error = errorMsg != null ? Option.Some(errorMsg) : Option.None<Exception>();
+        }
     }
```
aparajit-pratap, line 185: Does `public EvaluationCompletedEventArgs(bool evaluationTookPlace, Exception errorMsg = null)` need to be deprecated as a result of adding this new constructor?

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
aparajit-pratap: So `messages` here are any messages related to warnings, errors or info that need to be displayed on the node?

mjkkirschner: I think these are just runtime warnings and build warnings:
https://github.com/DynamoDS/Dynamo/blob/94326a0a020257303029bdfb25f0f79f9118b396/src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs#L712

src/DynamoCore/Models/DynamoModelEventArgs.cs
mjkkirschner: I don't think we *need* though we could, I've made the new constructor internal for now, I think we should initialize the `MessageKeys` to an empty list probably though.

aparajit-pratap (approved):

src/DynamoCore/Core/PropertyChangeManager.cs
```diff
@@ -0,0 +1,53 @@
...
+        protected IEnumerable<string> PropertyNames { get; set; }
+        /// <summary>
+        /// Disable property change notifications for a list of a props.
+        /// The previous set of suppressed notifications will be cleared.
+        /// This API is meant to be used only for a short time and Dispose should be
+        /// called on the object this call returns.
+        /// </summary>
+        /// <param name="props"></param>
+        /// <returns></returns>
+        public PropertyChangeManager SetPropsToSuppress(IEnumerable<string> props)
```
pinzart90, line 27: What about using a param array specifier 'params'. It would make it easier to use

mjkkirschner: good idea.

mjkkirschner:
@pinzart90 made the requested change, unfortunately, I cannot seem to find write a test that fails before my changes, though the differences in manual testing are obvious.

src/DynamoCore/Core/PropertyChangeManager.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            PropertyNames = props;
+            return this;
+        }
+        /// <summary>
+        /// Disposing this class will allow prop notifications to be raised.
+        /// </summary>
+        public void Dispose()
+        {
+            PropertyNames = null;
```
pinzart90, line 37: Do we need to trigger the property change notification on dispose?
If we do not...then may e we should specify it in the comments

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -750,8 +750,10 @@ private void OnUpdateGraphCompleted(AsyncTask task)
                 var node = workspace.Nodes.FirstOrDefault(n => n.GUID == guid);
                 if (node == null)
                     continue;
-
-                node.Warning(message.Value); // Update node warning message.
+                using (node.PropertyChangeManager.SetPropsToSuppress(nameof(NodeModel.ToolTipText), nameof(NodeModel.State)))
+                {
+                    node.Warning(message.Value); // Update node warning message.
+                }
```
pinzart90, line 756: So we do not need the property change notification at all here? We are changing model data without notifying the ui ...

mjkkirschner: yes, that is exactly the point, we need to completely avoid triggering the prop notifications for these changes and trigger the ui update ourselves at the correct time.

src/DynamoCore/Core/PropertyChangeManager.cs
mjkkirschner: No, we do not want to trigger the updates on dispose - or at least that will not make this class useful for the problem I am trying to solve with how I am currently using it.

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
mjkkirschner: see changes to `hwm_EvaluationCompleted`

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -750,8 +750,10 @@ private void OnUpdateGraphCompleted(AsyncTask task)
                 var node = workspace.Nodes.FirstOrDefault(n => n.GUID == guid);
                 if (node == null)
                     continue;
-
-                node.Warning(message.Value); // Update node warning message.
+                using (node.PropertyChangeManager.SetPropsToSuppress(nameof(NodeModel.ToolTipText), nameof(NodeModel.State)))
```
aparajit-pratap, line 753: I did not know you can declare an array by simply specifying a comma-separated list of values.

mjkkirschner: Tibi suggested using the params modifier which allows this.

src/DynamoCore/Core/PropertyChangeManager.cs
pinzart90: Sounds like we could improve this class in the future to also have a method called `SetPropsToDelayNotification` that could only trigger the properties that were delayed (on dispose).

src/DynamoCore/Core/NotificationObject.cs
```diff
@@ -20,6 +22,11 @@ public abstract class NotificationObject : INotifyPropertyChanged
         /// <param name="propertyName">The property that has a new value.</param>
         protected virtual void RaisePropertyChanged(string propertyName)
         {
+            //TODO profile this?
```
aparajit-pratap, line 25: I see O(n) worst-case complexity as it would need to search for the property in `PropertyNames`.

aparajit-pratap (approved):

pinzart90 (approved):

mjkkirschner:
profiling looks good, merging.

PR #12563, opened 2022-01-20
Dynamo UI should not block during concurrent save/run operations- attempt 2

mjkkirschner:
### Purpose
while investigating https://jira.autodesk.com/browse/DYN-4521  I found that the new `ExternalFiles` implementation causes a UI hang when when called while the graph is executing. (more specifically when the `LiveRunner` has a lock on its `mutexObject` field) This is commonly during graph executions.

This is attempt 2 adding a new internal method to EngineController/LiverRunner etc to run an `Action` only if the internal `LiveRunner` mutex is not already owned by some thread. 
**This means that save will still occur during graph execution as before but ExternalFile ref calculation will be skipped if the lock is already owned.**

⚠️ Also note I have not yet verified this is the same issue as in Alias.

I've also removed the `HeartBeat` thread as it appears it was only used for instrumentation and logging uptime to the registry which doesn't appear to be used anywhere.

To test this is quite simple, run a long running graph and hit the save icon.

### Declarations

### Release Notes

Fixes a UI hang when saving at the same time the graph is running.


src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -825,46 +825,51 @@ private List<INodeLibraryDependencyInfo> GetExternalFiles()
...
             if (this is HomeWorkspaceModel homeWorkspaceModel)
             {
-                foreach (var node in nodes)
+                homeWorkspaceModel.EngineController.RunActionIfLiveRunnerNotBusy(() =>
                 {
-                    externalFilesDictionary.TryGetValue(node.GUID, out var serializedDependencyInfo);
-
-                    // Check for the file path string value at each of the output ports of all nodes in the workspace. 
-                    foreach (var port in node.OutPorts)
+                    this.Log("!!!!!!!!!computing external deps");
```
mjkkirschner, line 830: guess we can remove this or make it more formal. ;)

pinzart90:
Here are a couple of questions notes I could think up (a bit more general than the target of this PR)
```
1. We should analyze all critical sections for thread contention or deadlocks/race conditions
2. New APIs to access critical sections with timeout for waiting ? - Anyone accessing critical sections needs to be aware that the threads may freeze
3. All access from UI thread to critical sections should be analyzed and mitigated. All new calls to critical sections should take care not to block the UI
4. Why does a read operation for a single node's value have to wait for the entire graph to finish running ?
5. Maybe we can minimize the critical sections to reduce contention
```

PR #12564, opened 2022-01-20
Dynamo UI should not block during concurrent save/run operations- attempt 1

mjkkirschner:
### Purpose
while investigating https://jira.autodesk.com/browse/DYN-4521 I found that  the new `ExternalFiles` implementation causes a UI hang when when called while the graph is executing. (more specifically when the `LiveRunner` has a lock on its `mutexObject` field) This is commonly during graph executions.

This is attempt 1 using an existing simple bool field (`HomeWorkspaceModel.executingTask`) to track if a graph run is in progress. I'm not sure how robust a solution it is, and it's not covering all cases because a client could call some other `EngineController` or `LiveRunner` method that grabs this lock - though it will likely work for most cases, is simple, and works in my repro case.

**This means that save will still occur during graph execution as before but ExternalFile ref calculation will be skipped if the graph is executing.**

~Also note I have not yet verified this is the same issue as in Alias.~ I have verified this fixes the hang in Alias. (well at least the one I experienced during testing).

I've also removed the `HeartBeat` thread as it appears it was only used for instrumentation and logging uptime to the registry which doesn't appear to be used anywhere.

To test this is quite simple, run a long running graph and hit the save icon.

### Declarations

### Release Notes

Fixes a UI hang when saving at the same time the graph is running.


mjkkirschner:
⚠️ 
This doesn't work in some situations - for example see the failing test - It looks like `executingTask` is set to true AFTER it's set to false, so the state is actually inverted. I think this is because the tests put the scheduler into synchronous mode (like Revit) - I don't think this solution will work well in those contexts.

Maybe it could be adapted with another bool, for example `RunEnabled` seems to be reset at the correct time... not sure how it will work with different thread scenarios though... leaning towards solution 2 for now.

mjkkirschner:
`RunEnabled` seems to work, waiting for tests.

pinzart90:
I would rather go with this option.
The second one could probably be made a bit more generic and we could try to look at other similar situations

pinzart90 (approved):
LGTM

mjkkirschner:
this passed on the serial CI as well.

mjkkirschner:
spoke with @pinzart90 about this, given the short timeline we agreed that while this PR does not cover all cases, it's sufficient and it's better to have less changes to test.

@aparajit-pratap @sm6srw please prioritize this PR over attempt 2.

I am working to verify this fixes the reported alias hang.

pinzart90 (approved):
LGTM

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -817,13 +818,18 @@ internal List<INodeLibraryDependencyInfo> ExternalFiles
...
         {
             var externalFiles = new Dictionary<object, DependencyInfo>();
 
-            // Computes the external file references if the Workspace Model is a HomeWorkspaceModel.
-            // The workspace should be executed for the external references to be computed because the node output port values are needed.
-            if (this is HomeWorkspaceModel homeWorkspaceModel)
+            // If an execution is in progress we'll have to wait for it to be done before we can gather the
+            // external file references as this implementation relies on the output values of each node.
+            //instead just bail to avoid blocking the UI.
+            if (this is HomeWorkspaceModel homeWorkspaceModel && homeWorkspaceModel.RunSettings.RunEnabled)
```
aparajit-pratap, line 832: If retrieving external files is skipped here due to a graph running, say when saving a graph, then what action will cause this computation to actually run again at a later time?

aparajit-pratap: In other words, how do we ensure that this is scheduled to run after the graph finishes executing?

PR #12574, opened 2022-01-24
Adjust the proxy ports for the groups that have CBN.

reddyashish:
### Purpose

These changes are to adjust the proxy port positions for the collapsed groups that have code block nodes in it.

This fix would not resolve the port design issues as we would need to have multiple controls to adjust based on the port size. In the current design, the ports get overlapped but after this fix, the ports will be stacked below each other. 

Before:
image-2021-12-11-21-33-16-684

After: 
Screen Shot 2022-01-24 at 8 51 45 PM

### Declarations

### Release Notes

Fix the proxy ports positions for groups that have CBN's. 

### Reviewers

@QilongTang @SHKnudsen @Amoursol

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
@@ -696,7 +696,12 @@ namespace Dynamo.ViewModels
             {
                 case PortType.Input:
                     return new Point2D(Left, y);
-                case PortType.Output:         
+                case PortType.Output:
+                    if (portModel.Owner is CodeBlockNodeModel)
+                    {
+                        // Special case because code block outputs are smaller than regular outputs.
+                        return new Point2D(Left + Width, y + 6);
+                    }
                     return new Point2D(Left + Width, y);
             }
             return new Point2D();
--- a/src/DynamoCoreWpf/Views/Core/AnnotationView.xaml
+++ b/src/DynamoCoreWpf/Views/Core/AnnotationView.xaml
@@ -724,7 +724,6 @@
                               Grid.Row="0"
                               Canvas.ZIndex="20"
                               HorizontalContentAlignment="Left"
-                              Style="{StaticResource InOutPortControlStyle}"
                               ItemsSource="{Binding Path=InPorts}"
                               Margin="-25,10">
                 </ItemsControl>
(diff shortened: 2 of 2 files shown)


QilongTang (approved):

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -696,7 +696,12 @@ private Point2D CalculatePortPosition(PortModel portModel, int proxyPortIndex)
             {
                 case PortType.Input:
                     return new Point2D(Left, y);
-                case PortType.Output:         
+                case PortType.Output:
+                    if (portModel.Owner is CodeBlockNodeModel)
+                    {
+                        // Special case because code block outputs are smaller than regular outputs.
+                        return new Point2D(Left + Width, y + 6);
```
QilongTang, line 703: Assuming there is a shift in this case, is there a way we can get the port to be closer to the group edge?

mjkkirschner: can this value be based on a ratio instead of an exact pixel value?

src/DynamoCoreWpf/Views/Core/AnnotationView.xaml
```diff
@@ -724,7 +724,6 @@
                               Grid.Row="0"
                               Canvas.ZIndex="20"
                               HorizontalContentAlignment="Left"
-                              Style="{StaticResource InOutPortControlStyle}"
```
mjkkirschner, line 727: what consequences does removing these styles have?

reddyashish: This styling doesn't make any additional difference to other ports as this same style will be inherited from the node ports(they have this same styling)

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
reddyashish: @QilongTang If we adjust the margin more, it will displace ports coming from other nodes. This margin has been adjusted considering the area used for port snapping.

reddyashish:
Merging this after talking to @mjkkirschner and @QilongTang. Will be creating a followup task to investigate further after the 2.13.1 hotfix.

PR #12582, opened 2022-01-25
performance graphs should be in manual run mode.

mjkkirschner:
Python replication graph "OPEN" is failing on the perf CI since novemberish. 
It appears this graph has its run mode set to automatic so it's likely that it's runtime was affected by some changes to python updates / removal...

The underlying custom node graph that is referenced by this graph did have its engine type changed to `cpython` 2 months ago.

if this causes a failure on the CI still, we may need to update the baseline for the "RUN" perf test for this graph.

### DIFF
--- a/tools/Performance/DynamoPerformanceTests/graphs/PythonReplication.dyn
+++ b/tools/Performance/DynamoPerformanceTests/graphs/PythonReplication.dyn
@@ -79,7 +79,7 @@
       "HasRunWithoutCrash": true,
       "IsVisibleInDynamoLibrary": true,
       "Version": "2.11.0.3854",
-      "RunType": "Automatic",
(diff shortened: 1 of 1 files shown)


mjkkirschner:
https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoPerformance-CI/job/master/1093/console

mjkkirschner:
@pinzart90 please see the linked performance job results, it appears that the baselines were incorrect - ie the OPEN time included the runtime and the RUN time was essentially nothing. 63 micro seconds - man thats fast! 🙃 .
I will send another PR to that repo updating the baselines for these tests.

pinzart90 (approved):

PR #12586, opened 2022-01-26
DYN-3861 CRASH: ArgumentNullException in UpdateRenderPackageAsyncTask.Initialize

sm6srw:
### Purpose

Add some defensive code for avoiding crash.

### Declarations

### Release Notes

### Reviewers

### FYIs

### DIFF
--- a/src/DynamoCore/Graph/Nodes/NodeModel.cs
+++ b/src/DynamoCore/Graph/Nodes/NodeModel.cs
@@ -2546,7 +2546,15 @@ namespace Dynamo.Graph.Nodes
             };
 
             var task = new UpdateRenderPackageAsyncTask(scheduler);
-            if (!task.Initialize(initParams)) return false;
+            try
+            {
+                if (!task.Initialize(initParams)) return false;
+            }
+            catch (ArgumentNullException e)
+            {
+                Log(e.ToString());
+                return false;
+            }
 
             task.Completed += OnRenderPackageUpdateCompleted;
             scheduler.ScheduleForExecution(task);
--- a/src/Libraries/Watch3DNodeModelsWpf/Watch3DNodeViewCustomization.cs
+++ b/src/Libraries/Watch3DNodeModelsWpf/Watch3DNodeViewCustomization.cs
@@ -46,9 +46,6 @@ namespace Watch3DNodeModelsWpf
             var dynamoViewModel = nodeView.ViewModel.DynamoViewModel;
             watch3dModel = model;
 
-            var renderingTier = (RenderCapability.Tier >> 16);
-            if (renderingTier < 2) return;
-
             var dynamoModel = dynamoViewModel.Model;
 
             var vmParams = new Watch3DViewModelStartupParams(dynamoModel);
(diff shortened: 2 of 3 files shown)


src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -2546,7 +2546,15 @@ internal void RequestValueUpdate(EngineController engine)
             };
 
             var task = new UpdateRenderPackageAsyncTask(scheduler);
-            if (!task.Initialize(initParams)) return false;
+            try
+            {
+                if (!task.Initialize(initParams)) return false;
```
aparajit-pratap, line 2551: Can't we simply add a null check for `initParams` if that's what's going on here?

src/Libraries/Watch3DNodeModelsWpf/Watch3DNodeViewCustomization.cs
```diff
@@ -46,9 +46,6 @@ public void CustomizeView(Watch3D model, NodeView nodeView)
             var dynamoViewModel = nodeView.ViewModel.DynamoViewModel;
             watch3dModel = model;
 
-            var renderingTier = (RenderCapability.Tier >> 16);
```
mjkkirschner, line 49: interesting, why remove these, I think they are checking for shader model, texture size etc?

sm6srw: This is unrelated, sort of. Just wanted to see if removing this causes any failures. The problem is, keeping this check prevents Watch3D nodes to work in software mode. It is not needed as far as I can tell.

mjkkirschner: ahh, got it, it may make more sense to check device info directly or something like.

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -2546,7 +2546,15 @@ internal void RequestValueUpdate(EngineController engine)
...
+            }
+            catch (ArgumentNullException e)
```
aparajit-pratap, line 2553: which argument is null? if we know, we can just introduce a null check, can we?

sm6srw: I have not been able to recreate this so don't know actually. It is one of the argument checks in Initialize that throws and the argument name is passed to the exception. Will we actually see what argument it is now with our recent updates to our exceptions logging? One other option would be to not throw at all in Initialize and just return false if anything is null.

aparajit-pratap: By looking at the `Initialize` method it appears that the `AstIdentifierForPreview` property of the nodemodel could be `null`. If so, could add a null check for it [here](https://github.com/DynamoDS/Dynamo/blob/01994440fdc5eb78f34ddb15a482b4bef4027859/src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs#L96)?

sm6srw: The call stack indicates an `ArgumentNullException`. My guess is that this happens at the very beginning of the initialize method where we check the arguments [here](https://github.com/DynamoDS/Dynamo/blob/01994440fdc5eb78f34ddb15a482b4bef4027859/src/DynamoCore/Scheduler/UpdateRenderPackageAsyncTask.cs#L77)?

src/Libraries/Watch3DNodeModelsWpf/Watch3DNodeViewCustomization.cs
sm6srw: I will revert this and move it to a separate PR.

aparajit-pratap (approved):

PR #12588, opened 2022-01-26
Creation of an undo action to unpin the node

filipeotero:
### Purpose

When a note is pinned in a node, it is not possible to undo this action, bugging the interface by just repositioning or deleting the note.
To fix this, an action was created to be subscribed when a note is pinned. 

unpinNodeUndo

### Declarations

### Reviewers

@QilongTang 

### FYIs

@RobertGlobant20 @jesusalvino


src/DynamoCore/Graph/Notes/NoteModel.cs
```diff
@@ -11,6 +11,14 @@ namespace Dynamo.Graph.Notes
     /// </summary>
     public class NoteModel : ModelBase
     {
+        /// <summary>
+        /// 
+        /// </summary>
+        public event Action<ModelBase> UndoRequest;
```
mjkkirschner, line 17: in general, make the access modifiers for members as conservative as possible (internal, private etc) - if you make things public we need to maintain them until the next major release - this is very difficult.

src/DynamoCore/Graph/Notes/NoteModel.cs
```diff
@@ -11,6 +11,14 @@ namespace Dynamo.Graph.Notes
...
+
+        private bool createdPinNode;
```
mjkkirschner, line 19: it seems you can just use auto properties here instead of the field.
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/auto-implemented-properties


in addition /// summary tags are good, sometimes even for internal members.

src/DynamoCore/Graph/Notes/NoteModel.cs
```diff
@@ -40,11 +48,13 @@ public NodeModel PinnedNode
             get { return pinnedNode; }
             set
             {
+                CreatedPinNode = true;
```
mjkkirschner, line 51: instead of creating a new member, can't you just check that this is not null?

src/DynamoCore/Graph/Notes/NoteModel.cs
```diff
@@ -111,10 +122,16 @@ protected override void DeserializeCore(XmlElement nodeElement, SaveContext cont
             Text = helper.ReadString("text", "New Note");
             X = helper.ReadDouble("x", 0.0);
             Y = helper.ReadDouble("y", 0.0);
+            createdPinNode = helper.ReadBoolean("createdPinNode");
 
             // Notify listeners that the position of the note has changed, 
             // then parent group will also redraw itself.
             ReportPosition();
+            if(createdPinNode && UndoRequest != null)
```
mjkkirschner, line 130: what's happening here is not obvious, and it seems strange to raise an event `UndoRequest` every time a note is deserialized, I know this method is currently used during undo/redo and copy paste constructors... maybe you could do a small write up in the PR description of what the issue is, and we can brainstorm other ways to fix this.

mjkkirschner:
@filipeotero  I've added some comments - but at first glance, the solution seems a bit confusing - some general thoughts:
* tests for the case you are fixing.
* more detailed PR description with what the issue is and how this change fixes it.
* there are some test failures to address.

src/DynamoCore/Graph/Notes/NoteModel.cs
mjkkirschner: in any case, I would definitely add a comment here if end up keeping this solution.

src/DynamoCore/Graph/Notes/NoteModel.cs
filipeotero: I was following how the properties were defined in the class. I see your point, changed to internal.

src/DynamoCore/Graph/Notes/NoteModel.cs
filipeotero: Fixed. Thanks!

src/DynamoCore/Graph/Notes/NoteModel.cs
filipeotero: In this case, checking if `pinnedNode` is null will not work because I need to check if It's the first action after pinning the note.

src/DynamoCore/Graph/Notes/NoteModel.cs
filipeotero: I changed this code to the ActionNode method as a suggestion. Let me know your opinion about it.

filipeotero:
> @filipeotero I've added some comments - but at first glance, the solution seems a bit confusing - some general thoughts:
> 
> * tests for the case you are fixing.
> * more detailed PR description with what the issue is and how this change fixes it.
> * there are some test failures to address.

Do you mean creating unit tests?

mjkkirschner:
> > @filipeotero I've added some comments - but at first glance, the solution seems a bit confusing - some general thoughts:
> > 
> > * tests for the case you are fixing.
> > * more detailed PR description with what the issue is and how this change fixes it.
> > * there are some test failures to address.
> 
> Do you mean creating unit tests?

yes

src/DynamoCore/Graph/Notes/NoteModel.cs
QilongTang: @filipeotero Is it because without check, we could also accidentally undo the text input of note? Would you eleborate?

test/DynamoCoreWpfTests/NoteViewTests.cs
```diff
@@ -172,6 +174,9 @@ public void NoteIsPlacedAbovePinnedNode()
 
             var noteModel = noteView.ViewModel.Model;
             var nodeModel = nodeView.ViewModel.NodeModel;
+
+            Thread.Sleep(5000);
```
QilongTang, line 178: Why do we need this line?

src/DynamoCore/Graph/Notes/NoteModel.cs
```diff
@@ -11,6 +11,16 @@ namespace Dynamo.Graph.Notes
...
     {
+        /// <summary>
+        /// This action is triggered when undo command is pressed and a node is pinned
+        /// </summary>
+        internal event Action<ModelBase> UndoRequest;
+
+        /// <summary>
+        /// Checks if the node has just been pinned
+        /// </summary>
+        internal bool CreatedPinNode { get; set; }
```
QilongTang, line 22: Also if we have to use this new boolean, it might be good to follow the code convention to start the bool with `Is`, e.g. `IsNodeJustPinned`

src/DynamoCore/Core/UndoRedoRecorder.cs
```diff
@@ -389,9 +389,14 @@ private void UndoActionGroup(XmlElement actionGroup)
...
                         if (toBeUpdated != null)
                         {
-
                             RecordActionInternal(newGroup, toBeUpdated, modelActionType);
                             undoClient.ReloadModel(element);
+
+                            if(toBeUpdated is Graph.Notes.NoteModel)
+                            {
+                                var noteModel = (Graph.Notes.NoteModel)toBeUpdated;
+                                noteModel.TryToSubscribeUndoNote();
```
QilongTang, line 398: Is it because the `pinnedNode` is not a serializable property so we have to make an exception here?

src/DynamoCore/Graph/Notes/NoteModel.cs
```diff
@@ -112,11 +120,25 @@ protected override void DeserializeCore(XmlElement nodeElement, SaveContext cont
             X = helper.ReadDouble("x", 0.0);
             Y = helper.ReadDouble("y", 0.0);
 
+            if(pinnedNode != null)
+                pinnedNode.GUID = helper.ReadGuid("pinnedNode");
```
QilongTang, line 124: Can we move the code from undoredorecorder here?

src/DynamoCore/Core/UndoRedoRecorder.cs
QilongTang: Again, I would love to see if we can remove the special case here

src/DynamoCore/Graph/Notes/NoteModel.cs
filipeotero: yes, sure!

QilongTang:
Waiting for PR checks to pass, LGTM

QilongTang (approved):

QilongTang (approved):

mjkkirschner:
@QilongTang I just wanted to revist this after seeing the demo - doesn't this seem like it should be solvable with the undo/redo recorder using an undo group?

QilongTang:
@mjkkirschner Can you be more specific? Do you mean to undo both the note and node in a single action?

mjkkirschner:
yes, when the pin happens create an undo group for the creation of both of them and add it to the undo stack.

PR #12609, opened 2022-02-06
Dyn 4533 selection button style

jesusalvino:
### Purpose

Adding and Applying the standard styling to the Dynamo Core Selection Button

### Declarations

### Reviewers

@QilongTang

### FYIs

@RobertGlobant20 @filipeotero


QilongTang:
hi @jesusalvino Thanks for the changes, did you figure out how to place selection node in DynamoSandbox and would you attach a UI difference screenshot as well?

src/Libraries/CoreNodeModelsWpf/Controls/ElementSelectionControl.xaml
```diff
@@ -7,7 +7,75 @@
              mc:Ignorable="d" 
              d:DesignHeight="100" d:DesignWidth="200">
     <UserControl.Resources>
-        <converters:SelectionButtonContentConverter x:Key="SelectionButtonContentConverter" />
+        <ResourceDictionary>
```
mjkkirschner, line 10: @jesusalvino @QilongTang please see this comment I made on Long's recent PR - https://github.com/DynamoDS/Dynamo/pull/11958#discussion_r797938320

TLDR - I think this is an anti pattern and we should move the resource to the shared resource dictionary, not create it inside the control's dictionary or resources element.

mjkkirschner: the relevant quote is:

> A ResourceDictionary defined inside of a [UserControl](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.usercontrol) carries a penalty. The platform creates a copy of such a ResourceDictionary for every instance of the UserControl. If you have a UserControl that is used a lot, then move the ResourceDictionary out of the UserControl and put it at the page level.

src/Libraries/CoreNodeModelsWpf/Controls/ElementSelectionControl.xaml
```diff
@@ -7,7 +7,75 @@
...
+                                        <SolidColorBrush Color="#FFFFFF" Opacity="0.5"/>
+                                    </Border.BorderBrush>
+
+                                    <TextBlock x:Name="textBlock"
+                                       HorizontalAlignment="Center"
+                                       Margin="0,1,0,0"
+                                       VerticalAlignment="Center"
+                                       FontFamily="{StaticResource ArtifaktElementRegular}"
+                                       FontSize="14px"
+                                       Foreground="#FFFFFF"
```
mjkkirschner, line 55: for all the colors defined here, it would be better to use the colors that are already defined in the dynamo modern and other resource dictionaries so that changing the color in the single place will update all controls, for example if we want to make themeing easy, or support different color schemes for accessibility.

src/Libraries/CoreNodeModelsWpf/Controls/ElementSelectionControl.xaml
QilongTang: I agree, converter could stay but the style can be put in https://github.com/DynamoDS/Dynamo/blob/master/src/Notifications/Styles/DynamoModern.xaml

src/Libraries/CoreNodeModelsWpf/Controls/ElementSelectionControl.xaml
jesusalvino: Done, thanks for the advice.

src/Libraries/CoreNodeModelsWpf/Controls/ElementSelectionControl.xaml
jesusalvino: Done, thanks for the advice.

jesusalvino:
> hi @jesusalvino Thanks for the changes, did you figure out how to place selection node in DynamoSandbox and would you attach a UI difference screenshot as well?

Per the team's answer in the channel there is not a way to do that, We were researching a bit about that but it look like need more time. the expected behavior has been attached.

imagen


SelectNode

mjkkirschner:
It looks good to me - but I'm not really sure what it's supposed to look like :) (ie why is there an OK button?) did you apply this style to some dialog just for testing in sandbox?

Would be good to verify how it looks in Revit, could be done after merge.
Also please - quickly go through the checkboxes in the PR template if applicable.

Thanks for addressing the comments. 👍

mjkkirschner (approved):

jesusalvino:
> 

ohh, the OK is just for comparison purpose, the expected behavior is the [Select] one, yes it was applied only for testing in Sandbox.

PR #12615, opened 2022-02-08
Thread safe warnings

pinzart90:
### Purpose
Crash, how to reproduce:
1. It looks like `RuntimeStatus.Warnings` can be accessed through `LogWarningMessageEvents.OnLogWarningMessage`
2. LibG calls `LogWarningMessageEvents.OnLogWarningMessage` every time the produced geometry seems to exceed the ASM modeler extents/bounds
3. There are some code paths in Dynamo that call into LibG APIs directly from the UI thread (ex. In NodeManipulator)
4. `RuntimeStatus.Warnings` are called a lot from WorkerThreads (that execute the graph)
5. It can happen that the UI thread call into Warnings at the same time as the Worker thread

Seems like the warnings List is accessed from multiple threads. 
This is an attempt to make the warnings thread safe

### Declarations

### Release Notes

Thread safe warnings

### Reviewers

### FYIs


src/Engine/ProtoCore/RuntimeStatus.cs
```diff
@@ -56,7 +57,7 @@ public struct WarningEntry
     public class RuntimeStatus
     {
         private ProtoCore.RuntimeCore runtimeCore;
-        private List<Runtime.WarningEntry> warnings;
+        private ConcurrentStack<Runtime.WarningEntry> warnings;
```
pinzart90, line 60: ConcurrentStack<> seemed to be the closest thread safe collection implementation to List<>

src/Engine/ProtoCore/RuntimeStatus.cs
```diff
@@ -70,7 +71,16 @@ public IOutputStream WebMessageHandler
             set;
         }
 
+        [Obsolete("Property will be deprecated in Dynamo 3.0. Please use WarningEntries instead")]
         public IEnumerable<Runtime.WarningEntry> Warnings
+        {
+            get
+            {
+                return warnings.ToList();
```
mjkkirschner, line 79: I'm confused, this had no setter, and this getter copies the list of Warnings - where did you find concurrent access to the underlying collection where the existing getter caused a problem?

src/Engine/ProtoCore/RuntimeStatus.cs
mjkkirschner: may be worth profiling this vs just a simple lock:
https://stackoverflow.com/questions/19326197/net-4-0-concurrent-collection-performance

I think it depends on the use case of course, also not sure this stack overflow question is the best source of benchmarks for this.

src/Engine/ProtoCore/RuntimeStatus.cs
pinzart90: I managed to cause a crash a couple of times ...but not reliably. The Warnings List was being modified on one thread while it was also being iterated on another thread. 
Happened while moving MouseManipulators around...

src/Engine/ProtoCore/RuntimeStatus.cs
pinzart90: Yeah..I was just thinking maybe a lock is enough...

pinzart90: Although, If we always make a copy on the getter...it might be too slow

src/Libraries/DesignScriptBuiltin/DesignScriptBuiltin.csproj
```diff
@@ -11,10 +11,13 @@
...
   <ItemGroup>
-    <PackageReference Include="System.Collections.Immutable" Version="5.0.0" >
-      <ExcludeAssets>runtime</ExcludeAssets>
+    <PackageReference Include="System.Collections.Immutable" Version="5.0.0">
       <GeneratePathProperty>true</GeneratePathProperty>
+      <IncludeAssets>none</IncludeAssets>
     </PackageReference>
+    <Reference Include="System.Collections.Immutable">
+      <HintPath>$(PkgSystem_Collections_Immutable)\lib\portable-net45+win8+wp8+wpa81\System.Collections.Immutable.dll</HintPath>
+    </Reference>
```
pinzart90, line 20: By including `none` and adding an assembly reference to the immutable dll we are forcing `DesignScriptBuiltin.csproj` to reference the `portable-net45...` dll...and not whatever targetframework specific dll it automatically found before..

src/Engine/ProtoCore/RuntimeStatus.cs
```diff
@@ -70,35 +72,50 @@ public IOutputStream WebMessageHandler
             set;
         }
 
-        public IEnumerable<Runtime.WarningEntry> Warnings
+        public List<Runtime.WarningEntry> Warnings
```
pinzart90, line 75: this should be ok.
SHould **not** be a breaking change.

src/Engine/ProtoCore/RuntimeStatus.cs
```diff
@@ -70,35 +72,50 @@ public IOutputStream WebMessageHandler
...
         {
             get
             {
-                return warnings;
+                using (rwl.CreateReadLock())
+                {
+                    return warnings.ToList();
```
pinzart90, line 81: this is the slowest part.
There is no good way to ensure thread safety for a Linq deferred execution/query.
One other way would be to use more granular locking ... like the System.Collections.Concurrent Namespace

pinzart90: At the moment, Dynamo calls  the `Warnings` getter is only at the end of each Graph execution.
I t should not be that bad of a perf degradation

mjkkirschner:
@pinzart90 this seems okay to me, I guess the idea with the readwrite lock is that it should give slightly better performance when multiple threads are trying to read from the collection?

pinzart90:
> @pinzart90 this seems okay to me, I guess the idea with the readwrite lock is that it should give slightly better performance when multiple threads are trying to read from the collection?

Yes

mjkkirschner (approved):

src/Engine/ProtoCore/RuntimeStatus.cs
aparajit-pratap: How is this not a breaking change?

mjkkirschner: last sentence says the signature should not change.

mjkkirschner (changes requested):

src/Engine/ProtoCore/RuntimeStatus.cs
pinzart90: @mjkkirschner I thought the return type is not included in the method signature

pinzart90:
```
A return type of a method is not part of the signature of the method for the purposes of method overloading. However, it is part of the signature of the method when determining the compatibility between a delegate and the method that it points to.
```

pinzart90: So if somebody created a delegate that points to the getter/setter of the `Warnings` ..that might break
not sure if we should worry about this scenario

mjkkirschner: The only thing that quote indicates a signature does not include the return type, is when overloading - we're not overloading here, and it doesn't mention all the places that the signature does include the return type.

mjkkirschner: Is there any issue here with just keeping it IEnumerable?

src/DynamoUtilities/DynamoLock.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// </summary>
+        internal void UnlockForRead() => rwlock.ExitReadLock();
+        /// <summary>
+        /// Unlocks a write lock. Use this API if you called LockForWrite beforehand.
+        /// </summary>
+        internal void UnlockForWrite() => rwlock.ExitWriteLock();
+        /// <summary>
+        /// Unlocks an upgradeable read lock. Use this API if you called LockForUpgreadableRead beforehand.
+        /// </summary>
+        internal void UnlockForUpgreadableRead() => rwlock.ExitUpgradeableReadLock();
```
aparajit-pratap, line 41: `UnlockForUpgradeableRead`

aparajit-pratap (commented):
I think we can revert changes to test/DynamoCoreTests/DynamoCoreTests.csproj as they're only whitespace changes.

pinzart90:
> I think we can revert changes to test/DynamoCoreTests/DynamoCoreTests.csproj as they're only whitespace changes.

They actually fix formatting to be in line with the rest of the PackageReferences

src/DynamoUtilities/DynamoLock.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// Constructs a disposable object that locks for read and releases the lock when Dispose is called.
+        /// </summary>
+        /// <returns>New disposable object</returns>
+        internal IDisposable CreateReadLock() => Disposable.Create(() => { rwlock.EnterReadLock(); }, () => { rwlock.ExitReadLock(); });
+
+        /// <summary>
+        /// Constructs a disposable object that locks for read and allows lock for write in the same scope. Releases the lock when Dispose is called.
+        /// </summary>
+        /// <returns>New disposable object</returns>
+        internal IDisposable CreateUpgradeableReadLock() => Disposable.Create(() => { rwlock.EnterUpgradeableReadLock(); }, () => { rwlock.ExitUpgradeableReadLock(); });
```
aparajit-pratap, line 77: It doesn't look like this is being used, do we need it, and what is its purpose?

pinzart90: Not used now. But it might be useful. It exposes the remaining lock style.
It is useful when you have a code scope that does read and write. Using upgradeable is a faster and cleaner solution
Ex
```
getOrCreateStuff() {
   using(dynLock.CreateUpgradeableLock())
   {
       if (arr.HasItem()) {
           return arr.Item;
       }
       using(dynLock.CreateWriteableLock())
       {
          return arr.CreateItem();
       }
   }
}
```

test/DynamoCoreTests/MessageLogTests.cs
```diff
@@ -1,13 +1,43 @@
...
+        private Exception addWarningEx;
+        private int waitHandle1 = 0;
+        private int waitHandle2 = 0;
+
+        public void AddWarning()
+        {
+            try
+            {
+                ProtoCore.RuntimeCore runtimeCore = CurrentDynamoModel.EngineController.LiveRunnerRuntimeCore;
+                var warnings = runtimeCore.RuntimeStatus.Warnings.Where(x => x.Message.Contains("Test"));
```
aparajit-pratap, line 23: Where are these warnings containing the message "test" added?

test/DynamoCoreTests/MessageLogTests.cs
```diff
@@ -27,6 +57,36 @@ public void TestWarningMessageLog()
...
+
+                Thread thread1 = new Thread(AddWarning);
+                thread1.Start();
+
+                while (waitHandle2 == 0 && thread1.IsAlive)// wait for thread1 to reach the Warning iteration
+                {
+                    Thread.Sleep(100);
+                }
+
+                LogWarningMessageEvents.OnLogWarningMessage($"Test3");
```
aparajit-pratap, line 79: So at this point, a new warning is added on the main thread to the warnings list while at the same time `thread1` is iterating over the warnings list?

pinzart90: Correct

test/DynamoCoreTests/MessageLogTests.cs
```diff
@@ -27,6 +57,36 @@ public void TestWarningMessageLog()
...
+
+                Interlocked.Decrement(ref waitHandle1);
```
aparajit-pratap, line 81: What happens if you don't decrement this?

src/DynamoUtilities/DynamoLock.cs
pinzart90: Because `ReaderWriterLockSlim` does not have one.

test/DynamoCoreTests/MessageLogTests.cs
pinzart90: Deadlock, I guess

pinzart90: This decrement call to waitHandle1 ensures that the thread1::forLOop does not finish before `Test3` warning is added

src/DynamoUtilities/DynamoLock.cs
aparajit-pratap: I meant to ask why `ReaderWriterLockSlim` doesn't have one but that's okay.

aparajit-pratap (approved):

mjkkirschner (approved):

pinzart90:
> @pinzart90, just reading your PR description - thought occurs to me - this PR looks good, but to fix the original issue couldn't you also schedule the work to be done from the manipulator on to the scheduler/scheduler thread via the DynamoScheduler to avoid the crash?

@mjkkirschner It would solve the threading issue, but passing execution to a single thread might degrade performance. In our case the NodeManipulator uses LibG APIs to generate internal data and graphic data on mouse move (ex. the Gizmo graphics). This needs to have good performance because the gizmo graphics moves with the mouse.
Looking at it again...maybe there are other logic issues too. NodeManipulators call into LibG APIs outside GraphExecution contexts. But those LibG calls can generate warnings and those warnings are kind of assumed to be part of a Graph Execution.
Also I am still trying to figure out a good and generic way that we can guard against future threading issues....

pinzart90:
@mjkkirschner I just realized that maybe you were talking about always running the Warnings code in the Scheduler thread (like all APIs related to warnings would be async executed on the scheduler thread)
THat may work...but I am not sure how the logic inside the warnings will behave when the warning is async...and possibly called after the Graph/node execution

PR #12617, opened 2022-02-08
Revert "Add-ons section header in Library will always be displayed."

reddyashish:
Revert PR 12606 from master as it is causing a crash on library search.


reddyashish:
@QilongTang Merging this to unblock the library error.

PR #12638, opened 2022-02-16
Dyn3629

mjkkirschner:
### Purpose

Add logging when we encounter a conflict with a builtin library package, or a built in library package is set unloaded or scheduled to be unloaded.

related PR:
https://git.autodesk.com/Dynamo/Analytics.NET/pull/48

### Declarations

### Release Notes

Adds analytics for builtin package conflicts

### DIFF
--- a/src/DynamoPackages/Package.cs
+++ b/src/DynamoPackages/Package.cs
@@ -83,7 +83,7 @@ namespace Dynamo.PackageManager
 
         internal void SetAsLoaded() { state = StateTypes.Loaded; errorMessage = ""; }
         internal void SetAsError(string msg = "") { state = StateTypes.Error; errorMessage = msg; }
-        internal void SetAsUnloaded() { state = StateTypes.Unloaded; /*TODO logging?*/ errorMessage = ""; }
+        internal void SetAsUnloaded() { state = StateTypes.Unloaded; errorMessage = ""; }
         internal void ResetState() { state = StateTypes.None; }
 
         internal void SetScheduledForDeletion() { scheduledState = ScheduledTypes.ScheduledForDeletion; }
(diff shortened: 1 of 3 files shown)


QilongTang (approved):

src/DynamoPackages/Package.cs
```diff
@@ -599,6 +600,8 @@ internal void UninstallCore(CustomNodeManager customNodeManager, PackageLoader p
                 if (BuiltInPackage)
                 {
                     LoadState.SetAsUnloaded();
+                    Analytics.TrackEvent(Actions.BuiltInPackageConflict, Categories.PackageManagerOperations, $"{Name } {versionName} set unloaded");
```
aparajit-pratap, line 603: @mjkkirschner what indicates that this code path is a package conflict scenario?

src/DynamoPackages/Package.cs
```diff
@@ -528,6 +528,7 @@ internal void MarkForUninstall(IPreferences prefs)
         {
             if (BuiltInPackage) 
             {
+                Analytics.TrackEvent(Actions.BuiltInPackageConflict, Categories.PackageManagerOperations, $"{Name } {versionName} marked to be unloaded");
```
aparajit-pratap, line 531: same question here - what indicates that this code path is a package conflict scenario?

mjkkirschner: thanks for pointing these out @aparajit-pratap, my thought was that it would be good to note if any builtin packages were being uninstalled or unloaded at all - while it's not necessarily because of a conflict it seems likely, and grouping them all together under one analytics category seemed simpler, the extended description in analytics does call out exactly what the operation was.

We can get rid of these I guess.

PR #12647, opened 2022-02-22
input data deserialization proposal C.

mjkkirschner:
### Purpose

Most closely resembles proposal C from this wiki: https://wiki.autodesk.com/pages/viewpage.action?pageId=1204716652
does the following:
* just wraps input deserialization in a try catch, and forget about inputs which fail to deserialize for any reason.
* introduces new `Type2` property to `NodeInputs` so that old dynamo versions don't try to deserialize them (and fail). We now serialize both `Type` and `Type2`. and `Type` no longer serializes as hostSelection or dropdownSelection.
* ⚠️ An important finding I made while looking at this section of code is that **Dynamo doesn't care what the values are in the input section** when deserializing!!!!!!!, really only the name and the existence in the section are important, the others are just thrown away... for that reason it doesn't really matter what we return from this deserialization - at least not at the current time.

This set of changes fixes two issues:
* files from the future with new InputTypes won't break file deserialization
* stop producing files that legacy dynamo versions can't open.

### TODO - 
- [x] test that Type is not serialized as hostSelection or dropdownSelection for our nodes.
- [x] test that unknown type does not cause deserialization failure

### Release Notes

* no longer produce files that cannot be read by dynamo versions < 2.12.
* serialize a new `Type2` property on graph `Inputs` - this should be used instead of `Type`

### Reviewers

### FYIs

### DIFF
--- a/src/DynamoCore/Graph/Nodes/NodeInputData.cs
+++ b/src/DynamoCore/Graph/Nodes/NodeInputData.cs
@@ -8,11 +8,17 @@ using Newtonsoft.Json.Converters;
 
 namespace Dynamo.Graph.Nodes
 {
+    // NodeInputTypes is now serialized twice as NodeInputData.Type and NodeInputData.Type2.
+    // This enables us to add new values to the enum while allowing previous dynamo versions to successfully
+    // load new files with those new types. This is the case because by default json.net ignores missing properties.
+    // So Type2 is not deserialized at all in previous versions of Dynamo.
+    // Type's setter limits the possible values to a subset of the enum to avoid clients setting this to a value that would break file
+    // deserialization in previous dynamo versions.
+    // TODO We should unify these properties (Type and Type2) when possible (Dynamo 3.x) 
+
     /// <summary>
     /// Possible graph input types. 
     /// </summary>
-    [Obsolete("please use InputTypes instead, will be removed in 3.x. Will not be updated with new types." +
-        "Serializing hostSelection or dropdownSelection will produce broken dynamo files. ")]
     [JsonConverter(typeof(StringEnumConverter))]
     public enum NodeInputTypes
     {
@@ -26,38 +32,20 @@ namespace Dynamo.Graph.Nodes
         colorInput,
         [EnumMember(Value = "date")]
         dateInput,
+        [Obsolete()]
         [EnumMember(Value = "selection")]
         selectionInput,
+        /// <summary>
+        /// Should not be used when setting NodeInputData.Type, use selection instead, can be used with Type2.
+        /// </summary>
         [EnumMember(Value = "hostSelection")]
         hostSelection,
+        /// <summary>
+        /// Should not be used when setting NodeInputData.Type, use selection instead, can be used with Type2.
+        /// </summary>
         [EnumMember(Value = "dropdownSelection")]
         dropdownSelection
     };
-    /// <summary>
-    /// Possible graph input types. This Enum replaces NodeInputTypes. 
-    /// </summary>
-    [JsonConverter(typeof(StringEnumConverter))]
-    public enum InputTypes
-    {
-        [EnumMember(Value = "number")]
-        numberInput,
-        [EnumMember(Value = "boolean")]
-        booleanInput,
-        [EnumMember(Value = "string")]
-        stringInput,
-        [EnumMember(Value = "color")]
-        colorInput,
-        [EnumMember(Value = "date")]
-        dateInput,
-        [Obsolete("Use hostSelection or dropdownSelection instead")]
-        [EnumMember(Value = "selection")]
-        selectionInput,
-        [EnumMember(Value = "hostSelection")]
-        hostSelection,
-        [EnumMember(Value = "dropdownSelection")]
-        dropdownSelection
-    };
-
 
     /// <summary>
     /// Represents a node which acts as a UI input for the graph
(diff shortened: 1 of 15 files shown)


src/DynamoCore/Graph/Nodes/NodeInputData.cs
```diff
@@ -49,7 +77,11 @@ public class NodeInputData
         /// <summary>
         /// The type of input this node is.
         /// </summary>
+        [Obsolete("Obsolete, this member has been replaced by Type2, which may contain new input types.")]
         public NodeInputTypes Type { get; set; }
+        /// The type of input this node is.
+        /// </summary>
+        public InputTypes Type2 { get; set; }
```
pinzart90, line 84: Maybe `InputType` instead of `Type2` ?

src/DynamoCore/Graph/Nodes/NodeInputData.cs
```diff
@@ -104,6 +136,18 @@ public class NodeInputData
             { typeof(Int64),NodeInputTypes.numberInput},
             {typeof(float),NodeInputTypes.numberInput},
         };
+        private static Dictionary<Type, InputTypes> dotNetTypeToInputType = new Dictionary<Type, InputTypes>
```
pinzart90, line 139: can't we reuse the old dictionary `dotNetTypeToNodeInputType` ?

mjkkirschner: How? It returns types from the old enum? What are you thinking?

mjkkirschner: and it's still used to set `Type` to the correct values since we serialize both of them for the time being.

pinzart90: Is it a requirement that they stay the same until only one remains ?

src/Libraries/CoreNodeModels/DropDown.cs
```diff
@@ -74,7 +74,9 @@ public override NodeInputData InputData
                 {
                     Id = this.GUID,
                     Name = this.Name,
-                    Type = NodeInputTypes.dropdownSelection,
+                    //because selection makes more sense than defaulting to number...
+                    Type = NodeInputTypes.selectionInput,
+                    Type2 = InputTypes.dropdownSelection,
```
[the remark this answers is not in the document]
pinzart90: or even stop serializing it ...

src/DynamoCore/Graph/Nodes/NodeInputData.cs
mjkkirschner: `Is it a requirement that they stay the same until only one remains ?` I think that is best from an API point of view, we don't know who is using it external to adsk - it's a public property in what we called a "schema" and in. the JSON anyone could be parsing.

src/DynamoCore/Graph/Nodes/NodeInputData.cs
pinzart90: ok

pinzart90:
Maybe a few comments in code. Somewhere near the old and new enum types.
I would like to see some comments on why there are 2 enums with the same values...what is planned for the future and how to proceed if you need to add new values

pinzart90 (approved):
More comments then LGTM

mjkkirschner:
@pinzart90 I will add comments in code

mjkkirschner:
Hey @pinzart90 - I gave the PR a bit more thought while writing up the comments and adding tests, my latest changes only use a single enum, I originally wanted to use two to make sure that users couldn't set `Type` to the bad values, but I didn't end up doing that anyway because it would break deserialization - I think I have a good solution now using Type's setter.

PTAL!

src/DynamoCore/Graph/Nodes/NodeInputData.cs
```diff
@@ -104,6 +138,8 @@ public class NodeInputData
             { typeof(Int64),NodeInputTypes.numberInput},
             {typeof(float),NodeInputTypes.numberInput},
         };
+
+        [Obsolete("To be removed in Dynamo 3.x")]
```
aparajit-pratap, line 142: Why is this being obsoleted?

mjkkirschner: It's a public method, but it looks like an implementation detail...

aparajit-pratap: If it's going to be replaced with a method having a different name or made internal that's fine. After seeing the obsolete attribute I thought it was no longer needed but was confused since it's still being used in a number of places.

src/DynamoCore/Graph/Nodes/NodeInputData.cs
```diff
@@ -21,16 +32,21 @@ public enum NodeInputTypes
...
         [EnumMember(Value = "selection")]
         selectionInput,
+        /// <summary>
+        /// Should not be used when setting NodeInputData.Type, use selection instead, can be used with Type2.
+        /// </summary>
         [EnumMember(Value = "hostSelection")]
         hostSelection,
+        /// <summary>
+        /// Should not be used when setting NodeInputData.Type, use selection instead, can be used with Type2.
+        /// </summary>
```
pinzart90, line 45: It just occurred to me that maybe we could have used the `[NonSerialized()]` attribute for the  `hostSelection` and `dropdownSelection` members. This would only make sense with 2 (duplicate) enum classes

pinzart90 (approved):
LGTM

PR #12653, opened 2022-02-24
DYN-4586 Count of error/warning increases on source node upon copy/pasting a node

filipeotero:
### Purpose

This PR is to solve the issue of increasing the number of warnings in the nodes after the node has been copied and pasted. 
During the investigation of this issue, it was noticed the bug not only happens with copy/paste actions but also by creating, linking, and deleting a node.

warningNodeActionsExample

First, in this part of the code, there can be noticed that the warning messages related to a node are being concatenated at a single string. With this, the warning messages of the same are not being separated.

```
   // Runtime warnings take precedence over build warnings.
            foreach (var warning in updateTask.RuntimeWarnings)
            {
                var message = string.Join(Environment.NewLine, warning.Value.Select(w => w.Message));
                messages.Add(warning.Key, message);
            }
```

The reason for the bug was that at every end of the workspace evaluation, all the messages (even the processed ones), were being added again in the `NodeMessages ` list.

So, to avoid that, and assuming that the difference between a new message and a message that has already been added to the list is the 
string itself, it was added a previous verification if the string already exists in the list. It avoids the incrementation of the counter and maintains multiple messages for the same node as shown below:

warningNodeSolution

### Declarations

### Reviewers

@QilongTang 

### FYIs

@jesusalvino 
@RobertGlobant20

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
@@ -774,7 +774,6 @@ namespace Dynamo.ViewModels
             DynamoViewModel.Model.PropertyChanged += Model_PropertyChanged;
             DynamoViewModel.Model.DebugSettings.PropertyChanged += DebugSettings_PropertyChanged;
 
-
             //Do a one time setup of the initial ports on the node
             //we can not do this automatically because this constructor
             //is called after the node's constructor where the ports
@@ -1193,8 +1192,7 @@ namespace Dynamo.ViewModels
 
             bool hasErrorOrWarning = NodeModel.IsInErrorState || NodeModel.State == ElementState.Warning || NodeModel.State == ElementState.PersistentWarning;
 
-            if (!(NodeModel.WasInvolvedInExecution && hasErrorOrWarning)) return;
-            
+            if (!(NodeModel.WasInvolvedInExecution && hasErrorOrWarning)) return;            
             if (ErrorBubble == null) BuildErrorBubble();
 
             if (!WorkspaceViewModel.Errors.Contains(ErrorBubble)) return;
@@ -1219,20 +1217,16 @@ namespace Dynamo.ViewModels
             {
                 DynamoViewModel.UIDispatcher.Invoke(() =>
                 {
-                    var messageNode = new InfoBubbleDataPacket(style, topLeft, botRight, content, connectingDirection);
-                    ErrorBubble.NodeMessages.Add(messageNode);
+                    ErrorBubble.NodeMessages.Add(new InfoBubbleDataPacket(style, topLeft, botRight, content, connectingDirection));
                     WarningBarColor = GetWarningColor();
                 });
             }
             else
             {
-                var messageNode = new InfoBubbleDataPacket(style, topLeft, botRight, content, connectingDirection);
-                ErrorBubble.NodeMessages.Add(messageNode);
+                ErrorBubble.NodeMessages.Add(new InfoBubbleDataPacket(style, topLeft, botRight, content, connectingDirection));
                 WarningBarColor = GetWarningColor();
             }
-
-            ErrorBubble.ChangeInfoBubbleStateCommand.Execute(InfoBubbleViewModel.State.Pinned);
-            
+            ErrorBubble.ChangeInfoBubbleStateCommand.Execute(InfoBubbleViewModel.State.Pinned);            
         }
 
         private void UpdateErrorBubblePosition()
--- a/src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
@@ -774,6 +774,7 @@ namespace Dynamo.ViewModels
             DynamoViewModel.Model.PropertyChanged += Model_PropertyChanged;
             DynamoViewModel.Model.DebugSettings.PropertyChanged += DebugSettings_PropertyChanged;
 
+
             //Do a one time setup of the initial ports on the node
             //we can not do this automatically because this constructor
             //is called after the node's constructor where the ports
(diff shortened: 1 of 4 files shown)


src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
```diff
@@ -1217,14 +1218,21 @@ public void UpdateBubbleContent()
             {
                 DynamoViewModel.UIDispatcher.Invoke(() =>
                 {
-                    ErrorBubble.NodeMessages.Add(new InfoBubbleDataPacket(style, topLeft, botRight, content, connectingDirection));
+                    var messageNode = new InfoBubbleDataPacket(style, topLeft, botRight, content, connectingDirection);
+                    if(ErrorBubble.NodeMessages.Count(x=> x.TextMessageExists(content)) == 0)
```
mjkkirschner, line 1222: so two thoughts here... 
1. This will stop duplicate messages from ever being added - I don't think this is the intended behavior - for example, what if we replicate over a list of inputs, we should get multiple error messages, some of which will be the same - please confirm @QilongTang @Amoursol 
2. This seems inefficient, you're checking all the messages to make sure you they are not the same, but you can bail as soon as you see any - thats if I am incorrect on point 1.

mjkkirschner:
Hi @filipeotero  - Please add tests for the bug being fixed check in with @QilongTang and @Amoursol on my question above regarding multiple messages being added as the intended behavior in some situations and tests seem important to make sure this does not crop up again.

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
Amoursol: Comment had via Slack and now resolved with clearer AC's :) Thanks for the feedback @mjkkirschner !

QilongTang: @mjkkirschner @Amoursol AC updated for Jira and additional code changes needed for this PR. Currently, warning/error can only be linked with node but not port so AC could be refined more after more investigations done by @filipeotero

aparajit-pratap: This is not a comment about this PR per se but to @mjkkirschner's point I don't think that we ever got multiple error messages for a node in the case of replication, we always reported the first one. I'm not sure if that behavior was supposed to change after visual refresh but it looks to me it didn't.

aparajit-pratap: About this PR, I'm not sure I understand why `UpdateBubbleContent` is being called for nodes that are not taking part in the evaluation in the first place. For example, if a node is copy-pasted or a new node is added, unless the node in question is participating in the incremental execution of the graph, why should its bubble contents even need to be updated. It seems wasteful. I would expect this function to be called only for the new or modified nodes or the nodes downstream of them.

aparajit-pratap: I think in line [1195](https://github.com/DynamoDS/Dynamo/pull/12653/files#diff-97f7c391a3f292149633414ed4ac50ce1a73a1439bcc3307d20da0fb043a6d78R1195) we are already checking if the node is not involved in the execution but not sure why are we also checking if the node does **not** have an error/warning state.

aparajit-pratap: @pinzart90 do you remember why you added the `!hasErrorOrWarning` condition [here](https://github.com/DynamoDS/Dynamo/blob/e504bebb2e30659ca7881261a7ffbe5efe9f4269/src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs#L1194)?
It looks like we should be able to simply skip updating the bubble content (adding new warnings) for nodes not involved in an execution.

pinzart90: This is the PR that changed that https://github.com/DynamoDS/Dynamo/pull/12346

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
```diff
@@ -1191,8 +1191,8 @@ public void UpdateBubbleContent()
             if (DynamoViewModel == null) return;
 
             bool hasErrorOrWarning = NodeModel.IsInErrorState || NodeModel.State == ElementState.Warning || NodeModel.State == ElementState.PersistentWarning;
-            if (!NodeModel.WasInvolvedInExecution && !hasErrorOrWarning) return;
 
+            if (!(NodeModel.WasInvolvedInExecution && hasErrorOrWarning)) return;
```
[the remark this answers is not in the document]
pinzart90: This would stop duplicate messages from being added

pinzart90:
I would say we need to revisit the Warning system to make sure that The UI (warningBubble) always reflects the NodeModel's backend data.

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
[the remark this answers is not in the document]
filipeotero: This looks similar to the previous implementation made. What do you think about that @QilongTang?

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
filipeotero: Hmmm I see, TY @pinzart90

QilongTang:
@filipeotero Does your current alone fix the bug?

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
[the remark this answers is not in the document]
QilongTang: I found this single line is also block us on https://jira.autodesk.com/browse/DYN-4557 as well. In that case, there is a node with a deprecation warning, on graph open, I got `WasInvolvedInExecution` false and `hasErrorOrWarning` true. In that case, the `UpdateBubbleContent()` again will be skipped so no warning message shown for user at all but Dynamo still says execution with warning.. I am a bit confused on what is the right fix in this case.. FYI: @aparajit-pratap @pinzart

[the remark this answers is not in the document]
filipeotero: Maybe this is a palliative solution but is going to be a problem if we have to count the number of messages (even the duplicated ones). As described in this [task](https://jira.autodesk.com/browse/DYN-4647).

aparajit-pratap:
I think I figured out the issue for at least code block nodes. The `InfoBubbleViewModel`'s `NodeMessages` property is not in sync with the number of warnings on the nodemodel (at least for CBNs). I'm in the process of fixing this for CBN. Once the fix is in I can test to see if it can apply to other nodes as well and fix this issue and DYN-4557.

QilongTang:
> I think I figured out the issue for at least code block nodes. The `InfoBubbleViewModel`'s `NodeMessages` property is not in sync with the number of warnings on the nodemodel (at least for CBNs). I'm in the process of fixing this for CBN. Once the fix is in I can test to see if it can apply to other nodes as well and fix this issue and DYN-4557.

Thank you @aparajit-pratap Then I will hold fire on it

QilongTang (approved):

QilongTang:
I discussed with @aparajit-pratap @pinzart90 on this and we noticed there are additional improvements we can make in terms of efficiency of how Dynamo updates warning bubble in the case only necessary. Since there are a lot of tasks depending on the changes here, we will merge and test together one by one and make sure all test cases are captured. Merging.

PR #12664, opened 2022-02-28
[23Q1T1] Fade-out Node Glyphs

dnenov:
### Purpose

The first step of implementing the changes to the Color Coding behavior of Dynamo Nodes. This PR adds fade-out behavior to the stack panel containing the visual-cue glyphs (preview/frozen).

Commit 4 - Glyphs Overlay Implementation (2 options for glyph position)

**(preferred)**

glyphs_over

glyphs_under

Commit 3
node_color_overlay
Commit 2
frozen_glyph_implementation
Commit 1
glyph_fade_out

### Declarations

### Release Notes

Updates the visual-cue glyph behavior when zoomed-in/out.

### Reviewers

@sm6srw 

### FYI
@Amoursol


sm6srw:
Looks good.  Looks like you are close to also enabling the display of the glyphs? Do you want to include that in this PR or in a separate PR. Can you also look into adding a few tests?

dnenov:
Thanks, Jorgen. Only the .xaml part left more or less, let's do it in the same PR. 

I'd love to add tests, but I have two main problems in that regard - a conceptual and a knowledge-based one. I started reading through the existing tests and getting a holistic enough view of the code base to allow me to write a test looks .. a bit intimidating at the moment. On the conceptual level, I would appreciate your support on what should be a well-placed test? I think I only have 1 method so far I'd like tested, but I am not sure if I am right or wrong. 

I will finish up the code and before I get on with the tests, I will maybe chase you up for a quick chat on Slack?

sm6srw:
Great, sounds like a plan. Ping me when you want to chat.

sm6srw (commented):
Looks great! Do you feel like we have enough test coverage or do you think we need some tests that checks the properties on the node model?

src/DynamoCoreWpf/Views/Core/NodeView.xaml
```diff
@@ -535,18 +549,90 @@
...
                 Grid.Row="1"
                 Grid.RowSpan="4"
                 Grid.ColumnSpan="3"
                 Margin="-8"
                 Background="{StaticResource NodeFrozenOverlayColor}"
-                Canvas.ZIndex="7"
+                Canvas.ZIndex="6"
+                CornerRadius="8,8,0,0"
+                IsHitTestVisible="False"
+                Opacity="{Binding Path=DataContext.Zoom,
```
mjkkirschner, line 562: can you verify that in addition to lowering the opacity if the zoom level is high, that you are also making the item collapsed? This should remove the elements from the visual tree, avoiding layout and render costs.

test/DynamoCoreWpfTests/NodeViewCustomizationTests.cs
```diff
@@ -314,9 +314,11 @@ public void WatchImageCoreContainsImage()
 
             // Starting from Dynamo 2.13, node view now comes with 
             // images like node icon, lacing image etc
-            Assert.AreEqual(3, imgs.Count());
+            // As of March 2022, we have 7 images per NodeView
+            // Images are named for ease of use
+            Assert.AreEqual(7, imgs.Count());
```
mjkkirschner, line 319: any noticeable increase in memory use?

src/DynamoCoreWpf/Views/Core/NodeView.xaml
dnenov: Sure thing, will do!

dnenov: I have introduced the changes, let me know if that hasn't translated into the desired increase in performance.

test/DynamoCoreWpfTests/NodeViewCustomizationTests.cs
sm6srw: I did a high level comparison watching the process memory in use with and without this change. My use case was to display 792 nodes. Without the change the average was ~1.23 Gb over three runs. ~1.24 Gb over three runs with the change. Looks OK to me.

dnenov: Thank you, Jorgen!

PR #12679, opened 2022-03-08
Do not manually remove the obj folder before Cleanup

pinzart90:
Let cleanup of the obj folder happen automatically

### Release Notes

### Reviewers

### FYIs

### DIFF
--- a/src/Config/CS_SDK.props
+++ b/src/Config/CS_SDK.props
@@ -63,10 +63,6 @@
   <Target Name="CleanIntermediaryFolder" Condition="Exists('$(ProjectDir)$(BaseIntermediateOutputPath)')" BeforeTargets="Clean" >
      <RemoveDir ContinueOnError="WarnAndContinue" Directories="$(ProjectDir)$(BaseIntermediateOutputPath)" /> <!-- int -->
   </Target>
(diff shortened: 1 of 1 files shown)


src/Config/CS_SDK.props
```diff
@@ -63,10 +63,6 @@
   <Target Name="CleanIntermediaryFolder" Condition="Exists('$(ProjectDir)$(BaseIntermediateOutputPath)')" BeforeTargets="Clean" >
      <RemoveDir ContinueOnError="WarnAndContinue" Directories="$(ProjectDir)$(BaseIntermediateOutputPath)" /> <!-- int -->
   </Target>
-  <Target Name="CleanObjFolder" Condition="Exists('$(ProjectDir)obj')" BeforeTargets="Clean">
-	 <RemoveDir ContinueOnError="WarnAndContinue" Directories="$(ProjectDir)obj" />
-	 <!-- obj -->
-  </Target>
```
pinzart90, line 69: 1. Intellisense uses the contents of the obj folder...so it is not a good if this folder is missing
2. The issue we are seeing happens because the obj folder is created when nuget restores the project packages...but the Clean Task is called after that.
We could, in theory, fix it by adding the proper BeforeTargets to also depend on the Nuget restore Target. But I could not find any documentation on it...so I opted to stop doing cleanup manually.
Cleanup is already being done automatically (although I am not sure how well)

PR #12080, opened 2021-09-28
Node Informational States

OliverEGreen:
### Purpose

This PR adds in the node informational states functionality.

It has been carried over from these PRs: #11997 and #12011

It reponds to this issue:
[DYN-3960] https://jira.autodesk.com/browse/DYN-3960

### Declarations

Check these if you believe they are true

- [x] The codebase is in a better state after this PR
- [x] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [x] The level of testing this PR includes is appropriate
- [x] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.
- [ ] This PR modifies some build requirements and the readme is updated

### Reviewers

@QilongTang 

### FYIs

@mjkkirschner

PR #12682, opened 2022-03-10
DYN-4582, DYN-4557: Fix warning issues with Code block nodes

aparajit-pratap:
### Purpose

- DYN-4582: All compile-time and runtime warnings are now displayed on a CBN as multiple unique warnings upon execution
- Copy-pasting a CBN with warnings correctly shows warnings on the new node
- DYN-4557 - Obsolete node warnings are displayed as expected
- Multiple warnings and errors are displayed as expected
- Treat dummy node warnings as persistent warnings

### Declarations

### Release Notes

All compile-time and runtime warnings are now displayed on a CBN correctly as multiple unique warnings 

### FYIs

### DIFF
--- a/src/DynamoCore/Graph/Nodes/NodeModel.cs
+++ b/src/DynamoCore/Graph/Nodes/NodeModel.cs
@@ -1594,7 +1594,7 @@ namespace Dynamo.Graph.Nodes
         private void ClearTooltipText()
         {
             ToolTipText = "";
-            infos.RemoveWhere(x => x.State == ElementState.Warning);
+            infos.RemoveWhere(x => x.State == ElementState.Warning || x.State == ElementState.Error);
         }
 
         private void ClearPersistentWarning()
@@ -1718,8 +1718,9 @@ namespace Dynamo.Graph.Nodes
         public void Error(string p)
         {
             State = ElementState.Error;
-            ToolTipText = p;
             infos.Add(new Info(p, ElementState.Error));
+
+            ToolTipText = p;
         }
 
         /// <summary>
--- a/src/DynamoCore/Graph/Nodes/NodeModel.cs
+++ b/src/DynamoCore/Graph/Nodes/NodeModel.cs
@@ -1625,23 +1625,38 @@ namespace Dynamo.Graph.Nodes
         /// </summary>
         internal void ClearTransientWarning(string t = null)
         {
-            if (State == ElementState.Warning)
+            if (State != ElementState.Warning) return;
+
+            bool cond = false;
+            infos.RemoveWhere(x =>
             {
-                if (string.IsNullOrEmpty(t) || infos.Any(x => x.Message == t && x.State == ElementState.Warning))
-                {// Called with null (or empty argument) or argument matched the existing transient warning
-                    infos.RemoveWhere(x => x.State == ElementState.Warning);
-                    if (!string.IsNullOrEmpty(persistentWarning))
-                    {// Still have persistent warnings then switch to the PersistentWarning state
-                        State = ElementState.PersistentWarning;
-                        ToolTipText = persistentWarning;
-                        infos.Add(new Info(persistentWarning, ElementState.PersistentWarning));
+                if (string.IsNullOrEmpty(t) && x.State == ElementState.Warning)
+                {
+                    cond = true;
+                }
+                else if (!string.IsNullOrEmpty(t))
+                {
+                    if (x.Message == t && x.State == ElementState.Warning)
+                    {
+                        cond = true;
                     }
-                    else
-                    {// No persistent warnings, go to default
-                        State = ElementState.Dead;
-                        ClearErrorsAndWarnings();
-                    }
-                } 
+                }
+                return cond;
+            });
+            if (cond)
+            {
+                if (!string.IsNullOrEmpty(persistentWarning))
+                {
+                    // Still have persistent warnings then switch to the PersistentWarning state
+                    State = ElementState.PersistentWarning;
+                    ToolTipText = persistentWarning;
+                }
+                else
+                {
+                    // No persistent warnings, go to default
+                    State = ElementState.Dead;
+                    ClearErrorsAndWarnings();
+                }
             }
         }
 
--- a/src/DynamoCore/Graph/Nodes/NodeModel.cs
+++ b/src/DynamoCore/Graph/Nodes/NodeModel.cs
@@ -29,6 +29,17 @@ using StringNode = ProtoCore.AST.AssociativeAST.StringNode;
 
 namespace Dynamo.Graph.Nodes
 {
+    internal struct Info
+    {
+        public string Message;
+        public Nodes.ElementState State;
+        internal Info(string message, ElementState state = ElementState.Active)
+        {
+            Message = message;
+            State =  state;
+        }
+    }
+
     public abstract class NodeModel : ModelBase, IRenderPackageSource<NodeModel>, IDisposable
     {
         #region private members
@@ -42,12 +53,9 @@ namespace Dynamo.Graph.Nodes
         private string name;
         private ElementState state;
         private string toolTipText = string.Empty;
+        private HashSet<Info> infos = new HashSet<Info>();
         private string description;
         private string persistentWarning = string.Empty;
-        private string transientWarning = string.Empty;
-
-        private bool areInputPortsRegistered;
-        private bool areOutputPortsRegistered;
 
         ///A flag indicating whether the node has been explicitly frozen.
         internal bool isFrozenExplicitly;
@@ -366,6 +374,14 @@ namespace Dynamo.Graph.Nodes
                 toolTipText = value;
                 RaisePropertyChanged(nameof(ToolTipText));
             }
+        }
+
+        /// <summary>
+        /// Collection of warnings, errors and info items applied to the NodeModel.
+        /// </summary>
+        internal HashSet<Info> Infos
+        {
+            get { return infos; }
         }
 
         /// <summary>
@@ -1578,11 +1594,13 @@ namespace Dynamo.Graph.Nodes
         private void ClearTooltipText()
         {
             ToolTipText = "";
+            infos.RemoveWhere(x => x.State == ElementState.Warning);
         }
 
         private void ClearPersistentWarning()
         {
             persistentWarning = String.Empty;
+            infos.RemoveWhere(x => x.State == ElementState.PersistentWarning);
         }
 
         /// <summary>
@@ -1593,7 +1611,6 @@ namespace Dynamo.Graph.Nodes
         {
             State = ElementState.Dead;
             ClearPersistentWarning();
-            transientWarning = string.Empty;
 
             SetNodeStateBasedOnConnectionAndDefaults();
             ClearTooltipText();
@@ -1610,13 +1627,14 @@ namespace Dynamo.Graph.Nodes
         {
             if (State == ElementState.Warning)
             {
-                if (string.IsNullOrEmpty(t) || string.Compare(t, transientWarning) == 0)
+                if (string.IsNullOrEmpty(t) || infos.Any(x => x.Message == t && x.State == ElementState.Warning))
                 {// Called with null (or empty argument) or argument matched the existing transient warning
-                    transientWarning = string.Empty;// Reset the transient warning
+                    infos.RemoveWhere(x => x.State == ElementState.Warning);
                     if (!string.IsNullOrEmpty(persistentWarning))
                     {// Still have persistent warnings then switch to the PersistentWarning state
                         State = ElementState.PersistentWarning;
                         ToolTipText = persistentWarning;
+                        infos.Add(new Info(persistentWarning, ElementState.PersistentWarning));
                     }
                     else
                     {// No persistent warnings, go to default
@@ -1701,6 +1719,7 @@ namespace Dynamo.Graph.Nodes
         {
             State = ElementState.Error;
             ToolTipText = p;
+            infos.Add(new Info(p, ElementState.Error));
         }
 
         /// <summary>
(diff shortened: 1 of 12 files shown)


src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -1717,15 +1717,15 @@ public void Warning(string p, bool isPersistent = false)
                 State = ElementState.PersistentWarning;
                 if (!string.Equals(persistentWarning, p))
                 {
-                    persistentWarning += p;
+                    persistentWarning += persistentWarning == "" ? p : $"{Environment.NewLine}{p}";
```
QilongTang, line 1720: We can probably use the same check here from below: `string.IsNullOrEmpty(persistentWarning)`

QilongTang (approved):
One comment then LGTM

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -1717,15 +1717,15 @@ public void Warning(string p, bool isPersistent = false)
...
                 }
                 ToolTipText = persistentWarning;
             }
             else
             {
                 State = ElementState.Warning;
                 transientWarning = p;
-                ToolTipText = string.IsNullOrEmpty(persistentWarning) ? p : string.IsNullOrEmpty(p) ? persistentWarning : $"{persistentWarning}\n{p}";
+                ToolTipText = string.IsNullOrEmpty(persistentWarning) ? p : string.IsNullOrEmpty(p) ? persistentWarning : $"{persistentWarning}{Environment.NewLine}{p}";
```
pinzart90, line 1728: Probably a larger task...but we should redesign the data here to somthing like a HashSet or a List
I would say there should only be a single Warnings enumerable. Each warning should have a UniqueId, Content, type/category, persistent/transient.
The Tooltip should be a UI property, I think, and it should be generated from the existing Warnings enumerable.

QilongTang: Agreed, some mature design should be in favor. Especially our team discussed per port and node warning system.

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
```diff
@@ -1204,26 +1210,40 @@ public void UpdateBubbleContent()
...
-            if (string.IsNullOrWhiteSpace(content)) return;
             const InfoBubbleViewModel.Direction connectingDirection = InfoBubbleViewModel.Direction.Bottom;
-            var data = new InfoBubbleDataPacket(style, topLeft, botRight, content, connectingDirection);
-
-            ErrorBubble.UpdateContentCommand.Execute(data);
+            var warningTexts = content.Split(new string[] { Environment.NewLine }, StringSplitOptions.None);
+            var packets = new List<InfoBubbleDataPacket>(warningTexts.Length);
+            foreach (var text in warningTexts)
+            {
+                var data = new InfoBubbleDataPacket(style, topLeft, botRight, text, connectingDirection);
```
mjkkirschner, line 1218: so the meat of this change is that the warning text is now separated by newline characters, and we add individual `InfoBubbleDataPacket`'s for each one?
I'm still a bit confused how this fixes the code block issue.

I also am curious if it affects other features that might rely on the warning being a single unseparated string, or might separate it later? For example the documentation browser features that search for help urls in the warnings?

src/DynamoCore/Graph/Nodes/NodeModel.cs
aparajit-pratap: @QilongTang the per port warning idea seems quite hard. I was discussing it with @filipeotero and was looking at say passing 3 nulls to the x, y, z inputs for a Point.ByCoordinates node. The node fails at the first null and bails so it seems unlikely that we can get it to warn for any subsequent invalid inputs.

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
aparajit-pratap: This solution actually doesn't work in all cases, as there could be some warning messages that contain newline characters so I'm looking to see if I could split warnings (InfoBubbleDataPackets) based on the href link, assuming there is a separate link for each warning. @mjkkirschner can you confirm if that's the case?

Also, can you think of a case where a non-CBN node can have more than one warning/error so that I can test it? I'm assuming that that will be a requirement going forwards with the info state thing.

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -847,6 +847,13 @@ public void Run()
                 // currently fire differently from one another. 
                 foreach (NodeModel nodeModel in task.ModifiedNodes)
                 {
+                    // CodeBlockNodes are unique in that they can exhibit two types of warnings and errors - 
+                    // those that are found at precompilation (compile time errors and warnings),
+                    // and those that are found at execution time (runtime errors and warnings). In the case
+                    // of CBNs therefore, we need to retain errors and warnings from both stages, which is why we cannot 
+                    // clear them here (before execution) as this will wipe out those from the precompilation stage.
+                    if (nodeModel is CodeBlockNodeModel) continue;
```
aparajit-pratap, line 855: @mjkkirschner this is the actual fix for the bug where the warning didn't show up on the node after executing it even though the graph said `Run completed with warnings.`

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
mjkkirschner: * only some warnings have href links, those that do, only have 1.
* a ZT node could have multiple errors potentially in the following ways (unconfirmed)
    * replication yielding different errors for different inputs.
    * Use of a method like https://github.com/DynamoDS/Dynamo/blob/master/src/NodeServices/MessageEvents.cs#L20 `OnLogWarningMessage` to raise a warning, and also a real warning encountered during execution of the node.

aparajit-pratap: Thanks. In my experience, I've never seen the replication yielding multiple errors show as multiple messages on a node. Can you reproduce that?

aparajit-pratap: In any case, I'm still thinking of a better solution for this.

mjkkirschner: I cannot reproduce it, I may be remembering behavior in 1.x or even earlier. I just tried back in 2.6 and cannot reproduce there, so let's chalk it up to remembering incorrectly.

Though... should that be the behavior? It does not seem correct.  Even if the evaluations yield different errors we only return the first.

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -42,12 +53,9 @@ public abstract class NodeModel : ModelBase, IRenderPackageSource<NodeModel>, ID
         private string name;
         private ElementState state;
         private string toolTipText = string.Empty;
+        private HashSet<Info> infos = new HashSet<Info>();
         private string description;
         private string persistentWarning = string.Empty;
```
pinzart90, line 58: do we still need `persistentWarning` as a separate property ?
Or can we look it/them up in the new `infos` hashset ?

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
QilongTang: After a lot of debugging, I found actually this `nodeModel.OnNodeMessagesClearing();` here will clear node warnings include persistent warnings as well... Still need to look at why that changed in 2.13, but I feel persistent warning is another case here we should not clear after execution, otherwise they simply get lost. Do you have any memory on that end?

aparajit-pratap: @QilongTang do you mean persistent warnings that are related to obsolete nodes also need to be excluded from clearing over here? Will check that.

QilongTang: It's fine @aparajit-pratap Not meant to put all bugs on you. If this PR is good to merge, we can always follow up

src/DynamoCore/Graph/Nodes/NodeModel.cs
QilongTang: I guess previously, warning bubble always fallback to this so persistent warnings can be isolated from execution warningsos for display

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
aparajit-pratap: Thanks for pointing this out @QilongTang, I actually think I have a better solution now that treats all persistent warning nodes consistently so I don't have to call out CBN's specifically. Will send an update after some more testing.

src/Libraries/CoreNodeModelsWpf/NodeViewCustomizations/DummyNode.cs
```diff
@@ -29,7 +29,7 @@ public void CustomizeView(DummyNode model, Dynamo.Controls.NodeView nodeView)
             RenderOptions.SetBitmapScalingMode(dummyNodeImage, BitmapScalingMode.HighQuality);
 
             nodeView.inputGrid.Children.Add(dummyNodeImage);
-            model.Warning(model.GetDescription());
+            model.Warning(model.GetDescription(), true);
```
aparajit-pratap, line 32: Make dummy node (unresolved node) warnings persistent over here to treat this type consistently with other types of persistent warnings.

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
```diff
@@ -1190,40 +1195,66 @@ public void UpdateBubbleContent()
...
             var botRight = new Point(NodeModel.X + NodeModel.Width, NodeModel.Y + NodeModel.Height);
 
+            const InfoBubbleViewModel.Direction connectingDirection = InfoBubbleViewModel.Direction.Bottom;
+            var packets = new List<InfoBubbleDataPacket>(NodeModel.Infos.Count);
+            foreach (var info in NodeModel.Infos)
+            {
+                var infoStyle = info.State == ElementState.Error ? InfoBubbleViewModel.Style.Error : InfoBubbleViewModel.Style.Warning;
+                var data = new InfoBubbleDataPacket(infoStyle, topLeft, botRight, info.Message, connectingDirection);
+                packets.Add(data);
+                //ErrorBubble.UpdateContentCommand.Execute(data);
```
QilongTang, line 1225: should we just remove this? I also found that we update bubble content below anyway it seems

QilongTang (commented):
Looks solid

aparajit-pratap:
@QilongTang I'll try to merge this today. I'm still testing some edge cases where I see strange behavior. I also need to fix the CI test regressions. Will update you.

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -842,14 +842,6 @@ public void Run()
                 // The workspace has been built for the first time
                 silenceNodeModifications = false;
 
-                // An event handler we can listen to on the NodeViewModel, to update the 
-                // node's informational state. This is required because Errors and Warnings 
-                // currently fire differently from one another. 
-                foreach (NodeModel nodeModel in task.ModifiedNodes)
-                {
-                    nodeModel.OnNodeMessagesClearing();
-                }
```
aparajit-pratap, line 851: This  was recently added in #12080 but doesn't seem necessary as `OnNodeMessagesClearing` is being called in `ClearErrorsAndWarnings` which keeps the bubble messages in sync with the info states on the node anyway.

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
```diff
@@ -1190,40 +1194,64 @@ public void UpdateBubbleContent()
...
             if (DynamoViewModel == null) return;
 
-            bool hasErrorOrWarning = NodeModel.IsInErrorState || NodeModel.State == ElementState.Warning || NodeModel.State == ElementState.PersistentWarning;
+            bool hasErrorOrWarning = NodeModel.IsInErrorState || NodeModel.State == ElementState.Warning;
+
+            // Persistent warnings should not be dismissed even if nodes are not involved in an execution as they can include:
+            // 1. Compile-time warnings in CBNs
+            // 2. Obsolete nodes with warnings
+            // 3. Dummy or unresolved nodes
+            if (NodeModel.State != ElementState.PersistentWarning && !NodeModel.IsInErrorState)
```
aparajit-pratap, line 1203: This condition is now consistent with the one [above](https://github.com/DynamoDS/Dynamo/pull/12682/files#diff-97f7c391a3f292149633414ed4ac50ce1a73a1439bcc3307d20da0fb043a6d78R836).

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -1578,11 +1594,13 @@ internal void DisconnectOutput(int portData, int inPortData, NodeModel nodeModel
         private void ClearTooltipText()
         {
             ToolTipText = "";
+            infos.RemoveWhere(x => x.State == ElementState.Warning || x.State == ElementState.Error);
```
pinzart90, line 1597: so now the clearing the tooltip clears all the warnings/errors ?
With a little refactor I think we can get rid of the Tooltip from the Model and move it/replace with the new InfoBubble system

QilongTang: @aparajit-pratap @pinzart90 This PR is getting bigger and bigger. Any chance we can merge and do follow up PR or they better be done together? I am asking since our team have some dependency on this

aparajit-pratap: @QilongTang ok, I'll merge this as-is and make a follow-up PR.

QilongTang: @aparajit-pratap @pinzart90 Appreciate it! Sorry for the noise, just would like to unblock some work on our end so we can work in parallel on them. Cheers.

pinzart90 (approved):

PR #12700, opened 2022-03-15
Add node info to Integer slider node

reddyashish:
### Purpose

This PR is a first prototype for the node info messages. https://jira.autodesk.com/browse/DYN-4543. This will be done for more nodes as per the needs.

The example here is for IntegerSlider node, which will show an info message when the value entered in the node is not in the int64 range. We will be using the same UpdateBubbleContent() to add the info message. 

node info integer slider

### Declarations

### Release Notes

Add node info to Integer slider node

### Reviewers

@QilongTang @aparajit-pratap @zeusongit

### DIFF
--- a/src/DynamoCore/Graph/Nodes/NodeModel.cs
+++ b/src/DynamoCore/Graph/Nodes/NodeModel.cs
@@ -1639,7 +1639,6 @@ namespace Dynamo.Graph.Nodes
         {
             infos.RemoveWhere(x => x.State == ElementState.Info);
             State = ElementState.Dead;
-            ClearTooltipText();
             OnNodeInfoMessagesClearing();
         }
 
--- a/src/DynamoCore/Graph/Nodes/NodeModel.cs
+++ b/src/DynamoCore/Graph/Nodes/NodeModel.cs
@@ -1638,7 +1638,6 @@ namespace Dynamo.Graph.Nodes
         public virtual void ClearInfoMessages()
         {
             infos.RemoveWhere(x => x.State == ElementState.Info);
-            State = ElementState.Dead;
             OnNodeInfoMessagesClearing();
         }
 
@@ -1771,8 +1770,6 @@ namespace Dynamo.Graph.Nodes
         {
             State = ElementState.Info;
             infos.Add(new Info(p, ElementState.Info));
-
-            ToolTipText = p;
         }
 
         /// <summary>
--- a/src/DynamoCore/Graph/Nodes/NodeModel.cs
+++ b/src/DynamoCore/Graph/Nodes/NodeModel.cs
@@ -166,6 +166,20 @@ namespace Dynamo.Graph.Nodes
             NodeMessagesClearing?.Invoke(this);
         }
 
+        /// <summary>
+        /// Event triggered whenever the node re-executes to clear its info messages
+        /// </summary>
+        public event Action<NodeModel> NodeInfoMessagesClearing;
+
+        /// <summary>
+        /// Fires on each node that is modified to clear info messages, when the graph executes.
+        /// </summary>
+        internal void OnNodeInfoMessagesClearing()
+        {
+            NodeInfoMessagesClearing?.Invoke(this);
+        }
+
+
         internal void OnNodeExecutionBegin()
         {
             NodeExecutionBegin?.Invoke(this);
@@ -332,7 +346,7 @@ namespace Dynamo.Graph.Nodes
             get { return state; }
             set
             {
-                if (value != ElementState.Error && value != ElementState.AstBuildBroken)
+                if (value != ElementState.Error && value != ElementState.Info && value != ElementState.AstBuildBroken)
                     ClearTooltipText();
 
                 // Check before settings and raising
@@ -1593,7 +1607,7 @@ namespace Dynamo.Graph.Nodes
 
         private void ClearTooltipText()
         {
-            ToolTipText = "";
+            ToolTipText = string.Empty;
             infos.RemoveWhere(x => x.State == ElementState.Warning || x.State == ElementState.Error);
         }
 
@@ -1615,6 +1629,18 @@ namespace Dynamo.Graph.Nodes
             SetNodeStateBasedOnConnectionAndDefaults();
             ClearTooltipText();
             OnNodeMessagesClearing();
+        }
+
+        /// <summary>
+        /// Clears the info messages that are generated when running the graph,
+        /// the State will be set to ElementState.Dead.
+        /// </summary>
+        public virtual void ClearInfoMessages()
+        {
+            infos.RemoveWhere(x => x.State == ElementState.Info);
+            State = ElementState.Dead;
+            ClearTooltipText();
+            OnNodeInfoMessagesClearing();
         }
 
         /// <summary>
@@ -1734,6 +1760,18 @@ namespace Dynamo.Graph.Nodes
         {
             State = ElementState.Error;
             infos.Add(new Info(p, ElementState.Error));
+
+            ToolTipText = p;
+        }
+
+        /// <summary>
+        /// Set an info on a node.
+        /// </summary>
+        /// <param name="p">The info text.</param>
+        public void Info(string p)
+        {
+            State = ElementState.Info;
+            infos.Add(new Info(p, ElementState.Info));
 
             ToolTipText = p;
         }
@@ -2758,7 +2796,8 @@ namespace Dynamo.Graph.Nodes
         Warning,
         PersistentWarning,
         Error,
-        AstBuildBroken
+        AstBuildBroken,
+        Info
     };
 
     /// <summary>
--- a/src/DynamoCore/Graph/Nodes/NodeModel.cs
+++ b/src/DynamoCore/Graph/Nodes/NodeModel.cs
@@ -179,7 +179,6 @@ namespace Dynamo.Graph.Nodes
             NodeInfoMessagesClearing?.Invoke(this);
         }
 
-
         internal void OnNodeExecutionBegin()
         {
             NodeExecutionBegin?.Invoke(this);
--- a/src/DynamoCore/Graph/Nodes/NodeModel.cs
+++ b/src/DynamoCore/Graph/Nodes/NodeModel.cs
@@ -1643,8 +1643,7 @@ namespace Dynamo.Graph.Nodes
         }
 
         /// <summary>
-        /// Clears the errors/warnings that are generated when running the graph,
-        /// the State will be set to ElementState.Dead.
+        /// Clears the errors/warnings that are generated when running the graph
         /// </summary>
         public virtual void ClearErrorsAndWarnings()
         {
--- a/src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
@@ -803,6 +803,7 @@ namespace Dynamo.ViewModels
             }
             
             logic.NodeMessagesClearing += Logic_NodeMessagesClearing;
+            logic.NodeInfoMessagesClearing += Logic_NodeInfoMessagesClearing;
 
             logic_PropertyChanged(this, new PropertyChangedEventArgs(nameof(IsVisible)));
         }
@@ -848,6 +849,34 @@ namespace Dynamo.ViewModels
             }
         }
 
+        /// <summary>
+        /// Clears the existing messages on a node before it executes and re-evalutes its warnings/errors. 
+        /// </summary>
+        /// <param name="obj"></param>
+        private void Logic_NodeInfoMessagesClearing(NodeModel obj)
+        {
+            if (DynamoViewModel.UIDispatcher != null)
+            {
+                DynamoViewModel.UIDispatcher.Invoke(() =>
+                {
+                    var itemsToRemove = ErrorBubble.NodeMessages.Where(x => x.Style == InfoBubbleViewModel.Style.Info).ToList();
+
+                    foreach (var itemToRemove in itemsToRemove)
+                    {
+                        ErrorBubble.NodeMessages.Remove(itemToRemove);
+                    }
+                });
+            }
+            else
+            {
+                var itemsToRemove = ErrorBubble.NodeMessages.Where(x => x.Style == InfoBubbleViewModel.Style.Info).ToList();
+                foreach (var itemToRemove in itemsToRemove)
+                {
+                    ErrorBubble.NodeMessages.Remove(itemToRemove);
+                }
+            }
+        }
+
         private void DismissedNodeMessages_CollectionChanged(object sender, NotifyCollectionChangedEventArgs e)
         {
             if (!(sender is ObservableCollection<InfoBubbleDataPacket> observableCollection)) return;
@@ -890,6 +919,7 @@ namespace Dynamo.ViewModels
             }
 
             NodeModel.NodeMessagesClearing -= Logic_NodeMessagesClearing;
+            NodeModel.NodeInfoMessagesClearing -= Logic_NodeInfoMessagesClearing;
             
             if (ErrorBubble != null) DisposeErrorBubble();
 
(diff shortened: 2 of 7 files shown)


src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
```diff
@@ -1204,9 +1207,16 @@ public void UpdateBubbleContent()
...
                 : InfoBubbleViewModel.Style.Warning;
 
+            style = NodeModel.State == ElementState.Info ? InfoBubbleViewModel.Style.Info : style;
+
             // NOTE!: If tooltip is not cached here, it will be cleared once the dispatcher is invoked below
             string content = NodeModel.ToolTipText;
-            if (string.IsNullOrWhiteSpace(content)) return;
+            if (string.IsNullOrWhiteSpace(content))
+            {
+                ErrorBubble.NodeMessages.Clear();
```
reddyashish, line 1216: When the tooltip message is empty or null, we need to clear the previous node messages, otherwise they will show up even the tooltip message is empty. Do you think this will change any previous behavior?

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -1617,6 +1631,18 @@ public virtual void ClearErrorsAndWarnings()
...
+        /// <summary>
+        /// Clears the info messages that are generated when running the graph,
+        /// the State will be set to ElementState.Dead.
+        /// </summary>
+        public virtual void ClearInfoMessages()
+        {
+            infos.RemoveWhere(x => x.State == ElementState.Info);
+            State = ElementState.Dead;
+            ClearTooltipText();
+            OnNodeInfoMessagesClearing();
```
reddyashish, line 1643: I have created a new method OnNodeInfoMessagesClearing to clear the info messages, instead of using ClearErrorsAndWarnings. The reason being, when we want to clear the info message for a node in future, it should not clear any current errors/warnings for that node.

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -1617,6 +1631,18 @@ public virtual void ClearErrorsAndWarnings()
...
 
+        /// <summary>
+        /// Clears the info messages that are generated when running the graph,
+        /// the State will be set to ElementState.Dead.
+        /// </summary>
+        public virtual void ClearInfoMessages()
+        {
+            infos.RemoveWhere(x => x.State == ElementState.Info);
+            State = ElementState.Dead;
+            ClearTooltipText();
```
zeusongit, line 1642: I think this method will be replaced by `ClearTransientWarningsAndErrors` from @aparajit-pratap 's [PR](https://github.com/DynamoDS/Dynamo/pull/12712)

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -2758,7 +2796,8 @@ public enum ElementState
         Warning,
         PersistentWarning,
         Error,
-        AstBuildBroken
+        AstBuildBroken,
+        Info
```
zeusongit, line 2800: @QilongTang @Amoursol @reddyashish I was thinking if we could reuse `warning` or `error` for integer overflow notifications, or is there a reason we are going for a new state `Info` here? Just curious.

src/DynamoCore/Graph/Nodes/NodeModel.cs
reddyashish: In the future, if a particular node has any warnings/errors and info messages on it, we don't want to clear the warnings/errors incase we want to just remove the info message. That's why I created a separate method for that use.

zeusongit: Yup, I understand that, I am referring to `ClearTooltipText ` here, that it is being removed

reddyashish: Call ClearTransientWarningsAndErrors instead of ClearTooltipText?

zeusongit: Yes, but after that PR gets merged.

reddyashish: ClearTransientWarningsAndErrors would remove the errors and warnings from infos list, which is not needed.

src/DynamoCore/Graph/Nodes/NodeModel.cs
Amoursol: @zeusongit Error is only used when the thing, such as a graph, or code block will not run (i.e. missing a package so you have dummy nodes). Warning is when something has occured that "may cause your graph to execute in a way you do not expect", but it is up to the end user to determine that. The new "Info" state is purely informational, a kind of FYI.

We added the info state for many reasons, but one strong one is the concern users have around warnings, thinking their graph has failed. This is especially problematic in Dynamo Player (Which surfaces a message stating "Your graph has ran with warnings") as the end users there may have on conception of Dynamo, what a node is, what an error is etc.

zeusongit: Gotcha, thanks 👍🏽

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -1593,7 +1607,7 @@ internal void DisconnectOutput(int portData, int inPortData, NodeModel nodeModel
 
         private void ClearTooltipText()
         {
-            ToolTipText = "";
+            ToolTipText = string.Empty;
```
aparajit-pratap, line 1610: `ToolTipText `has been deprecated in PR #12712, we are now only using `Infos` as the single source of truth (assuming we merge #12712 before this one of course).

src/DynamoCore/Graph/Nodes/NodeModel.cs
aparajit-pratap: @reddyashish `ClearToolTipText` has been renamed as `ClearTransientWarningsAndErrors`

zeusongit: > ClearTransientWarningsAndErrors would remove the errors and warnings from infos list, which is not needed.

`ClearToolTipText` is executing the same code as `ClearTransientWarningsAndErrors`
Only the name has been updated, as TooltipText property is removed now, so I suggest that we use `ClearTransientWarningsAndErrors` here.

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -1738,6 +1764,18 @@ public void Error(string p)
...
+        /// <summary>
+        /// Set an info on a node.
+        /// </summary>
+        /// <param name="p">The info text.</param>
+        public void Info(string p)
+        {
+            State = ElementState.Info;
+            infos.Add(new Info(p, ElementState.Info));
+
+            ToolTipText = p;
```
aparajit-pratap, line 1776: this is deprecated, it's no longer required to set this.

src/DynamoCore/Graph/Nodes/NodeModel.cs
aparajit-pratap: If this property is serialized to the dyn, it can break deserialization, @mjkkirschner?

src/DynamoCore/Graph/Nodes/NodeModel.cs
aparajit-pratap: Can you use the existing constructor instead of adding a new one?
```
internal Info(string message, ElementState state = ElementState.Active)
        {
            Message = message;
            State =  state;
        }
```

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
```diff
@@ -1219,13 +1252,16 @@ public void UpdateBubbleContent()
             foreach (var info in NodeModel.Infos)
             {
                 var infoStyle = info.State == ElementState.Error ? InfoBubbleViewModel.Style.Error : InfoBubbleViewModel.Style.Warning;
+                infoStyle = info.State == ElementState.Info ? InfoBubbleViewModel.Style.Info : infoStyle;
```
aparajit-pratap, line 1255: I don't think the info state should override the error state ⚠️

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
```diff
@@ -1219,13 +1252,16 @@ public void UpdateBubbleContent()
...
                 var data = new InfoBubbleDataPacket(infoStyle, topLeft, botRight, info.Message, connectingDirection);
                 packets.Add(data);
             }
             InfoBubbleViewModel.Style style = NodeModel.State == ElementState.Error
                 ? InfoBubbleViewModel.Style.Error
                 : InfoBubbleViewModel.Style.Warning;
 
+            style = NodeModel.State == ElementState.Info ? InfoBubbleViewModel.Style.Info : style;
```
aparajit-pratap, line 1263: Same here, the info state cannot override the error state. I think the hierarchy of states should be:
1. Error 2. Warning 3. Info
@Amoursol please confirm.

src/Libraries/CoreNodeModels/Input/IntegerSlider.cs
```diff
@@ -334,6 +335,18 @@ protected override bool UpdateValueCore(UpdateValueParams updateValueParams)
...
 
+        private void UpdateNodeInfo(string value)
+        {
+            if (IsValueInt64(value))
+            {
+                ClearInfoMessages();
+            }
+            else
+            {
+                Info(Resources.IntegerSliderInfoMessage);
```
aparajit-pratap, line 346: This should be changed to:
```
this.Infos.Add(new Info(Resources.IntegerSliderInfoMessage));
```

src/DynamoCore/Graph/Nodes/NodeModel.cs
aparajit-pratap: @zeusongit I think @reddyashish intends to _only_ clear info messages in this method, not other errors and warnings. If that's the case, `ClearTransientWarningsAndErrors` (`ClearToolTipText`) should _not_ be called.

aparajit-pratap: Also, why is the node's state being set to `ElementState.Dead`?

aparajit-pratap (changes requested):
Please make the necessary changes that are called out.

src/DynamoCore/Graph/Nodes/NodeModel.cs
zeusongit: Yes, in that case warnings and errors should not be removed, I think the misleading method name `ClearToolTipText ` may have created this confusion.

reddyashish: Removed the ClearToolTipText() to avoid confusion.

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
Amoursol: Correct @aparajit-pratap, the hierarchy of the graphical representation is as follows when **collapsed:** Error > Warning > Info

Beyond that, they should all exist there in a stack when **uncollapsed,** so when you hover over that hierarchy bases graphical representation you still see anything else that exists. Refer to the Figma here: https://www.figma.com/file/UIDVMHnpK4TL6gm1U7QKEh/Dynamo_Redesign%3A-Alerts?node-id=0%3A1

src/DynamoCore/Graph/Nodes/NodeModel.cs
reddyashish: @aparajit-pratap Removed the statement where it is being set to ElementState.Dead.

src/DynamoCore/Graph/Nodes/NodeModel.cs
reddyashish: Removed this statement.

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
reddyashish: Modified this to follow the 1.Error 2. Warning 3. Info hierarchy.

src/DynamoCore/Graph/Nodes/NodeModel.cs
reddyashish: I have added the Info() to NodeModel class to be consistent with the Error() and Warning(). I thought it would be easy for developers to just call that method on NodeModel class. What do you think?

I have removed the following statement.

> ToolTipText = p;

src/Libraries/CoreNodeModels/Input/IntegerSlider.cs
reddyashish: Created this Info() as we have a similar Error() and Warning() on NodeModel object.

reddyashish:
@aparajit-pratap Have addressed the comments and replied to why I created the Info(). Let me know your thoughts.

src/DynamoCore/Graph/Nodes/NodeModel.cs
reddyashish: Will this be break? I think reordering the elements in a enum will be a break. Not sure.

src/DynamoCore/Graph/Nodes/NodeModel.cs
QilongTang: I agree

QilongTang (approved):

QilongTang:
Just looking for a unit test or so, otherwise LGTM

src/DynamoCore/Graph/Nodes/NodeModel.cs
mjkkirschner: I don't think we serialize ElementState on the nodeModel - simply adding a new value to the enum is not an issue if we don't actually serialize and deserialize (as long as that enum is added to the end of the enum - no reordering allowed)

so in conclusion - 

make sure NodeModel.State is not serialized - or property that is of type `ElementState`

mjkkirschner:
@QilongTang @reddyashish  not so sure the latest commits are safe to merge - none of them has passed on master-5 they are timing out...

It could of course be the test machine - please check the machine / check for memory leaks or perf issues.

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
```diff
@@ -855,6 +856,36 @@ private void Logic_NodeMessagesClearing(NodeModel obj)
...
+        /// <param name="obj"></param>
+        private void Logic_NodeInfoMessagesClearing(NodeModel obj)
+        {
+            if (ErrorBubble == null) return;
+
+            if (DynamoViewModel.UIDispatcher != null)
+            {
+                DynamoViewModel.UIDispatcher.Invoke(() =>
+                {
+                    var itemsToRemove = ErrorBubble.NodeMessages.Where(x => x.Style == InfoBubbleViewModel.Style.Info).ToList();
```
aparajit-pratap, line 871: Is the ToList required here and below?

aparajit-pratap (approved):
@reddyashish 2 minor comments, then LGTM!

src/DynamoCoreWpf/ViewModels/Core/NodeViewModel.cs
reddyashish: Addressed this in the new PR: https://github.com/DynamoDS/Dynamo/pull/12734

reddyashish:
@QilongTang Added a test to the https://github.com/DynamoDS/Dynamo/pull/12734

reddyashish:
Closing this PR and will the merging the alternate one.

src/DynamoCore/Graph/Nodes/NodeModel.cs
reddyashish: Will be addressing this in a followup task.

QilongTang:
> @QilongTang Added a test to the #12734

Thanks!

PR #12711, opened 2022-03-17
DYN-4709-GroupStyles-HideDeleteButton

RobertGlobant20:
### Purpose

Fix for hiding the groupstyle delete button when is a default style otherwise will be visible.
I added a Visibility Binding to the TrashDelete button ( with the IsDefault property) then I added a Converter from bool to visibility, in this way if the IsDefault is true then the button will be Collapsed(not visible) otherwise will be Visible.

### Declarations

### Release Notes

Fix for hiding the groupstyle delete button when is a default style otherwise will be visible.

### Reviewers

@QilongTang

### DIFF
--- a/src/DynamoCoreWpf/UI/Converters.cs
+++ b/src/DynamoCoreWpf/UI/Converters.cs
@@ -3327,4 +3327,25 @@ namespace Dynamo.Controls
 
         #endregion
     }
+
+    /// <summary>
+    /// Converts a boolean to Visibility but inverted, then false will be Visible and true will be Collapsed
+    /// </summary>
+    [ValueConversion(typeof(bool), typeof(Visibility))]
+    public class InvertedBooleanToVisibilityConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType,
+                              object parameter, CultureInfo culture)
+        {
+            var boolValue = (bool)value;
+
+            return boolValue ? Visibility.Collapsed : Visibility.Visible;
+        }
+
+        public object ConvertBack(object value, Type targetType,
+            object parameter, CultureInfo culture)
+        {
+            return null;
+        }
+    }
 }
--- a/src/DynamoCoreWpf/UI/Converters.cs
+++ b/src/DynamoCoreWpf/UI/Converters.cs
@@ -3327,25 +3327,4 @@ namespace Dynamo.Controls
 
         #endregion
     }
-
-    /// <summary>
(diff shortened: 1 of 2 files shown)


src/DynamoCoreWpf/UI/Converters.cs
```diff
@@ -3327,4 +3327,25 @@ public object ConvertBack(object value, Type targetType, object parameter, Syste
 
         #endregion
     }
+
+    /// <summary>
+    /// Converts a boolean to Visibility but inverted, then false will be Visible and true will be Collapsed
+    /// </summary>
+    [ValueConversion(typeof(bool), typeof(Visibility))]
+    public class InvertedBooleanToVisibilityConverter : IValueConverter
```
QilongTang, line 3335: I thought there is already a `BoolToVisibilityCollapsedConverter`? I dont think we need to create a new one here

RobertGlobant20: @QilongTang for this case I need a converted that that does the inverse operation, in the BoolToVisibilityCollapsedConverter if the value is **true** returns **Visible** and if is **false** returns **Collapsed**, for this case is the opposite, in the InvertedBooleanToVisibilityConverter  if is **true** then returns **Collapsed** (meaning that property IsDefault = true so the button should be collapsed) otherwise returns **Visible**.
Please let me know what you think about it.

QilongTang: Then how about using `InverseBoolToVisibilityConverter`

zeusongit: `InverseBoolToVisibilityCollapsedConverter ` sounds good

RobertGlobant20: @QilongTang I removed the converter that I added due that there was one already existing doing the same functionality.
https://github.com/DynamoDS/Dynamo/pull/12711/commits/73d3cadf7a8143c56ff8113f7fe75b1fe015061c

QilongTang:
Thank you @RobertGlobant20 Would you attach a screenshot?

RobertGlobant20:
> Thank you @RobertGlobant20 Would you attach a screenshot?

@QilongTang here is a GIF after deleting the Converter that I added.
groupstyles_deletebutton_disabled2

QilongTang (approved):

PR #12712, opened 2022-03-18
Deprecate NodeModel.ToolTipText in favor of NodeModel.Infos

aparajit-pratap:
### Purpose

Deprecate `NodeModel.ToolTipText` in favor of `NodeModel.Infos` - After the introduction of the `Infos` property, `ToolTipText` was redundant. 
- Converted `Infos `from `HashSet` to `ObservableHashSet`, a new type
- Removed `persistentWarning` field from `NodeModel`

### Declarations

### Release Notes

Deprecate `NodeModel.ToolTipText`.

### FYIs


src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -38,6 +38,25 @@ internal Info(string message, ElementState state = ElementState.Active)
             Message = message;
             State =  state;
         }
+
+        public override bool Equals(object other)
+        {
+            if (other == null) return false;
+
+            var otherInfo = (Info)other;
+            return Message == otherInfo.Message && State == otherInfo.State;
```
mjkkirschner, line 47: null checks?

pinzart90: why do we need a custom override ?
Isn't the default good enough ? I think the default Equals compares all fields via reflection ...Are you concerned about performance ?

aparajit-pratap: I changed it to a class so now we can keep the `Equals` override and perform the `null` checks.

aparajit-pratap:
@pinzart90, @mjkkirschner, I'm not sure if I should stop assigning tooltips to the public `ToolTipText` property for fear of breaking external code that may be referencing it, any thoughts?

src/DynamoUtilities/ObservableHashSet.cs
```diff
@@ -0,0 +1,50 @@
...
+    internal class ObservableHashSet<T> : INotifyCollectionChanged
+    {
+        private readonly HashSet<T> set = new HashSet<T>();
+
+        public event NotifyCollectionChangedEventHandler CollectionChanged;
+
+        public void Add(T item)
+        {
+            set.Add(item);
+            CollectionChanged?.Invoke(this, new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset));
```
pinzart90, line 17: Please use the proper notification types. You are not using them now...but someone might base their code on these types
```
    //
    // Summary:
    //     Describes the action that caused a System.Collections.Specialized.INotifyCollectionChanged.CollectionChanged
    //     event.
    public enum NotifyCollectionChangedAction
    {
        //
        // Summary:
        //     An item was added to the collection.
        Add,
        //
        // Summary:
        //     An item was removed from the collection.
        Remove,
        //
        // Summary:
        //     An item was replaced in the collection.
        Replace,
        //
        // Summary:
        //     An item was moved within the collection.
        Move,
        //
        // Summary:
        //     The contents of the collection changed dramatically.
        Reset
```

pinzart90 (approved):
One comment. then LGTM

src/DynamoUtilities/ObservableHashSet.cs
aparajit-pratap: @pinzart90 this constructor for the `NotifyCollectionChangedEventArgs` requires it to be set to `Reset`. If not, it throws an exception. There are other constructors that allow passing other enum values along with the item that has changed, but since I'm not using that information, I'm using the simplest constructor.
https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.notifycollectionchangedeventargs.-ctor?view=net-6.0#system-collections-specialized-notifycollectionchangedeventargs-ctor(system-collections-specialized-notifycollectionchangedaction)

mjkkirschner:
> @pinzart90, @mjkkirschner, I'm not sure if I should stop assigning tooltips to the public `ToolTipText` property for fear of breaking external code that may be referencing it, any thoughts?

If the tooltip property no longer does something, then setting it probably won't be useful, maybe you can redirect it to add an entry to `Infos`?

You could also wait to expose the infos property until we hear a request that setting tooltips was used - do we have any idea if there are extensions or applications that manually set warnings?

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -53,9 +75,8 @@ public abstract class NodeModel : ModelBase, IRenderPackageSource<NodeModel>, ID
         private string name;
         private ElementState state;
         private string toolTipText = string.Empty;
```
pinzart90, line 77: why do we still need the `private string toolTipText` ?

aparajit-pratap: Removed.

mjkkirschner (approved):

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -366,20 +391,28 @@ public bool IsInErrorState
...
+        [Obsolete("This property is deprecated and will be removed in a future version of Dynamo.")]
         public string ToolTipText
         {
-            get { return toolTipText; }
+            get 
+            {
+                var builder = new System.Text.StringBuilder();
+                foreach(var info in Infos)
+                {
+                    builder.Append(info.ToString() + "\n");
```
pinzart90, line 402: AppendLine ? ...

pinzart90 (approved):

PR #12733, opened 2022-03-23
Mark workspace as modified when performing add/remove operation on Groups

zeusongit:
### Purpose

When a node/group is added to a group mark the workspace as modified.
Added test conditions to a test

### Declarations

### Release Notes

- Mark workspace as modified when performing add/remove operation on Groups

### Reviewers

@DynamoDS/dynamo


src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -1117,7 +1117,6 @@ private void OnModelRemovedFromGroup(object sender, EventArgs e)
 
         private void OnModelAddedToGroup(object sender, EventArgs e)
         {
-            Analytics.TrackEvent(Actions.AddedTo, Categories.GroupOperations, "Group");
```
QilongTang, line 1120: Did you intend to remove this?

zeusongit: Yes, this already tracked at:  https://github.com/DynamoDS/Dynamo/blob/0e16acb5bb0a270976eb2c4db80e8ea45031d4ed/src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs#L497

QilongTang: @zeusongit Would you double check? We have two ways of adding group to group

- Drag into
- Use context menu

I think this is a shared place unlike the other one you mentioned from debugging, maybe keep this one instead?

QilongTang (commented):
One comment then LGTM

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
zeusongit: Ah..sorry I missed that, putting it back. Thank you for catching it.

reddyashish (approved):

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
QilongTang: Hi @zeusongit I checked again, I do think this one is a shared one but the other place you pointed out seems like a duplicate to me: https://github.com/DynamoDS/Dynamo/blob/0e16acb5bb0a270976eb2c4db80e8ea45031d4ed/src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs#L497 because when adding group to group using context menu, both of them get called

zeusongit: I debugged with drag and drop group to group and only this one was called, so I put it pack, may be we shoul remove the other one then

PR #12743, opened 2022-03-28
Dummy Node Remove Color Overlays

dnenov:
### Purpose

A small change to the Dummy Node. The Dummy Node is used for Unresolved or Deprecated Nodes. It will effectively replace any node and has no functionality but to visually represent this state. As such, it does not require the visual candy used in normal nodes, and all the extra glyphs showing Frozen, Hidden, Warning, Error, or Info states were removed. 

### Declarations

### Release Notes

Visual update to the Dummy Node.

dummy node fix - frozen reads
dummy node fix

### Reviewers

@sm6srw 

### FYIs

@Amoursol

### DIFF
--- a/src/Libraries/CoreNodeModelsWpf/NodeViewCustomizations/DummyNode.cs
+++ b/src/Libraries/CoreNodeModelsWpf/NodeViewCustomizations/DummyNode.cs
@@ -34,17 +34,10 @@ namespace CoreNodeModelsWpf.Nodes
             nodeView.inputGrid.Children.Add(dummyNodeImage);
             model.Warning(model.GetDescription(), true);
 
-            UIElement child = nodeView.grid.FindName("zoomGlyphsGrid") as UIElement;
-            nodeView.grid.Children.Remove(child);
-
-            child = nodeView.grid.FindName("nodeColorOverlayZoomIn") as UIElement;
-            nodeView.grid.Children.Remove(child);
-
-            child = nodeView.grid.FindName("nodeBorder") as UIElement;
-            nodeView.grid.Children.Remove(child);
-
-            child = nodeView.grid.FindName("GlyphStackPanel") as UIElement;
-            nodeView.grid.Children.Remove(child);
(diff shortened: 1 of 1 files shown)


Amoursol:
FYI @QilongTang

Amoursol:
@dnenov - Can you just double check that the "Dummy Node" still correctly shows the Frozen state as per the image below when not zoomed out? 

image

src/Libraries/CoreNodeModelsWpf/NodeViewCustomizations/DummyNode.cs
```diff
@@ -24,12 +25,26 @@ public void CustomizeView(DummyNode model, Dynamo.Controls.NodeView nodeView)
...
+            nodeView.grid.Children.Remove(child);
+
+            child = nodeView.grid.FindName("nodeColorOverlayZoomIn") as UIElement;
+            nodeView.grid.Children.Remove(child);
+
+            child = nodeView.grid.FindName("nodeBorder") as UIElement;
+            nodeView.grid.Children.Remove(child);
+
+            child = nodeView.grid.FindName("GlyphStackPanel") as UIElement;
+            nodeView.grid.Children.Remove(child);
```
sm6srw, line 47: I am debating with myself if this is the best way to do this but I think this is OK as long as this is the only node where this kind of exception needs to be added.

dnenov: > I am debating with myself if this is the best way to do this but I think this is OK as long as this is the only node where this kind of exception needs to be added.

I wasn't quite sure if that was the way to go, when I found out how the paperclip icon was implemented I thought it was a bit hacky. 

Initially, I went a different route through the HomeWorkspaceViewModel checking if the node was of type DummyNode to then suppress the glyphs. However, the glyph container visibility already works quite heavily, and I didn't want to overcomplicate it. Furthermore, I was making a lot more changes to the code compared to removing the glyph container only in the DummyNode.

sm6srw (approved):

PR #12763, opened 2022-04-01
Allow pkgs to partially load if some of the assemblies are good

pinzart90:
Please Note:

### Purpose

### Declarations

### Release Notes

### Reviewers

### FYIs

### DIFF
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -1420,9 +1420,18 @@ namespace Dynamo.Models
            // I think this is consistent with the current behavior - IE today - if a nodeModel exists in an assembly, the rest of the assembly 
            // is not imported as ZT - the same will be true if the assembly contains a NodeViewCustomization.
 
-           // TODO(mjk) draw up matrix of current behaviors which nodeLib flag can control.
-            if (!NodeModelAssemblyLoader.ContainsNodeModelSubType(assem)
-                && !(NodeModelAssemblyLoader.ContainsNodeViewCustomizationType(assem)))
+            bool nodeModelOrNodeView;
+            try
+            {
+                nodeModelOrNodeView = !NodeModelAssemblyLoader.ContainsNodeModelSubType(assem)
+                && !(NodeModelAssemblyLoader.ContainsNodeViewCustomizationType(assem));
+            } 
+            catch (Exception ex) {
+                throw new Exceptions.LibraryLoadFailedException(assem.Location, ex.Message);
+            }
+
+            // TODO(mjk) draw up matrix of current behaviors which nodeLib flag can control.
+            if (nodeModelOrNodeView)
             {
                 if (suppressZeroTouchLibraryLoad)
                 {
@@ -1430,7 +1439,7 @@ namespace Dynamo.Models
                 }
                 else
                 {
-                    LibraryServices.ImportLibrary(assem.Location);
+                    LibraryServices.ImportLibrary(assem.Location, false);
                 }
 
                 return;


src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1420,17 +1420,26 @@ internal void LoadNodeLibrary(Assembly assem, bool suppressZeroTouchLibraryLoad
...
+            if (nodeModelOrNodeView)
             {
                 if (suppressZeroTouchLibraryLoad)
                 {
                     LibraryServices.LoadNodeLibrary(assem.Location, false);
                 }
                 else
                 {
-                    LibraryServices.ImportLibrary(assem.Location);
+                    LibraryServices.ImportLibrary(assem.Location, false);
```
pinzart90, line 1442: The default is false. This makes the intention clearer.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1420,17 +1420,26 @@ internal void LoadNodeLibrary(Assembly assem, bool suppressZeroTouchLibraryLoad
...
+            {
+                nodeModelOrNodeView = !NodeModelAssemblyLoader.ContainsNodeModelSubType(assem)
+                && !(NodeModelAssemblyLoader.ContainsNodeViewCustomizationType(assem));
+            } 
+            catch (Exception ex) {
+                throw new Exceptions.LibraryLoadFailedException(assem.Location, ex.Message);
+            }
+
+            // TODO(mjk) draw up matrix of current behaviors which nodeLib flag can control.
+            if (nodeModelOrNodeView)
```
pinzart90, line 1434: there are 2 ways to go about this problem.
This is the one with the least code changes (safest for a hotfix I think).
In this solution we simply throw a specific type of Exception: `LibraryLoadFailedException`
This type of exception is already handled by PackageLoader and is interpreted as an acceptable failure whcih allows the pkg load operation to continue.

Another solution would be to write more code in the PackageLoader to handle exceptions in a smarter way...more changes would be needed

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1420,17 +1420,26 @@ internal void LoadNodeLibrary(Assembly assem, bool suppressZeroTouchLibraryLoad
            // I think this is consistent with the current behavior - IE today - if a nodeModel exists in an assembly, the rest of the assembly 
            // is not imported as ZT - the same will be true if the assembly contains a NodeViewCustomization.
 
-           // TODO(mjk) draw up matrix of current behaviors which nodeLib flag can control.
-            if (!NodeModelAssemblyLoader.ContainsNodeModelSubType(assem)
-                && !(NodeModelAssemblyLoader.ContainsNodeViewCustomizationType(assem)))
+            bool nodeModelOrNodeView;
+            try
+            {
+                nodeModelOrNodeView = !NodeModelAssemblyLoader.ContainsNodeModelSubType(assem)
```
pinzart90, line 1426: These checks (ex `ContainsNodeModelSubType`) actually iterate over all the types in the assembly...so in the case of RhythmRevit.dll....it would throw a ReflectionTypeLoadException exception

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1420,17 +1420,26 @@ internal void LoadNodeLibrary(Assembly assem, bool suppressZeroTouchLibraryLoad
...
+                && !(NodeModelAssemblyLoader.ContainsNodeViewCustomizationType(assem));
+            } 
+            catch (Exception ex) {
+                throw new Exceptions.LibraryLoadFailedException(assem.Location, ex.Message);
```
pinzart90, line 1430: any other type of Exception would disrupt the package loading operation.
Exceptions are managed in the same way inside `LibraryServices.LoadNodeLibrary`

src/DynamoCore/Models/DynamoModel.cs
mjkkirschner, line 1426: shouldn't this variable be name be inverted `notNodeModelOrNodeView`  or `DoesNotContain_NodeModelOrNodeView`

src/DynamoCore/Models/DynamoModel.cs
mjkkirschner: so this solution still relies on enumeration order of the binaries, or was that not actually an issue?

pinzart90: Yes..enumeration is still an issue. That would need to be handled on the Loader side.
I'll add that code to the PR..(I forgot about that part)

src/DynamoPackages/PackageLoader.cs
```diff
@@ -258,13 +260,29 @@ private void TryLoadPackageIntoLibrary(Package package)
...
+                            if (ex is LibraryLoadFailedException)
+                            {
+                                // Managed exception
+                                // We can still try to load other parts of the package
+                                Log(ex.GetType() + ": " + ex.Message);
+                            }
+                            else
+                            {
+                                // Everything else is considered fatal to the package loading operation.
+                                fatalError = true;
```
pinzart90, line 275: I guess I am silently discarding `AssemblyBlockedException` exceptions here...
Not sure how to make those fit in this scenario...
Maybe if all assemblies fail....with the same exception type `AssemblyBlockedException`...then we should re throw that ?...

src/DynamoPackages/PackageLoader.cs
```diff
@@ -292,22 +310,25 @@ private void TryLoadPackageIntoLibrary(Package package)
...
-            {
-                if (e is DynamoServices.AssemblyBlockedException)
+
+                if (e is CustomNodePackageLoadException ce)
+                {
+                    Package originalPackage =
+                        localPackages.FirstOrDefault(x => x.CustomNodeDirectory == ce.InstalledPath);
+                    OnConflictingPackageLoaded(originalPackage, package);
+                }
+                else if (e is DynamoServices.AssemblyBlockedException)
```
mjkkirschner, line 326: shouldn't this be an `if`?

src/DynamoCore/Models/DynamoModel.cs
aparajit-pratap: Same question as Mike.

src/DynamoCore/Models/DynamoModel.cs
aparajit-pratap: You could also do:
```
LibraryServices.ImportLibrary(assem.Location, isExplicitlyImportedLib : false);
```
to indicate what the false is for.

src/DynamoPackages/PackageLoader.cs
aparajit-pratap: Yes, I think we should. Can you please do that?

aparajit-pratap: I think it's cleaner if you do:
```
catch (LibraryLoadFailedException ex)
{ 
...
}
catch (Exception ex)
{
...
}
```

src/DynamoPackages/PackageLoader.cs
```diff
@@ -292,22 +310,25 @@ private void TryLoadPackageIntoLibrary(Package package)
                 PythonServices.PythonEngineManager.Instance.
                     LoadPythonEngine(package.LoadedAssemblies.Select(x => x.Assembly));
             }
-            catch (CustomNodePackageLoadException e)
+            catch (Exception e)
```
aparajit-pratap, line 313: I think a better coding practice is to catch the most derived exceptions first, then the generic ones later like:
```
catch (CustomNodePackageLoadException e)
{ ... }
catch (DynamoServices.AssemblyBlockedException) 
{ ... }
catch (Exception e)
{ ... }
```

pinzart90: Normally yes, that is cleaner.
In this case there code that has to run for all exception types.
This way I avoid code duplication

src/DynamoPackages/PackageLoader.cs
pinzart90: well...it cannot be both exception types at once.
`else if` seems to enforce that fact a bit clearer

src/DynamoPackages/PackageLoader.cs
pinzart90: refactored some more and I put back the try catch code as it was

src/DynamoPackages/PackageLoader.cs
pinzart90: not an issue anymore

src/DynamoCore/Models/DynamoModel.cs
pinzart90: I kept the name and reversed the condition.

src/DynamoPackages/PackageLoader.cs
```diff
@@ -260,11 +262,38 @@ private void TryLoadPackageIntoLibrary(Package package)
...
+                        {
+                            failedNodeLibs.Add(assem.Assembly);
+                        }
                     }
                 }
 
+                if (loadedNodeLibs.Count > 0)
+                {
+                    // Try to load any valid node libraries regardless package state
+                    PackagesLoaded?.Invoke(loadedNodeLibs);
```
aparajit-pratap, line 283: Why did you change the sequence of loading this  by moving it up here from [line 290](https://github.com/DynamoDS/Dynamo/pull/12763/files#diff-7ae21a0e381958737cc115ada84cbd9f96efe3315928cbc44b367967923cf55bL290)?

src/DynamoPackages/PackageLoader.cs
```diff
@@ -260,11 +262,38 @@ private void TryLoadPackageIntoLibrary(Package package)
...
+
+                if (blockedAssemblies.Count > 0)
+                {
+                    throw new Exception("The following assemblies are blocked : " + string.Join(", ", blockedAssemblies.Select(x => Path.GetFileName(x.Location))));
+                }
+
+                // Generic fatal error
+                if (failedNodeLibs.Count > 0)
+                {
+                    throw new Exception("Failed to load the following assemblies : " + string.Join(", ", failedNodeLibs.Select(x => Path.GetFileName(x.Location))));
```
aparajit-pratap, line 294: Shouldn't this be `new LibraryLoadFailedException`?

src/DynamoPackages/PackageLoader.cs
pinzart90: so that I do not have to add it in 2 places (when it succeeds and in the catch blocks)

src/DynamoPackages/PackageLoader.cs
```diff
@@ -260,11 +262,38 @@ private void TryLoadPackageIntoLibrary(Package package)
...
 
+                if (loadedNodeLibs.Count > 0)
+                {
+                    // Try to load any valid node libraries regardless package state
+                    PackagesLoaded?.Invoke(loadedNodeLibs);
+                }
+
+                if (blockedAssemblies.Count > 0)
+                {
+                    throw new Exception("The following assemblies are blocked : " + string.Join(", ", blockedAssemblies.Select(x => Path.GetFileName(x.Location))));
```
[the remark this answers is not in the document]
pinzart90: correct. Thanks

src/DynamoPackages/PackageLoader.cs
pinzart90: well..it could be..but I don't think it matters...since we do not catch that in any special way in this try catch

src/DynamoPackages/PackageLoader.cs
pinzart90: To avoid code duplication...I can put it here (line 283) or in a finally block.

johnpierson:
Thank you all for your work on this. I am thrilled to see this as a "feature" rather than a "Random convenience" lol 😁

PR #12768, opened 2022-04-04
DYN-4020 : improve CLI handling

BogdanZavu:
### Purpose
We need to enhance the CLI handling so that applications like DynamoPlayer and GenerativeDesign can manage execution context more easily.

### Declarations

### Reviewers
@mjkkirschner @saintentropy @QilongTang

### DIFF
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -436,6 +436,12 @@ namespace Dynamo.Models
         /// </summary>
         public DynamoModelState State { get; internal set; } = DynamoModelState.NotStarted;
 
+        /// <summary>
+        /// CLIMode indicates if we are running in DynamoCLI or DynamoWPFCLI mode.
+        /// Note that in CLI mode Scheduler is synchronous.
+        /// </summary>
+        public bool CLIMode { get; private set; }
+
         protected virtual void PreShutdownCore(bool shutdownHost)
         {
         }
@@ -475,6 +481,7 @@ namespace Dynamo.Models
             IAuthProvider AuthProvider { get; set; }
             IEnumerable<IExtension> Extensions { get; set; }
             TaskProcessMode ProcessMode { get; set; }
+            bool CLIMode { get; set; }
 
             /// <summary>
             /// If true, the program does not have a UI.
@@ -502,6 +509,7 @@ namespace Dynamo.Models
             public TaskProcessMode ProcessMode { get; set; }
             public bool IsHeadless { get; set; }
             public string PythonTemplatePath { get; set; }
+            public bool CLIMode { get; set; }
             /// <summary>
             /// Default Python script engine
             /// </summary>
@@ -578,6 +586,7 @@ namespace Dynamo.Models
             Context = config.Context;
             IsTestMode = config.StartInTestMode;
             IsHeadless = config.IsHeadless;
+            CLIMode = config.CLIMode;
 
             DebugSettings = new DebugSettings();
             Logger = new DynamoLogger(DebugSettings, pathManager.LogDirectory, IsTestMode);
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -440,7 +440,7 @@ namespace Dynamo.Models
         /// CLIMode indicates if we are running in DynamoCLI or DynamoWPFCLI mode.
         /// Note that in CLI mode Scheduler is synchronous.
         /// </summary>
-        public bool CLIMode { get; private set; }
+        public bool CLIMode { get; internal set; }
 
         protected virtual void PreShutdownCore(bool shutdownHost)
         {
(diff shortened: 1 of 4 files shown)


saintentropy:
LGTM @BogdanZavu

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -481,7 +481,6 @@ public interface IStartConfiguration
             IAuthProvider AuthProvider { get; set; }
             IEnumerable<IExtension> Extensions { get; set; }
             TaskProcessMode ProcessMode { get; set; }
-            bool CLIMode { get; set; }
```
mjkkirschner: you can't remove a member from a public interface like this. It's an API break.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -509,7 +508,6 @@ public struct DefaultStartConfiguration : IStartConfiguration
             public TaskProcessMode ProcessMode { get; set; }
             public bool IsHeadless { get; set; }
             public string PythonTemplatePath { get; set; }
-            public bool CLIMode { get; set; }
```
mjkkirschner: same here.

mjkkirschner (changes requested):
There are API breaking changes in this PR that need to be addressed.

src/DynamoCore/Scheduler/DynamoSchedulerInternals.cs
```diff
@@ -145,6 +158,7 @@ private void ReprioritizeTasksInQueue()
 
         private void ProcessTaskInternal(AsyncTask asyncTask)
         {
+            HasTaskInProgress = true;
```
mjkkirschner, line 161: @BogdanZavu I saw that previously this was marked volatile.  I've never seen that modifier used in c#, only c (when doing embedded micro programming) - I found the following info:
https://stackoverflow.com/questions/72275/when-should-the-volatile-keyword-be-used-in-c

now, I see that the there is no concurrency primitives used, locks etc. Do we need to consider wrapping this accessor in a lock, or will the scheduler threadbe the only thread to interact with it?

mjkkirschner:
@BogdanZavu - sorry, I was looking at an older commit somehow regarding the API breaks, thanks for fixing them.
I have a few more requests.

1. Why do we need to make hasPendingTasks threadsafe, and is there a more idiomatic way than volatile? 
2. Can you add a test for HasPendingTasks?
3. Why should CLIMode be public?

BogdanZavu:
@mjkkirschner just to clarify, there is no API breakage, right?
1. It is possible to be called from another thread, as the main thread or other thread in a player like execution contexts. I'll use lock instead of volatile. Initially, I was thinking using lock is an overkill here.
2. Yes, actually I think this is related to HasPendingTasks so I think modifying the existing tests should do it
3. We will activate certain functionality in our extension based on CLIMode. We can make this internal to our DLLs but I think it's good to be in the public API overall.

BogdanZavu:
@mjkkirschner just to clarify, there is no API breakage, right?
1. It is possible to be called from another thread, like the main thread in various player execution contexts. I'll use lock instead of volatile, initially, I was thinking using lock is an overkill here.
2. Yes or I'll modify the existing HasPendingTasks tests
3. So that others can enable functionality that is available only in CLI mode, we can make it internal to our DLLs but it looks like something good to be public.

src/DynamoCore/Scheduler/DynamoSchedulerInternals.cs
BogdanZavu: I kind of disagree with that statement, especially for a simple bool var case like ours but let's not spend the weekend debating over it :) I will use lock

test/DynamoCoreTests/NodeToCodeTest.cs
```diff
@@ -1629,7 +1629,7 @@ protected void MutationTest(string dynFilePath)
 
             RunModel(dynFilePath);
             // Block until all tasks are executed
-            while (CurrentDynamoModel.Scheduler.HasPendingTasks) ;
+            while (CurrentDynamoModel.Scheduler.HasPendingTasks || CurrentDynamoModel.Scheduler.HasTaskInProgress);
```
aparajit-pratap, line 1632: Was there any specific reason to add the `HasTaskInProgress` flag? Is the absence of it causing any issues where the check for pending tasks is insufficient? It appears that checking for this new flag will ensure that the last task (task queue should be empty when the last task is executing) also finishes, correct?

BogdanZavu: Yes

sm6srw (approved):
Nice, LGTM!

test/DynamoCoreTests/NodeToCodeTest.cs
aparajit-pratap: Can you elaborate what kind of issues you're seeing without this flag? Curious to know.

BogdanZavu: It is exactly like you described it so basically, we were trying to retrieve some information too early before a task was completed. 
The issue was always there just that in Revit context we are relying on OnIdle thread to execute/interrogate DynamoModel and workspace and since we are in the same process. The OnIddle ensured there was nothing left to do since we are in the same process. 
Now in CLIMode we might want to do these checks from another thread so need to check to make sure all tasks are finished before we perform an interrogation.

PR #12789, opened 2022-04-08
NodeAutocomplete UI fix

reddyashish:
### Purpose

JIRA:  https://jira.autodesk.com/browse/DYN-4454

### Declarations

### Release Notes

### Reviewers
@QilongTang @zeusongit

### DIFF
--- a/src/DynamoCoreWpf/Controls/IncanvasSearchControl.xaml
+++ b/src/DynamoCoreWpf/Controls/IncanvasSearchControl.xaml
@@ -71,7 +71,7 @@
                                    DockPanel.Dock="Left"
                                    FontFamily="{StaticResource ArtifaktElementRegular}"
                                    FontSize="14"
-                                   Foreground="{StaticResource DeafultSearchTextBlockText}"
+                                   Foreground="{StaticResource DefaultSearchTextBlockText}"
                                    IsHitTestVisible="False"
                                    Text="{x:Static resx:Resources.SearchTextBlockText}"
                                    Visibility="{Binding Path=SearchText, Converter={StaticResource NonEmptyStringToCollapsedConverter}}" />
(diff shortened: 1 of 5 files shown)


QilongTang:
@reddyashish Any screenshot of what it looks like now?

reddyashish:
Screen Shot 2022-04-11 at 10 46 06 AM

QilongTang (approved):

reddyashish:
Looking into it now.

zeusongit:
As per AC for this task, does this PR also handle long strings and truncate them using the ellipsis(...)?
Also, aabi pointed out that this happens in a localized setting, is this fixed in a localized environment as well?

reddyashish:
@zeusongit yes this is fixed in a localized environment. If the string is long, it does use ellipsis like in the screenshot below
Screen Shot 2022-04-12 at 10 18 27 AM


The tests have passed here: https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/956/

QilongTang:
@reddyashish The screenshot looks a bit strange, the entry highlight is less wide than the dialog overall?

zeusongit (approved):

Amoursol:
> Yes, that is because the width of the dialog is different for different languages to compensate the header text. Using Auto for width, does't work here because it will highlight only till the width of the item string. So this width is fixed. what do you think?

The static size of the width, is that "Limited" by the other components, or fixed by itself? For example, if we were to make it say, twice as long currently, would the English UI look the same? And then would the German UI look complete? 

If possible we would want the full horizontal background color to sit appropriately.

Amoursol:
> @Amoursol If we make it twice as long currently, the english UI would be too long in width.

Ok gotcha, so it's not bound and will overlap and look funny. If there are no other viable options, we can run with this.

QilongTang:
Discussed with @reddyashish that we may want to make the internal entry width more dynamic

reddyashish:
@QilongTang @Amoursol 

Looks like this with the new changes:

Screen Shot 2022-04-20 at 8 54 56 AM
Screen Shot 2022-04-20 at 8 55 26 AM
Screen Shot 2022-04-20 at 8 55 34 AM

QilongTang (approved):

PR #12790, opened 2022-04-08
WIP Start implementing feature flags for toy feature

mjkkirschner:
### Purpose

This PR onboards Dynamo onto launch darkly, it controls a popup at startup. The feature flag is enabled for debug builds, and disabled for production (release) builds.

PR does the following:
* adds a new project DynamoFeatureFlags that references the newest launch darkly client side dot net sdk nuget package.
* ~updates compiler unsafe to avoid conflict at build time~ ... latest changes revert this as LD is resolving the same version with Tibi's suggestion.
* Starts the launch darkly client if - analytics is not disabled by the configuration file, or the DisableAnalytics flag - otherwise LD will start, so it's not controllable via any UI setting, only by commandline, config file etc.
* We use the stable instrumentation id that is generated for analytics, if this cannot be created or loaded, we use a shared key. Shared key will also be used during testing etc.
* All users are logged as anonymous, which means that we cannot see them on the LD dashboard, but we can do targeted role outs / AB testing as long as the shared key is not used.
*  mobile sdk keys are embedded into the config file for the new project - according to LD docs, this is safe with no risk, these keys are not full sdk keys, they're more like a url for our dev/prod environments. 
* To avoid a slow startup the LD client is started async, so you can't immediately check feature flags unless you want to block - I have not implemented that, but it should be possible to add to the `CheckFeatureFlag<T>` method.

### New Binaries:

- LaunchDarkly.ClientSdk.dll                             - manual netstandard2.0
- LaunchDarkly.CommonSdk.dll                       - manual netstandard2.0
- LaunchDarkly.EventSource.dll                        - manual netstandard2.0
- LaunchDarkly.InternalSdk.dll                          - manual netstandard2.0
- LaunchDarkly.JsonStream.dll                          - manual netstandard2.0
- LaunchDarkly.Logging.dll                               - manual netstandard2.0
- ~Microsoft.Bcl.AsyncInterfaces.dll~
- System.Buffers.dll                                           - automatic net461
- System.Memory.dll                                         - automatic net461
- System.Numerics.Vectors.dll                           - automatic net461
- System.Runtime.CompilerServices.Unsafe.dll - automatic net461
- ~System.Text.Encodings.Web.dll~
- ~System.Text.Json.dll~
- ~System.Threading.Tasks.Extensions.dll~
- ~System.ValueTuple.dll~

package refs for the DynamoFeatureFlags Project:
```
'DynamoFeatureFlags' has the following package references
   [netstandard2.0]: 
   Top-level Package                Requested   Resolved
   > LaunchDarkly.ClientSdk         2.0.1       2.0.1   
   > NETStandard.Library      (A)   [2.0.3, )   2.0.3   

   Transitive Package                            Resolved
   > LaunchDarkly.CommonSdk                      5.5.0   
   > LaunchDarkly.EventSource                    4.1.3   
   > LaunchDarkly.InternalSdk                    2.3.2   
   > LaunchDarkly.JsonStream                     1.0.3   
   > LaunchDarkly.Logging                        1.0.1   
   > Microsoft.NETCore.Platforms                 1.1.0   
   > System.Buffers                              4.5.1   
   > System.Collections.Immutable                1.7.1   
   > System.Memory                               4.5.4   
   > System.Numerics.Vectors                     4.5.0   
   > System.Runtime.CompilerServices.Unsafe      4.5.3   
```

Screen Shot 2022-04-12 at 7 10 09 PM

### TODO:
- [ ] tests
- [ ] in a host like Revit where Dynamo is stopped and restarted, need to test if disposal/restarting of the LD client is smooth.
- [ ] if LD deps are missing, ensure that DynamoCore still starts fine.
- [ ] satisfied with new binaries this PR pulls in.
- [x] compiler.unsafe version alignment?
- [x] reference to system.configuration.dll in new csproj looks very wrong. (hint path)
- [x] if the user is logged in, we might want to add more attributes like email - so we can target specific emails for some features - we should add the private attribute flag if so to avoid storing any PII in LD.

### Declarations

### Release Notes

Add initial feature flags controls in Dynamo.

### Reviewers

### FYIs

### DIFF
--- a/src/DynamoFeatureFlags/DynamoFeatureFlags.csproj
+++ b/src/DynamoFeatureFlags/DynamoFeatureFlags.csproj
@@ -1,13 +1,20 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <ImportGroup Label="PropertySheets">
     <Import Project="$(SolutionDir)Config/CS_SDK.props" />
   </ImportGroup>
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
+    <!--force copying nuget netstd2 dependencies-->
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="LaunchDarkly.ClientSdk" Version="2.0.1" />
+    <PackageReference Include="LaunchDarkly.ClientSdk" Version="2.0.1">
+      <!--using this to avoid copying net461 deps of launch darkly -->
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+    <!--to force LD package to reference system.collections.immutable of the same version. -->
+    <PackageReference Include="System.Collections.Immutable" Version="5.0.0" />
   </ItemGroup>
 
 </Project>
--- a/src/DynamoFeatureFlags/DynamoFeatureFlags.csproj
+++ b/src/DynamoFeatureFlags/DynamoFeatureFlags.csproj
@@ -4,16 +4,44 @@
   </ImportGroup>
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
-    <!--force copying nuget netstd2 dependencies-->
-    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="LaunchDarkly.ClientSdk" Version="2.0.1">
-      <!--using this to avoid copying net461 deps of launch darkly -->
-      <PrivateAssets>all</PrivateAssets>
-    </PackageReference>
-    <!--to force LD package to reference system.collections.immutable of the same version. -->
+    <!--We want to us the netstandard2.0 flavor of launch darkly-->
+    <!--But we want nuget to resolve the non-ld assemblies (ex. the immutable pkg) -->
+    <PackageReference Include="LaunchDarkly.ClientSdk" Version="2.0.1" IncludeAssets="none" GeneratePathProperty="true" PrivateAssets="all" />
+    <Reference Include="LaunchDarkly.ClientSdk">
+      <HintPath>$(PkgLaunchDarkly_ClientSdk)/lib/netstandard2.0/LaunchDarkly.ClientSdk.dll</HintPath>
+      <CopyLocal>True</CopyLocal>
+    </Reference>
+    <PackageReference Include="LaunchDarkly.CommonSdk" Version="5.5.0" IncludeAssets="none" GeneratePathProperty="true" PrivateAssets="all" />
+    <Reference Include="LaunchDarkly.CommonSdk">
+      <HintPath>$(PkgLaunchDarkly_CommonSdk)/lib/netstandard2.0/LaunchDarkly.CommonSdk.dll</HintPath>
+      <CopyLocal>True</CopyLocal>
+    </Reference>
+    <PackageReference Include="LaunchDarkly.EventSource" Version="4.1.3" IncludeAssets="none" GeneratePathProperty="true" PrivateAssets="all" />
+    <Reference Include="LaunchDarkly.EventSource">
+      <HintPath>$(PkgLaunchDarkly_EventSource)/lib/netstandard2.0/LaunchDarkly.EventSource.dll</HintPath>
+      <CopyLocal>True</CopyLocal>
+    </Reference>
+    <PackageReference Include="LaunchDarkly.InternalSdk" Version="2.3.2" IncludeAssets="none" GeneratePathProperty="true" PrivateAssets="all" />
+    <Reference Include="LaunchDarkly.InternalSdk">
+      <HintPath>$(PkgLaunchDarkly_InternalSdk)/lib/netstandard2.0/LaunchDarkly.InternalSdk.dll</HintPath>
+      <CopyLocal>True</CopyLocal>
+    </Reference>
+    <PackageReference Include="LaunchDarkly.JsonStream" Version="1.0.3" IncludeAssets="none" GeneratePathProperty="true" PrivateAssets="all" />
+    <Reference Include="LaunchDarkly.JsonStream">
+      <HintPath>$(PkgLaunchDarkly_JsonStream)/lib/netstandard2.0/LaunchDarkly.JsonStream.dll</HintPath>
+      <CopyLocal>True</CopyLocal>
+    </Reference>
+    <PackageReference Include="LaunchDarkly.Logging" Version="1.0.1" IncludeAssets="none" GeneratePathProperty="true" PrivateAssets="all" />
+    <Reference Include="LaunchDarkly.Logging">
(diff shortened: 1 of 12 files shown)


src/DynamoFeatureFlags/DynamoFeatureFlags.csproj
```diff
@@ -0,0 +1,19 @@
...
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="LaunchDarkly.ClientSdk" Version="2.0.1" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Reference Include="System.Configuration">
+      <HintPath>..\..\..\..\..\..\..\..\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.8\System.Configuration.dll</HintPath>
```
mjkkirschner, line 15: not sure whats up with this... we should not need a hint path for a bcl library.

mjkkirschner: indeed, I will look at this:
```
D:\a\Dynamo\Dynamo\Dynamo\src\DynamoFeatureFlags\FeatureFlagsManager.cs(25,30): error CS0103: The name 'ConfigurationManager' does not exist in the current context [D:\a\Dynamo\Dynamo\Dynamo\src\DynamoFeatureFlags\DynamoFeatureFlags.csproj]

    411 Warning(s)
    1 Error(s)
```

mjkkirschner: oh, this is apparently because this functionality requires:
https://www.nuget.org/packages/system.configuration.configurationmanager/
when targeting .net standard 2
@pinzart90 did you run into this anywhere else?

mjkkirschner: also - from my reading it appears I can for the time being move to .net 4.8 target safely, though we'll need to figure this out eventually.

pinzart90: System.Configuration was not included in the nestandard list of assemblies/specifications (probably because they wanted to encourage other portable/lightweight solutions - like .json file configs)

They did make it available as a package though (as you mentioned).
I see 3 options here:
1. Use the PackageReference to system.config.manager
    This means that the net48 builds will have the new dll (even though net48 has its own flavor of system.configuration with different APIs)
2. Do multi targeting - net48;net5.0(or 6.0)
    This would require #id #def inside the code to use the different APIs
3. Use a handwritten way (with no external deps)
    Ex. like we do in LibG and [PackageManagerClient](https://github.com/DynamoDS/PackageManagerClient/blob/master/src/GregClient/Utility/Utilities.cs#L44)

src/DynamoFeatureFlags/DynamoFeatureFlags.csproj
```diff
@@ -0,0 +1,19 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <ImportGroup Label="PropertySheets">
+    <Import Project="$(SolutionDir)Config/CS_SDK.props" />
+  </ImportGroup>
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="LaunchDarkly.ClientSdk" Version="2.0.1" />
```
pinzart90, line 10: are we OK with bringing in a bunch of extra dependencies to the dynamo build ?
(I know we tried to avoid this in the past)

mjkkirschner: I'm not sure, do you think it's worth exploring wrapping up the launch darkly client in a separate process the same way we do for the markdown2htmlCLI tool? This would isolate the dependencies from DynamoCore /  the host process.

Performance might be a bit worse, but flags shouldn't be getting evaluated very frequently anyway.

pinzart90: From my point of view, LD in a separate process would make sense only if extra dll dependencies are deal breaker.
Some of the extra dependencies will be brought in by the move to dotnet6 anyway.

I did also see a more lightweight REST API client https://www.nuget.org/packages/LaunchDarkly.Api/
Not sure if this would a better alternative ...The API does look more difficult to use ...

mjkkirschner:
I believe that api is a non starter in our case as it requires full sdk keys that must remain secret - ie it’s for a custom automation tool, not a client process - but I will double check.

IMO we can start by referencing the ld client to keep things straight forward and then move to out of process if we run into conflicts in the future. It’s not clear how well features like local storage will work wrapped in another process- we might need to contend with other gotchas like that for conflicts that don’t exist.

src/DynamoFeatureFlags/DynamoFeatureFlags.csproj
mjkkirschner: went with option 3

src/DynamoFeatureFlags/DynamoFeatureFlags.csproj
```diff
@@ -0,0 +1,20 @@
...
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="LaunchDarkly.ClientSdk" Version="2.0.1">
+      <!--using this to avoid copying net461 deps of launch darkly -->
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+    <!--to force LD package to reference system.collections.immutable of the same version. -->
+    <PackageReference Include="System.Collections.Immutable" Version="5.0.0" />
```
pinzart90, line 17: I am not sure how well `CopyLocalLockFileAssemblies` will work vs same pacakgeReferences in other projects. 
Ex CopyLocalLockFileAssemblies = true will pull down the dotnetstandard flavor of all pacakgeReferences in the csproj. But there are other Dynamo projects which could have the same pacakgeRefereces (but with no copylocallock setting)
ex. 
1. System.Collections.Immutable@5.0.0 referenced in featureFlags csproj will copy over System.Collections.Immutable.dll from the netstandard2.0 subfolder
2. System.Collections.Immutable@5.0.0 referenced from designscriptbuiltin would copy over System.Collections.Immutable.dll from net461 (closest target framework moniker to the projects tfm)

src/DynamoFeatureFlags/FeatureFlagsManager.cs
```diff
@@ -0,0 +1,144 @@
...
+                if (ldClient == null || !ldClient.Initialized)
+                {
+                    MessageLogged($"feature flags client not initalized for requested flagkey: {flagkey}, returning default value: {defaultval}");
+                    return defaultval;
+                }
+
+                object output = default(T);
+                switch (default(T))
+                {
+                    case bool _:
```
aparajit-pratap, line 95: Are we testing for type here? If so, can we also use `typeof` as in:
```
switch (typeof(T))
{
    case typeof(bool):
...
}
```

mjkkirschner: Screen Shot 2022-04-18 at 11 58 27 AM

also fails on dotnet 6 unfortunately, mixing generic and switch case is not a great idea I guess.

PR #12801, opened 2022-04-14
wrap launch darkly in CLI to avoid bringing in new deps

mjkkirschner:
## Updated 4/18 to short lived CLI process.

An alternative to: 
https://github.com/DynamoDS/Dynamo/pull/12790

in this PR instead of pulling LD sdk and its dependencies into the same process as Dynamo/Host - it's wrapped in a CLI following a very similar pattern to the markdown2html conversion tool.

### How it works:
* currently, the CLI tool is started at dynamo startup, initializes a connection with LD, asks LD for all the current user's known flags and their values, and will immediately send back a json object containing that data.
```
launch darkly initalized
2022-04-19 18:50:24Z : LD startup time: 897 
2022-04-19 18:50:26Z : <<<<<InitDone>>>>>
2022-04-19 18:50:27Z : feature flag exe starting
2022-04-19 18:50:30Z : <<<<<Sod>>>>>
2022-04-19 18:50:32Z : {"EasterEggIcon1":true,"EasterEggMessage1":"Hello fellow dynamo developers, have a great day."}
2022-04-19 18:50:34Z : <<<<<Eod>>>>>
```
I've added the `start of data` messages so I can safely log other information.

The process then shuts down.

If somehow Dynamo shuts down during the brief window where the flags were being retrieved, the CLI will also shut down if the host process is no longer running. It checks this every 10 seconds.

The CLI tool/wrapper are started on another thread using `Task.Run` to avoid blocking the startup of DynamoModel/DynamoView - this makes synchronizing when to check flags a bit complicated, so I've added an event that can be subscribed to - still, it's not foolproof as the event may have fired before you have a chance to subscribe to it, in this case the `checkFlag` API can be queried directly.

If we encounter a case in the future where we absolutely must control DynamoModel/DynamoView constructors via feature flags, then we can just remove the `Task.Run` and wait for the flags to be retrieved synchronously, and get rid of the event. 

Screen Shot 2022-04-20 at 4 05 02 PM

### issues/TODO:
- [ ] The LaunchDarklyClient.Sdk refuses to run when its been merged with ILMerge - I have not tried merging only some of the deps- but it happily works if all the deps are in a separate **DynamoFeatureFlags** folder.
- [x] Tests
- [x] Test manually in Revit to make sure no issues with Task.Run/sync context.
- [x] std.error is redirected and read async to avoid the buffer filling up in a long running process.

### Declarations

### Release Notes

Add FeatureFlagsManager internal component for querying live feature flags.

### Reviewers

### FYIs


pinzart90:
Maybe we can have a self destruct functionality inside the launchDarkly process. In case it becomes a zombie..
It could check for the existance of a specific process (the dynamo process). If the dynamo process is not alive anymore...then self destruct.
We can pass the dynamo process id or name when the LD process is created...

mjkkirschner:
> Maybe we can have a self destruct functionality inside the launchDarkly process. In case it becomes a zombie..
> It could periodically check for the existence of the parent process (the dynamo process). If the dynamo process is not alive anymore...then self destruct.
> We can pass the dynamo process id or name when the LD process is created...

good idea!

I'm looking at the build failure. It's not obvious to me which path doesn't exist.

pinzart90:
Regarding deadlocks or lack of timeouts...
One other option (not saying it is better/faster) would be to make the separate process an http server

src/Tools/DynamoFeatureFlags/DynamoFeatureFlags.csproj
```diff
@@ -0,0 +1,30 @@
...
+    <Exec Condition=" '$([MSBuild]::IsOSPlatform(`Windows`))' == 'True'" Command="if not exist $(SharedOutputPath) mkdir $(SharedOutputPath)" />
+    <!-- the ILMergePath property points to the location of ILMerge.exe console application -->
+    <!--<Exec Condition=" '$([MSBuild]::IsOSPlatform(`Windows`))' == 'True'" Command="$(ILMergeConsolePath) /log $(OutputPath)DynamoFeatureFlags.exe /allowDup /lib:$(OutputPath) /out:$(SharedOutputPath)DynamoFeatureFlags.exe CommandLine.dll LaunchDarkly.ClientSdk.dll LaunchDarkly.CommonSdk.dll LaunchDarkly.EventSource.dll LaunchDarkly.internalSdk.dll LaunchDarkly.JsonStream.dll LaunchDarkly.Logging.dll Microsoft.Bcl.AsyncInterfaces.dll System.Buffers.dll System.Collections.Immutable.dll System.Memory.dll System.Numerics.Vectors.dll System.Runtime.CompilerServices.Unsafe.dll System.Text.Encodings.Web.dll System.Text.Json.dll System.Threading.Tasks.Extensions.dll System.ValueTuple.dll " />-->
+    <Exec Condition=" '!$([MSBuild]::IsOSPlatform(`Windows`))' == 'True' " Command="mkdir -p $(SharedOutputPath)" />
+    <Exec Condition=" '$([MSBuild]::IsOSPlatform(`Windows`))' == 'True' " Command="xcopy /y  $(OutputPath) $(SharedOutputPath)" />
```
pinzart90, line 27: not sure I understand what you are trying to do here...
Looks like this is the cause of the build failure

pinzart90: if you are just trying to box up the feature flag related dlls into a subfolder...then override the output path
```
<OutputPath>$(OutputPath)\DynamoFeatureFlags\</OutputPath>
```

pinzart90: There should be no need to copy things around...

mjkkirschner: yep, since ilmerge doesn't work here anyhow, I'll get rid of  the xcopy.

sm6srw:
Is there any fundamental difference in this implementation compared to the markdown tool? I kind of remember that I had very few issues with zombies in the end with the markdown tool. One thing that could cause this is if the process creates threads that is not properly destroyed. Maybe something that happens in the LD SDK?

src/Tools/DynamoFeatureFlags/Program.cs
```diff
@@ -0,0 +1,100 @@
...
+                            Console.WriteLine(startOfDataToken);
+                            CheckFeatureFlag();
+                        }
+                        Console.WriteLine(endOfDataToken);
+                    }
+
+                }
+                catch (Exception e) when (e is IOException || e is OutOfMemoryException || e is ArgumentOutOfRangeException)
+                {
+                    // Exit process
```
sm6srw, line 51: We should get here when the main process exit and the pipe closes. We could be too restrictive here. Can you temporary comment out `when` and see if that helps?

mjkkirschner: tested this on both sides (dynamo and the cli program), no difference.

In my latest commit I've implemented a simple watchdog on a timer as suggested by @pinzart90 - this works in my testing.

mjkkirschner:
moved to draft while I try to reproduce the test hang locally.

mjkkirschner:
this is no longer timing out the build machines - but there is a single failure, looking at it.

src/Tools/DynamoFeatureFlags/Program.cs
```diff
@@ -0,0 +1,119 @@
...
+{
+
+    class Options
+    {
+        [Option('u', "userkey", Required = false, HelpText = "stable user key, if not provided, a shared key will be used.")]
+        public string UserKey { get; set; }
+        [Option('m', "mobilekey", Required = false, HelpText = "mobile key for dynamo feature flag env. Do not use a full sdk key. If not provided loaded from config.")]
+        public string MobileKey { get; set; }
+        [Option('p',"proccessID",Required = true, HelpText= "parent proccess id, if this proccess is no longer running, this application will exit.")]
+        public int ProccessID { get; set; }
```
aparajit-pratap, line 20: typo: ProcessID

src/Tools/DynamoFeatureFlags/Program.cs
```diff
@@ -0,0 +1,119 @@
...
+            //System.Diagnostics.Debugger.Launch();
+            Parser.Default.ParseArguments<Options>(args).WithParsed(ops =>
+            {
+                try
+                {
+                    //start watchdog
+                    hostProccessId = ops.ProccessID;
+                    var timer = new System.Timers.Timer(10000);
+                    timer.AutoReset = true;
+                    timer.Elapsed += Timer_Elapsed;
```
aparajit-pratap, line 40: So this checks for the parent process every 10 sec?

mjkkirschner: yep exactly.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -644,9 +647,38 @@ protected DynamoModel(IStartConfiguration config)
...
             // If user skipped analytics from assembly config, do not try to launch the analytics client
+            // or the feature flags client.
             if (!areAnalyticsDisabledFromConfig && !Dynamo.Logging.Analytics.DisableAnalytics)
             {
                 AnalyticsService.Start(this, IsHeadless, IsTestMode);
+
+
+                //run process startup/reading on another thread so we don't block dynamo startup.
+                //if we end up needing to control aspects of dynamo model or view startup that we can't make
+                //event based/async then just run this on main thread - ie get rid of the Task.Run()
```
aparajit-pratap, line 658: By `if we end up needing to control aspects of dynamo model or view startup`, do you mean a case where we'll need to check for feature flags at some point before the model is instantiated?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1199,7 +1253,9 @@ public void Dispose()
             CustomNodeManager.MessageLogged -= LogMessage;
             CustomNodeManager.Dispose();
             MigrationManager.MessageLogged -= LogMessage;
-
+            FeatureFlags.MessageLogged -= LogMessageWrapper;
+            DynamoUtilities.DynamoFeatureFlagsManager.FlagsRetrieved -= CheckFeatureFlagTest;
+            FeatureFlags.Dispose();
```
aparajit-pratap, line 1258: Isn't it going to be disposed at the end of the `using` anyway?

src/DynamoUtilities/DynamoFeatureFlagsManager.cs
```diff
@@ -0,0 +1,117 @@
...
+    /// Which is itself a wrapper around LaunchDarkly.
+    /// Not thread safe.
+    /// </summary>
+    internal class DynamoFeatureFlagsManager : CLIWrapper
+    {
+        private string relativePath = Path.Combine("DynamoFeatureFlags", "DynamoFeatureFlags.exe");
+        private Dictionary<string, object> AllFlagsCache { get; set; }//TODO lock is likely overkill.
+        private SynchronizationContext syncContext;
+        internal static event Action FlagsRetrieved;
+        public override void Dispose()
```
aparajit-pratap, line 27: Does this need to override the base class Dispose? The base Dispose is killing the process anyway.

src/DynamoUtilities/Md2Html.cs
```diff
@@ -7,48 +7,178 @@
...
+        protected bool started;
+        internal event Action<string> MessageLogged;
+
+        public virtual void Dispose()
+        {
+            process.ErrorDataReceived -= Process_ErrorDataReceived;
+            KillProcess();
+        }
+
+        protected virtual void StartProces(string relativeEXEPath, string argString)
```
aparajit-pratap, line 28: typo: StartProcess

src/DynamoCore/Models/DynamoModel.cs
mjkkirschner: yes, and where we cannot recreate or reset that aspect of the model or view.

src/DynamoCore/Models/DynamoModel.cs
mjkkirschner: yes, this is a leftover from the other implementation, will get rid of it.

src/DynamoUtilities/DynamoFeatureFlagsManager.cs
mjkkirschner: good spot, can remove this.

mjkkirschner:
one failure
```
MESSAGE:
TearDown : System.InvalidOperationException : No process is associated with this object.
+++++++++++++++++++
STACK TRACE:
--TearDown
   at System.Diagnostics.Process.EnsureState(State state)
   at System.Diagnostics.Process.EnsureState(State state)
   at System.Diagnostics.Process.GetProcessHandle(Int32 access, Boolean throwIfExited)
   at System.Diagnostics.Process.Kill()
   at Dynamo.Utilities.CLIWrapper.KillProcess() in c:\WorkspaceDynamo\src\DynamoUtilities\Md2Html.cs:line 77
   at Dynamo.Models.DynamoModel.Dispose() in c:\WorkspaceDynamo\src\DynamoCore\Models\DynamoModel.cs:line 1271
   at Dynamo.Models.DynamoModel.ShutDownCore(Boolean shutdownHost) in c:\WorkspaceDynamo\src\DynamoCore\Models\DynamoModel.cs:line 455
   at Dynamo.Models.DynamoModel.ShutDown(Boolean shutdownHost) in c:\WorkspaceDynamo\src\DynamoCore\Models\DynamoModel.cs:line 420
   at Dynamo.ViewModels.DynamoViewModel.PerformShutdownSequence(ShutdownParams shutdownParams) in c:\WorkspaceDynamo\src\DynamoCoreWpf\ViewModels\Core\DynamoViewModel.cs:line 2930
   at DynamoCoreWpfTests.DynamoTestUIBase.Exit() in c:\WorkspaceDynamo\test\DynamoCoreWpfTests\DynamoTestUIBase.cs:line 124
``` looking at it.

**Comment by mjkkirschner** [comment]
Tests passed now, running master-15 serial tests as well. PTAL!

File: src/Dynamo.All.sln [hunk 9]
```diff
@@ -905,6 +907,14 @@ Global
...
 		{133FC760-5699-46D9-BEA6-E816B5F01016}.Release|x64.ActiveCfg = Release|Any CPU
 		{133FC760-5699-46D9-BEA6-E816B5F01016}.Release|x64.Build.0 = Release|Any CPU
+		{94CF053D-BB66-4FC7-883B-48F072701BA9}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{94CF053D-BB66-4FC7-883B-48F072701BA9}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{94CF053D-BB66-4FC7-883B-48F072701BA9}.Debug|x64.ActiveCfg = Debug|Any CPU
+		{94CF053D-BB66-4FC7-883B-48F072701BA9}.Debug|x64.Build.0 = Debug|Any CPU
+		{94CF053D-BB66-4FC7-883B-48F072701BA9}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{94CF053D-BB66-4FC7-883B-48F072701BA9}.Release|Any CPU.Build.0 = Release|Any CPU
+		{94CF053D-BB66-4FC7-883B-48F072701BA9}.Release|x64.ActiveCfg = Release|Any CPU
+		{94CF053D-BB66-4FC7-883B-48F072701BA9}.Release|x64.Build.0 = Release|Any CPU
```
pinzart90, line 917: do we want to build this tool by default ....with the rest of the solution  ?

src/DynamoCore/Logging/AnalyticsService.cs
```diff
@@ -13,7 +13,7 @@ class AnalyticsService
     {
         // Use the Analytics.Core interface so that we do not have to load the ADP assembly at this time.
         private static IAnalyticsUI adpAnalyticsUI;
-
+        private static IAnalyticsClient dynamoAnalyticsClient;
```
pinzart90, line 16: a bit weird that we need to store the same static instance in 2 places. (on is in DynamoServices proj class Analytics)
Maybe just make the DynamoServices one as internal ?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -2744,6 +2799,10 @@ private void LogMessage(ILogMessage obj)
         {
             Logger.Log(obj);
         }
+   private void LogMessageWrapper(string m)
```
pinzart90, line 2802: indentation/formatting

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -2770,7 +2829,7 @@ select string.IsNullOrEmpty(function.Namespace)
         }
 #endif
 
-        private void AddNodeTypeToSearch(TypeLoadData typeLoadData)
+    private void AddNodeTypeToSearch(TypeLoadData typeLoadData)
```
pinzart90, line 2832: indentation/formatting

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -873,6 +905,28 @@ protected DynamoModel(IStartConfiguration config)
             DynamoReady(new ReadyParams(this));
         }
 
+        private void CheckFeatureFlagTest()
+        {
+            if (DynamoModel.FeatureFlags.CheckFeatureFlag<bool>("EasterEggIcon1", false))
```
pinzart90, line 910: we will need to store the possible feature flag names in some way (config file?) right ?

mjkkirschner: why?

src/Dynamo.All.sln
mjkkirschner: We do so for the other CLI tools, why wouldn't we?

src/Dynamo.All.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.32106.194
+# Visual Studio Version 17
+VisualStudioVersion = 17.1.32228.430
```
QilongTang, line 4: required?

src/DynamoCore/Logging/AnalyticsService.cs
mjkkirschner: I'll take a look at improving this, yeah, I can make it internal I think.

src/Dynamo.All.sln
mjkkirschner: will try to revert - should be fiiiiine.

src/Tools/DynamoFeatureFlags/DynamoFeatureFlags.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <ImportGroup Label="PropertySheets">
+    <Import Project="$(SolutionDir)Config/CS_SDK.props" />
+  </ImportGroup>
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <OutputPath>$(OutputPath)\DynamoFeatureFlags\</OutputPath>
```
QilongTang, line 7: Remind me, the output will not be part of DynamoCore right?

mjkkirschner: It will be placed in `Bin\DynamoFeatureFlags` subfolder so it will be part of DynamoCore release zip and artifactory build - but it won't pollute the root bin folder.

src/DynamoCore/Models/DynamoModel.cs
mjkkirschner: I should mention that the keys, once created in launch darkly, are immutable.

src/DynamoCore/Models/DynamoModel.cs
mjkkirschner: this actually fixes the indentation.

src/Dynamo.All.sln
pinzart90: I guess if the tool does not directly depend on any other Dynamo projects it may have a slower dev cycle than the rest.
But I guess it is a small project (few cs files) so it does not really slow down the build process that much

src/DynamoCore/Models/DynamoModel.cs
pinzart90: In our discussions it seemed necessary to provide a way for feature flags to work offline (Like for ALIAS)
I may have misunderstood.
It is fine for now regardless.

pinzart90 (approved):
Some comments. 
LGTM

sm6srw (approved):
LGTM!

PR #12802, opened 2022-04-14
Dyn 4715 package name special characters

jesusalvino:
### Purpose

Prevent user from enter some special characters as part of the package name in the Publish Package feature.

### Declarations

###Reviewers
@QilongTang

###FYIs
@RobertGlobant20 @filipeotero


jesusalvino:
SpecialCharacters

src/DynamoCore/Search/SearchDictionary.cs
```diff
@@ -323,11 +323,17 @@ private static string[] SplitOnWhiteSpace(string s)
             return s.Split(null);
         }
 
-        private static bool ContainsSpecialCharacters(string element)
+        public static Char[] SpecialAndInvalidCharacters()
         {
-            return element.Contains("*") || element.Contains(".") || element.Contains(" ")
-                || element.Contains("\\");
+            return new[] { '*', '.', ' ', '\\', '/', ':', '?', '<', '>', '|', '"' };
```
jesusalvino, line 328: I have included the blank space as invalid character for a package in order to keep the previous implementation of the function, notice there are already created packages that contain blank spaces.

QilongTang: Is there a .NET API which can return these?

src/DynamoCoreWpf/UI/Themes/Modern/DynamoModern.xaml
```diff
@@ -723,6 +757,94 @@
         </Setter>
     </Style>
 
+    <Style x:Key="InputStyleWithIcon" TargetType="{x:Type TextBox}">
```
QilongTang, line 760: What are the change to this style?

src/DynamoCore/Search/SearchDictionary.cs
zeusongit: So, I think this could be problematic, what happens when a user wants to publish a new version for a package that already has spaces in the name, will that get flagged?
Also, why are we invalidating spaces?

src/DynamoCoreWpf/ViewModels/PackageManager/PublishPackageViewModel.cs
```diff
@@ -770,6 +804,8 @@ public PublishPackageViewModel( DynamoViewModel dynamoViewModel ) : this()
         {
             this.dynamoViewModel = dynamoViewModel;
             KnownHosts = initializeHostSelections();
+            //By Default the warning state of the Visual Settings tab (Group Styles section) will be disabled
+            isWarningEnabled = false;
```
QilongTang, line 808: The comments seems wrong?

src/DynamoCoreWpf/ViewModels/PackageManager/PublishPackageViewModel.cs
```diff
@@ -1777,6 +1818,15 @@ private bool CheckPackageValidity()
...
             return true;
         }
-    }
 
+        /// <summary>
+        /// This method will enable the warning icon next to the GroupName TextBox and other buttons needed
+        /// </summary>
+        /// <param name="warningMessage">Message that will be displayed when the mouse is over the warning</param>
+        internal void EnableGroupStyleWarningState(string warningMessage)
+        {
```
QilongTang, line 1827: Why is group style involved..?

src/DynamoCoreWpf/ViewModels/PackageManager/PublishPackageViewModel.cs
```diff
@@ -696,6 +696,40 @@ public PackageUploadRequestBody BaseVersionHeader
             }
         }
 
+        private bool isWarningEnabled;
+        /// <summary>
+        /// This flag will be in true when the Style that user is trying to add already exists (otherwise will be false - Default)
```
zeusongit, line 701: wrong comment?

src/DynamoCore/Search/SearchDictionary.cs
jesusalvino: > Is there a .NET API which can return these?

I was tented to use the function Path.GetInvalidFileNameChars but it returns more characters that we need

imagen

jesusalvino: > So, I think this could be problematic, what happens when a user wants to publish a new version for a package that already has spaces in the name, will that get flagged? Also, why are we invalidating spaces?

As I mention, the previous version of the function invalids the space, I will remove it.

src/DynamoCoreWpf/UI/Themes/Modern/DynamoModern.xaml
jesusalvino: > What are the changes to this style?

adds an extra space to the right side of the input to draw the Icon that will be displayed if required

src/DynamoCore/Search/SearchDictionary.cs
QilongTang: The `Path.GetInvalidFileNameChars` looks solid to me, and it seems it includes space as well? @zeusongit I am confused, Do we allow space before even though it was included as the invalid char check?

src/DynamoCoreWpf/ViewModels/PackageManager/PublishPackageViewModel.cs
jesusalvino: Done.

src/DynamoCoreWpf/ViewModels/PackageManager/PublishPackageViewModel.cs
jesusalvino: Done.

src/DynamoCoreWpf/ViewModels/PackageManager/PublishPackageViewModel.cs
jesusalvino: Done.

src/DynamoCore/Search/SearchDictionary.cs
jesusalvino: @QilongTang Actually the ContainsSpecialCharacters function hadn't been used before this branch, at least in Dynamo solution.

QilongTang: Perfect, let's replace it

QilongTang:
Regression also to look at
```
[Dynamo.Tests.Search.SearchDictionaryTest.TestContainsSpecialCharacters](https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/5684/testReport/junit/Dynamo.Tests.Search/SearchDictionaryTest/TestContainsSpecialCharacters/)
```

src/DynamoCore/Search/SearchDictionary.cs
jesusalvino: Done

src/DynamoCore/Search/SearchDictionary.cs
```diff
@@ -323,11 +323,33 @@ private static string[] SplitOnWhiteSpace(string s)
...
+                    filteredChars.Add(someChar.ToString());
+                }
+            }
+
+            char[] result = new char[filteredChars.Count];
+            for (int i = 0; i < filteredChars.Count; i++)
+            {
+                result[i] = char.Parse(filteredChars[i]);
+            }
+            return result;
```
QilongTang, line 344: Two suggestions I have:
- An easier way may I suggest is to convert the char[] to string and replace the space with string.null and then convert it back to char[].
-  Also currently this SpecialAndInvalidCharacters will be calculated runtime everytime we publish a package, can we initialize it as a static property only once and get back to it everytime we are doing package name check?

jesusalvino: The SpecialAndInvalidCharacters function was reduced to a single line

QilongTang (approved):

zeusongit:
So just to be clear this will not affect the packages which already have spaces and want to publish a new version, but will prohibit future package names to include spaces, right? @QilongTang @jesusalvino 
I think we should also make a special note of it in the next release.

mjkkirschner:
> So just to be clear this will not affect the packages which already have spaces and want to publish a new version, but will prohibit future package names to include spaces, right? @QilongTang @jesusalvino I think we should also make a special note of it in the next release.

What about users publishing new versions of existing packages with spaces? Won't this stop them from updating existing packages?

src/DynamoCore/Search/SearchDictionary.cs
```diff
@@ -323,11 +323,18 @@ private static string[] SplitOnWhiteSpace(string s)
             return s.Split(null);
         }
 
-        private static bool ContainsSpecialCharacters(string element)
+        public static Char[] SpecialAndInvalidCharacters()
         {
-            return element.Contains("*") || element.Contains(".") || element.Contains(" ")
-                || element.Contains("\\");
+            // Excluding white spaces and uncommon characters, only keeping the displayed in the Windows alert
```
QilongTang, line 328: FYI.. @zeusongit @mjkkirschner see this comment, space is excluded from .NET API

QilongTang:
> > So just to be clear this will not affect the packages which already have spaces and want to publish a new version, but will prohibit future package names to include spaces, right? @QilongTang @jesusalvino I think we should also make a special note of it in the next release.
> 
> What about users publishing new versions of existing packages with spaces? Won't this stop them from updating existing packages? - same question as @zeusongit 👍

See my comment in code

src/DynamoCore/Search/SearchDictionary.cs
```diff
@@ -323,11 +323,18 @@ private static string[] SplitOnWhiteSpace(string s)
             return s.Split(null);
         }
 
-        private static bool ContainsSpecialCharacters(string element)
+        public static Char[] SpecialAndInvalidCharacters()
```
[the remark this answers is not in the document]
jesusalvino: The reason why it's not private is because this function is reused in the PublishPackageViewModel in order to display to the user the invalid characters.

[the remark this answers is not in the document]
QilongTang: I see, thanks!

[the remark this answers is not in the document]
reddyashish: Can we make it internal then?

src/DynamoCore/Search/SearchDictionary.cs
zeusongit: Thanks for the clarification @QilongTang 👍🏼

zeusongit (approved):

src/DynamoCore/Search/SearchDictionary.cs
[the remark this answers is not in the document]
jesusalvino: > Can we make it internal then?

Done.

reddyashish (approved):

QilongTang (approved):

QilongTang:
@jesusalvino Looks like there are still regressions to deal with before we can merge

src/DynamoCore/Configuration/PathManager.cs
```diff
@@ -131,6 +131,29 @@ internal static string BuiltinPackagesDirectory
             }
         }
 
+        private static char[] invalidCharacters = null;
+        /// <summary>
+        /// List of the Windows invalid characters
+        /// </summary>
+        internal static char[] InvalidCharacters
```
QilongTang, line 138: Any reference to this property in this PR yet?

src/DynamoCore/Configuration/PathManager.cs
```diff
@@ -131,6 +131,29 @@ internal static string BuiltinPackagesDirectory
...
+        {
+            get
+            {
+                if (invalidCharacters == null)
+                {
+                    invalidCharacters = Search.SearchDictionary<object>.SpecialAndInvalidCharacters();
```
QilongTang, line 144: Looking at this PR again, I still would not prefer this way. The original getter of invalid chars lives inside of search dictionary which seems strange. Can you move this to https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoUtilities/PathHelper.cs and have `ContainsSpecialCharacters` refer to this property?

src/DynamoCore/Search/SearchDictionary.cs
```diff
@@ -323,11 +323,18 @@ private static string[] SplitOnWhiteSpace(string s)
             return s.Split(null);
         }
 
+        internal static Char[] SpecialAndInvalidCharacters()
```
QilongTang, line 326: I would delete this function

QilongTang (changes requested):
More changes requested @jesusalvino

src/DynamoCore/Configuration/PathManager.cs
```diff
@@ -131,6 +131,29 @@ internal static string BuiltinPackagesDirectory
...
+                }
+                return invalidCharacters;
+            }
+            set
+            {
+                if (invalidCharacters != value)
+                {
+                    invalidCharacters = value;
+                }
+            }
```
QilongTang, line 154: I dont think we need the setter here. One step forward, maybe we can just remove these since we already have the static function in `PathHelper`, what do you think?

QilongTang (approved):

PR #12813, opened 2022-04-20
fix iron python's API breaking change

pinzart90:
https://jira.autodesk.com/browse/DYN-4877
ISsue:
Latest (2.14.0) version of the IronPython package is not compatible with DynamoCore2.13

### Purpose

Fix API breaking change for the IronPython package

### Declarations

### Release Notes

Fix IronPython breaking API change

### Reviewers

### FYIs

### DIFF
--- a/src/Libraries/DSIronPython/pkg.json
+++ b/src/Libraries/DSIronPython/pkg.json
@@ -1,0 +1,19 @@
+{
+    "license": "",
+    "file_hash": null,
+    "name": "DynamoIronPython2.7",
+    "version": "2.5.0",
+    "description": "*** This is the official version of the IronPython Extension which will load the IronPython2 evaluation engine for Dynamo. *** \r\n\r\nIf you see the following warning 'The configured Python engine could not be found' when you try to run a Python Script node which is set to use IronPython2, the host application for Dynamo (Such as Revit or Civil 3d) will have chosen to exclude IronPython2 as default installed content and you will need to download this extension in order to enable IronPython2 again.\r\n\r\nYou have the ability to either download this IronPython2 Extension or use the alternative out-of-the-box CPython3 engine and migrate your legacy IronPython2 code to CPython3. While there are differences in language the Python basis is the same and we have provided a Migration Assistant to help you migrate your code to Python3.",
+    "group": "",
+    "keywords": [ "python", "ironpython" ],
+    "dependencies": [],
+    "host_dependencies": [ "Revit", "Civil 3D", "Alias", "Advance Steel", "FormIt" ],
+    "contents": "",
+    "engine_version": "2.13.0",
+    "engine": "dynamo",
+    "engine_metadata": "This pacakge is compatible with Dynamo2.13.0 release or newer",
+    "site_url": "https://dynamobim.org/",
+    "repository_url": "https://github.com/DynamoDS/Dynamo",
+    "contains_binaries": true,
+    "node_libraries": []
+}
--- a/src/NodeServices/PythonServices.cs
+++ b/src/NodeServices/PythonServices.cs
@@ -193,7 +193,7 @@ namespace Dynamo.PythonServices
             }
             catch (Exception e)
             {
-                System.Diagnostics.Debug.WriteLine($"Failed to load {CPythonAssemblyName} from {a.GetName().Name} with error {e.Message}");
+                System.Diagnostics.Debug.WriteLine($"Failed to load {CPythonAssemblyName} with error: {e.Message}");
             }
         }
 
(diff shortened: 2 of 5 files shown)


mjkkirschner:
@pinzart90 I think this will break compatibility with Dynamo 2.14 which is already released, can we check for both somehow?

pinzart90:
> @pinzart90 I think this will break compatibility with Dynamo 2.14 which is already released, can we check for both names somehow?

PythonEngineManager.IronPython2EngineName returns the same string ("IronPython2")
It should not break compatibility with 2.14 or the latest Dynamo 

I can check/test anyway

mjkkirschner:
Ah, I see, so the API break is not the name change, but the addition of the new property? 
`IronPython2EngineName` I guess I would have expected to see a missing member exception or something in the logs where the package fails to work.

mjkkirschner (approved):
looks good, I agree though that if you have time, exposing the exception to the log would be useful.

src/NodeServices/PythonServices.cs
```diff
@@ -218,9 +231,9 @@ private void LoadPythonEngine(Assembly assembly)
                     AvailableEngines.Add(engine);
                 }
             }
-            catch
+            catch(Exception ex)
             {
-                //Do nothing for now
+                throw new Exception($"Failed to add a Python engine from assembly {assembly.GetName().Name} with error {ex.Message}");
```
pinzart90, line 236: PackageLoader should catch this exception

src/NodeServices/PythonServices.cs
```diff
@@ -173,24 +173,36 @@ private PythonEngineManager()
 
             // We check only for the default python engine because it is the only one loaded by static references.
             // Other engines can only be loaded through package manager
-            LoadPythonEngine(AppDomain.CurrentDomain.GetAssemblies().
-                FirstOrDefault(a => a != null && a.GetName().Name == CPythonAssemblyName));
+            LoadDefaultPythonEngine(AppDomain.CurrentDomain.GetAssemblies().
+               FirstOrDefault(a => a != null && a.GetName().Name == CPythonAssemblyName));
 
-            AppDomain.CurrentDomain.AssemblyLoad += new AssemblyLoadEventHandler((object sender, AssemblyLoadEventArgs args) =>
+            AppDomain.CurrentDomain.AssemblyLoad += new AssemblyLoadEventHandler((object sender, AssemblyLoadEventArgs args) => LoadDefaultPythonEngine(args.LoadedAssembly));
```
mjkkirschner, line 179: what is this syntax called?

src/Libraries/DSIronPython/pkg.json
```diff
@@ -0,0 +1,19 @@
...
+    "host_dependencies": [ "Revit", "Civil 3D", "Alias", "Advance Steel", "FormIt" ],
+    "contents": "",
+    "engine_version": "2.13.0",
+    "engine": "dynamo",
+    "engine_metadata": "This pacakge is compatible with Dynamo2.13.0 release or newer",
+    "site_url": "https://dynamobim.org/",
+    "repository_url": "https://github.com/DynamoDS/Dynamo",
+    "contains_binaries": true,
+    "node_libraries": []
+}
```
pinzart90, line 19: @Amoursol @mjkkirschner any opinions on what needs to be communicated through the pkg json ?

src/Libraries/DSIronPython/pkg.json
```diff
@@ -0,0 +1,19 @@
...
+    "group": "",
+    "keywords": [ "python", "ironpython" ],
+    "dependencies": [],
+    "host_dependencies": [ "Revit", "Civil 3D", "Alias", "Advance Steel", "FormIt" ],
+    "contents": "",
+    "engine_version": "2.13.0",
+    "engine": "dynamo",
+    "engine_metadata": "This pacakge is compatible with Dynamo2.13.0 release or newer",
```
mjkkirschner, line 14: is this a real property that will make it up to the package manager, or this is just a comment?

pinzart90, line 14: I just added this but I am not sure who/what consumes the engine_metadata...

mjkkirschner: @zeusongit any idea?

pinzart90: I can move it to the description if you think it is valuable info ...

pinzart90: Ok, any opinions on the wording ?
I was thinking 
```
-------------------------------------------------
DynamoIronPython2.7 compatibility matrix
-------------------------------------------------
IronPython version    |    Dynamo version      |
|            2.1.0              |        [2.13.0, 2.13.x]      |
|            2.4.0              |        >=2.14.0              |
|            2.5.0              |        >=2.13.0              |
--------------------------------------------------
```

mjkkirschner: 👍

pinzart90: I am not sure how backwards compatible these versions are...
Should we try to find the minimum supported Dynamo versions ? 
Or is 2.13.0 good enough as a minimum supported version ?
Also should I add the rest of the versions too ? 1.0.0 and 2.0.0 ?
Also should I say something about the broken versions ? ...2.2.0 and 2.3.0 ?

aparajit-pratap: If you specify engine version to be `"engine_version": "2.13.0"`, doesn't it mean that the package is compatible with Dynamo version >= 2.13.0?

mjkkirschner: I believe the only thing it really means, is that the package was published from that version of Dynamo.

pinzart90: We do show a warning in Dynamo if the pkg engine_vers is greater than Dynamo's version
https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs#L812

pinzart90: also there is no good way to show the compatibility matrix as I have created it here. The description is just string from a JSON. I would have to use a simpler way to get our point across

mjkkirschner: can't the string in json contain escape characters?
https://ozh.github.io/ascii-tables/

zeusongit: I have not seen this property used anywhere

pinzart90: they can...but the html side of things will need some work to be able to interpret those escaped characters. 
I think the style of some of the tags needs to change

pinzart90: This is taking too much time. I propose that I publish the new package with just the engine property updated to reflect the compatibility (that would be pkg.engine_version: "2.13.0").
I can later update the description only (I think there is a way to just update the metadata of a package for super users)

pinzart90: @Amoursol @mjkkirschner do you agree ?

Amoursol: @pinzart Agreed, let's do a partial fix now to get this clearer for our users, and do a holistic one afterwards.

PR #12815, opened 2022-04-20
DYN-4800: Bump Python version to 3.9.12

sm6srw:
### Purpose

This pull request does:
* Bump pythonnet to `2.5.2`
* Bump embedded python to `3.9.12`

### Declarations

### Release Notes

### Reviewers

### FYIs


src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
```diff
@@ -376,6 +376,8 @@ public DSCPythonCodeCompletionProviderCore()
             BasicVariableTypes.Add(Tuple.Create(LIST_VARIABLE, typeof(PyList)));
             BasicVariableTypes.Add(Tuple.Create(DICT_VARIABLE, typeof(PyDict)));
 
+            Python.Included.Installer.SetupPython().Wait();
```
mjkkirschner, line 379: why is this necessary? The DSCPython engine should do this already, should it not?
I think it's worth testing this change in Revit context if it's really required.

pinzart90: There is a note in Dynamo
```
NOTE: Calling SetupPython multiple times will add the install location to the path many times,
        /// potentially causing the environment variable to overflow.
```
I think we might be doing this already ...maybe we should try to only run `SetupPython()` only once ...(keep a isAlreadySetup flag somewhere)
Please search for all occurrences of `SetupPython()`

sm6srw: I actually was able to get in here without having python initialized. And we are already calling this method in multiple places. I will take a look.

src/Libraries/DSCPython/CPythonEvaluator.cs
```diff
@@ -336,7 +336,7 @@ import os
         /// NOTE: Calling SetupPython multiple times will add the install location to the path many times,
         /// potentially causing the environment variable to overflow.
         /// </summary>
-        private static void InstallPython()
+        internal static void InstallPython()
```
sm6srw, line 339: I made it internal. Would it be OK to make this public?

src/Libraries/DSCPython/DSCPythonCodeCompletionProviderCore.cs
sm6srw: We are now calling SetupPython() every where. That method maintain a just do it once flag.

sm6srw:
@mjkkirschner PTAL!

src/Libraries/DSCPython/CPythonEvaluator.cs
mjkkirschner: since it has a side effect that is potentially dangerous I'd prefer if extensions could not call it.

mjkkirschner (approved):

pinzart90 (approved):

PR #12819, opened 2022-04-21
Set the default color in the color palette when using node autocomplete.

reddyashish:
### Purpose

This is a followup to the task: https://jira.autodesk.com/browse/DYN-4753.

When the color palette node is placed using the node autocomplete, the default color(black) was not being set. 

The reason for this was that there was an exception being thrown in here: https://github.com/DynamoDS/Dynamo/blob/master/src/Libraries/CoreNodeModelsWpf/NodeViewCustomizations/ColorPalette.cs#L42 and the view customization would not finish. 
The exception is a custom one that is related to UndoRedo operation:
https://github.com/DynamoDS/Dynamo/blob/a496720a5ee33b4d7ed93d358eed560121bc9ac8/src/DynamoCore/Core/UndoRedoRecorder.cs#L267 

The undo redo operation around this node works fine and I see this WorkspaceModel.RecordModelForModification() being called only for couple other nodes. There is a TODO on that API, saying we should refactor that and move the code to the WorkspaceViewModel. For now, I made sure that this would be the last step in node view customization so that everything else will be executed even if this throws an exception. We could also catch this exception and handle it accordingly. Any thoughts?

### Declarations

### Release Notes

### Reviewers
@QilongTang @zeusongit

### DIFF
--- a/src/Libraries/CoreNodeModelsWpf/NodeViewCustomizations/ColorPalette.cs
+++ b/src/Libraries/CoreNodeModelsWpf/NodeViewCustomizations/ColorPalette.cs
@@ -39,9 +39,6 @@ namespace CoreNodeModelsWpf.Nodes
 
             //kick off ui to match initial model state.
             this.ColorPaletteNode_PropertyChanged(ColorPaletteUINode, new PropertyChangedEventArgs("DsColor"));
-
-            var undoRecorder = viewNode.ViewModel.WorkspaceViewModel.Model.UndoRecorder;
-            WorkspaceModel.RecordModelForModification(colorPaletteNode, undoRecorder);
         }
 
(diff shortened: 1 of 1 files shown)


QilongTang:
I am fine with this being a temp fix, it seems this issue exist for all node customization applied to node from autocomplete though. @zeusongit what do you think?

zeusongit:
I think the fix is good enough for now, but we should file a task to handle the exception and not add another undo recorder that throws the exception.

reddyashish:
@QilongTang @zeusongit I have removed that API call and tested it locally. This is specific to the color palette node.

reddyashish:
All tests passed: https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/969/

QilongTang (approved):

PR #12824, opened 2022-04-25
Fix crash while installing package from view extension

reddyashish:
### Purpose

This fixes the crash mentioned in the JIRA: https://jira.autodesk.com/browse/DYN-4848 

The crash was happening due to the owner object not being loaded when installing the package from view extension. The owner is the package manager search view and this was happening as the package manager search dialog is closed before installing the package. So when the owner is not loaded, initializing the dialog box with the owner object would fail. We will be setting the dialog box's owner only when the owner is loaded.

### Declarations

### Release Notes

Fix crash while installing package from view extension

### Reviewers
@QilongTang @zeusongit


reddyashish:
All tests passed here: https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/

PR #12826, opened 2022-04-25
Updating the colors of the ports according to the Function State

filipeotero:
### Purpose

This PR contains the code improvement that changes the colors of the input/output of the nodes by checking the class and library of the object to define if the node is a 'Function'.

colorInputOutputFunction

### Declarations

### Reviewers

@mjkkirschner @QilongTang @Amoursol 

### FYIs

@RobertGlobant20 @jesusalvino

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Core/InPortViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/InPortViewModel.cs
@@ -22,10 +22,11 @@ namespace Dynamo.ViewModels
         private SolidColorBrush portValueMarkerColor = new SolidColorBrush(Color.FromArgb(255, 204, 204, 204));
 
         private bool portDefaultValueMarkerVisible;
+        private bool isReturningDefaultValue;
 
         private static SolidColorBrush PortValueMarkerBlue = new SolidColorBrush(Color.FromRgb(106, 192, 231));
         private static SolidColorBrush PortValueMarkerRed = new SolidColorBrush(Color.FromRgb(235, 85, 85));
-        private static SolidColorBrush PortValueMarkerGrey = new SolidColorBrush(Color.FromRgb(153, 153, 153));        
+        private static SolidColorBrush PortValueMarkerGrey = new SolidColorBrush(Color.FromRgb(153, 153, 153));
 
         private static readonly SolidColorBrush PortBackgroundColorKeepListStructure = new SolidColorBrush(Color.FromRgb(83, 126, 145));
         private static readonly SolidColorBrush PortBorderBrushColorKeepListStructure = new SolidColorBrush(Color.FromRgb(168, 181, 187));
@@ -151,8 +152,19 @@ namespace Dynamo.ViewModels
         public InPortViewModel(NodeViewModel node, PortModel port) : base(node, port)
         {
             port.PropertyChanged += PortPropertyChanged;
+            node.NodeModel.PropertyChanged += NodeModel_PropertyChanged;
 
             RefreshPortDefaultValueMarkerVisible();
+        }
+
+        private void NodeModel_PropertyChanged(object sender, System.ComponentModel.PropertyChangedEventArgs e)
+        {
+            switch (e.PropertyName)
+            {
+                case nameof(node.NodeModel.CachedValue):
+                    RefreshPortColors();
+                    break;
+            }
         }
 
         public override void Dispose()
@@ -162,7 +174,7 @@ namespace Dynamo.ViewModels
         }
 
         internal override PortViewModel CreateProxyPortViewModel(PortModel portModel)
-        {            
+        {
             portModel.IsProxyPort = true;
             return new InPortViewModel(node, portModel);
         }
@@ -211,7 +223,7 @@ namespace Dynamo.ViewModels
             var useLevel = (bool)parameter;
             var command = new DynamoModel.UpdateModelValueCommand(
                 Guid.Empty, node.NodeLogic.GUID, "UseLevels", string.Format("{0}:{1}", port.Index, useLevel));
-            
+
             node.WorkspaceViewModel.DynamoViewModel.ExecuteCommand(command);
             RaisePropertyChanged(nameof(UseLevelSpinnerVisible));
         }
@@ -233,7 +245,7 @@ namespace Dynamo.ViewModels
             var keepListStructure = (bool)parameter;
             var command = new DynamoModel.UpdateModelValueCommand(
                 Guid.Empty, node.NodeLogic.GUID, "KeepListStructure", string.Format("{0}:{1}", port.Index, keepListStructure));
-            
+
             node.WorkspaceViewModel.DynamoViewModel.ExecuteCommand(command);
         }
 
@@ -276,9 +288,16 @@ namespace Dynamo.ViewModels
         /// </summary>
         protected override void RefreshPortColors()
         {
-            bool isPythonNode = node.NodeModel is PythonNodeModels.PythonNode;
-            //Is in function state
-            if (node.NodeModel.IsPartiallyApplied && !isPythonNode)
+            //Boolean that defines if the node has a default value from the cache value
+            isReturningDefaultValue = !string.IsNullOrEmpty(node.NodeModel.CachedValue?.StringData);
+            
+            //This variable checks if the node is a function class
+            bool isFunctionNode = node.NodeModel.CachedValue != null &&
+                                    node.NodeModel.CachedValue.Data == null &&
+                                    !node.NodeModel.CachedValue.IsNull &&
+                                    node.NodeModel.CachedValue.Class != null;
+
+            if (isFunctionNode)
             {
                 if (node.NodeModel.AreAllOutputsConnected)
                 {
@@ -288,12 +307,12 @@ namespace Dynamo.ViewModels
                 }
                 else
                 {
-                    SetupDefaultPortColorValues();
+                    SetupDefaultPortColorValues(isFunctionNode);
                 }
             }
             else
             {
-                SetupDefaultPortColorValues(isPythonNode);
+                SetupDefaultPortColorValues();
             }
         }
 
@@ -305,15 +324,16 @@ namespace Dynamo.ViewModels
                 {
                     PortValueMarkerColor = PortValueMarkerGrey;
                 }
-                else 
+                else
                 {
                     SetupDefaultPortColorValues();
                 }
             }
         }
 
-        private void SetupDefaultPortColorValues(bool isPythonNode = false)
+        private void SetupDefaultPortColorValues(bool isFunctionNode = false)
         {
+            
             // Special case for keeping list structure visual appearance
             if (port.UseLevels && port.KeepListStructure && port.IsConnected)
             {
@@ -322,7 +342,7 @@ namespace Dynamo.ViewModels
                 PortBorderBrushColor = PortBorderBrushColorKeepListStructure;
             }
             // Port has a default value, shows blue marker
-            else if (UsingDefaultValue && DefaultValueEnabled || isPythonNode)
+            else if (UsingDefaultValue && DefaultValueEnabled || (isReturningDefaultValue && !isFunctionNode))
             {
                 PortValueMarkerColor = PortValueMarkerBlue;
                 PortBackgroundColor = PortBackgroundColorDefault;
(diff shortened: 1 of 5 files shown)


mjkkirschner:
As I mentioned in the thread, this is not unique to python nodes, but any class that inherits from NodeModel.

Amoursol:
> > As I mentioned in the thread, this is not unique to python nodes, but any class that inherits from NodeModel.
> 
> Sorry, I've understood that this is an issue that happens for Python Nodes because it has default values when created. Could you be more specific? Do we need to check the output result of the node then? If yes, is there any specific property that returns the type of the return?

@filipeotero as Mike suggested, this same problem will exist on other nodes. Take the following example of the "Color Range" node:

image

This has an issue with the "Function" grey-bar output also. It also has an issue with the input vertical bars too which we didn't catch for Global Launch :( Essentially, the input ports are also satisfied, but show as unsatisfied.

FYI @QilongTang

filipeotero:
Updated the code to check the Result of the node by getting the `CachedValue` of the model.
Some examples of the nodes after this implementation:

image

@Amoursol Are there more examples to be covered for this issue? Thanks!
FYI: @QilongTang @mjkkirschner

src/DynamoCoreWpf/ViewModels/Core/InPortViewModel.cs
```diff
@@ -276,8 +288,16 @@ private void RefreshPortDefaultValueMarkerVisible()
         /// </summary>
         protected override void RefreshPortColors()
         {
-            //Is in function state
-            if (node.NodeModel.IsPartiallyApplied)
+            //Boolean that defines if the node has a default value from the cache value
+            isReturningDefaultValue = !string.IsNullOrEmpty(node.NodeModel.CachedValue?.StringData);
```
mjkkirschner, line 292: I don't understand this logic, many nodes return strings?

src/DynamoCoreWpf/ViewModels/Core/InPortViewModel.cs
```diff
@@ -276,8 +288,16 @@ private void RefreshPortDefaultValueMarkerVisible()
...
+            
+            //This variable checks if the node is a function class
+            bool isFunctionNode = node.NodeModel.CachedValue != null &&
```
mjkkirschner, line 295: I think - there are better options - this is too fragile and could break. Let me look.

src/DynamoCoreWpf/ViewModels/Core/OutPortViewModel.cs
```diff
@@ -294,8 +305,14 @@ private void OnMouseLeftButtonDownOnContext(object parameter)
         }
 
         protected override void RefreshPortColors()
-        {            
-            if (node.NodeModel.IsPartiallyApplied)
+        { 
+            //This variable checks if the node is a function class
+            bool isFunctionNode = node.NodeModel.CachedValue != null &&
```
mjkkirschner, line 310: you should put this somewhere you can share it - but lets find something better to replace it with.

mjkkirschner: @aparajit-pratap do you know of any nice ways to check if the `NodeModel.CachedValue and its underlying stackValue` is a DS Function?

It appears that `StackValue` actually has a `IsFunctionPointer` and `IsFunctionIndex` method - would those be useful to expose in MirrorData as a public method - like `IsFunction`?

https://github.com/DynamoDS/Dynamo/blob/master/src/Engine/ProtoCore/Reflection/MirrorData.cs#L275

What is the actual object type when we see a `Function` in the preview bubble, is it a FunctionObject class?

aparajit-pratap: The actual object is a DesignScript class called `Function`, which is defined in `FunctionObject.ds`. The stackvalue will most likely be an object pointer.

src/DynamoCoreWpf/ViewModels/Core/InPortViewModel.cs
filipeotero: Instead of getting the info of a Function State from isPartiallyApplied, I'm getting the info from the CachedValue because It's what is appearing in the little console below the node. The idea was to identify if the node has a result, even though It's a null result. 
Another thing is that I had to differentiate between the output 'Function' (in which the inputs have to change the color between red/blue and the output grey) and other ones (inputs always blue and output disabled). 
I would love to hear any other suggestions from you guys!
Thanks.

src/DynamoCoreWpf/ViewModels/Core/OutPortViewModel.cs
filipeotero: Awesome, I will add that in the next commit. Also, I think we can remove the `isReturningDefaultValue ` boolean as it's looking quite confusing and appears to have the same objective.

src/DynamoCoreWpf/ViewModels/Core/InPortViewModel.cs
aparajit-pratap: I don't think we need this flag `isReturningDefaultValue`.

src/DynamoCoreWpf/ViewModels/Core/InPortViewModel.cs
```diff
@@ -321,7 +342,7 @@ private void SetupDefaultPortColorValues()
                 PortBorderBrushColor = PortBorderBrushColorKeepListStructure;
             }
             // Port has a default value, shows blue marker
-            else if (UsingDefaultValue && DefaultValueEnabled)
+            else if (UsingDefaultValue && DefaultValueEnabled || (isReturningDefaultValue && !isFunctionNode))
```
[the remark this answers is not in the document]
mjkkirschner: I don't understand how you can prove anything about the default value from the return value of the output ports -

I think it's important to point out that the python node executes because it doesn't NEED its inputs to be satisfied to execute, not because it has default values, it doesn't have default values for its ports, nor do other explicit NodeModel nodes, like ColorRange etc. 

Consider an arbitrary function in a familiar language like javascript, you can call a function and not pass all the arguments, the function still executes.

I'm not yet getting how default values factor into the requirements.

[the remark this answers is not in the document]
mjkkirschner: just saw your comment above @filipeotero regarding getting rid of `isReturningDefaultValue` - sounds good to me.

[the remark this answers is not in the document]
aparajit-pratap: I think, more strictly speaking, @filipeotero is trying to determine if a node's input ports are "satisfied" (not necessarily default) by inspecting the node output after it has run, and if it returns a non-`Function` value, he's assuming all ports to be satisfied, whereas if it returns a `Function`, for them to be unsatisfied (if not connected). @filipeotero can you confirm if I'm understanding this correctly.

src/Engine/ProtoCore/Reflection/MirrorData.cs
```diff
@@ -21,6 +21,14 @@ public class MirrorData
             /// </summary>
             private StackValue svData;
 
+            public bool IsFunction
+            {
+                get
+                {
+                    return svData.IsPointer && Class.Name.Equals("Function") && Class.LibraryMirror.Name.Equals("FunctionObject.ds");
```
aparajit-pratap, line 28: Do we have constants already defined for "Function" and "FunctionObject.ds"? Can you search for these in the codebase? If not, I think we should create them in a common utility class and make them accessible in the `ProtoCore` project at least.

src/DynamoCoreWpf/ViewModels/Core/InPortViewModel.cs
```diff
@@ -276,8 +290,11 @@ private void RefreshPortDefaultValueMarkerVisible()
         /// </summary>
         protected override void RefreshPortColors()
         {
-            //Is in function state
-            if (node.NodeModel.IsPartiallyApplied)
+            //This variable checks if the node is a function class
+            isFunctionNode = node.NodeModel.CachedValue != null &&
+                node.NodeModel.CachedValue.IsFunction;
```
aparajit-pratap, line 295: Shouldn't this be:
```
isFunctionNode = node.NodeModel.IsPartiallyApplied || (node.NodeModel.CachedValue != null &&
                node.NodeModel.CachedValue.IsFunction);
```

src/DynamoCoreWpf/ViewModels/Core/OutPortViewModel.cs
aparajit-pratap: Same here, shouldn't this be:
```
bool isFunctionNode = node.NodeModel.IsPartiallyApplied || node.NodeModel.CachedValue != null && node.NodeModel.CachedValue.IsFunction;
```

src/Engine/ProtoCore/Reflection/MirrorData.cs
filipeotero: I didn't find any constants defined for that. Do you have a class in mind for that?

src/DynamoCoreWpf/ViewModels/Core/InPortViewModel.cs
filipeotero: `node.NodeModel.IsPartiallyApplied ` will not make sense for now because there are cases where the node doesn't return a function ever, like the python node.

src/DynamoCoreWpf/ViewModels/Core/InPortViewModel.cs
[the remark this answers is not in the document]
filipeotero: yes @aparajit-pratap, you are right.

src/DynamoCoreWpf/ViewModels/Core/InPortViewModel.cs
mjkkirschner: hmm, I actually think it should be
```
isFunctionNode = node.NodeModel.IsPartiallyApplied && (node.NodeModel.CachedValue != null &&
                node.NodeModel.CachedValue.IsFunction);```

DS is a dynamic language, we could probably write a function which returns another function, and it would now turn ports blue, even if it was not partially unsatisfied.

File: src/Engine/ProtoCore/Reflection/MirrorData.cs (hunk 5 above)
Comment (aparajit-pratap) [reply src/Engine/ProtoCore/Reflection/MirrorData.cs:28]: `ProtoCore.Utils.CoreUtils` maybe

File: src/DynamoCoreWpf/ViewModels/Core/InPortViewModel.cs (hunk 6 above)
Comment (aparajit-pratap) [reply src/DynamoCoreWpf/ViewModels/Core/InPortViewModel.cs:295]: @filipeotero the check was `if (node.NodeModel.IsPartiallyApplied)` earlier, so by removing this check, won't you break behavior? I think this check is necessary for those nodes for which you can determine their "unsatisfied" state even before execution.

Comment (aparajit-pratap) [reply src/DynamoCoreWpf/ViewModels/Core/InPortViewModel.cs]: I mean a case like this that was working earlier in manual mode will fail won't it:
default_value

File: src/Engine/ProtoCore/Reflection/MirrorData.cs (hunk 5 above)
Comment (filipeotero) [reply src/Engine/ProtoCore/Reflection/MirrorData.cs]: Ok, done. Thanks!

File: src/Engine/ProtoCore/Utils/CoreUtils.cs [hunk 7]
```diff
@@ -7,6 +7,9 @@ namespace ProtoCore.Utils
 {
     public static class CoreUtils
     {
+        public static readonly string NodeFunctionClassName = "Function";        
+        public static readonly string NodeLibraryMirronFunctionObjectName = "FunctionObject.ds";
```
aparajit-pratap, line 11: I would name them `FunctionObjectClass` and `FunctionObjectLibrary`.

filipeotero: Changing in the next commit... Thanks!

src/DynamoCoreWpf/ViewModels/Core/InPortViewModel.cs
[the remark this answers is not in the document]
Amoursol: Correct @filipeotero @aparajit-pratap - it's about port satisfaction. This simply means the node will run as intended. On non NodeModel Nodes, this happens when you wire in or unwire a connector. Doesn't have to wait for node execution.

QilongTang (approved):

QilongTang:
Triggered tests running at: https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/5912/

QilongTang:
No new regressions were introduced in this PR. Merging.

PR #12848, opened 2022-05-02
As a Dynamo integrator I want to control where my nodes are located in the library

mjkkirschner:
### Purpose
https://jira.autodesk.com/browse/DYN-3407

For built-in packages Dynamo host product integrators should have a way of modifying the library layout spec so that their packages show up in the main sections of the library, and not the "add-on" section.

This PR does the following:
* introduces a recursive `MergeLayoutSpecs` method. This method can merge sections and categories of two layout specs objects. It does not merge data, only child elements. For example, it won't combine `IncludeInfo` or `IconUrl` of elements, it will only merge their child element collections.
* builtin packages that are present on startup and contain a `pkg/extra/layoutspecs.json` file, will trigger a layout spec application request.
* If the `libraryViewExtension` is loaded , it will handle the request and attempt to merge the two layout specs.
* NodeSearchElements's loaded from builtin packages directory will now be marked with the `BuiltIn` flag, in addition to the `Packaged` flag.
* BuiltIn package items in the library will have the prefix `bltinpkg://` as opposed to `pkg://` for normal packages. This is the prefix that should be applied to the paths that integrators put in their layoutspec file.
* tests are added.
* LibraryViewExtensionTests.csproj is reenabled for build in debug and release, and I've added a few tests to it, I've also removed `LibraryViewExtension(CEF)` as a dep and instead pointed to `LibraryViewExtensionMSWebbrowser`. Most of the tests for the old library have been commented out as these simply don't make sense or don't compile. It's a bigger task to get these working again, and we'll be tossing the MSWebBrowser implementation soon anyway for WebView2 hopefully.
* PR to buildscripts so these tests run as part of master-15 build: https://git.autodesk.com/Dynamo/DynamoBuildscripts/pull/29

Here is an example of moving the zero touch dynamo samples package into a category named `Revit`

Screen Shot 2022-04-28 at 4 41 24 PM

### Declarations

### Release Notes

Internal API for Dynamo integrators to modify the library layout for their builtin packages.

### Reviewers
...


src/LibraryViewExtension/Handlers/NodeItemDataProvider.cs
```diff
@@ -84,8 +84,12 @@ protected virtual object CreateObjectForSerialization(IEnumerable<NodeSearchElem
         /// </summary>
         public static string GetFullyQualifiedName(NodeSearchElement element)
         {
+            if (element.ElementType.HasFlag(ElementTypes.Packaged) && element.ElementType.HasFlag(ElementTypes.BuiltIn))
```
pinzart90, line 87: why not add this condition inside the if at line 88 ?
```
if (element.ElementType.HasFlag(ElementTypes.Packaged))
{
   if (element.ElementType.HasFlag(ElementTypes.BuiltIn) {
       return string.Format("{0}{1}.{2}", "bltinpkg://", element.FullCategoryName, element.Name);
    }
    //Use FullCategory and name as read from _customization.xml file
                return string.Format("{0}{1}.{2}", "pkg://", element.FullCategoryName, element.Name);
```

mjkkirschner: I don't like using nested if statements if avoidable, if you much prefer your sample code will make the switch.

mjkkirschner:
latest tests finally ran and did not introduce any new regressions!

src/DynamoCoreWpf/Extensions/ViewExtensionManager.cs
```diff
@@ -65,13 +88,38 @@ private IViewExtension RequestLoadViewExtensionHandler(string extensionPath)
...
         }
 
+        private void RequestApplyLayoutSpecHandler(string specJSON)
+        {
+            Log($"an extension requested application of {specJSON} layout spec");
+              try
+                {
+                //try to combine the layout specs.
+                    var customizationService = extensionManager.Service<ILibraryViewCustomization>();
+                //TODO if the layoutspec is empty, we're calling this too early, we can retry after x seconds?
```
sm6srw, line 98: Have you seen this happening?

sm6srw (approved):
LGTM!

src/DynamoCoreWpf/Extensions/ViewExtensionManager.cs
mjkkirschner: yes, but only before when I was calling this code from the ExtensionsManager (before view extensions like libraryjs were loaded)

pinzart90 (approved):
LGTM

PR #12867, opened 2022-05-10
DYN-4767 run the graph

filipeotero:
### Purpose

This PR contains the Step 5 of the onboarding guide which includes executing the graph by highlighting the 'Run' button to the user.
It's also included in this PR the fix for the bug DYN-4931.

Step5RunGraph

### Declarations

### Reviewers

@QilongTang 

### FYIs

@RobertGlobant20


src/DynamoCoreWpf/UI/GuidedTour/GuidesValidationMethods.cs
```diff
@@ -713,6 +713,10 @@ internal static void ManagePortConnectionEvents(Step stepInfo, StepUIAutomation
 
         private static void OutputPortViewModel_PropertyChanged(object sender, System.ComponentModel.PropertyChangedEventArgs e)
         {
+            if (CurrentExecutingStep.UIAutomation == null
+                || CurrentExecutingStep.UIAutomation.Count() < 3
+                || CurrentExecutingStep.UIAutomation[2].Parameters == null) return;
```
QilongTang, line 718: I forgot to ask why do we target `UIAutomation[2]` specifically? This makses this function less generic seems

src/DynamoCoreWpf/UI/GuidedTour/GuidesValidationMethods.cs
```diff
@@ -743,44 +747,50 @@ private static void OutputPortViewModel_PropertyChanged(object sender, System.Co
...
             //If any backward action is triggered, the created needs to be deleted 
             if (currentFlow == GuideFlow.BACKWARD && lastCreatedNode != null)
             {
                 var stepMainWindow = CurrentExecutingStep.MainWindow as Window;
+
+                foreach (var connector in lastCreatedNode.AllConnectors.ToList())
+                {
+                    connector.Delete();
+                }
+
```
QilongTang, line 761: It's strange that you need to do this dispose specifically, I would expect the existing API on node or workspace level to handle it, I will also take a look at this

src/DynamoCoreWpf/UI/GuidedTour/GuidesValidationMethods.cs
```diff
@@ -872,5 +882,31 @@ private static NodeView GetNotConnectedNode(Window stepMainWindow, string nodeTy
             return byOriginUserAdded;
         }
 
+
+        internal static void SubscribeRunButtonNextStep(Step stepInfo, StepUIAutomation uiAutomationData, bool enableFunction, GuideFlow currentFlow)
+        {
+            CurrentExecutingStep = stepInfo;
+            if (uiAutomationData.Parameters.Any() && uiAutomationData.Parameters[0] is string)
```
RobertGlobant20, line 889: I think you should assign the uiAutomationData.Parameters[0] to a variable so you can use it in several parts of the function, also you should include a validation for checking that Parameters is not null and the number of parameter are more than 1 or equals than 2.

src/DynamoCoreWpf/UI/GuidedTour/GuidesValidationMethods.cs
```diff
@@ -872,5 +882,31 @@ private static NodeView GetNotConnectedNode(Window stepMainWindow, string nodeTy
...
+            {
+                Button runButton = GuideUtilities.FindChildInVisualTree(stepInfo.MainWindow, uiAutomationData.Parameters[0].ToString()) as Button;
+
+                if(runButton != null)
```
RobertGlobant20, line 893: For not having several nested if statements you can use something like:
if(runButton == null) return;

src/DynamoCoreWpf/UI/GuidedTour/GuidesValidationMethods.cs
filipeotero: It's in the sequence defined in the JSON file. 
Gets this parameter:
```
{
		"Sequence": 3,
		"ControlType": "function",
		"Name": "RemovePortConnector",
		"Parameters": [ "NextButton", "outputPortControl", "CoordinateSystem.ByOrigin" ],
		"Action": "execute"
	}
```

src/DynamoCoreWpf/UI/GuidedTour/GuidesValidationMethods.cs
QilongTang: I see now, there is API on NodelViewModel level but not on NodeModel level

src/DynamoCoreWpf/UI/GuidedTour/GuidesValidationMethods.cs
```diff
@@ -872,5 +882,31 @@ private static NodeView GetNotConnectedNode(Window stepMainWindow, string nodeTy
...
+                    }
+                    else
+                    {
+                        runButton.Click -= RunButton_Click;
+                    }
+                }
+            }
+        }
+
+        private static void RunButton_Click(object sender, RoutedEventArgs e)
```
RobertGlobant20, line 907: I just found that we can pass a complete class as argument to event handlers when you can pass the complete Step or the UIAutomation instance if is needed (so there is no need to use CurrentExecutingStep.UIAutomation[2]) , so you can use this feature instead of using CurrentExecutingGuide, see this example of passing a complete class as parameter, just check if is useful and also I there is no problem during the unsubscribe line.
image

RobertGlobant20 (approved):
LGTM with some comments

src/DynamoCoreWpf/UI/GuidedTour/GuidesValidationMethods.cs
QilongTang: Cool, is there a way to not hard code `UIAutomation[2]`? Is it in relation with current step number? e.g. current step number is 4 so that we should look for `sequence 3`

filipeotero: Good tip! I will look for the sequence. Thanks.

src/DynamoCoreWpf/UI/GuidedTour/GuidesValidationMethods.cs
filipeotero: In this case, I'm expecting one parameter. I will assign it to a variable and check if it's null. Thanks!

src/DynamoCoreWpf/UI/GuidedTour/GuidesValidationMethods.cs
```diff
@@ -713,7 +713,15 @@ internal static void ManagePortConnectionEvents(Step stepInfo, StepUIAutomation
 
         private static void OutputPortViewModel_PropertyChanged(object sender, System.ComponentModel.PropertyChangedEventArgs e)
         {
-            var buttonName = CurrentExecutingStep.UIAutomation[2].Parameters[0] as string;
+            if (CurrentExecutingStep.UIAutomation == null) return;
+
+            var currentUiAutomation = CurrentExecutingStep.UIAutomation.FirstOrDefault(x => x.Sequence == 3);
```
QilongTang, line 718: This still seems a bit hardcoded specifically for this step, is there a way to make this function not specific to this step?

filipeotero: I updated the code by searching for its method name. I guess in this way it will be less hardcoded. Please, let me know your opinion.

reddyashish (approved):

QilongTang:
There is no real regressions

QilongTang (approved):

PR #12893, opened 2022-05-17
Adjust NodeLibrary column header width

zeusongit:
### Purpose

Increase max-width for NodeLibrary column header

Screen Shot 2022-05-17 at 1 08 17 PM

English:
Screen Shot 2022-05-18 at 1 15 15 PM

Chinese:
Screen Shot 2022-05-18 at 1 19 36 PM

Russian:
Screen Shot 2022-05-18 at 1 22 20 PM

### Declarations

### Release Notes

Increase max-width for NodeLibrary column header

### DIFF
--- a/src/DynamoCoreWpf/Views/PackageManager/PublishPackageView.xaml
+++ b/src/DynamoCoreWpf/Views/PackageManager/PublishPackageView.xaml
@@ -706,7 +706,7 @@
                           Style="{StaticResource DataGrid}">
                     <DataGrid.Columns>
                         <DataGridCheckBoxColumn MinWidth="85"
-                                                MaxWidth="85"
+                                                MaxWidth="130"
                                                 Binding="{Binding Path=IsNodeLibrary, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}"
                                                 ElementStyle="{StaticResource CheckBoxStyle}"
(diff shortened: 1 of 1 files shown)


reddyashish:
This one: https://github.com/DynamoDS/Dynamo/blob/d852c2576eb5e2aa0e25ad1f0fe5818b335a7a56/src/DynamoCoreWpf/UI/Converters.cs#L1262

zeusongit:
@QilongTang @reddyashish The min and max value was fixed before, I increased the max width as per the longest localized string (FR) here. The width values is the converter are in a different range of 300-400, but the width used here are in 130-150 range. Therefore the same converter cannot be used here. I can create a new converter with new range, but I have tested the change in other languages and it seems to work fine for now.

QilongTang:
> @QilongTang @reddyashish The min and max value was fixed before, I increased the max width as per the longest localized string (FR) here. The width values in the converter are in a different range of 300-400, but the width used here are in 130-150 range. Therefore the same converter cannot be used here. I can create a new converter with new range, but I have tested the change in other languages and it seems to work fine for now.

Cool, thank you for the clarification! Glad this way it works OOTB

QilongTang (approved):

zeusongit:
Attached more screenshots in some other languages

reddyashish (approved):

reddyashish (approved):

PR #12895, opened 2022-05-18
JIL Execution Cleanup and Speed Optimizations Part 1

saintentropy:
### Purpose

JIL Execution in the VM is responsible for a small but important subset of node execution.  While most nodes today are executed via the FFI (Foreign Function Interpreter) mechanisms which invoke C# code, some basic utility methods especially in Math category are written in native design script.  This type of function is handled by the `JILFunctionEndPoint`.  While many of these enhancements are targeted at this execution path, they also yield net improvements to the general execution of all EndPoint types.  This is especially true in the `POP_Handler`, `DEP_Handler`, `SetupExecutive`, and `RestoreFromCall` methods within the VM's Executive.  For a specific test graph with a mixture of geometry and mathematical operations, these  enhancements specifically reduce overall `UpdateGraph` run time by 35% (17s to 13s).  The net impact is also dramatically reduces temp memory allocation.  For the case of this specific test graph the temporary memory allocation associated with `UpdateGraph` went from 11.3gb to 3.6gb.  In summary this PR optimizes the execution of functions handled via the JIL Endpoint but has a net improvement to all node types.

Specifically this PR does the following

* Clean up dependency on `CurrentStackFrame` property of `RuntimeMemory`.  This getter creates a new `StackFrame` object to reference a subset of items at a specific location in the VM's Stack.  The `CurentStackFrame` property is utilized 99% of the time from the `IsGlobalScope` method.  This issue is `IsGlobalScope` can be called tens of millions of times during a Graph Execution run which creates a new `StackFrame` object every time the `CurrentStackFrame` property is referenced.  This optimization simply removes the need to allocate a temporary `StackValue` object when the data which is needed can be easily referenced directly from the Stack.  This optimization represents the majority of the extra gigabytes of temporary allocation described above in the sample graph performance delta.

* Refactor `RestoreFromCall` to not allocate empty list until after the required check of `runtimeCore.Options.RunMode == InterpreterMode.Expression`.  This allocation is done repeatedly with no items added to the collection. The call later to check the list via Any() checks can be refactored to a null check.

* Fast path for `GetGraphNodesAtScope` when asking for a Invalid ClassIndex and ProcessIndex (ie -1).  This is another function that can be called millions of times during a UpdateGraph run.  Many calls that are routed through this function are looking for the same item in the `graphNodeMap` dictionary.  This optimization creates a shortcut when the lookup is asking for the specific case of the invalid `ClassIndex` and `ProcessIndex` that avoids accessing the object from the dictionary.  Note, in the case of this function, caching the previous lookup would not speed up the lookup as the method usage typically alternates between values.

* Refactor `UpdateGraph` to not allocate a temporary list of GraphNodes. 

### Declarations

### Reviewers

@sm6srw @aparajit-pratap 

### FYIs

@jasonstratton @mjkkirschner


saintentropy:
@jasonstratton @aparajit-pratap Closing https://github.com/DynamoDS/Dynamo/pull/12153 in favor of this.  This removes the commit refactoring `JILFunctionEndPoint` to allow caching of the `Interpreter` which also had a failing test.  This PR now is exclusively allocation optimizations and should have no impact on exectuion.

src/Engine/ProtoCore/AssociativeGraph.cs
```diff
@@ -1527,8 +1527,20 @@ public DependencyGraph(Core core)
...
         {
+            if (classIndex == Constants.kInvalidPC && procIndex == Constants.kInvalidPC)
+            {
+                if (cachedNodes == null)
+                {
+                    graphNodeMap.TryGetValue(GetGraphNodeKey(classIndex, procIndex), out cachedNodes);
+                }
+
+                return cachedNodes;
+            }
```
aparajit-pratap, line 1542: Dictionary lookup is basically constant time, so how much time does caching really save?

src/Engine/ProtoCore/DSASM/Executive.cs
```diff
@@ -1422,7 +1428,6 @@ private int UpdateGraph(int exprUID, bool isSSAAssign)
                 }
                 gnode.isActive = false;
             }
-            return reachableGraphNodes.Count;
```
aparajit-pratap, line 1425: Why don't we keep this `return `statement, that way we can just say `return GraphUpdateImpl(...);` above.

src/Engine/ProtoCore/DSASM/Executive.cs
```diff
@@ -4485,10 +4489,10 @@ private void RETCN_Handler(Instruction instruction)
         private List<bool> RetrieveExecutionStatesFromStack(int localSize, int paramSize)
         {
             // Retrieve the execution execution states 
-            List<bool> execStateRestore = new List<bool>();
```
aparajit-pratap, line 4488: This is constant time again. Does this really save a whole lot?

src/Engine/ProtoCore/RuntimeMemory.cs
```diff
@@ -294,32 +294,101 @@ public int CurrentConstructBlockId
...
+                        for (int sourceIndex = fp - 1, destIndex = 0;
+                            sourceIndex >= fp - StackFrame.StackFrameSize;
+                            sourceIndex--, destIndex++)
                         {
                             frame[destIndex] = Stack[sourceIndex];
                         }
-                        var stackFrame = new StackFrame(frame);
+
+                        stackFrame = new StackFrame(frame);
                         return stackFrame;
```
aparajit-pratap, line 331: This is a redundant `return` as you're returning the `stackFrame `outside the `if` block anyway.

src/Engine/ProtoCore/RuntimeMemory.cs
```diff
@@ -294,32 +294,101 @@ public int CurrentConstructBlockId
...
+                get
+                {
+                    if (FramePointer - StackFrame.StackFrameSize >= startFramePointer)
                     {
-                        return null;
+                        return false;
                     }
+
+                    return true;
+                }
```
aparajit-pratap, line 348: Why can't we simply say:
```
get { stackFrame == null; }
```

src/Engine/ProtoCore/DSASM/Mirror/ExecutionMirror.cs
```diff
@@ -295,9 +295,9 @@ private int GetSymbolIndex(string name, out int ci, ref int block, out SymbolNod
 
             if (runtimeCore.DebugProps.DebugStackFrameContains(DebugProperties.StackFrameFlagOptions.FepRun))
             {
-                ci = runtimeCore.watchClassScope = rmem.CurrentStackFrame.ClassScope;
-                functionIndex = rmem.CurrentStackFrame.FunctionScope;
-                functionBlock = rmem.CurrentStackFrame.FunctionBlock;
```
aparajit-pratap, line 300: Do we still need these properties on `StackFrame` or can they be deprecated/removed?

src/Engine/ProtoCore/AssociativeGraph.cs
saintentropy: I think it is about scale. This is a nested dictionary lookup as there is another dictionary look up in `GetGraphNodeKey`.  In my sample graph `GetGraphNodesAtScope` is called 3 million times which adds up.

src/Engine/ProtoCore/DSASM/Executive.cs
saintentropy: Yes that makes sence.  I will update that.

src/Engine/ProtoCore/DSASM/Mirror/ExecutionMirror.cs
saintentropy: I do not think so.  The main concern was the `CurrentStackFrame` object.  It was always generating a new `StackFrame` object which was happening millions of times.  If you already have a `StackFrame` object these properties are fine.  I am only shortcutting the instantiation for the `CurrentStackFrame` wrapper class.

src/Engine/ProtoCore/RuntimeMemory.cs
saintentropy: Yep good catch

src/Engine/ProtoCore/RuntimeMemory.cs
saintentropy: I don't want to initialize the `stackFrame` object unless you actually need a `StackFrame` object.  That is the big time penalty here.  All of these essentially copy the implementation from StackFrame here to ovoid the getter for the CurrentStackFrame.  On my sample graph it is called 21million times.

src/Engine/ProtoCore/RuntimeMemory.cs
```diff
@@ -294,32 +294,101 @@ public int CurrentConstructBlockId
                 }
             }
 
+
+            private int fp;
+            private StackFrame stackFrame;
```
aparajit-pratap, line 299: rename to `currentStackFrame`

aparajit-pratap (approved):

PR #12897, opened 2022-05-18
DYN-4803 MouseWheel GroupStyle

filipeotero:
### Purpose

The objective of this PR is to fix the bug that the scrollbar of the group styles is not working with the mouse wheel.
For this, a MouseWheelScrollPreview property was added in the scrollview item to force the bar to be scrolled.

MouseScrollBugFix

### Declarations

### Reviewers

@QilongTang

### DIFF
--- a/src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
+++ b/src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
@@ -335,5 +335,19 @@ namespace Dynamo.Wpf.Views
                 viewModel.DisableTrustWarnings = (bool)(sender as ToggleButton).IsChecked;
             }
         }
+
+        private void ScrollViewer_PreviewMouseWheel(object sender, MouseWheelEventArgs e)
+        {
+            ScrollViewer scrollviewer = sender as ScrollViewer;
+            if (e.Delta > 0)
+            {
+                scrollviewer.LineUp();
+            }
+            else
+            {
+                scrollviewer.LineDown();
+            }
+            e.Handled = true;
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
```diff
@@ -335,5 +335,19 @@ private void DisableTrustWarningsChecked(object sender, RoutedEventArgs e)
                 viewModel.DisableTrustWarnings = (bool)(sender as ToggleButton).IsChecked;
             }
         }
+
+        private void ScrollViewer_PreviewMouseWheel(object sender, MouseWheelEventArgs e)
```
QilongTang, line 339: Interesting, isn't this functionality defined by WPF by default for us? Just curious

filipeotero: Yeah, it's weird because the action of the mouse wheel is being triggered but it's not scrolling without forcing it. I found this solution through this StackOverflow question: https://stackoverflow.com/questions/25366784/wpf-scrollviewer-mousewheel-not-working-with-stackpanel.

QilongTang: I feel that is the case so I was hoping we can address on the xaml level but this is also fine. @filipeotero Do you mind take a look at the xaml to see if there is anything fishy? If none, we can just use this way

filipeotero: Sure, I will take a look

filipeotero: Hi @QilongTang, didn't find anything in the XAML that helps with the fix. Should we proceed in this way?

PR #12930, opened 2022-05-25
DYN-4840 Avoid duplicated terms of use if the menu item is clicked

filipeotero:
### Purpose

This PR fixes duplicated terms of use if the _packages_ menu item is clicked by verifying if it has already been opened.
Related to the task DYN-4840.

TermsOfUseDuplicatedBugFix

### Declarations

### Reviewers

@QilongTang

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
@@ -47,6 +47,8 @@ namespace Dynamo.ViewModels
         private readonly PackageManagerClient packageManagerClient;
         private readonly AuthenticationManager authenticationManager;
 
+        private static TermsOfUseView termsOfUseView;
+
         public TermsOfUseHelper(TermsOfUseHelperParams touParams)
         {
             if (touParams == null)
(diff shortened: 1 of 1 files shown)


src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerClientViewModel.cs
```diff
@@ -47,6 +47,8 @@ public class TermsOfUseHelper
         private readonly PackageManagerClient packageManagerClient;
         private readonly AuthenticationManager authenticationManager;
 
+        private static TermsOfUseView termsOfUseView;
```
QilongTang, line 50: I am usually not a fan of keeping an instance/reference of view in view model since most of Dynamo code should follow MVVM model. Can you think of a different approach? For example, making the TermsOfUseView view a singleton?

filipeotero: We can create a flag that checks if the view is created.

QilongTang: Please give it a try

QilongTang (approved):

PR #12934, opened 2022-05-26
GraphicDataProvider Fix: Regression for pointer referencing non-clr objects

saintentropy:
### Purpose

This PR fixes a regression associated with task https://jira.autodesk.com/browse/DYN-4229 and the https://github.com/DynamoDS/Dynamo/pull/11920.  This adds back the check for edge cases where some Pointers are not associated with CLR Objects.  That is determined by checking the associated Class within the VM ClassTable is an external library.  

### Declarations

### Reviewers

@sm6srw @aparajit-pratap 

### FYIs

@QilongTang @Amoursol

### DIFF
--- a/src/Engine/ProtoCore/Reflection/GraphicDataProvider.cs
+++ b/src/Engine/ProtoCore/Reflection/GraphicDataProvider.cs
@@ -147,21 +147,24 @@ namespace ProtoCore.Mirror
         
         internal object GetCLRObject(StackValue svData, RuntimeCore runtimeCore)
         {
+            if (null == runtimeCore.DSExecutable.classTable)
+                return null;
+
+            //The GetCLRObject function is typically utilized retrieve a clrObject from a StackValue of type pointer.
+            //There is an edge cases for pointers where the pointer references an  a non CLR object.  This code
+            //checks for this edge case by verifying that the requested StackValue pointer is associated with an
+            //imported library.  An example is the "Function" pointer which does not have an associated CLRObject.
+            //In that case, the return value should be null.
+            var classNode = runtimeCore.DSExecutable.classTable.GetClassNodeAtIndex(svData.metaData.type);
+            if (classNode != null  && !classNode.IsImportedClass)
+            {
+                return null;
+            }
+
             if (marshaler != null)
             {
                 return marshaler.UnMarshal(svData, null, interpreter, typeof(object));
             }
-            
-            if (null == runtimeCore.DSExecutable.classTable)
-                return null;
-
-            IList<ClassNode> classNodes = runtimeCore.DSExecutable.classTable.ClassNodes;
-            if (null == classNodes || (classNodes.Count <= 0))
-                return null;
-
-            ClassNode classnode = runtimeCore.DSExecutable.classTable.ClassNodes[svData.metaData.type];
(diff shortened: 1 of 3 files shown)


sm6srw (approved):

src/Engine/ProtoCore/Reflection/GraphicDataProvider.cs
```diff
@@ -147,21 +147,24 @@ internal List<IGraphicItem> GetGraphicItems(DSASM.StackValue svData, RuntimeCore
...
 
-            IList<ClassNode> classNodes = runtimeCore.DSExecutable.classTable.ClassNodes;
-            if (null == classNodes || (classNodes.Count <= 0))
+            //The GetCLRObject function is typically utilized to retrieve a ClrObject from a StackValue of type pointer.
+            //There is an edge cases for pointers where the pointer references a non CLR object.  This code
+            //checks for this edge case by verifying that the requested StackValue pointer is associated with an
+            //imported library.  An example is the "Function" pointer which does not have an associated CLRObject.
+            //In that case, the return value should be null.
+            var classNode = runtimeCore.DSExecutable.classTable.GetClassNodeAtIndex(svData.metaData.type);
+            if (classNode != null  && !classNode.IsImportedClass)
```
aparajit-pratap, line 159: Is checking if it's not an imported class a sufficient condition to determine that it's a function object? My concern is that there could be (if not today) in the future, other DS types that are not treated as imported classes for which we may not want to return null.

aparajit-pratap: Can we use this condition instead?

aparajit-pratap: Based on our discussion, I think it's okay to leave this as-is for now. LGTM!

aparajit-pratap (approved):

aparajit-pratap:
Can we add a test case so that we don't miss a regression like this in the future?

src/Engine/ProtoCore/Reflection/GraphicDataProvider.cs
saintentropy: This may have been a pre-existing issue then.  This fix just returns the function to its previous behavior.

saintentropy: This may have been a pre-existing issue then.  This fix just returns the function to its previous behavior.

saintentropy:
Yep let me take a look

test/DynamoCoreTests/Nodes/ListTests.cs
```diff
@@ -665,6 +665,19 @@ public void SortBy_SimpleTest()
...
+        {
+            string openPath = Path.Combine(TestDirectory, @"core\list\SortBy_SimpleTest.dyn");
+            RunModel(openPath);
+
+            var preview = GetPreviewValueInString("42ac0cec-442f-4e4a-b629-1260f6db3d86");
+            Assert.AreEqual("Function()", preview);
+            
+            var value = GetPreviewValue("42ac0cec-442f-4e4a-b629-1260f6db3d86");
+            Assert.Null(value);
+        }
```
aparajit-pratap, line 678: This test seems to pass even before this fix. Can you create one that fails before this fix?

PR #12935, opened 2022-05-26
Dyn 4977 crash during installation

jesusalvino:
### Purpose

Excluding wrong package versions in order to prevent  Dynamo from Crash
https://jira.autodesk.com/browse/DYN-4977

### Declarations

###Reviewers
@QilongTang

###FYIs
@RobertGlobant20 @filipeotero

### DIFF
--- a/src/DynamoPackages/PackageManagerClient.cs
+++ b/src/DynamoPackages/PackageManagerClient.cs
@@ -78,7 +78,7 @@ namespace Dynamo.PackageManager
             return FailFunc.TryExecute(() => {
                 var nv = HeaderCollectionDownload.ByEngine("dynamo");
                 var pkgResponse = this.client.ExecuteAndDeserializeWithContent<List<PackageHeader>>(nv);                
-                CleanPackageWithWrongVersions(pkgResponse.content);
+                CleanPackagesWithWrongVersions(pkgResponse.content);
                 return pkgResponse.content;
             }, new List<PackageHeader>());
         }
@@ -87,7 +87,7 @@ namespace Dynamo.PackageManager
         /// For every package Checks its versions and exclude the ones with errors
         /// </summary>
         /// <param name="packages"></param>
-        void CleanPackageWithWrongVersions(List<PackageHeader> packages)
+        void CleanPackagesWithWrongVersions(List<PackageHeader> packages)
         {
             foreach (var package in packages)
             {
--- a/src/DynamoPackages/PackageManagerClient.cs
+++ b/src/DynamoPackages/PackageManagerClient.cs
@@ -91,12 +91,12 @@ namespace Dynamo.PackageManager
         {
             foreach (var package in packages)
             {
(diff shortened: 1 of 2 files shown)


src/DynamoPackages/PackageManagerClient.cs
```diff
@@ -77,11 +77,31 @@ internal IEnumerable<PackageHeader> ListAll()
...
 
+        /// <summary>
+        /// For every package Checks its versions and exclude the ones with errors
+        /// </summary>
+        /// <param name="packages"></param>
+        void CleanPackagesWithWrongVersions(List<PackageHeader> packages)
+        {
+            foreach (var package in packages)
+            {
+                bool packageHasWrongVersion = package.versions.Count(v => v.full_dependency_versions == null) > 0;
```
QilongTang, line 94: @zeusongit Can you remind me what does `full_dependency_versions` mean here?

zeusongit: it contains all version of packages listed in `full_dependency_ids` these are packages on which the current package is dependent upon.

zeusongit: I talked to greg marr about this. It was probably a DB error where a partial duplication occurred somehow.

QilongTang: @zeusongit So this change is expected? Do we need to apply this elsewhere?

jesusalvino: There are 70 packages under the same scenario, I have added more details here https://jira.autodesk.com/browse/DYN-4977  , I'm checking in-depth in order to find the root cause

reddyashish: Is the packages list here, a list of dependent packages?

jesusalvino: > Is the packages list here, a list of dependent packages?

not, they are packages that also have wrong data in their versions.

reddyashish (approved):

QilongTang:
Regression
```
Dynamo.PackageManager.Tests.PackageManagerClientTests.ListAll_ReturnsNonEmptyListWhenPackagesAreAvailable
```
@zeusongit @reddyashish Regardless of the DPM server side improvement, should we still merge this one on client side?

zeusongit:
I think we should not base our results on `full_dependencies` as described in my previous comment. 

However regarding the failing test it will always fail because the response provided in the test is an empty record, and the new condition will filter it out.
We can add the `url` property in the response so the `ListAll` allows it and then we can have the test pass as usual.

QilongTang:
@zeusongit Make sense, so `url` is more guaranteed to exist than `full_dependencies`?

zeusongit (approved):
I think you will still need to update the test, other than that LGTM

jesusalvino:
imagen

The related Test has been updated

PR #12941, opened 2022-05-27
Fix Publish package remove item bug

zeusongit:
### Purpose

Fix a bug in Dynamo Publish Package window, where user was unable to remove an item after publishing it it the next version. The problem was that the PackageContent list was being refreshed, bringing all the concatenated files back, including the deleted ones, therefore now we are updating each separate file list as well, so that a PackageContent refresh won't affect the delete action.

Revit_Pl22aPmiz0

### Declarations

### Release Notes

### Reviewers

### FYIs

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/PackageManager/PublishPackageViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/PackageManager/PublishPackageViewModel.cs
@@ -1335,15 +1335,28 @@ namespace Dynamo.PackageManager
         {
             if (!(parameter is PackageItemRootViewModel packageItemRootViewModel)) return;
 
-            if (packageItemRootViewModel.FileInfo == null || packageItemRootViewModel.FileInfo == null)
-            {
-                PackageContents.Remove(PackageContents
-                    .First(x => x.Name == packageItemRootViewModel.Name));
-                return;
-            }
+            string fileName = packageItemRootViewModel.FileInfo == null ? packageItemRootViewModel.Name : packageItemRootViewModel.FileInfo.FullName;
+            string fileType = packageItemRootViewModel.DependencyType.ToString();
 
-            PackageContents.Remove(PackageContents
-                .First(x => x.FileInfo?.FullName == packageItemRootViewModel.FileInfo.FullName));
+            if (fileName.ToLower().EndsWith(".dll") || fileType.Equals("Assembly"))
+            {
+                Assemblies.Remove(Assemblies
+                    .First(x => x.Name == Path.GetFileNameWithoutExtension(fileName)));
+            }
+            else if (fileType.Equals("CustomNode"))
+            {
+                CustomNodeDefinitions.Remove(CustomNodeDefinitions
+                    .First(x => x.DisplayName == fileName));
+            }
+            else
+            {
+                AdditionalFiles.Remove(AdditionalFiles
+                    .First(x => x == fileName));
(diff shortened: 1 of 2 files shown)


QilongTang:
@zeusongit So this is not a new regression?

zeusongit:
> @zeusongit So this is not a new regression?

No, apparently this used to happen before as well, I guess somehow was never discovered.

QilongTang (commented):
Looks good but can we add a unit test since we never knew this regressed?

src/DynamoCoreWpf/ViewModels/PackageManager/PublishPackageViewModel.cs
```diff
@@ -1335,15 +1335,28 @@ private void RemoveItem(object parameter)
         {
             if (!(parameter is PackageItemRootViewModel packageItemRootViewModel)) return;
 
-            if (packageItemRootViewModel.FileInfo == null || packageItemRootViewModel.FileInfo == null)
+            string fileName = packageItemRootViewModel.FileInfo == null ? packageItemRootViewModel.Name : packageItemRootViewModel.FileInfo.FullName;
+            string fileType = packageItemRootViewModel.DependencyType.ToString();
+
+            if (fileName.ToLower().EndsWith(".dll") || fileType.Equals(DependencyType.Assembly))
```
reddyashish, line 1341: If the filetype is DependencyType.Assembly, I suppose it should be a dll. Do we need both the checks?

reddyashish (approved):
Looks good to me with one comment.

src/DynamoCoreWpf/ViewModels/PackageManager/PublishPackageViewModel.cs
zeusongit: Yes, this was the case before, but I noticed that we do not get filename with `.dll` extension, therefore added another check that will satisfy this condition, putting an `&&` will not pass if the filename is without extension.

mjkkirschner: what about packages that ship .exes?

zeusongit: According to the current flow, they will end up in additional files

zeusongit: Do you want me to add `.exe` extension to assemblies?

zeusongit: Do exes have Dependency Type as Assembly?

mjkkirschner: I think not - according to:
https://github.com/DynamoDS/Dynamo/blob/74709f73df1b87a1d4d3997da947dac3dea3697f/src/DynamoPackages/Package.cs#L335

we only scan files ending in `*.dll` and assign those types as assembly. I guess like you said above .exes would just be another file in extra.

zeusongit: I published a test package with exes, it ends up in additional files, so we are good.
thanks you for the review @reddyashish @mjkkirschner

QilongTang (approved):

PR #12944, opened 2022-05-30
Automatic Execution Bug Fixes

dnenov:
### Purpose

Changes and additions to the GraphNodeManager Extension. After the initial review of the extension, a number of items have been found that need an update. 

- [ ] Focus around Node not working in Automatic Mode 
- [ ] Change on Node attributes (Output/Input, Frozen, Rename, etc.) should trigger an update 
- [ ] Add 'x' to delete search bar text
- [ ] DataGrid column separators - visible and correctly aligned
- [ ] Handle a case when a Node is both in Warning and Error states at the same time
- [ ] Scroll should be possible for graphs with many Nodes

### Declarations

### Release Notes

- focus around node when selecting from the UI now correctly works when in Automatic mode
- items of the datagrid now correctly updates on Node changes

### Reviewers

@reddyashish 

### FYIs
@mjkkirschner
@Amoursol

### DIFF
--- a/src/DynamoCoreWpf/Extensions/ViewModelCommandExecutive.cs
+++ b/src/DynamoCoreWpf/Extensions/ViewModelCommandExecutive.cs
@@ -41,6 +41,16 @@ namespace Dynamo.Wpf.Extensions
         }
 
         /// <summary>
+        /// Search for an element by its ID and focus the view on it
+        /// </summary>
+        /// <param name="objectID"></param>
+        public void FocusNodeCommand(string objectID)
+        {
+            dynamoViewModel.CurrentSpaceViewModel.FocusNodeCommand.Execute(objectID);
+        }
+
+
+        /// <summary>
         /// Force re-execute all nodes in the current workspace
         /// </summary>
         /// <param name="showErrors">Should errors be shown?</param>
--- a/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
@@ -1946,9 +1946,15 @@ namespace Dynamo.ViewModels
             }
         }
 
-        internal void ShowElement(NodeModel e)
+        /// <summary>
+        /// Attempts to find a NodeModel and focuses the view around it
+        /// Default boolean introduced to allow for this method to be used in Automatic mode
+        /// </summary>
+        /// <param name="e"></param>
+        /// <param name="automatic"></param>
+        internal void ShowElement(NodeModel e, bool automatic = true)
         {
-            if (HomeSpace.RunSettings.RunType == RunType.Automatic)
+            if (HomeSpace.RunSettings.RunType == RunType.Automatic && automatic)
                 return;
 
             if (!model.CurrentWorkspace.Nodes.Contains(e))
--- a/src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
@@ -1416,6 +1416,37 @@ namespace Dynamo.ViewModels
             return true;
         }
 
+        /// <summary>
+        /// Selects an Element by ID and focuses the view around it
+        /// </summary>
+        /// <param name="id"></param>
+        private void FocusNode(object id)
+        {
+            try
+            {
+                var node = DynamoViewModel.Model.CurrentWorkspace.Nodes.First(x => x.GUID.ToString() == id.ToString());
+                if(node != null)
+                {
+                    //select the element
+                    DynamoSelection.Instance.ClearSelection();
+                    DynamoSelection.Instance.Selection.Add(node);
+
+                    DynamoViewModel.ShowElement(node, false);
+
+                    return;
+                }
+            }
+            catch
+            {
+                DynamoViewModel.Model.Logger.Log(Wpf.Properties.Resources.MessageFailedToFindNodeById);
+            }
+        }
+
+        private static bool CanFocusNode(object id)
+        {
+            return !string.IsNullOrEmpty(id.ToString());
+        }
+
         private void FindById(object id)
         {
             try
--- a/src/GraphNodeManagerViewExtension/ViewModels/NodeViewModel.cs
+++ b/src/GraphNodeManagerViewExtension/ViewModels/NodeViewModel.cs
@@ -367,8 +367,63 @@ namespace Dynamo.GraphNodeManager.ViewModels
         public NodeViewModel(NodeModel node)
         {
             NodeModel = node;
+
+            NodeModel.Modified += NodeModel_Modified;
+            NodeModel.PropertyChanged += NodeModel_PropertyChanged;
         }
-        
+
+        private void NodeModel_PropertyChanged(object sender, System.ComponentModel.PropertyChangedEventArgs e)
+        {
+            EvaluateNode(sender as NodeModel, e.PropertyName);
+        }
+
+        private void NodeModel_Modified(NodeModel nodeModel)
+        {
+            EvaluateNode(nodeModel);
+        }
+
+        /// <summary>
+        /// In case a Node has been modified in any way, update the affected property 
+        /// </summary>
+        /// <param name="nodeModel"></param>
+        /// <param name="propertyName"></param>
+        private void EvaluateNode(NodeModel nodeModel, string propertyName = "Modified")
+        {
+            switch (propertyName)
+            {
+                case "Name":
+                    RaisePropertyChanged(nameof(Name));
+                    break;
+                case "IsVisible":
+                    RaisePropertyChanged(nameof(StatusIsHidden));
+                    break;
+                case "IsSetAsInput":
+                    RaisePropertyChanged(nameof(StateIsInput));
+                    break;
+                case "IsSetAsOutput":
+                    RaisePropertyChanged(nameof(StateIsOutput));
+                    break;
+                case "IsInErrorState":
+                    RaisePropertyChanged(nameof(IssuesHasError));
+                    break;
+                case "IsFrozen":
+                    RaisePropertyChanged(nameof(StatusIsFrozen));
+                    break;
+                case "State":
+                    RaisePropertyChanged(nameof(State));
+                    break;
+            }
+        }
+
+        /// <summary>
+        /// Detach from all event handlers
+        /// </summary>
+        public void Dispose()
+        {
+            NodeModel.Modified -= NodeModel_Modified;
+            NodeModel.PropertyChanged -= NodeModel_PropertyChanged;
+        }
+
         #endregion
 
     }
--- a/src/DynamoCoreWpf/Commands/WorkspaceCommands.cs
+++ b/src/DynamoCoreWpf/Commands/WorkspaceCommands.cs
@@ -18,6 +18,7 @@ namespace Dynamo.ViewModels
         private DelegateCommand setZoomCommand;
         private DelegateCommand resetFitViewToggleCommand;
         private DelegateCommand findByIdCommand;
+        private DelegateCommand focusNodeCommand;
         private DelegateCommand alignSelectedCommand;
         private DelegateCommand setArgumentLacingCommand;
         private DelegateCommand findNodesFromSelectionCommand;
@@ -146,6 +147,18 @@ namespace Dynamo.ViewModels
                     findByIdCommand = new DelegateCommand(FindById, CanFindById);
 
                 return findByIdCommand;
+            }
+        }
+
+        [JsonIgnore]
+        public DelegateCommand FocusNodeCommand
+        {
+            get
+            {
+                if (focusNodeCommand == null)
+                    focusNodeCommand = new DelegateCommand(FocusNode, CanFocusNode);
+
+                return focusNodeCommand;
             }
         }
 
--- a/src/GraphNodeManagerViewExtension/GraphNodeManagerViewModel.cs
+++ b/src/GraphNodeManagerViewExtension/GraphNodeManagerViewModel.cs
@@ -15,9 +15,9 @@ using Dynamo.PackageManager;
 using Dynamo.Utilities;
 using Dynamo.ViewModels;
 using Microsoft.Practices.Prism.Commands;
-using Newtonsoft.Json;
 using DelegateCommand = Dynamo.UI.Commands.DelegateCommand;
 using NodeViewModel = Dynamo.GraphNodeManager.ViewModels.NodeViewModel;
+using Newtonsoft.Json;
 
 namespace Dynamo.GraphNodeManager
 {
@@ -281,7 +281,7 @@ namespace Dynamo.GraphNodeManager
             commandExecutive.ExecuteCommand(command, uniqueId, "GraphNodeManager");
 
             // Focus on selected
-            viewModelCommandExecutive.FindByIdCommand(nodeViewModel.NodeModel.GUID.ToString());
+            viewModelCommandExecutive.FocusNodeCommand(nodeViewModel.NodeModel.GUID.ToString());
         }
 
         /// <summary>
@@ -531,12 +531,25 @@ namespace Dynamo.GraphNodeManager
         public void Dispose()
         {
             UnsubscribeWorkspaceEvents(CurrentWorkspace);
+
+            DisposeNodes();
+
             viewLoadedParams.CurrentWorkspaceChanged -= OnCurrentWorkspaceChanged;
             viewLoadedParams.CurrentWorkspaceCleared -= OnCurrentWorkspaceCleared;
             NodesCollection.Filter -= NodesCollectionViewSource_Filter;
             GraphNodeManagerView = null;
         }
 
+        /// <summary>
+        /// Dispose of each NodeViewModel individually
+        /// </summary>
+        private void DisposeNodes()
+        {
+            foreach(var nvm in Nodes)
+            {
+                nvm.Dispose();
+            }
+        }
         #endregion
     }
 }


reddyashish:
All tests passing now: https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/994/

reddyashish:
The build on master-5 failed due to a VM issue. Will trigger the job again.

QilongTang:
@reddyashish Let me know if these improvements need to be included for demo

reddyashish:
@QilongTang, Looks good but needed some clarification from Deyan before merging. I was thinking, we will wait for him to come back and then merge.

src/GraphNodeManagerViewExtension/ViewModels/NodeViewModel.cs
```diff
@@ -366,8 +366,19 @@ private bool IsNodeEmptyList(MirrorData mirrorData)
...
             if (mirrorData == null || !mirrorData.IsCollection) return false;
 
-            var list = mirrorData.GetElements();
-            return !list.Any();
+            try
+            {
+                var list = mirrorData.GetElements();
+                return !list.Any();
+            }
+            catch(NullReferenceException)
```
zeusongit, line 374: The above null check for `mirrorData ` did not resolve this? I believe if `mirrorData ` was null it would have returned false from line 367.

dnenov: Actually, no. It turns out that ``mirrorData`` still throws a NullReferenceException even when it's not **null**.  It happens here 
array_null
-> the ``array`` can still be null.

dnenov: Do you think I should, as part of this PR, make changes to the ``MirrorData`` class inside the ``GetElements()`` method to make sure this case is covered?

mjkkirschner: I think you can just also add a check for MirrorData.IsNull() this is different than checking if the mirrorData wrapper is null, it checks the stack value.

zeusongit: I see that we already check for `IsCollection` inside the `GetElements` method, we should also include the appropriate null check inside it as well, and then just use the output from it. This will also cover any future issues we may get into while using this class.

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
```diff
@@ -1416,6 +1416,37 @@ private static bool CanResetFitViewToggle(object o)
...
+            }
+            catch
+            {
+                DynamoViewModel.Model.Logger.Log(Wpf.Properties.Resources.MessageFailedToFindNodeById);
+            }
+        }
+
+        private static bool CanFocusNode(object id)
+        {
+            return !string.IsNullOrEmpty(id.ToString());
```
aparajit-pratap, line 1447: Shouldn't this return true or false depending on if the node is found for a valid id?

src/GraphNodeManagerViewExtension/ViewModels/NodeViewModel.cs
dnenov: @mjkkirschner @zeusongit I just pushed the commit with the alternative solution we have discussed for your review. Let me know if you think that's not a good way of handling things and I will amend accordingly.

src/GraphNodeManagerViewExtension/ViewModels/GridNodeViewModel.cs
```diff
@@ -388,21 +380,10 @@ private bool IsNodeEmptyList(MirrorData mirrorData)
...
-            }
-            catch (NullReferenceException)
-            {
-                return false;
-            }
-            catch (Exception)
-            {
-                return false;
-            }
+            if (string.Equals(mirrorData.Class.Name, "Function")) return true;
```
mjkkirschner, line 385: I would keep the try catch, or also check name for null/empty

src/DynamoCoreWpf/Extensions/ViewModelCommandExecutive.cs
```diff
@@ -40,6 +40,16 @@ public void FindByIdCommand(string objectID)
             dynamoViewModel.CurrentSpaceViewModel.FindByIdCommand.Execute(objectID);
         }
 
+        /// <summary>
+        /// Search for an element by its ID and focus the view on it
```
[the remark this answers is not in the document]
dnenov: Yup, done! Thanks for the note, @aparajit-pratap.

src/GraphNodeManagerViewExtension/ViewModels/GridNodeViewModel.cs
```diff
@@ -364,21 +364,13 @@ public Guid NodeGuid
         /// <returns></returns>
         private bool IsNodeEmptyList(MirrorData mirrorData)
         {
-            if (mirrorData == null || !mirrorData.IsCollection) return false;
+            if (mirrorData.IsNull || mirrorData == null || !mirrorData.IsCollection) return false;
 
-            try
```
mjkkirschner: I don't think there is anything wrong with keeping these try catches in areas of code where you know it is hard to avoid null refs.

Do you expect many exceptions? If the exceptions are not constant, adding a try is not a large performance cost.
https://stackoverflow.com/questions/1308432/do-try-catch-blocks-hurt-performance-when-exceptions-are-not-thrown

src/GraphNodeManagerViewExtension/ViewModels/GridNodeViewModel.cs
dnenov: Thanks for pointing that out! I added a ``IsNullOrEmpty`` check for the mirrorData.Class.Name before using it as per your suggestion, @mjkkirschner.

src/Engine/ProtoCore/Reflection/MirrorData.cs
```diff
@@ -169,6 +169,9 @@ public IEnumerable<MirrorData> GetElements()
                     return null;
 
                 var array = runtimeCore.Heap.ToHeapObject<DSArray>(svData);
+                if (array == null)
+                    return null;
```
aparajit-pratap, line 173: I wouldn't recommend changing this method as if this case occurred before, it would have been addressed by now.

src/GraphNodeManagerViewExtension/ViewModels/GridNodeViewModel.cs
dnenov: The information regarding the performance cost of try catches is awesome! I will use that in the future, but I think, for this particular case, it was the unhandled exception that was causing the issue and made me use try/catch blocks before your comments. Fixing the root cause solves the problem altogether, I think!

src/Engine/ProtoCore/Reflection/MirrorData.cs
dnenov: Ah ... @aparajit-pratap  @mjkkirschner  I will then revert back to my original solution using try/catches within my own code. I can guarantee you that this is an issue, though, and I can demonstrate it for you, if you need me to.

src/GraphNodeManagerViewExtension/ViewModels/GridNodeViewModel.cs
```diff
@@ -56,7 +61,11 @@ public string Name
                 name = NodeModel?.Name;
                 return name;
             }
-            internal set => name = value;
+            internal set
+            {
+                name = value;
```
aparajit-pratap, line 66: I think a good practice to save unnecessary property changed notifications is to first check if `name != value`, then go ahead with it. Only in a few cases, this does not apply but that needs to be seen on a case-by-case basis.

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
```diff
@@ -1416,6 +1416,37 @@ private static bool CanResetFitViewToggle(object o)
...
+                    DynamoSelection.Instance.Selection.Add(node);
+
+                    DynamoViewModel.ShowElement(node, false);
+
+                    return;
+                }
+            }
+            catch
+            {
+                DynamoViewModel.Model.Logger.Log(Wpf.Properties.Resources.MessageFailedToFindNodeById);
```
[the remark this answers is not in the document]
dnenov: This code is an almost identical duplication of the ``FindById`` and ``CanFindById`` methods that already existed. The modification was done to circumvent a limitation with Automatic mode and has been discussed with @QilongTang. 

image

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
dnenov: Same as the above: 

This code is an almost identical duplication of the ``FindById`` and ``CanFindById`` methods that already existed. The modification was done to circumvent a limitation with Automatic mode and has been discussed with @QilongTang. 

image

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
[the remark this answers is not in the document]
aparajit-pratap: Well, in that case, I would question those methods too :)
Also, why can't we share the implementation, why two identical functions that do the same thing?

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
aparajit-pratap: I'm not sure what you mean by `The modification was done to circumvent a limitation with Automatic mode`. @QilongTang ?

src/Engine/ProtoCore/Reflection/MirrorData.cs
aparajit-pratap: In that case, I would be interested in seeing that demo. If what you're saying is true, I guess there's no harm in adding the `null` check. It's just surprising to me that we didn't catch this before. I'll try to sync up later.

src/GraphNodeManagerViewExtension/ViewModels/GridNodeViewModel.cs
```diff
@@ -318,8 +404,19 @@ private bool IsNodeOutputFunction(MirrorData mirrorData)
...
+                return false;
+            }
+            catch (NullReferenceException)
+            {
+                return false;
+            }
+            catch (Exception)
+            {
+                return false;
+            }
```
aparajit-pratap, line 419: I don't think we need this. We can use the `IsFunction` property on the `MirrorData`class itself that tells us if it's a `Function `or not.

src/GraphNodeManagerViewExtension/ViewModels/GridNodeViewModel.cs
```diff
@@ -305,8 +380,19 @@ private bool IsNodeEmptyList(MirrorData mirrorData)
...
-            return !list.Any();
+            try
+            {
+                var list = mirrorData.GetElements();
+                return !list.Any();
+            }
+            catch(NullReferenceException)
+            {
+                return false;
+            }
```
aparajit-pratap, line 391: I don't think we need this `catch`.

src/GraphNodeManagerViewExtension/ViewModels/GridNodeViewModel.cs
```diff
@@ -305,8 +380,19 @@ private bool IsNodeEmptyList(MirrorData mirrorData)
...
+            catch (Exception)
+            {
+                return false;
+            }
```
aparajit-pratap, line 395: I think this entire function makes sense to expose on `MirrorData` itself.

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
QilongTang: @aparajit-pratap These are used in view commands and the only available one we have now does not work in automatic run mode, it was set to exclude automatic mode specifically. So we decide to make a new command almost identical but enable node focus in automatic mode. The function does not have return type because it is how the function usually work for commands

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -1946,9 +1946,15 @@ internal void FocusCustomNodeWorkspace(Guid symbol)
...
+        /// <summary>
+        /// Attempts to find a NodeModel and focuses the view around it
+        /// Default boolean introduced to allow for this method to be used in Automatic mode
+        /// </summary>
+        /// <param name="e"></param>
+        /// <param name="automatic"></param>
+        internal void ShowElement(NodeModel e, bool automatic = true)
         {
-            if (HomeSpace.RunSettings.RunType == RunType.Automatic)
+            if (HomeSpace.RunSettings.RunType == RunType.Automatic && automatic)
```
aparajit-pratap, line 1957: I'm afraid, this looks hacky. Why can't we say:
```
if(automatic)
{
HomeSpace.RunSettings.RunType == RunType.Automatic;
return;
}
```

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
```diff
@@ -1416,6 +1416,37 @@ private static bool CanResetFitViewToggle(object o)
...
+            try
+            {
+                var node = DynamoViewModel.Model.CurrentWorkspace.Nodes.First(x => x.GUID.ToString() == id.ToString());
+                if(node != null)
+                {
+                    //select the element
+                    DynamoSelection.Instance.ClearSelection();
+                    DynamoSelection.Instance.Selection.Add(node);
+
+                    DynamoViewModel.ShowElement(node, false);
```
aparajit-pratap, line 1434: Why call `ShowElement` at all, if all you're going to do here is return from it?

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
aparajit-pratap: Ok, but why not log the same message (`MessageFailedToFindNodeById`) even if the node is not found (is null) and there is no exception thrown?

src/GraphNodeManagerViewExtension/ViewModels/GridNodeViewModel.cs
dnenov: Sorry, @aparajit-pratap , I didn't quite get what you meant there. I have implemented some of the modifications based on your comments from above, could you tell me what you think should happen next?

src/GraphNodeManagerViewExtension/ViewModels/GridNodeViewModel.cs
dnenov: Brilliant! Thanks for that.

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
dnenov: Hi! Do you maybe want to have a quick Slack catch-up? These last few things are legacy code for me. I know they are used in TuneUp extension, which is not something I have worked on. Maybe it might be better to judge the extent of this PR - if you think it's a good time to resolve some of these issues, I agree, and let's go over the code and agree on what needs to happen and I will implement it.

src/Engine/ProtoCore/Reflection/MirrorData.cs
dnenov: Here is a bit of an awkward gif to showcase the issue. I have temporarily added a try/catch inside the ``GetElements()`` of MirrorData. It was done using a graph called DynamoOnboardingGuide_HouseCreationDS.dyn which Sol Amour provided me with and I can send out.

The error is thrown when trying to operate on ``array.Values`` with ``array`` being null. 

mirrordata_nullexception

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
dnenov: We just want to focus and zoom around a Node that is being selected by the User through the UI of the extension. ShowElement() ultimately triggers CenterViewOnElement() method from the WorkspaceView.xaml.cs

image

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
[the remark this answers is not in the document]
dnenov: Just to answer the previous question, you were right, the null check was not correct. It's now removed.

src/GraphNodeManagerViewExtension/Controls/SearchBoxControl.xaml
```diff
@@ -107,6 +112,28 @@
...
+                                    Name="SearchClearButton"
+                                    Style="{StaticResource ButtonStyle1}"
+                                    Visibility="{Binding ElementName=SearchTextBox, 
+                                                         Path=Text, 
+                                                         Converter={StaticResource NonEmptyStringToCollapsedConverter}}"
+                                    Width="16px"
+                                    Height="16px"
+                                    HorizontalAlignment="Center"
+                                    VerticalAlignment="Center"
+                                    ToolTip="Clear search box .."
```
reddyashish, line 130: Can we move this to resource string so that we wont have localization issues.

src/GraphNodeManagerViewExtension/ViewModels/GridNodeViewModel.cs
```diff
@@ -364,11 +453,81 @@ private string GetNodeMessage(string message)
...
+        }
+
+        /// <summary>
+        /// In case an Input Node is frozen, only the input node triggers RaisePropertyChange
+        /// All the downstream Nodes won't RaisePropertyChange for IsFrozen (although they will report correctly that they are)
+        /// We need to manually update the view in this case
+        /// </summary>
+        /// <param name="nodeModel"></param>
+        private void UpdateDownstreemNodes(NodeModel nodeModel)
```
reddyashish, line 514: UpdateDownstreamNodes*

src/GraphNodeManagerViewExtension/Controls/SearchBoxControl.xaml
QilongTang: Make sense! Looks like our localization team will need to make additional effort after 6/17..

QilongTang:
> @QilongTang Not sure why the self-service is not being triggered on master-5. It dint run for last couple of this PR commits.

Yeah, feel free to run on https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/

reddyashish:
Self-service job passed: https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/1032/

QilongTang:
@reddyashish Since we are not going to address comment by @aparajit-pratap on this, Feel free to approve and merge when you think proper

reddyashish:
https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/1033/.

PR #12956, opened 2022-06-02
Return correct type indices for primitive types referred to by their CLR type names

aparajit-pratap:
### Purpose

Return correct type indices for primitive types referred to by their CLR type names - this permits a node like `List.RemoveIfNot` to work with not just the DS types passed as input, but also the C# type. For e.g. it now works for both the DS type `int` but also the equivalent C# type, `System.Int64` type.

### Declarations

### Release Notes

### FYIs

### DIFF
--- a/src/Engine/ProtoCore/Lang/Type.cs
+++ b/src/Engine/ProtoCore/Lang/Type.cs
@@ -121,6 +121,16 @@ namespace ProtoCore
         public Dictionary<ProtoCore.DSASM.AddressType, int> addressTypeClassMap { get; set; }
         private static Dictionary<PrimitiveType, string> primitiveTypeNames;
 
+        private static readonly Dictionary<string, string> clrToDSTypeMap = new Dictionary<string, string>
+        {
+            { $"{nameof(System)}.{nameof(String)}", "string" },
+            { "long", "int" },
+            { $"{nameof(System)}.{nameof(Int64)}", "int" },
+            { $"{nameof(System)}.{nameof(Double)}", "double" },
+            { $"{nameof(System)}.{nameof(Boolean)}", "bool" },
+            { $"{nameof(System)}.{nameof(Char)}", "char" }
+        };
+
         public TypeSystem()
         {
             SetTypeSystem();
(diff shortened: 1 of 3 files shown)


src/Engine/ProtoCore/Lang/Type.cs
```diff
@@ -121,6 +121,16 @@ public class TypeSystem
         public Dictionary<ProtoCore.DSASM.AddressType, int> addressTypeClassMap { get; set; }
         private static Dictionary<PrimitiveType, string> primitiveTypeNames;
 
+        private static readonly Dictionary<string, string> clrToDSTypeMap = new Dictionary<string, string>
+        {
+            { $"{nameof(System)}.{nameof(String)}", "string" },
+            { "long", "int" },
```
mjkkirschner, line 127: where does the `long` type name come from given that `int64` is the actual .net type name?

mjkkirschner: considering that this dictionary is only created once - maybe it's not so bad to do:
`typeof(Int64).FullName`

mjkkirschner (approved):

src/Engine/ProtoCore/Lang/Type.cs
```diff
@@ -314,7 +324,13 @@ public string GetType(int UID)
         public int GetType(string ident)
         {
             Validity.Assert(null != classTable);
-            return classTable.IndexOf(ident);
+            var index = classTable.IndexOf(ident);
+            if (index != Constants.kInvalidIndex) return index;
+
+            if(clrToDSTypeMap.TryGetValue(ident, out string dsType))
+                return classTable.IndexOf(dsType);
```
mjkkirschner, line 331: while I think this PR looks good - I am curious if you think there could be performance implications here, this `IndexOf` call doesn't seem to be straight forward, it seems to do partial matching? So in the worst case we'll now do 2x the lookups in some pathologically bad cases, do you think will come up rarely, and if so is this method `classTable.IndexOf` linear in complexity or something worse?

src/Engine/ProtoCore/Lang/Type.cs
aparajit-pratap: I was thinking that someone could manually type `long`, but that's highly unlikely.

aparajit-pratap: I can remove it.

src/Engine/ProtoCore/Lang/Type.cs
aparajit-pratap: I think right now, it's only going to take place for a node like `RemoveIfNot `where there is the possibility of checking for a C# type.

mjkkirschner:
LGTM

PR #12961, opened 2022-06-03
finding the ports by ID insted of an index

filipeotero:
### Purpose

It's included in this PR the fix for a bug at the outputs when a group is collapsed. 

fixOutPutCollapsedBar

### Declarations

### Reviewers

@QilongTang 

### FYIs

@RobertGlobant20

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
@@ -759,14 +759,18 @@ namespace Dynamo.ViewModels
                 .ToList();
 
             var newPortViewModels = new List<PortViewModel>();
-            for (int i = 0; i < groupPortModels.Count(); i++)
+            int index = 0;
+            foreach (var group in groupPortModels)
             {
-                var model = groupPortModels.ElementAt(i);
-                newPortViewModels.Add(originalPortViewModels[i].CreateProxyPortViewModel(model));
-
-                // calculate new position for the proxy outports
-                model.Center = CalculatePortPosition(model, i);
-                model.Owner.ReportPosition();
+                var originalPort = originalPortViewModels.FirstOrDefault(x => x.PortModel.GUID == group.GUID);
+                if (originalPort != null)
+                {
+                    var portViewModel = originalPort.CreateProxyPortViewModel(group);
+                    newPortViewModels.Add(portViewModel);
+                    // calculate new position for the proxy outports
+                    group.Center = CalculatePortPosition(group, index);
+                }
+                index++;
             }
 
             return newPortViewModels;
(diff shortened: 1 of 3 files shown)


reddyashish:
How would it look when the subgroup is collapsed and then the parent group is collapsed?

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -759,14 +759,18 @@ private List<PortViewModel> CreateProxyOutPorts(IEnumerable<PortModel> groupPort
...
+            int index = 0;
+            foreach (var group in groupPortModels)
             {
-                var model = groupPortModels.ElementAt(i);
-                newPortViewModels.Add(originalPortViewModels[i].CreateProxyPortViewModel(model));
-
-                // calculate new position for the proxy outports
-                model.Center = CalculatePortPosition(model, i);
-                model.Owner.ReportPosition();
+                var originalPort = originalPortViewModels.FirstOrDefault(x => x.PortModel.GUID == group.GUID);
```
reddyashish, line 765: I am just curious, how does this make a difference. Weren't were iterating through all the output-ports in the first place? or is this just refactoring?

filipeotero: Hi, thanks for the observation. Previously we were interacting through all the ports without controlling its owner. Some ports were wrongly set to another group or node, causing the bug. By searching its GUID we make sure the right node/group is referenced.

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -771,13 +771,18 @@ private List<PortViewModel> CreateProxyOutPorts(IEnumerable<PortModel> groupPort
                 .ToList();
 
             var newPortViewModels = new List<PortViewModel>();
-            for (int i = 0; i < groupPortModels.Count(); i++)
+            int index = 0;
+            foreach (var group in groupPortModels)
```
zeusongit, line 775: Any specific reason to switch to this foreach loop? we already had index variable in the previous loop.

filipeotero:
> Looks good to me after the conflicts are resolved. One minor suggestion: If we can have little vertical margin between the CBN ports, I feel it would be better.

I also think it looks better. Here it is:

image

QilongTang:
@reddyashish @zeusongit Is this PR approval?
Also @filipeotero Would you record a final version of the gif for me so I can add for the demo? We can use the nested group case, collapse subgroup first then parent group one

zeusongit (approved):

reddyashish (approved):

QilongTang:
Thanks @zeusongit and @reddyashish for the review. Merging.

PR #12974, opened 2022-06-07
collection performance

pinzart90:
### Purpose

Improve performance when opening graphs
https://jira.autodesk.com/browse/DYN-5005

Changes:
  Performance improvement when dealing with iterations

### Declarations

### Release Notes

### Reviewers

### FYIs

### DIFF
--- a/src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
+++ b/src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
@@ -2352,6 +2352,50 @@ namespace Dynamo.Graph.Workspaces
         /// <summary>
         /// Updates a workspace model with extra view information. When loading a workspace from JSON,
         /// the data is split into two parts, model and view. This method sets the view information.
+        /// </summary>
+        /// <param name="workspaceViewInfo">The extra view information from the workspace to update the model with.</param>
+        public void UpdateWithExtraWorkspaceViewInfo(ExtraWorkspaceViewInfo workspaceViewInfo)
+        {
+            if (workspaceViewInfo == null)
+                return;
+
+
+            X = workspaceViewInfo.X;
+            Y = workspaceViewInfo.Y;
+            Zoom = workspaceViewInfo.Zoom;
+
+            OnCurrentOffsetChanged(
+                this,
+                new PointEventArgs(new Point2D(X, Y)));
+
+            Dictionary<Guid, NodeModel> nodeMap = Nodes.ToDictionary(x => x.GUID);
+
+            // This function loads standard nodes
+            LoadNodes(workspaceViewInfo.NodeViews, ref nodeMap);
+
+            // This function loads notes from the Notes array in the JSON format
+            // NOTE: This is here to support early JSON graphs
+            // IMPORTANT: All notes must be loaded before annotations are loaded to
+            //            ensure that any contained notes are contained properly
+            LoadLegacyNotes(workspaceViewInfo.Notes);
+
+            // This function loads notes from the Annotations array in the JSON format
+            // that have an empty nodes collection
+            // IMPORTANT: All notes must be loaded before annotations are loaded to
+            //            ensure that any contained notes are contained properly
+            LoadNotesFromAnnotations(workspaceViewInfo.Annotations);
+
+            // This function loads ConnectorPins to the corresponding connector models.
+            LoadConnectorPins(workspaceViewInfo.ConnectorPins);
+
+            // This function loads annotations from the Annotations array in the JSON format
+            // that have a non-empty nodes collection
+            LoadAnnotations(workspaceViewInfo.Annotations, ref nodeMap);
+        }
+
+        /// <summary>
+        /// Updates a workspace model with extra view information. When loading a workspace from JSON,
+        /// the data is split into two parts, model and view. This method sets the view information.
         /// This overload allows to 'move' the incoming when placing them
         /// </summary>
         /// <param name="workspaceViewInfo"></param>
@@ -2371,8 +2415,10 @@ namespace Dynamo.Graph.Workspaces
                 this,
                 new PointEventArgs(new Point2D(X, Y)));
 
+            Dictionary<Guid, NodeModel> nodeMap = Nodes.ToDictionary(x => x.GUID);
+
             // This function loads standard nodes
-            LoadNodes(workspaceViewInfo.NodeViews, offsetX, offsetY);
+            LoadNodes(workspaceViewInfo.NodeViews, ref nodeMap, offsetX, offsetY);
 
             // This function loads notes from the Notes array in the JSON format
             // NOTE: This is here to support early JSON graphs
@@ -2391,7 +2437,7 @@ namespace Dynamo.Graph.Workspaces
 
             // This function loads annotations from the Annotations array in the JSON format
             // that have a non-empty nodes collection
-            LoadAnnotations(workspaceViewInfo.Annotations);
+            LoadAnnotations(workspaceViewInfo.Annotations, ref nodeMap);
         }
 
         private void LoadNodes(IEnumerable<ExtraNodeViewInfo> nodeViews, ref Dictionary<Guid, NodeModel> nodeMap, double offsetX = 0.0, double offsetY = 0.0)


src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
```diff
@@ -279,20 +284,20 @@ public bool IsCursorForced
...
 
         [JsonIgnore]
-        public ObservableCollection<ConnectorPinViewModel> Pins { get; } = new ObservableCollection<ConnectorPinViewModel>();
+        public SmartObservableCollection<ConnectorPinViewModel> Pins { get; } = new SmartObservableCollection<ConnectorPinViewModel>();
 
         [JsonIgnore]
-        public ObservableCollection<InfoBubbleViewModel> Errors { get; } = new ObservableCollection<InfoBubbleViewModel>();
-        public ObservableCollection<AnnotationViewModel> Annotations { get; } = new ObservableCollection<AnnotationViewModel>();
+        public SmartObservableCollection<InfoBubbleViewModel> Errors { get; } = new SmartObservableCollection<InfoBubbleViewModel>();
+        public SmartObservableCollection<AnnotationViewModel> Annotations { get; } = new SmartObservableCollection<AnnotationViewModel>();
```
pinzart90, line 300: Converting to a more derived type is not considered a breaking change

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
```diff
@@ -496,11 +496,24 @@ public WorkspaceViewModel(WorkspaceModel model, DynamoViewModel dynamoViewModel)
...
             // sync collections
-            foreach (NodeModel node in Model.Nodes) Model_NodeAdded(node);
-            foreach (NoteModel note in Model.Notes) Model_NoteAdded(note);
-            foreach (AnnotationModel annotation in Model.Annotations) Model_AnnotationAdded(annotation);
-            foreach (ConnectorModel connector in Model.Connectors) Connectors_ConnectorAdded(connector);
-            
+            using (Errors.DeferCollectionReset())
+            using (Nodes.DeferCollectionReset())
+            {
+                foreach (NodeModel node in Model.Nodes) Model_NodeAdded(node);
```
pinzart90, line 502: sadly we cannot simply use AddRange or RemoveRange here...because the binding WPF controls enforce the expectation that CollectionChanged of type Add or Remove should only have 1 changed item (not an array)
So the next best thing is to use Reset

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -2478,8 +2495,7 @@ private void LoadAnnotation(ExtraAnnotationViewInfo annotationViewInfo)
                     continue;
 
                 // NOTE: Some nodes may be annotations and not be found here
-                var nodeModel = Nodes.FirstOrDefault(node => node.GUID == guidValue);
-                if (nodeModel == null)
+                if (!nodeMap.TryGetValue(guidValue, out NodeModel nodeModel))
```
mjkkirschner, line 2498: 👍

src/DynamoCoreWpf/ViewModels/Core/ConnectorViewModel.cs
```diff
@@ -480,18 +480,14 @@ public double EndDotSize
 
         public NodeViewModel Nodevm
         {
-            get
-            {
-                return workspaceViewModel.Nodes?.FirstOrDefault(x => x.NodeLogic.GUID == model.Start.Owner.GUID);
```
mjkkirschner, line 485: yikes

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
```diff
@@ -472,6 +471,7 @@ public WorkspaceViewModel(WorkspaceModel model, DynamoViewModel dynamoViewModel)
             Model.NoteRemoved += Model_NoteRemoved;
             Model.NotesCleared += Model_NotesCleared;
 
+            Model.AnnotationsAdded += Model_AnnotationsAdded;
```
mjkkirschner, line 474: maybe a comment on the difference between these events and where they might be triggered?

src/DynamoUtilities/SmartObservableCollection.cs
```diff
@@ -133,7 +135,15 @@ protected override void ClearItems()
         {
             if (Items.Count > 0)
             {
-                base.ClearItems();
+                CheckReentrancy();
```
mjkkirschner, line 138: won't this throw?

pinzart90: Yes...but only if you use the BlockReentrancy API beforehand

mjkkirschner: so then what is the goal of this call?

pinzart90: I am overriding the base class method.
The base class (ObservableCollection) method supports CheckReentrancy.
I decided to keep supporting it in the overridden method  (maybe someone will actually use it later on)...

mjkkirschner: 👍 thanks for the explanation - makes sense.

pinzart90: CheckReentrancy is designed to throw if someone tries to modify the Collection while it is still being modified

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
pinzart90: removed since it does not help performance that much (at least in this case - when loading a workspace)

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
pinzart90: removed since it does not help performance that much in this case 
The WPF control (that binds to these collections) is not created yet (at this stage)

src/DynamoUtilities/SmartObservableCollection.cs
```diff
@@ -24,15 +27,22 @@ public class SmartObservableCollection<T> : ObservableCollection<T>
...
+  
+        public SmartObservableCollection(IEnumerable<T> collection) : base()
+        {
+            try
+            {
+                //Replace internal IList (from Collection<T>.items) with a thread safe IList(ThreadSafeList)
+                var itemsField = typeof(Collection<T>).GetField("items", BindingFlags.NonPublic | BindingFlags.Instance);
+                itemsField?.SetValue(this, new ThreadSafeList<T>(items: collection));
+            }
+            catch { }
```
pinzart90, line 44: If this seems like a bit of a hack...that's because it is (there is no protected or public way to change the inner field `items`)

Pretty safe though...All I am doing is replacing the internal (private) IList `items` that Collection<T> is using under the hood. The `items` field is only accessed as an IList...so passing another implementation of IList will work ok
Here is the source code for Collection<T> https://referencesource.microsoft.com/#mscorlib/system/collections/objectmodel/collection.cs,281923b8611114ec

pinzart90: The only risk I see...is if Microsoft decides to change their internal workings and maybe rename or remove the private field `Collection<T>.items`

In that case we could leave the SmartObservableCollection<T> to be thread-unsafe...

pinzart90: The alternative would be to loose compatibility with the ObservableCollection class (SmartObservableCollection would not be derived from ObservableCollection anymore)

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
```diff
@@ -750,32 +770,30 @@ private void unsubscribeNodeEvents(NodeViewModel nodeViewModel)
...
-                nodeViewModel = Nodes.First(x => x.NodeLogic == node);
-                Errors.Remove(nodeViewModel.ErrorBubble);
+                //unsub the events we attached below in NodeAdded.
+                unsubscribeNodeEvents(nodeViewModel);
+
                 Nodes.Remove(nodeViewModel);
-            }
-            //unsub the events we attached below in NodeAdded.
-            this.unsubscribeNodeEvents(nodeViewModel);
-            nodeViewModel.Dispose();
```
aparajit-pratap, line 762: Don't we still need this Dispose?

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
```diff
@@ -1560,19 +1578,13 @@ private void RefreshViewOnSelectionChange(object sender, NotifyCollectionChanged
...
         {
-            ViewModelBase foundModel = (Connectors.FirstOrDefault(c => c.ConnectorModel.GUID == modelGuid)
-                ?? Nodes.FirstOrDefault(node => node.NodeModel.GUID == modelGuid) as ViewModelBase)
-                ?? (Notes.FirstOrDefault(note => note.Model.GUID == modelGuid)
-                ?? Annotations.FirstOrDefault(annotation => annotation.AnnotationModel.GUID == modelGuid) as ViewModelBase);
-
-            return foundModel;
+            return ViewModelMap.TryGetValue(modelGuid, out ViewModelBase viewModel) ? viewModel : null;
+        }
+        internal VmType GetViewModel<VmType>(Guid modelGuid) where VmType : ViewModelBase
```
aparajit-pratap, line 1585: How about renaming this to `GetViewModelAs`?

src/DynamoUtilities/EnumerableExtensions.cs
```diff
@@ -467,5 +468,30 @@ where attribute.Name.Equals("value")
 
             return string.Empty;
         }
+
+        public static bool TryGetValueAs<Key, Value, ValueAs>(this IDictionary<Key, Value> dictionary, Key key, out ValueAs valueAs) where ValueAs : Value
+        {
+            if (dictionary.TryGetValue(key, out Value value))
+            {
+                valueAs = (ValueAs)value;
```
aparajit-pratap, line 476: Is it possible that this cast fails in some cases?

src/DynamoUtilities/EnumerableExtensions.cs
```diff
@@ -467,5 +468,30 @@ where attribute.Name.Equals("value")
...
+                return true;
+            }
+
+            valueAs = default;
```
aparajit-pratap, line 480: Sorry, not familiar with this syntax, what is default here?

src/DynamoUtilities/EnumerableExtensions.cs
```diff
@@ -467,5 +468,30 @@ where attribute.Name.Equals("value")
...
+            return false;
+        }
+
+
+        public static bool TryRemoveAs<Key, Value, ValueAs>(this ConcurrentDictionary<Key, Value> dictionary, Key key, out ValueAs valueAs) where ValueAs : Value
```
aparajit-pratap, line 485: Why is it that you have `ConcurrentDictionary` as the first parameter here while in the previous function you're using `IDictionary`?

src/DynamoUtilities/EnumerableExtensions.cs
aparajit-pratap: Is assigning `null `not an option here?

src/DynamoUtilities/SmartObservableCollection.cs
```diff
@@ -24,15 +27,25 @@ public class SmartObservableCollection<T> : ObservableCollection<T>
...
+        public SmartObservableCollection(List<T> list) : this(list as IEnumerable<T>)
         {}
+  
+        public SmartObservableCollection(IEnumerable<T> collection) : base()
+        {
+            try
+            {
+                //Replace internal IList (from Collection<T>.items) with a thread safe IList(ThreadSafeList)
+                var itemsField = typeof(Collection<T>).GetField("items", BindingFlags.NonPublic | BindingFlags.Instance);
+                itemsField.SetValue(this, new ThreadSafeList<T>(items: collection));
```
aparajit-pratap, line 42: How expensive is this?

aparajit-pratap: Instead of doing this manipulation, cannot we derive this smart class from a thread-safe observable collection of some sort, or does that not exist?

src/DynamoUtilities/ThreadSafeList.cs
```diff
@@ -0,0 +1,168 @@
...
+using System.Collections;
+using System.Collections.Generic;
+
+namespace DynamoUtilities
+{
+    /// <summary>
+    /// Thread safe List<T>
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    internal class ThreadSafeList<T> : IList<T>
```
aparajit-pratap, line 11: I find it strange that we have to innovate our very own data structures and cannot use something off-the-shelf here?

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
pinzart90: we do...good catch

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
pinzart90: sure

src/DynamoUtilities/EnumerableExtensions.cs
pinzart90: It is possible...I'll adjust the code to handle that

src/DynamoUtilities/EnumerableExtensions.cs
pinzart90: If I tweak the type constraints...then it is

src/DynamoUtilities/EnumerableExtensions.cs
pinzart90: I will switch both to ConcurrentDictionary since that is the only place they are used at the moment

src/DynamoUtilities/SmartObservableCollection.cs
```diff
@@ -24,15 +27,25 @@ public class SmartObservableCollection<T> : ObservableCollection<T>
...
-        public SmartObservableCollection() : base()
+        public SmartObservableCollection() : this(null)
         {}
 
-        public SmartObservableCollection(List<T> list)
-            : base(list)
-        {}
-
-        public SmartObservableCollection(IEnumerable<T> collection) : base(collection)
+        public SmartObservableCollection(List<T> list) : this(list as IEnumerable<T>)
```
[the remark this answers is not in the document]
pinzart90: It is calling another constructor signature of the same class.
I cannot call the same constructor again ..it would be an infinite cycle

src/DynamoUtilities/SmartObservableCollection.cs
pinzart90: I could not find a thread-safe observable collection out of the box.
And the initial direction I took was to keep the SmartObservableCollection class compatible with (derived from) the ObservableCollection class (for API compatibility)

src/DynamoUtilities/ThreadSafeList.cs
pinzart90: There are some thread safe collections of sorts out of the box..,but none that implement IList interface

I chose to create a thread safe IList so that I could replace the inner IList in SmartObservableCollection.ObservableCollection.Collection base class.

src/DynamoUtilities/ThreadSafeList.cs
```diff
@@ -0,0 +1,168 @@
...
+        private readonly DynamoLock _lock = new DynamoLock();
+
+        internal ThreadSafeList()
+        {
+            innerList = new List<T>();
+        }
+
+        internal ThreadSafeList(IList<T> items)
+        {
+            innerList = items != null ? new List<T>(items) : new List<T>();
```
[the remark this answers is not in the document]
pinzart90: I guess it is a matter of preference.
I chose to look at it as a copy constructor.
I did not want to have a situation where the ThreadSafeList references an array that could be manipulated some where else too...Maybe that would defeat the thread safety-ness...

src/DynamoUtilities/ThreadSafeList.cs
```diff
@@ -0,0 +1,168 @@
...
+        {
+            _lock.LockForWrite();
+            innerList.Clear();
+            _lock.UnlockForWrite();
+        }
+
+        public bool Contains(T item)
+        {
+            _lock.LockForRead();
+            bool containts = innerList.Contains(item);
```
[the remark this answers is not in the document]
pinzart90: done

src/DynamoUtilities/ThreadSafeList.cs
```diff
@@ -0,0 +1,168 @@
...
+        public void RemoveAt(int index)
+        {
+            _lock.LockForWrite();
+            innerList.RemoveAt(index);
+            _lock.UnlockForWrite();
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return new TSEnumerator(innerList.GetEnumerator(), _lock);
```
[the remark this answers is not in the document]
pinzart90: TSEnumerator creates a ReadLock for the lifetime of the enumerator
Basically a read lock will be created for iterations of ThreadSafeLists.
Ex
foreach (var item in threadSafeList){...}
at the beginning of the for loop ... the  threadSafeList.GetEnumerator() method will be called. => readLock will be created
At the end of the enumeration, the Enumerator will be disposed ...so readLock will be unlocked

src/DynamoUtilities/SmartObservableCollection.cs
pinzart90: This reflection code is only called when the class is constructed

pinzart90: I had some more comments on this matter in the Conversation tab..but I will copy it here

If this seems like a bit of a hack...that's because it is (there is no protected or public way to change the inner field `items`)

The purpose of doing this forced `set` is to allow SmartObservableCollection instances to be thread safe even when passed around as base class variables.
Ex. 
ObservableCollection obs = new SmartObservableCollection();
Collection obs = new SmartObservableCollection();
In both cases interacting with obs will be thread safe.

All I am doing is replacing the internal (private) IList items that Collection is using under the hood with a ThreaSafeList instance. The items field is only accessed as an IList...so passing another implementation of IList will work ok
Here is the source code for Collection https://referencesource.microsoft.com/#mscorlib/system/collections/objectmodel/collection.cs,281923b8611114ec

The only risk I see...is if Microsoft decides to change their internal workings and maybe rename or remove the private field Collection<T>.items
In that case we now throw an exception....or we could leave the SmartObservableCollection to be thread-unsafe...
 
Alternatives that I can think of:
1. Lose API compatibility with the ObservableCollection and Collection classes and implement(or re-use) an entirely different thread safe collection class
2. Keep API compatibility with the ObservableCollection and Collection classes but loose thread safety when passed around as those types of class variables
3. Give up on thread safety at this level.

pinzart90:
> Curious if all the locking and unlocking for each and every smart collection operation has any performance impact? It looks like in most cases, it may be overkill to make it thread safe so aggressively if that's the case, especially if such cross-threading scenarios will be rare.

The SmartObservableCollection is not meant to be a replacement for all Lists/dictionaries or for highly sensitive areas.
It should mostly be used for WPF bindings (and optionally when dealing with multiple threads).
As an example:
Scheduler thread: Model.smartObservableItems => changed when running a graph
UI thread: View => BindTo Model.smartObservableItems

SmartObservableCollection is already executing a lot of instructions on every collection operation (ex. property change notifications, collection change notifications). Adding locks to the mix should not impact performance in a noticeable way.
I can run performance tests to confirm.

src/DynamoUtilities/EnumerableExtensions.cs
```diff
@@ -469,28 +469,37 @@ where attribute.Name.Equals("value")
             return string.Empty;
         }
 
-        public static bool TryGetValueAs<Key, Value, ValueAs>(this IDictionary<Key, Value> dictionary, Key key, out ValueAs valueAs) where ValueAs : Value
+        /// <summary>
+        /// Returns true if a value of type ValueAs is found at key in the ConcurrentDictionary
+        /// Returns false otherwise. The output value 'valueAs' will be null in this case.
+        /// </summary>
+        public static bool TryGetValueAs<Key, Value, ValueAs>(this ConcurrentDictionary<Key, Value> dictionary, Key key, out ValueAs valueAs) where ValueAs : class
```
aparajit-pratap, line 476: Why did you need to change `where ValueAs : Value` to `where ValueAs : class`?

aparajit-pratap (approved):

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -2469,6 +2469,8 @@ private void LoadAnnotation(ExtraAnnotationViewInfo annotationViewInfo)
 
             var text = annotationViewInfo.Title;
 
+            Dictionary<Guid, NodeModel> nodeMap = Nodes.ToDictionary(x => x.GUID);
```
mjkkirschner, line 2472: Could this be done once before all the Annotations were loaded? Is this just to be sure that new nodes were not loaded between some annotations being loaded?

src/DynamoUtilities/SmartObservableCollection.cs
mjkkirschner: This makes me wonder, why this does not exist already, and if we are in for some quite difficult to debug WPF threading interactions with bindings.

mjkkirschner (commented):
so there are some fundamental changes to the core data structures that are used to bind the view models to the views in this PR.

What type of performance improvement did you notice, and can you characterize where those improvements came from?

I think it is also worth discussing why the smart collection type should be thread safe, I guess because the other collections used locks beforehand and we do not want to break any existing callers?

src/DynamoUtilities/SmartObservableCollection.cs
pinzart90: class Collection<T> can handle passing  custom IList implementations 
```
public Collection(IList<T> list) {
            if (list == null) {
                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.list);
            }
            items = list;
        }
```
ObservableCollection<T> : Collection<T> always tries to create a new list that it then passes to the base class (probably to avoid referencing other ObservableCollections ...)
```
        public ObservableCollection(List<T> list)
            : base((list != null) ? new List<T>(list.Count) : list)
        {
            // Workaround for VSWhidbey bug 562681 (tracked by Windows bug 1369339).
            // We should be able to simply call the base(list) ctor.  But Collection<T>
            // doesn't copy the list (contrary to the documentation) - it uses the
            // list directly as its storage.  So we do the copying here.
            // 
            CopyFrom(list);
        }
```
All list management APIs from ObservableCollections are defered to its base class Collection...
All this said...I do not think we will have issues with `difficult to debug WPF threading interactions with bindings`.

src/DynamoUtilities/SmartObservableCollection.cs
pinzart90: reverted change

src/DynamoUtilities/SmartObservableCollection.cs
pinzart90: Reverted change

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
pinzart90: done

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
```diff
@@ -683,65 +687,81 @@ void Connectors_ConnectorDeleted(ConnectorModel c)
...
             {
+                ViewModelMap.TryRemove(annotationViewModel.AnnotationModel.GUID, out _);
                 annotationViewModel.Dispose();
             }
             Annotations.Clear();
         }
 
         void Model_NodesCleared()
         {
-            lock (Nodes)
```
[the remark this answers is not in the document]
pinzart90: lock has been reintroduced as it was

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
```diff
@@ -750,32 +770,32 @@ private void unsubscribeNodeEvents(NodeViewModel nodeViewModel)
...
-            NodeViewModel nodeViewModel;
-            lock (Nodes)
+            if (ViewModelMap.TryRemoveAs(node.GUID, out NodeViewModel nodeViewModel))
             {
-                nodeViewModel = Nodes.First(x => x.NodeLogic == node);
-                Errors.Remove(nodeViewModel.ErrorBubble);
+                //unsub the events we attached below in NodeAdded.
+                unsubscribeNodeEvents(nodeViewModel);
+
                 Nodes.Remove(nodeViewModel);
```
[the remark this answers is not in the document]
pinzart90: Collections will now sync somewhat automatically ( except when calling Clear on the collections).

pinzart90:
Idea for more improvement is to DelayGraphExecution when calling `DynamoModel::GetConnectorsToAddAndDelete`. At this time...when hooking up a new node to an already connected node, Dynamo creates and schedules 2 UpdateGraphAsyncTasks.

mjkkirschner:
> Idea for more improvement is to DelayGraphExecution when calling `DynamoModel::GetConnectorsToAddAndDelete`. At this time...when hooking up a new node to an already connected node, Dynamo creates and schedules 2 UpdateGraphAsyncTasks.
see
https://jira.autodesk.com/browse/DYN-5128

PR #12978, opened 2022-06-07
Update the type identifier colors to match DS color mapping.

reddyashish:
### Purpose

The type identifier colors have been changed to match the DS look.

Screen Shot 2022-06-07 at 5 13 19 PM

### Declarations

### Release Notes
Update the type identifier colors to match DS color mapping.

### Reviewers
@QilongTang @Amoursol

### DIFF
--- a/src/DynamoCoreWpf/UI/Converters.cs
+++ b/src/DynamoCoreWpf/UI/Converters.cs
@@ -3309,7 +3309,7 @@ namespace Dynamo.Controls
                         return resourceDictionary["PrimaryCharcoal200Brush"] as SolidColorBrush;
                 };
             }
-            return resourceDictionary["PrimaryCharcoal200Brush"] as SolidColorBrush;
+            return resourceDictionary["boolLabelBackground"] as SolidColorBrush;
         }
         public object ConvertBack(object value, Type targetType, object parameter, System.Globalization.CultureInfo culture)
         {
--- a/src/DynamoCoreWpf/UI/Themes/Modern/DynamoColorsAndBrushes.xaml
+++ b/src/DynamoCoreWpf/UI/Themes/Modern/DynamoColorsAndBrushes.xaml
@@ -93,13 +93,13 @@
 
     <!-- NodeLabel colors -->
     <SolidColorBrush x:Key="stringLabelBackground"
-                     Color="#c3ae9d" />
+                     Color="#E0A86F" />
     <SolidColorBrush x:Key="numberLabelBackground"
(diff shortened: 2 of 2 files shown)


src/DynamoCoreWpf/UI/Themes/Modern/DynamoColorsAndBrushes.xaml
```diff
@@ -93,13 +93,13 @@
 
     <!-- NodeLabel colors -->
     <SolidColorBrush x:Key="stringLabelBackground"
-                     Color="#c3ae9d" />
```
QilongTang, line 96: Are these colors already defined elsewhere?

src/DynamoCoreWpf/UI/Converters.cs
```diff
@@ -3309,7 +3309,7 @@ public object Convert(object value, Type targetType, object parameter, System.Gl
                         return resourceDictionary["PrimaryCharcoal200Brush"] as SolidColorBrush;
                 };
             }
-            return resourceDictionary["PrimaryCharcoal200Brush"] as SolidColorBrush;
```
QilongTang, line 3312: hmm why the default is color for boolean?

reddyashish: This is for 'null' objects. PrimaryCharcoal200Brush is for default.

reddyashish:
@Amoursol For the Point.ByCoordinates(), the value returned is a single string "Point.ByCoordinates....". So if we want to style it with multiple colors, we may need to do some string transformations and apply it. A lot of use-cases would overlap.

QilongTang:
> @Amoursol For the Point.ByCoordinates(), the value returned is a single string "Point.ByCoordinates....". So if we want to style it with multiple colors, we may need to do some string transformations and apply it. A lot of use-cases would overlap.

In this case, I am fine with @Amoursol 's suggestion to keep the whole string of `Point.ByCoordinates`'s color unchanged

reddyashish:
Changed it to regular color:

Screen Shot 2022-06-08 at 11 57 33 AM

QilongTang (approved):

Amoursol:
> 

Looks much better! Can we also make sure the "List" value at the top isn't colored? it seems to be 😊 Might also pay to check some nested lists etc. to make sure no other inadvertent colors sneak through.

reddyashish:
Screen Shot 2022-06-08 at 4 02 05 PM

PR #12987, opened 2022-06-09
Minimally invasive notification counter

pinzart90:
### Purpose
Generates log of all notifications generated in Dynamo specifically:
1. WPF binding PropertyChanged events
2. Model NotificationObject's PropertyChanged events
3. ObservableCollection CollectionChanged events

A notification counter that generates/updates a `notifications.log` file in Dynamo's executing assembly folder (every 5 seconds - if any notifications exist). The log file can be deleted at any time (even while Dynamo is running)
Notification logs will always try to be merged/added back to existing logs in the `notifications.log`  folder. Deleting the  `notifications.log` basically allow you to do a fresh count of notifications (should wait 5 seconds after deleting to make sure no cached notifications are added)

Debug builds will have a new dll in the bin folder - 0Harmony.dll
Release builds will not have the new dll.

Instructions to use:
Debug builds : 
- Create/update the debug.config file (in Dynamo's executing assembly folder) to include the `NotificationCounter` debug mode.
- After Dynamo starts, the debug mode can be switched off.

Release builds :
- Copy the 0Harmony.dll from [nuget](https://www.nuget.org/api/v2/package/Lib.Harmony/2.2.1) in Dynamo's executing assembly folder
- Create/update the debug.config file (in Dynamo's executing assembly folder) to include the `NotificationCounter` debug mode.

Example debug.config
```
<?xml version="1.0" encoding="utf-8"?>
<DebugModes>
    <DebugMode name="NotificationCounter" enabled="true"/>
</DebugModes>
```

### Declarations

### Release Notes

### Reviewers

### FYIs

### DIFF
--- a/src/Tools/NotificationCounter/NotificationCounter.cs
+++ b/src/Tools/NotificationCounter/NotificationCounter.cs
@@ -124,6 +124,7 @@ namespace Dynamo.Utilities
                 {
                     patchedCoreNotifications = true;
                     PatchCoreNotifications(assembly.GetType("Dynamo.Core.NotificationObject"));
+                    PatchDynamoModelNotifications(assembly.GetType("Dynamo.Models.DynamoModel"));
                 }
                 if (name == "Microsoft.Practices.Prism" && !patchUINotifications)
                 {
@@ -176,6 +177,16 @@ namespace Dynamo.Utilities
         }
 
         private static void PatchCoreNotifications(Type classType)
+        {
+            var notificationChanged = classType.GetMethod("OnPropertyChanged", BindingFlags.Instance | BindingFlags.NonPublic, null, new Type[] { typeof(string) }, null);
+            var postfix = typeof(NotificationCounter).GetMethod(nameof(CountCoreNotifications), System.Reflection.BindingFlags.Static | System.Reflection.BindingFlags.NonPublic);
+            if (postfix != null && notificationChanged != null)
+            {
+                harmonyIntance.Patch(notificationChanged, null, new HarmonyMethod(postfix));
+            }
+        }
+
+        private static void PatchDynamoModelNotifications(Type classType)
         {
             var notificationChanged = classType.GetMethod("OnPropertyChanged", BindingFlags.Instance | BindingFlags.NonPublic, null, new Type[] { typeof(string) }, null);
             var postfix = typeof(NotificationCounter).GetMethod(nameof(CountCoreNotifications), System.Reflection.BindingFlags.Static | System.Reflection.BindingFlags.NonPublic);


src/DynamoCore/Utilities/NotificationCounter.cs
```diff
@@ -0,0 +1,213 @@
...
+                PatchFromAssmebly(loadedAssemblies.FirstOrDefault(x => x.GetName().Name == "DynamoCore"));
+                PatchFromAssmebly(loadedAssemblies.FirstOrDefault(x => x.GetName().Name == "Microsoft.Practices.Prism"));
+                PatchObservableCollection();
+            }
+            catch { }
+        }
+
+        private static void Timer_Elapsed(object sender, ElapsedEventArgs e)
+        {
+            try
```
mjkkirschner, line 63: why the locks? Will a timer ever fire concurrently?

src/DynamoCore/Utilities/NotificationCounter.cs
```diff
@@ -0,0 +1,213 @@
...
+                lock (logFile)
+                {
+                    if (File.Exists(logFile))
+                    {
+                        var oldContent = File.ReadAllLines(logFile);
+                        mergeLogContents(localNotifications, oldContent);
+                    }
+                }
+
+                var local2 = localNotifications.Select(x => Tuple.Create(x.Key, x.Value.Item1, x.Value.Item2)).ToList();
```
mjkkirschner, line 87: I know this is just developer code, but named tuples are quite nice for readability.

src/DynamoCore/DynamoCore.csproj
```diff
@@ -21,9 +21,11 @@
     <Reference Include="System.Configuration" />
   </ItemGroup>
   <ItemGroup>
-	<PackageReference Include="DynamoVisualProgramming.LibG_227_0_0" Version="2.13.0.8745" GeneratePathProperty="true">
-		<ExcludeAssets>all</ExcludeAssets>
-	</PackageReference>
+    <PackageReference Include="Lib.Harmony" Version="2.2.1" ExcludeAssets="runtime" Condition="'$(Configuration)' != 'Debug'" />
+    <PackageReference Include="Lib.Harmony" Version="2.2.1" ExcludeAssets="none" Condition="'$(Configuration)' == 'Debug'" />
+	  <PackageReference Include="DynamoVisualProgramming.LibG_227_0_0" Version="2.13.0.8745" GeneratePathProperty="true">
```
pinzart90, line 26: Debug builds will copy the harmony dll to the bin folder.
Release builds will not copy it.

QilongTang:
Our team is also touching the notifications display. Is this meant to change how the current `Notifications` view extension works?

mjkkirschner:
@QilongTang I'm pretty sure "Notifications" here refers to property changes /binding updates in WPF.

mjkkirschner:
@pinzart90 do you think we could call this something like binding events or something like that instead of notifications to disambiguate this from the user notifications work?

src/DynamoCore/DynamoCore.csproj
```diff
@@ -21,6 +21,8 @@
     <Reference Include="System.Configuration" />
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="Lib.Harmony" Version="2.2.1" ExcludeAssets="runtime" Condition="'$(Configuration)' != 'Debug'" />
```
mjkkirschner, line 24: so in release builds we exclude the runtime assets, and in debug builds we don't... we need the non debug condition so the build succeeds?

I don't know if this means we should add harmony to the about box, license info etc since even though we don't distribute it in release builds we still pull it.

pinzart90: Ugh..you're right. That would need legal clarification.
I rather have devs need to pull it manually than go through another round with legal

mjkkirschner: haha, what about using reflection to call it so theres no compile time dep?

mjkkirschner: the other thing to consider @pinzart90 is that MIT is a preapproved license so I think you can just use it and update the license info without discussion.

pinzart90: `about using reflection to call it so theres no compile time dep`
Makes development really tough ...
I updated the about box, as you suggested

LICENSE.txt
```diff
@@ -530,6 +530,16 @@ Licensing information: © Autodesk, Inc. All Rights Reserved.
 
 The Artifakt font software is Autodesk proprietary and confidential, and may be used only by authorized users and only for Autodesk business purposes. Any use not authorized by Autodesk is not permitted and is an infringement of Autodesk's intellectual property rights as well as a breach of your agreement with Autodesk. Go to https://brand.autodesk.com/brand-system/typography for detailed usage guidelines on when and how to use the Artifakt designer collection.
 
+Lib.Harmony v.2.2.1
+Copyright (c) 2017 Andreas Pardeike
```
mjkkirschner, line 534: can you add the MIT license type here as well so it's easy to identify without comparing legalese?

src/DynamoCore/DynamoCore.csproj
```diff
@@ -128,7 +130,6 @@
       <NodeHelpFiles Include="$(SolutionDir)..\doc\distrib\NodeHelpFiles\**\*.*" />
       <OpenSourceLicenses Include="$(SolutionDir)..\doc\distrib\Open Source Licenses\**\*.*" />
       <LocalizedResources Include="$(SolutionDir)..\extern\Localized\**\*.*" />
-      <ExternAnalytics Include="$(SolutionDir)..\extern\Analytics.NET\*.*" />
```
mjkkirschner, line 131: another one I missed, why is this removed? Should it have been removed as part of your other PR?

src/DynamoCore/Utilities/NotificationCounter.cs
```diff
@@ -0,0 +1,235 @@
...
+
+namespace Dynamo.Utilities
+{
+    internal class NotificationLog
+    {
+        internal SortedSet<int> objects = new SortedSet<int>();
+        internal int counter;
+    }
+
+    internal static class NotificationCounter
```
mjkkirschner, line 19: I think this name is still confusing for devs working on notifications and will be in the future as well, especially since this class has no summary.

mjkkirschner (commented):
just a few more comments!

also if you reboot dynamo will the notifications log be merged between sessions?

mjkkirschner:
@pinzart90 it also appears that the test job for this PR has failed a few times in a row while running the tests, log is not telling me much but have not deep dived, it seems to be running on dyn_perf_002 - maybe take a look on that machine.

PR #12927, opened 2022-05-25
DYN-4880: Dynamo crash upon clicking Packages > Arkance Systems Settings

sm6srw:
### Purpose

We can't prevent the crash in DYN-4880 but we can improve our messaging when users install packages that targets other hosts.

image

Normal case: Warn if a package target other hosts but not the current host
Special case: Sandbox, warn if a package target only one known host (a conservative way to not warn for packages that works for all hosts). 

### Declarations

Check these if you believe they are true

- [X] The codebase is in a better state after this PR
- [X] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [X] The level of testing this PR includes is appropriate
- [X] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [X] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.
- [ ] This PR modifies some build requirements and the readme is updated

### Release Notes

Improves messaging when a user install a package that target a different environment.

### Reviewers

(FILL ME IN) Reviewer 1  (If possible, assign the Reviewer for the PR)

(FILL ME IN, optional) Any additional notes to reviewers or testers.

### FYIs

(FILL ME IN, Optional) Names of anyone else you wish to be notified of

PR #12990, opened 2022-06-09
raise notification at startup if installed package targets a different host.

mjkkirschner:
### Purpose
https://jira.autodesk.com/browse/DYN-4884
Follow up for the work @sm6srw started:  https://jira.autodesk.com/browse/DYN-4880

PR does the following:
* refactors host conflict check into `package manager extension` and out of `package manager client view model`.
* updates some tests
* At startup, raises notifications if the check fails for any packages that installed already. I went with notifications over message boxes to avoid irritating/overwhelming users because these notifications will show up at each startup.

Screen Shot 2022-06-09 at 6 01 17 PM
notification after install.
camber notification install
notification when adding package path.
camber notification add pack dialog

In addition to these cases, startup also generates a notification of course.

### Declarations

### Release Notes

Add notifications at startup for packages that target a different host than current Dynamo context.

### Reviewers

### FYIs

### DIFF
--- a/src/DynamoCore/Extensions/ReadyParams.cs
+++ b/src/DynamoCore/Extensions/ReadyParams.cs
@@ -77,6 +77,9 @@ namespace Dynamo.Extensions
             get { return commandExecutive ?? (commandExecutive = new ExtensionCommandExecutive(dynamoModel)); }
         }
 
+        /// <summary>
+        /// HostInfo object, Useful to determine what host context Dynamo is running in.
+        /// </summary>
         internal HostAnalyticsInfo HostInfo => dynamoModel.HostAnalyticsInfo;
 
 
--- a/src/DynamoCore/Extensions/ReadyParams.cs
+++ b/src/DynamoCore/Extensions/ReadyParams.cs
@@ -76,7 +76,10 @@ namespace Dynamo.Extensions
         {
             get { return commandExecutive ?? (commandExecutive = new ExtensionCommandExecutive(dynamoModel)); }
         }
-        
+
+        internal HostAnalyticsInfo HostInfo => dynamoModel.HostAnalyticsInfo;
+
+
         /// <summary>
         /// Event that is raised when the Dynamo Logger logs a notification.
         /// This event passes the notificationMessage to any subscribers
--- a/src/DynamoCoreWpf/Properties/Resources.en-US.resx
+++ b/src/DynamoCoreWpf/Properties/Resources.en-US.resx
@@ -3235,9 +3235,6 @@ You can manage this in Preferences -&gt; Security.</value>
   <data name="MessagePackageTargetOtherHostShort" xml:space="preserve">
     <value>Package targets a different host than current host application.</value>
   </data>
-  <data name="String1" xml:space="preserve">
-    <value />
-  </data>
   <data name="TitlePackageTargetOtherHost" xml:space="preserve">
     <value>Package Host Error</value>
   </data>
--- a/src/DynamoCoreWpf/Properties/Resources.en-US.resx
+++ b/src/DynamoCoreWpf/Properties/Resources.en-US.resx
@@ -3229,4 +3229,16 @@ You can manage this in Preferences -&gt; Security.</value>
   <data name="TrustLocationSkippedNotification" xml:space="preserve">
     <value>No trusted location has been added.</value>
   </data>
+  <data name="MessagePackageTargetOtherHosts2" xml:space="preserve">
+    <value>The package or one of its dependencies targets a different environment, such as Revit, Civil 3D, Advance Steel, Alias or FormIt. This can cause instability and unexpected problems.</value>
+  </data>
+  <data name="MessagePackageTargetOtherHostShort" xml:space="preserve">
+    <value>Package targets a different host than current host application.</value>
+  </data>
+  <data name="String1" xml:space="preserve">
+    <value />
+  </data>
+  <data name="TitlePackageTargetOtherHost" xml:space="preserve">
+    <value>Package Host Error</value>
+  </data>
 </root>
--- a/src/DynamoCoreWpf/Properties/Resources.en-US.resx
+++ b/src/DynamoCoreWpf/Properties/Resources.en-US.resx
@@ -3154,10 +3154,10 @@ To install the latest version of a package, click Install. \n
     <value>Settings</value>
   </data>
   <data name="FileTrustWarningPopupText" xml:space="preserve">
-    <value>This file is stored in an untrusted location. Do you want to run this file?</value>
+    <value>This file is stored in an untrusted location. Do you want to open this file?</value>
   </data>
   <data name="FileTrustWarningPopupTitle" xml:space="preserve">
-    <value>Run external file?</value>
+    <value>Open external file?</value>
   </data>
   <data name="FileTrustWarningPopupTooltip1" xml:space="preserve">
     <value>External content may bring risks to your system security.</value>
@@ -3221,5 +3221,12 @@ You can manage this in Preferences -&gt; Security.</value>
   </data>
   <data name="FooterNotificationWarningTooltip" xml:space="preserve">
     <value>Click to cycle through nodes with warnings.</value>
+  </data>
+  <data name="TrustLocationAddedNotification" xml:space="preserve">
+    <value>You have trusted file location:
+{0}</value>
+  </data>
+  <data name="TrustLocationSkippedNotification" xml:space="preserve">
+    <value>No trusted location has been added.</value>
   </data>
 </root>
--- a/src/DynamoCoreWpf/Properties/Resources.resx
+++ b/src/DynamoCoreWpf/Properties/Resources.resx
@@ -3216,4 +3216,13 @@ You can manage this in Preferences -&gt; Security.</value>
   <data name="TrustLocationSkippedNotification" xml:space="preserve">
     <value>No trusted location has been added.</value>
   </data>
+  <data name="MessagePackageTargetOtherHosts2" xml:space="preserve">
+    <value>The package or one of its dependencies targets a different environment, such as Revit, Civil 3D, Advance Steel, Alias or FormIt. This can cause instability and unexpected problems.</value>
+  </data>
+  <data name="MessagePackageTargetOtherHostShort" xml:space="preserve">
+    <value>Package targets a different host than current host application.</value>
+  </data>
+  <data name="TitlePackageTargetOtherHost" xml:space="preserve">
+    <value>Package Host Error</value>
+  </data>
 </root>
--- a/src/DynamoCoreWpf/Properties/Resources.resx
+++ b/src/DynamoCoreWpf/Properties/Resources.resx
@@ -3141,10 +3141,10 @@ To install the latest version of a package, click Install. \n
     <value>Settings</value>
   </data>
   <data name="FileTrustWarningPopupText" xml:space="preserve">
-    <value>This file is stored in an untrusted location. Do you want to run this file?</value>
+    <value>This file is stored in an untrusted location. Do you want to open this file?</value>
   </data>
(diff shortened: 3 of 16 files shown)


src/DynamoPackages/PackageManagerExtension.cs
```diff
@@ -341,6 +347,36 @@ private void OnPackageRemoved(Package package)
...
+            }
+            else
+            {
+                // Warn if there are packages targeting other hosts but not our host
+                var otherHosts = knownHosts.Except(new List<string>() { Host });
+                containsPackagesThatTargetOtherHosts = newPackageHeaders.Any(x =>
+                {
+                    // Is our host in the list?
+                    // If not, is any other host in the list?
+                    return x.host_dependencies != null && !x.host_dependencies.Contains(Host) && otherHosts.Any(y => x.host_dependencies != null && x.host_dependencies.Contains(y));
```
zeusongit, line 372: Curious here, why do we need to check `otherHosts `? Can't we just check if our host is in the list and if not display the available list of hosts?

mjkkirschner: I think his comment sums it up nicely:
`// Warn if there are packages targeting other hosts but not our host`

QilongTang: I guess by doing this we can enable unloading these packages in the future at some point of Dynamo startup? It would be good to know the reason behind this for sure

src/DynamoCore/Extensions/ReadyParams.cs
```diff
@@ -76,7 +76,10 @@ public virtual ICommandExecutive CommandExecutive
         {
             get { return commandExecutive ?? (commandExecutive = new ExtensionCommandExecutive(dynamoModel)); }
         }
-        
+
+        internal HostAnalyticsInfo HostInfo => dynamoModel.HostAnalyticsInfo;
```
QilongTang, line 80: Would you add some comments to this and what it means?

mjkkirschner: yes, I will, but note that I have made it internal on purpose for the time being.

src/DynamoPackages/PackageManagerExtension.cs
```diff
@@ -341,6 +347,36 @@ private void OnPackageRemoved(Package package)
...
+            var containsPackagesThatTargetOtherHosts = false;
+
+            // Known hosts
+            var knownHosts = PackageManagerClient.GetKnownHosts();
+
+            // Sandbox, special case: Warn if any package targets only one known host
+            if (String.IsNullOrEmpty(Host))
+            {
+                containsPackagesThatTargetOtherHosts =
+                    newPackageHeaders.Any(y => y.host_dependencies != null && y.host_dependencies.Intersect(knownHosts).Count() == 1);
```
aparajit-pratap, line 362: Why check only if the package targets one host? In other words, shouldn't this check be `... && y.host_dependencies.Intersect(knownHosts).Count() >= 1`

src/DynamoPackages/PackageManagerExtension.cs
aparajit-pratap: In:
```
return x.host_dependencies != null && !x.host_dependencies.Contains(Host) && otherHosts.Any(y => x.host_dependencies != null && x.host_dependencies.Contains(y));
```
why do we need to check `x.host_dependencies != null` twice?

src/DynamoPackages/PackageManagerExtension.cs
mjkkirschner: this is a special case for sandbox - see the comment, because `host_dependencies` does not have a `core` or `sandbox` entry, the usual pattern is for authors to select *all the hosts* when they write a package for core. If a package only targets one host, there is a good chance this package does not work on sandbox.

mjkkirschner: IMO `host_dependencies` need some redesign at some point in the future.

src/DynamoPackages/PackageManagerExtension.cs
mjkkirschner: @aparajit-pratap good point, I think we can get rid of that check inside the `Any` statement.

src/DynamoCoreWpf/Properties/Resources.resx
```diff
@@ -3216,4 +3216,13 @@ You can manage this in Preferences -&gt; Security.</value>
...
   </data>
+  <data name="MessagePackageTargetOtherHosts2" xml:space="preserve">
+    <value>The package or one of its dependencies targets a different environment, such as Revit, Civil 3D, Advance Steel, Alias or FormIt. This can cause instability and unexpected problems.</value>
+  </data>
+  <data name="MessagePackageTargetOtherHostShort" xml:space="preserve">
+    <value>Package targets a different host than current host application.</value>
+  </data>
+  <data name="TitlePackageTargetOtherHost" xml:space="preserve">
+    <value>Package Host Error</value>
+  </data>
```
aparajit-pratap, line 3227: Aren't one or more of these messages already added previously here: https://github.com/DynamoDS/Dynamo/pull/12927?

aparajit-pratap (approved):

src/DynamoCoreWpf/Properties/Resources.resx
mjkkirschner: no, they are slightly different, @sm6srw's messages include a question - do you want to proceed? etc - since this is a notification and not a message box, and we've already loaded the package, that did not make sense to me.

aparajit-pratap:
Last question, can we perform these checks for case #2 `When a new package directory is added` pointed out by Jorgen here https://jira.autodesk.com/browse/DYN-4884?

mjkkirschner:
@aparajit-pratap PTAL - implemented notifications when adding package paths.

src/DynamoCoreWpf/Properties/Resources.en-US.resx
```diff
@@ -3229,4 +3229,13 @@ You can manage this in Preferences -&gt; Security.</value>
   <data name="TrustLocationSkippedNotification" xml:space="preserve">
     <value>No trusted location has been added.</value>
   </data>
+  <data name="MessagePackageTargetOtherHosts2" xml:space="preserve">
+    <value>The package or one of its dependencies targets a different environment, such as Revit, Civil 3D, Advance Steel, Alias or FormIt. This can cause instability and unexpected problems.</value>
```
aparajit-pratap, line 3233: Can we be specific about which host specifically the package is targeting rather than saying it could be one of the many?

src/DynamoPackagesWPF/PackageManagerViewExtension.cs
```diff
@@ -55,6 +59,19 @@ public string UniqueId
             remove { layouthandler -= value; }
         }
 
+        Action<NotificationMessage> notificationLogged;
+        event Action<NotificationMessage> INotificationSource.NotificationLogged
+        {
+            add
+            {
+             notificationLogged+=value;
```
aparajit-pratap, line 67: formatting

src/DynamoPackagesWPF/PackageManagerViewExtension.cs
```diff
@@ -64,9 +81,15 @@ public void Dispose()
         public void Loaded(ViewLoadedParams viewLoadedParams)
         {
             RequestLoadLayoutSpecs(packageManager.PackageLoader.LocalPackages);
+            var packagesToCheck = packageManager?.PackageLoader.LocalPackages;
+            if(packagesToCheck != null)
+            {
+                RaisePackageHostNotifications(packagesToCheck);
+            }
```
aparajit-pratap, line 88: Can you explain again why are we raising them here again?

mjkkirschner: This hook (`IViewExtension.Loaded`) occurs when the `PackageManagerViewExtension` is loaded, this is basically at startup time, so there are a set of packages that were just loaded that we need to check. 

`PackageLoaded_Handler` will handle late loaded extensions. Since this is a view extension, this class is not even loaded until *after* the packageManager finishes loading the packages in the users's package locations at startup.

It's the same pattern for the other things this extension checks, layoutspecs, view extensions etc.

src/DynamoCoreWpf/Properties/Resources.en-US.resx
aparajit-pratap: Yes, I think I meant just the first host.

src/DynamoPackagesWPF/PackageManagerViewExtension.cs
aparajit-pratap: Won't the `packagesToCheck` already be checked for hosts (using `CheckIfPackagesTargetOtherHosts`) by the package manager when it's loading the packages for the first time? If not, shouldn't we be doing the check at package load time?

mjkkirschner: No it won't be done by the package manager at package load time, the method is in that class so someone can use it form there, but I decided to defer it in this case, because at that point the view is not loaded so notifications / message boxes will not be able to be raised.

I put these checks here so that we ensure the view is loaded, that means other extensions like `NotificationsViewExtension` are ready to actually log notifications.

If we move some of the checks to `packageManagerExtension` then we will need to build up a list of notifications and send them to NotificationsViewExtension at some point later.

I can do that, but it seemed simpler to have all the checks done here - unless you think it is important to do these checks in headless mode as well - the big caveat being, the `NotificationViewExtension` is not loaded in headless mode so sending notifications won't do much! It will still log.

aparajit-pratap: Understood, thanks for the explanation. I overlooked the fact that none of the view extensions would be ready at that point in time. I'm okay with this approach.

aparajit-pratap (approved):
@mjkkirschner thanks for responding to my questions. As far as the message is concerned about the non-matching host, feel free to disregard it if it's too difficult a refactor. It's a minor point as I thought it might be nice to have that additional info for the user as to what host their downloaded package is actually targeting.

mjkkirschner:
@aparajit-pratap @QilongTang  given the timing, I will merge this, and cherry pick to 2.15 given this task is scoped for release in jira.

I can attempt to make improvements to the message in master.

mjkkirschner:
pass here:
https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/6375/

I ran it a few more times and it looks like another one of those issues with checking `ContextMenuPopUp.IsOpen` being flaky... it does not fail consistently, fairly certain we've seen this failure before. 
I don't think it's related to these changes.

mzjensen:
Hi @mjkkirschner,

I noticed this new notification now that Dynamo Core 2.15 is shipped with the latest Civil 3D update (2023.1). It's convenient that you were doing your testing with my package as well! I find the message a little confusing because it says that my package targets a different host, which would be true if it was installed for Sandbox. But why does this message show in Dynamo for Civil 3D? Is it a bug, or do I need to change something in my pkg.json file? My ultimate goal is for this notification to _not_ appear for Civil 3D users so they don't wonder if they did something wrong.

mjkkirschner:
Hi @mzjensen that sounds like a bug because it looks like your package is already marks a host dependency on Civil3d. I will file a bug to investigate.

mzjensen:
@mjkkirschner - sounds good, thanks. FWIW, I also see the same message when installing the IronPython 2.7 package, which doesn't seem right.

mjkkirschner:
yes, I think if you were to debug you would see that Civil 3D is providing an unexpected host name...

mjkkirschner:
at the moment I'm not clear if this will work - but as a workaround you could try not setting the host dependencies to anything, and publishing locally to see if that works to avoid the notification.

PR #12997, opened 2022-06-14
DYN-5018 wires disappear when moving group

filipeotero:
### Purpose

In this PR, the position of the wires is calculated based on the height of the port that is created after the group is collapsed.
Previously, we were assuming that all the ports have the same height, causing a mispositioning of the wires. 

image

image

Positionwires

### Declarations

### Reviewers

@QilongTang @pinzart @reddyashish

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
@@ -594,13 +594,14 @@ namespace Dynamo.ViewModels
             LoadGroupStylesFromPreferences(preferenceSettings.GroupStyleItemsList);
         }
 
-        
+
         /// <summary>
-            /// Creates input ports for the group based on its Nodes.
-            /// Input ports that either is connected to a Node outside of the
-            /// group, or has a port that is not connected will be used for the group.
-            /// </summary>
-        internal void SetGroupInputPorts()
+        /// Creates input ports for the group based on its Nodes.
+        /// Input ports that either is connected to a Node outside of the
+        /// group, or has a port that is not connected will be used for the group.
+        /// This function appends to the inputs
+        /// </summary>
+        private void SetGroupInputPorts()
         {
             List<PortViewModel> newPortViewModels;
 
@@ -635,8 +636,9 @@ namespace Dynamo.ViewModels
         /// <summary>
         /// Creates output ports for the group based on its Nodes.
         /// Output ports that are not connected will be used for the group.
+        /// This function appends to the outports
         /// </summary>
-        internal void SetGroupOutPorts()
+        private void SetGroupOutPorts()
         {
             List<PortViewModel> newPortViewModels;
 
--- a/src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
@@ -249,7 +249,7 @@ namespace Dynamo.ViewModels
                 OutPorts.Clear();
                 if (value)
                 {
-                    this.ShowGroupContents();
+                    this.ShowGroupContents();       
                 }
                 else
                 {
@@ -725,13 +725,10 @@ namespace Dynamo.ViewModels
             return verticalOffset + (proxyPortIndex * portHeight) + portVerticalMidPoint;
         }
 
-        private Point2D CalculatePortPosition(PortModel portModel, int proxyPortIndex)
+        private Point2D CalculatePortPosition(PortModel portModel, double verticalPosition)
         {
             double groupHeaderHeight = Height - ModelAreaRect.Height;
-
-            double offset = GetPortVerticalOffset(portModel, proxyPortIndex);
-            double y = Top + groupHeaderHeight + offset;
-
+            double y = Top + groupHeaderHeight + verticalPosition + verticalOffset + portVerticalMidPoint;
             switch (portModel.PortType)
             {
                 case PortType.Input:
@@ -740,7 +737,7 @@ namespace Dynamo.ViewModels
                     if (portModel.Owner is CodeBlockNodeModel)
                     {
                         // Special case because code block outputs are smaller than regular outputs.
-                        return new Point2D(Left + Width, y + 6);
+                        return new Point2D(Left + Width, y - 8);
                     }
                     return new Point2D(Left + Width, y);
             }
@@ -755,15 +752,19 @@ namespace Dynamo.ViewModels
                 .ToList();
 
             var newPortViewModels = new List<PortViewModel>();
-            for (int i = 0; i < groupPortModels.Count(); i++)
+            double verticalPosition = 0;
+            foreach (var group in groupPortModels)
             {
-                var model = groupPortModels.ElementAt(i);
-                newPortViewModels.Add(originalPortViewModels[i].CreateProxyPortViewModel(model));
-
-                // calculate new position for the proxy inports.
-                model.Center = CalculatePortPosition(model, i);
+                var originalPort = originalPortViewModels.FirstOrDefault(x => x.PortModel.GUID == group.GUID);
+                if (originalPort != null)
+                {
+                    var portViewModel = originalPort.CreateProxyPortViewModel(group);
+                    newPortViewModels.Add(portViewModel);
+                    // calculate new position for the proxy outports
+                    group.Center = CalculatePortPosition(group, verticalPosition);
+                    verticalPosition += originalPort.Height;
+                }
             }
-
             return newPortViewModels;
         }
 
@@ -775,7 +776,7 @@ namespace Dynamo.ViewModels
                 .ToList();
 
             var newPortViewModels = new List<PortViewModel>();
-            int index = 0;
+            double verticalPosition = 0;
             foreach (var group in groupPortModels)
             {
                 var originalPort = originalPortViewModels.FirstOrDefault(x => x.PortModel.GUID == group.GUID);
@@ -784,9 +785,9 @@ namespace Dynamo.ViewModels
                     var portViewModel = originalPort.CreateProxyPortViewModel(group);
                     newPortViewModels.Add(portViewModel);
                     // calculate new position for the proxy outports
-                    group.Center = CalculatePortPosition(group, index);
+                    group.Center = CalculatePortPosition(group, verticalPosition);
+                    verticalPosition += originalPort.Height;
                 }
-                index++;
             }
 
             return newPortViewModels;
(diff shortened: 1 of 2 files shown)


src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -260,6 +262,15 @@ public bool IsExpanded
                 AddGroupToGroupCommand.RaiseCanExecuteChanged();
                 RaisePropertyChanged(nameof(IsExpanded));
                 RedrawConnectors();
+                ReportNodesPosition();
```
pinzart90, line 265: `node.ReportPosition` is an expensive operation when dealing with a lot of nodes under the same group.
Please check the performance with (there should be a .dyn you can use in this [jira taks](https://jira.autodesk.com/browse/DYN-4600)

Also take a look at what code paths might be calling the IsExpanded setter (to get a better idea what to test) - could be direct function calls or RaisePropertyChanges...

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -746,15 +752,19 @@ private List<PortViewModel> CreateProxyInPorts(IEnumerable<PortModel> groupPortM
...
-            for (int i = 0; i < groupPortModels.Count(); i++)
+            double verticalPosition = 0;
+            foreach (var group in groupPortModels)
             {
-                var model = groupPortModels.ElementAt(i);
-                newPortViewModels.Add(originalPortViewModels[i].CreateProxyPortViewModel(model));
-
-                // calculate new position for the proxy inports.
-                model.Center = CalculatePortPosition(model, i);
+                var originalPort = originalPortViewModels.FirstOrDefault(x => x.PortModel.GUID == group.GUID);
```
mjkkirschner, line 758: I'm, confused by then, but not super familiar with this code, when would the entire guid of a group equal the guid of a port model?

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -245,9 +245,11 @@ public bool IsExpanded
             set
             {
                 annotationModel.IsExpanded = value;
```
pinzart90, line 247: I think a better option would be to listen to the model.IsExpanded OnPropertyChanged...instead of the entire setter
(this can be done in another PR) 

annotationModel.PropertyChanged += OnModelPropertyChanged;
OnModelPropertyChanged(sender, e) {
   if (e.PropertyName == "IsExpanded") {
                      InPorts.Clear();
                OutPorts.Clear();
                if (value)
                {
                    this.ShowGroupContents();       
                }
                else
                {
                    this.SetGroupInputPorts();
                    this.SetGroupOutPorts();
                    this.CollapseGroupContents(true);
                    RaisePropertyChanged(nameof(NodeContentCount));
                }
                WorkspaceViewModel.HasUnsavedChanges = true;
                .............the whole setter content............
   }
}

ANd only keep the getter

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -1081,10 +1097,12 @@ private void model_PropertyChanged(object sender, System.ComponentModel.Property
                     RaisePropertyChanged("Width");
                     RaisePropertyChanged(nameof(ModelAreaRect));
                     UpdateAllGroupedGroups();
+                    UpdateProxyPortsPosition();
```
pinzart90, line 1100: check performance here...
When moving a group (drag with mouse) ...these properties can get changed a lot of times

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
filipeotero: It is for when the group is uncollapsed and the annotation inside it needs to report the position. Otherwise, that's what happens:
annotationNode

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
filipeotero: It's the group's port. I will change the variable to groupPort to avoid confusion. Thanks.

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
filipeotero: Ok, let's do that on another PR

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
filipeotero: Ok, I will check an alternative for this.

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
pinzart90: If there is no performance penalty in doing "node.ReportPosition" then it is fine.
The fix would be to make the Connectors redraw when the corresponding NodeModels(start or end) change their IsProxyPort flag
https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/ViewModels/Core/ConnectorViewModel.cs#L1209

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
pinzart90: check if performance is an issue first...then check alternatives
No use in doing premature optimization

filipeotero:
Instead of calling the `UpdateProxyPortsPosition` method every time the width changes, the method is being called on `GroupTextBox_OnTextChanged ` which was updating of the width of the group, but not the center of its ports.

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -536,11 +547,15 @@ private bool CanChangeFontSize(object obj)
         public AnnotationViewModel(WorkspaceViewModel workspaceViewModel, AnnotationModel model)
         {
             annotationModel = model;
+            var modelBase = (ModelBase)model;
```
pinzart90, line 550: why do you need to cast it to ModelBase ?

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -536,11 +547,15 @@ private bool CanChangeFontSize(object obj)
...
+
             this.WorkspaceViewModel = workspaceViewModel;
             this.preferenceSettings = WorkspaceViewModel.DynamoViewModel.PreferenceSettings;
             model.PropertyChanged += model_PropertyChanged;
             model.RemovedFromGroup += OnModelRemovedFromGroup;
             model.AddedToGroup += OnModelAddedToGroup;
+            modelBase.PropertyChanged += ModelBase_PropertyChanged;
```
pinzart90, line 557: there already is a handler for the model PropertyChange event
```
model.PropertyChanged += model_PropertyChanged;
```

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -625,7 +649,6 @@ internal void SetGroupInputPorts()
         /// </summary>
         internal void SetGroupOutPorts()
         {
-            OutPorts.Clear();
```
pinzart90, line 628: why did you remove `OutPorts.Clear();` ?

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -581,14 +596,23 @@ public AnnotationViewModel(WorkspaceViewModel workspaceViewModel, AnnotationMode
...
-        /// Input ports that either is connected to a Node outside of the
-        /// group, or has a port that is not connected will be used for the group.
-        /// </summary>
+            /// Creates input ports for the group based on its Nodes.
+            /// Input ports that either is connected to a Node outside of the
+            /// group, or has a port that is not connected will be used for the group.
+            /// </summary>
         internal void SetGroupInputPorts()
         {
-            InPorts.Clear();
```
pinzart90, line 591: why did you remove `InPorts.Clear();` ?

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
filipeotero: Removed

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -581,14 +596,23 @@ public AnnotationViewModel(WorkspaceViewModel workspaceViewModel, AnnotationMode
             LoadGroupStylesFromPreferences(preferenceSettings.GroupStyleItemsList);
         }
 
+        private void ModelBase_PropertyChanged(object sender, System.ComponentModel.PropertyChangedEventArgs e)
+        {
+            switch (e.PropertyName)
+            {
+                case nameof(annotationModel.Position):
+                    UpdateProxyPortsPosition();
```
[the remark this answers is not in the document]
filipeotero: removed

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
filipeotero: Same here

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
```diff
@@ -581,14 +594,14 @@ public AnnotationViewModel(WorkspaceViewModel workspaceViewModel, AnnotationMode
...
+        
         /// <summary>
-        /// Creates input ports for the group based on its Nodes.
-        /// Input ports that either is connected to a Node outside of the
-        /// group, or has a port that is not connected will be used for the group.
-        /// </summary>
+            /// Creates input ports for the group based on its Nodes.
+            /// Input ports that either is connected to a Node outside of the
+            /// group, or has a port that is not connected will be used for the group.
+            /// </summary>
```
pinzart90, line 602: looks like the formatting got screwed up

src/DynamoCoreWpf/ViewModels/Core/AnnotationViewModel.cs
pinzart90: so now `SetGroupInputPorts` and `SetGroupOutPorts` will always append ports to `OutPorts` and `InPorts`.  I hope nobody uses these functions expecting the port arrays to be `re-set` ...as the func names suggest. Can you make these functions private and maybe add to the description that they always append to the respective port arrays ?

filipeotero: Sure, editted.

pinzart90 (approved):

QilongTang (approved):

PR #12693, opened 2022-03-14
temp fix for builds, remove tsplines nodes docs

mjkkirschner:
### Purpose

Currently performance CI job, and WPF visualization - perf machine test jobs are down - it appears both have to do with the new docs files that have extremely long file names, like: 
`doc/distrib/NodeHelpFiles/Autodesk.DesignScript.Geometry.TSpline.TSplineSurface.ByBoxLengths(origin, width, length, height, xSpans, ySpans, zSpans, symmetry, inSmoothMode).md`

This PR simply removes the 3 files with the longest file names. There are some others, but I want to remove the minimal set for now.

I will file a followup task to consider solutions to this issue in the docs browser and docs generator tools. I'm thinking something like generating a unique short name using a guid, standard hash, or some short name scheme etc. 

There are other solutions as well for the nodes we maintain, like storing them in a binary file/resx which the docs browser parses at runtime.

### Declarations

Check these if you believe they are true

- [ ] The codebase is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.
- [ ] This PR modifies some build requirements and the readme is updated

### Release Notes

(FILL ME IN) Brief description of the fix / enhancement. **Mandatory section** 

### Reviewers

(FILL ME IN) Reviewer 1  (If possible, assign the Reviewer for the PR)

(FILL ME IN, optional) Any additional notes to reviewers or testers.

### FYIs

(FILL ME IN, Optional) Names of anyone else you wish to be notified of

PR #13023, opened 2022-06-16
Revert "Revert "temp fix for builds, remove tsplines nodes docs (#12693)""

mjkkirschner:
Reverts DynamoDS/Dynamo#12984

a whole bunch of reverts - @QilongTang found this still causes a failure in the Binary diff job:
https://autodesk.slack.com/archives/C1866AFQB/p1655407837498549

we'll have to revisit after 2.15 - in any case these files are empty anyway and don't contain any useful documentation currently.


QilongTang:
Thanks @mjkkirschner I would like to try this change first on `mater` to see if the problem for binary diff can be fixed

QilongTang (approved):

PR #13055, opened 2022-06-21
Re-add allowrankreduction attribute to Dictionary.Values

aparajit-pratap:
### Purpose

Fixes a case like this that would fail. It now works as expected.
image

### Declarations

This introduces an API break but I think it's a necessary fix. Alternatively, we can add a new property with a different but similar name to `Values` and deprecate `Values`.

### Release Notes

### Reviewers

@saintentropy 

### FYIs

@Amoursol what do you think?

### DIFF
--- a/src/Libraries/DesignScriptBuiltin/Dictionary.cs
+++ b/src/Libraries/DesignScriptBuiltin/Dictionary.cs
@@ -59,8 +59,7 @@ namespace DesignScript
             ///     Produces the values in a Dictionary.
             /// </summary>
             /// <returns name="values">Values of the dictionary</returns>
-            // [AllowRankReduction]
-            // TODO: Consider adding this attribute in 3.0 (DYN-1697)
+            [AllowRankReduction]
             public IEnumerable<object> Values => D.Values;
 
             /// <summary>


Amoursol:
@aparajit-pratap the API break would only be negative if a user was relying on the failure in this case, correct? If so, that should be little to no exposure, and the benefit should outweigh it.

aparajit-pratap:
> @aparajit-pratap the API break would only be negative if a user was relying on the failure in this case, correct? If so, that should be little to no exposure, and the benefit should outweigh it.

@Amoursol sorry, I wasn't clear with my question. I think there might be other consequences of the API break such as user code relying on this property to be broken as well. However, what I wanted from you was, if you were okay with the idea of introducing a new node with a different name than `Values` that we could expose while deprecating `Values` if we decided we should avoid the API break at all costs.

aparajit-pratap:
> LGTM!

@sm6srw any idea how severe is the API break if a new attribute is added to an API, in this case, the public property? Will user code need to be rebuilt?

sm6srw:
> 
Looks like it is API breaking according to https://stackoverflow.com/questions/1456785/a-definitive-guide-to-api-breaking-changes-in-net. I assume a recompile is needed? This is not something I have run into.

src/Engine/ProtoCore/FFI/CLRDLLModule.cs
```diff
@@ -634,18 +634,6 @@ private static bool SupressesImport(MemberInfo member, Dictionary<MethodInfo, At
...
-        {
-            object[] atts = method.GetCustomAttributes(false);
-            foreach (var item in atts)
-            {
-                if (item is AllowRankReductionAttribute)
-                    return true;
-            }
-
-            return false;
-        }
```
aparajit-pratap, line 647: unused

aparajit-pratap:
@Amoursol @sm6srw I have avoided the behaviour change by removing the `AllowRankReduction` attribute. Instead, I applied the `ArbitraryDimensionArrayImport` attribute to the return value of `Dictionary.Values` that fixes the issue as well as does not change the original behaviour, aka, a list of a single value still returns as a list of a single value.

Amoursol:
> @Amoursol @sm6srw I have avoided the behaviour change by removing the `AllowRankReduction` attribute. Instead, I applied the `ArbitraryDimensionArrayImport` attribute to the return value of `Dictionary.Values` that fixes the issue as well as does not change the original behaviour, aka, a list of a single value is still returned as a list of a single value.

Brilliant, thanks @aparajit-pratap! This should safely allow us to make this change.

src/Engine/ProtoCore/FFI/CLRDLLModule.cs
```diff
@@ -634,18 +634,6 @@ private static bool SupressesImport(MemberInfo member, Dictionary<MethodInfo, At
             return false;
         }
 
-        private bool AllowsRankReduction(MethodInfo method)
```
sm6srw, line 637: I am a little confused about that this method now can be deleted. Was it not used before?

aparajit-pratap: Yes, this method is unused so I'm deleting it.

src/Libraries/DesignScriptBuiltin/Dictionary.cs
```diff
@@ -59,9 +59,11 @@ private Dictionary(ImmutableDictionary<string, object> dict)
             ///     Produces the values in a Dictionary.
             /// </summary>
             /// <returns name="values">Values of the dictionary</returns>
-            // [AllowRankReduction]
-            // TODO: Consider adding this attribute in 3.0 (DYN-1697)
-            public IEnumerable<object> Values => D.Values;
+            public IEnumerable<object> Values
+            {
+                [return: ArbitraryDimensionArrayImport]
```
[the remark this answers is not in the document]
aparajit-pratap: `ArbitraryDimensionArrayImport` and `AllowRankReduction` attributes are slightly different in behaviour. The first one can be used for function parameters and return values and allows arbitrary ranked arrays and mixed rank arrays from being passed and returned to and from functions. The latter attribute can only be applied to _return_ value of a function. It returns a single value in the case that a function returning a list returns a list of a single value (reduces the rank of a return value).

The thing that both attributes are similar in is that both of them convert the value to arbitrary rank, if it is an array (and if it's an array containing more than one value in the case of `AllowRankReduction`).

In this case, we don't want to change the behaviour of `Dictionary.Values` as far as possible, so we don't want to use `AllowRankReduction` but use `ArbitraryArrayDimensionImport` attribute instead so we can support returning arbitrarily ranked arrays.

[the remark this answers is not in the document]
aparajit-pratap: `ArbitraryArrayDimensionImport` in the past even though it could be applied to function return values and function parameters, we didn't add support to make it handle return values. So, in this PR, I've added that support as well as applied it to `Dictionary.Values`.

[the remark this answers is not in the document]
sm6srw: Sounds good. And this approach will not break the API?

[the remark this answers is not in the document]
mjkkirschner: I think this is fine, can you name the syntax though, my googling is coming up short - `[return: ArbitraryDimensionArrayImport]`

src/Engine/ProtoCore/FFI/CLRDLLModule.cs
```diff
@@ -1342,13 +1331,22 @@ public FFIMethodAttributes(MethodBase method, Dictionary<MethodInfo, Attribute[]
...
             {
-                attributes = method.GetCustomAttributes(false).Cast<Attribute>().ToArray();
+                attributes.AddRange(method.GetCustomAttributes(false).Cast<Attribute>());
+            }
+            if(method is MethodInfo minfo)
+            {
+                var returnTypeAtts = minfo.ReturnTypeCustomAttributes;
+                if (returnTypeAtts != null)
+                {
+                    attributes.AddRange(returnTypeAtts.GetCustomAttributes(false).Cast<Attribute>());
```
mjkkirschner, line 1348: so this seems to be compressing a bit of information - we seem to take the return type attribute and then assign it to a list of attributes for the entire node (method) - whats the purpose of making it an attribute of the `return` value if we just add it to the same list of attributes on the `FFIMethodAttributes` ?

mjkkirschner:
@aparajit-pratap it looks good to me, I'm a bit confused by the purpose of using the `return` qualifier when assigning the attribute since it appears we just lump it together with other attributes, but I don't think it hurts anything unless we might want to support that same attribute on the whole method in some cases where it does something else?

src/Engine/ProtoCore/FFI/CLRDLLModule.cs
aparajit-pratap: I guess we could make it an attribute on the entire method as we do for `AllowRankReduction`. Currently, the `ArbitraryDimensionArrayImport` attribute's usage is for parameters and _return_ values, and if we were to change, it, it would need to be applied to methods and properties as we do for `AllowRankReduction`, not sure if that is a bigger change and/or API break. I believe with the return qualifier it also makes it clearer that it's applied to the return value and not to something else on the method.

PR #13073, opened 2022-06-29
DYN-5018 Collapsed group moving fix

filipeotero:
### Purpose

This PR is to fix the bug when the user moves a collapsed group horizontally.
The issue was occurring because of an unusual change in the width value. To fix that was created a variable to keep the previous value of the group's width when collapsed.

wirebug

### Declarations

### Reviewers

@QilongTang @pinzart

### DIFF
--- a/src/DynamoCore/Graph/Annotations/AnnotationModel.cs
+++ b/src/DynamoCore/Graph/Annotations/AnnotationModel.cs
@@ -23,6 +23,8 @@ namespace Dynamo.Graph.Annotations
         private const double ExtendSize = 10.0;
         private const double ExtendYHeight = 5.0;
 
+        double lastExpandedWidth = 0;
+
         #region Properties
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


QilongTang:
Is there a Jira task linked or just a bug you found?

src/DynamoCore/Graph/Annotations/AnnotationModel.cs
```diff
@@ -498,9 +500,19 @@ internal void UpdateBoundaryFromSelection()
...
+
+                if (IsExpanded)
+                {
+                    Width = Math.Max(region.Width, TextMaxWidth + ExtendSize);                    
+                    lastExpandedWidth = Width;
+                }
+                else
+                {
+                    //If the annotation is not expanded, than it will remain the same width of the last time it was expanded
+                    Width = lastExpandedWidth;
```
QilongTang, line 514: Would there be any edge case that the new width should not be set to last expanded width?

src/DynamoCore/Graph/Annotations/AnnotationModel.cs
```diff
@@ -498,9 +500,19 @@ internal void UpdateBoundaryFromSelection()
...
+                Height = this.ModelAreaHeight + TextBlockHeight;
+
+                if (IsExpanded)
+                {
+                    Width = Math.Max(region.Width, TextMaxWidth + ExtendSize);                    
+                    lastExpandedWidth = Width;
+                }
+                else
+                {
+                    //If the annotation is not expanded, than it will remain the same width of the last time it was expanded
```
zeusongit, line 513: ~~than~~ -> then
`expanded, then it will `

src/DynamoCore/Graph/Annotations/AnnotationModel.cs
QilongTang: Thinking about it, maybe no edge case since we currently disable adding new elements into group when it is collapsed

QilongTang (approved):
LGTM once comments addressed

src/DynamoCore/Graph/Annotations/AnnotationModel.cs
filipeotero: That's right, the width will be the same when the group is collapsed.

src/DynamoCore/Graph/Annotations/AnnotationModel.cs
filipeotero: whops, thanks!

PR #13078, opened 2022-06-30
Workspace References refactoring

reddyashish:
### Purpose

This PR is to refactor some of the code related to workspace references view extension. 

Task: https://jira.autodesk.com/browse/DYN-5028

In the current workflow, the workspace references are computed when a workspace is saved(to serialize the new data) and when the data needs to be updated on the view extension. To avoid doing it twice, a new flag is introduced that would control the workspace references computation. After this change, if the user clicks on "recalculate" button, the references would be computed only if the workspace has unsaved changes. On the other hand, if the user saves the workspace, the references are computed during serialization, saved in a list and that list is displayed in the view extension. 
When a workspace is opened, we want to compute the references even though there are no saved changes. This can done by using the 'ForceComputeWorkspaceReferences' flag. The same is the case when the view extension is opened using the menu item or when a new package is loaded after downloading the missing package references.

Also cleaned up some of the related code.

### Declarations

### Release Notes
Workspace References refactoring

### Reviewers
@QilongTang @zeusongit @aparajit-pratap @mjkkirschner

### DIFF
--- a/src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
+++ b/src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
@@ -935,6 +935,9 @@ namespace Dynamo.Graph.Workspaces
             return externalFiles.Values.ToList<INodeLibraryDependencyInfo>();
         }
 
+        /// <summary>
+        /// This flag will indicate if the workspace references should be computed again.
+        /// </summary>
         internal bool ForceComputeWorkspaceReferences
         {
             get { return forceComputeWorkspaceReferences; }
--- a/src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
+++ b/src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
@@ -238,8 +238,16 @@ namespace Dynamo.Graph.Workspaces
         private bool hasNodeInSyncWithDefinition;
         protected Guid guid;
         private HashSet<Guid> dependencies = new HashSet<Guid>();
-
         private int delayGraphExecutionCounter = 0;
+
+        // For workspace references view extension.
+        private bool forceComputeWorkspaceReferences;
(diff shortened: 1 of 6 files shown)


src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -283,7 +287,6 @@ private void DownloadPackage(object sender, RoutedEventArgs e)
         internal void DownloadSpecifiedPackageAndRefresh(PackageDependencyInfo info)
         {
             packageInstaller.DownloadAndInstallPackage(info);
-            DependencyRegen(currentWorkspace);
```
reddyashish, line 286: This is not needed as the DependencyRegen is being called after the package is loaded

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -876,10 +935,14 @@ private List<INodeLibraryDependencyInfo> GetExternalFiles()
             return externalFiles.Values.ToList<INodeLibraryDependencyInfo>();
         }
 
-        private Dictionary<Guid, PackageInfo> nodePackageDictionary = new Dictionary<Guid, PackageInfo>();
-        private Dictionary<Guid, DependencyInfo> localDefinitionsDictionary = new Dictionary<Guid, DependencyInfo>();
-        private Dictionary<Guid, DependencyInfo> externalFilesDictionary = new Dictionary<Guid, DependencyInfo>();
-
+        internal bool ForceComputeWorkspaceReferences
```
QilongTang, line 938: Would you add some comments to this?

QilongTang:
Thanks for the PR clarification, good stuff

QilongTang (approved):

zeusongit (approved):
LGTM

PR #13092, opened 2022-07-08
Emit arbitrary arrays

aparajit-pratap:
### Purpose

This PR adds support for the emission of DS arrays as MSIL to convert them to equivalent C# arrays. Where possible for constant arrays (arrays containing primitive values), it emits typed arrays. For all other cases `object[]` is emitted, for example, for most mixed types and for combinations of single values and nested arrays or for nested arrays with differing ranks in an outer array.

1. If arrays contain a mix of `int `and `double`, a `double` array is emitted
2. `FunctionLookup` method now returns a collection of `MethodBase` objects in order to support overloaded methods that have the same class and method names and the same number of parameters. The final function endpoint among the list of `methodbase `objects is expected to be determined at runtime (future work).
3. Add switch to select engine from test framework as well as from command line arg
4. Added function `EmitCoercionCode` to coerce constant arguments at compile time to fit to match function parameter types they are passed to. This can be built upon to support more coercion cases.

Some work is still WIP as noted below. I will send a follow-up PR for that. This PR is so that I can unblock the team for subsequent tasks depending on this. 

WIP:
1. Compile-time coercion of types passed to function calls. - this has been done for one case, i.e. conversion of double to long
2. Coercion of `char `to `string` in the case of mixed `char` and `string` arrays - leaving for future work

### Declarations

### Release Notes

### Reviewers

### FYIs


src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -22,13 +22,13 @@ public class CodeGenIL
...
         private StreamWriter writer;
-        private Dictionary<int, MethodBase> methodCache = new Dictionary<int, MethodBase>();
+        private Dictionary<int, IEnumerable<MethodBase>> methodCache = new Dictionary<int, IEnumerable<MethodBase>>();
         private CompilePass compilePass;
 
         private enum CompilePass
         {
-            GlobalFuncSig,
-            GlobalFuncBody,
+            MethodLookup,
```
mjkkirschner, line 30: 👍

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -37,6 +36,7 @@ public static IList ReplicationLogic(MethodBase mi, IList args, string[][] repli
             //Turn the replication guides into a guide -> List args data structure
             var partialInstructions = Replicator.BuildPartialReplicationInstructions(partialReplicationGuides);
 
+            var mi = mInfos.FirstOrDefault();
```
mjkkirschner, line 39: whats this about? Is this the followup you referred to?

aparajit-pratap: Yes.

mjkkirschner (approved):

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -152,6 +154,87 @@ private MethodBase FunctionLookup(IList args)
...
+        {
+            var keys = arrayTypes.Keys;
+            if (keys.Count == 1)
+            {
+                // There is only a single type in the array, return it.
+                return keys.FirstOrDefault();
+            }
+            if(keys.Count == 2)
+            {
+                if(keys.Any(x => x == typeof(long)) && keys.Any(x => x == typeof(double)))
```
pinzart90, line 226: all of these `.Any` queries will actually iterate through the `keys` array.
Do you think it may be faster to iterate once and check for all these expected types ?
i,e 
bool isLong, isDouble, isChar, isString = false
foreach(Type key in keys) {
   isLong |= key == typeof(long)
   ....
}
Not sure this is going to matter much...

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -152,6 +154,87 @@ private MethodBase FunctionLookup(IList args)
             return mi;
         }
 
+        private static Dictionary<Type, int> GetTypeStatisticsForArray(ExprListNode array)
+        {
+            var usageFreq = new Dictionary<Type, int>();
```
pinzart90, line 159: is the frequency information actually used anywhere ?
Dictionary<Type, int> - the integer

aparajit-pratap: As of now, I think it's just the number of keys in the dictionary that is important.

aparajit-pratap: Are you suggesting I use a hashset or something instead?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -152,6 +154,88 @@ private MethodBase FunctionLookup(IList args)
...
+                // There is only a single type in the array, return it.
+                return arrayTypes.FirstOrDefault();
+            }
+            if(arrayTypes.Count == 2)
+            {
+                bool isLong = false;
+                bool isDouble = false;
+                bool isChar = false;
+                bool isString = false;
+                foreach (var type in arrayTypes)
```
pinzart90, line 220: if the length of the array is only 2...I guess my suggestion will not improve performance at all.

pinzart90: I initially though we were iterating the entire array (the items not the types).

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -152,6 +154,88 @@ private MethodBase FunctionLookup(IList args)
...
+                if (exp is ExprListNode eln)
+                {
+                    var subArray = GetTypeStatisticsForArray(eln);
+                    var t = GetOverallTypeForArray(subArray);
+
+                    if (t != typeof(object))
+                    {
+                        t = t.MakeArrayType();
+                    }
+                    arrayTypes.Add(t);
```
pinzart90, line 172: why not make the array type in the case of object too ?
type would be : `object[]`

aparajit-pratap: I'm not sure why I added this condition. Looks like I can safely make it `object[]` as well. Will try that.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -152,6 +154,88 @@ private MethodBase FunctionLookup(IList args)
...
+                bool isChar = false;
+                bool isString = false;
+                foreach (var type in arrayTypes)
+                {
+                    isLong = isLong || type == typeof(long);
+                    isDouble = isDouble || type == typeof(double);
+                    isChar = isChar || type == typeof(char);
+                    isString = isString || type == typeof(string);
+                }
+                if (isLong && isDouble)
```
pinzart90, line 227: I think these conditions will only work because the arrayTypes length is 2.
If it were more than 2, I think the conditions would need to change....to check if the arrayTypes is made up of only (long or double) or only (char or str).


Do you agree ?

aparajit-pratap: Yes. So are you saying, I should change it so that it works in the more general cases?

aparajit-pratap: I think I'm going to leave it as-is for now.

src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -1682,17 +1692,15 @@ private void SynchronizeInternal(GraphSyncData syncData)
                 // Get AST list that need to be executed
                 var finalDeltaAstList = changeSetComputer.GetDeltaASTList(syncData);
 
+                if (!DSExecutionEngine)
+                {
+                    CompileAndExecuteMSIL(finalDeltaAstList);
```
sm6srw, line 1697: ✔️

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -422,28 +506,59 @@ private Type EmitIdentifierListNode(AssociativeNode node)
...
             {
                 EmitOpCode(OpCodes.Dup);
                 EmitOpCode(OpCodes.Ldc_I4, ++elCount);
                 var t = DfsTraverse(el);
-                //if (t == typeof(int) || t == typeof(long) || t == typeof(double) || t == typeof(bool) || t == typeof(char))
-                //{
-                //    EmitOpCode(OpCodes.Box, t);
-                //}
-                EmitOpCode(OpCodes.Stelem_R8);
+                if (ot == typeof(object) && (t == typeof(int) || t == typeof(long) || t == typeof(double) || t == typeof(bool) || t == typeof(char)))
```
pinzart90, line 528: So is the idea here to check for non-array types ?

If that is the case...could we not use the `IsArray` property of the `Type` class ?

https://docs.microsoft.com/en-us/dotnet/api/system.type.isarray?redirectedfrom=MSDN&view=net-6.0#System_Type_IsArray

test/DynamoCoreTests/Performance/PerformanceTests.cs
```diff
@@ -65,7 +65,7 @@ private void DoWorkspaceOpenAndCompare(string filePath)
 
             Setup();
 
-            OpenModel(filePath);
+            OpenModel(filePath, dsExecution: false);
```
sm6srw, line 68: 👍

src/Engine/EmitMSIL/CodeGenIL.cs
aparajit-pratap: The idea here is to check for primitive (value) types to see if they need to be boxed to convert to reference type.

src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -1682,17 +1692,15 @@ private void SynchronizeInternal(GraphSyncData syncData)
...
                 changeSetApplier.Apply(runnerCore, runtimeCore, changeSetComputer.csData);
 
-                ///////////////////////////////////////////////////
-                Dictionary<string, IList> input = new Dictionary<string, IList>();
-                var codeGenIL = new EmitMSIL.CodeGenIL(input, @"D:\GitHub\Dynamo\src\Engine\EmitMSIL\bin\Debug\opCodes.txt");
-                codeGenIL.Emit(finalDeltaAstList);
-
-                ///////////////////////////////////////////////////
-
                 if (finalDeltaAstList.Any() || anyForcedExecutedNodes || changeSetComputer.csData.ModifiedNodesForRuntimeSetValue.Any())
```
sm6srw, line 1704: This is still confusing to me.  So when DSExecutionEngine is true, this condition will in most cases be false? At least now in the beginning?

sm6srw (approved):
Great, this looks good to me.

src/Engine/EmitMSIL/CodeGenIL.cs
pinzart90: should be fine as is now. Maybe just a comment ...if someone else starts adding more code in there.

src/Engine/EmitMSIL/CodeGenIL.cs
pinzart90: I think there is a `IsValueType` property on Types

pinzart90 (approved):
LGTM

PR #13100, opened 2022-07-11
WIP emit range expressions

mjkkirschner:
### Purpose

This PR adds partial support for generating/evaluating range expressions in the MSIL based engine by doing the following:

1. adds some plain c# range generation methods to protocore in builtinfunctionendpoint.cs - the existing ones use `StackValue` which is a DS VM wrapper type.
2. starts refactoring the existing range methods to call the new ones - this could be useful if we plan to have 2 vms for a while, in that case it would be nice if we only had one implementation.
3. adds an entry point to call these new methods from DSCoreNodes - 😢 - I prefer to put them in `DesignScriptBuiltins` but had some cyclic dependency issues, they can be resolved, but it was lower priority as long as we keep everything internal.
4. When an `RangeExpressionASTNode` is encountered we generate a function call node to the selected static entry point `GenerateRangeILInt` or `double` etc mentioned above.

miscellaneous changes

* adds a new method `CodeGenIL` - `EmitAndExecute` which is used for testing - it can probably be combined with the existing `Emit` - I wanted to check with @aparajit-pratap before modifying that method and its the entry point into the engine today.
* Adds some simple CodeGenMSIL range tests - to accomplish this I had emit nothing for `importASTNodes` - but this works as the actual import is done in the parser... 😬 (I can see why that might be necessary) - but it seems fishy.
* sets opcodes.txt to relative path.
* implement dispose on `CodeGen`

### Current issues and open questions:
- [ ] ~~alphabet ranges are not supported~~ deferring this - see below.
- [ ] ~~refactor existing range methods to call new plain c# ones and convert to/from stack values - this will let us use the existing tests.~~ because we have deferred implementing the dynamic logic that was previously performed with `stackValue` we cannot easily switch the old vm to use the new methods - we can do this later when we implement alphabet ranges, I will file a followup.
- [ ] ~~the current range implementation returns object[] and has to box the ints/doubles/chars - I think the only way around this to to return some kind of value type array using generics, a stackValue union or base type, or split all the range methods into overloads for int/double/char - but I think for this to work the VM would need to have overloads working and we'd have 3 entry points with the correct return types - I also think we would not be able to avoid all boxing.~~ For the time being we decided to use type information at compile time to select a particular range generation method - if we don't have type information than we always generate a double array. When we add alphabetic ranges it seems we'll need to implement a boxing dynamic range method. For now it simpler to coerce double[] and long[] array types rather than object[] anyway, especially since the intention for this task was to make testing other nodes/replication easier.
- [x] I think DSCore nodes is the wrong place for the generate range entry point.
- [x] needs more tests

### Declarations

### Release Notes

### Reviewers

### FYIs


src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -700,11 +743,63 @@ private void EmitLanguageBlockNode(AssociativeNode node)
...
+                toNode,
+                stepNode ?? new NullNode(),
+                op,
+                AstFactory.BuildBooleanNode(stepNode != null),
+                AstFactory.BuildBooleanNode(hasAmountOperator),
+            };
+            //TODO shortest lacing?
+            //TODO without more refactoring we cannot reference the names of the generate range functions. fix that.
+            //maybe move to designscriptbuiltins and use type forwarding for enum?
+            var rangeExprFunc = AstFactory.BuildFunctionCall("GenerateRangeIL", arguments);
```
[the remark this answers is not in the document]
mjkkirschner: yes, we discussed in the meeting and this was one of the options, I will send you the recording link - but this only works when the AST actually defines for us what the types are - in many cases the AST nodes will be identifiers with no type information.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -700,11 +743,63 @@ private void EmitLanguageBlockNode(AssociativeNode node)
...
+            var idlist = new IdentifierListNode()
+            {
+                LeftNode = new IdentifierNode("DSCore.RangeHelpers"),
+                RightNode = rangeExprFunc
+            };
+            //we want to cache the call to generate range, so traverse down in any case.
+            var t = DfsTraverse(idlist);
```
pinzart90, line 794: You could also create a FunctionCallNode and call EmitFunctionCall directly right ?
```
            var fnNode = new FunctionCallNode()
            {
                FormalArguments = arguments,
                Function = rangeExprFunc
            };

            EmitFunctionCallNode(fnNode);
```
Or is there something useful going on in creating an IdentifierListNode...?

mjkkirschner:
> @mjkkirschner What happens today if array is connected to input of another nodes? Not supported yet correct?

I just added a test for this - and yes you get a object of type[] cannot be converted to IEnumerable<double> - it doesn't matter if the underlying types are `double` or `int` - some type coercion is needed I assume.

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: `className` is only set during identifierListNode creation.

aparajit-pratap:
> @mjkkirschner What happens today if array is connected to input of another nodes? Not supported yet correct?

This is now supported - at least the creation of the arrays is. See https://github.com/DynamoDS/Dynamo/pull/13092

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -90,6 +91,45 @@ public void Emit(List<AssociativeNode> astList)
...
+        {
+            //TODO refactor with above method.
+            compilePass = CompilePass.MethodLookup;
+            foreach (var ast in astList)
+            {
+                DfsTraverse(ast);
+            }
+
+            // 1. Create assembly builder (dynamic assembly)
+            var asm = BuilderHelper.CreateAssemblyBuilder("DynamicAssembly", false);
```
aparajit-pratap, line 104: I guess we could add another option here to `CreateAssemblyBuilder` to only create an assembly with `AssemblyBuilderAccess.Run` or `AssemblyBuilderAccess.RunAndCollect` option in this case. See https://github.com/DynamoDS/Dynamo/blob/4f5610ff713d90a842c691429648ef98ab3477c0/src/Engine/EmitMSIL/BuilderHelper.cs#L40.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -146,7 +186,7 @@ private IEnumerable<MethodBase> FunctionLookup(IList args)
 
                     // TODO: Add check for constructorinfo objects
 
-                    if (mi != null)
+                    if (mi != null && mi.Count()>0)
```
aparajit-pratap, line 189: `mi.Any()`

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -159,7 +199,7 @@ private IEnumerable<MethodBase> FunctionLookup(IList args)
                     //}
                 }
             }
-            if (mi == null)
+            if (mi != null && mi.Count() > 0 == null)
```
aparajit-pratap, line 202: Do you mean `if( mi == null || (mi != null && mi.Count() > 0) )`?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -416,10 +456,13 @@ private void EmitDynamicBlockNode(int block)
...
         }
-
+        
         private void EmitImportNode(AssociativeNode node)
         {
-            throw new NotImplementedException();
+            //doing absolutely nothing is actually
+            //enough to import binaries.
+            //TODO do other important things here!
+            //see: ProtoAssociative.CodeGen.EmitImportNode
```
aparajit-pratap, line 465: @mjkkirschner do you think it could be useful to implement this for a separate, initial compile pass, say `CompilePass.ImportAssemblies` to prepopulate class and function tables like we do today (see `CLRModuleType.ParseSystemType`)? We can then use this information to our benefit during code emission and/or at runtime.

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: thanks, I think this was a copy paste error

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -509,10 +552,19 @@ private Type EmitIdentifierListNode(AssociativeNode node)
...
+                className = iln.LeftNode.ToString();
+            }
+            else
+            {
+                var ident = iln.LeftNode as IdentifierNode;
+                if (ident == null) throw new ArgumentException("Left node of IdentifierListNode is expected to be an identifier.");
 
-            className = ident.Value;
+                className = ident.Value;
+            }
```
aparajit-pratap, line 567: I think this entire logic seems flaky. We should be able to use this existing utility method here, `CoreUtils.GetIdentifierExceptMethodName(IdentifierListNode identList)`. Can you try that?

aparajit-pratap: https://github.com/DynamoDS/Dynamo/blob/e5e6d90511ab8ea84bd45d193071c3006e96c1cb/src/Engine/ProtoCore/Utils/CoreUtils.cs#L518

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -700,11 +752,63 @@ private void EmitLanguageBlockNode(AssociativeNode node)
...
+            var idlist = new IdentifierListNode()
+            {
+                LeftNode = new IdentifierNode("DSCore.RangeHelpers"),
+                RightNode = rangeExprFunc
+            };
+            //we want to cache the call to generate range, so traverse down in any case.
+            var t = DfsTraverse(idlist);
+            if (compilePass == CompilePass.MethodLookup)
+            {
+                return null;
```
aparajit-pratap, line 806: I suppose this is not required as we're returning in the next line anyway.

src/Engine/ProtoCore/Lang/BuiltInFunctionEndPoint.cs
```diff
@@ -1032,29 +1032,98 @@ internal static double Map(double rangeMin, double rangeMax, double inputValue)
...
+                return GenerateRangeByStepNumberInt(start, end, stepnum).Select(x => StackValue.BuildInt(x)).ToArray();
+            }
+            else
+            {
+                return GenerateRangeByStepNumberDouble(start, end, stepnum).Select(x => StackValue.BuildDouble(x)).ToArray();
+            }
+        }
+
+      
+        internal static long[] RangeExpressionInt(long start, long end, double step,int op,bool hasStep, bool hasAmountOp )
```
aparajit-pratap, line 1097: If the step is `double`, e.g. `1..5..1.5`, won't the generated array be a double array?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -700,11 +752,63 @@ private void EmitLanguageBlockNode(AssociativeNode node)
...
+                toNode,
+                stepNode ?? new NullNode(),
+                op,
+                AstFactory.BuildBooleanNode(stepNode != null),
+                AstFactory.BuildBooleanNode(hasAmountOperator),
+            };
+            //TODO shortest lacing?
+            //TODO without more refactoring we cannot reference the names of the generate range functions. fix that.
+            //maybe move to designscriptbuiltins and use type forwarding for enum?
+            var rangeExprFunc = AstFactory.BuildFunctionCall("GenerateRangeIL", arguments);
```
aparajit-pratap, line 796: Where is the `GenerateRangeIL` method defined? I only see a `GenerateRangeILInt`, and a `GenerateRangeILDouble`.

src/Engine/ProtoCore/Lang/BuiltInFunctionEndPoint.cs
mjkkirschner: not in all cases - there is a specific existing case for the form start..#amount..step - where step can be a double, but if the value can be represented by an integer than the range is an int array.

mjkkirschner: yes, for now I will implement a fallback boxing case.

src/Engine/EmitMSIL/CodeGenIL.cs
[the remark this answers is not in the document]
mjkkirschner: I've taken this approach, PTAL - I have yet to implement the fallback boxing method, but it will basically be the first version of this code that I wrote using objects.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -706,11 +747,160 @@ private void EmitLanguageBlockNode(AssociativeNode node)
...
+
+            var range = node as RangeExprNode;
+            var fromNode = range.From;
+            var toNode = range.To;
+            var stepNode = range.Step;
+            var stepOp = range.StepOperator;
+            var hasAmountOperator = range.HasRangeAmountOperator;
+
+            //TODO we may want to do this check again at runtime.
+            if(stepNode is DoubleNode && stepOp == DSASM.RangeStepOperator.Number)
```
aparajit-pratap, line 802: Does this mean we cannot do something like `1..10..#2.3;`?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -706,11 +747,160 @@ private void EmitLanguageBlockNode(AssociativeNode node)
...
+            {
+                throw new ArgumentException(Resources.kInvalidAmountInRangeExpression);
+            }
+
+
+            var methodName = unselectedToken;
+
+            var isIntStep = stepNode is IntNode stpInt ||
+                (hasAmountOperator && stepOp == DSASM.RangeStepOperator.StepSize && stepNode is DoubleNode stpDB && Math.Truncate(stpDB.Value) == stpDB.Value) ||
```
aparajit-pratap, line 811: What exactly is the purpose of `hasAmountOperator`?

aparajit-pratap: Just so I understand - if we make the condition before the `||` this: `stepNode is intNode && hasAmountOperator && stepOp == DSASM.RangeStepOperator.StepSize`, and the from and to nodes are also integers, then it will always mean generate an integer range, correct?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -706,11 +747,160 @@ private void EmitLanguageBlockNode(AssociativeNode node)
...
+                        methodName = doubleRangeMethodName;
+                    }
+                }
+            }
+            if (methodName == unselectedToken)
+            {
+                //if we still have not been able to determine the type of range to generate, temporarily generate a double range.
+                //TODO when we add alphabetic ranges we'll want to check for string/char types in the above logic but,
+                //if we still get to this line we'll need to call a dynamic version of generate range that boxes objects.
+                methodName = doubleRangeMethodName;
```
aparajit-pratap, line 859: ✔️

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: refactored this a bit.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -52,6 +55,33 @@ public CodeGenIL(IDictionary<string, IList> input, string filePath)
         }
 
         public IDictionary<string, IList> Emit(List<AssociativeNode> astList)
+        {
+            var compileResult = CompileAstToDynamicType(astList, AssemblyBuilderAccess.RunAndCollect);
```
aparajit-pratap, line 59: This should be `RunAndSave`.

mjkkirschner: thanks.

src/Engine/EmitMSIL/CodeGenIL.cs
aparajit-pratap: Yes.

src/Libraries/DesignScriptBuiltin/RangeHelpers.cs
```diff
@@ -0,0 +1,378 @@
...
+
+
+        internal static long[] RangeExpressionInt(long start, long end, double step, int op, bool hasStep, bool hasAmountOp)
+        {
+            if (hasAmountOp && !hasStep)
+            {
+                throw new ArgumentException(Resources.kNoStepSizeInAmountRangeExpression);
+            }
+
+            return GenerateNumericRangeCoreInt(start, end, step, op, hasStep, hasAmountOp);
```
aparajit-pratap, line 69: Why not move this code up to the public method? That way we can reduce a few methods, in this case, `RangeExpressionInt` and `RangeExpressionDouble`.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -706,11 +726,160 @@ private void EmitLanguageBlockNode(AssociativeNode node)
...
+
+            
+            var rangeExprFunc = AstFactory.BuildFunctionCall(methodName, arguments);
+            var idlist = new IdentifierListNode()
+            {
+                LeftNode = new IdentifierNode(typeof(Builtin.RangeHelpers).FullName),
+                RightNode = rangeExprFunc
+            };
+            //we want to cache the call to generate range, so traverse down in any case.
+            var t = DfsTraverse(idlist);
```
aparajit-pratap, line 877: Are the cached generate range methods ever going to be used after retrieving them from the cache?

src/Libraries/DesignScriptBuiltin/RangeHelpers.cs
mjkkirschner: These methods are also used from the existing DS range methods- as I started to refactor there - maybe I can combine the methods you have suggested and call the public ones instead .. unsure.

src/Engine/EmitMSIL/CodeGenIL.cs
aparajit-pratap: Does this mean, that the range methods will be looked up only the first time a range expr is encountered and subsequently they will simply be returned from the cache? I think I got it.

src/Libraries/DesignScriptBuiltin/RangeHelpers.cs
aparajit-pratap: Ah, ok, then this is fine.

aparajit-pratap (approved):

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: yes, sorry I should probably remove this comment, it was from before when the methodlookup phase made it down here.

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: I will file a followup on this.

PR #13102, opened 2022-07-12
Pin Wire from Right-click Context Menu

dnenov:
### Purpose

An additional feature to the context menu of the connection wire. When Right-clicked, now will display the option to Pin wire 

pin right-click context

### Declarations

### Release Notes

- added right-click context menu to allow Pin wire
- the wire will be pinned at the mouse point, similar to the existing mouse hoover behavior

### Reviewers

@reddyashish 

### FYIs

@Amoursol

### DIFF
--- a/src/DynamoCoreWpf/Properties/Resources.Designer.cs
+++ b/src/DynamoCoreWpf/Properties/Resources.Designer.cs
@@ -466,6 +466,15 @@ namespace Dynamo.Wpf.Properties {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to Pin Wire.
+        /// </summary>
+        public static string ConnectorContextMenuHeaderPinConnector {
+            get {
+                return ResourceManager.GetString("ConnectorContextMenuHeaderPinConnector", resourceCulture);
+            }
+        }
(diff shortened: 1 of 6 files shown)


reddyashish (approved):

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
```diff
@@ -1518,6 +1518,53 @@ private static bool CanDoGraphAutoLayout(object o)
...
+        /// <param name="isHidden"></param>
+        private void ShowHideAllWires(List<NodeModel> nodeModels, bool isHidden)
+        {
+            if (!nodeModels.Any()) return;
+
+            foreach (var nodeModel in nodeModels)
+            {
+                var connectors = nodeModel.AllConnectors;
+                var enumerator = connectors.GetEnumerator();
+                while (enumerator.MoveNext())
```
reddyashish, line 1559: I was thinking may be using an enumerator for connectors list could be slow and could affect graphs with lot of connectors. Can we just iterate the list normally?

PR #13111, opened 2022-07-14
Update PackageManagerClient.cs

zeusongit:
### Purpose

https://jira.autodesk.com/browse/DYN-5103

When a user right clicks a package in Dynamo package manager preference tab, we check if the user is an author of that package or not in order to enable/disable Deprecate/Undeprecate option for that package.

Created a cache in Dynamo to record the result so that we do not call the API again and again for one session.

### Declarations

### Release Notes

Created a cache in Dynamo to record the result so that we do not call the API again and again for one session.

### Reviewers

@QilongTang

### DIFF
--- a/src/DynamoPackages/PackageManagerClient.cs
+++ b/src/DynamoPackages/PackageManagerClient.cs
@@ -31,6 +31,11 @@ namespace Dynamo.PackageManager
         ///     The directory where new packages are created during the upload process.
         /// </summary>
         private readonly string packageUploadDirectory;
+
+        /// <summary>
+        ///     The dictionay stores the package name corresponding to the boolean result of whether the user is an author of that package or not.
+        /// </summary>
+        private Dictionary<string, bool> packageMaintainers;
        
         /// <summary>
         ///     The URL of the package manager website
@@ -47,6 +52,7 @@ namespace Dynamo.PackageManager
             this.packageUploadDirectory = packageUploadDirectory;
             this.uploadBuilder = builder;
             this.client = client;
+            this.packageMaintainers = new Dictionary<string, bool>();
         }
 
         internal bool Upvote(string packageId)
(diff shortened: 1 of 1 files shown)


mjkkirschner:
do we check this again / refresh if the user logs out / logs in as another user?

zeusongit:
> do we check this again / refresh if the user logs out / logs in as another user?

Not tested that scenario in revit, but it will break in that case, as Dictionary will have stale values. 
Can user remain in Dynamo and logout? Because if it needs a restart then this could work. 
I will test this and update.

src/DynamoPackages/PackageManagerClient.cs
```diff
@@ -277,9 +283,15 @@ internal PackageManagerResult Undeprecate(string name)
 
         internal bool DoesCurrentUserOwnPackage(Package package,string username) 
         {
+            bool value;
+            if (this.packageMaintainers.Count > 0 && this.packageMaintainers.TryGetValue(package.Name, out value)) {
```
QilongTang, line 287: If I remember correctly, you can use the following to save the definition? 
```
f (this.packageMaintainers.Count > 0 && this.packageMaintainers.TryGetValue(package.Name, out bool value))
```

src/DynamoPackages/PackageManagerClient.cs
```diff
@@ -277,9 +283,15 @@ internal PackageManagerResult Undeprecate(string name)
...
+                return value;
+            }
             var pkg = new PackageInfo(package.Name, new Version(package.VersionName));
             var mnt = GetPackageMaintainers(pkg);
-            return (mnt != null) && (mnt.maintainers.Any(maintainer => maintainer.username.Equals(username)));
+            value = (mnt != null) && (mnt.maintainers.Any(maintainer => maintainer.username.Equals(username)));
```
QilongTang, line 292: Just curious if `maintainers` could be null because it will crash Dynamo written this way

zeusongit: `mnt.maintainers` cannot be null

src/DynamoPackages/PackageManagerClient.cs
zeusongit: I had to use `value` below at line 293 so I used a variable for it.

zeusongit:
I checked in Revit, we cannot sign out without quitting, not sure about other integrators.

Screen Shot 2022-07-14 at 4 03 39 PM
.

QilongTang (commented):
LGTM does it make sense to add a unit test for this update?

zeusongit:
> LGTM does it make sense to add a unit test for this update?

Unit test is not helpful, as we call the parent method for result, internally it uses cached or API response, with no way of knowing the difference.

QilongTang (approved):

PR #13122, opened 2022-07-18
DYN-5084  Checking if the ports are empty or null

filipeotero:
### Purpose

This Pr includes the improvement for some edge cases where the Inports/Outports are wrongly colored when run at manual mode.

Ex2

Ex1

image

### Declarations

### Reviewers

@QilongTang @Amoursol


QilongTang (approved):

filipeotero:
> I have a question about the second gif in description. When the connector for watch node is removed, shouldn't the port show the red color as it is missing the value.

Hi, thanks for noticing that. This was happening because the watch node was having the same behavior of a function node. 
According to the AC's of the DYN-5084, watch node shouldn't show grey bars under any circumstances.
By that, I've added a check specifically for the watch nodes.

reddyashish:
I agree on the grey bar. I was just not clear what the red color bar on the port indicate? Missing the value or something else?

filipeotero:
> I agree on the grey bar. I was just not clear what the red color bar on the port indicate? Missing the value or something else?

Yes. It indicates when the port has no values.

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -917,7 +917,12 @@ public bool DisplayLabels
         [JsonIgnore]
         public bool AreAllOutputsConnected
         {
-            get { return outPorts.All(p => p.IsConnected); }
+            get { 
+                return 
+                    outPorts != null &&
+                    outPorts.Count > 0 &&
+                    outPorts.All(p => p.IsConnected);
```
reddyashish, line 924: I think the All() checks for null and empty elements because it would return false if all the elements do not satisfy the condition. Did you notice any edge cases?

reddyashish (approved):
LGTM with one comment.

src/DynamoCore/Graph/Nodes/NodeModel.cs
filipeotero: There is an edge case where a code block is created in manual mode and this property returns true even though the outports have zero elements. It assumes that all the outports are connected, but there aren't any.
debugCode

PR #13125, opened 2022-07-19
Replication logic

pinzart90:
This PR implements replication logic (part of it anyway)
Most of the replication functions are copied from the old VM. The new functions are mostly located right under the original ones.

New structures introduced:
- `CLRStackValue` => equivalent to StackValue
   Wraps the CLR objects so that we can store useful metadata (like rank, ProtoCore.Type info etc)
-  `CLRFunctionEndpoint` => equivalent to FunctionEndpoint
   Wraps the MethodBase objects and stores useful metadata (like rank, ProtoCore.Type info etc)
- `MSILRuntimeCore` - equivalent to RuntimeCore 
   Holds type information like ClassTable. Makes it easier to mimic the old VM functions
   
 Marshaler changes:
 I introduced new virtual methods (that can handle CLRStackValues) to the existing Marshaler classes ...so there should be no breaking changes.
 
 The deferred work is marked in code with TODO_MSIL. I will create jira tasks once we finalize the PR 
 TODO : Add task to support int32 data types.


src/Engine/EmitMSIL/Replication.cs
```diff
@@ -133,5 +130,200 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
...
+            return result;
+        }
+
+        // TODO: Look into array conversion performance 
+        private static object[] convertToArray(object o)
+        {
+            System.Type type = o.GetType();
+            if (o is Array oldArr)
+            {
+                Array newArr = Array.CreateInstance(typeof(object), oldArr.Length);
```
mjkkirschner, line 168: why do we need to create a new array if o is already an array?

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -133,5 +130,206 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
...
+            }
+            else if (o is IEnumerable oEnum)
+            {
+                oIEnum = oEnum.Cast<object>();
+            }
+            else
+            {
+                oIEnum = new List<object>() { o };
+            }
+            return oIEnum;
```
pinzart90, line 178: all these cast<object>() calls will do boxing and they will probably copy the items too...
Not sure if there is a better way to do this...

mjkkirschner: what about trying to use `Memory<T>`? I  think it has a constructor from an array and the you can use:
MemoryMarshal.ToEnumerable...
https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.memorymarshal.toenumerable?view=net-6.0

These types are only useful I guess if the `memory` object is really a view onto the array, which I thought I had read it was at some point...

src/Engine/ProtoCore/DSASM/InstructionSet.cs
```diff
@@ -110,6 +110,113 @@ public struct MetaData
...
+        {
+            get;
+            set;
+        }
+
+        public ProtoCore.Type ProtoType { get; set; }
+
+        public System.Type Type => Value?.GetType();
+
+        public object Value { get; set; }
```
aparajit-pratap, line 140: Does this mean we need to box value types?

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
```diff
@@ -1133,6 +1226,31 @@ private StackValue CreateDSObject(object obj, ProtoCore.Runtime.Context context,
...
+                var classTable = runtimeCore.ClassTable;
+                mCachedType = classTable.IndexOf(GetTypeName(objType));
+
+                //Recursively get the base class type if available.
+                while (mCachedType == -1 && objType != null)
+                {
+                    objType = objType.BaseType;
+                    if (null != objType)
+                        mCachedType = classTable.IndexOf(GetTypeName(objType));
+                }
```
aparajit-pratap, line 1247: Why are we assigning the most base type to `mCachedType` here?

src/Engine/ProtoCore/Lang/FunctionGroup.cs
```diff
@@ -177,6 +178,26 @@ public void CopyPublic(List<FunctionEndPoint> rhs)
             return ret;
         }
 
+        internal static Dictionary<CLRFunctionEndPoint, int> GetConversionDistances(IEnumerable<CLRFunctionEndPoint> functionEndPoints,
```
aparajit-pratap, line 181: Are these `functionEndPoints` (plural) the overloaded methods? Can we simplify the replication logic by ignoring method overloads for the current phase of work?

src/Engine/ProtoCore/Lang/FunctionGroup.cs
```diff
@@ -292,18 +307,91 @@ public static bool CheckInvalidArrayCoersion(FunctionEndPoint fep, List<StackVal
                 if (isNotExactTypeMatch && argumentsNotNull && recievingTypeNotAVar && isNotConvertible)// && !isCalleeVar)
                 {
                     return true; //It's an invalid coersion
+                }
+            }
+            return false;
+        }
+        /*
+        internal static bool CheckInvalidArrayCoersion(CLRFunctionEndPoint fep, List<CLRStackValue> reducedSVs, bool allowArrayPromotion)
```
aparajit-pratap, line 315: Why is this commented out?

src/Engine/ProtoCore/Lang/Replication/Replicator.cs
```diff
@@ -337,7 +337,52 @@ public static List<StackValue> EstimateReducedParams(List<StackValue> formalPara
...
         }
+        
+        /// <summary>
+        /// Compute the effects of the replication guides on the formal parameter lists
+        /// The results of this loose data, and will not be correct on jagged arrays of hetrogenius types
+        /// </summary>
+        /// <param name="formalParams"></param>
+        /// <param name="replicationInstructions"></param>
+        /// <returns></returns>
+        internal static List<CLRStackValue> EstimateReducedParams(List<CLRStackValue> formalParams, List<ReplicationInstruction> replicationInstructions)
```
aparajit-pratap, line 348: These "formalParams" are the arguments, correct?

src/Engine/ProtoCore/DSASM/InstructionSet.cs
```diff
@@ -110,6 +110,111 @@ public struct MetaData
         public int type;
     }
 
+    /// <summary>
+    /// Wrapper over all values handled by the MSIL/CLR engine.
+    /// </summary>
+    internal struct CLRStackValue
```
pinzart90, line 116: Simplified version of existing StackValue structure in the old VM.
Reasons for creating/using this new struct:
1. Simplifies the copy/paste process from old VM logic to new one.
2. Replication needs access to particular information that we can store once rather than have it compute every time (ex rank)

src/Engine/ProtoCore/DSASM/InstructionSet.cs
```diff
@@ -110,6 +110,111 @@ public struct MetaData
...
+
+        public bool IsNull => Value == null;
+
+        // Equivalent to StackValue.metaData.type
+        public int TypeUID { get; set; }
+
+        public System.Type Type => Value?.GetType();
+
+        // TODO_MSIL: Figure out how bad boxing/unboxing is for performance
+        public object Value { get; set; }
```
pinzart90, line 139: boxing/unboxing will still be an issue.
Marshaling/Unmarshaling is the place where most of the boxing/unboxing will appen

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -2,17 +2,107 @@
...
+using ProtoCore.Exceptions;
+using ProtoCore.Properties;
+using ProtoCore.DSASM;
+using ProtoCore.Utils;
 
 namespace EmitMSIL
 {
     public class Replication
     {
+        internal static List<CLRStackValue> MarshalFunctionArguments(IList args, MSILRuntimeCore runtimeCore)
```
mjkkirschner, line 17: how would you describe what these marshal methods are doing? What are they marshaling between, since we now have a compiler instead of a VM? Or does this PR bring back a lot of a the machinery of the VM?

Asked another way, why do we need the runtime core, class table etc?

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -2,17 +2,107 @@
...
+                }
+                catch (System.InvalidCastException ex)
+                {
+                    runtimeCore.LogWarning(ProtoCore.Runtime.WarningID.AccessViolation, ex.Message);
+                    return null;
+                }
+                catch (InvalidOperationException)
+                {
+                    string message = String.Format(Resources.kFFIFailedToObtainObject, paramType.Name, formalParams[i].CLRInfo.Member.DeclaringType.Name, formalParams[i].CLRInfo.Member.Name);
+                    runtimeCore.LogWarning(ProtoCore.Runtime.WarningID.AccessViolation, message);
```
mjkkirschner, line 82: do these warnings get surfaced somewhere?

src/Engine/ProtoCore/DSASM/InstructionSet.cs
mjkkirschner: maybe related to:
https://jira.autodesk.com/browse/DYN-5152

@aparajit-pratap noted yesterday that these types are still not returned from `ReplicationLogic` but only used internally to compute function endpoints / replication behavior - is that still the case?

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
```diff
@@ -430,7 +513,114 @@ private ICollection ToICollection(StackValue dsObject, ProtoCore.Runtime.Context
...
+                else if (arrayType.UnderlyingSystemType == typeof(bool[]))
+                {
+                    return UnMarshal<bool>(dsObject, runtimeCore);
+                }
+            }
+
+            var dsArrayValues = dsObject.Value as IList<CLRStackValue>;
+
+            //  use arraylist instead of object[], this allows us to correctly capture
```
mjkkirschner, line 541: comment seems out of date?

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -23,7 +113,13 @@ public class Replication
         /// <returns></returns>
         public static IList ReplicationLogic(IEnumerable<MethodBase> mInfos, IList args, string[][] replicationAttrs)
         {
-            var reducedArgs = ReduceArgs(args);
+            // Static instance of MSILRuntimeCore
+            // TODO_MSIL: FIgure out how and when to set this runtimeCore (CodeGen should have it too)
```
mjkkirschner, line 117: what do you imagine the codegen will use the runtime core for?

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -36,33 +132,35 @@ public static IList ReplicationLogic(IEnumerable<MethodBase> mInfos, IList args,
...
+            List<CLRFunctionEndPoint> resolvedFeps;
+            List<ReplicationInstruction> replicationInstructions;
+            ComputeFeps(reducedArgs, feps, partialInstructions, runtimeCore, out resolvedFeps, out replicationInstructions);
+
+            var finalFep = SelectFinalFep(resolvedFeps, reducedArgs);
 
-            // Testing invoking method without replication
             object result;
-            if (mi.IsStatic)
+            if (replicationInstructions.Count == 0)
```
mjkkirschner, line 144: do you believe it would be difficult to compute this case at compile time? (no replication)
https://jira.autodesk.com/browse/DYN-5151

src/Engine/ProtoCore/Lang/FunctionGroup.cs
pinzart90: Yes, `functionEndPoints` (here) represent overaloaded methods. However, it will only contain a single functionEndpoint because overloaded methods are not analyzed yet.
Since I already took the time to convert these methods to the new VM...I would like to keep them...even though they do not do much for now

src/Engine/ProtoCore/Lang/FunctionGroup.cs
pinzart90: it is new code...I will remove it for now, since I haven't gotten around to it yet

src/Engine/ProtoCore/Lang/Replication/Replicator.cs
pinzart90: YEs...I kept the name from the old vm function.
(as a general rule, if the array type is CLRStackValue...then it represents the arguments)

src/Engine/ProtoCore/DSASM/InstructionSet.cs
pinzart90: That is still the case.
But this does incur some performance penalties due to the extra Marshaling/Unmarshaling that has to happen for every returned value and input argument

src/Engine/EmitMSIL/Replication.cs
pinzart90: CodeGEn can add more type information  (class, primitives, methods etc) to the RuntimeCore so that it can be later used during runtime.
The old VM populates its Type information during EMITImportStatement and EMitClassDefinition stages of the compiler.

src/Engine/EmitMSIL/Replication.cs
pinzart90: only Stdout for now

src/Engine/ProtoCore/DSASM/InstructionSet.cs
pinzart90: yes..we still use boxing/unboxing with this approach

src/Engine/EmitMSIL/Replication.cs
pinzart90: We are marshaling between the pure CLR objects and the CLRStackValue structure.
We need the CLRStackValue, runtime core and class table only for performance reasons (and maybe to simplify the copy/paste process).

We need type information (that can represent CLR and DesingScript types - ex `var`), rank, coercions scores etc at runtime. Having this information pre-processed and stored at compile time ...will improve performance at runtime.

pinzart90: At the moment I am doing most of this work at runtime...just to avoid the extra work of doing it at compile time.
But a lot of the performance gain comes when we pre-process this info at compile time

src/Engine/EmitMSIL/Replication.cs
pinzart90: It should be relatively easy...if we refactor the replication functions a bit...so that we can re-use them at compile time

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -2,17 +2,107 @@
...
+            foreach (var arg in args)
+            {
+                var protoType = ProtoFFI.CLRObjectMarshaler.GetProtoCoreType(arg.GetType());
+                CLRStackValue dsObj = marshaller.Marshal(arg, protoType, runtimeCore);
+                stackValues.Add(dsObj);
+            }
+            return stackValues;
+        }
+
+        internal static List<object> UnmrshalFunctionArguments(List<CLRFunctionEndPoint.ParamInfo> formalParams, List<CLRStackValue> args, MSILRuntimeCore runtimeCore)
```
pinzart90, line 30: `UnmrshalFunctionArguments` is copied from the CLRFFIFunctionPointer.Execute function (the marshaling/conversion parts)

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -2,17 +2,107 @@
...
+                {
+                    string message = String.Format(Resources.kFFIFailedToObtainObject, paramType.Name, formalParams[i].CLRInfo.Member.DeclaringType.Name, formalParams[i].CLRInfo.Member.Name);
+                    runtimeCore.LogWarning(ProtoCore.Runtime.WarningID.AccessViolation, message);
+                    return null;
+                }
+            }
+            return values;
+        }
+
+        internal static List<CLRFunctionEndPoint> ConvertMethodsToFEPs(IEnumerable<MethodBase> methods)
```
pinzart90, line 89: This is new code...that wraps MethodBase instances with `CLRFunctionEndPoint`
Similar to old VM's `FunctionEndPoint` but a lot simpler

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -133,5 +231,324 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
...
+                    arguments,
+                    replicationInstructions,
+                    candidatesWithCastDistances,
+                    candidateFunctions,
+                    candidatesWithDistances);*/
+
+            return compliantTarget;
+        }
+
+        private static void ComputeFeps(List<CLRStackValue> arguments,
```
pinzart90, line 270: Copied over from `CallSite.ComputeFeps`. Not much changed

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -133,5 +231,324 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
...
+            #endregion
+
+            resolvedFeps = new List<CLRFunctionEndPoint>();
+            replicationInstructions = instructions;
+        }
+
+        private static CLRFunctionEndPoint SelectFinalFep(IEnumerable<CLRFunctionEndPoint> functionEndPoints, List<CLRStackValue> formalParameters)
+        {
+            // TODO_MSIL: Determine final function endpoint here based on fitting runtime args to function parameters
+            return functionEndPoints.FirstOrDefault();
```
pinzart90, line 356: This is where method overload resolution should happen

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -133,5 +231,324 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
...
+        }
+
+        private static object ExecWithZeroRI(CLRFunctionEndPoint finalFep, List<CLRStackValue> formalParameters, MSILRuntimeCore runtimeCore)
```
pinzart90, line 359: copied over from CallSite.ExecWithZeroRI

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -133,5 +231,324 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
...
+            {
+                return o.Value as IList<CLRStackValue>;
+            }
+            else
+            {
+                return new List<CLRStackValue>() { o };
+            }
+        }
+
+        private static object ExecWithRISlowPath(CLRFunctionEndPoint finalFep, List<CLRStackValue> formalParameters,
```
pinzart90, line 404: copied over from CallSite.ExecWithRISlowPath

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -2032,6 +2032,63 @@ public static StackValue PerformReturnTypeCoerce(ProcedureNode procNode, Runtime
             }
         }
 
+        internal static CLRStackValue PerformReturnTypeCoerce(CLRFunctionEndPoint procNode, CLRStackValue ret, MSILRuntimeCore runtimeCore)
```
pinzart90, line 2035: copied from `StackValue PerformReturnTypeCoerce` right above

src/Engine/ProtoCore/DSASM/InstructionSet.cs
```diff
@@ -110,6 +110,112 @@ public struct MetaData
...
+        // TODO_MSIL: Figure out if we need to use an AddressType, like StackValue
+        public bool IsEnumerable => Type != typeof(string) && ArrayUtils.IsEnumerable(Type);
+
+        public bool IsDouble => Value is double;
+
+        // TODO_MSIL: Figure out how to set/use this flag;
+        public bool IsDefaultArgument => false;
+
+        // TODO_MSIL: Figure out if we need to use an AddressType, like StackValue
+        public bool IsPointer => Type != typeof(string) && !ArrayUtils.IsEnumerable(Type) && !Type.IsValueType && Type.IsClass;
```
pinzart90, line 127: Not sure I have these covered well enough...
This flag probably needs to be explicitly set during MSILEmit phase (when we get to that...)

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -837,13 +840,20 @@ private Type EmitRangeExprNode(AssociativeNode node)
...
+            AssociativeNode dummyStepNode = AstFactory.BuildDoubleNode(1);
             var arguments = new List<AssociativeNode>
             {
                 fromNode,
                 toNode,
-                stepNode ?? new NullNode(),
+                // TODO_MSIL: Figure out a better solution for this scenario.
+                // Use DoubleNode(1) because standard replication cannot handle null to value type coerce
+                // The old VM handles builtin functions (like range expr) in a special way...that does not try coercion
+                hasStep ? stepNode : dummyStepNode,//NullNode()
```
pinzart90, line 854: FYI @mjkkirschner

pinzart90:
Single test fails on jenkins ... but passes locally
```
SetUp : System.IO.IOException : The process cannot access the file 'D:\jenkins\workspace\Dynamo\DynamoSelfServe\pullRequestValidation\Dynamo\bin\AnyCPU\Release\OpCodesTEST.txt' because it is being used by another process.
```

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -837,13 +840,20 @@ private Type EmitRangeExprNode(AssociativeNode node)
                     op = new IntNode(-1);
                     break;
             }
+
+            bool hasStep = stepNode != null;
+            // The value of the dymy DoubleNode does not matter since the hasStep boolean will be false.
```
sm6srw, line 845: Spelling

sm6srw (approved):
This is a lot. Looks good to me.

src/Engine/ProtoCore/Lang/Type.cs
```diff
@@ -361,6 +369,16 @@ public static StackValue ClassCoerece(StackValue sv, Type targetType, RuntimeCor
             return sv;
         }
 
+        internal static CLRStackValue ClassCoerece(CLRStackValue sv, Type targetType)
```
aparajit-pratap, line 372: spelling: ClassCoerce

src/Engine/ProtoCore/Utils/ClassUtils.cs
```diff
@@ -51,8 +51,33 @@ public static int GetUpcastCountTo(ClassNode from, ClassNode to, RuntimeCore run
                 return int.MaxValue;
 
             return upcastChain.IndexOf(toID);
+        }
 
+        // TODO_MSIL: replace with the existing ClassNode signature of this function when ClassNode are properly populated in MSIL.
+        internal static int GetUpcastCountTo((int id, System.Type type) from, (int id, System.Type type) target)
```
aparajit-pratap, line 57: Why do we need to replace with `ClassNode` if this implementation works?

src/Engine/ProtoCore/RuntimeCore.cs
```diff
@@ -378,4 +378,54 @@ public void RemoveExecutedAstGuids()
...
+            if (typeUID == ProtoCore.DSASM.Constants.kInvalidIndex)
+            {
+                typeUID = CLRModuleType.GetProtoCoreType(type, null).UID;
+            }
+            return typeUID;
+        }
+
+        internal int GetCoercionScore(int srcType, int targetType)
+        {
+            return ClassTable?.ClassNodes[srcType]?.GetCoercionScore(targetType) ?? (int)ProcedureDistance.NotMatchScore;
```
aparajit-pratap, line 428: You said you're not populating the classtable currently for `MSILRuntimeCore` and are using the old VM's classtable, correct? Can you point to the place in this PR where you're doing so?

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -133,5 +242,324 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
             return new List<List<ReplicationGuide>>();
         }
 
+        private static CLRFunctionEndPoint GetCompliantFEP(
+            List<CLRStackValue> arguments,
+            IEnumerable<CLRFunctionEndPoint> funcGroup,
+            List<ReplicationInstruction> replicationInstructions,
```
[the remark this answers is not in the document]
pinzart90: No real reason. My thinking was that if the list is only iterated in a function body...then I pass it as an IEnumerable...
I can pass it everywhere as List...to be consistent

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -133,5 +242,324 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
...
+            }
+            else
+            {
+                result = finalFep.method.Invoke(args[0], args.Skip(1).ToArray());
+            }
+
+            var marshaller = ProtoFFI.CLRDLLModule.GetMarshaler(runtimeCore);
+            CLRStackValue dsRetValue = marshaller.Marshal(result, ProtoFFI.CLRObjectMarshaler.GetProtoCoreType(finalFep.ReturnType), runtimeCore);
+
+            if (!dsRetValue.IsExplicitCall)
```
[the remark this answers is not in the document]
pinzart90: I am not sure. I marked the flag with TODO to figure it out

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -133,5 +242,324 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
...
+            ReplicationInstruction ri = replicationInstructions[0];
+
+            if (ri.Zipped)
+            {
+                ZipAlgorithm algorithm = ri.ZipAlgorithm;
+
+                //For each item in this plane, an array of the length of the minimum will be constructed
+
+                //The size of the array will be the minimum size of the passed arrays
+                List<int> repIndecies = ri.ZipIndecies;
```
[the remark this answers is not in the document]
pinzart90: THat is part of the ReplicationInstruction struct, used by the old VM. I did not want to add even more changes if not critical

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -133,5 +242,324 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
...
+                    default:
+                        throw new ReplicationCaseNotCurrentlySupported(Resources.AlgorithmNotSupported);
+                }
+
+                bool hasEmptyArg = false;
+                foreach (int repIndex in repIndecies)
+                {
+                    // TODO: Investigate convertToArray performance
+                    var subParameters = getSubParameters(formalParameters[repIndex]).ToArray();
+                    parameters.Add(subParameters.ToArray());
```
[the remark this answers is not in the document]
pinzart90: correct. I will remove the redundant call

src/Engine/ProtoCore/DSASM/InstructionSet.cs
```diff
@@ -110,6 +110,112 @@ public struct MetaData
...
+        public bool IsDouble => Value is double;
+
+        // TODO_MSIL: Figure out how to set/use this flag;
+        public bool IsDefaultArgument => false;
+
+        // TODO_MSIL: Figure out if we need to use an AddressType, like StackValue
+        public bool IsPointer => Type != null && Type != typeof(string) && !ArrayUtils.IsEnumerable(Type) && !Type.IsValueType && Type.IsClass;
+
+        // TODO_MSIL: Figure out how to set/use this flag;
+        public bool IsExplicitCall => false;
```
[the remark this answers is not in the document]
pinzart90: I will investigate in a separate task

src/Engine/ProtoCore/DSASM/InstructionSet.cs
```diff
@@ -110,6 +110,112 @@ public struct MetaData
...
+        #region Converters
+        public CLRStackValue ToBoolean()
+        {
+            switch (TypeUID)
+            {
+                case (int)PrimitiveType.Bool:
+                    return this;
+
+                case (int)PrimitiveType.Integer:
+                    return new CLRStackValue(Value != null, (int)PrimitiveType.Bool);
```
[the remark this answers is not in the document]
pinzart90: Correct. thanks

src/Engine/ProtoCore/DSASM/InstructionSet.cs
```diff
@@ -110,6 +110,112 @@ public struct MetaData
...
+                default:
+                    return CLRStackValue.Null;
+            }
+        }
+
+        public CLRStackValue ToInteger()
+        {
+            switch (TypeUID)
+            {
+                case (int)PrimitiveType.Integer:
```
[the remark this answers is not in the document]
pinzart90: I will add a separate task for this.

src/Engine/ProtoCore/Lang/FunctionEndPoint.cs
```diff
@@ -7,6 +8,217 @@
...
+                    }
+                }
+
+                if (expectedType.IsIndexable && (expectedType.IsIndexable == args[i].IsEnumerable))
+                {
+                    //In case of conversion from double to int, add a conversion score.
+                    //There are overloaded methods and the difference is the parameter type between int and double.
+                    //Add this to make it call the correct one. - Randy
+                    bool bContainsDouble = ArrayUtils.ContainsDoubleElement(args[i]);
+                    if (expectedType.UID == (int)PrimitiveType.Integer && bContainsDouble)
```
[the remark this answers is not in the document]
pinzart90: I will add a separate task for this.

src/Engine/ProtoCore/Lang/FunctionGroup.cs
```diff
@@ -321,6 +329,23 @@ public static bool CheckInvalidArrayCoersion(FunctionEndPoint fep, List<StackVal
...
+        internal static Dictionary<CLRFunctionEndPoint, int> GetCastDistances(IEnumerable<CLRFunctionEndPoint> funcGroup, List<CLRStackValue> formalParams, List<ReplicationInstruction> replicationInstructions)
+        {
+            Dictionary<CLRFunctionEndPoint, int> ret = new Dictionary<CLRFunctionEndPoint, int>();
+
+            //@PERF: Consider parallelising this
+
+            IEnumerable<CLRFunctionEndPoint> feps = funcGroup;
+            List<CLRStackValue> reducedParamSVs = Replicator.EstimateReducedParams(formalParams, replicationInstructions);
+
+            foreach (var fep in feps)
```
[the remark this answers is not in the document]
pinzart90: will do

src/Engine/ProtoCore/Lang/Type.cs
pinzart90: will do

src/Engine/ProtoCore/Utils/ClassUtils.cs
pinzart90: To re-use the pre-existing method (if possible)
Not sure if the ClassNode can do anything more (represent something more) than System.Type can.
Maybe pure DesignScript classes ? not sure if we can define classes in design script...And if we can...maybe we can simply create System.Type instead of a ClassNode, but since the ClassNode already exists...we might as well use it

src/Engine/ProtoCore/RuntimeCore.cs
pinzart90: Look at the method above this one
`GetProtoCoreTypeID`
```
            int typeUID = ClassTable.IndexOf(CLRObjectMarshaler.GetTypeName(type));
            if (typeUID == ProtoCore.DSASM.Constants.kInvalidIndex)
            {
                typeUID = CLRModuleType.GetProtoCoreType(type, null).UID;
            }
```
`CLRModuleType.GetProtoCoreType(type, null)`  - CLRModuleType has most of the types loaded by the old VM

image

src/Engine/ProtoCore/Utils/ClassUtils.cs
aparajit-pratap: Yes, we have builtin DesignScript classes to implement some built-in nodes. See BuiltIn.ds and FunctionObject.ds. FunctionObject.ds is mainly to create partial functions and support nodes that accept function pointers and we might be doing away with those so it may not be useful to continue supporting it. BuiltIn.ds has some `List` nodes that are implemented in DesignScript. These might need to be ported to C# functions.

aparajit-pratap:
> Single test fails on jenkins ... but passes locally
> 
> ```
> SetUp : System.IO.IOException : The process cannot access the file 'D:\jenkins\workspace\Dynamo\DynamoSelfServe\pullRequestValidation\Dynamo\bin\AnyCPU\Release\OpCodesTEST.txt' because it is being used by another process.
> ```

I think we can skip generating this file in test mode as it's purpose is only for development and debugging.

src/Engine/ProtoCore/Utils/ClassUtils.cs
pinzart90: Well I guess we can keep this method then (I will remove the TODO)
But I am not sure if we can/should entirely get rid of the ClassNode (and simply replace it with System.Type ).
I guess we can figure that out later

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -133,5 +242,324 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
...
+            if (ri.Zipped)
+            {
+                ZipAlgorithm algorithm = ri.ZipAlgorithm;
+
+                //For each item in this plane, an array of the length of the minimum will be constructed
+
+                //The size of the array will be the minimum size of the passed arrays
+                List<int> repIndecies = ri.ZipIndecies;
+
+                //this will hold the heap elements for all the arrays that are going to be replicated over
```
aparajit-pratap, line 436: This comment doesn't seem to apply anymore.

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
pinzart90: actually that still applies (maybe not be best way..but it still works)

src/Engine/EmitMSIL/Replication.cs
pinzart90: removed

aparajit-pratap (approved):
Few more questions, basically LGTM.

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
```diff
@@ -430,7 +515,122 @@ private ICollection ToICollection(StackValue dsObject, ProtoCore.Runtime.Context
...
+
+            return arrList;
+        }
+
+        internal override CLRStackValue Marshal(object obj, ProtoCore.Type type, ProtoCore.MSILRuntimeCore runtimeCore)
+        {
+            var collection = obj as IEnumerable;
+            Validity.Assert(null != collection, "Expected IEnumerable object for marshaling as collection");
+            if (null == collection)
+                return CLRStackValue.Null;
```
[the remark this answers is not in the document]
pinzart90: Looks like it. Copy pasted from original.
I will remove the unreachable code

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
```diff
@@ -430,7 +515,122 @@ private ICollection ToICollection(StackValue dsObject, ProtoCore.Runtime.Context
...
+            List<CLRStackValue> svs = new List<CLRStackValue>();
+            foreach (var item in collection)
+            {
+                if (item is CLRStackValue oldSv)
+                {
+                    svs.Add(oldSv);
+                    continue;
+                }
+                var newSv = primitiveMarshaler.Marshal(item, GetApproxDSType(item), runtimeCore);
+                svs.Add(newSv);
```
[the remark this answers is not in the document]
pinzart90: This logic was copied from the original:
1. If the array items to be marshaled are already CLRStackValues (boxed as objects) then no need to marshal again and we simply keep the existing values (this is the oldSv)
2. If the array items are not CLRStackValues then we marshal them.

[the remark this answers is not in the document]
pinzart90: maybe old vs new are not the best variable names in this case

pinzart90:
Only 1 failure that passes locally: `pullRequestValidation\Dynamo\bin\AnyCPU\Release\OpCodesTEST.txt' because it is being used by another`
I am going to merge

mjkkirschner:
@pinzart90 I think I've addressed the concurrent test file access on my pr.

PR #13126, opened 2022-07-19
DYN-5096: Create and emit calls to common Built-in methods e.g. Math operators

sm6srw:
### Purpose

This pull request does:

* add implementations for common operators.
* use `object` types.

Next possible step:
Use overloads when argument types are known at compile time.

### Declarations

### Release Notes

### Reviewers

### FYIs


src/Engine/EmitMSIL/Operators.cs
```diff
@@ -0,0 +1,244 @@
+﻿using System;
```
sm6srw, line 1: This file should probably be somewhere else. Or?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -135,51 +134,63 @@ private IEnumerable<MethodBase> FunctionLookup(IList args)
...
+
+                            if (prop != null)
+                            {
+                                mi = prop.GetAccessors().ToList();
+                            }
+                        }
+
+                        // TODO: Add check for constructorinfo objects
+
+                        if (mi != null)
```
mjkkirschner, line 174: I found that sometimes the lookups above can return an empty list and then we end up with an empty list in the method cache.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -135,51 +134,63 @@ private IEnumerable<MethodBase> FunctionLookup(IList args)
...
+                        //if (method != null)
+                        //{
+                        //    argTypes = method.GetParameters().Select(p => p.ParameterType).ToList();
+                        //    return method.ReturnType;
+                        //}
+                    }
+                }
+                else
+                {
+                    mi = Operators.GetOperatorMethod(methodName);
```
mjkkirschner, line 189: does this always return null if it can't find an operator?

src/Engine/EmitMSIL/Operators.cs
```diff
@@ -0,0 +1,244 @@
...
+using System.Threading;
+using ProtoCore.Utils;
+
+namespace EmitMSIL
+{
+    public class Operators
+    {
+        private static string GetOperatorMethodName(string name)
+        {
+            if (name[0] == '%')
```
mjkkirschner, line 13: I feel like protocore/ the parser has a bunch of this logic already perhaps?

src/Engine/EmitMSIL/Operators.cs
```diff
@@ -0,0 +1,244 @@
...
+            {
+                return a <= b;
+            }
+
+            return false;
+        }
+
+        public static bool IsOperator(string name)
+        {
+            if (name.Length > 0)
```
mjkkirschner, line 225: same here, I think you can replace this with checks on the AST nodes directly instead of doing string comparisons.

src/Engine/EmitMSIL/Operators.cs
```diff
@@ -0,0 +1,244 @@
...
+            {
+                name = name.Remove(0, 1);
+            }
+
+            return name[0].ToString().ToUpper() + name.Substring(1);
+        }
+
+        private static bool ToBoolean(dynamic a)
+        {
+            if (a is bool ab)
```
mjkkirschner, line 23: just a style thing, I think you can `switch` on type now, though we may not yet use a c# version that supports that.

src/Engine/EmitMSIL/Operators.cs
```diff
@@ -0,0 +1,244 @@
...
+
+        public static dynamic Mul(dynamic a, dynamic b)
+        {
+            return a * b;
+        }
+
+        public static dynamic Div(dynamic a, dynamic b)
+        {
+            //division is always carried out as a double
+            double lhs = a;
```
mjkkirschner, line 79: why?

src/Engine/EmitMSIL/Operators.cs
```diff
@@ -0,0 +1,244 @@
...
+
+            if (a is char ac)
+            {
+                return ac != 0;
+            }
+
+            return false;
+        }
+
+        public static dynamic Add(dynamic a, dynamic b)
```
mjkkirschner, line 61: now that you've implemented all of this as dynamic, it looks pretty clean, I wonder what the performance is vs using object.

This still boxes but I think by using the DLR we get the benefits of cacheing the DLR performs when given different argument types.

src/Engine/EmitMSIL/CodeGenIL.cs
sm6srw: It currently throws when that happens so not an issue. But I should make this consistent.

src/Engine/EmitMSIL/Operators.cs
sm6srw: Thanks, I wll give it a try.

src/Engine/EmitMSIL/Internals.cs
```diff
@@ -0,0 +1,235 @@
...
+                return true;
+            }
+
+            return false;
+        }
+
+        public static dynamic Add(dynamic a, dynamic b)
+        {
+            return a + b;
+        }
```
aparajit-pratap, line 45: One thing I see with this is that it could cause some differences with the old `Add` method in Dynamo. For example, in this case, an `int` can be added with a `char` and still give a valid result (the resulting ASCII value), whereas in the old VM, it would return null, just because it was implemented differently.
image

aparajit-pratap: @mjkkirschner trying to remember, is this approach something you tried for `GenerateRangeIL `functions, where you could define the inputs having dynamic type?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -135,51 +134,63 @@ private IEnumerable<MethodBase> FunctionLookup(IList args)
...
+
+                            if (prop != null)
+                            {
+                                mi = prop.GetAccessors().ToList();
+                            }
+                        }
+
+                        // TODO: Add check for constructorinfo objects
+
+                        if (mi != null || mi.Any())
```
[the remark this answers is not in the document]
sm6srw: Oops, thanks!

src/Engine/EmitMSIL/Internals.cs
sm6srw: Yes, I think many (all) of these new methods are not fully compatible in what they support and how they fail.  Something we need to look at.

mjkkirschner: @aparajit-pratap no I didn't try dynamic, I did use `object` as my attempt at a dynamic range creation conversion, but then we decided to file followups:
1. evaluate stackvalue,object vs dynamic performance: https://jira.autodesk.com/browse/DYN-5110
2. Implement dynamic range generation for character ranges: https://jira.autodesk.com/browse/DYN-5109

where the idea was that doing 1 would give us some direction in how to implement 2. 

Maybe @sm6srw and @pinzart90 are kind of investigating 1 implicitly now though.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -20,6 +18,7 @@ public class CodeGenIL:IDisposable
         private ILGenerator ilGen;
         internal string className;
         internal string methodName;
+        private bool isInternal;
```
mjkkirschner, line 21: maybe deserves a comment?

src/Engine/EmitMSIL/Internals.cs
```diff
@@ -0,0 +1,235 @@
...
+                methodName = op.ToString();
+            }
+            else if (CoreUtils.TryGetUnaryOperator(name, out UnaryOperator un))
+            {
+                methodName = un.ToString();
+            }
+
+            if (methodName.Length > 0)
+            {
+                methodName = methodName[0].ToString().ToUpper() + methodName.Substring(1);
```
mjkkirschner, line 221: ??

src/Engine/EmitMSIL/CodeGenIL.cs
aparajit-pratap: I think `builtin` would be a better name as it is used currently in the VM.

src/Engine/EmitMSIL/Operators.cs
aparajit-pratap: @sm6srw is the only difference implementation-wise that we save the need for emitting opcodes for boxing and unboxing?

src/Engine/EmitMSIL/CodeGenIL.cs
sm6srw: Both are used in the current VM but I think Internal is the most common. For example, the method for checking for these are `CoreUtils.IsInternalMethod()`. But I am OK changing to BuiltIn.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -611,8 +623,14 @@ private Type EmitFunctionCallNode(AssociativeNode node)
...
             var args = fcn.FormalArguments;
             var numArgs = args.Count;
+            isInternal = CoreUtils.IsInternalMethod(methodName);
 
-            if(compilePass == CompilePass.MethodLookup)
+            if (isInternal)
+            {
+                className = nameof(Internals);
+            }
+
```
aparajit-pratap, line 632: Why don't we call this inside FunctionLookup itself? This way we won't need to store `isInternal` as a class member.

src/Engine/EmitMSIL/Internals.cs
```diff
@@ -0,0 +1,233 @@
...
+            {
+                return a > b;
+            }
+
+            return false;
+        }
+
+        public static bool lt(dynamic a, dynamic b)
+        {
+            if (a is double || b is double || a is long || b is long)
```
aparajit-pratap, line 179: These functions assume all integer arguments to be `long`, which means we need to convert all returned `Int32`s to long. I actually don't think that's necessary anymore. We must convert them to `long` only when needed, which means that we should support all these methods for int32 as well.

src/Engine/EmitMSIL/Internals.cs
```diff
@@ -0,0 +1,233 @@
...
+            {
+                return a <= b;
+            }
+
+            return false;
+        }
+
+        public static IEnumerable<MethodBase> GetInternalMethod(string name)
```
aparajit-pratap, line 207: This looks like it's always returning a single MethodBase object, no?

src/Engine/ProtoCore/Utils/CoreUtils.cs
```diff
@@ -254,6 +254,20 @@ public static bool TryGetOperator(string methodName, out Operator op)
             return System.Enum.TryParse(realMethodName, out op);
         }
 
+        public static bool TryGetUnaryOperator(string methodName, out UnaryOperator op)
+        {
+            Validity.Assert(null != methodName);
+            if (!methodName.StartsWith(DSASM.Constants.kInternalNamePrefix))
```
pinzart90, line 260: couldn't we use `CoreUtils.IsInternalMethod(methodName)` instead here ?

src/Engine/EmitMSIL/Operators.cs
pinzart90: I am curious about this too. I will try to read up on the benefits of dynamic vs just using object.

src/Engine/EmitMSIL/CodeGenIL.cs
sm6srw: The reason is that `KeyGen` needs a consistent class name and I tried to avoid calling `CoreUtils.IsInternalMethod(methodName)` inside that method but that is maybe OK?

src/Engine/EmitMSIL/Internals.cs
sm6srw: Thanks. This has confused me. I have never seen anything than `long` so far so thought that was on purpose.

src/Engine/EmitMSIL/Internals.cs
sm6srw: For now yes.

sm6srw: I was kind of preparing for overloads.

src/Engine/ProtoCore/Utils/CoreUtils.cs
sm6srw: Yes,  we could. I just copied the code from the function above for normal operators.

src/Engine/EmitMSIL/Internals.cs
aparajit-pratap: I needed to keep it long when emitting integer nodes to maintain the same behaviour with the old DS engine specs.

sm6srw: Ok, so there is no need for me to support `int32` in these methods?

src/Engine/EmitMSIL/Operators.cs
sm6srw: Now using objects.

src/Engine/EmitMSIL/Internals.cs
sm6srw: Now using objects.

mjkkirschner (approved):

src/Engine/EmitMSIL/BuiltIn.cs
```diff
@@ -0,0 +1,383 @@
...
+                    return !Double.IsNaN(ad) && !ad.Equals(0.0);
+
+                case string str:
+                    return !string.IsNullOrEmpty(str);
+
+                case char ac:
+                    return ac != 0;
+            }
+
+            if (a is Pointer)
```
aparajit-pratap, line 36: What is `Pointer`? Is it a C# type?

src/Engine/EmitMSIL/BuiltIn.cs
```diff
@@ -0,0 +1,383 @@
...
+            if ((a is long || a is int) && (b is long || b is int))
+            {
+                var al = Convert.ToInt64(a);
+                var bl = Convert.ToInt64(b);
+                return HandleOverflow(() => checked(al + bl), () => al + bl);
+            }
+
+            if ((a is long || a is double || a is int) && (b is long || b is double || b is int))
+            {
+                return Convert.ToDouble(a) + Convert.ToDouble(b);
```
aparajit-pratap, line 87: At this point, I believe, either a is double or b is double. Do we really need to convert the other number to double before we add? Won't it work even without any conversion?

src/Engine/EmitMSIL/BuiltIn.cs
```diff
@@ -0,0 +1,383 @@
...
+            if (a is int ai && b is int bi)
+            {
+                return HandleOverflow(() => checked(ai + bi), () => ai + bi);
+            }
+
+            if ((a is long || a is int) && (b is long || b is int))
+            {
+                var al = Convert.ToInt64(a);
+                var bl = Convert.ToInt64(b);
+                return HandleOverflow(() => checked(al + bl), () => al + bl);
```
aparajit-pratap, line 82: At this point, I think with `a` or `b` is `long`. Won't the addition automatically work and return the result as `long`? I don't think the conversion to `long` is needed in this case.

src/Engine/EmitMSIL/BuiltIn.cs
sm6srw: I think so. This is from the old implementation.

src/Engine/EmitMSIL/BuiltIn.cs
```diff
@@ -0,0 +1,383 @@
...
+        {
+            switch (a)
+            {
+                case int ai:
+                    return HandleOverflow(() => checked(-ai), () => -ai);
+                case long al:
+                    return HandleOverflow(() => checked(-al), () => -al);
+                case double ad:
+                    return -ad;
+            }
```
aparajit-pratap, line 208: Can we write `if(a is long || a is int || a is double) return -a;`?

src/Engine/EmitMSIL/BuiltIn.cs
sm6srw: Same here, trying to simplify the logic. Remember that a and b are objects.

src/Engine/EmitMSIL/BuiltIn.cs
sm6srw: No, does not work because `a` is an object.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -611,8 +624,14 @@ private Type EmitFunctionCallNode(AssociativeNode node)
             methodName = fcn.Function.Name;
             var args = fcn.FormalArguments;
             var numArgs = args.Count;
+            isBuiltIn = CoreUtils.IsInternalMethod(methodName);
```
aparajit-pratap, line 627: Why can't you call `IsInternalMethod` inside `FunctionLookup` itself? This way, you won't have to create the isBuiltIn member.

src/Engine/EmitMSIL/BuiltIn.cs
sm6srw: Is is a 'class System.Reflection.Pointer`.

aparajit-pratap (approved):
LGTM except for one comment about the need to have `isBuiltIn`.

src/Engine/EmitMSIL/CodeGenIL.cs
sm6srw: The code currently relies on that the `className` member is stable for built in methods (it is initially null or unchanged for builtin methods). className is used in 'KeyGen' and in `Function Lookup`.

An alternative would be to make a property of 'className' and implement it as:
`internal string ClassName { get => CoreUtils.IsInternalMethod(methodName) ? nameof(BuiltIn) : className; set => className = value; }`

Do you think this approach is better?

sm6srw: Ok, I got rid of the flag (without making a property of `className`).

PR #13127, opened 2022-07-20
Rank 1 array coercion

aparajit-pratap:
### Purpose

This PR adds coercion between `int[]`, `long[]`, and `double[]` and between number arrays and `List<int/long/double>`. It also adds coercion between `IList` and `IEnumerable<int/long/double>`

These should cover most cases of coercion of collections of numerical types. There could be some that have been left out, which we can return to at a later point in time.

TODO:

- Currently, 1 test fails which expects a string of a single character, whereas it's returned as a character if the string has only one char in the new engine. 

### Declarations

### Release Notes

### FYIs


src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -115,16 +113,182 @@ public CodeGenIL(IDictionary<string, IList> input, string filePath)
...
+
+            // arr.CopyTo(newarr, 0);
+            EmitOpCode(OpCodes.Callvirt, mInfo);
+
+            EmitOpCode(OpCodes.Ldloc, newArrIndex);
+
+            return t;
+        }
+
+        private Type EmitArrayCoercion<S, T>()
```
mjkkirschner, line 202: any reason not to have more descriptive type names?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -115,16 +113,182 @@ public CodeGenIL(IDictionary<string, IList> input, string filePath)
...
+                EmitOpCode(OpCodes.Stelem_I4);
+            }
+            else if (typeof(T) == typeof(long))
+            {
+                EmitOpCode(OpCodes.Conv_I8);
+                EmitOpCode(OpCodes.Stelem_I8);
+            }
+            if (typeof(T) == typeof(double))
+            {
+                EmitOpCode(OpCodes.Conv_R8);
```
mjkkirschner, line 263: is it safe to call this method if S is already a double? (I may not be understanding the intention here)

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -115,16 +113,182 @@ public CodeGenIL(IDictionary<string, IList> input, string filePath)
...
+            {
+                EmitOpCode(OpCodes.Ldelem_I8);
+            }
+            else if (typeof(S) == typeof(int))
+            {
+                EmitOpCode(OpCodes.Ldelem_I4);
+            }
+            if (typeof(T) == typeof(int))
+            {
+                EmitOpCode(OpCodes.Conv_I4);
```
mjkkirschner, line 253: do these conversion opcodes end up looking like casts? Or do they convert at the bit level?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -115,16 +113,182 @@ public CodeGenIL(IDictionary<string, IList> input, string filePath)
...
+        {
+            MethodInfo mInfo = null;
+            // Find length for IList to be coerced (already on top of eval stack), len
+            var prop = typeof(ICollection).GetProperties(BindingFlags.Public | BindingFlags.Instance).Where(
+                                p => p.Name == nameof(ICollection.Count)).FirstOrDefault();
+
+            if (prop != null)
+            {
+                mInfo = prop.GetAccessors().FirstOrDefault();
+            }
```
mjkkirschner, line 173: if this is null, shouldn't we throw an exception?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -115,16 +113,182 @@ public CodeGenIL(IDictionary<string, IList> input, string filePath)
...
+            //var mInfo = typeof(ICollection).GetMethod(nameof(ICollection.get_Count));
+            EmitOpCode(OpCodes.Callvirt, mInfo);
+
+            // var newarr = new T[len];
+            EmitOpCode(OpCodes.Newarr, typeof(T));
+
+            // Declare new array to store coerced values
+            var t = typeof(T).MakeArrayType();
```
mjkkirschner, line 181: but a `system.array` does not implement `IEnumerable<T>` ?

src/Engine/EmitMSIL/CodeGenIL.cs
aparajit-pratap: I guess so. Will add it.

src/Engine/EmitMSIL/CodeGenIL.cs
aparajit-pratap: This basically makes type `T` into type `T[]`.

src/Engine/EmitMSIL/CodeGenIL.cs
aparajit-pratap: Do you mean if this function is called as `EmitArrayCoercion<double, double>`?

aparajit-pratap: The intention here is basically to convert the source type (S) into the target (T) double type.

src/Engine/EmitMSIL/CodeGenIL.cs
aparajit-pratap: No reason, I just didn't think it was common to do that.

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: sorry, should have been more clear - but `T[]` does not implement `IEnumerable<T>` so it's not clear how this method could work to convert `IList` to `IEnumerable<T>`

aparajit-pratap: It doesn't, I attempted writing `EmitIListCoercion` to do that, which I haven't gotten to work yet.

aparajit-pratap: Here, I'm emitting code to initialize a new array of type `T[]`, then take, the `IList` and call `CopyTo` on it to copy its elements into the new array like: `ilist.CopyTo(newarr, 0);`

I think `IEnumerable<T>` is assignable from `T[]`, which is why I chose to convert the IList to `T[]`.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -115,16 +113,180 @@ public CodeGenIL(IDictionary<string, IList> input, string filePath)
...
         private Type EmitCoercionCode(Type argType, ParameterInfo param)
         {
+            if (param.ParameterType.IsAssignableFrom(argType)) return argType;
+
             if(argType == typeof(double) && param.ParameterType == typeof(long))
             {
                 EmitOpCode(OpCodes.Conv_I8);
                 return typeof(long);
             }
+            if (argType == typeof(double) || argType == typeof(long) && param.ParameterType == typeof(int))
```
pinzart90, line 123: maybe enclose the argType conditions in parenthesis ?
if ( `(`argType == typeof(double) || argType == typeof(long) `)` &&

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -115,16 +113,180 @@ public CodeGenIL(IDictionary<string, IList> input, string filePath)
...
+            {
+                EmitOpCode(OpCodes.Conv_I4);
+                return typeof(int);
+            }
+            if(argType == typeof(double[]) && typeof(IEnumerable<int>).IsAssignableFrom(param.ParameterType))
```
pinzart90, line 128: I think you can cover more cases with `IsAssignableFrom` for the `argType` check
`typeof(System.Collections.Generic.IEnumerable<int/double/long/etc>).IsAssignableFrom(argType.GetType())` instead of specifically checking for array of type.

pinzart90: But I guess you would need to support convertion of IEnumerables
Ex. `EmitIEnumerableCoercion<Src, Trg>()`

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -115,16 +113,180 @@ public CodeGenIL(IDictionary<string, IList> input, string filePath)
...
+            //{
+            //    return EmitIListCoercion<long>();
+            //}
             // TODO: Add more coercion cases here.
 
             return argType;
         }
 
+        // TODO: Does not work yet (intended to coerce IList to IEnumerable<T>)
+        private Type EmitIListCoercion<T>()
```
pinzart90, line 163: maybe convert this method to cover  IEnumerable<srcType> to IEnumerable<trgType>

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: @aparajit-pratap you are correct:
```
Single-dimensional arrays implement the [System.Collections.Generic.IList<T>](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ilist-1?view=net-6.0), [System.Collections.Generic.ICollection<T>](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.icollection-1?view=net-6.0), [System.Collections.Generic.IEnumerable<T>](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ienumerable-1?view=net-6.0), [System.Collections.Generic.IReadOnlyList<T>](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ireadonlylist-1?view=net-6.0) and [System.Collections.Generic.IReadOnlyCollection<T>](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ireadonlycollection-1?view=net-6.0) generic interfaces. The implementations are provided to arrays at run time, and as a result, the generic interfaces do not appear in the declaration syntax for the [Array](https://docs.microsoft.com/en-us/dotnet/api/system.array?view=net-6.0) class. In addition, there are no reference topics for interface members that are accessible only by casting an array to the generic interface type (explicit interface implementations). The key thing to be aware of when you cast an array to one of these interfaces is that members which add, insert, or remove elements throw [NotSupportedException](https://docs.microsoft.com/en-us/dotnet/api/system.notsupportedexception?view=net-6.0).
```

it even does implement those interfaces for single dimension arrays, but it's injected at runtime... what?

aparajit-pratap: Unfortunately, however, I'm getting an invalidcastexception, which I cannot seem to understand why. When I try the same exact code in linqpad it works.

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: just so I am clear, this doesn't work yet, but if it did work, it would only work incases were both the List/IEnumerable are of the same `T`?

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: I guess like this:
https://stackoverflow.com/questions/19497765/equivalent-of-cs-reinterpret-cast-in-c-sharp

😬

mjkkirschner: https://benbowen.blog/post/fun_with_makeref/

src/Engine/EmitMSIL/CodeGenIL.cs
aparajit-pratap: Correct.

mjkkirschner: have you tried with an `List<T>` ?

aparajit-pratap: Yes, I probably could try emitting:
```
var list = ilist.Cast<T>().ToList();
```

aparajit-pratap: Will try that, thanks.

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -46,10 +46,15 @@ public static IList ReplicationLogic(IEnumerable<MethodBase> mInfos, IList args,
...
             }
+            else if(mi.IsConstructor)
+            {
+                result = (mi as ConstructorInfo).Invoke(reducedArgs.ToArray());
+            }
             else
             {
                 result = mi.Invoke(reducedArgs[0], reducedArgs.Skip(1).ToArray());
             }
+            if (result is IList) return (IList)result;
```
pinzart90, line 57: so arrays/lists should not get wrapped in yet another list?
But single values should still get wrapped ?

pinzart90 (approved):
one question, then
LGTM

sm6srw (approved):

src/Engine/EmitMSIL/Replication.cs
aparajit-pratap: Single values need to get wrapped as the return type of the method is `IList`. I think we might be able to avoid it if we can avoid calling ReplicationLogic for non-replicated cases.

PR #13128, opened 2022-07-20
Pinned Note on Dismissed Warnings

dnenov:
### Purpose

The purpose of this PR is to rectify an issue where a pinned Note would remain in the same offset location after a Warning (dialog) has been dismissed. Normally, a Note would correctly get pushed up to provide space for the Warning tooltip dialog. However, when a Warning is dismissed, the space provided is no longer needed. This PR fixes this issue.

A secondary purpose of this PR is to make sure a Warning Dialog is always displayed on top, even in case of a more recent UI element (note). Currently, a note added after a Warning has been registered will appear on top of the Warning, which should not be the case. (Postponed)

A third potential goal is to allow to copy Connector Pins when copying/pasting Nodes with Pinned Connectors. (WIP)

**Pin Dismissed Warnings**
pin dismissed warnings

### Declarations

### Release Notes

- pinned Note correctly aligns to the Node once a Warning is dismissed

### Reviewers

@reddyashish 
@mjkkirschner 

### FYIs

@Amoursol

### DIFF
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -2522,7 +2522,12 @@ namespace Dynamo.Models
                                 connector.End != null && connector.End.Owner.IsSelected
                                     && !ClipBoard.Contains(connector));
 
+                var pins =
+                    connectors.SelectMany(x => x.ConnectorPinModels);
+
                 ClipBoard.AddRange(connectors);
+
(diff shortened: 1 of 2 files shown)


reddyashish (approved):
After some discussion, we will be addressing the ability to copy Connector Pins when copying/pasting Nodes with Pinned Connectors in a different task.

PR #13138, opened 2022-07-26
Handling the Custom Nodes and Trusted Location message

jesusalvino:
### Purpose

Handling the Trusted location message when a custom node space is opened.
https://jira.autodesk.com/browse/DYN-5089

### Declarations

### Reviewers
@QilongTang 

### FYIs
@RobertGlobant20 @filipeotero

### DIFF
--- a/src/DynamoCoreWpf/Controls/NotificationsControl.xaml.cs
+++ b/src/DynamoCoreWpf/Controls/NotificationsControl.xaml.cs
@@ -43,11 +43,6 @@ namespace Dynamo.Wpf.Controls
             var hsvm = (HomeWorkspaceViewModel)((DynamoViewModel)DataContext).HomeSpaceViewModel;            
             // Commented this after MAGN - 8423
             //hsvm.ClearWarning();
-            if (hsvm.IsHomeSpace && !hsvm.IsCurrentSpace)
-            {
-                hsvm.DynamoViewModel.FileTrustViewModel.ShowWarningPopup = false;
-                hsvm.ClearWarning();
-            }
         }
 
         private void StackPanel_MouseDown(object sender, MouseButtonEventArgs e)
--- a/src/DynamoCoreWpf/Controls/NotificationsControl.xaml.cs
+++ b/src/DynamoCoreWpf/Controls/NotificationsControl.xaml.cs
@@ -42,7 +42,7 @@ namespace Dynamo.Wpf.Controls
         {
             var hsvm = (HomeWorkspaceViewModel)((DynamoViewModel)DataContext).HomeSpaceViewModel;            
             // Commented this after MAGN - 8423
-            //hsvm.ClearWarning();
+            // hsvm.ClearWarning();
         }
 
         private void StackPanel_MouseDown(object sender, MouseButtonEventArgs e)
--- a/src/DynamoCoreWpf/Controls/NotificationsControl.xaml.cs
+++ b/src/DynamoCoreWpf/Controls/NotificationsControl.xaml.cs
@@ -40,7 +40,7 @@ namespace Dynamo.Wpf.Controls
         /// <param name="e"></param>
         void window_PreviewMouseDown(object sender, MouseButtonEventArgs e)
         {
-            var hsvm = (HomeWorkspaceViewModel)((DynamoViewModel)DataContext).HomeSpaceViewModel;            
+            var hsvm = (HomeWorkspaceViewModel)((DynamoViewModel)DataContext).HomeSpaceViewModel;
             // Commented this after MAGN - 8423
             // hsvm.ClearWarning();
         }
--- a/src/DynamoCoreWpf/Controls/NotificationsControl.xaml.cs
+++ b/src/DynamoCoreWpf/Controls/NotificationsControl.xaml.cs
@@ -40,9 +40,14 @@ namespace Dynamo.Wpf.Controls
         /// <param name="e"></param>
         void window_PreviewMouseDown(object sender, MouseButtonEventArgs e)
         {
-            var hsvm = (HomeWorkspaceViewModel)((DynamoViewModel)DataContext).HomeSpaceViewModel;
+            var hsvm = (HomeWorkspaceViewModel)((DynamoViewModel)DataContext).HomeSpaceViewModel;            
             // Commented this after MAGN - 8423
-            // hsvm.ClearWarning();
+            //hsvm.ClearWarning();
+            if (hsvm.IsHomeSpace && !hsvm.IsCurrentSpace)
+            {
+                hsvm.DynamoViewModel.FileTrustViewModel.ShowWarningPopup = false;
+                hsvm.ClearWarning();
+            }
         }
 
         private void StackPanel_MouseDown(object sender, MouseButtonEventArgs e)
--- a/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
@@ -111,7 +111,13 @@ namespace Dynamo.ViewModels
 
         public bool ViewingHomespace
         {
-            get { return model.CurrentWorkspace == HomeSpace; }
+            get {
+                if (model.CurrentWorkspace != HomeSpace)
+                {
+                    onNotViewingHomeSpace();
+                }
+                return model.CurrentWorkspace == HomeSpace; 
+            }
         }
 
         public int ScaleFactorLog
@@ -805,6 +811,15 @@ namespace Dynamo.ViewModels
             if (NodeViewReady != null)
             {
                 this.NodeViewReady(nodeView, new EventArgs());
+            }
+        }
+
+        internal event EventHandler NotViewingHomeSpace;
+        internal void onNotViewingHomeSpace()
+        {
+            if (NotViewingHomeSpace != null)
+            {
+                this.NotViewingHomeSpace(null, new EventArgs());
             }
         }
 
--- a/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
@@ -111,9 +111,7 @@ namespace Dynamo.ViewModels
 
         public bool ViewingHomespace
         {
-            get {                
-                return model.CurrentWorkspace == HomeSpace; 
-            }
+            get { return model.CurrentWorkspace == HomeSpace; }
         }
 
         public int ScaleFactorLog
@@ -1105,7 +1103,7 @@ namespace Dynamo.ViewModels
                         this.PublishCurrentWorkspaceCommand.RaiseCanExecuteChanged();
                     RaisePropertyChanged("IsPanning");
                     RaisePropertyChanged("IsOrbiting");
-                    //RaisePropertyChanged("RunEnabled");                    
+                    //RaisePropertyChanged("RunEnabled");
                     break;
 
                 case "EnablePresetOptions":
--- a/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
@@ -111,11 +111,7 @@ namespace Dynamo.ViewModels
 
         public bool ViewingHomespace
         {
-            get {
-                if (model.CurrentWorkspace != HomeSpace)
-                {
-                    onNotViewingHomeSpace();
-                }
+            get {                
                 return model.CurrentWorkspace == HomeSpace; 
             }
         }
@@ -814,15 +810,6 @@ namespace Dynamo.ViewModels
             }
         }
 
-        internal event EventHandler NotViewingHomeSpace;
-        internal void onNotViewingHomeSpace()
-        {
-            if (NotViewingHomeSpace != null)
-            {
-                this.NotViewingHomeSpace(null, new EventArgs());
-            }
-        }
-
         #region Event handler destroy/create
 
         protected virtual void UnsubscribeAllEvents()
@@ -1118,7 +1105,7 @@ namespace Dynamo.ViewModels
                         this.PublishCurrentWorkspaceCommand.RaiseCanExecuteChanged();
                     RaisePropertyChanged("IsPanning");
                     RaisePropertyChanged("IsOrbiting");
-                    //RaisePropertyChanged("RunEnabled");
+                    //RaisePropertyChanged("RunEnabled");                    
                     break;
 
                 case "EnablePresetOptions":
--- a/src/DynamoCoreWpf/ViewModels/Core/HomeWorkspaceViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/HomeWorkspaceViewModel.cs
@@ -270,6 +270,7 @@ namespace Dynamo.Wpf.ViewModels.Core
             if(IsHomeSpace && !IsCurrentSpace)
             {
                 SetCurrentWarning(NotificationLevel.Mild, string.Empty);
+                DynamoViewModel.FileTrustViewModel.ShowWarningPopup = false;
                 return;
             }
             if (RunSettings.ForceBlockRun)
@@ -329,7 +330,7 @@ namespace Dynamo.Wpf.ViewModels.Core
         /// <param name="hasErrors"></param>
         private void UpdateFooterItems(bool hasInfo, bool hasWarnings, bool hasErrors)
         {
-            
+
             if (hasErrors)
                 FooterNotificationItems[0].NotificationCount = Model.Nodes.Count(n => n.State == ElementState.Error);
             else
--- a/src/DynamoCoreWpf/ViewModels/Core/HomeWorkspaceViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/HomeWorkspaceViewModel.cs
@@ -278,7 +278,6 @@ namespace Dynamo.Wpf.ViewModels.Core
             if(IsHomeSpace && !IsCurrentSpace)
             {
                 SetCurrentWarning(NotificationLevel.Mild, string.Empty);
-                DynamoViewModel.FileTrustViewModel.ShowWarningPopup = false;
                 return;
             }
             if (RunSettings.ForceBlockRun)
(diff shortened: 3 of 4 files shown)


jesusalvino:
trusted

src/DynamoCoreWpf/Views/FileTrust/FileTrustWarning.xaml.cs
```diff
@@ -135,9 +135,10 @@ private void YesButton_Click(object sender, RoutedEventArgs e)
                 if (dynViewModel.PreferenceSettings.AddTrustedLocation(fileTrustWarningViewModel.DynFileDirectoryName))
                     dynViewModel.MainGuideManager.CreateRealTimeInfoWindow(string.Format(Properties.Resources.TrustLocationAddedNotification, fileTrustWarningViewModel.DynFileDirectoryName));
             }
-            if (dynViewModel.CurrentSpaceViewModel.RunSettingsViewModel.Model.RunType != RunType.Manual)
+            // The maximum workspaces always will be Two, The first one for the Dynamo file and the second one for a Custom Node
+            if (dynViewModel.Workspaces[0].RunSettingsViewModel.Model.RunType != RunType.Manual)
```
mjkkirschner, line 139: I think it's safer to not assume this won't change, and instead check that the workspace is a home workspace.

QilongTang: I agree

QilongTang (approved):

QilongTang:
ALthough this change is good, can you hide the trust warning when switching to a different workspace which is a custom node workspace?

jesusalvino:
> Although this change is good, I don't think we should display trust warning in the bug senario from the first place. Can you hide the trust warning when switching to a different workspace which is a custom node workspace?
trusted2

jesusalvino:
> Although this change is good, I don't think we should display trust warning in the bug senario from the first place. Can you hide the trust warning when switching to a different workspace which is a custom node workspace?

trusted4

src/DynamoCoreWpf/Controls/NotificationsControl.xaml.cs
```diff
@@ -40,9 +40,14 @@ void NotificationsControl_Loaded(object sender, RoutedEventArgs e)
...
-            var hsvm = (HomeWorkspaceViewModel)((DynamoViewModel)DataContext).HomeSpaceViewModel;
+            var hsvm = (HomeWorkspaceViewModel)((DynamoViewModel)DataContext).HomeSpaceViewModel;            
             // Commented this after MAGN - 8423
-            // hsvm.ClearWarning();
+            //hsvm.ClearWarning();
+            if (hsvm.IsHomeSpace && !hsvm.IsCurrentSpace)
+            {
+                hsvm.DynamoViewModel.FileTrustViewModel.ShowWarningPopup = false;
+                hsvm.ClearWarning();
+            }
```
QilongTang, line 50: hmm why do we need to touch this piece of code?

src/DynamoCoreWpf/ViewModels/Core/HomeWorkspaceViewModel.cs
```diff
@@ -270,6 +270,7 @@ internal void UpdateRunStatusMsgBasedOnStates()
             if(IsHomeSpace && !IsCurrentSpace)
             {
                 SetCurrentWarning(NotificationLevel.Mild, string.Empty);
+                DynamoViewModel.FileTrustViewModel.ShowWarningPopup = false;
```
QilongTang, line 273: From the gif, there will be a flash of trust warning displayed then disappear.. Less desirable, can we hide it even earlier?

src/DynamoCoreWpf/Controls/NotificationsControl.xaml.cs
jesusalvino: it forces to hide the untrusted message and bubble when we start to work in a just created custom node workspace, not only when we switch the workspaces.

QilongTang:
@jesusalvino Nice tries but I would love something more granular, for example, you can touch the visibility at https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs#L220 before executing the command to switch tabs. Or you  can subscribe to `ViewingHomespace` property change to determine the visibility of file trust warning. From my point of view, touching visibility after the tab is switched is less desirable to me.

jesusalvino:
> @jesusalvino Nice tries but I would love something more granular, for example, you can touch the visibility at https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs#L220 before executing the command to switch tabs. Or you can subscribe to `ViewingHomespace` property change to determine the visibility of file trust warning. From my point of view, touching visibility after the tab is switched is less desirable to me.

Done

trustedx

src/DynamoCoreWpf/Views/FileTrust/FileTrustWarning.xaml.cs
```diff
@@ -134,10 +139,10 @@ private void YesButton_Click(object sender, RoutedEventArgs e)
             {
                 if (dynViewModel.PreferenceSettings.AddTrustedLocation(fileTrustWarningViewModel.DynFileDirectoryName))
                     dynViewModel.MainGuideManager.CreateRealTimeInfoWindow(string.Format(Properties.Resources.TrustLocationAddedNotification, fileTrustWarningViewModel.DynFileDirectoryName));
-            }
-            if (dynViewModel.CurrentSpaceViewModel.RunSettingsViewModel.Model.RunType != RunType.Manual)
+            }            
+            if ((dynViewModel.HomeSpaceViewModel as HomeWorkspaceViewModel).RunSettingsViewModel.Model.RunType != RunType.Manual)
             {
-                dynViewModel.Model.CurrentWorkspace.RequestRun();
+                (dynViewModel.HomeSpaceViewModel as HomeWorkspaceViewModel).Model.RequestRun();
```
QilongTang, line 145: If we hide this dialog correctly, we should not run into case where `Yes` button is clicked on non-HomeSpaceViewModel. That was why I was asking you to hide it since it should not display in the first place. I dont think we need this check here..

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -111,7 +111,13 @@ public Point TransformOrigin
 
         public bool ViewingHomespace
         {
-            get { return model.CurrentWorkspace == HomeSpace; }
+            get {
+                if (model.CurrentWorkspace != HomeSpace)
+                {
+                    onNotViewingHomeSpace();
```
QilongTang, line 117: Making calls or throwing events in getter usually is not desired. We sometimes do it in setter but since in Dynamo we already have OnPropertyChanged() function ready for this, I would move this to there, there is a switch block to handle property change and what would be the consequences. We just need to flip the visibility based on the value of this property I think.

src/DynamoCoreWpf/Controls/NotificationsControl.xaml.cs
```diff
@@ -40,9 +40,9 @@ void NotificationsControl_Loaded(object sender, RoutedEventArgs e)
         /// <param name="e"></param>
         void window_PreviewMouseDown(object sender, MouseButtonEventArgs e)
         {
-            var hsvm = (HomeWorkspaceViewModel)((DynamoViewModel)DataContext).HomeSpaceViewModel;
+            var hsvm = (HomeWorkspaceViewModel)((DynamoViewModel)DataContext).HomeSpaceViewModel;            
             // Commented this after MAGN - 8423
-            // hsvm.ClearWarning();
+            //hsvm.ClearWarning();
```
QilongTang, line 45: Would you revert the change to this file? In general, there should be a space after the `//` for comments

src/DynamoCoreWpf/ViewModels/Core/HomeWorkspaceViewModel.cs
QilongTang: I dont think we still need this fix here in addition to other changes?

src/DynamoCoreWpf/Views/FileTrust/FileTrustWarning.xaml.cs
```diff
@@ -47,6 +47,11 @@ public FileTrustWarning(DynamoView dynamoViewWindow)
             SetUpPopup();
 
             HomeWorkspaceModel.WorkspaceClosed += CloseWarningPopup;
+            dynViewModel.NotViewingHomeSpace += DynViewModel_NotViewingHomeSpace; ;
```
QilongTang, line 50: unsubscribe.. But I think this solution is not general enough since it only handles the `NotViewingHomeSpace` case.. I have mentioned reasoning above.

src/DynamoCoreWpf/ViewModels/Core/HomeWorkspaceViewModel.cs
```diff
@@ -128,10 +128,18 @@ public HomeWorkspaceViewModel(HomeWorkspaceModel model, DynamoViewModel dynamoVi
             hwm.SetNodeDeltaState +=hwm_SetNodeDeltaState;
 
             dynamoViewModel.Model.ShutdownStarted += Model_ShutdownStarted;
-
+            dynamoViewModel.PropertyChanged += DynamoViewModel_PropertyChanged;
```
QilongTang, line 131: Thanks, we will also need to unsubscribe it somewhere

src/DynamoCoreWpf/ViewModels/Core/HomeWorkspaceViewModel.cs
```diff
@@ -128,10 +128,18 @@ public HomeWorkspaceViewModel(HomeWorkspaceModel model, DynamoViewModel dynamoVi
...
             SetupFooterNotificationItems();
         }
 
+        private void DynamoViewModel_PropertyChanged(object sender, PropertyChangedEventArgs e)
+        {
+            if (e.PropertyName == "CurrentSpace" && !((sender as DynamoViewModel).HomeSpaceViewModel as HomeWorkspaceViewModel).IsCurrentSpace)
```
QilongTang, line 137: hmm why not just watch `ViewingHomespace`? This seems to be the same check

src/DynamoCoreWpf/ViewModels/Core/HomeWorkspaceViewModel.cs
```diff
@@ -128,10 +128,18 @@ public HomeWorkspaceViewModel(HomeWorkspaceModel model, DynamoViewModel dynamoVi
...
+            {
+                ClearWarning();
+            }
```
QilongTang, line 140: Do we need an else case to make it visible when switching back from custom node workspace?

src/DynamoCoreWpf/Views/FileTrust/FileTrustWarning.xaml.cs
```diff
@@ -47,6 +47,15 @@ public FileTrustWarning(DynamoView dynamoViewWindow)
...
             HomeWorkspaceModel.WorkspaceClosed += CloseWarningPopup;
+            dynViewModel.PropertyChanged += DynViewModel_PropertyChanged;
+        }
+
+        private void DynViewModel_PropertyChanged(object sender, System.ComponentModel.PropertyChangedEventArgs e)
+        {
+            if (e.PropertyName == "CurrentSpace" && !((sender as DynamoViewModel).HomeSpaceViewModel as HomeWorkspaceViewModel).IsCurrentSpace)
+            {
+                IsOpen = false;
+            }
```
QilongTang, line 58: Do we need an else case to make it visible again when switching back?

src/DynamoCoreWpf/Views/FileTrust/FileTrustWarning.xaml.cs
```diff
@@ -47,6 +47,15 @@ public FileTrustWarning(DynamoView dynamoViewWindow)
             SetUpPopup();
 
             HomeWorkspaceModel.WorkspaceClosed += CloseWarningPopup;
+            dynViewModel.PropertyChanged += DynViewModel_PropertyChanged;
+        }
+
+        private void DynViewModel_PropertyChanged(object sender, System.ComponentModel.PropertyChangedEventArgs e)
+        {
+            if (e.PropertyName == "CurrentSpace" && !((sender as DynamoViewModel).HomeSpaceViewModel as HomeWorkspaceViewModel).IsCurrentSpace)
```
QilongTang, line 55: hmm why not just watch ViewingHomespace? This seems to be the same check

src/DynamoCoreWpf/ViewModels/Core/HomeWorkspaceViewModel.cs
```diff
@@ -128,10 +128,18 @@ public HomeWorkspaceViewModel(HomeWorkspaceModel model, DynamoViewModel dynamoVi
...
-
+            dynamoViewModel.PropertyChanged += DynamoViewModel_PropertyChanged;
             SetupFooterNotificationItems();
         }
 
+        private void DynamoViewModel_PropertyChanged(object sender, PropertyChangedEventArgs e)
+        {
+            if (e.PropertyName == "CurrentSpace" && !((sender as DynamoViewModel).HomeSpaceViewModel as HomeWorkspaceViewModel).IsCurrentSpace)
+            {
+                ClearWarning();
```
QilongTang, line 139: Can you explain to me again why we need to do this? Wouldn't warning be cleared forever after user switching to custom node workspace?

jesusalvino: That's true for the switching scenario. However the reason why we need that is to hide it immediately the custom node workspace is created (it remains visible from the HomeWorkspace)

src/DynamoCoreWpf/Views/FileTrust/FileTrustWarning.xaml.cs
jesusalvino: No because the HomeworkSpace do that here UpdateRunStatusMsgBasedOnStates , it looks like the trust ui was focused only with one Workspace.

src/DynamoCoreWpf/ViewModels/Core/HomeWorkspaceViewModel.cs
QilongTang: @jesusalvino if user switch back from custom node workspace to homeworkspace, would warnings show up again?

QilongTang (commented):
Looks much cleaner, LGTM once the remaining comments addressed

src/DynamoCoreWpf/Views/FileTrust/FileTrustWarning.xaml.cs
jesusalvino: Done

src/DynamoCoreWpf/ViewModels/Core/HomeWorkspaceViewModel.cs
jesusalvino: Done, also in the FileTrustWarning

src/DynamoCoreWpf/Views/FileTrust/FileTrustWarning.xaml.cs
jesusalvino: Done

src/DynamoCoreWpf/Controls/NotificationsControl.xaml.cs
jesusalvino: Done

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
jesusalvino: Done

src/DynamoCoreWpf/Views/FileTrust/FileTrustWarning.xaml.cs
jesusalvino: Done

src/DynamoCoreWpf/ViewModels/Core/HomeWorkspaceViewModel.cs
jesusalvino: Warnings show up again in the HomeworkSpace only if is required, it's already handled by the HomeworkSpace in fact. For the reported scenario, yes.

jesusalvino:
> Looks much cleaner, LGTM once the remaining comments addressed

Done

src/DynamoCoreWpf/ViewModels/Core/HomeWorkspaceViewModel.cs
```diff
@@ -128,10 +128,18 @@ public HomeWorkspaceViewModel(HomeWorkspaceModel model, DynamoViewModel dynamoVi
...
 
             dynamoViewModel.Model.ShutdownStarted += Model_ShutdownStarted;
-
+            dynamoViewModel.PropertyChanged += DynamoViewModel_PropertyChanged;
             SetupFooterNotificationItems();
         }
 
+        private void DynamoViewModel_PropertyChanged(object sender, PropertyChangedEventArgs e)
+        {
+            if (e.PropertyName == "CurrentSpace" && !(sender as DynamoViewModel).ViewingHomespace)
```
mjkkirschner, line 137: use `nameof` to avoid spelling mistakes and make refactoring easier.

jesusalvino: > use `nameof` to avoid spelling mistakes and make refactoring easier.

Done

jesusalvino: @QilongTang Kindly, Could you let me know if we have any pending here please ?

QilongTang (approved):
LGTM

src/DynamoCoreWpf/ViewModels/Core/HomeWorkspaceViewModel.cs
```diff
@@ -128,10 +128,18 @@ public HomeWorkspaceViewModel(HomeWorkspaceModel model, DynamoViewModel dynamoVi
...
 
             dynamoViewModel.Model.ShutdownStarted += Model_ShutdownStarted;
-
+            dynamoViewModel.PropertyChanged += DynamoViewModel_PropertyChanged;
             SetupFooterNotificationItems();
         }
 
+        private void DynamoViewModel_PropertyChanged(object sender, PropertyChangedEventArgs e)
+        {
+            if (e.PropertyName == nameof(DynamoViewModel.CurrentSpace) && !(sender as DynamoViewModel).ViewingHomespace)
```
QilongTang, line 137: I think you can check `PropertyName == nameof(DynamoViewModel.ViewingHomespace)` directly but this also works

QilongTang:
Waiting for PR checks and also checking @jesusalvino is the last gif still valid?

jesusalvino:
> Waiting for PR checks and also checking @jesusalvino is the last gif still valid?

yes, the last gif is valid since the behavior is the same with the refactoring

QilongTang:
passed at https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/1043

PR #13141, opened 2022-07-27
Implement compile time indexing for arrays

mjkkirschner:
### Purpose

This PR attempts to compile simple index expressions into MSIL opcodes.

logic is as follows:

1. Introduces a new compile phase - `GatherTypeInfo` - when this phase is active, we gather all variable type info as well as type info for every AST node we can - this is useful later in determining if we have enough type info to emit indexing operations. I'm sure it will come in handy in other places as well. We don't emit any IL while this phase is active. 
1. When we encounter a function call named `Builtin.ValueAtIndex` we attempt to transform this function call and its arguments into opcodes for loading the arguments onto the stack, and emitting the correct index operation.
2. For arrays we emit `ldelem` - or a variant depending on the type of array, we must know the type of arrayElement to emit the correct opcode.
3. For `IList`, we should try to emit a `callvirt` opcode for `get_Item` - While we do generate these in this PR, in practice this does not work, because `ILists` are produced as part of `replicationLogic` function calls - ⚠️ instead we generate a call to `ValueAtIndexDynamic`.
4. For `IDictionary` - we generate `get_Item` calls using `callvirt` though it's unclear still when we will encounter these types - likely from non replicated calls we make in the future.  https://jira.autodesk.com/browse/DYN-5151 
5. for Builtin.Dictionary we don't generate anything yet - there is a follow up https://jira.autodesk.com/browse/DYN-5153 - but for the moment this works by emitting calls to `ValueAtIndexDynamic` which checks the runtime type and calls `Dictionary.ValueAtKey`.
7. If we fail to emit index code, because we don't know the type to index, or it's an object for example we generate a `ValueAtIndexDynamic` replicated call.
8. I've added some new micro tests and refactored the `microTest` class so it is reusable.

❗ 
I think it's important to note that because of issue 4 that we will emit `IList` indexing code when we really shouldn't, so we currently bail when we encounter an `IList` and instead generate a replicated call to `ValueAtIndexDynamic`

### Followups (some of these are not really related to indexing, but are blockers):
* ~~accessing a string from a string[] - bizarrely produces an output value of `IList`- the IL compiles fine but when attempting to do anything to the IList fails at runtime... ⚠️ (there is a failing test)~~ -fixed as part of this PR.
* ~~somehow address point 3/4 above - essentially, we need more type information at compile time, or we should wrap our replicated call return values in a known return type - as it is now, it's unclear what level of indexing to generate when we encounter an `IList` - Is it a valid `IList`? Or is it an outer wrapper `IList` because this was the result of a function call?~~ partially fixed by @aparajit-pratap - we don't wrap `IList` with another `IList` - but we still lose type info for single items and generic collections etc.
   ~~1. non replicated calls, or simple replicated calls should return type information from the `EmitFunctionCall` method - then we'll know type info for many cases, like dictionary initialization, or range expression initialization.~~ followup filed.
    2. ~~don't wrap lists or other collections in `IList` when replicating (we could wrap them in something less general)- this is less useful, but will still help us index correctly into `IList` results.~~ (done by Aparajit)
    3. other ideas here?
* ~~Fix Dictionary initialization - promote single values to arrays so we can do `x = {"key": "val"}` - not 100% necessary, can call the constructor directly, though even when calling the constructor directly arguments are reduced in some cases.~~ followup filed.
* ~~our Dictionary type does not inherit from Dictionary or have an indexer we can use via MSIL.~~ - followup filed.
* ~~negative indexing - this will require emitting opcodes with branch instructions and testing the index value at runtime.~~ followup filed.
* ~~indexing using an array of indices - there is a failing test for this.~~ followup filed.

### Tasks filed:
* https://jira.autodesk.com/browse/DYN-5151 - emit type information for function calls (don't rely on replicationLogic)
* https://jira.autodesk.com/browse/DYN-5152 - Replicated Function calls wrapping output in IList causes issues with indexing.
* https://jira.autodesk.com/browse/DYN-5153 - builtin.dictionary indexing.
* https://jira.autodesk.com/browse/DYN-5154 - negative indexing.
* https://jira.autodesk.com/browse/DYN-5155 - indexing with an array of indices.
* https://jira.autodesk.com/browse/DYN-5157 - dictionary construction issues - I think this is a general coercion problem.

### Declarations

### Release Notes

### Reviewers

### FYIs


aparajit-pratap:
> In DS the Builtin.Dictionary type is not a Dictionary, it does not have an indexer etc - keys can only be accessed today by function call without modifying this core type.
Can we not call `ValueAtKey` on the dictionary object (`Builtin.Dictionary`) for this? I believe there are no overloads for `ValueAtKey`.

mjkkirschner:
> > In DS the Builtin.Dictionary type is not a Dictionary, it does not have an indexer etc - keys can only be accessed today by function call without modifying this core type.
> 
> Can we not call `ValueAtKey` on the dictionary object (`Builtin.Dictionary`) for this? I don't believe there are any overloads for `ValueAtKey`.
yes, that is an option, but there are blockers
1. because of the above issues outlined in (4) - we will not know that the output type of a dictionary constructor is a dictionary. `not having return type information from function calls - since Dictionary initialization is done via function calls... we end up calling the IList indexing code, and trying to index with a string key... wreaking havoc.`
2. let's assume 1. is resolved somehow, and we can now detect that the object we're indexing is a `Builtin.Dictionary.` - we could then emit a call to `ValueAtKey` - I guess it could be faster than letting the eventual overload logic determine an overload.

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -50,7 +50,7 @@ public static IList ReplicationLogic(IEnumerable<MethodBase> mInfos, IList args,
             {
                 result = mi.Invoke(reducedArgs[0], reducedArgs.Skip(1).ToArray());
             }
-
+            //TODO return lists/arrays without extra wrapping in next PR.
             return new[] { result };
```
aparajit-pratap, line 54: Will this help in some cases (I think you suggested something like this):
```
if (result is IList) return (IList)result;

return new[] { result };
```

aparajit-pratap: It prevents adding an extra level of nesting.

mjkkirschner: yes, I think it resolves a minimal set of cases, like this test which is currently failing:
```
 public void IndexIntoRange_WithIdent()
        {
            var dscode = @"
import(""DesignScriptBuiltin.dll"");
a = 0..100..10; //a is an ILIST because of replication?
b = 5;
c=a[b];";
            var ast = ParserUtils.Parse(dscode).Body;
            Assert.IsNotEmpty(output);
            Assert.AreEqual(40, output.Values.ToList()[2][0]);
        }

```

test/CodeGenILTests/IndexingMicroTests.cs
```diff
@@ -0,0 +1,217 @@
...
+c = 5;
+d=a[b][c];";
+            var ast = ParserUtils.Parse(dscode).Body;
+            var output = codeGen.EmitAndExecute(ast);
+            Assert.IsNotEmpty(output);
+            Assert.AreEqual(4, output.Values.ToList()[3][0]);
+        }
+
+        [Test]
+        [Category("Failure")]//to accomplish this either need to generate a function call, or emit conditional il.
```
aparajit-pratap, line 124: For constant arrays, can you not get the length, and subtract the -ve index from the length to get the actual index?

test/CodeGenILTests/IndexingMicroTests.cs
```diff
@@ -0,0 +1,217 @@
...
+a = [0,1,2,3,4,5,6,7,8,9,10,11,12];
+b = -5;
+c=a[b];";
+            var ast = ParserUtils.Parse(dscode).Body;
+            var output = codeGen.EmitAndExecute(ast);
+            Assert.IsNotEmpty(output);
+            Assert.AreEqual(8, output.Values.ToList()[2][0]);
+        }
+        [Test]
+        [Category("Failure")]//I can't understand this failure, the IL is valid but the loading the element retrieves an IList instead of a string...
```
aparajit-pratap, line 138: Does your IL look like this:
image

test/CodeGenILTests/IndexingMicroTests.cs
mjkkirschner: @aparajit-pratap yes, that is possible if we know all values ahead of time, but if the value is an identifier then we don't currently have the capability to see if it's negative - instead what I propose is to emit MSIL which tests if the value is negative at runtime, this will make indexing cost a few more operations each but we can still avoid the cost of a function call.

Another approach is to use a try/catch and fallback to negative indexing in the catch block.
please see:
https://jira.autodesk.com/browse/DYN-5154

test/CodeGenILTests/IndexingMicroTests.cs
mjkkirschner: yes, I actually think value is being loaded to the stack correctly, but stored in the output dictionary incorrectly somehow. I will try to write a test case that does not involve indexing but still fails.
https://jira.autodesk.com/browse/DYN-5156

test/CodeGenILTests/IndexingMicroTests.cs
```diff
@@ -0,0 +1,217 @@
...
+            var ast = ParserUtils.Parse(dscode).Body;
+            var output = codeGen.EmitAndExecute(ast);
+            Assert.IsNotEmpty(output);
+            Assert.AreEqual(40, output.Values.ToList()[2][0]);
+        }
+        #endregion
+        #region dictionary
+        [Test]
+        [Category("Failure")]//this is failing because key and val are single items and dictionaries are initialized with lists of keys and vals.
+        //once that is resolved it will fail because we don't have enough type info (wrapped by function call logic IList)
```
aparajit-pratap, line 187: I'm not sure I understand this. You're saying the AST comes in as
``` 
a = Dictionary.ByKeysValues(var_keys, var_values); // var_values is a list of a single key, ...
b = var_key; // var_key is a string
c = Dictionary.ValueAtKey(a, b);
```
Assuming we know `a` is a dictionary, and we can find out the type of `b` to be a string, why can't we emit a function call to `Dictionary.ValueAtKey`. Am I missing something?

aparajit-pratap: I understand that currently we don't have a way to know the type of `a`, but what's the other issue?

test/CodeGenILTests/IndexingMicroTests.cs
aparajit-pratap: Can this be solved using your idea of not just caching types in the `variables ` dictionary but also values of the identifiers in it?

mjkkirschner: I think that would help solve the issue partially - but it won't help when the index is returned from a function call - let's say we have a program like:
```
a = [1,2,3,4];
b = getRandomInt(-1,1);
c = a[b];
```
by looking at the types, we will decide that we can compile this index operation, we can't know what the value is, and half the time we'll generate index calls that will throw at runtime.

We can either:
1. implement index IL that checks if the value at runtime is negative 
2. stop supporting negative indexing.
3. don't generate IL index ops unless we index using constants.
4. only perform indexing by using function calls (get rid of this PR)

test/CodeGenILTests/IndexingMicroTests.cs
mjkkirschner: `Dictionary.ByKeysValues(var_keys, var_values);` expects 2 arrays, and when called with a single key/val pair, it throws. I guess the old VM promoted it to array.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -603,6 +607,112 @@ private void EmitArray<T>(Type arrType, IEnumerable<T> items, Action<T, int> ite
...
+            {
+                if (t.IsGenericType)
+                {
+                    return (true, t.GenericTypeArguments[1]);
+                }
+                else
+                {
+                    return (true, typeof(object));
+                }
+            }
```
aparajit-pratap, line 640: Does there need to be a TODO here like `TryEmitIndexingForDictionary()`... before returning?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -603,6 +607,112 @@ private void EmitArray<T>(Type arrType, IEnumerable<T> items, Action<T, int> ite
...
+            return true;
+        }
+
+        private bool TryEmitIndexingForDictionary(AssociativeNode array, AssociativeNode index, Type collectionType, Type listElementType)
+        {
+           //TODO - Dictionaries in DS will always(almost?) be the DS Dictionary wrapper - which is not a Dictionary
+           //we have to either give it an indexer and call that or make a call to ValueAtKey()
+           //if we think we will encounter regular IDictionary here than we can still implement this.
+
+            return false;
```
aparajit-pratap, line 692: If we have an `IDictionary `returned from an arbitrary function call and the next AST is something like:
`Builtin.Get.ValueAtIndex(dict, "key");`, I suppose this could be converted to an opcode that directly indexes into an `IDictionary`, assuming we know that `dict `is an `IDictionary`.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -487,7 +487,7 @@ private void EmitBinaryExpressionNode(AssociativeNode node)
                 EmitOpCode(OpCodes.Ldarg_2);
                 EmitOpCode(OpCodes.Ldstr, lNode.Value);
                 // if t is a single value, wrap it in an array of the single value.
-                if (!typeof(IEnumerable).IsAssignableFrom(t))
+                if (!typeof(IEnumerable).IsAssignableFrom(t) || typeof(string).IsAssignableFrom(t))
```
aparajit-pratap, line 490: So is this when you encounter an AST initializing a string array - an `ExprListNode `of strings?

aparajit-pratap: Or is it when emitting the individual strings in that expression list?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -409,22 +444,29 @@ private void EmitOpCode(OpCode opCode, MethodBase mBase)
...
         private void EmitOpCode(OpCode opCode, double val)
         {
+            if (!EmitEnabled) return;
             ilGen.Emit(opCode, val);
             writer.WriteLine($"{opCode} {val}");
         }
 
         private void EmitOpCode(OpCode opCode, long val)
         {
+            if (!EmitEnabled) return;
```
aparajit-pratap, line 461: Do we really need this additional flag or can we just say `if(compilePass = CompilePass.GatherTypeInfo) return;`?

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: no, this is when we encounter a single string - `system.string` implements `IEnumerable` so it previously did not get wrapped in an outer `IList` - weird things then happen, because the output Dictionary is of type `<String,IList>` - at runtime the CLR in some cases will show that the value is an `IList` and in others it will show that it's a list, very strange things happen when trying to access the underlying object - I guess because it's a reference that it can work until you try to resolve the reference, I would have though it would fail when the CLR jited it.

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: will look into simplifying it

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -603,6 +607,112 @@ private void EmitArray<T>(Type arrType, IEnumerable<T> items, Action<T, int> ite
...
+            if (t == null)
+            {
+                for (int i = 0; i < toPopCount; i++)
+                {
+                    EmitOpCode(OpCodes.Pop);
+                }
+
+                return (false, null);
+            }
+            else if (t == typeof(IDictionary))
```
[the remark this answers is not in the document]
pinzart90: How/Who how would a  `IDictionary` type get here in `t` ?
for IList....I see that EmitFunctionCall returns typeof(IList)...so that I understand.

Shouldn't we check typeof(IDictionary).IsAssignableFrom(t) ?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -251,7 +249,15 @@ private Type EmitIListCoercion<T>()
             EmitOpCode(OpCodes.Ldloc, counterIndex);
 
             // Load array to be coerced.
-            EmitOpCode(OpCodes.Ldloc, localVarIndex);
+            var currentVarIndex = localVarIndex;
+            if (arg is IdentifierNode ident)
+            {
+                currentVarIndex = variables[ident.Value].Item1;
+            }
```
aparajit-pratap, line 256: If arg isn't an Identifier, how will this work? There will be no dictionary entry, nor will `localVarIndex` be correct.

mjkkirschner: yes, thats why I've added the TODO below for the other case where `emitExpressionListNode` is used.

aparajit-pratap: ok, looks good.

aparajit-pratap: So, the reason why you're seeing a regression with that test case in your PR is that the current `localVarIndex` is not pointing the array to be coerced but to something else that has been added to the top of the stack?

mjkkirschner: it's because the `localIndexVar` variable is incorrect when we get to the `emitMSIL` phase - it's already been incremented in the `GatherTypeInfo` phase.

this change avoids both issues though because we lookup the associated index for the specific variable, not just whatever happens to be the last variable we declared.

mjkkirschner: I am still working through similar issues with the operators tests though, so maybe more fixes will be required here.

aparajit-pratap: So with this fix, can you reset the localVarIndex after the GatherTypeInfo compile pass?

mjkkirschner: With the current implementation it will not help - let me explain.
1. in this PR - in the `GatherTypeInfo` phase - we increment `localvarIndex` and declare variables in the variables dictionary.
2. If we declare / increment those variables again, we'll throw exceptions, there is an invariant in the current code that variables should not be redefined.
3. If we do a reset between phases, then the local var index will just always be -1 during`emitIL` phase.

... I guess we could just increment the `localVarIndex` during `emitIL` phase so at least it is correct, but not touch the `variables` dictionary.

do you want me to investigate that so it's a bit safer to use the `localVarIndex` during `emitIL` or alternatively we could add some documentation to this field?

aparajit-pratap: I was under the impression that variables are being declared and added to the dictionary in both phases, and that `localVarIndex` continues incrementing from where it left off in the `emitIL `phase. If I understand correctly, it sounds like now, no variables are declared in the `emitIL` phase any longer and the `variables `dictionary and `localVarIndex` are up-to-date at the end of the `GatherTypeInfo` phase. If that's the case, it would be unsafe to use `localVarIndex` in the `emitIL `phase altogether. We would need to query the `variables` dictionary using the identifier name in all places instead as you've done here.

mjkkirschner: yes, in the current PR is it not safe to use the `localVarIndex` in the `emitIL` phase at all. 
I see a few alternatives to make this less of an issue.

1. try to reset and re-increment the localVarIndex in the `emitIL` phase... it's likely still going to lead to bugs like we saw today though when larger programs start declaring multiple variables. (I think the use of the index from the `variables` dictionary is a safer approach in general).
2. document that the localVarIndex should not be used in `emitIL`
3. wrap localVarIndex in a getter and throw when trying to access in `emitIL`

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -905,6 +963,138 @@ private void EmitArray<T>(Type arrType, IEnumerable<T> items, Action<T, int> ite
...
+                    TryEmitIndexingForIList(fcn.FormalArguments[0], fcn.FormalArguments[1], t, typeof(object));
+                    return (true, typeof(object));
+                }
+            }
+            EmitILComment("NOT ENOUGH TYPE INFO TO EMIT INDEXING");
+            return (false, null);
+        }
+
+
+        private bool TryEmitIndexingForIList(AssociativeNode array, AssociativeNode index, Type collectionType, Type listElementType)
```
sm6srw, line 1049: Looks like these methods always will succeed now and return `true`. Should we drop the `Try` part and make them void?

src/Engine/EmitMSIL/CodeGenIL.cs
aparajit-pratap: I don't like option 1. I'm fine with option 2. 
I wonder if it's even necessary to continue keeping `localVarIndex` as a class member. I guess I'll need to review this PR again.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -100,6 +102,13 @@ public CodeGenIL(IDictionary<string, IList> input, string filePath)
...
 
+            compilePass = CompilePass.GatherTypeInfo;
+            // 5. Traverse AST and use ILGen to gather what type info we can.
+            foreach (var ast in astList)
+            {
+                DfsTraverse(ast);
+            }
+
             compilePass = CompilePass.emitIL;
             // 5. Traverse AST and use ILGen to emit code for Execute method
```
aparajit-pratap, line 113: // 6.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -150,48 +159,53 @@ private Type EmitCoercionCode(Type argType, ParameterInfo param)
...
             {
-                return EmitIListCoercion<long>();
+                return EmitIListCoercion<long>(arg);
             }
             // TODO: Add more coercion cases here.
 
             return argType;
         }
 
         // TODO: Does not work yet (intended to coerce IList to IEnumerable<T>)
```
aparajit-pratap, line 201: Oops, I forgot to delete this comment. It's not true any longer.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -319,7 +341,15 @@ private Type EmitIListCoercion<T>()
             EmitOpCode(OpCodes.Ldloc, counterIndex);
 
             // Load array to be coerced.
-            EmitOpCode(OpCodes.Ldloc, localVarIndex);
+            var currentVarIndex = localVarIndex;
+            if(arg is IdentifierNode ident)
+            {
+            currentVarIndex = variables[ident.Value].Item1;
```
aparajit-pratap, line 347: indentation

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -646,53 +676,72 @@ public Type DfsTraverse(Node n)
                     EmitGroupExpressionNode(node);
                     break;
             }
+            if(compilePass == CompilePass.GatherTypeInfo)
+            {
+                if (aSTTypeInfoMap.ContainsKey(node.ID))
```
aparajit-pratap, line 681: Can we rename this to follow camelcase: `astTypeInfomap`, or simply rename to `typeInfoMap`?

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: done

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -285,8 +308,18 @@ private Type EmitIListCoercion<T>()
         }
 
         // Coerce int/long/double arrays to IEnumerable<T> or IList<T>
-        private Type EmitArrayCoercion<Source, Target>(Type ienumerableParamType)
+        private Type EmitArrayCoercion<Source, Target>(AssociativeNode arg,Type ienumerableParamType)
         {
+            if (compilePass == CompilePass.GatherTypeInfo)
+            {
+                var returnType = typeof(Target[]);
+                if (typeof(IList<Target>).IsAssignableFrom(ienumerableParamType))
```
aparajit-pratap, line 316: I think this should be changed from `IList` to `List` (I made the same mistake earlier) as `IList<T>` is assignable from `T[]`.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -905,6 +977,133 @@ private void EmitArray<T>(Type arrType, IEnumerable<T> items, Action<T, int> ite
...
+            }
+            EmitILComment("NOT ENOUGH TYPE INFO TO EMIT INDEXING");
+            return (false, null);
+        }
+
+
+        private void EmitIndexingForIList(AssociativeNode array, AssociativeNode index, Type collectionType, Type listElementType)
+        {
+            var indexT = DfsTraverse(index);
+            var mi = typeof(IList).GetMethod("get_Item", BindingFlags.Instance | BindingFlags.Public);
```
aparajit-pratap, line 1066: I remember trying this but it not working for me, so I ended up doing this instead:
```
prop = typeof(IList).GetProperties(BindingFlags.Public | BindingFlags.Instance).Where(
                                    p => p.Name == "Item").FirstOrDefault();
            mInfo = prop.GetAccessors().FirstOrDefault();
```

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -939,16 +1168,22 @@ private Type EmitFunctionCallNode(AssociativeNode node)
             EmitOpCode(OpCodes.Call, keygen);
 
             var local = DeclareLocal(typeof(IEnumerable<MethodBase>));
-
-            EmitOpCode(OpCodes.Ldloca, local);
+            //local could be null if we are not emitting currently.
+            if(local != null)
```
aparajit-pratap, line 1172: Can you clarify under what circumstances can this be null?

mjkkirschner: yes, during `GatherTypeInfo` - `DeclareLocal` will return null.

I think we should really remove all use of `ilGen` directly - and wrap all methods, I noticed it's being used in the new coercion methods, if you are okay with it I will wrap those calls in helpers so we can just return from them if the compiler phase is not emitIL?

essentially, it's not good to write anything to the IL stream until we are ready, since we can't remove it.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1064,10 +1064,12 @@ private void EmitArray<T>(Type arrType, IEnumerable<T> items, Action<T, int> ite
         private void EmitIndexingForIList(AssociativeNode array, AssociativeNode index, Type collectionType, Type listElementType)
         {
             var indexT = DfsTraverse(index);
-            var mi = typeof(IList).GetMethod("get_Item", BindingFlags.Instance | BindingFlags.Public);
```
aparajit-pratap: Did `GetMethod` not work? I used `GetProperties` but I was curious if you could get it to work using `GetMethod`.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -21,6 +21,9 @@ public class CodeGenIL : IDisposable
         // True for internal methods (for example operators and unary operators)
         private IDictionary<string, IList> input;
         private IDictionary<string, IList> output;
+        /// <summary>
+        /// counter for local variables, should only be used directly during MethodLookup or GatherTypeInfo phase. It will incorrect during EmitIL phase. 
+        /// </summary>
```
aparajit-pratap, line 26: I think this can only be used in one phase. If it's used in `MethodLookup` phase, we'll be unable to use it in any subsequent phase, correct? Perhaps we should just mention that it's meant for use in `GatherTypeInfo` phase only.

mjkkirschner: okay, yes, that is more clear.

src/Engine/EmitMSIL/CodeGenIL.cs
aparajit-pratap: Ah, ok.

aparajit-pratap (approved):

pinzart90 (approved):
LGTM

PR #13142, opened 2022-07-28
Dyn 5107 webview2 update in guided tours

RobertGlobant20:
### Purpose

I replaced the WebBrowser component used in Guided Tours by WebView2.
The WebBrowserWindow was removed due that the component was inserted directly in the PopupWindow.
Also I noticed that some code for hiding the Guided Tour Popups when Dynamo is in background was producing a crash in the WebView2 component, then I removed the code in GuidesManager.cs.
In PopupWindow.xaml.cs added the code needed for loading the html page in the WebView2 compoonent.

### Declarations

### Release Notes

I replaced the WebBrowser component used in Guided Tours by WebView2.

### Reviewers

@QilongTang


RobertGlobant20:
This is a GIF showing the Package guide using the WebView2 component
PackagesGuideWebView2
.

src/DynamoCoreWpf/Views/GuidedTour/PopupWindow.xaml.cs
```diff
@@ -75,10 +84,13 @@ private void EnsureStandardPopupAlignment()
 
         private void PopupWindow_Closed(object sender, EventArgs e)
         {
-            if (webBrowserWindow != null)
-                webBrowserWindow.IsOpen = false;
+            if (webBrowserComponent != null)
+            {
+                webBrowserComponent.Visibility = Visibility.Collapsed;
```
QilongTang, line 89: @RobertGlobant20 Can you explain when is the `webBrowserComponent` disposed?

src/DynamoCoreWpf/Views/GuidedTour/PopupWindow.xaml.cs
```diff
@@ -89,11 +101,77 @@ private void PopupWindow_Opened(object sender, EventArgs e)
...
+            webBrowserComponent.Width = popupViewModel.Width;
+            //The height is subtracted by a const that sums the height of the header and footer of the popup
+            var heightBottom = bottomGrid.ActualHeight;
+            var heightTitle = titleGrid.ActualHeight;
+            //popupBordersOffSet * 2 because is one offset at the top and the other one at the bottom of the Popup
+            webBrowserComponent.Height = popupViewModel.Height - (heightBottom + heightTitle + popupBordersOffSet * 2);
+            contentGrid.Children.Add(webBrowserComponent);
+            Grid.SetRow(webBrowserComponent, 1);
+
+            LoadWebBrowser(hostControlInfo.HtmlPage);
```
QilongTang, line 122: @RobertGlobant20 Can these code be reused for future tasks, e.g. replacing library view and documentation browser? Just curious

src/DynamoCoreWpf/Views/GuidedTour/PopupWindow.xaml.cs
```diff
@@ -89,11 +101,77 @@ private void PopupWindow_Opened(object sender, EventArgs e)
...
+        /// </summary>
+        /// <param name="htmlPage">Contains filename and resources to be loaded in page</param>
+        private async void LoadWebBrowser(HtmlPage htmlPage)
+        {
+            var bodyHtmlPage = ResourceUtilities.LoadContentFromResources(htmlPage.FileName, GetType().Assembly, false, false);
+
+            bodyHtmlPage = LoadResouces(bodyHtmlPage, htmlPage.Resources);
+            bodyHtmlPage = LoadResourceAndReplaceByKey(bodyHtmlPage, "#fontStyle", mainFontStylePath);
+            await webBrowserComponent.EnsureCoreWebView2Async();
+            webBrowserComponent.NavigateToString(bodyHtmlPage);
```
QilongTang, line 136: These looks like something we can reuse in the future

QilongTang (commented):
A few comments then LGTM

src/DynamoCoreWpf/Views/GuidedTour/PopupWindow.xaml.cs
```diff
@@ -89,11 +101,77 @@ private void PopupWindow_Opened(object sender, EventArgs e)
...
+        }
+
+        /// <summary>
+        /// Loads resource from a dictionary and replaces its key by an embedded file
+        /// </summary>
+        /// <param name="bodyHtmlPage">Html page string</param>
+        /// <param name="resources">Resources to be loaded</param>
+        /// <returns></returns>
+        private string LoadResouces(string bodyHtmlPage, Dictionary<string, string> resources)
```
filipeotero, line 145: Maybe a good thing is to add those methods in the Utils class, so they can be reused. (For example in our notifications extension method). Let me know your opinion about it.

filipeotero (approved):
LGTM

src/DynamoCoreWpf/Views/GuidedTour/PopupWindow.xaml.cs
RobertGlobant20: @QilongTang the WebView2 control is created every time the Popup is opened and disposed when the Popup is Hidden (specifically in the Step.Hide() method that is executed when we click the next or prev Popup button), the problem with Popup.IsOpen = false is that it generates an async call to Popup.OnClose() so if we don't dispose the WebView2 instance before we close the PackageManagerSearch, the WebView2 crashes (in the image below you will see that we set IsOpen = false and then we execute the UIAutomation Steps - inside one of this we are closing PackageManagerSearch). 
image

QilongTang: @RobertGlobant20 Gotcha, thanks!

src/DynamoCoreWpf/Views/GuidedTour/PopupWindow.xaml.cs
RobertGlobant20: The methods LoadWebBrowser, LoadResouces and LoadResourceAndReplaceByKey were moved to the ResourceUtilities class so they can be used in other places, please check the next commit:
https://github.com/DynamoDS/Dynamo/pull/13142/commits/f8c0b4b52c2de77495455df98ec5bdb11fdc496f

src/DynamoCoreWpf/Views/GuidedTour/PopupWindow.xaml.cs
RobertGlobant20: The methods LoadWebBrowser, LoadResouces and LoadResourceAndReplaceByKey were moved to the ResourceUtilities class so they can be used in other places, please check the next commit:
https://github.com/DynamoDS/Dynamo/pull/13142/commits/f8c0b4b52c2de77495455df98ec5bdb11fdc496f

src/DynamoCoreWpf/Views/GuidedTour/PopupWindow.xaml.cs
RobertGlobant20: The methods LoadWebBrowser, LoadResouces and LoadResourceAndReplaceByKey were moved to the ResourceUtilities class so they can be used in other places, please check the next commit:
https://github.com/DynamoDS/Dynamo/pull/13142/commits/f8c0b4b52c2de77495455df98ec5bdb11fdc496f

QilongTang:
Thank you @RobertGlobant20 , I will need to build your fork to understand the performance impact. Although we talked about replacing the browser for Guided Tour being easiest :) My gut feeling is that, it actually impact performance most since other places, the WebView2 component will not be initialized and disposed with such frequency..

RobertGlobant20:
> Thank you @RobertGlobant20 , I will need to build your fork to understand the performance impact. Although we talked about replacing the browser for Guided Tour being easiest :) My gut feeling is that, it actually impact performance most since other places, the WebView2 component will not be initialized and disposed with such frequency..

@QilongTang yes, as you said I was thinking that this should be the easiest one but there was an issue happening when passing from Step7 to Step8 in the Packages guide that took me more time. 
As I mentioned before the WebView2 instance was crashing, the main problem was that the code that I added for hiding the Popups when Dynamo is in background didn't allow me to see the real bug then I was able to see that the real problem was not disposing the WebView2 component in a specific time. We need to dispose the WebView2 instance because when showing the Step7 the Popup.PlacementTarget = PackageManagerSearch but when passing to Step8 we close the Popup trying to open a new one, I guess .NET has a mechanism that tries to call the WebView2.IsVisible() method when the PackageManagerSearch is closed (due that is the Popup.PlacementTarget) but the WebView2 instance was already disposed when we closed the Popup setting the IsOpen = false (async call to the protected Popup.OnClose() method), then that's the reason why it was crashing, after disposing the WebView2 it worked correctly.

Something that we will need to research is what happens when the WebView2 is hidden or collapsed, since I noticed that the CoreWebView2 is disposed automatically (also applies if is inside a Popup and is hidden, the CoreWebView2 instance is disposed), so we need to re-initialized again calling EnsureCoreWebView2Async and reloading the page, so probably in flows in which we are hiding the LibraryView or the DocumentationBrowser we will need to do that.

PR #13174, opened 2022-08-04
DYN-4833-Group-NodesAlignment

RobertGlobant20:
### Purpose

Fixing nodes alignment when they are inside a Group
Before this fix all the  alignment calculations were considering all the nodes selected then also was considering the Group (AnnotationModel), so when aligning to Left, Right, X-Distribute, Y-Distribute the Group was being moved some pixels so the right, left or over X-axis and Y-axis.
My fix was not considering the Group in the selected items so if we are trying to align all the nodes inside the Group, the group location won't change (just if is necessary in some cases like Left or Right alignment).

### Declarations

### Release Notes

Fixing nodes alignment when they are inside a Group

### Reviewers

@QilongTang 

### FYIs

@avidit

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
@@ -1007,7 +1007,7 @@ namespace Dynamo.ViewModels
 
         public double GetSelectionAverageX()
         {
-            return DynamoSelection.Instance.Selection.Where((x) => x is ILocatable)
+            return DynamoSelection.Instance.Selection.Where((x) => !(x is AnnotationModel) && x is ILocatable)
                            .Cast<ILocatable>()
                            .Select((x) => x.CenterX)
                            .Average();
@@ -1015,7 +1015,7 @@ namespace Dynamo.ViewModels
 
         public double GetSelectionAverageY()
         {
-            return DynamoSelection.Instance.Selection.Where((x) => x is ILocatable)
+            return DynamoSelection.Instance.Selection.Where((x) => !(x is AnnotationModel) && x is ILocatable)
                            .Cast<ILocatable>()
                            .Select((x) => x.CenterY)
                            .Average();
@@ -1023,7 +1023,7 @@ namespace Dynamo.ViewModels
 
         public double GetSelectionMinX()
         {
-            return DynamoSelection.Instance.Selection.Where((x) => x is ILocatable)
+            return DynamoSelection.Instance.Selection.Where((x) => !(x is AnnotationModel) && x is ILocatable)
                            .Cast<ILocatable>()
                            .Select((x) => x.X)
                            .Min();
@@ -1031,7 +1031,7 @@ namespace Dynamo.ViewModels
 
         public double GetSelectionMinY()
         {
-            return DynamoSelection.Instance.Selection.Where((x) => x is ILocatable)
+            return DynamoSelection.Instance.Selection.Where((x) => !(x is AnnotationModel) && x is ILocatable)
                            .Cast<ILocatable>()
                            .Select((x) => x.Y)
                            .Min();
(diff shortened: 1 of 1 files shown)


RobertGlobant20:
This is a GIF **before mi fix,** showing Alignment Right and Left
AfterFix_AlignGroupRight_Left

RobertGlobant20:
This is a GIF before my fix showing alignment X and Y Distribute.
AlignGroupDistributeX_Y

QilongTang:
Thanks! @RobertGlobant20 Please also attach a gif.

I do want to check with Sol if these features meant to align everything in the selection, it seems we found that is quite complex and less feasible at this point

RobertGlobant20:
This is a GIF **AFTER** mi fix, showing Alignment Right and Left
AfterFix_AlignGroupRight_Left

RobertGlobant20:
This is a GIF **AFTER** mi fix, showing alignment X and Y Distribute.
AfterFix_AlignGroupDistributeX_Y

RobertGlobant20:
> Thanks! @RobertGlobant20 Please also attach a gif.
> 
> I do want to check with @Amoursol if these features meant to align everything in the selection, it seems we found that is quite complex and less feasible at this point

@QilongTang also please consider that the main problem was that the group was jumping to left, right or up after the alignment and after this fix this shouldn't be happening.

Amoursol:
@QilongTang @RobertGlobant20 Removing the group from the Selection of what to align is perfect as the group itself should be **located based on the contents**, not controlling the location :) 

It's hard to see the differences in the GIF's... what exactly is the difference? If you have a Group selected as part of the alignment, it is ignored now?

QilongTang:
> @QilongTang @RobertGlobant20 Removing the group from the Selection of what to align is perfect as the group itself should be **located based on the contents**, not controlling the location :)
> 
> It's hard to see the differences in the GIF's... what exactly is the difference? If you have a Group selected as part of the alignment, it is ignored now?

Yes, previously the group will jump ahead. I think that is a bug not mentioned by GG but @RobertGlobant20 decided to fix in addition to the position of nodes.

QilongTang (approved):

QilongTang:
Before merging, I think the gifs looks good. I also want to confirm with you @RobertGlobant20 are those bugs only happening when nodes are inside the groups so group selection become a main factor of the bug?

RobertGlobant20:
> @QilongTang @RobertGlobant20 Removing the group from the Selection of what to align is perfect as the group itself should be **located based on the contents**, not controlling the location :)
> 
> It's hard to see the differences in the GIF's... what exactly is the difference? If you have a Group selected as part of the alignment, it is ignored now?

@Amoursol the main idea is when aligning the nodes (by selecting one by one or by selecting the group) the alignment behavior should be the same (the group should not be jumping).

RobertGlobant20:
> > @QilongTang @RobertGlobant20 Removing the group from the Selection of what to align is perfect as the group itself should be **located based on the contents**, not controlling the location :)
> > It's hard to see the differences in the GIF's... what exactly is the difference? If you have a Group selected as part of the alignment, it is ignored now?
> 
> Yes, previously the group will jump ahead. I think that is a bug not mentioned by GG but @RobertGlobant20 decided to fix in addition to the position of nodes.

@QilongTang I think the jumping ahead bug is mentioned by GG in the Jira task but is not very clear in the description (I think both issues are related the jumping ahead and the nodes position).

QilongTang:
@RobertGlobant20 Cool, glad to see this feature works and nodes are not overlapping.

RobertGlobant20:
> Before merging, I think the gifs looks good. I also want to confirm with you @RobertGlobant20 are those bugs only happening when nodes are inside the groups so group selection become a main factor of the bug?

@QilongTang yes, the problem is happening only when nodes are inside the group (because it was considering the group in the calculation of the span and spacing of each node), when the nodes are out of the group there is no problem (the calculation methods used are the same).

QilongTang:
Passed at https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/1044

PR #13183, opened 2022-08-08
Fixes to support coercion of temporary arrays and ILists

aparajit-pratap:
### Purpose
Fixes to support coercion of temporary arrays and ILists

### Declarations

### Release Notes

### FYIs

### DIFF
--- a/src/Engine/EmitMSIL/CodeGenIL.cs
+++ b/src/Engine/EmitMSIL/CodeGenIL.cs
@@ -200,7 +200,7 @@ namespace EmitMSIL
             }
             else
             {
-                localBuilder = DeclareLocal(typeof(IList), "array to coerce");
+                localBuilder = DeclareLocal(typeof(IList), "IList to coerce");
                 currentVarIndex = localBuilder.LocalIndex;
 
                 EmitOpCode(OpCodes.Stloc, currentVarIndex);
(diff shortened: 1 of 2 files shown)


mjkkirschner (approved):

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -191,6 +191,22 @@ private Type EmitCoercionCode(AssociativeNode arg, Type argType, ParameterInfo p
...
+            {
+                currentVarIndex = variables[ident.Value].Item1;
+            }
+            else
+            {
+                localBuilder = DeclareLocal(typeof(IList), "IList to coerce");
+                currentVarIndex = localBuilder.LocalIndex;
+
+                EmitOpCode(OpCodes.Stloc, currentVarIndex);
+                EmitOpCode(OpCodes.Ldloc, currentVarIndex);
```
pinzart90, line 207: do we really need to do a pop(stloc) then push(ldloc) ? seems redundant...but maybe that is how it's done?

pinzart90 (approved):
One question then LGTM

src/Engine/EmitMSIL/CodeGenIL.cs
aparajit-pratap: Yes, we do, since we need to store the value at the top of the stack in the local var at `currentVarIndex`, so it can be loaded from that index later when we need it.

PR #13185, opened 2022-08-08
Replication logic

pinzart90:
Original PR - which was reverted due to failures - https://github.com/DynamoDS/Dynamo/pull/13125#issue-1309521124

This PR implements replication logic (part of it anyway)
Most of the replication functions are copied from the old VM. The new functions are mostly located right under the original ones.

New structures introduced:

CLRStackValue => equivalent to StackValue
Wraps the CLR objects so that we can store useful metadata (like rank, ProtoCore.Type info etc)
CLRFunctionEndpoint => equivalent to FunctionEndpoint
Wraps the MethodBase objects and stores useful metadata (like rank, ProtoCore.Type info etc)
MSILRuntimeCore - equivalent to RuntimeCore
Holds type information like ClassTable. Makes it easier to mimic the old VM functions
Marshaler changes:
I introduced new virtual methods (that can handle CLRStackValues) to the existing Marshaler classes ...so there should be no breaking changes.

The deferred work is marked in code with TODO_MSIL. I will create jira tasks once we finalize the PR
TODO : Add task to support int32 data types.


src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1425,13 +1428,20 @@ private Type EmitRangeExprNode(AssociativeNode node)
                     op = new IntNode(-1);
                     break;
             }
+
+            bool hasStep = stepNode != null;
```
mjkkirschner, line 1432: this is because `stepNode` always has a value?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1425,13 +1428,20 @@ private Type EmitRangeExprNode(AssociativeNode node)
...
+            // The value of the dummy DoubleNode does not matter since the hasStep boolean will be false.
+            AssociativeNode dummyStepNode = AstFactory.BuildDoubleNode(1);
             var arguments = new List<AssociativeNode>
             {
                 fromNode,
                 toNode,
-                stepNode ?? new NullNode(),
+                // TODO_MSIL: Figure out a better solution for this scenario.
+                // Use DoubleNode(1) because standard replication cannot handle null to value type coerce
+                // The old VM handles builtin functions (like range expr) in a special way...that does not try coercion
```
mjkkirschner, line 1441: EmitFunctionCall could do something special for builtin nodes - as it does when it encounters operators or indexing calls...

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -2,17 +2,116 @@
...
+                    if (arg.IsDefaultArgument)
+                        param = System.Type.Missing;
+                    else
+                    {
+                        param = marshaller.UnMarshal(arg, paramType, runtimeCore);
+                        marshaller.OnDispose(arg);
+                    }
+
+                    /*TODO_MSIL: Figure out how to set/use these flags
+                    if (paraminfos[i].KeepReference && opArg.IsReferenceType)
```
mjkkirschner, line 62: if we can avoid writing our own GC, I think we no longer need this flag.

src/Engine/ProtoCore/Lang/FunctionEndPoint.cs
```diff
@@ -7,6 +8,231 @@
...
+            }
+
+            if (args.Count != FormalParams.Count)
+            {
+                return (int)ProcedureDistance.MaxDistance;
+            }
+
+            int distance = (int)ProcedureDistance.MaxDistance;
+            // Jun Comment:
+            // Default args not provided by the caller would have been pushed by the call instruction as optype = DefaultArs
```
mjkkirschner, line 159: does this comment still apply?

src/Engine/ProtoCore/Lang/Replication/Replicator.cs
```diff
@@ -337,7 +337,52 @@ public static List<StackValue> EstimateReducedParams(List<StackValue> formalPara
 
             return reducedParamTypes;
         }
+        
+        /// <summary>
+        /// Compute the effects of the replication guides on the formal parameter lists
+        /// The results of this loose data, and will not be correct on jagged arrays of hetrogenius types
```
mjkkirschner, line 343: spelling (lose)

src/Engine/ProtoCore/Lang/Type.cs
```diff
@@ -130,6 +135,9 @@ public class TypeSystem
             { typeof(char).FullName, "char" }
         };
 
+        private static TypeSystem instance;
+        internal static TypeSystem Instance => instance != null ? instance : new TypeSystem();
```
mjkkirschner, line 139: maybe will cause issues with tests

test/CodeGenILTests/ReplicationTests.cs
```diff
@@ -0,0 +1,255 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CodeGenILTests
+{
+    class ReplicationTests
```
mjkkirschner, line 14: you should be able to inherit from microTest and you'll get the setup / teardown for free.

src/Engine/EmitMSIL/CodeGenIL.cs
pinzart90: I will add a task for this case

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
```diff
@@ -1413,4 +1919,23 @@ public int GetHashCode(StackValue obj)
...
+        {
+            return object.ReferenceEquals(x.Value, y.Value) && x.TypeUID == y.TypeUID;
+        }
+
+        public int GetHashCode(CLRStackValue obj)
+        {
+            unchecked
+            {
+                var hash = 0;
+                hash = (hash * 397) ^ obj.Value?.GetHashCode() ?? 0;
```
pinzart90, line 1935: Always return hash `0` for null values ..

pinzart90: I hope that does not clash with another hash code...

src/Engine/ProtoCore/Lang/FunctionEndPoint.cs
pinzart90: I did not try to adapt the old functions to the new VM except for the bare minimum.
We can adjust comments and redundancies later on

src/Engine/ProtoCore/Lang/Replication/Replicator.cs
pinzart90: copy pasted form the old VM (I only tried to adapt the bare minimum)
We can do spell checks later (There are definitely not the only ones)

src/Engine/ProtoCore/Lang/Type.cs
pinzart90: removed

test/CodeGenILTests/ReplicationTests.cs
pinzart90: done

pinzart90:
@mjkkirschner @aparajit-pratap FYI
Since the revert:
I implemented a couple more replication functions (bare minimum) so that the existing tests would pass

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1172,11 +1172,11 @@ private Type EmitFunctionCallNode(AssociativeNode node)
...
                     }
                     //if we fail to emit indexing at compile time, 
-                    //emit a function call to ValueAtIndexDynamic to avoid overload issues.
+                    //emit a function call to ValueAtIndex.
                     else
                     {
-                        methodName = nameof(DesignScript.Builtin.Get.ValueAtIndexDynamic);
-                        EmitILComment("NOT ENOUGH TYPE INFO TO EMIT INDEXING, EMIT VALUEATINDEXDYNAMIC FUNCTION CALL");
+                        methodName = nameof(DesignScript.Builtin.Get.ValueAtIndex);
+                        EmitILComment("NOT ENOUGH TYPE INFO TO EMIT INDEXING, EMIT ValueAtIndex FUNCTION CALL");
```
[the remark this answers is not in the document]
mjkkirschner: hmm, thats okay with me, I hope we can implement compile time function calls without replication soon though.

mjkkirschner (approved):
Hey @pinzart90 - some thoughts on this:

1. In general, I think this is a reasonable approach to get to a state where we can test performance and correctness of many more graphs.
2. I think this introduces a ton of complexity that I was excited to leave behind in the new design - but maybe that was wishful thinking.
3. I think it would be good as part of this work to immediately bring in some .dyn tests into the performance folder so we can take advantage of this work to get metrics for more complex graphs vs the old vm.
4. I am curious for your thoughts on where to go next with replication specifically. For example, now that we have similar runtime replication functionality as in the old VM - does it make sense to try shrinking these new classes, tearing out anything we can to while maintaining correctness? Or does it make more sense to leave this runtime behavior as is, and focus instead on putting as much cleverness as possible into the compiler (compile time replication, compile time direct function calls, specific type info from function calls etc)

pinzart90:
> Hey @pinzart90 - some thoughts on this:
> 
> 1. In general, I think this is a reasonable approach to get to a state where we can test performance and correctness of many more graphs.
> 2. I think this introduces a ton of complexity that I was excited to leave behind in the new design - but maybe that was wishful thinking.
> 3. I think it would be good as part of this work to immediately bring in some .dyn tests into the performance folder so we can take advantage of this work (and @sm6srw's) to get metrics for more complex graphs vs the old vm.
> 4. I am curious for your thoughts on where to go next with replication specifically. For example, now that we have similar runtime replication functionality as in the old VM - does it make sense to try shrinking these new classes, tearing out anything we can while maintaining correctness? Or does it make more sense to leave this runtime behavior as is, and focus instead on putting as much cleverness as possible into the compiler (compile time replication, compile time direct function calls, specific type info from function calls etc)?

My intention (for the short term - until we get performance metrics) was to get replication working as much as possible like the old VM, without too much investment in making it leaner/simpler.
If the new VM does inherently provide simpler logic (like GC), then we can definitely  use that from the start (instead of bringing in the more complicated parts of the old VM). It did not seem to me like the new VM brought anything to the table regarding replication. If I missed something in this category, then let me know and I can take another look.

My thoughts on next steps are:
a. Get the new VM (replication) to as state where it can do as much as possible so that we can compare performance.
b. Do not bring in redundant logic from the old VM (like GC). Use the new VM inherent capabilities as much as possible 
c. Once performance is recorded and we get approval to proceed, then start to think about ways to make replication simpler/leaner and easier to understand (with a close eye on what features we want to keep working)

test/CodeGenILTests/ReplicationTests.cs
```diff
@@ -138,5 +137,119 @@ public void MSIL_ReplicationGuides_BuiltinMethods()
...
+                        }
+                    },
+                    new Object[] {
+                        new Object[] {
+                            new FFITarget.ReplicationTestA() { X = 1, Y = 2, Z = 4 },
+                            new FFITarget.ReplicationTestA() { X = 1, Y = 2, Z = 5 }
+                        },
+                        new Object[] {
+                            new FFITarget.ReplicationTestA() { X = 1, Y = 3, Z = 4 },
+                            new FFITarget.ReplicationTestA() { X = 1, Y = 3, Z = 5 }
```
aparajit-pratap, line 220: I think the result should be:
0,2,4
0,2,5

0,3,4
0,3,5

1,2,4
1,2,5

1,3,4
1,3,5

src/Engine/EmitMSIL/CodeGenIL.cs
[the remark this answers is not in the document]
aparajit-pratap: In that case, can we remove the implementation for `ValueAtIndexDynamic` form `BuiltIn.cs`?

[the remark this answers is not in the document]
aparajit-pratap: @pinzart90, won't we have the same issue with the builtin operators that @sm6srw added in this case, where we are replicating over those operators?

test/CodeGenILTests/ReplicationTests.cs
pinzart90: yes...that looks correct.
I'll take a look

pinzart90: the test assertion function was wrong. Fixed directly in branch

src/Engine/EmitMSIL/CodeGenIL.cs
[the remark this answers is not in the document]
pinzart90: yes we should remove `ValueAtIndexDynamic`.
If @sm6srw 's builtin operators are only supposed to handle rank 0 arguments...then all should be fine. If you pass it a list of args, then it will (and should) replicate.

I did not see any arrays or lists to be handled in those builtin operators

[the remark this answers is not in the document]
aparajit-pratap: Ok, so as long as the parameter is supposed to be a single value and is of type object, it's fine, got it.

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
```diff
@@ -682,7 +682,7 @@ public override object UnMarshal(StackValue dsObject, ProtoCore.Runtime.Context
             // If the target type is an IDictionary, marshal to that type
             if (expectedCLRType == typeof(IDictionary))
             {
-                return ToIDictionary(dsObject, context, dsi, expectedCLRType);
+                return ToIDictionary(primitiveMarshaler.GetDictionary(dsObject), expectedCLRType);
```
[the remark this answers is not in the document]
pinzart90: Just a refactor so that we can share more code with the CLRStackValue

src/Engine/ProtoCore/Lang/FunctionEndPoint.cs
```diff
@@ -231,6 +231,80 @@ internal int ComputeTypeDistance(List<CLRStackValue> args, MSILRuntimeCore runti
...
+                if (FormalParams[i].UID != formalParameters[i].TypeUID)
+                    return false;
+            }
+            return true;
+        }
+
+        public bool DoesPredicateMatch(List<CLRStackValue> formalParameters, List<ReplicationInstruction> replicationInstructions)
+        {
+            return true;
+        }
```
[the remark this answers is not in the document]
pinzart90: we can

PR #13186, opened 2022-08-08
DYN-4959 Performance profiling of current state of single graph execution vs. single graph execution of dynamic assembly

sm6srw:
Please Note:

### Purpose

### Declarations

### Release Notes

### Reviewers

### FYIs


src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -1195,6 +1196,7 @@ public partial class LiveRunner : ILiveRunner, IDisposable
     {
         private IDictionary<string, IList> graphOutput;
         internal bool DSExecutionEngine = true;
+        internal long ExecutionTime = 0;
```
mjkkirschner, line 1199: I think if we make this a timespan it's less possible to get confused about units.

test/DynamoCoreTests/Performance/PerformanceTests.cs
```diff
@@ -85,6 +108,12 @@ private void DoWorkspaceOpenAndCompare(string filePath)
             var wcd2 = new serializationTestUtils.WorkspaceComparisonData(ws2, CurrentDynamoModel.EngineController, dsExecution: false);
 
             serializationTestUtils.CompareWorkspaceModelsMSIL(wcd1, wcd2);
+
+            var newEngineExecutionTime = model.EngineController.ExecutionTime;
+
+            Console.WriteLine("Execution time old Engine={0} ms, new Engine={1} ms", oldEngineExecutionTime, newEngineExecutionTime);
```
mjkkirschner, line 114: does this definitely make it into the test log?

I don't think this is going to give us very stable metrics when run with the parallel test job - this job will run in serial, but other threads on the machine will be running other tests...

I guess we'll have to run this these tests either with the master-15 self serve, locally, or create another test job.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -53,19 +54,23 @@ public CodeGenIL(IDictionary<string, IList> input, string filePath)
             writer = new StreamWriter(filePath);
         }
 
-        public IDictionary<string, IList> Emit(List<AssociativeNode> astList)
+        public (IDictionary<string, IList>, long) Emit(List<AssociativeNode> astList)
         {
             var compileResult = CompileAstToDynamicType(astList, AssemblyBuilderAccess.RunAndSave);
```
mjkkirschner, line 59: so, later we'll wrap this some other timer? But this method now only returns a single result. I guess even though it's public, we're free to change it until this is more than a proof of concept.

test/DynamoCoreTests/Performance/PerformanceTests.cs
sm6srw: I will check if this goes into the log now. Yes, I was more thinking about using these numbers locally now.

sm6srw: Yes, the numbers do show up in the log. I added an example to the PR.

src/Engine/EmitMSIL/CodeGenIL.cs
sm6srw: I will add and use an Internal Property instead.

aparajit-pratap (approved):
@sm6srw I believe we can tweak this performance testing to measure more fine-grained metrics, for example, only compile time of old engine vs. new engine. Thoughts?

sm6srw:
> @sm6srw I believe we can tweak this performance testing to measure more fine-grained metrics, for example, only compile time of old engine vs. new engine. Thoughts?

I know we talked about it and yes, it should not be hard to do. Do you think we should do that now or should I create a follow up task (that we may have already)?

pinzart90:
@sm6srw so we are testing the time taken to compile+execute for the old VM vs just execute of the new VM right ?
Maybe just add a new log for the new VM to include compile time too ?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -70,7 +72,12 @@ public CodeGenIL(IDictionary<string, IList> input, string filePath)
             var t = compileResult.tbuilder.CreateType();
```
pinzart90, line 72: maybe add a new stopwatch that will include `CompileAstToDynamicType` and `mi.Invoke(null, new object[] { null, methodCache, output });` ?
We could have a `CompileAndExecuteTime` separate from the current ExecutionTime

sm6srw:
I added separate measurements for compile and execute for the new engine. I have not dived into trying to get separate numbers for the old engine. Do we need it? I can file that as a separate task.

src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -1680,7 +1680,7 @@ private void CompileAndExecuteForDeltaExecution(List<AssociativeNode> astList)
             CompileAndExecute(dispatchASTList);
             PostExecution();
             timer.Stop();
-            ExecutionTime = timer.Elapsed;
+            CompileAndExecutionTime = (new TimeSpan(0), timer.Elapsed);
```
aparajit-pratap, line 1683: To generate compile time only and execution time only for old engine, we can add another timer in `bool LiveRunner.CompileAndExecute(string code)`, something like:
```
private bool CompileAndExecute(string code)
        {
            // TODO Jun: Revisit all the Compile functions and remove the blockId out argument
// compiler timer start
            bool succeeded = runner.CompileAndGenerateExe(code, runnerCore, new ProtoCore.CompileTime.Context());
// compile timer end
            if (succeeded)
            {
// execution timer start
                Execute(!string.IsNullOrEmpty(code));
// execution timer end
            }
            return succeeded;
        }
```

aparajit-pratap (approved):

mjkkirschner (approved):
looks good, those some of these numbers make think
1 - we need larger test cases.
2- we have some profiling to do.

PR #13195, opened 2022-08-09
CER API

pinzart90:
https://jira.autodesk.com/browse/DYN-4679

Provide a way for host integrators to set the CER (crash report tool) location so that it does not need to be found on disk automatically (which may cause performance issues...or may not be found).

PR changes:
1. Added new debugmode to crash on demand (only on debug builds)
3. Added APIs to set the CER location

Proposed API usage:
1. In process:  
   Implement the new interface `IStartConfigCrashReporter`
   ex
   ```
      public interface IRevitStartConfiguration : IStartConfiguration, IStartConfigCrashReporter
      { }

        public struct RevitStartConfiguration : IRevitStartConfiguration
        {
            public CrashReporterStartupOptions CRStartConfig { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
        }
   ```
   Existing OnRequestsCrashPrompt calls will also try to show the CER if the CER tool is found. Default options are to collect everything.
   
   Some CER options can be controlled by calling `OnRequestsCrashPrompt` with `CrashErrorReportArgs`
   ```
       OnRequestsCrashPrompt(null, new CrashErrorReportArgs() { SendSettingsFile = false });
   ```
4. Out of process:
   Use the cli arguments `cr` (ex. dynamosandbox.exe cr = C:/Program Files/Autodesk/Revit)

### DIFF
--- a/src/DynamoCLI/Program.cs
+++ b/src/DynamoCLI/Program.cs
@@ -92,7 +92,10 @@ namespace DynamoCLI
                 model = Dynamo.Applications.StartupUtils.MakeModel(true, string.Empty, cmdLineArgs.AnalyticsInfo);
             }
 
-            model.CERLocation = cmdLineArgs.CrashReportLocation;
+            if (!string.IsNullOrEmpty(cmdLineArgs.CERLocation))
+            {
+                model.CERLocation = cmdLineArgs.CERLocation;
+            }
 
             model.ShutdownCompleted += (m) => { ShutDown(); };
 
--- a/src/DynamoCLI/Program.cs
+++ b/src/DynamoCLI/Program.cs
@@ -20,6 +20,11 @@ namespace DynamoCLI
                     Dynamo.Logging.Analytics.DisableAnalytics = true;
                 }
 
+                if (!string.IsNullOrEmpty(cmdLineArgs.CrashReportLocation))
+                {
+                    DynamoModel.CrashToolLocation = cmdLineArgs.CrashReportLocation;
+                }
+
                 if (cmdLineArgs.KeepAlive)
                 {
                     var thread = new Thread(() => RunKeepAlive(cmdLineArgs));
--- a/src/DynamoCLI/Program.cs
+++ b/src/DynamoCLI/Program.cs
@@ -20,11 +20,6 @@ namespace DynamoCLI
                     Dynamo.Logging.Analytics.DisableAnalytics = true;
                 }
 
-                if (!string.IsNullOrEmpty(cmdLineArgs.CrashReportLocation))
-                {
-                    DynamoModel.CrashToolLocation = cmdLineArgs.CrashReportLocation;
-                }
-
                 if (cmdLineArgs.KeepAlive)
                 {
                     var thread = new Thread(() => RunKeepAlive(cmdLineArgs));
@@ -96,6 +91,8 @@ namespace DynamoCLI
             {
                 model = Dynamo.Applications.StartupUtils.MakeModel(true, string.Empty, cmdLineArgs.AnalyticsInfo);
             }
+
+            model.CERLocation = cmdLineArgs.CrashReportLocation;
 
             model.ShutdownCompleted += (m) => { ShutDown(); };
 
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -166,6 +166,11 @@ namespace Dynamo.Models
         /// </summary>
         public static bool EnableMigrationLogging { get; set; }
 
+        /// <summary>
+        /// The Autodesk CrashReport tool location on disk (directory that contains the "senddmp.exe")
+        /// </summary>
+        public static string CrashToolLocation;
+
         #endregion
 
         #region public properties
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -482,6 +482,7 @@ namespace Dynamo.Models
         {
         }
 
+        // TODO_Dynamo3.0: Replace the IStartConfiguration with a class or struct instance in order to avoid future breaking changes for every new option added.
         public interface IStartConfiguration
         {
             string Context { get; set; }
@@ -504,6 +505,9 @@ namespace Dynamo.Models
             bool IsHeadless { get; set; }
         }
 
+        /// <summary>
+        /// Options used to customize the CER (crash error reporting) experience.
+        /// </summary>
         public struct CrashReporterStartupOptions
         {
             /// <summary>
@@ -512,6 +516,10 @@ namespace Dynamo.Models
             public string CERLocation { get; set; }
         }
 
+        // Remove this interface in Dynamo3.0 and merge it back into IStartConfiguration.
+        /// <summary>
+        /// Use this interface to set the CER (crash error reporting) tool path. 
+        /// </summary>
         public interface IStartConfigCrashReporter
         {
             /// <summary>
@@ -1293,6 +1301,10 @@ namespace Dynamo.Models
                 PreferenceSettings.RequestUserDataFolder -= pathManager.GetUserDataFolder;
                 PreferenceSettings.MessageLogged -= LogMessage;
             }
+
+#if DEBUG
+            CurrentWorkspace.NodeAdded -= CrashOnDemand.CurrentWorkspace_NodeAdded;
+#endif
 
             LogWarningMessageEvents.LogWarningMessage -= LogWarningMessage;
             foreach (var ws in _workspaces)
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -166,11 +166,6 @@ namespace Dynamo.Models
         /// </summary>
         public static bool EnableMigrationLogging { get; set; }
 
-        /// <summary>
-        /// The Autodesk CrashReport tool location on disk (directory that contains the "senddmp.exe")
-        /// </summary>
-        public static string CrashToolLocation;
-
         #endregion
 
         #region public properties
@@ -502,6 +497,22 @@ namespace Dynamo.Models
             /// No update checks or analytics collection should be done.
             /// </summary>
             bool IsHeadless { get; set; }
+        }
+
+        public struct CrashReporterStartupOptions
+        {
+            /// <summary>
+            /// The Autodesk CrashReport tool location on disk (directory that contains the "senddmp.exe")
+            /// </summary>
+            public string CERLocation { get; set; }
+        }
+
+        public interface IStartConfigCrashReporter
+        {
+            /// <summary>
+            /// CERLocation
+            /// </summary>
+            CrashReporterStartupOptions CRStartConfig { get; set; }
         }
 
         /// <summary>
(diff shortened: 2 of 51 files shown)


src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -166,6 +166,11 @@ public partial class DynamoModel : IDynamoModel, IDisposable, IEngineControllerM
         /// </summary>
         public static bool EnableMigrationLogging { get; set; }
 
+        /// <summary>
+        /// The Autodesk CrashReport tool location on disk (directory that contains the "senddmp.exe")
+        /// </summary>
+        public static string CrashToolLocation;
```
mjkkirschner, line 172: so how would `DynamoRevit` set this?

pinzart90: I will add a public setter and getter
Any other suggestions that may be a better fit for the API ?
Maybe adding it to Preferences or to expand the StartConfiguration ?

mjkkirschner: I think it makes sense in the start configuration, though I know that is a bit of a pain - preferences does not make as much a sense to me as it's not a user preference but an integration one.

src/DynamoCoreWpf/Utilities/CrashReportTool.cs
```diff
@@ -29,21 +29,27 @@ internal CrashReportArgs(DynamoViewModel dynamoViewModel)
...
-            try
+            if (!string.IsNullOrEmpty(model.CERLocation) && 
+                File.Exists(Path.Combine(model.CERLocation, CERExe)))
+            {
+                CERLocation = model.CERLocation;
+            }
+            else
             {
-                CerToolDir = GetCERToolLocation();
+                CERLocation = FindCERToolInInstallLocations();
```
mjkkirschner, line 41: this won't ever throw?

src/DynamoCoreWpf/Utilities/CrashReportTool.cs
```diff
@@ -170,20 +176,15 @@ private static string GetCERToolLocation()
...
         internal static void OnCrashReportWindow(CrashReportArgs args)
         {
             DynamoModel model = args.viewModel?.Model;
-            if (string.IsNullOrEmpty(CerToolDir))
+
+            var cerTool = Path.Combine(CERLocation, CERExe);
+            if (string.IsNullOrEmpty(cerTool) || !File.Exists(cerTool))
             {
-                model?.Logger?.LogError($"CERLocation was not provided in the DynamoCore.dll.config");
+                model?.Logger?.LogError($"The CER tool was not found");
```
mjkkirschner, line 187: worth logging what the cerTool path was for debug purposes?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -1305,6 +1305,13 @@ private void Selection_CollectionChanged(object sender, NotifyCollectionChangedE
 
         private void Controller_RequestsCrashPrompt(object sender, CrashPromptArgs args)
         {
+            if (CrashReportTool.ShowCrashErrorReportWindow(
+                (args is CrashErrorReportArgs cerArgs) ? cerArgs : 
+                new CrashErrorReportArgs(dynamoViewModel, null)))
+            {
+                return;
+            }
+            // Backup crash reporting dialog (in case ADSK CER is not found)
```
pinzart90, line 1314: we need to keep the old crash reporting UI because CER may not always exist on the user machine. (sad complication because we were not allowed to bundle CER with Dynamo)

src/DynamoCoreWpf/Utilities/CrashReportTool.cs
```diff
@@ -144,47 +143,64 @@ internal static string CreateMiniDumpFile(string outputDir = null)
...
-        // Calls external CER tool (with UI)
-        internal static void OnCrashReportWindow(CrashReportArgs args)
+        /// <summary>
+        /// Calls external CER tool (with UI)
+        /// </summary>
+        /// <param name="args"></param>
+        /// <returns>True if the CER tool process was successfully started. False otherwise</returns>
+        internal static bool ShowCrashErrorReportWindow(CrashErrorReportArgs args)
         {
+            if (DynamoModel.FeatureFlags?.CheckFeatureFlag("CER", false) == false)
```
pinzart90, line 189: still under feature flag ?
@Amoursol

src/DynamoCoreWpf/Utilities/CrashReportTool.cs
```diff
@@ -12,38 +13,36 @@
...
+        public CrashErrorReportArgs(DynamoViewModel viewModel, Exception e) : base(e) 
         {
-            viewModel = dynamoViewModel;
+            this.viewModel = viewModel;
         }
+
+        public bool SendLogFile { get; set; } = true;
+        public bool SendSettingsFile { get; set; } = true;
+        public bool SendDynFile { get; set; } = true;
+        public bool SendRecordedCommands { get; set; } = true;
```
pinzart90, line 35: can we collect all these by default with CER ? 
Is this more of a legal question ? I ca bring this up
FYI @Amoursol

src/DynamoCLI/Program.cs
```diff
@@ -92,6 +92,11 @@ private static DynamoModel StartupDynamo(StartupUtils.CommandLineArguments cmdLi
                 model = Dynamo.Applications.StartupUtils.MakeModel(true, string.Empty, cmdLineArgs.AnalyticsInfo);
             }
 
+            if (!string.IsNullOrEmpty(cmdLineArgs.CERLocation))
+            {
+                model.CERLocation = cmdLineArgs.CERLocation;
```
mjkkirschner, line 97: so this will override whatever was passed in the startup config?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -499,6 +504,22 @@ public interface IStartConfiguration
...
 
+        public struct CrashReporterStartupOptions
+        {
+            /// <summary>
+            /// The Autodesk CrashReport tool location on disk (directory that contains the "senddmp.exe")
+            /// </summary>
+            public string CERLocation { get; set; }
+        }
+
+        public interface IStartConfigCrashReporter
```
mjkkirschner, line 515: TODO to get rid of this interface at 3.x?

src/DynamoCoreWpf/Utilities/CrashReportTool.cs
```diff
@@ -194,7 +210,13 @@ internal static void OnCrashReportWindow(CrashReportArgs args)
                 var cerDir = Directory.CreateDirectory(Path.Combine(Path.GetTempPath(), "DynamoCER_Report_" +
                     DateTime.Now.ToUniversalTime().ToString("yyyy-MM-dd-HH-mm-ss")));
 
-                using (Scheduler.Disposable.Create(() => { cerDir.Delete(); }) )
+                using (Scheduler.Disposable.Create(() => {
+                    // Cleanup
+                    foreach (FileInfo file in cerDir.EnumerateFiles())
```
mjkkirschner, line 215: these have the potential to throw, maybe wrap in a try catch?

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -1305,6 +1305,13 @@ private void Selection_CollectionChanged(object sender, NotifyCollectionChangedE
 
         private void Controller_RequestsCrashPrompt(object sender, CrashPromptArgs args)
         {
+            if (CrashReportTool.ShowCrashErrorReportWindow(
+                (args is CrashErrorReportArgs cerArgs) ? cerArgs :
```
mjkkirschner, line 1309: what condition is this handling?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -913,6 +939,10 @@ protected DynamoModel(IStartConfiguration config)
                 }
             }
 
+#if DEBUG
+            CurrentWorkspace.NodeAdded += CrashOnDemand.CurrentWorkspace_NodeAdded;
```
mjkkirschner, line 943: I think it would be good to unsubscribe this at shudown/dispose.

test/DynamoCoreTests/Models/DynamoModelEventsTest.cs
```diff
@@ -365,14 +366,20 @@ public void TestOnRequestsCrashPrompt()
...
+            CurrentDynamoModel.OnRequestsCrashPrompt(null, cerArgs);
+            CurrentDynamoModel.OnRequestsCrashPrompt(null, cArgs);
             CurrentDynamoModel.OnRequestsCrashPrompt(this, crashArgs);
 
             //Assert
             //Unsubcribe from event
             CurrentDynamoModel.RequestsCrashPrompt -= CurrentDynamoModel_RequestsCrashPrompt;
             //This will validate that the local handler was executed and set the flag in true
-            Assert.IsTrue(requestsCrashPrompt);
+            Assert.AreEqual(3, requestsCrashPrompt);
```
mjkkirschner, line 382: should the test reset `requestsCrashPrompt`?

mjkkirschner (changes requested):
It looks good Tibi, I think everything is basically good to go, except unsubscribing the new event handler.

src/DynamoCLI/Program.cs
pinzart90: Yes.

src/DynamoCore/Models/DynamoModel.cs
pinzart90: will add

src/DynamoCoreWpf/Utilities/CrashReportTool.cs
pinzart90: It is...but it might be useful to try catch the cleanup separately...since the outer catch will cause the backup (old) crash prompt to be displayed

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
pinzart90: `Controller_RequestsCrashPrompt` can be called with `CrashPromptArgs`(existing API) and with `CrashErrorReportArgs` (new API added). For both cases we want to call into CER first (if CER is not found, then we go with the backup crash prompt)
`CrashErrorReportArgs` is derived from `CrashErrorReportArgs` and has a bunch of flags to customize CER.
If `Controller_RequestsCrashPrompt` is called with `CrashErrorReportArgs`, then we pass it to CER.
If `Controller_RequestsCrashPrompt` is called with `CrashPromptArgs`, then we call CER with a new `CrashErrorReportArgs` (defaults to send everything it can to CER), basically ignoring the `CrashPromptArgs` args.

I can add some of this info to the APIs

test/DynamoCoreTests/Models/DynamoModelEventsTest.cs
pinzart90: teardown / setup should already be doing cleanup

src/DynamoCore/Models/DynamoModel.cs
pinzart90: done

mjkkirschner (approved):

src/DynamoCoreWpf/Utilities/CrashReportTool.cs
Amoursol: @pinzart we will need to check with Legal about any of these things (Unless it was correct as per Standup yesterday with Jackie). Assume that if users opt into CER, that these will be covered.

PR #13196, opened 2022-08-09
[POC] Embed notification backend URL in notification web-app

zeusongit:
### Purpose

This is a proof of concept demonstrating one way of embedding notification service URL in web-app.
We are using the same technique as loading the javascript in WebView2 handler - string replacement to embed the URL.
The URL is read from an application config file.
```
<?xml version="1.0"?>
<configuration>
  <appSettings>
    <add key="notificationAddress" value="https://d2xm4bcf3at21r.cloudfront.net/dynNotifications.json"/>
  </appSettings>
<startup><supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.8"/></startup></configuration>
```

The app reads the key `notificationAddress` and replaces the URL hard-coded in the web-app.

We can update the mock URL in the web-app to any other string.

UPDATE:
With some help from @filipeotero I was able to re-render notifications after url update.
The notifications are fetched from https://ddehnr4ewobxc.cloudfront.net/dynNotifications.json
DynamoSandbox_M9IXz2u1C4

### Declarations

### Release Notes

- Add config file to provide notification URL

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoUtilities/PathHelper.cs
+++ b/src/DynamoUtilities/PathHelper.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Configuration;
 using System.IO;
 using System.Linq;
 using System.Security.AccessControl;
@@ -389,6 +390,33 @@ namespace DynamoUtilities
             string directoryPath = FormatDirectoryPath(directory);
             
             return subdirPath.StartsWith(directoryPath, StringComparison.OrdinalIgnoreCase);
+        }
+
+        /// <summary>
+        /// Returns the path configured for package manager to retrieve packages from 
+        /// as defined inside config file
+        /// </summary>
+        /// <returns>Path that will be used to fetch packages</returns>
+        public static string getServiceBackendAddress(object o, string serviceKey)
+        {
+            string url = null;
+            if (o != null)
+            {
+                var path = o.GetType().Assembly.Location;
+                var config = ConfigurationManager.OpenExeConfiguration(path);
+                var key = config.AppSettings.Settings[serviceKey];
+
+                if (key != null)
+                {
+                    url = key.Value;
+                }
+
+                if (!Uri.IsWellFormedUriString(url, UriKind.Absolute))
+                {
+                    throw new ArgumentException("Incorrectly formatted URL provided for the service.", "url");
+                }
+            }
+            return url;
         }
     }
 }
--- a/src/DynamoUtilities/PathHelper.cs
+++ b/src/DynamoUtilities/PathHelper.cs
@@ -396,6 +396,9 @@ namespace DynamoUtilities
         /// Returns the path configured for package manager to retrieve packages from 
         /// as defined inside config file
         /// </summary>
+        /// <param name="o">The "this" object from where the function is being called from.</param>
+        /// <param name="serviceKey">Service or feature for which the address is being requested. 
+        /// It should match the key specified in the config file.</param>
         /// <returns>Path that will be used to fetch packages</returns>
         public static string getServiceBackendAddress(object o, string serviceKey)
         {
--- a/src/Notifications/NotificationCenterController.cs
+++ b/src/Notifications/NotificationCenterController.cs
@@ -23,9 +23,9 @@ namespace Dynamo.Notifications
         private static readonly string htmlEmbeddedFile = "Dynamo.Notifications.node_modules._dynamods.notifications_center.build.index.html";
         private static readonly string jsEmbeddedFile = "Dynamo.Notifications.node_modules._dynamods.notifications_center.build.index.bundle.js";
 
-        private event Action<ILogMessage> MessageLogged;
+        private DynamoLogger logger;
 
-        internal NotificationCenterController(DynamoView dynamoView)
+        internal NotificationCenterController(DynamoView dynamoView, DynamoLogger dynLogger)
         {
             var shortcutBar = dynamoView.ShortcutBar;
             var notificationsButton = (Button)shortcutBar.FindName("notificationsButton");
@@ -45,7 +45,8 @@ namespace Dynamo.Notifications
             this.notificationsButton.Click += NotificationsButton_Click;
 
             notificationUIPopup.webView.EnsureCoreWebView2Async();
-            notificationUIPopup.webView.CoreWebView2InitializationCompleted += WebView_CoreWebView2InitializationCompleted;            
+            notificationUIPopup.webView.CoreWebView2InitializationCompleted += WebView_CoreWebView2InitializationCompleted;
+            logger = dynLogger;
         }
 
         private void WebView_CoreWebView2InitializationCompleted(object sender, Microsoft.Web.WebView2.Core.CoreWebView2InitializationCompletedEventArgs e)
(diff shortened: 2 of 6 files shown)


src/Notifications/NotificationCenterController.cs
```diff
@@ -21,6 +23,8 @@ public class NotificationCenterController
         private static readonly string htmlEmbeddedFile = "Dynamo.Notifications.node_modules._dynamods.notifications_center.build.index.html";
         private static readonly string jsEmbeddedFile = "Dynamo.Notifications.node_modules._dynamods.notifications_center.build.index.bundle.js";
 
+        private event Action<ILogMessage> MessageLogged;
```
mjkkirschner, line 26: not sure this will do anything until you 
1. implement ILogSource and
2. I think Dynamo only pays attention to `ILogSource` objects if they are extensions - so you could relay these messages through the notification extension and make it implement `ILogSource` - that should get hooked up automatically since it's a ViewExtension.

src/Notifications/NotificationCenterController.cs
```diff
@@ -55,10 +59,14 @@ private void WebView_CoreWebView2InitializationCompleted(object sender, Microsof
...
             }
 
+            //Fetch notification URL from config
+            string notificationURL = FetchNotificationURL();
+
             using (Stream stream = assembly.GetManifestResourceStream(jsEmbeddedFile))
             using (StreamReader reader = new StreamReader(stream))
             {
                 var jsString = reader.ReadToEnd();
+                jsString = jsString.Replace("http://demo9540080.mockable.io/notifications", notificationURL);
```
QilongTang, line 69: This is fine for POC, I would love to maybe consider adding a config on the WebApp side as well to record the end point value so Dynamo is just overwriting that config instead of string replacement

src/Notifications/NotificationCenterController.cs
```diff
@@ -91,6 +99,35 @@ private void NotificationsButton_Click(object sender, RoutedEventArgs e)
             notificationUIPopup.IsOpen = !notificationUIPopup.IsOpen;
             if (notificationUIPopup.IsOpen)
                 notificationUIPopup.webView.Focus();
-        }        
+        }
+
+        private string FetchNotificationURL()
```
QilongTang, line 104: Is there a similar function on package manager side? If there is, I expect us to re-use the function. Maybe we can put this as a shared utility function

QilongTang (commented):
Overall I think this way of deployment is fine so we can maybe fine tuning this approach to make it more mature. Curious if rest of team think the same.

src/Notifications/NotificationCenterController.cs
zeusongit: There isn't I created this, but I can try to put it in utility and use it on both locations.

mjkkirschner:
I don't understand why we need to do this for a simple string, doesn't webView2 have apis like the WebBrowser control for calling methods on the javascript side with a primitive object like a string? (invoke script etc?)

src/Notifications/NotificationCenterController.cs
QilongTang: If there isn't, would you find out how we read the endpoint value for the package manager? Would love to align

zeusongit: It is the same code: https://github.com/DynamoDS/Dynamo/blob/20851b250a91203bab7275d78bc60b34982a7bcf/src/DynamoPackages/PackageManagerExtension.cs#L123
but without a method

QilongTang: @zeusongit Gotcha 👍🏻

src/Notifications/NotificationCenterController.cs
zeusongit: Do you mean the web-app reads a config file before bundling it? because that is possible.
We can provide the URL to the web-app using environment variables that are read from a config file. Then on build the URL will be imported in the bundled JS file. Am I getting this right?

src/Notifications/NotificationCenterController.cs
zeusongit: Ok so implementing `ILogSource` in notification extension and then relaying the object in controller to log messages then?

src/Notifications/NotificationCenterController.cs
QilongTang: yeah, one thing I am unsure of is whether we want to expose the env or config file to Dynamo project (which means it will be part of the bundle) or like @mjkkirschner suggested, to have an API so we can just call to update the value

filipeotero: If we are going to use the approach, maybe we should set the URL (or the key value) in a const variable.

filipeotero:
I guess this works well, but I think a more concise approach is to define the URL though calling a javascript method that is provided by _ExecuteScriptAsync_. It's more complicated but we will need to do something like it in the future anyways (sending the local notification for example).

QilongTang:
`v0.0.11` publish has been finished so you can validate the change officially locally

zeusongit:
> `v0.0.11` publish has been finished so you can validate the change officially locally

New changes has been validated.

src/DynamoUtilities/PathHelper.cs
```diff
@@ -390,5 +391,32 @@ internal static bool IsSubDirectoryOfDirectory(string subdirectory, string direc
             
             return subdirPath.StartsWith(directoryPath, StringComparison.OrdinalIgnoreCase);
         }
+
+        /// <summary>
+        /// Returns the path configured for package manager to retrieve packages from 
+        /// as defined inside config file
+        /// </summary>
+        /// <returns>Path that will be used to fetch packages</returns>
+        public static string getServiceBackendAddress(object o, string serviceKey)
```
QilongTang, line 400: Would you add comments for params as well?

src/Notifications/NotificationCenterController.cs
```diff
@@ -105,6 +113,27 @@ private void NotificationsButton_Click(object sender, RoutedEventArgs e)
...
+        }
+
+        private async void InvokeJS(string script)
+        {
+            await notificationUIPopup.webView.CoreWebView2.ExecuteScriptAsync(script);
+        }
+
+        /// <summary>
+        /// Invokes the script on the notification web-app side to update the URL to fetch notifications from 
+        /// and that will trigger a re-render of the panel. If a URL us provided then that will be used
```
QilongTang, line 125: `If a URL was provided`?

src/Notifications/NotificationCenterController.cs
```diff
@@ -64,8 +69,11 @@ private void WebView_CoreWebView2InitializationCompleted(object sender, Microsof
                 htmlString = htmlString.Replace("mainJs", jsString);
             }
 
-            if(notificationUIPopup.webView.CoreWebView2 != null)
+            if (notificationUIPopup.webView.CoreWebView2 != null)
+            {
                 notificationUIPopup.webView.CoreWebView2.NavigateToString(htmlString);
+                RefreshNotifications();
```
QilongTang, line 75: Just want to be clear, this currently does not work immediately as you described?

QilongTang (commented):
A couple comments then LGTM, The reported failure seems to be a system fluke.

src/Notifications/NotificationCenterController.cs
zeusongit: Nope, after the meeting @filipeotero helped in getting it to work, so it does re-renders now.

QilongTang: Thank you both for looking into this and getting it to work 👍🏻

QilongTang (approved):

PR #13219, opened 2022-08-15
don't initialize the direct manipulation extension when asm is not loaded

mjkkirschner:
### Purpose
Dynamo crashes when LibG is not loaded and a direct manipulator is created.
https://jira.autodesk.com/browse/DYN-5023

* check that node output is not null or a function - if so, just return before trying to create any direct manipulators.
* inside of the direct manipulation extension startup code, just return if `IsASMLoaded` is false.
* Note that I do think this `IsASMLoaded` is accurate in all situations, it seems to really only be set in CLI or sandbox startup - but thats probably sufficient since in-process hosts preload ASM themselves in most cases. 

Screen Shot 2022-08-15 at 3 00 59 PM

I will send another PR with a test - want to get this in first.

### Declarations

### Release Notes

Fixed crash when node was moved which was going to create direct manipulator.

### DIFF
--- a/src/DynamoManipulation/DynamoManipulationExtension.cs
+++ b/src/DynamoManipulation/DynamoManipulationExtension.cs
@@ -8,6 +8,7 @@ using Dynamo.Extensions;
 using Dynamo.Graph.Nodes;
 using Dynamo.Graph.Nodes.ZeroTouch;
 using Dynamo.Graph.Workspaces;
+using Dynamo.Logging;
 using Dynamo.Models;
 using Dynamo.Visualization;
 using Dynamo.Wpf.Extensions;
@@ -16,7 +17,7 @@ using DoubleSlider = CoreNodeModels.Input.DoubleSlider;
 
 namespace Dynamo.Manipulation
 {
-    public class DynamoManipulationExtension : IViewExtension
+    public class DynamoManipulationExtension : IViewExtension,ILogSource
     {
         private IEnumerable<NodeModel> manipulatorNodes;
         private ManipulatorDaemon manipulatorDaemon;
@@ -24,6 +25,8 @@ namespace Dynamo.Manipulation
         private IWorkspaceModel workspaceModel;
         //Keeps track of Node Guids for which event handler is attached.
         private HashSet<Guid> trackedNodeGuids = new HashSet<Guid>();
+
+        public event Action<ILogMessage> MessageLogged;
 
         internal IWatch3DViewModel BackgroundPreviewViewModel { get; private set; }
         internal IRenderPackageFactory RenderPackageFactory { get; private set; }
@@ -118,18 +121,28 @@ namespace Dynamo.Manipulation
 
         public void Dispose()
         {
-            manipulatorDaemon.KillAll();
+            manipulatorDaemon?.KillAll();
 
             UnregisterEventHandlers();
         }
 
         public void Startup(ViewStartupParams viewStartupParams)
         {
+            if (!viewStartupParams.IsGeometryLibraryLoaded)
+            {
+                MessageLogged(LogMessage.Info($"GeometryLibrary was not loaded, skipping {nameof(DynamoManipulationExtension)}"));
+                return;
+            };
             manipulatorDaemon = ManipulatorDaemon.Create(new NodeManipulatorFactoryLoader());
         }
 
         public void Loaded(ViewLoadedParams viewLoadedParams)
         {
+            if (!viewLoadedParams.StartupParams.IsGeometryLibraryLoaded)
+            {
+                MessageLogged(LogMessage.Info($"GeometryLibrary was not loaded, skipping {nameof(DynamoManipulationExtension)}"));
+                return;
+            };
             viewParams = viewLoadedParams;
 
             WorkspaceModel = viewParams.CurrentWorkspaceModel;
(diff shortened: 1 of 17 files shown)


src/DynamoManipulation/NodeManipulator.cs
```diff
@@ -498,15 +498,24 @@ private RenderPackageCache GenerateRenderPackages()
             // and we must return immediately before proceeding with further calls to ProtoGeometry
             if (IsNodeNull(Node.CachedValue)) return packages;
 
-            AssignInputNodes();
-            
-            active = UpdatePosition();
-
-            if (!IsEnabled())
+            //it's possible the the node's return value is not null, but libG was not loaded correclty
+            //it could be a function, null check is not sufficent, wrap use of libG in a try catch.
```
aparajit-pratap, line 502: Instead of a try-catch, can we add a check for the node being a function inside `IsEnabled `and return false if that's the case?

aparajit-pratap: `NodeModel.CachedValue.IsFunction`

mjkkirschner: no, I tried that already, it appears the direct manipulator should be created when the node returns a function so if we do that, direct manipulation will not work at all.

mjkkirschner: I guess my comment is not sufficient if it did not explain the issue, I will try to add more detail there,

sm6srw (approved):
LGTM!

src/DynamoManipulation/NodeManipulator.cs
aparajit-pratap: Direct manipulation is supposed to work for a node that is not connected if it's input port has a default arg or if it's connected to a slider. I'm not sure if it's supposed to work if it's a function.

pinzart90 (approved):

src/DynamoManipulation/NodeManipulator.cs
```diff
@@ -494,19 +494,31 @@ private RenderPackageCache GenerateRenderPackages()
                 return packages;
             }
 
-            // This check is required as for some reason LibG fails to load, geometry nodes are null
-            // and we must return immediately before proceeding with further calls to ProtoGeometry
+            // This check is required if for some reason LibG fails to load, geometry nodes are null
+            // and we must return immediately before proceeding with further calls to ProtoGeometry.
+            // It's possible the the node's return value is not null, but libG was still not loaded correclty.
+            // For example we could be returning a function, and checking for null or function is not sufficent here
+            // as direct manipulators should be created for function outputs. Additionally, checking for Function does not
```
aparajit-pratap, line 501: `direct manipulators should be created for function outputs` - I'm not sure I agree with this statement. Can you explain why you think manipulators should be created for function outputs?

aparajit-pratap (commented):
@mjkkirschner can we prevent loading the DM view extension altogether if ASM fails to preload? I think @QilongTang did this for the guided tour, if I'm not wrong.

src/DynamoManipulation/NodeManipulator.cs
```diff
@@ -494,19 +494,31 @@ private RenderPackageCache GenerateRenderPackages()
...
-            // This check is required as for some reason LibG fails to load, geometry nodes are null
-            // and we must return immediately before proceeding with further calls to ProtoGeometry
+            // This check is required if for some reason LibG fails to load, geometry nodes are null
+            // and we must return immediately before proceeding with further calls to ProtoGeometry.
+            // It's possible the the node's return value is not null, but libG was still not loaded correclty.
+            // For example we could be returning a function, and checking for null or function is not sufficent here
+            // as direct manipulators should be created for function outputs. Additionally, checking for Function does not 
+            // indicate that LibG failed to load.
+            // Instead wrap use of libG in a try catch.
             if (IsNodeNull(Node.CachedValue)) return packages;
```
aparajit-pratap, line 505: I think apart from testing for the node to be null, adding the check to see if it's returning a function, and returning if so, will work here, i.e. before even calling `AssignInputNodes`.

src/DynamoManipulation/NodeManipulator.cs
mjkkirschner: @aparajit-pratap let me do some more testing with the function check, I may have mis understood when a manipulator should be created, in which case the function || null check may work.

mjkkirschner:
thanks @aparajit-pratap for correcting my misunderstanding here - I attempted to resolve both suggestions you gave.

mjkkirschner:
tests running: https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/6802/

src/DynamoManipulation/DynamoManipulationExtension.cs
```diff
@@ -118,18 +121,28 @@ public string Name
...
+            {
+                MessageLogged(LogMessage.Info($"GeometryLibrary was not loaded, skipping {nameof(DynamoManipulationExtension)}"));
+                return;
+            };
             manipulatorDaemon = ManipulatorDaemon.Create(new NodeManipulatorFactoryLoader());
         }
 
         public void Loaded(ViewLoadedParams viewLoadedParams)
         {
+            if (!viewLoadedParams.StartupParams.IsGeometryLibraryLoaded)
```
aparajit-pratap, line 141: Do we need the `IsGeometryLibraryLoaded` check in both `Startup `and `Loaded `functions?

aparajit-pratap (approved):

src/DynamoManipulation/DynamoManipulationExtension.cs
mjkkirschner: yes, I'd like to avoid partially initializing as much as possible.

PR #13228, opened 2022-08-18
Don't import sat file as if Dynamo is always in meters.

mjkkirschner:
### Purpose

This PR consumes a new build of LibG and adds new SAT Import nodes that allow specifying how the internal units of the .SAT should be interpreted relative to the "dynamo units" - **If Dynamo had a unit system, what units would it be set to when importing this file**. Units are specified using forge unit types. 
❗ the default - if no unit is supplied is to treat the .sat file as unitless and import the geometry without scaling - likely what the user wants anyway.
- [x] consumes a new build of LibG with new SAT import methods 
- [x] adds new nodes to layout spec and puts them into Geometry category where existing sat import nodes were.
- [x] deprecates existing sat import nodes, but they will open in old graphs.
- [x] Figure out how to rename the node to `Geometry.ImportFromSatByUnits` and lose the class name. (worst case I need to write a nodeModel node).

relies on this PR:
https://git.autodesk.com/Dynamo/LibG/pull/1331

Screen Shot 2022-08-19 at 4 38 42 PM
Screen Shot 2022-08-19 at 4 38 52 PM

### Declarations

### Release Notes

Added new .SAT and .SAB Import nodes that allow better control of how sat units are interpreted. 

### Reviewers

### FYIs


mjkkirschner:
These build failures are expected until we merge the LibG PR and update LibG again.

src/Libraries/GeometryColor/ImportSatWithUnits.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Linq;
+using System.Collections.Generic;
+using System.IO;
+using Autodesk.DesignScript.Runtime;
+
+//TODO at a major host release verion - 2024 for example, rename this assembly to something like
+//ZeroTouchGeometryNodes and use type forward attribute to retain API compat.
+
+//This namespace and class name are a bit confusing - but this is done so that the Nodes show up in the UI as Geometry.NodeName.
+namespace Autodesk.DesignScript.Geometry.ImportHelpers
```
mjkkirschner, line 12: I'm not thrilled with this namespace and class name, but it works so that the Node appears to be in the Geometry category and class when in the UI and it's not brutal when using from c# or a CBN, though there could be issues disambiguating `Autodesk.DesignScript.Geometry.Geometry` from `Autodesk.DesignScript.Geometry.ImportHelpers.Geometry` ... A simpler solution is to just use a different class name here but I was not sure it was acceptable to have the Import nodes change their class name, even if they are still shown in the Geometry category in the library.

If that is acceptable, than I think it's a better, simpler solution for DS/C#/python users and doesn't confuse things too bad for node UI users much.

@aparajit-pratap let me know what you think.

src/Libraries/GeometryColor/ImportSatWithUnits.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// <param name="filePath">string file path to a .SAT file.</param>
+        /// <param name="dynamoUnit">a forge unit length, if left null, sat file will be imported as unitless</param>
+        /// <returns></returns>
+        [AllowRankReduction]
+        public static IEnumerable<DesignScript.Geometry.Geometry> ImportFromSATByUnits(string filePath, [DefaultArgument("null")] DynamoUnits.Unit dynamoUnit)
+        {
+            double mm_per_unit = -1;
+            if (dynamoUnit != null )
+            {
+                const string milimeters = "autodesk.unit.unit:millimeters";
```
aparajit-pratap, line 33: spelling: millimeters

src/Libraries/GeometryColor/ImportSatWithUnits.cs
```diff
@@ -0,0 +1,68 @@
...
+            if (dynamoUnit != null)
+            {
+                const string milimeters = "autodesk.unit.unit:millimeters";
+                var mm = Unit.ByTypeID($"{milimeters}-1.0.1");
+
+                if (!dynamoUnit.ConvertibleUnits.Contains(mm)){
+                    throw new Exception($"{dynamoUnit.Name} was not convertible to mm");
+                }
+                mm_per_unit = DynamoUnits.Utilities.ConvertByUnits(1, dynamoUnit, mm);
+            }
```
aparajit-pratap, line 64: I think we can refactor this common code into a helper function.

src/Libraries/GeometryColor/ImportSatWithUnits.cs
aparajit-pratap: @mjkkirschner this is extra work but how about exposing these nodes as nodemodel nodes like we do for `ExportWithUnits` - https://github.com/DynamoDS/Dynamo/blob/4819d581c7386adafc9558a5b8a74ef502873aaf/src/Libraries/GeometryUI/ExportWithUnits.cs#L15. This way we can make them seamlessly look like they're part of the Geometry class and category in the library.

src/LibraryViewExtensionMSWebBrowser/web/library/layoutSpecs.json
```diff
@@ -1454,11 +1454,14 @@
                       "inclusive": false
                     },
                     {
-                      "path": "GeometryColor.Autodesk.DesignScript.Geometry.ImportHelpers.Geometry",
-                      "inclusive": false
+                      "path": "Geometry.Geometry.ExportToSAT"
                     },
                     {
-                      "path": "Geometry.Geometry.ExportToSAT"
+                      "path": "Geometry.Geometry.Geometry.ImportFromSatAndUnits"
```
aparajit-pratap, line 1460: Why does the export node `Geometry.Geometry.Export`, whereas the new node is `Geometry.Geometry.Geometry...`

src/Libraries/GeometryUI/ExportWithUnits.cs
```diff
@@ -1,17 +1,103 @@
...
+using System.Xml;
 
 namespace GeometryUI
 {
+
+    //TODO move to own importwithunits file.
+    //TODO finish the descriptions etc with resx.
+
+    [NodeCategory(BuiltinNodeCategories.GEOMETRY)]
+    [NodeName("Geometry.ImportFromSatAndUnits")]
```
aparajit-pratap, line 20: How about naming this `ImportFromSATWithUnits`? It aligns with the analogous export class name.

src/Libraries/GeometryUI/ExportWithUnits.cs
```diff
@@ -1,17 +1,103 @@
...
+            if (!InPorts[0].IsConnected || (!InPorts[1].IsConnected && !InPorts[1].UsingDefaultValue))
+            {
+                return new[] { AstFactory.BuildAssignment(GetAstIdentifierForOutputIndex(0), AstFactory.BuildNullNode()) };
+            }
+
+            var funcNode = AstFactory.BuildFunctionCall("ImportHelpers", "ImportFromSATByUnits",
+                        new List<AssociativeNode> { inputAstNodes[0], inputAstNodes[1] });
+
+            return new[] { AstFactory.BuildAssignment(GetAstIdentifierForOutputIndex(0), funcNode) };
+        }
```
aparajit-pratap, line 57: What about creating a function object and all that boilerplate code if the node is not fully satisfied?

aparajit-pratap: And also lacing?

src/LibraryViewExtensionMSWebBrowser/web/library/layoutSpecs.json
mjkkirschner: Because the existing ExportToSAT node appears in the UI as `ExportToSAT` - but I thought we wanted this node to appear as `Geometry.ImportFromSATblahblahblah` - and I thought that, because the existing SAT import nodes look like: `Geometry.ImportFromSAT`
Screen Shot 2022-08-22 at 12 43 56 PM
Screen Shot 2022-08-22 at 12 43 50 PM

src/Libraries/GeometryUI/ExportWithUnits.cs
mjkkirschner: I was following the example from the export node which does not handle any of this, because this node doesn't have any UI I think it's more reasonable to expect that boiler plate behavior, will add.

src/LibraryViewExtensionMSWebBrowser/web/library/layoutSpecs.json
aparajit-pratap: Ok, great 👍

src/Libraries/GeometryColor/ImportHelpers.cs
```diff
@@ -40,22 +40,22 @@ public static IEnumerable<Geometry> ImportFromSATWithUnits(string filePath, [Def
...
-        /// Imports geometry from SAB filepath. Set the dynamoUnit input to match how you are 
+        /// Imports geometry from SAB byte array. Set the dynamoUnit input to match how you are 
         /// interperting the other numbers in your Dynamo file.
         /// </summary>
         /// <param name="buffer">SAB byte array</param>
         /// <param name="dynamoUnit">a forge unit length, if left null, sat file will be imported as unitless</param>
         /// <returns></returns>
         [AllowRankReduction]
-        public static IEnumerable<Geometry> DeserializeFromSABWithUnits(byte[] buffer, [DefaultArgument("null")] DynamoUnits.Unit dynamoUnit)
+        public static IEnumerable<Geometry> DeserializeFromSABWithUnits(byte[] buffer, DynamoUnits.Unit dynamoUnit)
```
aparajit-pratap, line 58: Isn't the default for mmPerUnit decided as -1, in which case, there should be a corresponding default argument for `DynamoUnits.Unit`?

src/Libraries/GeometryUI/ImportWithUnits.cs
```diff
@@ -2,90 +2,138 @@
...
 {
     //TODO finish the descriptions etc with resx.
 
     [NodeCategory(BuiltinNodeCategories.GEOMETRY)]
-    [NodeName("Geometry.ImportFromSatWithUnits")]
+    [NodeName("Geometry.ImportFromSATWithUnits")]
     [InPortTypes(new string[] { "var", "DynamoUnits.Unit" })]
-    //[NodeDescription("ExportToSATDescripiton", typeof(GeometryUI.Properties.Resources))]
-    //[NodeSearchTags("ExportWithUnitsSearchTags", typeof(GeometryUI.Properties.Resources))]
+    [InPortNames(new[] { "file|filePath","dynamoUnit" })]
```
aparajit-pratap, line 12: So there's just one nodemodel node for both file path and file object?

src/Libraries/GeometryColor/ImportHelpers.cs
mjkkirschner: it was not necessary because the node model already has a default, but I can add it back if you prefer both the internal and external method to have the same default.

src/Libraries/GeometryUI/ImportWithUnits.cs
mjkkirschner: yep, the correct overload is called at runtime, tested this a bunch and it seems solid.

src/Libraries/GeometryColor/ImportHelpers.cs
aparajit-pratap: I'm guessing that these methods could still be used in Python, code blocks, C# code, etc. so might be good to add them back.

PR #13240, opened 2022-08-24
Skip sending sensitive data

pinzart90:
PR changes:

Skip the following files until they are reviewed by the Data and Analytics team
1. Settings file
2. Log file

The dynamo file is considered IP so we will never send that through CER.
Send the call stack as it was recorded in the old crash prompt
Keep the Feature flag (as a production feature flag) so that we can still turn it off in case of something wrong.

### DIFF
--- a/src/AssemblySharedInfoGenerator/AssemblySharedInfo.cs
+++ b/src/AssemblySharedInfoGenerator/AssemblySharedInfo.cs
@@ -45,7 +45,7 @@ using System.Runtime.InteropServices;
 // to distinguish one build from another. AssemblyFileVersion is specified
 // in AssemblyVersionInfo.cs so that it can be easily incremented by the
 // automated build process.
-[assembly: AssemblyVersion("2.16.0.1640")]
+[assembly: AssemblyVersion("2.16.0.2355")]
 
 
 // By default, the "Product version" shown in the file properties window is
@@ -64,4 +64,4 @@ using System.Runtime.InteropServices;
 // You can specify all the values or you can default the Build and Revision Numbers 
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyFileVersion("2.16.0.1640")]
+[assembly: AssemblyFileVersion("2.16.0.2355")]
--- a/src/AssemblySharedInfoGenerator/AssemblySharedInfo.cs
+++ b/src/AssemblySharedInfoGenerator/AssemblySharedInfo.cs
@@ -45,7 +45,7 @@ using System.Runtime.InteropServices;
 // to distinguish one build from another. AssemblyFileVersion is specified
 // in AssemblyVersionInfo.cs so that it can be easily incremented by the
 // automated build process.
-[assembly: AssemblyVersion("2.16.0.2355")]
+[assembly: AssemblyVersion("2.16.0.2366")]
 
 
 // By default, the "Product version" shown in the file properties window is
@@ -64,4 +64,4 @@ using System.Runtime.InteropServices;
 // You can specify all the values or you can default the Build and Revision Numbers 
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyFileVersion("2.16.0.2355")]
+[assembly: AssemblyFileVersion("2.16.0.2366")]
--- a/src/DynamoCore/Core/CrashPromptArgs.cs
+++ b/src/DynamoCore/Core/CrashPromptArgs.cs
@@ -125,12 +125,12 @@ namespace Dynamo.Core
         /// <summary>
         /// Allow Dynamo to send the log file to the CER system.
         /// </summary>
-        public bool SendLogFile { get; set; } = true;
+        internal bool SendLogFile { get; set; } = false;
 
         /// <summary>
         /// Allow Dynamo to send the settings file to the CER system.
         /// </summary>
-        public bool SendSettingsFile { get; set; } = true;
+        internal bool SendSettingsFile { get; set; } = false;
 
         /// <summary>
         /// Allow Dynamo to send the recorded commands file to the CER system.
--- a/src/DynamoCore/Core/CrashPromptArgs.cs
+++ b/src/DynamoCore/Core/CrashPromptArgs.cs
@@ -116,6 +116,13 @@ namespace Dynamo.Core
         {}
 
         /// <summary>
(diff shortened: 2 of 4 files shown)


src/DynamoCoreWpf/Utilities/CrashReportTool.cs
```diff
@@ -245,10 +240,17 @@ internal static bool ShowCrashErrorReportWindow(DynamoViewModel viewModel, Crash
                         $@"session_start_count=\""0\"" session_clean_close_count=\""0\"" current_session_length=\""0\"" />";
                     }
 
+                    string dynConfig = string.Empty;
+                    string dynName = viewModel?.Model.CurrentWorkspace.Name;
+                    if (!string.IsNullOrEmpty(dynName))
+                    {
+                        dynConfig = $"/DWG {dynName}";
```
pinzart90, line 247: Send the name of the Dynamo file.
https://wiki.autodesk.com/pages/viewpage.action?spaceKey=NCERS&title=CER+Component+Integration+Guidance+-+6.1.x#CERComponentIntegrationGuidance6.1.x-DWG

src/DynamoCoreWpf/Utilities/CrashReportTool.cs
```diff
@@ -220,12 +216,11 @@ internal static bool ShowCrashErrorReportWindow(DynamoViewModel viewModel, Crash
...
-                    if (args.SendDynFile && model != null)
+                    if (args.HasDetails())
                     {
-                        var dynFilePath = Path.Combine(cerDir.FullName, "DynamoModel.dyn");
-                        model.CurrentWorkspace.Save(dynFilePath);
-
-                        filesToSend.Add(dynFilePath);
+                        var stackTracePath = Path.Combine(cerDir.FullName, "StackTrace.log");
+                        File.WriteAllText(stackTracePath, args.Details);
+                        filesToSend.Add(stackTracePath);
```
pinzart90, line 223: send the call stack as it was recorded in the old crash prompt

pinzart90:
Still waiting on word from Legal.
Once we get the ok, I will move this to Ready for Review

src/DynamoCore/Core/CrashPromptArgs.cs
```diff
@@ -116,19 +116,21 @@ public CrashErrorReportArgs(Exception e) : base(e)
...
-        public bool SendLogFile { get; set; } = true;
+        /// <param name="details">Crash details</param>
+        internal CrashErrorReportArgs(string details) : base(details)
+        { }
 
         /// <summary>
-        /// Allow Dynamo to send the settings file to the CER system.
+        /// Allow Dynamo to send the log file to the CER system.
         /// </summary>
-        public bool SendSettingsFile { get; set; } = true;
```
QilongTang, line 126: are these added in Dynamo 2.16 only so not API breaking?

src/DynamoCoreWpf/Utilities/CrashReportTool.cs
mjkkirschner: thanks!

src/DynamoCoreWpf/Utilities/CrashReportTool.cs
```diff
@@ -158,15 +158,11 @@ private static string FindCERToolInInstallLocations()
         /// <summary>
         /// Calls external CER tool (with UI)
         /// </summary>
+        /// <param name="viewModel">The Dynamo view model</param>
         /// <param name="args"></param>
         /// <returns>True if the CER tool process was successfully started. False otherwise</returns>
         internal static bool ShowCrashErrorReportWindow(DynamoViewModel viewModel, CrashErrorReportArgs args)
         {
-            if (DynamoModel.FeatureFlags?.CheckFeatureFlag("CER", false) == false)
```
mjkkirschner, line 165: instead of removing the flag check, why don't we just turn on the flag and keep this check until the next release - if something unexpected happens with the new feature then we can turn the flag back off instead of issuing a hotfix.

src/DynamoCoreWpf/Utilities/CrashReportTool.cs
pinzart90: Some context: 
@mjkkirschner noticed that the new CER UI cannot show the callstack (like the old prompt). IT was useful to quickly diagnose the problem. To view the callstack in CER, we will need to go to cer.autodesk.com and find the report (and sometimes download the callstack file, if cer.autodesk.com does not display).

not sure if it is a good UX...but we could try to show the callstack like we show the recorded commands (or even in the same document as the commands). 
At the moment the recorded commands (xml) will be opened in a separate process when a crash occurs.

@mjkkirschner @QilongTang any opinions?

src/DynamoCoreWpf/Utilities/CrashReportTool.cs
pinzart90: we could do that.
We would need a production feature flag

mjkkirschner: @Amoursol is currently the only "owner" but I think @jasonstratton could signup as well without any soc compliance issues.

Amoursol: > @Amoursol is currently the only "owner" but I think @jasonstratton could signup as well without any soc compliance issues.

Would be good for sure to have someone on the team with access! But always happy to help out where I can.

src/AssemblySharedInfoGenerator/AssemblySharedInfo.cs
```diff
@@ -45,7 +45,7 @@
 // to distinguish one build from another. AssemblyFileVersion is specified
 // in AssemblyVersionInfo.cs so that it can be easily incremented by the
 // automated build process.
-[assembly: AssemblyVersion("2.16.0.1640")]
+[assembly: AssemblyVersion("2.16.0.2355")]
```
mjkkirschner, line 48: can you revert this?

src/DynamoCoreWpf/Utilities/CrashReportTool.cs
mjkkirschner: I think as long as where to find the files gets documented somewhere it's not necessary to popup another window during a crash.

mjkkirschner (approved):

src/AssemblySharedInfoGenerator/AssemblySharedInfo.cs
pinzart90: done

PR #13245, opened 2022-08-25
DYN-5208-DynamoRevit-Crash-Fix

RobertGlobant20:
### Purpose

Fixing crash in Dynamo for Revit when initializing WebView2.
DocumentationBrowser was crashing in Dynamo for Revit due that WebView2 was trying to create the cache folder in the path "C:\Program Files\Autodesk\Revit 2023\AddIns\DynamoForRevit" due to permissions so in this fix I'm changing the folder to use the AppData for Revit "C:\Users\roberto.tellez\AppData\Roaming\Dynamo\Dynamo Revit\2.16".
The same medicin was applied for the WebView2 used in Guided Tours, when using the ResourceUtilities.LoadWebBrowser() method it will using the cache folder passed as parameter.

### Declarations

### Release Notes

Fixing crash in Dynamo for Revit when initializing WebView2.

### Reviewers

@QilongTang 

### FYIs

@avidit

### DIFF
--- a/src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
+++ b/src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
@@ -68,10 +68,6 @@ namespace Dynamo.DocumentationBrowser
             {
                 var docsDir = new DirectoryInfo(Path.Combine(pathManager.DynamoCoreDirectory, FALLBACK_DOC_DIRECTORY_NAME));
                 fallbackDocPath = docsDir.Exists ? docsDir : null;
-
-                //When executing DynamoSandbox the WebView2 cache folder will be located in the same directory than the executable
-                var userDataDir = new DirectoryInfo(pathManager.DynamoCoreDirectory);
-                webBrowserUserDataFolder = userDataDir.Exists ? userDataDir : null;
             }
 
             if (!string.IsNullOrEmpty(pathManager.HostApplicationDirectory))
@@ -81,11 +77,11 @@ namespace Dynamo.DocumentationBrowser
                 var hostAppDirectory = Directory.GetParent(pathManager.HostApplicationDirectory).FullName;
                 var docsDir = new DirectoryInfo(Path.Combine(hostAppDirectory, FALLBACK_DOC_DIRECTORY_NAME));
                 fallbackDocPath = docsDir.Exists ? docsDir : null;
-
-                //When executing Dynamo inside Revit the WebView2 cache folder will be located in the Revit AppData folder
-                var userDataDir = new DirectoryInfo(pathManager.UserDataDirectory);
-                webBrowserUserDataFolder = userDataDir.Exists ? userDataDir : null;
             }
+
+            //When executing Dynamo as Sandbox or inside any host like Revit, FormIt, Civil3D the WebView2 cache folder will be located in the AppData folder
+            var userDataDir = new DirectoryInfo(pathManager.UserDataDirectory);
+            webBrowserUserDataFolder = userDataDir.Exists ? userDataDir : null;
 
             if (this.BrowserView == null) return;
 
--- a/src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
+++ b/src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
@@ -24,6 +24,8 @@ namespace Dynamo.DocumentationBrowser
         private const string FALLBACK_DOC_DIRECTORY_NAME = "fallback_docs";
         //these fields should only be directly set by tests.
         internal DirectoryInfo fallbackDocPath;
+        //these fields should only be directly set by tests.
+        internal DirectoryInfo webBrowserUserDataFolder;
 
         internal DocumentationBrowserView BrowserView { get; private set; }
         internal DocumentationBrowserViewModel ViewModel { get; private set; }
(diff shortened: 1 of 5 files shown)


RobertGlobant20:
This GIF is showing the behavior of Dynamo in Revit.
fix_userdatafolder_webview2

src/DynamoCoreWpf/Views/GuidedTour/PopupWindow.xaml.cs
```diff
@@ -31,6 +31,11 @@ public partial class PopupWindow : Popup
         //Assembly path to the Resources folder
         private const string resourcesPath = "Dynamo.Wpf.Views.GuidedTour.HtmlPages.Resources";
 
+        /// <summary>
+        /// This property will be hold the path of the WebView2 cache folder, the value will change based in if DynamoSandbox is executed or Dynamo from Revit is executed
```
mjkkirschner, line 35: consider this generally, it's not just DynamoRevit, what about DynamoForCivil or Dynamo Formit etc?

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
```diff
@@ -67,18 +68,36 @@ public override void Startup(ViewStartupParams viewStartupParams)
...
+                var userDataDir = new DirectoryInfo(pathManager.DynamoCoreDirectory);
+                webBrowserUserDataFolder = userDataDir.Exists ? userDataDir : null;
             }
 
             if (!string.IsNullOrEmpty(pathManager.HostApplicationDirectory))
             {
-                var docsDir = new DirectoryInfo(Path.Combine(pathManager.HostApplicationDirectory, FALLBACK_DOC_DIRECTORY_NAME));
+                //when running over Revit HostApplicationDirectory = C:\Program Files\Autodesk\Revit 2023\AddIns\DynamoForRevit\Revit
+                //Then we need to remove the Revit folder from the path so we can find the fallback_docs directory.
+                var hostAppDirectory = Directory.GetParent(pathManager.HostApplicationDirectory).FullName;
```
QilongTang, line 81: We may want to double check on this if there is a better way. I will get back to you

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
```diff
@@ -67,18 +68,36 @@ public override void Startup(ViewStartupParams viewStartupParams)
...
+                var docsDir = new DirectoryInfo(Path.Combine(hostAppDirectory, FALLBACK_DOC_DIRECTORY_NAME));
                 fallbackDocPath = docsDir.Exists ? docsDir : null;
```
QilongTang, line 83: So for Sandbox, this `pathManager.HostApplicationDirectory` would be null? I think this only means this is not an in process integration, for case like FormIt, where it is out of process but still delivered in programe files, this may still not work

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
QilongTang: So for Sandbox, this pathManager.HostApplicationDirectory would be null? I think this only means this is not an in process integration, for case like FormIt, where it is out of process but still delivered in programe files, this may still not work

src/DynamoCoreWpf/Views/GuidedTour/PopupWindow.xaml.cs
RobertGlobant20: Comment updated: https://github.com/DynamoDS/Dynamo/pull/13245/commits/45e845bef4b113c5ebd934b2925f71e578253fab

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
RobertGlobant20: @QilongTang 
For other products like FormIt or Civil not sure which is the path of the Dynamo "fallback_doc" folder, I followed the implementation that I found in PackageDocumentationManager.cs (check link below) then if the fallback_doc folder location is different for each host (Revit, Civil, Formit) then the implementation done in PackageDocumentationManager is not working or hostDynamoFallbackDocPath = null  or there is another process creating the folder and moving all the md/jpg files to the HostApplicationDirectory path.

Should I install FormIt and Civil so I can test if images are shown in DocumentationBrowser?

https://github.com/DynamoDS/Dynamo/blob/79ebac3289a4474b4d3cba4d2280ed349ce6c710/src/DocumentationBrowserViewExtension/PackageDocumentationManager.cs#L66

RobertGlobant20: @QilongTang now the WebView2 cache folder will be always created in AppData, see the next commit:
https://github.com/DynamoDS/Dynamo/pull/13245/commits/e101e8a2699fc306239ef702b8ae408ba54a3078

src/DynamoCoreWpf/UI/GuidedTour/GuidesManager.cs
```diff
@@ -446,6 +457,11 @@ private Step CreateStep(Step jsonStepInfo, HostControlInfo hostControlInfo, int
                             Title = title
                         }
                     };
+                    var popupWindow = newStep.stepUIPopup as PopupWindow;
+                    if(popupWindow != null)
+                    {
+                        popupWindow.WebBrowserUserDataFolder = userDataFolder != null ? userDataFolder.FullName : string.Empty;
+                    }
```
QilongTang, line 464: Why do we need this code here? Shouldn't this be set one time only at the Welcome step?

QilongTang (commented):
LGTM with some comments

src/DynamoCoreWpf/UI/GuidedTour/GuidesManager.cs
RobertGlobant20: @RobertGlobant20 
1st question:
- we need to set the WebBrowserUserDataFolder  before calling the WebView2.EnsureCoreWebView2Async() method so it can set the cache folder in the right location, I put it here because in GuidesManager we have access to DynamoViewModel.Model.PathManager.

2nd question.
- I did a change that will be set only for popups that use html, check the commit below.

3rd question.
- Not all the Welcome steps use html then it will be set only for the ones using html (and by consequence WebView2).

commit:
https://github.com/DynamoDS/Dynamo/pull/13245/commits/b37dcc167897b8190e41c0a1985117abdae61855

QilongTang (approved):
Please see one of my comments about updating the comments about revit, I dont think it is up to date. Otherwise, LGTM

src/DynamoCoreWpf/UI/GuidedTour/GuidesManager.cs
mjkkirschner: @RobertGlobant20 are there any MSDN docs on how these webview2 startup options affect other instances of WebView2?

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
RobertGlobant20: @QilongTang fixed comment about Revit in the next commit:
https://github.com/DynamoDS/Dynamo/pull/13245/commits/2729c08968deabdad13b23ee962bae040d8c2fd1

src/DynamoCoreWpf/UI/GuidedTour/GuidesManager.cs
RobertGlobant20: @mjkkirschner MS documentation doesn't talk specifically about that but I think the startup options are specific for each WebView2 instance.
If you question is because guided tours and documentationbrowser will be using the same cache - user data folder (UDF) folder, MS Documentation says "It's possible for your host app to overlap them" but doesn't say anything about the consequences, I just opened the "User Interface" tour that opens the DocumentationBrowser at the same time than a Popup opened but I didn't see any problem (the Popup is not using html for this case), then once Notification Center is completed we will need to see what happen if we open the Notifications Center and the DocumentationBrowser at the same time.
https://docs.microsoft.com/en-us/microsoft-edge/webview2/concepts/user-data-folder?tabs=win32

mjkkirschner: @RobertGlobant20 thanks- I'm also curious because inside of Hosts like Revit, they will use WebView2 and so will the other addins running in the Host process. It's the wild west!

RobertGlobant20: @mjkkirschner well, we asked Revit about it but for my understanding they are still in the process of adopting WebView2 then they haven't deal with this feature yet so if Revit team decide to use another cache folder in AppData for WebView2 we can evaluate a change in Dynamo to use the same than them, another thing is that we need to ask if this feature of enable/disable cache for WebView2 will be enabled in the future (since I didn't find anything about it).
Also I checked with AutoCAD team and they said they create the folder in the user profile (consider that Autocad is using C++).
Thanks

QilongTang:
@RobertGlobant20 Please cherry-pick this PR

Issue #13168, opened 2022-08-03
Float Number Slider goes out of whack when dragging


aliasguru:
Surprised to not find a report about this here, but maybe people are using older versions of Dynamo, or don't compile themselves.
When I add a float slider to my node tree, no matter what boundaries I use, whenever I drag the slider, the `Min` and/or `Max` values change unexpectedly. See series of screenshots for illustration.

Narrowed it down to commit d09b5b23 causing this behavior. Simply reverting the commit resolves the problem for me.

## Dynamo version
RC2.15.0
the behavior is present in any release after commit d09b5b23 currently. However, I have not tested current `master`

## Operating system
Windows 10

## What did you do? 

lay down a Float Slider
![Start](https://user-images.githubusercontent.com/24734924/182583200-4d287e71-4c10-4c1c-9c7e-ddad54c9ae69.jpg)

start dragging the handle
![drag](https://user-images.githubusercontent.com/24734924/182583250-82d9ed5d-a4cb-42b2-8b20-4c073346f4fa.jpg)

enjoy changed boundaries (in this case, Max was changed while dragging)
![ridiculousValue](https://user-images.githubusercontent.com/24734924/182583308-ef768723-0372-432d-a73c-e244ea3cc57e.jpg)

## What did you expect to see?
Min and Max boundaries must not change, that's the point of them

## What did you see instead?
read above

mjkkirschner:
@reddyashish @QilongTang FYI

reddyashish:
@aliasguru I was not able to reproduce the max value change when dragging the slider. The max value will be changed only when a value(greater than current max) is manually entered in the box, which is expected or when the max value field is changed manually.
![slider](https://user-images.githubusercontent.com/43763136/183462758-e19e2839-70c8-4801-9fc0-9485791aefc1.gif)

aliasguru:
Thank you @reddyashish for investigating. I'm amazed it does not happen for you. First I thought this might be due to me compiling it myself, so to check that I downloaded the latest build from [here](https://dyn-builds-data.s3-us-west-2.amazonaws.com/DynamoCoreRuntime2.15.0.zip) and tried this example again. I just opened the Dynamo version and checked it is the latest one:

![Dynamo_Version](https://user-images.githubusercontent.com/24734924/183591901-31fee216-eb97-46af-9939-fac09c01ab2d.jpg)

Next, I just add in a new Number Slider, without changing anything:

[...]

PR #13248, opened 2022-08-25
The min and max value should not be changed when the slider value is changed by adjusting the slider.

reddyashish:
### Purpose

This change is in related to the issue mentioned here: https://github.com/DynamoDS/Dynamo/issues/13168

Even though no-one from the team was able to reproduce this, the user states that when the value in the slider is changed by moving the slider, the max value of the slider is set to a new value if the new value is greater than the previous max value. 

The public T Value{} property is only triggered when the value in slider is changed by dragging it. As the value in the slider cannot exceed the max value and the max/min value should not be changed in this case. So, we shouldn't be needing those conditions.

### Declarations

### Release Notes
The min and max value cannot be changed when the slider value is changed by adjusting the slider.

### Reviewers
@QilongTang @mjkkirschner

### DIFF
--- a/src/DocumentationBrowserViewExtension/DocumentationBrowserView.xaml.cs
+++ b/src/DocumentationBrowserViewExtension/DocumentationBrowserView.xaml.cs
@@ -125,8 +125,7 @@ namespace Dynamo.DocumentationBrowser
             //Due that the Web Browser(WebView2 - Chromium) security CORS is blocking the load of resources like images then we need to create a virtual folder in which the image are located.
             this.documentationBrowser.CoreWebView2.SetVirtualHostNameToFolderMapping(VIRTUAL_FOLDER_MAPPING, FallbackDirectoryName, CoreWebView2HostResourceAccessKind.DenyCors);
 
-            //This will remove special characters in paths (like <img src="/path")
-            string htmlContent = HttpUtility.UrlDecode(this.viewModel.GetContent());
+            string htmlContent = this.viewModel.GetContent();
 
             Dispatcher.BeginInvoke(new Action(() =>
             {
--- a/src/Tools/Md2Html/Md2Html.cs
+++ b/src/Tools/Md2Html/Md2Html.cs
@@ -81,7 +81,7 @@ namespace Md2Html
                     continue;
 
                 var imageName = Path.GetFileName(image.Url);
-                var absoluteImagePath = Path.Combine(VIRTUAL_FOLDER_MAPPING, imageName);
+                var absoluteImagePath = string.Join("/", VIRTUAL_FOLDER_MAPPING, imageName);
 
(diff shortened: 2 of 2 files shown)


QilongTang:
Although we cant validate on our end, but this looks reasonable. @aliasguru do you think you will be able to test this change locally and let us know?

mjkkirschner:
@QilongTang if this is merged, modifying the value directly past it's max value, will not update the max value.

reddyashish:
@mjkkirschner It will be changed here: https://github.com/DynamoDS/Dynamo/blob/1c83613d87e6031c141ff29245fc25b4e9a99cca/src/Libraries/CoreNodeModels/Input/SliderBase.cs#L97 when the value is updated. 

With the changes on this branch:
slider change

This public value property on sliderviewmodel is executed only when the value is dragged in the slider.

mjkkirschner:
@reddyashish ahh, thanks for the explanation.

aliasguru:
just tested it with the changes applied. Unfortunately, with this change the bug is back  :(
Now I went into Debugging, and found a lead. It seems like there is a problem with converting the Slider value to a float number later on, because my OS language is set to German. See this screenshot here, taken from within `CoreNodeModels`:

image

there is a `value` parameter, which comes in as a string, and shall be converted to a float. However, the string value is `10,7`, not `10.7` as you might expect! As a result, the `result` variable becomes `107` all of a sudden, explaining the explosion of numbers when I drag. See screenshot of variable content below:

image

That also explains why I only have it on the Number Slider, never on the Integer Slider. No comma / dot, no problem.
It puzzles me though, as they provide a `CultureInfo` as they should. As far as I can tell, you only reference to `CoreNodeModels`, it has not been created by you, right?

After the code has completed running, the slider indeed is at `107` instead of `10.7`:

image

At least now it's clear what's going on, and why only I have it and you can't reproduce. But how to fix it? The brute force way would be to replace any `,` with `.`, but this isn't the code you'd like to see in the neighborhood ususally:

    model.UpdateValue(new Dynamo.Graph.UpdateValueParams(nameof(Value), value.ToString().Replace(',', '.')));

This fixes the bug, but maybe there's a cleaner way with dealing with the German language?

mjkkirschner:
@aliasguru nice debugging - 
I think the behavior of the try parse is expected using invariant culture, IMO, what should be changed is that the string value that is passed to this conversion function should be in an invariant culture as well.

string replacement will just run into other edge cases unless it is also combined with culture checking, but it's not a good solution in general what if some data was serialized with an invariant culture even if the current culture is german for example.

If possible @reddyashish I would look into how this string is being created with current culture and not invariant culture to begin with.

aliasguru:
Am a bit confused now since this has been closed. Has there been a change to master? Or is a fix still coming?

PR #13252, opened 2022-08-26
add html license files from javascript projects

mjkkirschner:
### Purpose

Show license files for our javascript projects - these license files are generated using an adsk tool that produces html. To display them I used a webView2 browser and concatenated the html for each license nested inside of `<details>` elements that allow simple collapse/hide functionality.

To add new licenses they should be named like `componentName_license.html` and put into this folder in doc/distrib - thats all.

one downside is that now there are two slightly different about box scroll viewers, one for the RTF license.rtf and one of the html content.

If this is a problem, we can easily convert the .rtf file to html manually using microsoft word, and then concatenate all the html together - I think we'll need to keep the RTF file around for editing purposes and then manually create the .html for each edit, I did not do this because I'm not sure it's worth introducing the extra manual work into the release process - though I do like all the license files being an easily readable format - RTF is no longer improved, is hard to diff, and cannot be viewed without an RTF viewer.

Alternatively, we could just get rid of the .RTF entirely and just use the .html as our source of truth, it's fairly human readable. (of course we still need to ship the .RTF file for another release cycle in any case)

runtime conversion of rtf to html is possible, but out of scope for this pr, and fraught with edge cases.

Screen Shot 2022-08-25 at 9 22 39 PM

Screen Shot 2022-08-25 at 8 53 30 PM

### Declarations

### Release Notes

### Reviewers

### FYIs

### DIFF
--- a/src/DynamoCoreWpf/Views/About/AboutWindow.xaml.cs
+++ b/src/DynamoCoreWpf/Views/About/AboutWindow.xaml.cs
@@ -1,12 +1,15 @@
 ﻿using System;
 using System.Diagnostics;
 using System.IO;
+using System.Reflection;
 using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Documents;
 using System.Windows.Input;
 using System.Windows.Media;
 using Dynamo.ViewModels;
+using Microsoft.Web.WebView2.Core;
+using Microsoft.Web.WebView2.Wpf;
 
 namespace Dynamo.UI.Views
 {
@@ -16,6 +19,9 @@ namespace Dynamo.UI.Views
     public partial class AboutWindow : Window
     {
         bool ignoreClose;
+        private WebView2 htmlLicenseView;
+        static readonly string HTML_IMAGE_PATH_PREFIX = @"http://";
+        private const string ABOUT_BLANK_URI = "about:blank";
 
         public AboutWindow(DynamoViewModel dynamoViewModel)
         {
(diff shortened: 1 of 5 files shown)


src/DynamoCoreWpf/Views/About/AboutWindow.xaml.cs
```diff
@@ -37,6 +51,130 @@ private void HandleEsc(object sender, KeyEventArgs e)
                 Close();
         }
 
+        private void LoadHTMLAboutBoxContent()
+        {
+            string executingAssemblyPathName = Assembly.GetExecutingAssembly().Location;
+            string rootModuleDirectory = Path.GetDirectoryName(executingAssemblyPathName);
+            var htmlLicensePath = Path.Combine(rootModuleDirectory, "third_party_licenses_html");
```
QilongTang, line 58: So these htmls will be delivered instead of being dll resources, did I understand it correct?

QilongTang: Thank you for your answer @mjkkirschner I think this goes with my earlier concern, I was hoping they can be delivered with the npm package and embedded into project dll resources, just like the js file so we dont have to introduce new files. But I understand you started this with the way aligning more with the licence.rtf. I am fine with this and seems we can remove file at a later Global launch case.

mjkkirschner: hmm, I see, I can try packing these files into an assembly or assemblies and extracting from there... maybe we can use a simple naming scheme to automatically locate licenses that are embedded in any assembly, that way we don't need to have dependencies in dynamoCore on extension folders or files...

will think about it.

QilongTang: You can start from Notification Center, it will be more straightforward. If we update the cp step there, the html should be included automatically by Github Publish Action. Then it is only about how to consume it on Dynamo end. My 2 cents. I am supportive of using same naming or some schema, so DynamoCore only need to traverse all dlls and too out these block and append them one by one

mjkkirschner:
This will be addressed in https://jira.autodesk.com/browse/DYN-5230

PR #13256, opened 2022-08-26
DYN-5106-WebView2-DocumentationBrowser-Fix1

RobertGlobant20:
### Purpose

When displaying the documentation for NodeAutocomplete the GIFs were not showing correctly because the base64 content was filtered by HttpUtility.UrlDecode method then I removed this method call and also added code for joining the two sections of the URL, for this I'm using a call to  string.Join. After this fix the images for nodes and for node autocomplete are shown correctly.

### Declarations

### Release Notes

When displaying the documentation for NodeAutocomplete the GIFs were not showing correctly 

### Reviewers

@QilongTang 

### FYIs

@avidit

### DIFF
--- a/src/Tools/Md2Html/Md2Html.cs
+++ b/src/Tools/Md2Html/Md2Html.cs
@@ -81,7 +81,7 @@ namespace Md2Html
                     continue;
 
                 var imageName = Path.GetFileName(image.Url);
-                var absoluteImagePath = Path.Combine(VIRTUAL_FOLDER_MAPPING, imageName);
+                var absoluteImagePath = string.Join("/", VIRTUAL_FOLDER_MAPPING, imageName);
 
                 image.Url = $"{HTTP_IMAGE_PATH_PREFIX}{absoluteImagePath}";
             }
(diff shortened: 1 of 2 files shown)


RobertGlobant20:
This is a GIF showing documentation for nodes and nodeautocomplete.
node_autocomplete_documentation

src/Tools/Md2Html/Md2Html.cs
```diff
@@ -81,7 +81,7 @@ private static void ConvertRelativeLocalImagePathsToAbsolute(string mdFilePath,
                     continue;
 
                 var imageName = Path.GetFileName(image.Url);
-                var absoluteImagePath = Path.Combine(VIRTUAL_FOLDER_MAPPING, imageName);
+                var absoluteImagePath = string.Join("/", VIRTUAL_FOLDER_MAPPING, imageName);
```
QilongTang, line 84: What is the diff here do you mind?

RobertGlobant20: @QilongTang with Path.Combine generates something like "http://appassets%502my_image.png", then later I have to apply something like HttpUtility.UrlDecode(absoluteImagePath), so WebView2 can find the image in <img src=, but the problem is that when the src is a base64 content the HttpUtility.UrlDecode() method remove all special characters (that's why nodeautocomplete documentation GIFs are not appearing because has base64 content).
With  string.Join("/" it generates "http://appassets/my_image.png", then there is no need to use UrlDecode so in that case images in base64 and URL are working correctly.

QilongTang:
@RobertGlobant20 Please cherry-pick it

src/Tools/Md2Html/Md2Html.cs
mjkkirschner: @RobertGlobant20 this change has the downside that it's no longer cross platform - can you look at making the same bug fix but in a cross platform way where paths on linux and mac will work?

PR #13283, opened 2022-09-01
Notification bell hover color to match with rest of the dynamo elements

reddyashish:
### Purpose

Task: https://jira.autodesk.com/browse/DYN-5240

Before:
dyn-5240

After:
notification bell hover

### Declarations

### Release Notes
Notification bell hover color to match with rest of the dynamo elements

### Reviewers
@QilongTang @Amoursol

### DIFF
--- a/src/DynamoCoreWpf/Controls/ShortcutToolbar.xaml
+++ b/src/DynamoCoreWpf/Controls/ShortcutToolbar.xaml
@@ -186,18 +186,37 @@
                         <MenuItem.Header>
                             <Button x:Name="notificationsButton"
                                     BorderBrush="Transparent"
+                                    Margin="10,0,10,0"
                                     BorderThickness="0"
-                                    Background="Transparent"
                                     Cursor="Hand">
                                 <Button.ToolTip>
                                     <ToolTip Content="{x:Static p:Resources.NotificationCenterButtonTooltip}" Style="{StaticResource GenericToolTipLight}" />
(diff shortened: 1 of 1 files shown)


Amoursol:
@reddyashish can we please also change the icon on hover to blue? If @Jingyi-Wen is still on vacation, Helena should be able to help!

reddyashish:
For this use-case, we are not loading the bell icon as an image directly. I think that is because we are getting it from the HIG package. Dynamo loads it by converting it into font awesome code. There are only 2 images for bell outline, one for enable and the other was disable. Will confirm with @QilongTang.

QilongTang:
> For this use-case, we are not loading the bell icon as an image directly. I think that is because we are getting it from the HIG package. Dynamo loads it by converting it into font awesome code. There are only 2 images for bell outline, one for enable and the other was disable. Will confirm with @QilongTang.

Yes, fontawesome icons. We discussed at standup and will try change the forecolor when mouse hover. If it does not allow for fontawesome icon, then we will go with the current state.

reddyashish:
Foreground color applies to the text control and dint make any difference in our case.

QilongTang:
> @QilongTang Foreground color applies to the text control and did not make any difference in our case.

Thanks, I think I also tried that when I implement it. Make sense

QilongTang (approved):

QilongTang:
@reddyashish would you cherrypick this when you have time?

PR #13293, opened 2022-09-06
DYN-5245-DynamoRevit-CreateCustom UserDataFolder

RobertGlobant20:
### Purpose

Added code for creating the WebView2 cache folder in the AppData location because when tries to create it inside C:\Program Files\Autodesk\Revit 2023\AddIns\DynamoForRevit is crashing due to permissions then now will be created in the AppData location.

### Declarations

### Release Notes

Added code for creating the WebView2 cache folder in the AppData location

### Reviewers

@QilongTang 

### FYIs

@mjkkirschner

### DIFF
--- a/src/Notifications/NotificationCenterController.cs
+++ b/src/Notifications/NotificationCenterController.cs
@@ -15,6 +15,7 @@ using Dynamo.Notifications.View;
 using Dynamo.ViewModels;
 using Dynamo.Wpf.ViewModels.Core;
 using Newtonsoft.Json;
+using Microsoft.Web.WebView2.Wpf;
 
 namespace Dynamo.Notifications
 {
@@ -49,6 +50,7 @@ namespace Dynamo.Notifications
         private static readonly string htmlEmbeddedFile = "Dynamo.Notifications.node_modules._dynamods.notifications_center.build.index.html";
         private static readonly string jsEmbeddedFile = "Dynamo.Notifications.node_modules._dynamods.notifications_center.build.index.bundle.js";
         private static readonly string NotificationCenterButtonName = "notificationsButton";
+        internal DirectoryInfo webBrowserUserDataFolder;
 
         private readonly DynamoLogger logger;
         private string jsonStringFile;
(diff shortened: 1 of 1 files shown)


RobertGlobant20:
This is a GIF showing the notification on Dynamo for Revit but also showing the crash when closing Dynamo and notifications opened.
DynamoRevit_ShowingNotification_Crash

RobertGlobant20:
@QilongTang related to this question, the initialization in constructor was since the beginning I'm just creating a method named InitializeBrowserAsync() and call it in the constructor, do you think it should not be initialized in the constructor?
image

QilongTang:
> @QilongTang related to this question, the initialization in constructor was since the beginning I'm just creating a method named InitializeBrowserAsync() and call it in the constructor, do you think it should not be initialized in the constructor? image

No, I did not see the previous call. I think @mjkkirschner could be right that previously everything was async so the view tests could finish before WebView2 was required

src/Notifications/NotificationCenterController.cs
```diff
@@ -72,15 +78,29 @@ internal NotificationCenterController(DynamoView view, DynamoLogger dynLogger)
...
+            if (webBrowserUserDataFolder != null)
+            {
+                //This indicates in which location will be created the WebView2 cache folder
+                notificationUIPopup.webView.CreationProperties = new CoreWebView2CreationProperties()
+                {
+                    UserDataFolder = webBrowserUserDataFolder.FullName
+                };
+            }               
+            notificationUIPopup.webView.CoreWebView2InitializationCompleted += WebView_CoreWebView2InitializationCompleted;
+            await notificationUIPopup.webView.EnsureCoreWebView2Async();
```
QilongTang, line 103: This await should not be required right?

RobertGlobant20: @QilongTang well not sure if await is required or not but I knew that we should be awaiting async calls to finish otherwise we cannot control the flow of calls (if we need to use WebView2 later), for example in this case there is a call to RequestNotifications(); method before (in the original code was after), then not sure which problems could produce in the future but as you said let's test the behavior.
I will proceed to make the changes as you are suggesting and later we can do some experiments.
Thanks

QilongTang: @RobertGlobant20 I made the change in https://github.com/DynamoDS/Dynamo/pull/13295 and running tests. Because I see this as one of the big difference before/after in your previous PR

RobertGlobant20: FYI: commit: https://github.com/DynamoDS/Dynamo/pull/13293/commits/9dcf6020b3606e3edca8444efa57cfb6d5176052

QilongTang: @RobertGlobant20 For the `RequestNotifications()` call it should not affect WebView2 initialization, because the actual passing of information is bind to the event `NavigationCompleted`. So not until WebView2 was launched and navigated to the html, there would be no trial of passing info

QilongTang:
https://master-15.jenkins.autodesk.com/view/DYN/job/DYN-DevCI_Self_Service/1088/

QilongTang:
Tests passed merging. @RobertGlobant20 Please file the second PR based on this which would fix the crash

PR #13301, opened 2022-09-07
Dyn 5159 import export preferences

jesusalvino:
### Purpose

Import and Export Settings Features
https://jira.autodesk.com/browse/DYN-5159

### Declarations

### Reviewers
@QilongTang 

### FYIs
@RobertGlobant20 @filipeotero


QilongTang:
any gif of how it works now?

jesusalvino:
> any gif of how it works now?

Export

Export

Import

Import

src/DynamoCore/Configuration/PathManager.cs
```diff
@@ -686,6 +686,16 @@ private IEnumerable<string> LibrarySearchPaths(string library)
             yield return Path.GetFullPath(library);
         }
 
+        internal static string GetAppDataFolder()
```
QilongTang, line 689: What is the difference of this function v.s. e.g. `GetCommonDataFolder`?

src/DynamoCoreWpf/DynamoCoreWpf.csproj
```diff
@@ -1491,6 +1491,10 @@
     <EmbeddedResource Include="Views\GuidedTour\HtmlPages\Resources\DynamoPackages_1_1.png" />
     <Resource Include="UI\Images\default_info_16px.png" />
     <Resource Include="UI\Images\clock-icon.png" />
+    <Resource Include="UI\Images\question-default-16px.png" />
+    <Resource Include="UI\Images\question-hover-16px.png" />
+    <Resource Include="UI\Images\info-default-16px.png" />
+    <Resource Include="UI\Images\info-hover-16px.png" />
```
QilongTang, line 1497: I am surprised that we do not have these images already in the solution, are these new ones?

src/DynamoCoreWpf/ViewModels/Menu/TrustedPathViewModel.cs
```diff
@@ -171,6 +171,7 @@ private void CommitChanges(object param)
         internal void InitializeTrustedLocations()
         {
             TrustedLocations = new ObservableCollection<string>(settings?.TrustedLocations ?? new List<string>());
+            RaisePropertyChanged("");
```
QilongTang, line 174: what does this do?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -3156,6 +3156,12 @@ private void OnWorkspacePropertyChanged(object sender, PropertyChangedEventArgs
             if (args.PropertyName == "EnablePresetOptions")
                 OnPropertyChanged("EnablePresetOptions");
         }
+
+        public PreferenceSettings ImportPreferences(string newSettingsPath)
+        {
+            return PreferenceSettings.Load(newSettingsPath);
+        }
```
QilongTang, line 3163: Wrong section, this is the section of private functions. Also I think this can be internal method

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -3097,6 +3097,16 @@ private bool AskUserToSaveWorkspacesOrCancel(bool allowCancel = true)
...
 
+        /// <summary>
+        /// Import Preferences
+        /// </summary>
+        /// <param name="path">Path from the xml settings file to be imported</param>
+        /// <returns>A PreferencesSettings instance from the selected path</returns>
+        public PreferenceSettings ImportPreferences(string path)
+        {
+            return model.ImportPreferences(path);
+        }
```
QilongTang, line 3108: Same here, wrong section as this is the private function region

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -807,6 +807,43 @@ private void AddPythonEnginesOptions()
...
         public TrustedPathViewModel TrustedPathsViewModel { get; set; }
 
+        /// <summary>
+        /// Import Settings
+        /// </summary>
+        /// <param name="filePath"></param>
+        public void importSettings(string filePath)
+        {
+            PreferenceSettings newPreferences = dynamoViewModel.ImportPreferences(filePath);
+            newPreferences.CopyProperties(preferenceSettings);
```
QilongTang, line 817: Would you explain why some of the properties can be copied but others need to be set explicitly?

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -807,6 +807,43 @@ private void AddPythonEnginesOptions()
...
+            foreach (Wpf.ViewModels.Watch3D.DefaultWatch3DViewModel item in dynamoViewModel.Watch3DViewModels)
+            {
+                Interfaces.BackgroundPreviewActiveState preferenceItem = preferenceSettings.BackgroundPreviews.Where(i => i.Name == item.PreferenceWatchName).FirstOrDefault();
+                if (preferenceItem != null)
+                {
+                    item.Active = preferenceItem.IsActive;
+                }                
+            }
+
+            RaisePropertyChanged("");
```
QilongTang, line 844: This means?

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
```diff
@@ -978,17 +978,74 @@
...
+                                    MouseLeftButtonDown="importTextBlock_MouseLeftButtonDown"/>
+                                <Image Name="importToolTipIcon"
+                                    Grid.Column="0"
+                                    Margin="0,-5,0,0"
+                                    Width="16"
+                                    Height="16"
+                                    HorizontalAlignment="Center"
+                                    VerticalAlignment="Center"
+                                    Style="{StaticResource QuestionIcon}"
+                                    ToolTipService.ShowDuration="10000">
```
QilongTang, line 1010: If this is intentional, I wonder if this should be moved into `Style="{StaticResource GenericToolTipLight}"`

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
```diff
@@ -369,5 +373,62 @@ private void ScrollViewer_PreviewMouseWheel(object sender, MouseWheelEventArgs e
...
+                        selectedPathFile = Path.Combine(dialog.SelectedPath, PathManager.PreferenceSettingsFileName.Replace(".", "_" + ((DateTimeOffset)DateTime.UtcNow).ToUnixTimeSeconds().ToString() + "."));
+                    }
+
+                    File.Copy(currentSettingsFile, selectedPathFile);
+                    string argument = "/select, \"" + selectedPathFile + "\"";
+                    System.Diagnostics.Process.Start("explorer.exe", argument);
+                }
+                catch (Exception)
+                {
+                    throw;
```
QilongTang, line 429: We may need to log the exception content to Dynamo console here in case permission issue happens etc

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
```diff
@@ -369,5 +373,62 @@ private void ScrollViewer_PreviewMouseWheel(object sender, MouseWheelEventArgs e
...
+            System.Windows.Forms.DialogResult result = openFileDialog.ShowDialog();
+            if (result == System.Windows.Forms.DialogResult.OK)
+            {
+                try
+                {
+                    viewModel.importSettings(openFileDialog.FileName);
+                    Wpf.Utilities.MessageBoxService.Show(
+                       Res.ImportSettingsSuccessMessage, Res.ImportSettingsDialogTitle, MessageBoxButton.OK, MessageBoxImage.Information);
+                }
+                catch (LibraryLoadFailedException ex)
```
QilongTang, line 395: for other types of exceptions, what would happen?

QilongTang (commented):
Good progress, some questions and comments

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
```diff
@@ -369,5 +373,62 @@ private void ScrollViewer_PreviewMouseWheel(object sender, MouseWheelEventArgs e
             }
             e.Handled = true;
         }
+
+        private void importTextBlock_MouseLeftButtonDown(object sender, MouseButtonEventArgs e)
+        {
+            string[] fileFilter = { string.Format(Res.FileDialogImportSettingsFiles, "*.xml") };
+            System.Windows.Forms.OpenFileDialog openFileDialog = new System.Windows.Forms.OpenFileDialog();
```
RobertGlobant20, line 380: There is already a MessageBoxService (inside the MessageBoxUtilities.cs) that helps to show message boxes then you could use that class.

src/DynamoCore/Configuration/PathManager.cs
```diff
@@ -686,6 +686,16 @@ private IEnumerable<string> LibrarySearchPaths(string library)
...
+        {
+            var folder = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
+            var dynamoVersion = FileVersionInfo.GetVersionInfo(Assembly.GetExecutingAssembly().Location);
+            var appDataFolder = Path.Combine(Path.Combine(folder, "Dynamo", "Dynamo Core"),
+                $"{dynamoVersion.FileMajorPart}.{dynamoVersion.FileMinorPart}");
```
RobertGlobant20, line 694: Not sure which path you are trying to build here but I noticed that PathManager.UserDataDirectory has the path to:
C:\Users\roberto.tellez\AppData\Roaming\Dynamo\Dynamo Core\2.17
not sure if is useful.

RobertGlobant20 (commented):
LGTM with two comments

src/DynamoCore/Configuration/PathManager.cs
jesusalvino: Got it, using the Dynamo View Model reference instead

src/DynamoCoreWpf/DynamoCoreWpf.csproj
jesusalvino: Yes, They are new, We had already some similar but they weren't the same according to the expected design https://www.figma.com/file/dyhgWVlxDPJvOUTL07hUxa/Dynamo-Splash-Screen?node-id=207%3A91295

src/DynamoCoreWpf/ViewModels/Menu/TrustedPathViewModel.cs
jesusalvino: RaisePropertyChanged with no specific property notifies the UI all the properties data had changed so the user can see the new settings just after importing them

src/DynamoCore/Models/DynamoModel.cs
jesusalvino: Done

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
jesusalvino: I couldn't find the specific place/region for the public methods like DynamoModel has, just moving it after the last public

jesusalvino: Do you consider useful group them by access level as a next action item for later ?

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
jesusalvino: Because they don't have implemented the set method in their properties, that means their Can't be Written, That's why I need to access to the fields using the explicit methods like SetTrustWarningsDisabled and SetTrustedLocations.

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
jesusalvino: I have tried adding the ShowDuration in the GenericToolTipLight style but it didn't work, we need to define it per each kind of control that use a Tooltip, after researching the only way to apply a generic ShowDuration in the app is in the program main as the below image, Please let me know if I apply this change or just keep the current one setting an explicit duration to that image like other places doing. 

imagen

imagen

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
jesusalvino: Done.

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
jesusalvino: It will handling as a generic one.

src/DynamoCore/Configuration/PathManager.cs
jesusalvino: Done.

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
jesusalvino: Yes I use it when I handling the exception messages, In the highlighted case it needs to use the openFileDialog to grab a file as other places Does, thanks for the recommendation.

src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -14,6 +15,30 @@
...
 namespace Dynamo.Configuration
 {
+    static class ExtensionMethods
+    {
+        /// <summary>
+        /// Copy Properties from a PreferenceSettings instance to another
+        /// </summary>
+        /// <param name="source"></param>
+        /// <param name="destination"></param>
+        public static void CopyProperties(this PreferenceSettings source, PreferenceSettings destination)
```
mjkkirschner, line 25: can we make this internal?

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
mjkkirschner: maybe you could add internal setters, unless there was a reason we didn't add those @pinzart90 ?

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
```diff
@@ -369,5 +373,62 @@ private void ScrollViewer_PreviewMouseWheel(object sender, MouseWheelEventArgs e
...
+
+            if (dialog.ShowDialog() == System.Windows.Forms.DialogResult.OK)
+            {
+                string selectedPathFile = Path.Combine(dialog.SelectedPath, PathManager.PreferenceSettingsFileName);
+
+                try
+                {
+                    if (File.Exists(selectedPathFile))
+                    {
+                        selectedPathFile = Path.Combine(dialog.SelectedPath, PathManager.PreferenceSettingsFileName.Replace(".", "_" + ((DateTimeOffset)DateTime.UtcNow).ToUnixTimeSeconds().ToString() + "."));
```
mjkkirschner, line 419: 1. can you split this up into multiple lines, it's hard to read in github
2. seem like it deserves a comment.

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
```diff
@@ -369,5 +373,62 @@ private void ScrollViewer_PreviewMouseWheel(object sender, MouseWheelEventArgs e
...
+                    }
+
+                    File.Copy(dynViewModel.Model.PathManager.PreferenceFilePath, selectedPathFile);
+                    string argument = "/select, \"" + selectedPathFile + "\"";
+                    System.Diagnostics.Process.Start("explorer.exe", argument);
```
mjkkirschner, line 424: will this actually work on systems with only edge?

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
```diff
@@ -369,5 +373,62 @@ private void ScrollViewer_PreviewMouseWheel(object sender, MouseWheelEventArgs e
...
+
+                try
+                {
+                    if (File.Exists(selectedPathFile))
+                    {
+                        selectedPathFile = Path.Combine(dialog.SelectedPath, PathManager.PreferenceSettingsFileName.Replace(".", "_" + ((DateTimeOffset)DateTime.UtcNow).ToUnixTimeSeconds().ToString() + "."));
+                    }
+
+                    File.Copy(dynViewModel.Model.PathManager.PreferenceFilePath, selectedPathFile);
+                    string argument = "/select, \"" + selectedPathFile + "\"";
```
mjkkirschner, line 423: I think using Path.Combine is more idiomatic

src/DynamoCore/Configuration/PreferenceSettings.cs
pinzart90: looks like the class is internal (by default)
But it making the method itself internal makes it clearer I guess

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
pinzart90: I could not find a lot of documentation on this behavior, but there are a few mentions here and there (fyi We do not support this behavior for our inhouse model notifications) 
Maybe we should use String.Empty  instead of ""

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
pinzart90: Well trusted locations was designed to have a very restrictive API in order to make it as secure as possible (i.e make it hard to be hacked)
In order for Trusted location to be set, we would first need to check if it is valid (exists, read access, redundant paths etc)
We do have a example of setting the trusted locations blindly (without checks) but that is used only by tests.

We could have the setter be private and then try to get it with the proper reflection binding flags but there might still be issues.

Possible issues of adding the setter (internal or private) and/or using it by import :
1. Someone might use the setter without considering the implications
2. Using it in the import logic (without checking validity) might bring in bad paths.

jesusalvino: Those expected validations per each Location are handling when the settings is loaded here https://github.com/DynamoDS/Dynamo/blob/63e5c2f78f92a0144da7ae63fac4c855b3a28f1b/src/DynamoCore/Configuration/PreferenceSettings.cs#L368  by the helper https://github.com/DynamoDS/Dynamo/blob/63e5c2f78f92a0144da7ae63fac4c855b3a28f1b/src/DynamoUtilities/PathHelper.cs#L325

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
jesusalvino: Done.

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
```diff
@@ -369,5 +373,62 @@ private void ScrollViewer_PreviewMouseWheel(object sender, MouseWheelEventArgs e
             }
             e.Handled = true;
         }
+
+        private void importTextBlock_MouseLeftButtonDown(object sender, MouseButtonEventArgs e)
+        {
+            string[] fileFilter = { string.Format(Res.FileDialogImportSettingsFiles, "*.xml") };
```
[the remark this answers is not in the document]
jesusalvino: Done.

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
jesusalvino: Done.

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
jesusalvino: Are you referring to Microsoft Edge? if so, actually it doesn't deal with any browser, it opens the Windows Explorer based on the recently exported file and select it

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
jesusalvino: That variable means an argument for the System.Diagnostics.Process.Start instead a path.

mjkkirschner:
The tests seem to be timing out for this branch.

jesusalvino:
> The tests seem to be timing out for this branch.

I see all clean now, What's the next step in order to complete the merge ?

reddyashish (commented):
LGTM. Adding a unit test would be useful.

reddyashish (approved):

mjkkirschner:
tests are not only a very good idea, they seem like a requirement for a new feature like this that interacts with user data.

src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -14,6 +15,30 @@
...
+    {
+        /// <summary>
+        /// Copy Properties from a PreferenceSettings instance to another
+        /// </summary>
+        /// <param name="source"></param>
+        /// <param name="destination"></param>
+        internal static void CopyProperties(this PreferenceSettings source, PreferenceSettings destination)
+        {
+            // Iterate the Properties of the destination instance and populate them from their source counterparts  
+            PropertyInfo[] destinationProperties = destination.GetType().GetProperties();
```
mjkkirschner, line 28: I'll mention this again, `var` makes refactoring much easier and it makes the code easier to read for others because it reduces the line noise.

please use var where possible, if not for those reasons, because it matches the existing style of the rest of the source as well.

src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -14,6 +15,30 @@
...
+
+            foreach (PropertyInfo destinationPi in destinationProperties)
+            {
+                PropertyInfo sourcePi = source.GetType().GetProperty(destinationPi.Name);
+
+                if (destinationPi.GetCustomAttributes(typeof(System.ObsoleteAttribute), true).Length == 0 && destinationPi.CanWrite)
```
mjkkirschner, line 34: so if a property is obsolete it's not copied? This should go in the description/summary of the method.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -2867,6 +2867,11 @@ public void ClearCurrentWorkspace()
             OnWorkspaceCleared(CurrentWorkspace);
         }
 
+        internal PreferenceSettings ImportPreferences(string newSettingsPath)
```
mjkkirschner, line 2870: is this really necessary?

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -807,6 +807,43 @@ private void AddPythonEnginesOptions()
...
+            var engine = PythonEnginesList.FirstOrDefault(x => x.Equals(preferenceSettings.DefaultPythonEngine));
+            SelectedPythonEngine = string.IsNullOrEmpty(engine) ? Res.DefaultPythonEngineNone : preferenceSettings.DefaultPythonEngine;
+            dynamoViewModel.RenderPackageFactoryViewModel.MaxTessellationDivisions = preferenceSettings.RenderPrecision;
+            dynamoViewModel.RenderPackageFactoryViewModel.ShowEdges = preferenceSettings.ShowEdges;
+            PackagePathsForInstall = null;
+            PackagePathsViewModel?.InitializeRootLocations();
+
+            dynamoViewModel.IsShowingConnectors = preferenceSettings.ShowConnector;
+            dynamoViewModel.IsShowingConnectorTooltip = preferenceSettings.ShowConnectorToolTip;
+            foreach (Wpf.ViewModels.Watch3D.DefaultWatch3DViewModel item in dynamoViewModel.Watch3DViewModels)
```
mjkkirschner, line 835: var please

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
```diff
@@ -369,5 +373,66 @@ private void ScrollViewer_PreviewMouseWheel(object sender, MouseWheelEventArgs e
...
+            if (dialog.ShowDialog() == System.Windows.Forms.DialogResult.OK)
+            {
+                string selectedPathFile = Path.Combine(dialog.SelectedPath, PathManager.PreferenceSettingsFileName);
+
+                try
+                {
+                    if (File.Exists(selectedPathFile))
+                    {
+                        string uniqueId = ((DateTimeOffset)DateTime.UtcNow).ToUnixTimeSeconds().ToString();
+                        string suffixPlusDot = "_" + uniqueId + ".";
```
mjkkirschner, line 421: small thing but you can do `$_{uniqueId}.`

src/DynamoCore/Configuration/PreferenceSettings.cs
jesusalvino: if a property is obsolete it just will be skipped, both old and new instance are created from the same class (defined property), I will create a test for it.

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -3064,6 +3064,16 @@ public bool PerformShutdownSequence(ShutdownParams shutdownParams)
             return true;
         }
 
+        /// <summary>
+        /// Import Preferences
+        /// </summary>
+        /// <param name="path">Path from the xml settings file to be imported</param>
+        /// <returns>A PreferencesSettings instance from the selected path</returns>
+        public PreferenceSettings ImportPreferences(string path)
```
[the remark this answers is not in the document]
jesusalvino: yes for keeping the separate responsibilities, there is already an instance created of it, It just uses it.

src/DynamoCore/Configuration/PreferenceSettings.cs
jesusalvino: Deone.

src/DynamoCore/Models/DynamoModel.cs
jesusalvino: Done.

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
jesusalvino: Done.

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
jesusalvino: Done.

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
QilongTang: Can you remind me the default value of this setting? If not too different, I think the current implementation would be fine based on your investigation

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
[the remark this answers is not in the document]
QilongTang: To me it seems we dont really need this function, I think that is what @mjkkirschner suggested. Can we remove this function and rely on `PreferencesSetting.ImportPreferences(path);` where needed?

QilongTang (commented):
some comments then LGTM with unit test

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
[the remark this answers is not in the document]
jesusalvino: > To me it seems we dont really need this function, I think that is what @mjkkirschner suggested. Can we remove this function and rely on `PreferencesSetting.ImportPreferences(path);` where needed?

Done.

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
jesusalvino: 5 seconds as I recall, is fall short time to read the tooltip message, that's why I'm setting it at 10 seconds.

test/DynamoCoreTests/Configuration/PreferenceSettingsTests.cs
```diff
@@ -199,5 +203,136 @@ public void TestSerializationTrustedLocations()
 
             Assert.IsTrue(settingsLoaded.IsTrustedLocation(Path.GetTempPath()));
         }
+
+        private List<string> PropertiesWithStaticValues()
+        {
+            List<string> staticProperties = new List<string>() { "PythonTemplateFilePath","DefaultPythonEngine","NodeSearchTagSizeLimit","IronPythonResolveTargetVersion"};
```
QilongTang: If really required to be hardcoded, why not just make this a static member of the class?

QilongTang (commented):
LGTM with one comment

QilongTang:
Also the new test does not seem to pass from the CI: [Dynamo.Tests.Configuration.PreferenceSettingsTests.TestImportCopySettings](https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/7120/testReport/Dynamo.Tests.Configuration/PreferenceSettingsTests/TestImportCopySettings/)

test/DynamoCoreTests/Configuration/PreferenceSettingsTests.cs
jesusalvino: refactored.

src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -914,5 +938,14 @@ internal void OnMessageLogged(ILogMessage msg)
...
         #endregion
+
+        /// <summary>
+        /// List of static Fields to be excluded for evaluation due to their access level
+        /// </summary>
+        /// <returns></returns>
+        public List<string> StaticFields()
+        {
+            return typeof(PreferenceSettings).GetMembers(BindingFlags.Static | BindingFlags.NonPublic).OfType<FieldInfo>().Select(field => field.Name).ToList();
+        }
```
QilongTang, line 949: Where is this used?

jesusalvino: in the Unit Test, just updated.

QilongTang (approved):

QilongTang:
LGTM waiting for PR checks to finish

PR #13313, opened 2022-09-15
add nunit test adapter for 2022

pinzart90:
Add nunit test adapter for Visual Studio 2022
Cleanup extern/Analytics folder and related tasks (System.Net.Http is part of the dotnet framework...so we do not need a copy in our releases). There is a version difference : Version from extern- 4.6.1055; version from net48 - 4.8.3671 but there should be no breaking changes.

image


aparajit-pratap (approved):

mjkkirschner:
for system http - I think since 2.17 is likely a global launch release we can now remove everything from the legacy remove me folder and all the copy steps.

pinzart90:
> for system http - I think since 2.17 is likely a global launch release we can now remove everything from the legacy remove me folder and all the copy steps.

I will create a new PR for that so that the intention/changeset is clear

PR #13317, opened 2022-09-16
HelixWatch3DViewModel new property added

dnenov:
### Purpose

This PR implements a new public property to HelixWatch3DViewModel. The property allows checking if a graph has procured Geometry after the run. 

The HelixWatch3DViewModel always renders a number of default elements. 
- Grid
- Axis
- Light

Anything more would mean new geometry has been rendered.

### Declarations

### Release Notes

- added a new public property to HelixWatch3DViewModel checking if any Geometry was rendered by the graph
- test added for new property
- minor bug-fix

### Reviewers

@reddyashish 

### FYIs

@mmisol 
@sm6srw


src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -409,6 +409,25 @@ public override bool IsGridVisible
...
+        /// Identifies if the Graph yields any rendered geometry
+        /// Any graph would always render at least 3 elements:
+        /// Headlight, Grid, and Axis
+        /// Anything more than the default elements is considered rendered geometry
+        /// </summary>
+        public bool HasRenderedGeometry
+        {
+            get
+            {
+                return Element3DDictionary.Count(g =>
```
mjkkirschner, line 423: 1. this seems quite wasteful in terms of performance, but it's unclear to me from this PR what the anticipated use of this API is - can you give some context?

2. In addition, it appears one could call this API in and get incorrect data, ie, when is it safe to call this API? Is there an event that it should be called after, a specific task in the scheduler? From the UI thread? After a run? This should be included in the summary.

3. Do you consider geometry with preview off to be rendered? Maybe add a test for that case and describe the behavior in the summary.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -1337,11 +1338,22 @@ private void DynamoViewModelRequestSaveImage(object sender, ImageSaveEventArgs e
...
+            var dpiY = 0.0;
 
+            // dpi aware, otherwise incorrect images are created
+            try
+            {
+                var scale = VisualTreeHelper.GetDpi(this);
+                dpiX = scale.PixelsPerInchX;
+                dpiY = scale.PixelsPerInchY;
+            }
+            catch (Exception)
```
mjkkirschner, line 1351: it would be good to log this I think.

src/VisualizationTests/HelixWatch3DViewModelTests.cs
```diff
@@ -721,6 +721,25 @@ public void HelixWatch3DViewModel_ChangeBackgroundVisibility_CanNavigateButtonsA
             Assert.AreEqual(Visibility.Visible, currentWorkspace.statusBarPanel.Visibility, "Navigation buttons did not appear");
         }
 
+
+        [Test]
+        public void HelixWatch3DViewModel_HasRenderedGeometryTest()
```
mjkkirschner, line 726: 👍

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
dnenov: You are right, I did not think about that. My initial attempt was a simple `dictionary.Count > 3 `check, which was a bit vague. I thought this is more explicit, however, I realize it is quite computationally intensive for large graphs. 

It seems we always draw 3 default elements, which constitutes an 'empty' scene. 

I did think of hidden geometry, but in my use case, this still counts as an empty scene. 

Currently, there is no way to access this information from the context of an external extension hence the implementation of this property. I can revert back to the simpler check, which should address the performance concern I hope. In terms of scope, the intention is to query after a run has finished and the dispatcher has no remaining tasks. I can add this to the description.

mjkkirschner: well the context is important, if this method is called infrequently than maybe it does not matter - though iterating every single geometry object could be very slow depending on what is being drawn.

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
dnenov: Got it!

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
dnenov: I pushed the changed version, hopefully, this is a bit clearer and less misleading to use.

mjkkirschner (approved):

PR #13323, opened 2022-09-20
Return object instead of IList from ReplicationLogic

aparajit-pratap:
### Purpose

Return object instead of IList from ReplicationLogic

Hopefully this should address https://jira.autodesk.com/browse/DYN-5152

### Declarations

### Release Notes

### Reviewers

### FYIs

### DIFF
--- a/src/Engine/EmitMSIL/Replication.cs
+++ b/src/Engine/EmitMSIL/Replication.cs
@@ -251,9 +251,6 @@ namespace EmitMSIL
             {
                 result = ExecWithRISlowPath(finalFep, reducedArgs, replicationInstructions, runtimeCore);
             }
-            //if (result is IList) return (IList)result;
-
-            //return new[] { result };
             return result;
         }
 
--- a/src/Engine/EmitMSIL/Replication.cs
+++ b/src/Engine/EmitMSIL/Replication.cs
@@ -220,8 +220,6 @@ namespace EmitMSIL
             // TODO_MSIL: Emit these CLRStackValue's from the CodeGen stage.
             var stackValues = MarshalFunctionArguments(args, runtimeCore);
 
-            //var reducedArgs = ReduceArgs(stackValues);
-
             // TODO_MSIL: Emit these CLRFunctionEndpoint's from the CodeGen stage.
             var feps = CreateFEPs(mInfos);
 
@@ -253,23 +251,6 @@ namespace EmitMSIL
             }
             return result;
         }
-
-        //private static List<CLRStackValue> ReduceArgs(List<CLRStackValue> args)
-        //{
-        //    var reducedArgs = new List<CLRStackValue>();
-        //    foreach (var arg in args)
-        //    {
-        //        if (arg.IsEnumerable && (arg.Value as IList).Count == 1)
-        //        {
-        //            reducedArgs.Add((arg.Value as IList<CLRStackValue>)[0]);
-        //        }
-        //        else
-        //        {
-        //            reducedArgs.Add(arg);
-        //        }
-        //    }
-        //    return reducedArgs;
-        //}
 
         private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replicationAttrs)
         {
--- a/src/Engine/EmitMSIL/Replication.cs
+++ b/src/Engine/EmitMSIL/Replication.cs
@@ -566,7 +566,7 @@ namespace EmitMSIL
 
             if (o.IsEnumerable)
             {
-                return o.Value as IList<CLRStackValue> ?? new List<CLRStackValue>() { o };
+                return o.Value as IList<CLRStackValue>;
             }
             else
             {
--- a/src/Engine/EmitMSIL/Replication.cs
+++ b/src/Engine/EmitMSIL/Replication.cs
@@ -211,7 +211,7 @@ namespace EmitMSIL
         /// <param name="args"></param>
         /// <param name="replicationAttrs"></param>
         /// <returns></returns>
-        public static IList ReplicationLogic(IEnumerable<MethodBase> mInfos, IList args, string[][] replicationAttrs)
+        public static object ReplicationLogic(IEnumerable<MethodBase> mInfos, IList args, string[][] replicationAttrs)
         {
             // Static instance of MSILRuntimeCore
             // TODO_MSIL: FIgure out how and when to set this runtimeCore (CodeGen should have it too)
@@ -251,9 +251,10 @@ namespace EmitMSIL
             {
                 result = ExecWithRISlowPath(finalFep, reducedArgs, replicationInstructions, runtimeCore);
             }
-            if (result is IList) return (IList)result;
+            //if (result is IList) return (IList)result;
 
-            return new[] { result };
+            //return new[] { result };
+            return result;
         }
 
         private static List<CLRStackValue> ReduceArgs(List<CLRStackValue> args)
--- a/src/Engine/EmitMSIL/Replication.cs
+++ b/src/Engine/EmitMSIL/Replication.cs
@@ -220,7 +220,7 @@ namespace EmitMSIL
             // TODO_MSIL: Emit these CLRStackValue's from the CodeGen stage.
             var stackValues = MarshalFunctionArguments(args, runtimeCore);
 
-            var reducedArgs = ReduceArgs(stackValues);
+            //var reducedArgs = ReduceArgs(stackValues);
 
             // TODO_MSIL: Emit these CLRFunctionEndpoint's from the CodeGen stage.
             var feps = CreateFEPs(mInfos);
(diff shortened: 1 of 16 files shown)


pinzart90:
@aparajit-pratap Looks like you simply replace the IList with object.
I guess I am now thinking why we needed the IList in the first place ...

aparajit-pratap:
> @aparajit-pratap Looks like you simply replace the IList with object without any regressions (asides from mending a few tests) I guess I am now thinking why we needed the IList in the first place ...

@pinzart90 I also made a few changes in the replication code. PTAL.

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
```diff
@@ -456,20 +458,33 @@ protected T[] UnMarshal<T>(CLRStackValue dsObject, ProtoCore.MSILRuntimeCore run
             var dsElements = dsObject.Value as IList<CLRStackValue>;
```
aparajit-pratap, line 458: I think I need to add more checks such as this for everywhere that we're doing `stackValue.Value as IList<CLRStackValue>` since it can now be an `IList `of `object`.

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -587,7 +566,7 @@ private static IList<CLRStackValue> getSubParameters(CLRStackValue o)
 
             if (o.IsEnumerable)
             {
-                return o.Value as IList<CLRStackValue>;
+                return o.Value as IList<CLRStackValue> ?? new List<CLRStackValue>() { o };
```
pinzart90, line 569: I am trying to figure out why `o.Value` would not be `IList<CLRStackValue>`
The Marshaler should always convert enumerables to `IList<CLRStackValue>` I think...

aparajit-pratap: Well, in this particular function I still haven't encountered the case where `o.Value` is something other than `IList<CLRStackValue>` so I guess I can revert this but definitely in the other places I needed to check if the `Value` property was something other than `IList<CLRStackValue>`. 

Did your question pertain to just this function or all such changes in general?

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
mjkkirschner: should this be a helper?

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
```diff
@@ -456,20 +458,33 @@ protected T[] UnMarshal<T>(CLRStackValue dsObject, ProtoCore.MSILRuntimeCore run
...
+                    else
+                    {
+                        result.Add((T)obj);
+                    }
                 }
-                else
+            }
+            else
+            {
+                var enumerator = (dsObject.Value as IEnumerable).GetEnumerator();
```
mjkkirschner, line 484: is this any different from using a foreach loop?

test/CodeGenILTests/IndexingMicroTests.cs
```diff
@@ -168,7 +168,7 @@ public void IndexIntoArray_ObjectArray()
             var output = codeGen.EmitAndExecute(ast);
             Assert.IsNotEmpty(output);
             //TODO unclear why this index result has an extra level of nesting in the output dictionary. Replication maybe?
```
mjkkirschner, line 170: can we remove the TODO?

test/DynamoCoreTests/SerializationTests.cs
```diff
@@ -186,12 +187,20 @@ private void Init(WorkspaceModel workspace, EngineController controller, bool ds
...
-                                var objs = n.GetCLRValue(p.Index, controller).Cast<object>();
-                                var values = objs.Select(x => ProtoCore.Utils.CoreUtils.GetDataOfCLRValue(x)).ToList();
-                                if (values.Count == 1) { return values[0]; }
-                                return values;
+                                var obj = n.GetCLRValue(p.Index, controller);
+                                if (typeof(IEnumerable).IsAssignableFrom(obj.GetType()))
+                                {
+                                    var enumerator = (obj as IEnumerable).GetEnumerator();
+                                    var values = new List<object>();
+                                    while(enumerator.MoveNext())
```
mjkkirschner, line 197: can't use foreach?

src/Engine/EmitMSIL/Replication.cs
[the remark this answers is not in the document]
pinzart90: `o.IsEnumerable` specifically discards the case of string

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
aparajit-pratap: This was actually a question to @pinzart90 since I'm not clear if all clr stackvalues that are enumerable are expected to be `IList<CLRStackValue>`.

src/Engine/EmitMSIL/Replication.cs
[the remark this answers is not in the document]
mjkkirschner: thanks @pinzart90 I missed that this method is implemented on the `CLRStackValue` type.

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -220,15 +220,13 @@ public static IList ReplicationLogic(IEnumerable<MethodBase> mInfos, IList args,
             // TODO_MSIL: Emit these CLRStackValue's from the CodeGen stage.
             var stackValues = MarshalFunctionArguments(args, runtimeCore);
 
-            var reducedArgs = ReduceArgs(stackValues);
```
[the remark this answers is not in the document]
aparajit-pratap: `ReduceArgs` would assume that if there is an argument that is an `IList` of a single value, it would need to be reduced to a single value. Firstly, I believe this assumption is flawed, secondly, I don't believe it's required anymore given that we are not wrapping single values as an `IList `of a single value anymore.

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
```diff
@@ -600,35 +612,67 @@ internal override object UnMarshal(CLRStackValue dsObject, System.Type expectedC
...
 
             if (expectedCLRType.IsArray || expectedCLRType.IsGenericType)
             {
-                var list = collection as ArrayList;
-                if (null != list)
-                    return list.ToArray(elementType);
-            }
+                if (collection is Array) return collection;
 
+                if(elementType == typeof(int))
```
pinzart90, line 640: as an alternative we could use
```

                unsafe
                {
                    return Array.CreateInstance(elementType, collection.Count);
                }
 ```
 If we are ok with using unsafe...(the project would also need  `<AllowUnsafeBlocks>True</AllowUnsafeBlocks>`)

File: src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs [hunk 8]
```diff
@@ -600,35 +612,67 @@ internal override object UnMarshal(CLRStackValue dsObject, System.Type expectedC
...
+                {
+                    return ConvertToTypedArray<int>(collection);
+                }
+                if (elementType == typeof(long))
+                {
+                    return ConvertToTypedArray<long>(collection);
+                }
+                if (elementType == typeof(double))
+                {
+                    return ConvertToTypedArray<double>(collection);
```
pinzart90, line 650: another alternative is to use collection.Cast<T>().ToArray(); instead of the custom `ConvertToTypedArray` function you implemented.

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
aparajit-pratap: Is there any advantage to doing this? Is the advantage that it will support arbitrary types rather than just the primitive types that I have implemented?

pinzart90: Pretty much just arbitrary types and less code.

mjkkirschner: I'm just curious - is there some way to convert to a avoid iterating/copying the entire collection...
https://stackoverflow.com/questions/4972951/listt-to-t-without-copying

mjkkirschner:
I think theres something to be gained by profiling and investigating ways to improve the performance of collection marshaling - but that seems like it's out of scope for this PR.

mjkkirschner (approved):

pinzart90 (approved):
LGTM
Please use whatever conversion method you think best.

src/Engine/ProtoCore/FFI/CLRObjectMarshaler.cs
```diff
@@ -600,35 +612,67 @@ internal override object UnMarshal(CLRStackValue dsObject, System.Type expectedC
...
 
             if (expectedCLRType.IsGenericType && !expectedCLRType.IsInterface)
             {
                 if (!collection.GetType().IsArray)
                 {
-                    Validity.Assert(collection is ArrayList);
-                    collection = (collection as ArrayList).ToArray(elementType);
+                    collection = (collection as List<object>).ToArray();
                 }
                 return Activator.CreateInstance(expectedCLRType, new[] { collection });
```
aparajit-pratap, line 633: @pinzart90 by your argument, wouldn't this existing code also be unsafe as this relies on there being a suitable constructor for whichever type is passed as `expectedCLRType`?

pinzart90: Yes, I agree that it looks a bit unsafe.
Not sure what we could do, except try to get the constructor first and check if it exists
Or we could whitelist  the expectedType (Array, ArrayList, List etc)

PR #13327, opened 2022-09-21
null checking

pinzart90:
ALIAS reported issues (crash and broken background preview) with Dynamo 2.15 when using the DisableAnalytics flag set to TRUE
https://jira.autodesk.com/browse/DYN-5273


aparajit-pratap:
@pinzart90 what are the changes in the license.rtf file?

mjkkirschner:
@pinzart90 can you explain how the disable flag and feature flags are related to the background preview not working?

mjkkirschner:
> @pinzart90 what are the changes in the license.rtf file?

a good example of why switching to .html will be useful.

src/DynamoCore/Logging/AnalyticsService.cs
```diff
@@ -84,6 +84,12 @@ internal static bool IsADPOptedIn
 
         internal static bool IsADPAvailable()
         {
+            if (Analytics.DisableAnalytics ||
+                adpAnalyticsUI == null)
+            {
+                return false;
+            }
+
             return adpAnalyticsUI.IsProviderAvailable();
```
aparajit-pratap, line 93: Was this the reason for the crash? Is `adpAnalyticsUI` `null `if `DisableAnalytics` is `true`?

aparajit-pratap (approved):
Some questions.

src/DynamoCore/Logging/AnalyticsService.cs
pinzart90: Yes. `adpAnalyticsUI` will be null when `DisableAnalytics` is true. 
We do not initialize `adpAnalyticsUI` so that no analytics assemblies will be loaded.

pinzart90:
> @pinzart90 can you explain how the disable flag and feature flags are related to the background preview not working?

Looks like the FeatureFlag object is not initialized if the DisableAnalytics flag is turned on. Probably to avoid any network traffic (for ALIAS's benefit mostly)

mjkkirschner:
> > @pinzart90 can you explain how the disable flag and feature flags are related to the background preview not working?
> 
> Looks like the FeatureFlag object is not initialized if the DisableAnalytics flag is turned on. Probably to avoid any network traffic (for ALIAS's benefit mostly)

oh yeah - thats correct, completely forgot.

mjkkirschner (approved):

PR #13331, opened 2022-09-22
Add test selection node and corresponding test

aparajit-pratap:
### Purpose

Add test selection node to test Revit regression.

### Declarations

Check these if you believe they are true

- [ ] The codebase is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.
- [ ] This PR modifies some build requirements and the readme is updated

### Release Notes

(FILL ME IN) Brief description of the fix / enhancement. **Mandatory section** 

### Reviewers

(FILL ME IN) Reviewer 1  (If possible, assign the Reviewer for the PR)

(FILL ME IN, optional) Any additional notes to reviewers or testers.

### FYIs

(FILL ME IN, Optional) Names of anyone else you wish to be notified of

PR #13334, opened 2022-09-22
Dyn5272

mjkkirschner:
### Purpose

a continuation of @aparajit-pratap 's work started here:
https://github.com/DynamoDS/Dynamo/pull/13331
and here:
https://github.com/DynamoDS/Dynamo/pull/13221

We have issues using the input optimization that was put in place to skip recompiling primitive inputs when executing complex UI nodes where their AST produces primitive values sometimes and complex values other times - for example, the Revit Face selection node which returns null sometimes (a primitive) and the result of a function call other times.

We check the previous assignment statement for the node's output identifier to determine if the previous result was a primitive before attempting to apply any input optimization and skip compilation.

We have verified this fixes the Revit regressions with selection nodes using RTF.

### Declarations

### Release Notes

Fixes regression with selection nodes in Revit.

### Reviewers

### FYIs

### DIFF
--- a/src/Engine/ProtoScript/Runners/LiveRunner.cs
+++ b/src/Engine/ProtoScript/Runners/LiveRunner.cs
@@ -1022,7 +1022,7 @@ namespace ProtoScript.Runners
 
                     //Check if the subtree (ie a node in graph) is an input and has primitive Right hand Node type
                     if (redefinitionAllowed && st.IsInput && node is BinaryExpressionNode bne  && CoreUtils.IsPrimitiveASTNode(bne.RightNode)
-                        && CoreUtils.IsPrimitiveASTNode(prevBNE.RightNode) || prevBNE == null)
+                        && (CoreUtils.IsPrimitiveASTNode(prevBNE.RightNode) || prevBNE == null))
                     {
                         // An input node is not re-compiled and executed
                         // It is handled by the ChangeSetApply by re-executing the modified node with the updated changes
--- a/src/Engine/ProtoScript/Runners/LiveRunner.cs
+++ b/src/Engine/ProtoScript/Runners/LiveRunner.cs
@@ -1028,7 +1028,7 @@ namespace ProtoScript.Runners
                     //Check if the subtree (ie a node in graph) is an input and has primitive Right hand Node type, also
                     //ensure that the previous RHS of this assignment was a primitive value, or we'll have generated incorrect instructions.
                     if (redefinitionAllowed && st.IsInput && bne != null && CoreUtils.IsPrimitiveASTNode(bne.RightNode)
-                        && (CoreUtils.IsPrimitiveASTNode(prevBNE.RightNode) || prevBNE == null))
+                        && (prevBNE == null || CoreUtils.IsPrimitiveASTNode(prevBNE.RightNode)))
                     {
                         // An input node is not re-compiled and executed
                         // It is handled by the ChangeSetApply by re-executing the modified node with the updated changes
--- a/src/Engine/ProtoScript/Runners/LiveRunner.cs
+++ b/src/Engine/ProtoScript/Runners/LiveRunner.cs
@@ -1011,17 +1011,20 @@ namespace ProtoScript.Runners
                     subtree.ForceExecution = false;
 
                     BinaryExpressionNode prevBNE = null;
-                    if (node is BinaryExpressionNode bne1){
-                        var prevNode = st.AstNodes.OfType<BinaryExpressionNode>().Where(x => (x.LeftNode as IdentifierNode).Name == bne1.LeftNode.Name);
-                        if(prevNode.Count() > 1)
+                    var bne = node as BinaryExpressionNode;
+                    if (bne != null){
+                        foreach (var astNode in st.AstNodes)
                         {
-                            throw new Exception("1 prevnode expected");
+                            if(astNode is BinaryExpressionNode curBNE && curBNE.LeftNode is IdentifierNode id && id.Name == bne.LeftNode.Name)
+                            {
+                                prevBNE = curBNE;
+                            }
                         }
-                        prevBNE = prevNode.FirstOrDefault();
                     }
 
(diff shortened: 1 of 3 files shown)


src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -1010,8 +1010,19 @@ private List<AssociativeNode> GetModifiedNodes(Subtree subtree, bool redefinitio
...
+                        if(prevNode.Count() > 1)
+                        {
+                            throw new Exception("1 prevnode expected");
+                        }
+                        prevBNE = prevNode.FirstOrDefault();
+                    }
+
                     //Check if the subtree (ie a node in graph) is an input and has primitive Right hand Node type
-                    if (redefinitionAllowed && st.IsInput && node is BinaryExpressionNode bne  && CoreUtils.IsPrimitiveASTNode(bne.RightNode))
+                    if (redefinitionAllowed && st.IsInput && node is BinaryExpressionNode bne  && CoreUtils.IsPrimitiveASTNode(bne.RightNode)
```
aparajit-pratap, line 1024: If we don't want to make null node a primitive, we can remove the condition of bne.RightNode being a primitive I think.

src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -1010,8 +1010,19 @@ private List<AssociativeNode> GetModifiedNodes(Subtree subtree, bool redefinitio
...
 
+                    BinaryExpressionNode prevBNE = null;
+                    if (node is BinaryExpressionNode bne1){
+                        var prevNode = st.AstNodes.OfType<BinaryExpressionNode>().Where(x => (x.LeftNode as IdentifierNode).Name == bne1.LeftNode.Name);
+                        if(prevNode.Count() > 1)
+                        {
+                            throw new Exception("1 prevnode expected");
+                        }
+                        prevBNE = prevNode.FirstOrDefault();
+                    }
```
aparajit-pratap, line 1021: I think it's better to do this search here rather than outside the `if(!nodeFound)`, which was what I was doing.

src/Engine/ProtoScript/Runners/LiveRunner.cs
aparajit-pratap: I also think we can remove the `node is BinaryExpressionNode bne` condition as it is implied if `CoreUtils.IsPrimitiveASTNode(prevBNE.RightNode` is true

src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -1010,8 +1010,19 @@ private List<AssociativeNode> GetModifiedNodes(Subtree subtree, bool redefinitio
                     // It can then be handled normally regardless of its ForceExecution state
                     subtree.ForceExecution = false;
 
+                    BinaryExpressionNode prevBNE = null;
+                    if (node is BinaryExpressionNode bne1){
+                        var prevNode = st.AstNodes.OfType<BinaryExpressionNode>().Where(x => (x.LeftNode as IdentifierNode).Name == bne1.LeftNode.Name);
```
aparajit-pratap, line 1015: I would stick with using the `Value` property for IdentifierNode as it's specific to that node type while Name applies to the base `Node `type. May not make a difference but just for clarity sake.

mjkkirschner: okay, will do.

src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -1010,8 +1010,22 @@ private List<AssociativeNode> GetModifiedNodes(Subtree subtree, bool redefinitio
...
                     subtree.ForceExecution = false;
 
-                    //Check if the subtree (ie a node in graph) is an input and has primitive Right hand Node type
-                    if (redefinitionAllowed && st.IsInput && node is BinaryExpressionNode bne  && CoreUtils.IsPrimitiveASTNode(bne.RightNode))
+                    BinaryExpressionNode prevBNE = null;
+                    var bne = node as BinaryExpressionNode;
+                    if (bne != null){
+                        foreach (var astNode in st.AstNodes)
+                        {
+                            if(astNode is BinaryExpressionNode curBNE && curBNE.LeftNode is IdentifierNode id && id.Name == bne.LeftNode.Name)
```
aparajit-pratap, line 1018: minor but I would still use Value for identifier node as it's specific to that node type, I think you missed my previous comment to the same effect

src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -1010,8 +1010,22 @@ private List<AssociativeNode> GetModifiedNodes(Subtree subtree, bool redefinitio
...
+                            {
+                                prevBNE = curBNE;
+                            }
+                        }
+                    }
+
+                    //Check if the subtree (ie a node in graph) is an input and has primitive Right hand Node type, also
+                    //ensure that the previous RHS of this assignment was a primitive value, or we'll have generated incorrect instructions.
+                    if (redefinitionAllowed && st.IsInput && bne != null && CoreUtils.IsPrimitiveASTNode(bne.RightNode)
+                        && (CoreUtils.IsPrimitiveASTNode(prevBNE.RightNode) || prevBNE == null))
```
aparajit-pratap, line 1028: I thought the condition `prevBNE == null` needed to precede the second condition invoking the property on it, but I guess the compiler is smart enough to decide the order of evaluating the conditions :)

mjkkirschner: well, just in case, I'll fix that.

src/Engine/ProtoScript/Runners/LiveRunner.cs
mjkkirschner: should be addressed now.

aparajit-pratap (approved):

PR #13342, opened 2022-09-26
Dyn 5159 import validation

jesusalvino:
### Purpose
https://jira.autodesk.com/browse/DYN-5159
Adding Importing file validation

### Declarations

### Reviewers
@QilongTang 

### FYIs
@RobertGlobant20 @filipeotero

### DIFF
--- a/src/DynamoCore/Configuration/PreferenceSettings.cs
+++ b/src/DynamoCore/Configuration/PreferenceSettings.cs
@@ -951,12 +951,12 @@ namespace Dynamo.Configuration
         }
 
         /// <summary>
-        /// Indicates when an instance has been created from a preferences file correctly or not
+        /// Indicates when an instance has been created from a preferences file correctly, a support property
         /// </summary>
-        /// <returns></returns>
-        public bool IsCreatedFromValidFile()
+        [XmlIgnore]
+        public bool IsCreatedFromValidFile
         {
-            return this.isCreatedFromValidFile;
+            get { return isCreatedFromValidFile; }
         }
     }
 }
--- a/src/DynamoCore/Configuration/PreferenceSettings.cs
+++ b/src/DynamoCore/Configuration/PreferenceSettings.cs
@@ -52,7 +52,7 @@ namespace Dynamo.Configuration
         private bool isBackgroundGridVisible;
         private bool disableTrustWarnings = false;
         private bool isNotificationCenterEnabled;
-        private bool isValid = true;
+        private bool isCreatedFromValidFile = true;
 
         #region Constants
         /// <summary>
@@ -755,7 +755,7 @@ namespace Dynamo.Configuration
             {
                 if (settings == null)
                 {
-                    return new PreferenceSettings() { isValid = false };
+                    return new PreferenceSettings() { isCreatedFromValidFile = false };
                 }
             }
 
@@ -954,9 +954,9 @@ namespace Dynamo.Configuration
         /// Indicates when an instance has been created from a preferences file correctly or not
         /// </summary>
         /// <returns></returns>
-        public bool isValidInstance()
+        public bool IsCreatedFromValidFile()
         {
-            return this.isValid;
+            return this.isCreatedFromValidFile;
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -753,7 +755,7 @@ public static PreferenceSettings Load(string filePath)
             {
                 if (settings == null)
                 {
-                    return new PreferenceSettings();
+                    return new PreferenceSettings() { isValid = false };
```
mjkkirschner, line 758: this seems a little funny - if we do return a brand new `PreferenceSettings` object, it should be a valid preferences - just not the one that the user tried to load... maybe the bool needs a different name.

src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -52,6 +52,8 @@ public class PreferenceSettings : NotificationObject, IPreferences, IRenderPreci
         private bool isBackgroundGridVisible;
         private bool disableTrustWarnings = false;
         private bool isNotificationCenterEnabled;
+        private bool isValid = true;
```
mjkkirschner, line 55: why not just create an auto property, then you don't need to create the extra method below that is basically just a setter.

src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -947,5 +949,14 @@ public List<string> StaticFields()
         {
             return typeof(PreferenceSettings).GetMembers(BindingFlags.Static | BindingFlags.NonPublic).OfType<FieldInfo>().Select(field => field.Name).ToList();
         }
+
+        /// <summary>
+        /// Indicates when an instance has been created from a preferences file correctly or not
+        /// </summary>
+        /// <returns></returns>
+        public bool isValidInstance()
```
mjkkirschner, line 957: see above suggestion to just use properties.

test/DynamoCoreTests/Configuration/PreferenceSettingsTests.cs
```diff
@@ -320,6 +320,8 @@ public void TestImportCopySettings()
             var newSettings = PreferenceSettings.Load(newSettingslFilePath);
 
             // validation
+            Assert.IsTrue(newSettings.isValidInstance(), "The new settings file is invalid");
```
mjkkirschner, line 323: it would be good to have a negative test as well.

src/DynamoCore/Configuration/PreferenceSettings.cs
jesusalvino: Rename works, done.

src/DynamoCore/Configuration/PreferenceSettings.cs
jesusalvino: > why not just create an auto property, then you don't need to create the extra method below that is basically just a getter.
> 
> You can make the setter internal and the getter public.

Yh I was thinking about thatbut kindly: per my understanding, the reason of this class is serves as a mapping from each property to a setting xml file, that's the reason Why I didn't add and extra property just for validation aproach, please your thoughts.

test/DynamoCoreTests/Configuration/PreferenceSettingsTests.cs
jesusalvino: Done.

mjkkirschner:
Thanks for the explanation - you can use an attribute so that the property is ignored during serialization and deserialization- I think there are other examples of using it - it’s something like xml ignore - otherwise please add a comment explaining so no one makes a refactor with an unintended consequence.

src/DynamoCore/Configuration/PreferenceSettings.cs
jesusalvino: Done.

QilongTang:
Looks like the only failure was a fluke, restarted the job

QilongTang (approved):

QilongTang:
The new job finished without failures https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/7230/

PR #13354, opened 2022-09-29
Dyn4994 - part 1.5

mjkkirschner:
### Purpose

This attempts to add some graph tests to the performance folder. Unfortunately (or fortunately?) I found another couple issues while running these.

I'm not sure how to fix the most complex one - would like to discuss.

1. adds 2 graphs - one is quite long running - one still fails.
2. added some null checks in the test infrastructure so we can see null values in the results instead of the comparison blowing up.
3. found a simple case where importing a method with the signature `object FirstItem(IList a)` should be imported as `def FirstItem(a:var[]..[]):var[]..[]` - but is imported as `def FirstItem(a:var[]..[]):var` - this causes rank reduction errors at return.

### Declarations

### Release Notes

### Reviewers

### FYIs

### DIFF
--- a/src/Engine/EmitMSIL/Replication.cs
+++ b/src/Engine/EmitMSIL/Replication.cs
@@ -568,6 +568,8 @@ namespace EmitMSIL
             var marshaller = ProtoFFI.CLRDLLModule.GetMarshaler(runtimeCore);
             CLRStackValue dsRetValue = marshaller.Marshal(result, ProtoFFI.CLRObjectMarshaler.GetProtoCoreType(finalFep.ReturnType), runtimeCore);
 
+            //TODO this causes a test failure. (array reduction to var not allowed)
+            //this is always false, and the comment below does not make sense.
             if (!dsRetValue.IsExplicitCall)
             {
                 // An explicit call requires return coercion at the return instruction
--- a/src/Engine/EmitMSIL/Replication.cs
+++ b/src/Engine/EmitMSIL/Replication.cs
@@ -243,7 +243,7 @@ namespace EmitMSIL
             object result;
             if (replicationInstructions.Count == 0)
             {
-                result = ExecWithZeroRI(finalFep, stackValues, runtimeCore);
+                result = ExecWithZeroRI(finalFep, stackValues, runtimeCore,true);
             }
             else //replicated call
             {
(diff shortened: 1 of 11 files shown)


src/Engine/EmitMSIL/Replication.cs
```diff
@@ -568,9 +568,12 @@ private static object ExecWithZeroRI(CLRFunctionEndPoint finalFep, List<CLRStack
             var marshaller = ProtoFFI.CLRDLLModule.GetMarshaler(runtimeCore);
             CLRStackValue dsRetValue = marshaller.Marshal(result, ProtoFFI.CLRObjectMarshaler.GetProtoCoreType(finalFep.ReturnType), runtimeCore);
 
-            if (!dsRetValue.IsExplicitCall)
+            //TODO this causes a test failure. (array reduction to var not allowed)
+            //this is always false, and the comment below does not make sense.
+
+            // An explicit call requires return coercion at the return instruction
+            if (!isExplicitCall)
```
aparajit-pratap, line 575: I don't believe this concept of something being an explicit vs. an implicit call is relevant to us in the new engine anymore and strongly feel that we should avoid using this flag. I remember trying to work with this idea long ago when I was working on DesignScript but not touching it since then.

aparajit-pratap: It appears that it was originally used to distinguish between regular function calls vs. internal function calls like GC.

mjkkirschner: thanks for the info @aparajit-pratap - it turns out atleast for this problem identified here the real issue is that this function should return `var[]..[]` but that the FEP says it returns `var` - I think @pinzart90 will address it in his type system import work.

in the meantime, I've marked some of these tests failures, which should turn green after Tibi's work - added 2 graph perf cases (1 failing for the aforementioned reason), and re-enabled a few of them that started passing!

mjkkirschner: I do think I could construct a failing test case for the IsExplicitIssue though using the FFITarget ... maybe I will give that a shot so we have it tracked and can discuss how it should really be solved.

pinzart90 (approved):

src/Engine/EmitMSIL/Replication.cs
mjkkirschner: actually, I can't create a test case for it that doesn't also require returning `object`

sm6srw (approved):

PR #13361, opened 2022-09-30
expose CLRFunctionEndpoint to Compiler

pinzart90:
Expose CLRFunctionEndpoint to the rest of the VM
Do partial function resolution at compile time
Resolve all types by hooking into the old VM's type system (temporary solution)


src/Libraries/DesignScriptBuiltin/RangeHelpers.cs
```diff
@@ -31,7 +31,7 @@ public enum RangeStepOperator
 namespace Builtin
 {
     //TODO for now suppress import into DS VM.
-    [SupressImportIntoVM]
+    [IsVisibleInDynamoLibrary(false)]
```
pinzart90, line 34: @mjkkirschner I exposed these to the VM as a shortcut to getting access to these types from the same place (the old vm's type system)

src/Engine/ProtoCore/RuntimeCore.cs
```diff
@@ -424,5 +436,36 @@ internal int GetCoercionScore(int srcType, int targetType)
         {
             return ClassTable?.ClassNodes[srcType]?.GetCoercionScore(targetType) ?? (int)ProcedureDistance.NotMatchScore;
         }
+
+        internal FunctionGroup GetFuncGroup(string methodName, string className)
+        {
+            int classScope = ClassTable.IndexOf(className);
+            // try to use dynamic classScope
+            /*if (arguments.Count > 0)
```
pinzart90, line 444: this commented part is supposed to be called at runtime...which we do not do anymore

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -20,7 +22,8 @@ public class CodeGenIL : IDisposable
         internal string className;
         internal string methodName;
         private IDictionary<string, IList> input;
-
+        public ProtoCore.MSILRuntimeCore runtimeCore;
```
mjkkirschner, line 25: does this need to be public?

pinzart90: It does not. A lot of the new APIs do not need to be.
Making it here public made it easier to test out.

I can either make everything internal (or most of it)  ...or leave it as is and create a separate jira task for it

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -32,7 +35,7 @@ public class CodeGenIL : IDisposable
         /// </summary>
         private Dictionary<int, Type> astTypeInfoMap = new Dictionary<int, Type>();
         private StreamWriter writer;
-        private Dictionary<int, IEnumerable<MethodBase>> methodCache = new Dictionary<int, IEnumerable<MethodBase>>();
+        private Dictionary<int, IEnumerable<ProtoCore.CLRFunctionEndPoint>> methodCache = new Dictionary<int, IEnumerable<ProtoCore.CLRFunctionEndPoint>>();
```
mjkkirschner, line 38: I need to stare at this for a while - but let's say that we want to emit some IL that directly calls a function, without wrapping the return types in CLRStackValue and without using replication logic at runtime - does this change stop us from doing that?

pinzart90: Not really. It would just mean to call other APIs on the  ProtoCore.CLRFunctionEndPoint class instead of using MethodBase

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -57,13 +60,14 @@ public static int KeyGen(string className, string methodName, int numParameters)
...
-        public CodeGenIL(IDictionary<string, IList> input, string filePath)
+        public CodeGenIL(IDictionary<string, IList> input, string filePath, ProtoCore.MSILRuntimeCore runtimeCore)
         {
             this.input = input;
             writer = new StreamWriter(filePath);
+            this.runtimeCore = runtimeCore;
         }
 
-        public IDictionary<string, object> Emit(List<AssociativeNode> astList)
+        public IDictionary<string, object> Emit(List<AssociativeNode> astList, ProtoCore.MSILRuntimeCore runtimeCore)
```
aparajit-pratap, line 70: Why does this function need to be passed the runtime core, isn't already saved as a class member?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -115,7 +119,7 @@ public CodeGenIL(IDictionary<string, IList> input, string filePath)
             // 4. Create method ("Execute"), get ILGenerator 
             var execMethod = BuilderHelper.CreateMethod(type, "Execute",
                 System.Reflection.MethodAttributes.Static, typeof(void), new[] { typeof(IDictionary<string, IList>),
-                typeof(IDictionary<int, IEnumerable<MethodBase>>), typeof(IDictionary<string, object>)});
+                typeof(IDictionary<int, IEnumerable<ProtoCore.CLRFunctionEndPoint>>), typeof(IDictionary<string, object>), typeof(ProtoCore.MSILRuntimeCore)});
```
aparajit-pratap, line 122: Can we just add a `using ProtoCore` to avoid prefixing the namespace everywhere and make it less verbose?

pinzart90: that would cause a lot of conflicts with Type (System.Type and ProtoCore.Type, and others I think)

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -463,85 +467,127 @@ private Type EmitIListCoercion<T>(AssociativeNode arg)
             return t;
         }
 
-        private IEnumerable<MethodBase> FunctionLookup(IList args)
+        private IEnumerable<ProtoCore.CLRFunctionEndPoint> FunctionLookup(IList args)
         {
-            IEnumerable<MethodBase> mbs = null;
+            List<ProtoCore.CLRFunctionEndPoint> mbs = new List<ProtoCore.CLRFunctionEndPoint>();
```
aparajit-pratap, line 472: `var mbs = ...`

src/Engine/EmitMSIL/CodeGenIL.cs
pinzart90: will remove it

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -463,85 +467,127 @@ private Type EmitIListCoercion<T>(AssociativeNode arg)
...
+
+            int cid = runtimeCore.ClassTable.GetClassId(className);
+            if (!CoreUtils.IsInternalMethod(methodName))
             {
-                if (!CoreUtils.IsInternalMethod(methodName))
+                Validity.Assert(cid != DSASM.Constants.kInvalidIndex);
+            }
+
+            // Can this be used at compile time ? or does this need to be at runtime ? 
+            var fg = runtimeCore.GetFuncGroup(methodName, className);
```
aparajit-pratap, line 486: Can you explain more what you mean by this comment?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -463,85 +467,127 @@ private Type EmitIListCoercion<T>(AssociativeNode arg)
...
-                        //{
-                        //    argTypes = method.GetParameters().Select(p => p.ParameterType).ToList();
-                        //    return method.ReturnType;
-                        //}
-                    }
+                // Try to find a setter
+                if (fg == null)
+                {
+                    methodName = ProtoCore.DSASM.Constants.kSetterPrefix + methodName;
+                    fg = runtimeCore.GetFuncGroup(methodName, className);
```
aparajit-pratap, line 504: No harm with this change but I thought we didn't support setters. At least I haven't seen any being used.

src/Engine/EmitMSIL/CodeGenIL.cs
pinzart90: The original method (`CallSite::GetFuncGroup`) also handled FunctionGroup lookup by inspecting the arguments (stackValues), which only makes sense at runtime.
I converted the original method for usage at compile time only.

Basically I hope I am not missing any cases by dismissing the arguments analysis.
https://github.com/DynamoDS/Dynamo/blob/master/src/Engine/ProtoCore/Lang/CallSite.cs#L1070

src/Engine/EmitMSIL/CodeGenIL.cs
pinzart90: For some reason I thought they were handled before..(but that may just have been the getters)
I can remove it if that is the case

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -463,85 +467,127 @@ private Type EmitIListCoercion<T>(AssociativeNode arg)
...
+                var procNode = funcEnd.procedureNode;
+                Validity.Assert(procNode != null, "Expected to have a valid procedureNode");
+
+                bool isStaticOrConstructorOrInternal = procNode.IsStatic || procNode.IsConstructor || CoreUtils.IsInternalMethod(methodName);
+
+                bool matchingArgCount = isStaticOrConstructorOrInternal ?
+                    procNode.ArgumentTypes.Count == args.Count :
+                    (procNode.ArgumentTypes.Count + 1) == args.Count;
+
+                if (!matchingArgCount) {
```
aparajit-pratap, line 521: In which cases will `matchingArgCount `be `false`?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -463,85 +467,127 @@ private Type EmitIListCoercion<T>(AssociativeNode arg)
...
+                    parameterInfos = clrFFI.ReflectionInfo.GetParameters().Select(x => x.Info).ToArray();
+                    declaringType = clrFFI.ReflectionInfo.DeclaringType;
+                    fFIMemberInfo = clrFFI.ReflectionInfo;
+                }
+                else
+                {
+                    throw new NotImplementedException("Unkown FunctionEndpoint type. Not implemented yet");
+                }
+
+                List<ProtoCore.CLRFunctionEndPoint.ParamInfo> formalParams = new List<ProtoCore.CLRFunctionEndPoint.ParamInfo>();
```
aparajit-pratap, line 559: `var formalParams = `

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: @pinzart90 - for methods that are statically typed shouldn't we be able to lookup the correct function group just using types?

src/Engine/EmitMSIL/CodeGenIL.cs
pinzart90: image

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1201,16 +1248,16 @@ private Type EmitFunctionCallNode(AssociativeNode node)
             var keygen = typeof(CodeGenIL).GetMethod(nameof(CodeGenIL.KeyGen));
             EmitOpCode(OpCodes.Call, keygen);
 
-            var local = DeclareLocal(typeof(IEnumerable<MethodBase>), "cached MethodBase objects");
+            var local = DeclareLocal(typeof(IEnumerable<ProtoCore.CLRFunctionEndPoint>), "cached MethodBase objects");
```
aparajit-pratap, line 1251: update comment to "cached CLR function endpoints"

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: Oh I see, it's not what I was thinking - if I'm understanding that function correctly it's trying to find the first class in the inheritance chain that matches one of the non array arguments...

We should probably add some tests that try to call methods from one class on another derived object.

src/Engine/EmitMSIL/ExecuteIL.cs
```diff
@@ -60,7 +60,7 @@ class ExecuteIL
             key = CodeGenIL.KeyGen(className, methodName, args.Count);
             mi = methodCache[key];
             // 6. Emit call to ReplicationLogic by passing input args from previous steps
-            var d = Replication.ReplicationLogic(mi, args, guides);
+            var d = Replication.ReplicationLogic(null, args, guides, null);
```
aparajit-pratap, line 63: This code was initially written as an example so I could wrap around my head over what was being emitted by replicating it as actual C# code. Now it looks like this example isn't relevant anymore unless we want to instantiate CLRFunctionEndPoints and all those types. So I suggest removing it if it doesn't hold value anymore.

pinzart90: I can remove the entire File and the Program.cs file too

src/Engine/ProtoCore/DSASM/InstructionSet.cs
```diff
@@ -113,7 +113,7 @@ public struct MetaData
     /// <summary>
     /// Wrapper over all values handled by the MSIL/CLR engine.
     /// </summary>
-    internal struct CLRStackValue
+    public struct CLRStackValue
```
aparajit-pratap, line 116: We should keep everything as `internal` and just expose internals to other assemblies. Don't want end users to access any crap from `ProtoCore`, etc.

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: Looking at the history of this function, it was to implement dynamic dispatch... I agree that is probably not possible at compile time.

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -1094,14 +1096,14 @@ private FunctionGroup GetFuncGroup(RuntimeCore runtimeCore, List<StackValue> arg
...
         }
 
-        private FunctionGroup FirstFunctionGroupInInheritanceChain(RuntimeCore runtimeCore, int cidx)
+        internal static FunctionGroup FirstFunctionGroupInInheritanceChain(string methodName, int cidx, ClassTable classTable, FunctionTable globalFnTable)
         {
             FunctionGroup funcGroup = null;
-            var classNodes = runtimeCore.DSExecutable.classTable.ClassNodes;
-            var globalFuncTable = globalFunctionTable.GlobalFuncTable;
+            var classNodes = classTable.ClassNodes;
+            var globalFuncTable = globalFnTable.GlobalFuncTable;
```
aparajit-pratap, line 1106: What's the real advantage of making this function `static `or is it just cosmetic?

src/Engine/EmitMSIL/CodeGenIL.cs
pinzart90: var pnt = Point.ByCoordinates(0,0,0);
Geometry.Translate(pnt, 1,0,0);

Translate should be found as a Point method not a Geometry method ?
Not sure I understand the use case very well

src/Engine/ProtoCore/Lang/CallSite.cs
pinzart90: it is just a refactor so that it can be shared by the MSIL APIs

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: If point implemented its own `Translate` method - but it does not.

There are some examples @aparajit-pratap wrote up for 2.0 release:
https://github.com/DynamoDS/Dynamo/wiki/Dynamo-2.0-Language-Changes-Explained#solved-issues

src/Engine/ProtoCore/RuntimeCore.cs
aparajit-pratap: Are you saying this commented code is no longer required to be called at runtime in the new VM? If so, can we remove it or do we need to make sure before we can do so?

pinzart90: I was hoping we could use this API only at compile time. 
But it looks like there might be regressions with polymorphism if we skip the commented code

I might have to move it all at runtime again...

test/CodeGenILTests/IndexingMicroTests.cs
```diff
@@ -11,6 +11,14 @@ namespace CodeGenILTests
     [TestFixture]
     public class IndexingMicroTests : MicroTests
     {
+        protected override void GetLibrariesToPreload(ref List<string> libraries)
+        {
+            base.GetLibrariesToPreload(ref libraries);
+            libraries.Add("DesignScriptBuiltin.dll");
+            libraries.Add("DSCoreNodes.dll");
```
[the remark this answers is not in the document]
pinzart90: these are specified for the old VM's RuntimeCore.
The old vm has to preload all types so that the new VM can borrow them

test/CodeGenILTests/RangeMicroTests.cs
```diff
@@ -11,21 +11,39 @@ namespace CodeGenILTests
...
         [SetUp]
         public void Setup()
         {
+            liveRunner = new ProtoScript.Runners.LiveRunner();
+
+            List<string> libraries = new List<string>();
+            GetLibrariesToPreload(ref libraries);
+
+            liveRunner.ResetVMAndResyncGraph(libraries);
+            runtimeCore = new ProtoCore.MSILRuntimeCore(liveRunner.RuntimeCore);
```
aparajit-pratap, line 35: Is this a good long-term strategy to establish reliance on the old VM? It looks like it will be harder to split away from it later when we want to ditch it completely or ditch parts of it.

test/CodeGenILTests/ReplicationTests.cs
```diff
@@ -8,6 +8,14 @@ namespace CodeGenILTests
 {
     class ReplicationTests : MicroTests
     {
+        protected override void GetLibrariesToPreload(ref List<string> libraries)
+        {
+            base.GetLibrariesToPreload(ref libraries);
+            libraries.Add("DesignScriptBuiltin.dll");
+            libraries.Add("DSCoreNodes.dll");
```
aparajit-pratap, line 15: Already loaded in base.

test/CodeGenILTests/RangeMicroTests.cs
pinzart90: Well...if we do go with the new VM,  we will have to implement the EmitFunctionDefinition and EmitImportStatements and all the preloading logic.
This is a temporary solution, a shortcut to get access to all the loaded types

test/CodeGenILTests/IndexingMicroTests.cs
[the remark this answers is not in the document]
pinzart90: Oh..I get it now.
Will remove redundancies

src/Engine/ProtoCore/RuntimeCore.cs
pinzart90: We can also figure out polymorphism later on (probably will have to move some of the method resolution to runtime).
If this looks good enough for the go/no go testing, I can add a separate jira task to figure out polymorphism

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: I think long term, we'll need to find some way of both doing the most at compile time, and then falling back to runtime behavior as well, can this function be used in both places.
At compile time - we use this function as you are doing here, and then at runtime we use the version that checks the runtime args to potentially switch to another method?

test/CodeGenILTests/RangeMicroTests.cs
mjkkirschner: can we file a task to implement `EmitFunctionDefinition` and `EmitImportStatements`, add a TODO to the code, and replace these calls?

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -86,44 +87,14 @@ internal static List<object> UnmarshalFunctionArguments(List<CLRFunctionEndPoint
...
-                    formalParams.Add(new CLRFunctionEndPoint.ParamInfo() { CLRInfo = param, ProtoInfo = dsType });
-                }
-
-                CLRFunctionEndPoint fep = new CLRFunctionEndPoint(mInfo, formalParams);
-                
-                feps.Add(fep);
-            }
-
-            return feps;
-        }
```
aparajit-pratap, line 125: So this code has now basically been moved up to `FunctionLookup `at compile time?

pinzart90: yes

test/CodeGenILTests/IndexingMicroTests.cs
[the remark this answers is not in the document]
aparajit-pratap: `base.GetLibrariesToPreload(ref libraries);` is the call the base class, MicroTests' `GetLibrariesToPreload `function that already is loading these assemblies or am I missing something?

test/CodeGenILTests/RangeMicroTests.cs
pinzart90: will do

src/Engine/EmitMSIL/CodeGenIL.cs
pinzart90: sounds like a good approach to me.

src/Engine/ProtoCore/RuntimeCore.cs
aparajit-pratap: Yes, I feel some part of method resolution will always be needed at runtime. For now, we can create a task and defer it.

src/Engine/ProtoCore/Lang/FunctionEndPoint.cs
```diff
@@ -6,73 +6,71 @@
...
 using ProtoCore.Utils;
+using ProtoFFI;
 
 namespace ProtoCore
 {
     /// <summary>
     /// Wrapper over MethodBase.
     /// </summary>
-    internal sealed class CLRFunctionEndPoint
+    public sealed class CLRFunctionEndPoint
```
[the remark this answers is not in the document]
pinzart90: It is a bit more complicated to use internals with dynamically generated assemblies. 
I would rather not have to do that now. https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/w58ww7se(v=vs.100)?redirectedfrom=MSDN 
I only exposed the bare minimum as public.

src/Engine/ProtoCore/DSASM/InstructionSet.cs
pinzart90: see comment below

src/Engine/ProtoCore/Lang/FunctionEndPoint.cs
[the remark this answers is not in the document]
aparajit-pratap: I'm not sure what this means. Where does it specify that somethings needs to be public?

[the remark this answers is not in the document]
pinzart90: We are creating a new assembly + type + function here
https://github.com/DynamoDS/Dynamo/blob/emitMSIL/src/Engine/EmitMSIL/CodeGenIL.cs#L110-L119

This new method will try to access types from other assemblies (ProtoCore, etc). 
In normal situations we would use `InternalsVisibleTo` ...but that only works if the assembly has a strong name.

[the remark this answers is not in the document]
pinzart90: I guess another way would be to move CLRStackValue and CLRFunctionEndpoint (and everything that uses them) to the EmitMSIL project...but that might mean more work and potentially more code duplication

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -13,6 +13,7 @@
 
 namespace EmitMSIL
 {
+    [Obsolete("This is an internal class, do not use it.")]
     public class Replication
```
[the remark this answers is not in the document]
aparajit-pratap: Oh, sorry, this doesn't belong to the dynamic assembly either so whatever holds true for ProtoCore, holds true for EmitMSIL assembly as well.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -463,85 +467,118 @@ private Type EmitIListCoercion<T>(AssociativeNode arg)
...
-                    var assemblies = modules.Select(m => m.Assembly ?? (m.Module?.Assembly)).Where(m => m != null);
-                    foreach (var asm in assemblies)
+                    var mInfo = BuiltIn.GetInternalMethod(methodName);
+                    parameterInfos = mInfo.GetParameters();
+                    declaringType = mInfo.DeclaringType;
+                    fFIMemberInfo = new FFIMethodInfo(mInfo);
+                }
+                else if (funcEnd is FFIFunctionEndPoint ffiFep)
+                {
+                    FFIHandler handler = FFIFunctionEndPoint.FFIHandlers[ffiFep.activation.ModuleType];
```
mjkkirschner, line 529: is this ever anything but clr ffi?

src/Engine/ProtoCore/FFI/CLRFFIFunctionPointer.cs
```diff
@@ -267,6 +269,9 @@ public FFIMemberInfo ReflectionInfo
             private set;
         }
 
+        internal ProtoCore.Type[] ArgumementTypes => mArgTypes;
```
mjkkirschner, line 272: maybe you could a comment about why these getters exist?

src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -1270,7 +1270,12 @@ internal void CompileAndExecuteMSIL(List<AssociativeNode> finalDeltaAstList)
         {
             Dictionary<string, IList> input = new Dictionary<string, IList>();
             var assemblyPath = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location));
-            var codeGenIL = new EmitMSIL.CodeGenIL(input, Path.Combine(assemblyPath, "opCodes.txt"));
+
+            //TODO_MSIL: remove the dependency on the old VM by implementing
+            //necesary Emit functions(ex mitFunctionDefinition and EmitImportStatements and all the preloading logic)
+            msilRuntimeCore = msilRuntimeCore ?? new MSILRuntimeCore(runtimeCore);
```
mjkkirschner, line 1276: is there a jira link?

test/CodeGenILTests/ReplicationTests.cs
```diff
@@ -488,7 +492,7 @@ public void MSIL_Replication_BuiltinMethods()
             var output = codeGen.EmitAndExecute(ast);
             Assert.IsNotEmpty(output);
 
-            Assert.AreEqual(output["test1"], new Object[] { true, true });
+            Assert.AreEqual(output["test1"], false);
```
mjkkirschner, line 495: you verified this is the correct result?

mjkkirschner (approved):
looks good to me with a few comments.

test/CodeGenILTests/ReplicationTests.cs
pinzart90: yes

src/Engine/EmitMSIL/CodeGenIL.cs
pinzart90: that is what I've seen during testing. 
Not sure what other edge cases there could be

src/Engine/ProtoScript/Runners/LiveRunner.cs
pinzart90: not yet...I will add the to the PR description

PR #13367, opened 2022-10-03
Handle no-replication cases

aparajit-pratap:
### Purpose

This PR has the following changes:
1. Fixes local var index issues by relying solely on the `localbuilder`'s `localindex` returned by calling `ilGen.DeclareLocal` (which declares a new local variable) instead of incrementing a counter
2. Compares parameter-argument AST to determine if a function call is not replicating and emits direct call to function rather than redirecting the call via `ReplicationLogic`. 
3. Avoids saving dynamic assembly to disk for the test framework

TODOs:
- The call to `WillCallReplicate` happens twice, once during the `GatherTypeInfo` pass, and secondly, during the `emitIL` pass, which is redundant. I tried to cache the result into a dictionary with the AST ID as key but ran into issues as some ASTs are built during the emission steps (again redundantly) due to which their IDs change between compile passes.
- There are still some test scenarios that end up in a non-replicated call to `ExecWithZeroRI` inside of `ReplicationLogic `at runtime. These need to be investigated. This can be done by simply uncommenting the `Validity `assertion and seeing which tests fail, then debugging them.

### Declarations

### Release Notes

### Reviewers

### FYIs


src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -545,15 +594,16 @@ private IEnumerable<MethodBase> FunctionLookup(IList args)
             return mbs;
         }
 
-        private static HashSet<Type> GetTypeStatisticsForArray(ExprListNode array)
+        private HashSet<Type> GetTypeStatisticsForArray(ExprListNode array, ref int rank)
```
mjkkirschner, line 597: do you really want a `ref` type - wouldn't `out` suffice? I think you are only passing data out the method, not in both directions?

mjkkirschner: hmm, maybe I am wrong, I missed `rank += 1;`

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -888,10 +941,15 @@ private Type EmitBinaryExpressionNode(AssociativeNode node)
                         // variable being assigned already exists in dictionary.
                         throw new Exception("Variable redefinition is not allowed.");
                     }
-                    variables.Add(lNode.Value, new Tuple<int, Type>(++localVarIndex, t));
+                    variables.Add(lNode.Value, new Tuple<int, Type>(-1, t));
```
mjkkirschner, line 944: is this change made because you want to always use the index defined by the local builder?

aparajit-pratap: We still need to cache the local var index in this dictionary as we need to query it for the index in two places: in `EmitArrayCoercion` and in `EmitIEnumerableCoercion`.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1227,53 +1254,269 @@ private Type EmitFunctionCallNode(AssociativeNode node)
...
+
+        private static int GetEnumerableRank(Type type)
+        {
+            var genericArgs = type.GetGenericArguments();
+            if (genericArgs.Length == 0) return 0;
+
+            return 1 + GetEnumerableRank(genericArgs.FirstOrDefault());
+        }
+
+        private static int GetReductionDepth<T>(T val, Func<T, IEnumerable> asArr)
```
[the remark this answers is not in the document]
aparajit-pratap: I can rename it, it also returns the rank of something that can be an array. In this case, I'm using it only for an expression list associative AST so I could also turn it into a function specific to finding the rank of an AST rather than making it a generic function.

aparajit-pratap:
> @aparajit-pratap - so if I am understanding these changes correctly, we should never make it into this if statement: https://github.com/DynamoDS/Dynamo/blob/emitMSIL/src/Engine/EmitMSIL/Replication.cs#L213
> 
> maybe we could throw an exception or assert here to see if some cases still do?

@mjkkirschner I would not remove that `if` statement at this point. There can be some no-replication cases that are not caught at compile time due to inadequate type information, for e.g., if the arg type just shows up as an `object` at compile time, then I'm simply falling back to redirect the call to runtime `ReplicationLogic`.

mjkkirschner:
> > @aparajit-pratap - so if I am understanding these changes correctly, we should never make it into this if statement: https://github.com/DynamoDS/Dynamo/blob/emitMSIL/src/Engine/EmitMSIL/Replication.cs#L213
> > maybe we could throw an exception or assert here to see if some cases still do?
> 
> @mjkkirschner I would not remove that `if` statement at this point. There can be some no-replication cases that are not caught at compile time due to inadequate type information, for e.g., if the arg type just shows up as an `object` at compile time, then I'm simply falling back to redirect the call to runtime `ReplicationLogic`.

Hi @aparajit-pratap - definitely - I don't think we should remove the if statement either, I was suggesting adding an assert or analytics log there so we are made aware of cases that still his this if statement so we can improve compile time identification of those cases.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -201,135 +210,174 @@ private Type EmitCoercionCode(AssociativeNode arg, Type argType, Type param)
...
-            EmitOpCode(OpCodes.Clt);
+            EmitOpCode(OpCodes.Ldloc, enumeratorIndex);
 
-            EmitOpCode(OpCodes.Brtrue_S, loopBodyLabel.Value);
+            var finallyLabel = DefineLabel();
+            EmitOpCode(OpCodes.Brfalse_S, finallyLabel.Value);
+
+            EmitOpCode(OpCodes.Ldloc, enumeratorIndex);
+
+            mInfo = typeof(IDisposable).GetMethod(nameof(IDisposable.Dispose));
```
mjkkirschner, line 362: what is going on here? What needs to be disposed during coercion?

test/CodeGenILTests/ReplicationTests.cs
```diff
@@ -724,5 +720,43 @@ public void Value_Are_Correct_Reversed()
...
+            ";
+            var ast = ParserUtils.Parse(code).Body;
+            var output = codeGen.EmitAndExecute(ast);
+            Assert.IsNotEmpty(output);
+            Assert.AreEqual(new int[] { 1, 2, 4, 4 }, output["test1"]);
+            Assert.AreEqual(11, output["test2"]);
+        }
+
+        [Test]
+        public void IList_T_IEnumerable_Coerce()
```
mjkkirschner, line 744: perhaps it's now worth adding a test that asserts that the generated IL does not include any call instructions to ReplicationLogic?

src/Engine/EmitMSIL/CodeGenIL.cs
aparajit-pratap: This code is emitting the following `foreach ` loop and we need to dispose the enumerator used in the loop.
```
        var len = a.Count();    // where a is an IEnumerable<Source>
	var res = new Target[len];
	int c = 0;
	foreach(var i in a)
	{
		res[c] = i;
		c++;
	}
```

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1225,96 +1296,276 @@ private Type EmitFunctionCallNode(AssociativeNode node)
...
+                argIndex++;
+            }
+            return true;
+        }
 
-            // Emit call to load the runtimeCore argument
-            EmitOpCode(OpCodes.Ldarg_3);
+        private static int GetRank(Type type)
+        {
+            return type.IsArray ? GetArrayRank(type) : GetEnumerableRank(type);
```
pinzart90, line 1510: only for arrays ? not enumerables ?

aparajit-pratap: For enumerables, I use `GetEnumerableRank`.

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -210,6 +210,10 @@ public static object ReplicationLogic(List<CLRFunctionEndPoint> feps, IList args
             object result;
             if (replicationInstructions.Count == 0)
             {
+                // TODO: Ideally, we should not reach here as no-replication cases should ideally be all handled
+                // at compile-time. Uncomment the following Validity.Assert to debug these remaining cases.
+                //Validity.Assert(replicationInstructions.Count != 0,
+                //    "No-replication case, not expected to be handled inside ReplicationLogic function.");
```
mjkkirschner, line 216: @aparajit-pratap nice - should this be a jira task?

mjkkirschner (approved):

src/Engine/EmitMSIL/Replication.cs
aparajit-pratap: Will file it.

PR #13372, opened 2022-10-04
replication cases

pinzart90:
Implement the remaining cases of runtime replication.
Could not find valid tests for case 6.
New jira items created to figure out:
1. The this pointer context (when calling instance methods  ex. classInstance.DosSomething())
2. Global functions


src/Engine/EmitMSIL/Replication.cs
```diff
@@ -199,12 +200,25 @@ public static object ReplicationLogic(List<CLRFunctionEndPoint> feps, IList args
             //Turn the replication guides into a guide -> List args data structure
             var partialInstructions = Replicator.BuildPartialReplicationInstructions(partialReplicationGuides);
 
+            // TODO_MSIL: Figure it the best way to interpret/pass around the this pointer.
+            CLRStackValue thisPtr = CLRStackValue.Null;
+            bool staticFuncCall = feps.All(x => x.IsStatic);
```
mjkkirschner, line 205: is this existing logic or new - couldn't we have two candidate feps in the same class, one static and one not?

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -280,7 +294,29 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
...
+            CLRStackValue thisPtr, Dictionary<CLRFunctionEndPoint, int> candidatesWithDistances)
+        {
+            var candidateFunctions = new List<CLRFunctionEndPoint>();
+            foreach (var fep in candidatesWithDistances.Keys)
+            {
+                bool isClassInstanceMethod = fep.procedureNode != null &&
+                  fep.procedureNode.ClassID != Constants.kInvalidIndex &&//valid class
+                  !fep.procedureNode.IsConstructor && !fep.procedureNode.IsStatic;//not static and not constructor  
+
+                if (thisPtr.IsGlobal && isClassInstanceMethod)
```
mjkkirschner, line 307: huh? Can you explain, whats a global pointer in this context?

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -315,7 +351,114 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
...
+            {
+                return exactFeps.First();
+            }
+
+            //Walk the class tree structure to find the method
+            while (runtimeCore.ClassTable.ClassNodes[typeID].Base != Constants.kInvalidIndex)
+            {
+                typeID = runtimeCore.ClassTable.ClassNodes[typeID].Base;
+
+                foreach (var fep in feps)
```
mjkkirschner, line 401: can you add `{`

mjkkirschner: how does this logic factor into this comment:
`(most derived class should have priority).`

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -1286,15 +1286,19 @@ internal static FunctionGroup FirstFunctionGroupInInheritanceChain(string method
         private List<FunctionEndPoint> GetCandidateFunctions(StackFrame stackFrame,
```
mjkkirschner, line 1286: do we actually use this somewhere? How do we get stackframes?

src/Engine/EmitMSIL/Replication.cs
pinzart90: THis is new logic.
I do not think we can have mixed static/non-static function endpoints at the same time. 

Seems to me like:
Function calls ClassName.FunctionName should only resolve to static functions.
Function calls classInstance.Function should only resolve to non-static functions.
Please correct me if I am wrong. @mjkkirschner @aparajit-pratap 

I guess static-ness vs non-static-ness should be something we figure out at the compile stage. But it seemed that we could infer it at runtime replication stage too.

src/Engine/EmitMSIL/Replication.cs
pinzart90: thisPtr.IsGlobal is my attempt to represent the global function scope (temporary solution so that we can track the old VM functionality). It is false by default for now.

It is supposed to represent design script function definitions (the global scope)
ex
CodeBlock
```
def funcA(a:int) {
return a+a;
}
```
I can remove it too and we can just keep a TODO or something ....

src/Engine/ProtoCore/Lang/CallSite.cs
pinzart90: I made these changes because I found the conditions very confusing.
I hope my changes makes it a bit clearer

src/Engine/EmitMSIL/Replication.cs
pinzart90: this entire function has been copy pasted (and retrofitted) to the new VM.
For the most derived class priority :
I think the initial check  `feps.Where(x => x.procedureNode.ClassID == typeID);` represents "the most derived class".
If no exact match is found, then we go through the base classes

sm6srw (approved):

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -199,12 +200,25 @@ public static object ReplicationLogic(List<CLRFunctionEndPoint> feps, IList args
             //Turn the replication guides into a guide -> List args data structure
             var partialInstructions = Replicator.BuildPartialReplicationInstructions(partialReplicationGuides);
 
+            // TODO_MSIL: Figure it the best way to interpret/pass around the this pointer.
```
mjkkirschner, line 203: figure out (spelling)?

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -315,14 +351,121 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
             return compliantTarget;
         }
 
+        private static CLRFunctionEndPoint SelectFEPFromMultiple(
```
mjkkirschner, line 354: for these really big functions ... basically all the functions in this file, I really would not mind a summary :) - even if the methods are internal, trying to remember what they all do is not trivial for me.

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -407,13 +545,50 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
...
                 if (lowestCostFeps.Count > 0)
-                    // TODO_MSIL: implement SelectFEPFromMultiple
-                    compliantTarget = lowestCostFeps[0];//SelectFEPFromMultiple(stackFrame, runtimeCore, lowestCostFeps, formalParams);
+                    compliantTarget = SelectFEPFromMultiple(thisPtr, runtimeCore, lowestCostFeps, formalParams);
             }
             return compliantTarget;
         }
 
-        private static void ComputeFeps(List<CLRStackValue> arguments,
+        private static CLRFunctionEndPoint GetLooseCompliantFEP(
```
mjkkirschner, line 553: would appreciate summary of how this differs from `getCompliantFEP`

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -504,7 +679,7 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
 
                 foreach (List<ReplicationInstruction> replicationOption in replicationTrials)
                 {
-                    CLRFunctionEndPoint compliantTarget = GetCompliantFEP(arguments, funcGroup, replicationOption, runtimeCore, true);
+                    CLRFunctionEndPoint compliantTarget = GetCompliantFEP(thisPtr, arguments, funcGroup, replicationOption, runtimeCore, true);
```
mjkkirschner, line 682: `GetCompliantFEP` seems to call `GetConversionDistance` - which has a TODO related to array coercion when computing type distance - I wonder if this could be related to it catching all the test cases.

src/Engine/ProtoCore/DSASM/InstructionSet.cs
```diff
@@ -117,40 +117,43 @@ public struct MetaData
...
+        internal bool IsGlobal;// is global function
+
+        internal object Value { get; set; }
 
 
         internal CLRStackValue(object value, int protoType)
         {
             this.Value = value;
             this.TypeUID = protoType;
+            IsGlobal = false;
```
mjkkirschner, line 150: Is this set true anywhere?

src/Engine/ProtoCore/Lang/FunctionGroup.cs
```diff
@@ -187,30 +190,29 @@ public void CopyPublic(List<FunctionEndPoint> rhs)
...
                 // Member functions are overloaded with thisptr as the first
                 // parameter, so if member function replicates on the left hand
                 // side, the type matching should only be applied to overloaded
                 // member functions, otherwise should only be applied to original
                 // member functions.
 
-                // TODO_MSIL: figure out isInstance vs isGlobal
-                /*if (isInstance && context.IsReplicating != proc.IsAutoGeneratedThisProc)
+                // There are no scenarios yet where isReplicating is true in the MSIL VM.
+                bool isReplicating = false;
```
mjkkirschner, line 207: hmm, but shouldn't there be? should this still be a MSIL_TODO? We always replicate?

src/Engine/ProtoCore/Lang/FunctionGroup.cs
```diff
@@ -251,6 +253,32 @@ public void CopyPublic(List<FunctionEndPoint> rhs)
             return ret;
         }
 
+        internal static Dictionary<CLRFunctionEndPoint, int> GetLooseConversionDistances(
```
mjkkirschner, line 256: summary please!

src/Engine/ProtoCore/Lang/FunctionGroup.cs
```diff
@@ -307,6 +335,86 @@ public void CopyPublic(List<FunctionEndPoint> rhs)
             return ret;
         }
 
+        internal static bool CheckInvalidArrayCoersion(CLRFunctionEndPoint fep, List<CLRStackValue> reducedSVs, MSILRuntimeCore runtimeCore, bool allowArrayPromotion)
```
mjkkirschner, line 338: summary if possible.

src/Engine/ProtoCore/Lang/FunctionGroup.cs
```diff
@@ -307,6 +335,86 @@ public void CopyPublic(List<FunctionEndPoint> rhs)
...
+                if (!typ.IsIndexable)
+                    continue; //It wasn't an array param, skip
+
+                //Compute the type of target param
+                if (!allowArrayPromotion)
+                    Validity.Assert(reducedSVs[i].IsEnumerable, "This should be an array otherwise this shouldn't have passed previous tests");
+
+                if (!allowArrayPromotion)
+                {
+                    if (typ.rank != ArrayUtils.GetMaxRankForArray(reducedSVs[i]) &&
```
mjkkirschner, line 357: I think braces would make these nested if statements less prone to misreading or typos.

src/Engine/ProtoCore/Lang/FunctionGroup.cs
```diff
@@ -307,6 +335,86 @@ public void CopyPublic(List<FunctionEndPoint> rhs)
...
+                }
+                else
+                {
+                    if (typ.rank < ArrayUtils.GetMaxRankForArray(reducedSVs[i]) &&
+                        typ.rank != DSASM.Constants.kArbitraryRank)
+                        return true; //Invalid co-ercsion
+
+                }
+
+                Dictionary<ClassNode, int> arrayTypes = ArrayUtils.GetTypeStatisticsForArray(reducedSVs[i], runtimeCore);
```
mjkkirschner, line 369: var ?

src/Engine/ProtoCore/Lang/FunctionGroup.cs
```diff
@@ -307,6 +335,86 @@ public void CopyPublic(List<FunctionEndPoint> rhs)
...
+
+                //Compute the type of target param
+                if (!allowArrayPromotion)
+                    Validity.Assert(reducedSVs[i].IsEnumerable, "This should be an array otherwise this shouldn't have passed previous tests");
+
+                if (!allowArrayPromotion)
+                {
+                    if (typ.rank != ArrayUtils.GetMaxRankForArray(reducedSVs[i]) &&
+                        typ.rank != DSASM.Constants.kArbitraryRank)
+                        return true; //Invalid co-ercsion
```
mjkkirschner, line 359: spelling

src/Engine/ProtoCore/Lang/FunctionGroup.cs
```diff
@@ -307,6 +335,86 @@ public void CopyPublic(List<FunctionEndPoint> rhs)
...
+                }
+                else
+                {
+                    if (typ.rank < ArrayUtils.GetMaxRankForArray(reducedSVs[i]) &&
+                        typ.rank != DSASM.Constants.kArbitraryRank)
+                        return true; //Invalid co-ercsion
```
mjkkirschner, line 365: spelling

src/Engine/ProtoCore/Lang/FunctionGroup.cs
```diff
@@ -307,6 +335,86 @@ public void CopyPublic(List<FunctionEndPoint> rhs)
...
+                if (arrayTypes.Count == 0)
+                {
+                    //This was an empty array
+                    Validity.Assert(cn == null, "If it was an empty array, there shouldn't be a type node");
+                    cn = runtimeCore.ClassTable.ClassNodes[(int)PrimitiveType.Null];
+                }
+                else if (arrayTypes.Count == 1)
+                {
+                    //UGLY, get the key out of the array types, of which there is only one
+                    foreach (ClassNode key in arrayTypes.Keys)
```
mjkkirschner, line 382: then why use a foreach?

src/Engine/ProtoCore/Lang/FunctionGroup.cs
```diff
@@ -307,6 +335,86 @@ public void CopyPublic(List<FunctionEndPoint> rhs)
...
+                        cn = key;
+                }
+                else if (arrayTypes.Count > 1)
+                {
+                    ClassNode commonBaseType = ArrayUtils.GetGreatestCommonSubclassForArrayInternal(arrayTypes, runtimeCore.ClassTable);
```
mjkkirschner, line 387: var?

src/Engine/ProtoCore/Lang/FunctionGroup.cs
```diff
@@ -307,6 +335,86 @@ public void CopyPublic(List<FunctionEndPoint> rhs)
...
+
+                    if (commonBaseType == null)
+                        throw new ProtoCore.Exceptions.ReplicationCaseNotCurrentlySupported(
+                            string.Format(Resources.ArrayWithNotSupported, "{0C644179-14F5-4172-8EF8-A2F3739901B2}"));
```
mjkkirschner, line 391: are these guids still useful?

src/Engine/ProtoCore/Lang/FunctionGroup.cs
```diff
@@ -307,6 +335,86 @@ public void CopyPublic(List<FunctionEndPoint> rhs)
...
+
+                    cn = commonBaseType; //From now on perform tests on the commmon base type
+                }
+
+                ClassNode argTypeNode = classTable.ClassNodes[typ.UID];
```
mjkkirschner, line 396: var

src/Engine/ProtoCore/Lang/FunctionGroup.cs
```diff
@@ -307,6 +335,86 @@ public void CopyPublic(List<FunctionEndPoint> rhs)
...
+
+                //cn now represents the class node of the argument
+                //argTypeNode represents the class node of the argument
+
+                bool isNotExactTypeMatch = cn != argTypeNode;
+                bool argumentsNotNull = cn != runtimeCore.ClassTable.ClassNodes[(int)PrimitiveType.Null];
+                bool recievingTypeNotAVar = argTypeNode != runtimeCore.ClassTable.ClassNodes[(int)PrimitiveType.Var];
+                bool isNotConvertible = !cn.ConvertibleTo(typ.UID);
+
+                //bool isCalleeVar = cn == core.classTable.list[(int) PrimitiveType.kTypeVar];
```
mjkkirschner, line 406: why  is this  commented out?

src/Engine/ProtoCore/Lang/FunctionGroup.cs
```diff
@@ -251,6 +253,32 @@ public void CopyPublic(List<FunctionEndPoint> rhs)
             return ret;
         }
 
+        internal static Dictionary<CLRFunctionEndPoint, int> GetLooseConversionDistances(
+            List<CLRFunctionEndPoint> functionEndPoints,
+            List<CLRStackValue> formalParams,
+            List<ReplicationInstruction> replicationInstructions,
+            ClassTable classTable,
```
aparajit-pratap, line 260: I don't see this parameter being used in the function.

src/Engine/ProtoCore/Lang/FunctionGroup.cs
aparajit-pratap: spelling: `coercion`

src/Engine/ProtoCore/Lang/FunctionGroup.cs
```diff
@@ -307,6 +335,86 @@ public void CopyPublic(List<FunctionEndPoint> rhs)
...
+                    return true;
+
+                if (!typ.IsIndexable)
+                    continue; //It wasn't an array param, skip
+
+                //Compute the type of target param
+                if (!allowArrayPromotion)
+                    Validity.Assert(reducedSVs[i].IsEnumerable, "This should be an array otherwise this shouldn't have passed previous tests");
+
+                if (!allowArrayPromotion)
```
aparajit-pratap, line 355: I think we should be able to get rid of one `if (!allowArrayPromotion)`

src/Engine/ProtoCore/Lang/FunctionGroup.cs
aparajit-pratap: 😆

src/Engine/EmitMSIL/Replication.cs
pinzart90: I do not mean AST.
I mean the design script language. 
I do not think we will be allowed to call static methods from a class instance. Similarly we will not be allowed to call instance methods from the type name.

src/Engine/ProtoCore/Lang/FunctionGroup.cs
pinzart90: copy pasted from the old VM

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -280,7 +294,29 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
...
+            {
+                bool isClassInstanceMethod = fep.procedureNode != null &&
+                  fep.procedureNode.ClassID != Constants.kGlobalScope &&//valid class
+                  !fep.procedureNode.IsConstructor && !fep.procedureNode.IsStatic;//not static and not constructor  
+
+                if (thisPtr.IsGlobal && isClassInstanceMethod)
+                {
+                    // Filter out class instance methods when dealing with global function scope
+                    continue;
+                }
```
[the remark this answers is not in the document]
pinzart90: Yes..they do seem to be mutually exclusive. 
I do not know how the situation can arise. But that is what the old VM was filtering out. We can choose to remove this code.

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -315,14 +351,121 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
...
+            //    because static methods can only be hidden.
+            // 2. Method call from an instance of a class - valid this pointer.
+            // ex: classAInstance.method();
+            // 3. Function from the global scope - no this pointer and no class scope.
+            // ex: SomeGlobalFunction();
+            //
+            // All 3 cases will run through the same matching steps.
+
+            // A static function call has an invalid this pointer and a valid class scope;
+            bool isValidStaticFuncCall = !thisPtr.IsNull && staticFuncClassId != Constants.kInvalidIndex;
```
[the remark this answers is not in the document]
pinzart90: CLRStackValues do not have a `Pointer` field.
we can check `IsPointer == false`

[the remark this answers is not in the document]
pinzart90: bool isValidStaticFuncCall = !thisPtr.IsNull && !thisPtr.IsPointer && staticFuncClassId != Constants.kInvalidIndex

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -315,14 +351,121 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
...
+        private static CLRFunctionEndPoint SelectFEPFromMultiple(
+            CLRStackValue thisPtr,
+            MSILRuntimeCore runtimeCore,
+            List<CLRFunctionEndPoint> feps,
+            List<CLRStackValue> argumentsList)
+        {
+            Validity.Assert(thisPtr.IsNull || thisPtr.IsPointer,
+                            "this pointer wasn't null or a pointer. {89635B06-AD53-4170-ADA5-065EB2AE5858}");
+
+            var staticFuncClassId = feps.FirstOrDefault().procedureNode.ClassID;
```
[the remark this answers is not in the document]
pinzart90: classId is better, thanks

src/Engine/EmitMSIL/Replication.cs
pinzart90: I removed the flag from CLRStackValue. I left the isGLobal as hardcoded to false + TODO to figure it out later

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -96,8 +98,11 @@ internal static List<object> UnmarshalFunctionArguments(List<CLRFunctionEndPoint
...
-        private static CLRFunctionEndPoint SelectFinalFep(List<CLRFunctionEndPoint> functionEndPoint,
-                                        List<CLRStackValue> formalParameters, MSILRuntimeCore runtimeCore)
+        private static CLRFunctionEndPoint SelectFinalFep(
+            CLRStackValue thisPtr,
+            List<CLRFunctionEndPoint> functionEndPoint,
+            List<CLRStackValue> formalParameters,
+            MSILRuntimeCore runtimeCore)
         {
             List<ReplicationInstruction> replicationControl = new List<ReplicationInstruction>();
             //We're never going to replicate so create an empty structure to allow us to use
```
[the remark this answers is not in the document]
pinzart90: SelectFinalFep is no longer using a list of ReplicationInstruction.
Not sure what the comment was about (copied from the old VM).

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -564,14 +564,27 @@ private IEnumerable<ProtoCore.CLRFunctionEndPoint> FunctionLookup(IList args)
                 var procNode = funcEnd.procedureNode;
                 Validity.Assert(procNode != null, "Expected to have a valid procedureNode");
 
-                bool isStaticOrConstructorOrInternal = procNode.IsStatic || procNode.IsConstructor || CoreUtils.IsInternalMethod(methodName);
+                bool isInstanceMethod = !procNode.IsStatic &&
```
mjkkirschner, line 567: maybe `IsExternalInstanceMethod` ?

src/Engine/EmitMSIL/Replication.cs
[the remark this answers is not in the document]
mjkkirschner: perhaps add an assert or throw is this ever happens?

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -319,43 +355,180 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
...
+            {
+                typeID = runtimeCore.ClassTable.ClassNodes[typeID].Base;
+
+                foreach (var fep in feps)
+                    if (fep.procedureNode.ClassID == typeID)
+                        return fep;
+            }
+
+            //We weren't able to distinguish based on class hiearchy, try to sepearete based on array ranking
+            List<int> numberOfArbitraryRanks = new List<int>();
```
mjkkirschner, line 425: var?

mjkkirschner: do you think it could make sense performance wise to allocate these lists at the start of the while loop and reset them in each for loop?

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -200,18 +192,15 @@ public static object ReplicationLogic(List<CLRFunctionEndPoint> feps, IList args
...
-            // TODO_MSIL: Figure it the best way to interpret/pass around the this pointer.
+            // TODO_MSIL: Figure out the best way to interpret/pass around the this pointer.
             CLRStackValue thisPtr = CLRStackValue.Null;
-            bool staticFuncCall = feps.All(x => x.IsStatic);
-            if (!staticFuncCall)
+            bool staticFuncCalls = feps.All(x => x.IsStatic);
+            if (!staticFuncCalls)
             {
+                Validity.Assert(stackValues.Count > 0 && !stackValues[0].IsNull && stackValues[0].IsPointer,
+    "Expected a vcalid 'this' pointer");
```
aparajit-pratap: spelling: valid

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -319,43 +355,180 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
...
+            while (runtimeCore.ClassTable.ClassNodes[typeID].Base != Constants.kInvalidIndex)
+            {
+                typeID = runtimeCore.ClassTable.ClassNodes[typeID].Base;
+
+                foreach (var fep in feps)
+                    if (fep.procedureNode.ClassID == typeID)
+                        return fep;
+            }
+
+            //We weren't able to distinguish based on class hiearchy, try to sepearete based on array ranking
```
mjkkirschner, line 424: spelling separate

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -319,43 +355,180 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
...
+                    indicesOfSmallest.Add(i);
+                }
+                else if (numberOfArbitraryRanks[i] == smallest)
+                    indicesOfSmallest.Add(i);
+            }
+
+            Validity.Assert(indicesOfSmallest.Count > 0,
+                            "Couldn't find a fep when there should have been multiple: {EB589F55-F36B-404A-91DC-8D0EDC527E72}");
+
+            if (indicesOfSmallest.Count == 1)
```
mjkkirschner, line 458: so we are looking for the FEP with the least number of arbitrary rank params?

src/Engine/EmitMSIL/Replication.cs
mjkkirschner: ping!

src/Engine/ProtoCore/Lang/FunctionGroup.cs
mjkkirschner: @pinzart90 looks like you fixed this in the copied function, but not here:
`cn = arrayTypes.Keys.First();`

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -96,40 +96,41 @@ internal static List<object> UnmarshalFunctionArguments(List<CLRFunctionEndPoint
...
-            List<ReplicationInstruction> replicationControl = new List<ReplicationInstruction>();
-            //We're never going to replicate so create an empty structure to allow us to use
-            //the existing utility methods
-
             //Filter for exact matches
 
-            List<CLRFunctionEndPoint> exactTypeMatchingCandindates = new List<CLRFunctionEndPoint>();
-
-            foreach (CLRFunctionEndPoint possibleFep in functionEndPoint)
+            List<CLRFunctionEndPoint> exactTypeMatchingCandidates = new List<CLRFunctionEndPoint>();
```
aparajit-pratap, line 114: var

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -305,12 +343,10 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
                     arguments,
                     replicationInstructions);
 
-            // TODO_MSIL: implement GetCandidateFunctions;
-            List<CLRFunctionEndPoint> candidateFunctions = candidatesWithDistances.Keys.ToList();//CallSite.GetCandidateFunctions(candidatesWithDistances);
+            List<CLRFunctionEndPoint> candidateFunctions = GetCandidateFunctions(candidatesWithDistances);
```
aparajit-pratap, line 346: var

src/Engine/ProtoCore/Lang/CallSite.cs
```diff
@@ -1286,15 +1286,19 @@ internal static FunctionGroup FirstFunctionGroupInInheritanceChain(string method
...
                                                              Dictionary<FunctionEndPoint, int> candidatesWithDistances)
         {
-            List<FunctionEndPoint> candidateFunctions = new List<FunctionEndPoint>();
+            bool isGlobalFuncScope = stackFrame.ThisPtr.IsPointer &&
+                stackFrame.ThisPtr.Pointer == Constants.kInvalidIndex;
 
+            List<FunctionEndPoint> candidateFunctions = new List<FunctionEndPoint>();
```
aparajit-pratap, line 1292: var

aparajit-pratap (approved):

src/Engine/EmitMSIL/Replication.cs
pinzart90: yes, that is what it looks like to me too

src/Engine/EmitMSIL/Replication.cs
pinzart90: Maybe...I do not want to do such fine grain refactor (this is copy-pasted code)
I propose we refactor later when we get further ahead

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -319,43 +355,180 @@ private static List<List<ReplicationGuide>> ConstructRepGuides(string[][] replic
...
+                return feps.FirstOrDefault();
+            }
+
+            // TODO_MSIL: figure out the this ptr context
+            int typeID = /*thisPtr.TypeUID;*/Constants.kInvalidIndex;
+
+            // Try to match with feps belonging to the class scope (most derived class should have priority).
+            // In this case we simply select the function that belongs to the calling class.
+            // The assumption here is that all function end points in "feps" have already been checked that they have the same signature.
+            IEnumerable<CLRFunctionEndPoint> exactFeps = feps.Where(x => x.procedureNode.ClassID == typeID);
```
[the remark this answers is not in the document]
pinzart90: Not sure. I did not try to validate this.

src/Engine/EmitMSIL/CodeGenIL.cs
pinzart90: why not

PR #13362, opened 2022-09-30
add infrastructure to run existing tests against MSIL compiler

mjkkirschner:
### Purpose

These are the minimum changes needed to run and verify some of the existing replication tests against the new compiler/runtime.

It would be nice to come back to this later and create a runtime mirror so this runner could have a unified API - but I don't think it's worth it right now, and I'm not sure maintaining this complex runner is really worth it anyway in the longterm - for now this is just to get the tests running with no changes to them.

⚠️ 
I was wrong initially, the tests on master-5 actually did complete in just over 30 minutes, there are 3007 failures!
https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/7197/#showFailuresLink

I think thats a bit much to leave on, for now I propose leaving the bool set to use the old VM, and running these locally as we make progress.

Additionally we can leave a PR open that sets the bool to use the new compiler and we'll get test results automatically without polluting the test results of the other builds.

Screen Shot 2022-10-03 at 3 50 26 PM
Screen Shot 2022-09-30 at 2 10 48 PM

Many of these fail for two reasons:
1. We have not implemented EmitFunctionDeclarion etc
2. The current tests don't need to fully namespace functions, but we do - maybe @pinzart90's type reuse will fix this.

### Declarations

Check these if you believe they are true

- [ ] The codebase is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.
- [ ] This PR modifies some build requirements and the readme is updated

### Release Notes

(FILL ME IN) Brief description of the fix / enhancement. **Mandatory section** 

### Reviewers

(FILL ME IN) Reviewer 1  (If possible, assign the Reviewer for the PR)

(FILL ME IN, optional) Any additional notes to reviewers or testers.

### FYIs

(FILL ME IN, Optional) Names of anyone else you wish to be notified of

PR #13376, opened 2022-10-04
add infrastructure to run existing tests against MSIL compiler - attempt 2

mjkkirschner:
### Purpose

attempts to complete work started in:
https://github.com/DynamoDS/Dynamo/pull/13362

I needed to construct a live runner and set it up the same way @pinzart90 did in the micro tests - I've added the same TODOs matching the one in the core to remove this when the other ast nodes are implemented in codegenMSIL.

### Declarations

### Release Notes

### Reviewers

### FYIs

### DIFF
--- a/src/Engine/ProtoScript/Runners/ProtoScriptRunner.cs
+++ b/src/Engine/ProtoScript/Runners/ProtoScriptRunner.cs
@@ -345,9 +345,9 @@ namespace ProtoScript.Runners
             return succeeded;
         }
 
-        public Dictionary<string, object> CompileAndGenerateMSIL(string souce, EmitMSIL.CodeGenIL codegen)
+        public Dictionary<string, object> CompileAndGenerateMSIL(string source, EmitMSIL.CodeGenIL codegen)
         {
-            var ast = ParserUtils.Parse(souce).Body;
+            var ast = ParserUtils.Parse(source).Body;
             return codegen.EmitAndExecute(ast);
         }
 
--- a/test/CodeGenILTests/RangeMicroTests.cs
+++ b/test/CodeGenILTests/RangeMicroTests.cs
@@ -26,6 +26,8 @@ namespace CodeGenILTests
         [SetUp]
         public void Setup()
         {
+            //TODO_MSIL: remove the dependency on the old VM by implementing
+            //necesary Emit functions(ex mitFunctionDefinition and EmitImportStatements and all the preloading logic)
             liveRunner = new ProtoScript.Runners.LiveRunner();
 
             List<string> libraries = new List<string>();
--- a/test/Engine/ProtoTestFx/TestFrameWork.cs
+++ b/test/Engine/ProtoTestFx/TestFrameWork.cs
@@ -405,12 +405,26 @@ namespace ProtoTestFx.TD
             if (testMSILExecution)
             {
                 Console.WriteLine($"!!!!test {NUnit.Framework.TestContext.CurrentContext.Test.Name} being run with MSIL compiler!!!!");
-                var assemblyPath = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location));
-                var outputpath = Path.Combine(assemblyPath, "MSILTestOutput");
-                System.IO.Directory.CreateDirectory(outputpath);
-                var inputs = new Dictionary<string, IList>();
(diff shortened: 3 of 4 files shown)


src/Engine/ProtoScript/Runners/ProtoScriptRunner.cs
```diff
@@ -345,6 +345,12 @@ public bool CompileAndGenerateExe(string sourcecode, ProtoCore.Core compileCore,
             return succeeded;
         }
 
+        public Dictionary<string, object> CompileAndGenerateMSIL(string souce, EmitMSIL.CodeGenIL codegen)
```
aparajit-pratap, line 348: source

test/CodeGenILTests/RangeMicroTests.cs
```diff
@@ -26,6 +26,8 @@ protected virtual void GetLibrariesToPreload(ref List<string> libraries)
         [SetUp]
         public void Setup()
         {
+            //TODO_MSIL: remove the dependency on the old VM by implementing
+            //necesary Emit functions(ex mitFunctionDefinition and EmitImportStatements and all the preloading logic)
```
aparajit-pratap, line 30: Something to keep in mind is that assemblies are not just imported into the VM using import statements, there are other code paths in Dynamo currently that also preload them into the VM.

test/Engine/ProtoTestFx/TestFrameWork.cs
```diff
@@ -397,6 +402,32 @@ public virtual ExecutionMirror RunScriptSource(string sourceCode, string errorst
...
+            if (testMSILExecution)
+            {
+                Console.WriteLine($"!!!!test {NUnit.Framework.TestContext.CurrentContext.Test.Name} being run with MSIL compiler!!!!");
+                //TODO_MSIL: remove the dependency on the old VM by implementing
+                //necesary Emit functions(ex mitFunctionDefinition and EmitImportStatements and all the preloading logic)
+                //MSIL_TODO add list of libraries to load here for tests.
+                //It seems we have to load SOME library successfully or DSExecutable is null.
+                using (var liveRunner = new ProtoScript.Runners.LiveRunner())
+                {
+                    liveRunner.ResetVMAndResyncGraph(new string[] { "DesignScriptBuiltin.dll" });
```
aparajit-pratap, line 414: Why do you need to reset the VM here?

mjkkirschner: I was following the pattern @pinzart90 used here:
https://github.com/DynamoDS/Dynamo/blob/emitMSIL/test/CodeGenILTests/RangeMicroTests.cs#L34

If I did not call this method with some valid library then `DSExecutable` seemed to always be null.

test/CodeGenILTests/RangeMicroTests.cs
mjkkirschner: thanks, it's a good point - @pinzart90 did you have time to file a task on addressing these TODOs? It would be good to add @aparajit-pratap's point about preloading to the task AC.

test/Engine/ProtoTestFx/TestFrameWork.cs
aparajit-pratap: but can't you use the same instance of the old VM that is already initialized? I mean isn't `liveRunner.runtimeCore` and the rest of the stuff like `DSExecutable` etc. already initialized in these tests?

mjkkirschner: Not really, the tests start a testcore and then create a runtimecore sometime later, it's all bound up in the existing execute method and I didn't see the value in refactoring this - it's not obvious to me we're going to keep all this legacy infrastructure...

mjkkirschner: I can take a look at reusing it if you think it's worth it? I was hoping we could run these tests with only import statements at some point.

aparajit-pratap: Ok, thanks.

aparajit-pratap: > I can take a look at reusing it if you think it's worth it? I was hoping we could run these tests with only import statements at some point.

No, I don't think it's worth it if we are going to remove the dependency on the old runtime core at some point later.

aparajit-pratap (approved):

pinzart90 (approved):

test/CodeGenILTests/RangeMicroTests.cs
pinzart90: I'll ping you @mjkkirschner  when the task is created

PR #13401, opened 2022-10-17
Expose dynamic wrappers to compiler, defer marshaling to/from where possible.

mjkkirschner:
### Purpose
based on:
https://github.com/DynamoDS/Dynamo/pull/13379

The intent of this PR is to attempt to reduce the time required for marshaling and un-marshaling when making replicated calls via `ReplicationLogic` I'll refer to this as dynamic replication in the rest of this PR. (as opposed to replication unrolling at compile time which we don't currently support).

⚠️ 
### Notes:
1. ~This PR currently causes 5 test regressions related to coercion - dreaded invalid program exception - might need assistance looking into these.~ ✅ 
2. Performance gains seem moderate, I guess they are mostly from deferring unmarshaling until after execution where possible. This is kind of like cheating?
3.  ~~I think we can avoid calling `unmarshalFunctionArgs` before all direct function calls by checking if arg type is `CLRStackValue` - currently I just do it for all the args.~~ done. ✅ 
4. Yet to be seen if we should proceed further in this direction, IE is the complexity worth the perf gains.

### What does this PR do:

1. Short circuits marshaling arguments to replicated calls, if the value is already a `CLRStackValue`. This is now probable because of point 2.
2. We no longer un-marshal `CLRStackValues` back to plain c# objects after every replicated call. - Instead we defer this un-marshling until later.
3. Short circuit un-marshaling - if the value is not a CLRStackValue, we can just return.
4. **unmarshal case 1** - Introduces a new Dictionary wrapper type that un-marshals when a client tries to get a value at a key. This means that unmarshaling happens at test time or UI time, when the client asks for a specific value, we lazily un-marshal instead of doing it all up front. I think this will be especially valuable in a headless scenario where you only care about a few values, or for example in a UI scenario if we only show preview bubbles when a user interacts with them.
5. **unmarshal case 2** -We are attempting to make a non replicated direct function call - but one or more of our inputs was the result of a dynamic replicated call - now the argument is a `CLRStackValue` and we are forced to unmarshal before calling our function. This means we lose any benefit of points 1 and 2 - and incur a serious addition of complexity to our direct function emit code. Currently in this PR we **always** unmarshal when calling a direct function - but we could avoid this by only unmarshaling if an argument is a CLRStackValue or from a replicated call - I think we can figure this out at compile by checking the return type of that AST.

 
### TODO:
- [x] profile perf
- [ ]  profile memory of slightly larger wrapper type (stores FEP return type)
- [x]   fix tests. 😢 

### Declarations

### Release Notes

### Reviewers

### FYIs

### DIFF
--- a/src/Engine/EmitMSIL/BuiltIn.cs
+++ b/src/Engine/EmitMSIL/BuiltIn.cs
@@ -36,7 +36,7 @@ namespace EmitMSIL
             {
                 get
                 {
-                    var obj = Unmarshal(key);
+                    var obj = UnmarshalKey(key);
                     backingDict[key] = obj;
                     return obj;
                 }
@@ -53,7 +53,7 @@ namespace EmitMSIL
                 return marshaler.UnMarshal(clrwrapped, clrwrapped.CLRFEPReturnType, runtimeCore);
             }
 
-            private V Unmarshal(K key)
+            private V UnmarshalKey(K key)
             {
                 //TODO consider scanning for nested wrappers we missed.
                 if (backingDict[key] is CLRStackValue clrwrapped)
@@ -66,12 +66,21 @@ namespace EmitMSIL
 
             public ICollection<K> Keys => backingDict.Keys;
 
-            public ICollection<V> Values => backingDict.Keys.Select((x) =>
+            public ICollection<V> Values
             {
-                var obj = Unmarshal(x);
-                backingDict[x] = obj;
-                return obj;
-            }).ToArray();
+                get
+                {
+                    var output = new List<V>(backingDict.Count);
+                    //copy keys so we can mutate dictionary.
+                    foreach (var key in backingDict.Keys.ToList())
+                    {
+                        var obj = UnmarshalKey(key);
+                        output.Add(obj);
+                        backingDict[key] = obj;
+                    }
+                    return output;
+                }
+            }
 
 
             public int Count => backingDict.Count;
@@ -127,7 +136,7 @@ namespace EmitMSIL
             {
                 if (backingDict.TryGetValue(key, out _))
                 {
-                    value = Unmarshal(key);
+                    value = UnmarshalKey(key);
                     backingDict[key] = value;
                     return true;
                 }
--- a/src/Engine/EmitMSIL/BuiltIn.cs
+++ b/src/Engine/EmitMSIL/BuiltIn.cs
@@ -39,6 +39,7 @@ namespace EmitMSIL
 
             private V Unmarshal(K key)
             {
+                //TODO consider scanning for nested wrappers we missed.
                 if (backingDict[key] is CLRStackValue clrwrapped)
                 {
                     return marshaler.UnMarshal(clrwrapped, clrwrapped.CLRFEPReturnType, runtimeCore) as V;
@@ -103,12 +104,14 @@ namespace EmitMSIL
 
             public bool TryGetValue(K key, out V value)
             {
-                //TODO call
-                V intermediatevalue;
-                var result = backingDict.TryGetValue(key, out intermediatevalue);
-                // unmarshal intermediate val.
-                value = intermediatevalue;
-                return result;
+                if (backingDict.TryGetValue(key, out _))
+                {
+                    value = Unmarshal(key);
+                    return true;
+                }
+
+                value = null;
+                return false;
             }
 
             IEnumerator IEnumerable.GetEnumerator()
--- a/src/Engine/EmitMSIL/BuiltIn.cs
+++ b/src/Engine/EmitMSIL/BuiltIn.cs
@@ -1,12 +1,121 @@
-﻿using ProtoCore.DSASM;
+using ProtoCore.DSASM;
 using ProtoCore.Utils;
 using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
 using System.Reflection;
+using ProtoCore;
+using ProtoFFI;
+using Type = ProtoCore.Type;
 
 namespace EmitMSIL
 {
     public class BuiltIn
     {
+        /// <summary>
+        /// A Dictionary wrapper for MSIL engine outputs, any dynamic CLR stackvalues are unmarshaled before they are returned.
+        /// </summary>
+        /// <typeparam name="K"></typeparam>
+        /// <typeparam name="V"></typeparam>
+        internal class MSILOutputMap<K, V> : IDictionary<K, V> where V : class
+        {
+            private Dictionary<K, V> backingDict = new Dictionary<K, V>();
+            private MSILRuntimeCore runtimeCore;
+            private FFIObjectMarshaler marshaler;
+
+            public MSILOutputMap(MSILRuntimeCore runtimeCore)
+            {
+                this.runtimeCore = runtimeCore;
+                marshaler = ProtoFFI.CLRDLLModule.GetMarshaler(runtimeCore);
+
+            }
+
+            public V this[K key]
+            {
+                get => Unmarshal(key);
+                set => backingDict[key] = value;
+            }
+
+            private V Unmarshal(K key)
+            {
+                if (backingDict[key] is CLRStackValue clrwrapped)
+                {
+                    return marshaler.UnMarshal(clrwrapped, clrwrapped.CLRFEPReturnType, runtimeCore) as V;
+                }
+
+                return backingDict[key];
+            }
+
+            public ICollection<K> Keys => backingDict.Keys;
+
+            public ICollection<V> Values => backingDict.Keys.Select(x => Unmarshal(x)).ToArray();
+
+
+            public int Count => backingDict.Count;
+
+            public bool IsReadOnly => (backingDict as ICollection<KeyValuePair<K, V>>).IsReadOnly;
+
+            public void Add(K key, V value)
+            {
+                backingDict.Add(key, value);
+            }
+
+            public void Add(KeyValuePair<K, V> item)
+            {
+                (backingDict as ICollection<KeyValuePair<K, V>>).Add(item);
+            }
+
+            public void Clear()
+            {
+                backingDict.Clear();
+            }
+
+            public bool Contains(KeyValuePair<K, V> item)
+            {
+                return (backingDict as ICollection<KeyValuePair<K, V>>).Contains(item);
+            }
+
+            public bool ContainsKey(K key)
+            {
+                return backingDict.ContainsKey(key);
+            }
+
+            public void CopyTo(KeyValuePair<K, V>[] array, int arrayIndex)
+            {
+                (backingDict as ICollection<KeyValuePair<K, V>>).CopyTo(array, arrayIndex);
+            }
+
+            public IEnumerator<KeyValuePair<K, V>> GetEnumerator()
+            {
+                return backingDict.GetEnumerator();
+            }
+
+            public bool Remove(K key)
+            {
+                return backingDict.Remove(key);
+            }
+
+            public bool Remove(KeyValuePair<K, V> item)
+            {
+                return (backingDict as ICollection<KeyValuePair<K, V>>).Remove(item);
+            }
+
+            public bool TryGetValue(K key, out V value)
+            {
+                //TODO call
+                V intermediatevalue;
+                var result = backingDict.TryGetValue(key, out intermediatevalue);
+                // unmarshal intermediate val.
+                value = intermediatevalue;
+                return result;
+            }
+
+            IEnumerator IEnumerable.GetEnumerator()
+            {
+                return backingDict.GetEnumerator();
+            }
+        }
         private static object ToBoolean(object a)
         {
             switch (a)
--- a/src/Engine/EmitMSIL/BuiltIn.cs
+++ b/src/Engine/EmitMSIL/BuiltIn.cs
@@ -34,7 +34,12 @@ namespace EmitMSIL
 
             public V this[K key]
             {
-                get => Unmarshal(key);
+                get
+                {
+                    var obj = Unmarshal(key);
+                    backingDict[key] = obj;
+                    return obj;
+                }
                 set => backingDict[key] = value;
             }
             /// <summary>
@@ -61,7 +66,12 @@ namespace EmitMSIL
 
             public ICollection<K> Keys => backingDict.Keys;
 
-            public ICollection<V> Values => backingDict.Keys.Select(x => Unmarshal(x)).ToArray();
+            public ICollection<V> Values => backingDict.Keys.Select((x) =>
+            {
+                var obj = Unmarshal(x);
+                backingDict[x] = obj;
+                return obj;
+            }).ToArray();
 
 
             public int Count => backingDict.Count;
@@ -118,6 +128,7 @@ namespace EmitMSIL
                 if (backingDict.TryGetValue(key, out _))
                 {
                     value = Unmarshal(key);
+                    backingDict[key] = value;
                     return true;
                 }
 
--- a/src/Engine/EmitMSIL/BuiltIn.cs
+++ b/src/Engine/EmitMSIL/BuiltIn.cs
@@ -18,7 +18,8 @@ namespace EmitMSIL
         /// </summary>
         /// <typeparam name="K"></typeparam>
         /// <typeparam name="V"></typeparam>
-        internal class MSILOutputMap<K, V> : IDictionary<K, V> where V : class
+        [Obsolete("Internal, do not use")]
+        public class MSILOutputMap<K, V> : IDictionary<K, V> where V : class
         {
             private Dictionary<K, V> backingDict = new Dictionary<K, V>();
             private MSILRuntimeCore runtimeCore;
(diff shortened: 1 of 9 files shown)


src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -110,7 +110,7 @@ internal CodeGenIL(IDictionary<string, IList> input, string filePath, ProtoCore.
             // 1. Create assembly builder (dynamic assembly)
             var asm = BuilderHelper.CreateAssemblyBuilder("DynamicAssembly", false, access);
             // 2. Create module builder
-            var mod = BuilderHelper.CreateDLLModuleBuilder(asm, "DynamicModule");
+            var mod = BuilderHelper.CreateDLLModuleBuilder(asm, "DynamicAssembly");
```
aparajit-pratap, line 113: ?

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -38,7 +46,12 @@ internal static List<CLRStackValue> MarshalFunctionArguments(IList args, MSILRun
             return stackValues;
         }
 
-        internal static List<object> UnmarshalFunctionArguments(List<CLRFunctionEndPoint.ParamInfo> formalParams, List<CLRStackValue> args, MSILRuntimeCore runtimeCore)
+        public static IList<object> UnMarshalFunctionArguments2(IEnumerable<CLRFunctionEndPoint> feps, IEnumerable args, MSILRuntimeCore runtimeCore)
```
aparajit-pratap, line 49: why make this public?

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -16,7 +16,7 @@ namespace EmitMSIL
     [Obsolete("This is an internal class, do not use it.")]
     public class Replication
     {
-        internal static List<CLRStackValue> MarshalFunctionArguments(IList args, MSILRuntimeCore runtimeCore)
+        public static List<CLRStackValue> MarshalFunctionArguments(IList args, MSILRuntimeCore runtimeCore)
```
aparajit-pratap, line 19: why make this public?

src/Engine/EmitMSIL/Replication.cs
aparajit-pratap: Ok, I see you're emitting a call to this at compile time.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1405,26 +1407,111 @@ private Type EmitFunctionCallNode(AssociativeNode node)
...
+                        t = EmitCoercionCode(n, t, parameters[index - 1]);
+                    }
+
+                    if (t == null) return;
+
+                    if (t.IsValueType)
+                    {
+                        EmitOpCode(OpCodes.Box, t);
+                    }
+                });
```
aparajit-pratap, line 1472: Why does all this need to be done instead of simply `Type t = DfsTraverse(n);` and `EmitCoercion` in a for loop? Just trying to figure out what was wrong in the previous logic.

aparajit-pratap: Oh, I see this is to emit the args for `UnMarshal2`. It looks like all this is negating the benefits of a direction function call 🤔

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1405,26 +1407,111 @@ private Type EmitFunctionCallNode(AssociativeNode node)
...
+
+                //TODO if possible only marshal args that resulted from a replicated call.
+                
+                //emit marshal for args
+
+                // Emit call to load the runtimeCore argument
+                EmitOpCode(OpCodes.Ldarg_3);
+                //TODO cache this at start.
+                //now call marshall to convert any clrstackvalues to plain c# objs. 
+                var marhshallMethod = typeof(Replication).GetMethod(nameof(Replication.UnMarshalFunctionArguments2), BindingFlags.Static | BindingFlags.Public);
```
aparajit-pratap, line 1482: spelling `marshalMethod`

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1405,26 +1407,111 @@ private Type EmitFunctionCallNode(AssociativeNode node)
...
+                EmitOpCode(OpCodes.Call, marhshallMethod);
+
+                var unMarshaledArgsArray = DeclareLocal(marhshallMethod.ReturnType, "unMarshaledArgsArray");
+                if (unMarshaledArgsArray != null)
+                {
+                    EmitOpCode(OpCodes.Stloc, unMarshaledArgsArray.LocalIndex);
+                }
+                //now on the stack we have a list of objects, we need to iterate through it and push each item to the stack.
+                //emit a for loop which pushes to stack.
```
aparajit-pratap, line 1491: I believe the stloc in the previous line will pop the array from the stack.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1405,26 +1407,111 @@ private Type EmitFunctionCallNode(AssociativeNode node)
...
+
+                    if (t.IsValueType)
+                    {
+                        EmitOpCode(OpCodes.Box, t);
+                    }
+                });
+
+                //TODO if possible only marshal args that resulted from a replicated call.
+                
+                //emit marshal for args
```
aparajit-pratap, line 1476: you mean unmarshal right?

test/CodeGenILTests/IndexingMicroTests.cs
```diff
@@ -186,7 +186,7 @@ public void IndexIntoRange_WithIdent()
 b = 5;
 c=a[b];";
             var ast = ParserUtils.Parse(dscode).Body;
-            var output = codeGen.EmitAndExecute(ast);
+            var output = codeGen.Emit(ast);
```
aparajit-pratap, line 189: Did you change this to save the assembly to disk? I believe this should continue to be `EmitAndExecute`.

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: I'm not sure exactly whats going on locally - but if this is not named the same as the Dynamic Assembly file it ends up generating a separate module file that I cannot decompile, and the .dll fails to decompile.

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: well - thats not exactly clear yet - we **need** to do the un-marshaling somewhere between a replicated call and a non replicated one.

I have yet to get some hard numbers.

It definitely ends up making the code more complex - please see point 5 in the PR description for more info.

test/CodeGenILTests/IndexingMicroTests.cs
mjkkirschner: yep, just did this while testing, will revert.

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: yes, thanks.

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: thats @aparajit-pratap that makes sense - it should be okay though since we do 
`EmitOpCode(OpCodes.Ldloc, unMarshaledArgsArray.LocalIndex);` at each step in the unrolling below.

Though I should definitely clarify the comment.

src/Engine/EmitMSIL/CodeGenIL.cs
aparajit-pratap: I was able to decompile the separately generated DynamicModule.dll but found DynamicAssembly.dll to be more or less empty. In any case, I was unclear what was the real difference between them and it might be that both need to be named the same as you're doing here.

src/Engine/EmitMSIL/Replication.cs
```diff
@@ -38,7 +46,12 @@ internal static List<CLRStackValue> MarshalFunctionArguments(IList args, MSILRun
             return stackValues;
         }
 
-        internal static List<object> UnmarshalFunctionArguments(List<CLRFunctionEndPoint.ParamInfo> formalParams, List<CLRStackValue> args, MSILRuntimeCore runtimeCore)
+        public static IList<object> UnMarshalFunctionArguments2(IEnumerable<CLRFunctionEndPoint> feps, IEnumerable args, MSILRuntimeCore runtimeCore)
+        {
+            return UnmarshalFunctionArguments(feps.FirstOrDefault().FormalParams, (IList)args, runtimeCore).ToArray();
```
pinzart90, line 51: are we confident that the first FEP is good enough for these purposes ?
Should we add a TODO to figure out how to handle multiple FEPs ?

mjkkirschner: I am not confident - I will add a TODO and look at this as a potential cause of test failures perhaps.

src/Engine/EmitMSIL/BuiltIn.cs
```diff
@@ -1,12 +1,124 @@
...
+            public MSILOutputMap(MSILRuntimeCore runtimeCore)
+            {
+                this.runtimeCore = runtimeCore;
+                marshaler = ProtoFFI.CLRDLLModule.GetMarshaler(runtimeCore);
+
+            }
+
+            public V this[K key]
+            {
+                get => Unmarshal(key);
```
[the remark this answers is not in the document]
mjkkirschner: good idea.

src/Engine/EmitMSIL/Replication.cs
aparajit-pratap: As part of my task, I'm working on this to determine a single fep at compile time. If I'm not able to narrow it down to a single fep, I'm passing the list of feps for it to be determined at runtime.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1405,26 +1407,111 @@ private Type EmitFunctionCallNode(AssociativeNode node)
...
+                    for (int i = 0; i < args.Count; i++)
+                    {
+                        //load array
+                        EmitOpCode(OpCodes.Ldloc, unMarshaledArgsArray.LocalIndex);
+                        EmitOpCode(OpCodes.Ldc_I4, i);
+                        EmitOpCode(OpCodes.Ldelem_Ref);
+                        if (parameters[i].IsValueType)
+                        {
+                            EmitOpCode(OpCodes.Unbox_Any, parameters[i]);
+                        }
```
aparajit-pratap, line 1510: @mjkkirschner, if I understand this correctly, shouldn't this be something like:
```
EmitOpCode(OpCodes.Ldloc, unMarshaledArgsArray.LocalIndex);
 EmitOpCode(OpCodes.Ldc_I4, i);
if (parameters[i].IsValueType)
{
            EmitOpCode(OpCodes.Unbox_Any, parameters[i]);
            EmitOpCode(OpCodes.Ldelem, typeof(parameters[i]));
}
else EmitOpCode(OpCodes.Ldelem_Ref);

Comment (mjkkirschner) [reply src/Engine/EmitMSIL/CodeGenIL.cs:1510]: in your sample code, the object reference is not yet on the stack, wouldn't that unbox be applied to the constant `i` - not the value in the array?

Comment (aparajit-pratap) [reply src/Engine/EmitMSIL/CodeGenIL.cs:1510]: I'm not sure how unbox works but how about this:
```
EmitOpCode(OpCodes.Ldloc, unMarshaledArgsArray.LocalIndex);
 EmitOpCode(OpCodes.Ldc_I4, i);
EmitOpCode(OpCodes.Ldelem_Ref);
if (parameters[i].IsValueType)
{
            EmitOpCode(OpCodes.Unbox_Any, parameters[i]);
            EmitOpCode(OpCodes.Ldelem, typeof(parameters[i]));
}
```

Comment (aparajit-pratap) [reply src/Engine/EmitMSIL/CodeGenIL.cs:1510]: Also, what's the difference between unbox_any and unbox?

Comment (aparajit-pratap) [reply src/Engine/EmitMSIL/CodeGenIL.cs:1510]: Ok, I was wondering if a `ldelem` was required after the unboxing, so this clears it.

File: src/Engine/EmitMSIL/CodeGenIL.cs [hunk 12]
```diff
@@ -361,34 +363,19 @@ private Type EmitCoercionCode(AssociativeNode arg, Type argType, Type param)
...
             EmitOpCode(OpCodes.Ldloc, newArrIndex);
 
             localBuilder = DeclareLocal(t, "target array");
             var targetArrayIndex = localBuilder.LocalIndex;
             EmitOpCode(OpCodes.Stloc, targetArrayIndex);
-
-            var endLabel = DefineLabel();
-            EmitOpCode(OpCodes.Br_S, endLabel.Value);
-
-            MarkLabel(endLabel.Value, "label: end label");
```
aparajit-pratap, line 391: So all of this code removal is simply equivalent to removing the try and finally blocks?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1051,6 +1039,20 @@ private Type EmitExprListNode(AssociativeNode node)
                 Type t = DfsTraverse(el);
                 if (t == null) return;
 
+                //if this element is a CLRStackValue, we need to unmarshal it.
+                if (t == typeof(DSASM.CLRStackValue))
+                {
+                    //reorganize stack.
+                    EmitOpCode(OpCodes.Pop);
+                    EmitOpCode(OpCodes.Ldarg_2);
+                    DfsTraverse(el);
```
aparajit-pratap, line 1048: Can you explain what you're doing here? I'm confused.

aparajit-pratap:
> I think we can avoid calling unmarshalFunctionArgs before all direct function calls by checking if arg type is CLRStackValue - currently I just do it for all the args.

In such cases, you're still calling `unmarshalFunctionArgs` but at least you're short-circuiting the call by directly returning the object instead of going ahead with the unmarshaling in the way you've implemented `unmarshalFunctionArgs`, correct?

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: well, it removes the leave instruction, which if I am understanding correctly, jumped to the finally label before jumping to the leave label - it also removes the end label, we just fall into that code now instead of needing to branch unconditionally.

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: I added an `unmarshal` instance util function to the `output` dictionary since I was already cacheing a marshaler there -
this code pops the current arg off the stack, so we can put the `output` dictionary on top first, then adds back the arg - thats done because we want to call an instance method on the `output` dictionary.

I did not try it but perhaps I can use the `ASTTypeMap` or variables dictionary instead to avoid the pop and extra stack load.

mjkkirschner:
> > I think we can avoid calling unmarshalFunctionArgs before all direct function calls by checking if arg type is CLRStackValue - currently I just do it for all the args.
> 
> In such cases, you're still calling `unmarshalFunctionArgs` but at least you're short-circuiting the call by directly returning the object instead of going ahead with the unmarshaling in the way you've implemented `unmarshalFunctionArgs`, correct?

yes thats correct.

src/Engine/EmitMSIL/CodeGenIL.cs
aparajit-pratap: Ok, makes sense. Yes, I was wondering if you could avoid the first `DfsTraverse(el)` in line 1039 but I suppose this is okay for now.

aparajit-pratap: So, if I understand this correctly, you need to unmarshal each element in an array in addition to the top-level unmarshal you're doing at the function call level for each argument, correct?

mjkkirschner: so - maybe I am misunderstanding your question - but the reason I am adding any code here at all is a failing test - consider this case:

x = [1,2,3]
y = vaueAtIndex(x,1) // this replicates for some reason. - so `y` is a `CLRStackValue`
z  = [1,y];

z now contains `[int,CLRStackValue]`

We can't really defer this safely because the unmarshal methods always assume that `CLRStackValues` are always nested only under other `CLRStackValues` - so we unmarshal any wrappers we find when creating a list to avoid a situation where a plain c# object list might contain a CLRStackValue element.

aparajit-pratap: Cool, thanks!

aparajit-pratap (approved):

mjkkirschner:
in [this run](https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/7371/)  the only test that failed is this unrelated flaky UI test.

DynamoCoreWpfTests.CoreUserInterfaceTests.WorkspaceContextMenu_TestIfOpenOnRightClick

so PTAL! @pinzart90 @sm6srw

pinzart90 (approved):

PR #13403, opened 2022-10-17
Agd 2777 player dropdown inputs

twastvedt:
### Purpose
Fixes [AGD-2777](https://jira.autodesk.com/browse/AGD-2777)

When running a graph from Dynamo Player while the Dynamo UI was open, the value and UI display of dropdown nodes was not updated consistently.

### Declarations

### Reviewers

@saintentropy

### DIFF
--- a/src/Libraries/CoreNodeModels/DropDown.cs
+++ b/src/Libraries/CoreNodeModels/DropDown.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
@@ -105,7 +105,9 @@ namespace CoreNodeModels
                     selectedIndex = value;
                     selectedString = GetSelectedStringFromItem(Items.ElementAt(value));
                 }
+
                 RaisePropertyChanged("SelectedIndex");
+                RaisePropertyChanged("SelectedString");
             }
         }
 
@@ -120,15 +122,22 @@ namespace CoreNodeModels
             get => selectedString;
             set
             {
-                if (!string.IsNullOrEmpty(value) && value != selectedString)
+                if (value == selectedString)
+                {
+                    return;
+                }
+
+                if (!string.IsNullOrEmpty(value))
                 {
                     var item = Items.FirstOrDefault(i => GetSelectedStringFromItem(i).Equals(value));
                     // In the case that SelectedString deserialize after SelectedIndex
                     // With a valid item from search, get the index of item and replace the current one. 
                     // If no exact match found, fall back to use the default selectedIndex from deserialization.
-                    selectedIndex = item != null ?
-                        Items.IndexOf(item) :
-                        selectedIndex;
+                    if (item != null)
+                    {
+                        selectedIndex = Items.IndexOf(item);
+                        RaisePropertyChanged("SelectedIndex");
+                    }
                 }
 
                 selectedString = value;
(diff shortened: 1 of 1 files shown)


src/Libraries/CoreNodeModels/DropDown.cs
```diff
@@ -91,45 +90,54 @@ public override NodeInputData InputData
...
                     // In the case that SelectedString deserialize after SelectedIndex
                     // With a valid item from search, get the index of item and replace the current one. 
                     // If no exact match found, fall back to use the default selectedIndex from deserialization.
-                    selectedIndex = item != null ?
-                        Items.IndexOf(item) :
-                        selectedIndex;
+                    if (item != null)
+                    {
+                        selectedIndex = Items.IndexOf(item);
+                        RaisePropertyChanged("SelectedIndex");
```
QilongTang, line 139: no `RaisePropertyChanged("SelectedString");` here?

src/Libraries/CoreNodeModels/DropDown.cs
```diff
@@ -173,32 +181,23 @@ protected override void DeserializeCore(XmlElement nodeElement, SaveContext cont
...
             string value = updateValueParams.PropertyValue;
 
-            if (name == "Value" && value != null)
+            if (updateValueParams.PropertyName == "Value" && value != null)
             {
-                selectedIndex = ParseSelectedIndex(value, Items);
-                if (selectedIndex < 0)
-                {
-                    Warning(Dynamo.Properties.Resources.NothingIsSelectedWarning);
-                    selectedString = String.Empty;
```
QilongTang, line 195: Did I miss anything, the edge case here is skipped? I believe if the parse failed, the index will be negative

QilongTang (commented):
A few comments

src/Libraries/CoreNodeModels/DropDown.cs
twastvedt: We're now using the property setter to set SelectedIndex, so that we get the PropertyChanged event. That setter already has an out-of-bounds check. I notice it doesn't have the Warning though, I'll add that.

twastvedt: Unless there's a pathway which should be able to set SelectedIndex to -1 without raising a warning. I don't _think_ so...

src/Libraries/CoreNodeModels/DropDown.cs
twastvedt: it happens down below, line 144 (hidden in the diff)

src/Libraries/CoreNodeModels/DropDown.cs
```diff
@@ -91,45 +90,55 @@ public override NodeInputData InputData
...
         {
-            get { return selectedIndex; }
+            get => selectedIndex;
             set
             {
                 //do not allow selected index to
                 //go out of range of the items collection
                 if (value > Items.Count - 1 || value < 0)
                 {
+                    Warning(Dynamo.Properties.Resources.NothingIsSelectedWarning);
```
QilongTang, line 100: hmm please give it some testing, I think the default index is -1, so adding the warning here may affect default behavior of this Dynamo node, e.g. when user place it, it display a warning already?

QilongTang: If your testing is good, then it is probably safe to do so, I am not 100% sure :)

twastvedt: OK, thanks for noticing this. Adding it to the setter isn't good, as `SelectedIndex` is set to -1 when the dropdown is opened. I added the warning back into `UpdateValueCore`.

src/Libraries/CoreNodeModels/DropDown.cs
twastvedt: Added the warning back here. The rest of the edge case check is handled in the setter.

QilongTang: Thanks, should be good then

QilongTang:
Waiting for all PR checks to finish

twastvedt:
@QilongTang Yes we will!

PR #13410, opened 2022-10-19
Dyn 5288 node autocomplete improvements

jesusalvino:
### Purpose
Implement the story https://jira.autodesk.com/browse/DYN-5288

### Declarations

### Reviewers
@QilongTang 

### FYIs
@RobertGlobant20 @filipeotero


Amoursol:
@jesusalvino looking awesome as a first pass! A couple of comments:

- Overall, is the font used Artifakt for consistency with the rest of the UI? Looks different to my naked eye, but I could be wrong.
- For "Case 2", is there any reason we keep the graphic? The message states there are now nodes, but we are showing two options. This will be confusing to end-users.
- We should also have at the very least a "Watch node" as an output option to all nodes.

jesusalvino:
@Amoursol I have already discussed with @Jingyi-Wen about the Case 2:

Assuming that each node from the ML result could be "by use" (recently used) or "by recommendation", this case shows 2 results and they are only by use, that's why the "no recommendations message" is displayed, it probably rarely happens, please let me know if we are good with the implementation of the mockups, any change / update is welcome.

Amoursol:
> @Amoursol I have already discussed with @Jingyi-Wen about the Case 2:
> 
> Assuming that each node from the ML result could be "by use" (recently used) or "by recommendation", this case shows 2 results and they are only by use, that's why the "no recommendations message" is displayed, it probably rarely happens, please let me know if we are good with the implementation of the mockups, any change / update is welcome.

Ah, that makes sense. In this case, please do keep it as is!

src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
```diff
@@ -238,6 +238,27 @@ public ImageSource LargeIcon
...
+        /// </summary>
+        public Boolean IsByUse { get; set; }
+        /// <summary>
+        /// Indicates if the Node is part of the ML result per recommendation
+        /// </summary>
+        public Boolean IsByRecommendation
+        {
+            get
+            { return !IsByUse; }
+        }
```
QilongTang, line 260: This class is shared for incanvas search and node autocomplete search. How does your current changes affect the look for incanvas search results?

Also, to make the usage more appearant, maybe we can create a new class inherit from the current one?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -462,5 +544,30 @@ private static int GetTypeDistance(string typea, string typeb, ProtoCore.Core co
...
+
+        /// <summary>
+        /// Indicates if display the No recommendations / Low confidence message (image and texts)
+        /// </summary>
+        public bool DisplayNoRecommendationsLowConfidence { get; set; }
+
+        /// <summary>
+        /// Indicates if display the Low confidence option and Tooltip
+        /// </summary>
+        public bool DisplayLowConfidence { get; set; }
```
QilongTang, line 571: Do we need all of these properties? It seems most of them will have same/opposite value of `Models.NodeAutocompleteSuggestion.MLRecommendation`?

QilongTang (commented):
Looks good but I have two questions so wondering if we can further simplify/ specify things

src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
jesusalvino: I have tested the incanvas search and the new properties of the NodeSearchElementViewModel don't impact its look since in its UI implementation (SidebarGridStyleDictionary.xaml MemberTemplate) is being handling by the property ViewConfidenceScoreRecentUse, by default false. so it will keep hide the new mapped controls at least that we want to display them, that's the case when the user select the ML method.

About create a new class based on the current one  it means replace it in the NodeAutoCompleteSearchViewModel.cs and also in the SearchViewModel.cs so that mean several changes, per my understanding keeping the 4 new properties in the current class is valid since the search element will need to support its new source from the ML.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
jesusalvino: Yes, according to the Mockups we have different messages and behavior based on the type of the results: by use, recommendations and confidence score.

src/DynamoCoreWpf/ViewModels/Search/NodeSearchElementViewModel.cs
QilongTang: @jesusalvino My take is the 4 new properties do not apply to the base class but only search element for NodeAutoComplete, in this case, can we create a sub class with these additional properties?

QilongTang: After reviewing your current implementation, I think this is good for now

src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml.cs
```diff
@@ -294,16 +294,18 @@ private void DisplaySuggestions(object sender, RoutedEventArgs e)
             cm.IsOpen = true;
         }
 
-        private void onSuggestion_Click(object sender, RoutedEventArgs e)
+        private void OnSuggestion_Click(object sender, RoutedEventArgs e)
```
QilongTang, line 297: Fixed name violation

src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml.cs
```diff
@@ -294,16 +294,18 @@ private void DisplaySuggestions(object sender, RoutedEventArgs e)
...
+            if (selectedSuggestion.Name.Contains(nameof(Models.NodeAutocompleteSuggestion.MLRecommendation)))
             {
                 ViewModel.dynamoViewModel.PreferenceSettings.DefaultNodeAutocompleteSuggestion = Models.NodeAutocompleteSuggestion.MLRecommendation;
-            } else
+            }
+            else
             {
                 ViewModel.dynamoViewModel.PreferenceSettings.DefaultNodeAutocompleteSuggestion = Models.NodeAutocompleteSuggestion.ObjectType;
             }
+            ViewModel.ResetAutoCompleteSearchViewState();
```
QilongTang, line 308: Added a function to reset the view when switching back and forth

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -30,12 +34,107 @@ public class NodeAutoCompleteSearchViewModel : SearchViewModel
...
 
+        /// <summary>
+        /// For checking if the ML method is selected
+        /// </summary>
+        public bool IsDisplayingMLRecommendation
+        {
+            get
+            {
+                return dynamoViewModel.PreferenceSettings.DefaultNodeAutocompleteSuggestion == Models.NodeAutocompleteSuggestion.MLRecommendation;
+            }
```
QilongTang, line 45: removed setter and changed the view code to be one way binding

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -30,12 +34,107 @@ public class NodeAutoCompleteSearchViewModel : SearchViewModel
...
+            if (!IsDisplayingMLRecommendation)
+            {
+                NoRecommendationsOrLowConfidenceMessage = string.Empty;
+                NoRecommendationsOrLowConfidenceTitle = string.Empty;
+                LowConfidenceMessageAdditional = string.Empty;
+                DisplayNoRecommendationsLowConfidence = false;
+                DisplayLowConfidence = false;
+            }
+            PopulateAutoCompleteCandidates();
+        }
```
QilongTang, line 137: @jesusalvino Let me know if this is what you intend to reset the view, I would like to lazy call this function and not for each time when populating the results because that would be too frequent

QilongTang:
@jesusalvino After my changes, user should be able to switch back and forth as well, see below:

SwitchNodeAutoCompleteOption

PR #13415, opened 2022-10-20
DYN-5332 : handle cli mode in release no console

BogdanZavu:
### Purpose
Keep CLIs alive when there is no console available during run.

### Reviewers
@saintentropy @QilongTang

### DIFF
--- a/src/DynamoApplications/StartupUtils.cs
+++ b/src/DynamoApplications/StartupUtils.cs
@@ -104,6 +104,7 @@ namespace Dynamo.Applications
                 // and issue commands until the extension calls model.Shutdown().
                 bool keepAlive = false;
                 bool showHelp = false;
+                bool noConsoleCli = false;
 
                 // Disables all analytics (Google and ADP)
                 bool disableAnalytics = false;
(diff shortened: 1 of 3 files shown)


mjkkirschner:
instead of only writing to the console in debug, is there not some way of checking if a console is available?

BogdanZavu:
> instead of only writing to the console in debug, is there not some way of checking if a console is available?

There is something something but it's not great and not sure if there are potential issues when there is some sort of redirection from a parent process.
```
[DllImport("kernel32.dll")]
static extern IntPtr GetConsoleWindow();
```
But for the CLI we only care about seeing this during debugging anyway.
For troubleshoot in release we can write in the dynamo file log - but this is currently not working for the CLIs ( will deal with it in an upcoming PR ).
So unless you feel strongly about it I would leave it as is and write in the logs later.

mjkkirschner:
I'm just concerned about third parties that might be using the CLI - I guess the behavior will keep working, just the output will change?

BogdanZavu:
The behavior will be the same and will now work as expected in no console mode.
Keep alive behavior was introduced in 2.15 
Btw we need to cherry pick this in 2,16 branch as well.

mjkkirschner:
there is no current planned 2.16 release as far as I am aware - 2.16.1 hotfix just went out - so unless another release branch is cut and planned , and integrated (into Revit?)I’m not sure where cherry-picking will be to?

BogdanZavu:
So I take it back, there is a change in behavior in release when there is a console. I'll do it differently.
Oh I was hopping it will be cherry picked in RC2.16.1_master since that is currently in Revit ?

mjkkirschner:
@BogdanZavu - this looks ok to me, thanks for keeping the behavior stable - 
As I said above 2.16.1 is already released on our side, you'd need to start the process for a 2.16.2 release.

mjkkirschner (approved):

PR #13426, opened 2022-10-25
Cache will replicate

pinzart90:
Please Note:

### Purpose

### Declarations

### Release Notes

### Reviewers

### FYIs

### DIFF
--- a/src/Engine/EmitMSIL/CodeGenIL.cs
+++ b/src/Engine/EmitMSIL/CodeGenIL.cs
@@ -2005,5 +2005,10 @@ namespace EmitMSIL
             }
             throw new ArgumentException("Identifier node expected.");
         }
+
+        public void Dispose()
+        {
+        }
     }
+
 }
--- a/src/Engine/EmitMSIL/CodeGenIL.cs
+++ b/src/Engine/EmitMSIL/CodeGenIL.cs
@@ -35,7 +35,7 @@ namespace EmitMSIL
         private Dictionary<int, Type> astTypeInfoMap = new Dictionary<int, Type>();
         private StreamWriter writer;
         private Dictionary<int, IEnumerable<ProtoCore.CLRFunctionEndPoint>> methodCache = new Dictionary<int, IEnumerable<ProtoCore.CLRFunctionEndPoint>>();
-        private Dictionary<string, bool> willReplicateCache = new Dictionary<string, bool>();
+        private Dictionary<int, bool> willReplicateCache = new Dictionary<int, bool>();
         private CompilePass compilePass;
         internal (TimeSpan compileTime, TimeSpan executionTime) CompileAndExecutionTime;
 
@@ -1297,7 +1297,8 @@ namespace EmitMSIL
             //if the method name is builtin.valueAtIndex then don't emit a function call yet.
             //instead try to emit direct indexing... to do so, we'll need to wait until ilemit phase
             //so variable dictionary has valid data.
-            if (className == Node.BuiltinGetValueAtIndexTypeName && methodName == Node.BuiltinValueAtIndexMethodName)
+            if (className == Node.BuiltinGetValueAtIndexTypeName &&
+                methodName == Node.BuiltinValueAtIndexMethodName)
             {
                 if (compilePass == CompilePass.MethodLookup)
                 {
@@ -1337,26 +1338,8 @@ namespace EmitMSIL
 
             var isStaticOrCtor = clrFep.IsStatic || clrFep.IsConstructor;
 
-            var doesReplicate = WillCallReplicate(clrFep.Signature, parameters, isStaticOrCtor, args);
-
-            // TODO: Figure out a way to avoid calling WillCallReplicate twice -
-            // once in the GatherTypeInfo phase and again in the emitIL phase.
-            // We should be able to cache the result in the GatherTypeInfo compile pass
-            // and reuse it in the emitIL pass.
-            //if (compilePass == CompilePass.GatherTypeInfo)
-            //{
-            //    if(isRepCall.ContainsKey(node.ID))
-            //    {
-            //        throw new Exception($"ast {node.ID}:{node.Kind} already exists in replicated call map.");
-            //    }
-            //    isRepCall.Add(node.ID, WillCallReplicate(parameters, isStatic, args));
-            //}
-
-            //bool doesReplicate = false;
-            //if(!isRepCall.TryGetValue(node.ID, out doesReplicate))
-            //{
-            //    throw new Exception($"ast { node.ID }:{ node.Kind} does not exist in replicated call map.");
-            //}
+            args.Select(x => x.ID);
+            var doesReplicate = WillCallReplicate(node.ID, parameters, isStaticOrCtor, args);
             if (doesReplicate)
             {
                 EmitILComment("emit replicating call");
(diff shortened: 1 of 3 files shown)


src/Engine/ProtoScript/Runners/LiveRunner.cs
```diff
@@ -1270,13 +1270,11 @@ public void Dispose()
...
             Dictionary<string, IList> input = new Dictionary<string, IList>();
-            var assemblyPath = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location));
 
+            var assemblyPath = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location));
             //TODO_MSIL: remove the dependency on the old VM by implementing
             //necesary Emit functions(ex mitFunctionDefinition and EmitImportStatements and all the preloading logic)
-            msilRuntimeCore = msilRuntimeCore ?? new MSILRuntimeCore(runtimeCore);
-            var codeGenIL = new EmitMSIL.CodeGenIL(input, Path.Combine(assemblyPath, "opCodes.txt"), msilRuntimeCore);
-
+            codeGenIL = codeGenIL ?? new EmitMSIL.CodeGenIL(input, Path.Combine(assemblyPath, "opCodes.txt"), new MSILRuntimeCore(runtimeCore));
```
pinzart90, line 1277: re-using codeGen caches for every execution

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -101,41 +105,43 @@ internal CodeGenIL(IDictionary<string, IList> input, string filePath, ProtoCore.
...
-            var asm = BuilderHelper.CreateAssemblyBuilder("DynamicAssembly", false, access);
-            // 2. Create module builder
-            var mod = BuilderHelper.CreateDLLModuleBuilder(asm, "DynamicAssembly");
-            // 3. Create type builder (name it "ExecuteIL")
-            var type = BuilderHelper.CreateType(mod, "ExecuteIL");
-            // 4. Create method ("Execute"), get ILGenerator 
-            var execMethod = BuilderHelper.CreateMethod(type, "Execute",
-                System.Reflection.MethodAttributes.Static | System.Reflection.MethodAttributes.Private, typeof(void), new[] { typeof(IDictionary<string, IList>),
+            writer = new StreamWriter(logPath);
+            using (Disposable.Create(() => { writer.Close(); writer.Dispose(); }))
```
mjkkirschner, line 109: from what I can find online, dispose will also close the stream, so I think you're safe to just use a regular using statement over the writer.

src/Engine/ProtoCore/Lang/FunctionEndPoint.cs
```diff
@@ -46,6 +46,10 @@ internal List<ParamInfo> FormalParams
 
         public MemberInfo MemberInfo => method.Info;
 
+        private string sig = string.Empty;
+
+        internal string Signature => sig;
```
mjkkirschner, line 51: why do we need the field?

src/Engine/ProtoCore/Lang/FunctionEndPoint.cs
```diff
@@ -66,6 +70,25 @@ internal CLRFunctionEndPoint(FFIMemberInfo method, List<ParamInfo> formalParams,
             this.method = method;
             this.FormalParams = formalParams;
             this.procedureNode = procedureNode;
+
+            if (method.Info is MethodInfo mInfo)
+            {
+                sig += mInfo.ReturnType.FullName + " " + mInfo.DeclaringType.FullName + ":" + mInfo.Name;
```
mjkkirschner, line 76: IMO $"{}" interpolation is easier to read.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1691,25 +1697,33 @@ private static int GetArgumentRank(AssociativeNode val)
...
         }
 
-        private bool WillCallReplicate(List<Type> paramTypes, bool isStaticOrCtor, List<AssociativeNode> args)
+        private bool WillCallReplicate(string fepSig, List<Type> paramTypes, bool isStaticOrCtor, List<AssociativeNode> args)
         {
-            if (isStaticOrCtor)
+            if (willReplicateCache.TryGetValue(fepSig, out bool willReplicate))
             {
-                return !DoesParamArgRankMatch(paramTypes, args);
+                return willReplicate;
```
mjkkirschner, line 1704: so - if the arguments change won't this value potentially be incorrect?

I feel that either
1. We also need to put the arguments into the signature as the key, OR
2. We need to fallback to replication when making a direct call, maybe using a try catch that gets emitted?

aparajit-pratap (changes requested):
I'm not sure I understand this. The decision to replicate or not (true or false) depends on not just the function signature but also the arguments used in calling that function. It could be that in one case, the call replicates, while in the other it doesn't for the same signature, so I don't believe that using the function signature as the key is going to serve the purpose.

aparajit-pratap:
I tried using the AST node ID as the key for this cache (see commented code [here](https://github.com/DynamoDS/Dynamo/blob/f787a2481f27cf40d69faab27c0cdc8db42ae39f/src/Engine/EmitMSIL/CodeGenIL.cs#L1336)), the only problem with this approach is that the node ID changes midway as some ASTs are recreated between compilation passes, which isn't expected. I believe one way to do that is to also cache those ASTs so that they can be reused between compilation passes instead of recreating them again.

src/Engine/EmitMSIL/CodeGenIL.cs
pinzart90: Yep, totally forgot about the arguments

aparajit-pratap:
> InheritID

@mjkkirschner to use this, won't we need to cache the previously created ASTs to inherit their IDs anyway, in which case, we might as well, just skip the creation step and reuse these ASTs?

pinzart90:
@aparajit-pratap the only place I see that creates new FunctionCallNodes on the fly is in `EmitRangeExpression`. Weca stabilize the generated funcCall's ID by inheriting from the RangeExpressionNode. Do you think that will work?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1883,6 +1879,12 @@ private Type EmitRangeExprNode(AssociativeNode node)
 
 
             var rangeExprFunc = AstFactory.BuildFunctionCall(methodName, arguments);
+
+            // We need to keep the ID of the generated fuction call stable
+            // in order to make use of caches in between different compile stages.
+            // So just copy it from the rangeExpression node
+            rangeExprFunc.InheritID(range.ID);
```
pinzart90, line 1886: this is the only `InheritID` used.

aparajit-pratap: So this is AST is still created twice - once in `gatherTypeInfo` and once in `emitIL` phase?

pinzart90: yes

aparajit-pratap: I'm okay with this if you think that caching the AST will be too much work for little benefit.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1337,26 +1338,8 @@ private Type EmitFunctionCallNode(AssociativeNode node)
...
-            //        throw new Exception($"ast {node.ID}:{node.Kind} already exists in replicated call map.");
-            //    }
-            //    isRepCall.Add(node.ID, WillCallReplicate(parameters, isStatic, args));
-            //}
-
-            //bool doesReplicate = false;
-            //if(!isRepCall.TryGetValue(node.ID, out doesReplicate))
-            //{
-            //    throw new Exception($"ast { node.ID }:{ node.Kind} does not exist in replicated call map.");
-            //}
```
aparajit-pratap, line 1353: Please delete commented line 27 too.

aparajit-pratap (approved):
one question and one comment, then LGTM

src/Engine/EmitMSIL/CodeGenIL.cs
pinzart90: the cost to recreate the AST for `rangeExprFunc` seems to be very low

PR #13436, opened 2022-10-26
increase number of supported cases for direct function calls

mjkkirschner:
### Purpose

This PR begins to address https://jira.autodesk.com/browse/DYN-5321, but we'll keep this open as there are remaining known cases which are still failing. Unclear how to fix some of these at the moment and they require more discussion.

1. PR attempts to unify custom compile time will replicate logic with DFSTraversal which is already pretty good at determining compile time type information. This ends up simplifying the `WillReplicate` checks. Uses two compile passes to achieve this.
2. When determining if call will replicate also includes the selected `FEP` as a parameter so we can check what attributes each parameter has - these are useful when determining if a FEP will replicate - for example the `arbitraryRankArrayImport` attribute should make it so a parameter is not replicated over except with specific replication guides.
3. PR attempts to replace `ArrayTypeStatistics` logic in `EmitExprListNode` method with DFSTraversal - simplifies that code and uses two compile passes to achieve this.
4. introduces `StrictDirectFunctionCallValidation` on the `MSILRuntime` and CodeGenIL` objects, if these bools are set true exceptions will be thrown when direct function calls should made, but can't. If these values are false we'll fallback to replicated calls.

### Declarations

### Release Notes

### Reviewers

### FYIs


src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1612,46 +1588,77 @@ private bool DoesParamArgRankMatch(List<Type> parameterTypes, List<AssociativeNo
...
                 }
                 else return false;
                 argIndex++;
             }
             return true;
         }
 
-        private static int GetRank(Type type)
+        private static bool DoesParamArgRankMatchInner(ProtoCore.CLRFunctionEndPoint fep,Type p, Type t, int paramIndex)
         {
```
aparajit-pratap, line 1626: `DoesParamArgRankMatchInner(ProtoCore.CLRFunctionEndPoint fep,Type paramType, Type argType, int paramIndex)`

aparajit-pratap: Can we rename this function as the `inner` in its name sounded like it was matching the arg and param types of each element of an array?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1612,46 +1588,77 @@ private bool DoesParamArgRankMatch(List<Type> parameterTypes, List<AssociativeNo
...
+
+        private static int GetRank(ProtoCore.CLRFunctionEndPoint fep,Type type,int paramIndex)
+        {
+            if(type == typeof(IList)){
+                return -1;
+            }
+            if(type == typeof(object) && fep.ParamAttributes[paramIndex].Any(attr => attr is ArbitraryDimensionArrayImportAttribute))
+            {
+                return -1;
+            }
```
aparajit-pratap, line 1661: I believe a parameter of type `IEnumerable<T>`  with `ArbitraryDimensionImportAttribute` results in an arbitrary dimension array of type `T`, in which case the `type == typeof(object)` check doesn't seem quite right, does it?

mjkkirschner: Hi @aparajit-pratap, this check is specifically for a test that was calling `List.Contains` - this method and some of the other core list methods accept object parameters marked with this attribute:
https://github.com/DynamoDS/Dynamo/blob/c9a23ab9c9e4110ac798b33a9d4afad183718034/src/Libraries/CoreNodes/List.cs#L51

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1585,11 +1555,17 @@ private void EmitUnmarshalFunctionArgs(List<AssociativeNode> args, List<Type> pa
...
+        private bool DoesParamArgRankMatch(ProtoCore.CLRFunctionEndPoint fep, List<Type> parameterTypes, List<AssociativeNode> args)
         {
             var argIndex = 0;
+          
             foreach (var p in parameterTypes)
             {
+                var currentArg = args[argIndex];
+                if (compilePass == CompilePass.GatherTypeInfo)
+                {
+                    DfsTraverse(currentArg);
```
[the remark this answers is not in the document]
mjkkirschner: yes, exactly.

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: yeah, I'll think of something better.

src/Engine/EmitMSIL/CodeGenIL.cs
aparajit-pratap: I guess my question is how do we deal with parameters like this:
```
public void foo([ArbitraryDimensionArrayImport] IEnumerable<int> bar) { ... }
```
that are passed an `int` or list of any rank of `int`s as argument.

mjkkirschner: @aparajit-pratap I need to test it, but I agree it's a concern for example - `FirstItem(list:var[]..[]` should work with a list or a single item, I'm not sure what happens today though, I can add a test for that case at the minimum.

mjkkirschner: ah @aparajit-pratap to clarify, are you asserting that because `int` is a `valueType` that my check with `object` is actually too strict?

Now that I think about it - it does make sense that basically any use of this attribute on a parameter means it cannot replicate - EXCEPT in cases where there are replicate guides that force an iteration - does that sound right?

aparajit-pratap: 1. I believe there are 2 cases where ArbitraryDimension attribute applies, where it converts from:
a. `IList` to `var[]..[]` - I think you've taken care of this
b. `IEnumerable<T>` to `T[]..[]` - my question was about this case, how would you handle it as I don't think checking for `object` type would work here
2. I'm unsure what happens when it is applied to `object`. If it does, your check for the type being equal to `object` makes sense.
3. I agree, it would be good to test this logic with a single item arg passed to a parameter with the attribute

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1634,46 +1639,105 @@ private bool DoesParamArgRankMatch(List<Type> parameterTypes, List<AssociativeNo
...
+                    return false;
+                }
+
+            }
+            return true;
+        }
+
+        private static int GetRank(ProtoCore.CLRFunctionEndPoint fep,Type type,int paramIndex)
         {
+            //TODO (Alternatives to consider)
```
mjkkirschner, line 1723: I'd like some feedback on these alternatives
1. we use the imported protocore typesystem and just check if the imported rank is arbitrary
2. we try to minimally recreate the logic that the type system encodes to try to remove our dependency on it.

mjkkirschner:
Even though as you can see from the comments this PR leaves many failures when `StrictDirectFunctionCallValidation` is enabled, I would like to get this PR in as I believe it's an improvement on several fronts in code simplification and does fix a bunch of cases.

I can file another followup for the remaining classes of problems - I think they are bigger problems that will require discussion - in addition some are already in progress and will be resolved by compile time overload resolution.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1380,21 +1358,48 @@ private void EmitUnmarshalFunctionArgs(List<AssociativeNode> args, List<Type> pa
...
+        /// Checks rank, type, and attributes to make this determination.
+        /// </summary>
+        /// <param name="fep"></param>
+        /// <param name="parameterTypes"></param>
+        /// <param name="args"></param>
+        /// <returns></returns>
+        /// <exception cref="Exception"></exception>
+        private bool DoParamsRankMatchArgs(ProtoCore.CLRFunctionEndPoint fep, List<Type> parameterTypes, List<AssociativeNode> args)
         {
+            //TODO this can throw because it's possible that the FEP is incorrect
```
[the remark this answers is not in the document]
mjkkirschner: yes

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1408,46 +1413,105 @@ private bool DoesParamArgRankMatch(List<Type> parameterTypes, List<AssociativeNo
...
+                }
+
+            }
+            return true;
+        }
+
+        private static int GetRank(ProtoCore.CLRFunctionEndPoint fep,Type type,int paramIndex)
         {
+            //TODO (Alternatives to consider)
+            if (fep.procedureNode.ArgumentTypes.ElementAt(paramIndex).rank == DSASM.Constants.kArbitraryRank)
```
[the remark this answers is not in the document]
mjkkirschner: while yes, that is true, IMO in the case of an arbitrary rank parameter, it does not matter what the arg rank is, we can mark them as compatible as we should not replicate.

If you prefer the alternative check that uses the individual arg/param types then I can switch to that, though I expect it will eventually get more complex or perhaps I can find a way of converting the clr type to a dstype and querying whether it is arbitrary rank.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1408,46 +1413,105 @@ private bool DoesParamArgRankMatch(List<Type> parameterTypes, List<AssociativeNo
...
+                {
+                    return true;
+                }
+
+                var argRank = GetRank(fep,argType,paramIndex);
+                var paramRank = GetRank(fep,paramType, paramIndex);
+
+                //if the param is an arbitrary rank array and
+                //arg is some array type, replication should not occur.
+                if (paramRank == -1 && argRank > 0)
```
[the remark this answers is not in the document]
mjkkirschner: yes, rank 0 would just be a single item I believe.

One thing that is unclear to me -

If a parameter of rank -1 (arbitrary) will never replicate with a single item OR an array, then this check is unnecessary.

@aparajit-pratap is that always the case?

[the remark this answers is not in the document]
aparajit-pratap: Comparing a single parameter to an argument, if the parameter has the arbitrary rank attribute, then it doesn't matter what rank the argument has, it will not replicate. However, for a function with more than one parameter, there can be one parameter with the attribute, and another one that doesn't. In this case, the function can still replicate over the parameter that doesn't have the attribute. So only if `DoesParamRankMatchArg` returns true for all parameter-argument pairs for a function call, then we can decide not to replicate.

src/Engine/EmitMSIL/CodeGenIL.cs
aparajit-pratap: It looks like the `ArbitraryDimension` attribute applies to `object` as well, so you're right about checking for object type. Scratch (2) above.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1380,21 +1358,48 @@ private void EmitUnmarshalFunctionArgs(List<AssociativeNode> args, List<Type> pa
...
+            
             foreach (var p in parameterTypes)
             {
-                if (CoreUtils.IsPrimitiveASTNode(args[argIndex]) || args[argIndex] is CharNode)
+                var currentArg = args[argIndex];
+                if (compilePass == CompilePass.GatherTypeInfo)
+                {
+                    DfsTraverse(currentArg);
+                }
+                if (CoreUtils.IsPrimitiveASTNode(args[argIndex]) || currentArg is CharNode)
```
pinzart90, line 1394: maybe we can use `currentArg` instead of `args[argIndex]` here

src/Engine/EmitMSIL/CodeGenIL.cs
pinzart90, line 1475: can we use Constants.kArbitraryRank instead of -1 ?

src/Engine/EmitMSIL/CodeGenIL.cs
pinzart90: Well we did have discussions about removing the dependency on the old VM.
Not sure what that means now, maybe we can share some of that logic or re-write it.
I think we can use the protocore system now and we can separate out later

src/Engine/EmitMSIL/CodeGenIL.cs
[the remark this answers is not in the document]
mjkkirschner: ahh, good call, let me see if I can write a test for that case and address it.

mjkkirschner:
closing this in favor of emitMSIL-v2 version. Will try to address review comments there.

PR #13437, opened 2022-10-26
Leave RenderPackage data intact when passed to AggregateRenderPackages

saintentropy:
### Purpose

The purpose of this PR is to modify the behavior of the `AggregateRenderPackages` method.  Current method is triggered  based on the `NodeModel.RenderPackagesUpdated` event.  Currently a the `AggregateRenderPackages` modify the `RenderPackage` object it is handed via the event when it is processing data to add to the view.  The issue with this modification is that other consumers this event will receive a modified `RenderPackage` which does not contain the original data.  The PR changes the the RenderPackage data is processed so that it is not modified.

In testing this PR I also noticed some scenarios today which Render incorrectly.  Edge cases where a list is created combining objects that are rendered via instance data, objects with texture maps, and regular geometry.

Todo
- [x] Add tests for broken render scenarios 
- [x] Add tests to validate no changes to the RenderPackage

### Declarations

### Release Notes

Update API behavior for AggregateRenderPackages.
Fix regressions in background preview render for some list of objects.

### Reviewers

### FYIs

@twastvedt


mjkkirschner:
Hi @saintentropy 
1. please use the reviewers section in github when possible
2. It seems there are 2 failing tests here that are real, one is a UI test that is flaky. FYI @QilongTang 
3. What release is this intended for?

mjkkirschner:
@saintentropy to review this for correctness I think I'm going to need a more detailed description about what was changed/ how it works - at this point I have no memory of / bandwidth to get deep into this code.

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -1836,29 +1855,28 @@ internal virtual void AggregateRenderPackages(IEnumerable<HelixRenderPackage> pa
...
                     {
                         //For each range of mesh vertices add the mesh data and texture map to the scene
-                        var meshVertexCountTotal = 0;
                         for (var j = 0; j < rp.MeshVerticesRangesAssociatedWithTextureMaps.Count; j++)
                         {
                             var range = rp.MeshVerticesRangesAssociatedWithTextureMaps[j];
                             var startIndex = range.Item1; //Start mesh vertex index
                             var count = range.Item2 - range.Item1 + 1; //Count of mesh vertices
-                            var uniqueId = baseId + ":" + j + MeshKey;
+                            var uniqueId = baseId + ":" + j + MeshKey + "_Texture";
```
mjkkirschner, line 1863: oh I see, just to make debugging it easier?

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -1836,29 +1855,28 @@ internal virtual void AggregateRenderPackages(IEnumerable<HelixRenderPackage> pa
...
                             
-                            AddMeshData(uniqueId, rp,startIndex,count, drawDead, baseId, rp.TextureMapsList[j], rp.TextureMapsStrideList[j]);
+                            AddMeshData(uniqueId, m,startIndex,count, drawDead, baseId, rp.TextureMapsList[j], rp.TextureMapsStrideList[j], rp.Transform, rp.RequiresPerVertexColoration);
```
mjkkirschner, line 1865: can you split up these giant lines into two lines?

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -1869,45 +1887,98 @@ internal virtual void AggregateRenderPackages(IEnumerable<HelixRenderPackage> pa
...
-                        var vertexRanges =
-                            new List<(int start, int end)>(rp.MeshVertexRangesAssociatedWithInstancing.Values.ToList());
-
-                        RemoveMeshGeometryByRange(vertexRanges, m);
+                        //Otherwise, add ranges of mesh geometry to exclude for regions already generated related to instancing.
+                        MeshVertexRangesToRemove.AddRange(rp.MeshVertexRangesAssociatedWithInstancing.Values.ToList());
                     }
 
-                    AddMeshData(id, rp, 0, m.Positions.Count, drawDead, baseId, rp.Colors, rp.ColorsStride);
+                    //If mseh vertex ranges have been utilized previously for instantiating instanced geometry or multiple texture maps we only process the remaining mesh data
```
mjkkirschner, line 1920: mesh (sp)

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -1869,45 +1887,98 @@ internal virtual void AggregateRenderPackages(IEnumerable<HelixRenderPackage> pa
...
+                    //We clone the mesh object so that we do not modify the render package data.
+                    if (MeshVertexRangesToRemove.Any())
+                    {
+                        if (mCopy == null)
```
mjkkirschner, line 1924: why are these null checks necessary?

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -2339,15 +2440,24 @@ private void CreateOrUpdateDisplayLabels(string nodeId)
             }
         }
 
-        private DynamoGeometryModel3D CreateDynamoGeometryModel3D(HelixRenderPackage rp, bool isHitTestVisible = true, IEnumerable<byte> colors = null, int colorStride = 0)
+        /// <summary>
+        /// Create the Mesh Geometry Model object for the scenen
```
mjkkirschner, line 2444: scene sp

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -2381,17 +2491,25 @@ private DynamoGeometryModel3D CreateDynamoGeometryModel3D(HelixRenderPackage rp,
             return meshGeometry3D;
         }
 
-        private DynamoLineGeometryModel3D CreateLineGeometryModel3D(HelixRenderPackage rp, double thickness = 1.0, 
+        /// <summary>
+        /// Create the Line Geometry Model object for the scenen
```
mjkkirschner, line 2495: spelling

mjkkirschner (approved):
it seems okay, a few questions, and I think it would be good if we could create a regression test for the failure - maybe 2 subscribers to the event?

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
```diff
@@ -1787,45 +1791,60 @@ internal virtual void AggregateRenderPackages(IEnumerable<HelixRenderPackage> pa
...
                             //For each range of line vertices add the line data and instances to the scene
-                            var lineVertexCountTotal = 0;
                             var j = 0;
                             foreach (var item in rp.LineVertexRangesAssociatedWithInstancing)
                             {
                                 var range = item.Value;
                                 var startIndex = range.Item1; //Start line vertex index
                                 var count = range.Item2 - range.Item1 + 1; //Count of line vertices
-                                var uniqueId = baseId + ":" + j + LinesKey;
+                                var uniqueId = baseId + ":" + j + LinesKey + "_Instance";
```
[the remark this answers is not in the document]
saintentropy: Their current implementation has a collision on uniqueIds if a render package contains instanced mesh and multiple texture maps.  This change just makes sure the unique Ids for each type of helix geometry is actually unique.

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
saintentropy: See above but yes

src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs
saintentropy: ahh good catch... I was previously working on the copy in a few places.

src/VisualizationTests/HelixWatch3DViewModelTests.cs
```diff
@@ -1308,7 +1308,7 @@ public void InstancedMeshesAndLinesAreAddedToBackGroundPreviewForEachMatrixWhenS
             Assert.AreEqual(0, BackgroundPreviewGeometry.NumberOfVisibleCurves());
             ViewModel.RenderPackageFactoryViewModel.ShowEdges = true;
             //this graph displays a grid, cones, cone edges, cube instances, cube edge instances.
-            Assert.AreEqual(3, BackgroundPreviewGeometry.NumberOfVisibleCurves());
+            Assert.AreEqual(2, BackgroundPreviewGeometry.NumberOfVisibleCurves());
```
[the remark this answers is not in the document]
saintentropy: Yeah this is where their was a error in the original test.  `NumbersOfVisibleCurves` removes grids from the number you return.  So the expected behavior is 2.  Cube edges and Cone Edges.  Not sure what the third was before.  Probably an empty item.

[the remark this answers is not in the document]
saintentropy: Same for the axes but I can't remember if that is curve or mesh geometry.

src/DynamoCoreWpf/Rendering/HelixRenderPackage.cs
```diff
@@ -980,11 +980,13 @@ public bool ContainsTessellationId(Guid id)
         public MeshGeometry3D Mesh
         {
             get { return mesh; }
+            internal set { mesh = value; }
```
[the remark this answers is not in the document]
saintentropy: Yep no longer necessary

twastvedt:
> it seems okay, a few questions, and I think it would be good if we could create a regression test for the failure - maybe 2 subscribers to the event?

I'm looking into making this test. As far as I can tell I need to be able to modify the feature flags, as the issue only happens when the `graphics-primitive-instancing` flag is false, but in test mode that flag is set to true. I did a little looking around but didn't see another place where those flags are being modified. Does that sound right, and is that possible?

mjkkirschner:
> > it seems okay, a few questions, and I think it would be good if we could create a regression test for the failure - maybe 2 subscribers to the event?
> 
> I'm looking into making this test. As far as I can tell I need to be able to modify the feature flags, as the issue only happens when the `graphics-primitive-instancing` flag is false, but in test mode that flag is set to true. I did a little looking around but didn't see another place where those flags are being modified. Does that sound right, and is that possible?

Hi @twastvedt - maybe @saintentropy can confirm - I thought this issue only happened when instancing was true!?

twastvedt:
> Hi @twastvedt - maybe @saintentropy can confirm - I thought this issue only happened when instancing was true!?

OK, found a couple things.
1. Now remembering that Craig tried to tell me that I can just test the multiple texture maps part of this, as that doesn't depend on a feature flag. Doing that!
2. But, for reference, it looks like AggregateRenderPackages only modifies the render package if the geometry uses instancing but the flag is set to false (https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCoreWpf/ViewModels/Watch3D/HelixWatch3DViewModel.cs#L1868-L1869). Otherwise I _think_ it's just sending the render package with instancing embedded directly to Helix. I'm fuzzy on the details still though.

At any rate, I added a test which I verified fails before the PR, now passes. Still looking at the rendering edge cases.

src/VisualizationTests/HelixWatch3DViewModelTests.cs
```diff
@@ -1183,7 +1187,33 @@ public void CanTagGeometryWhenClickingArrayItemInWatchNode()
...
+        public void AggregateRenderPackagesDoesntMutateRenderPackage()
+        {
+            // Regression test for DYN-5329:
+            // HelixWatch3DViewModel.AggregateRenderPackages event handler shouldn't mutate render packages.
+
+            OpenVisualizationTest("MultipleTextureMaps.dyn");
+
+            var output = ViewModel.CurrentSpace.NodeFromWorkspace<CreateList>(Guid.Parse("04fcc9b7f80b43c99923b0dac8930e77"));
+
+            output.RenderPackagesUpdated += (NodeModel nodeModel, RenderPackageCache renderPackages) => {
```
mjkkirschner, line 1202: @twastvedt please unsub this handler when the test is over, it's simpler than reasoning about if you need to or not with a memory profiler.

twastvedt: Oh right. Yes, sorry.

twastvedt:
@mjkkirschner I think this is ready to go? Just wanted to give you the chance to look at the test I added, if you want to. It's pretty straightforward though. I added an image comparison test which checks for the render bug Craig discovered in the process of this PR. I verified that it renders something different (the bug) before this PR's changes.

mjkkirschner:
thanks @twastvedt - I will merge now, can you please cherry pick the squashed commit to 2.17?

PR #13453, opened 2022-11-01
dnenov-documentationbrowserviewextension-update

dnenov:
### Purpose

WIP Update to the DocumetnationBrowser View Extension

changes dec 2022

### Declarations

- [x] User-facing strings, if any, are extracted into `*.resx` files

### Release Notes

- now correctly undo/redo insert command
- will display toast notification if Example .dyn file is not found in fallback_folder
- now correctly nests inserted Example in a new group
- now shows the original name of the node if the node has been renamed (will no longer show the Node Type which provided the same value as the Node Name)
- visual style underline text to breadcrumbs
- breadcrumbs expand node library behavior 
- multiple bug fixes and small changes

#### Known Issues

- insert Example is triggered twice (will cause fake warning)
- nested group causes inaccuracies in placing the Example graph in the X direction
- after using the DocumentationBrowserViewExtension, the Md2Html process is triggered, but never killed in the Task Manager

### Reviewers

@Amoursol 
@reddyashish


src/AssemblySharedInfoGenerator/AssemblySharedInfo.tt
```diff
@@ -68,7 +68,7 @@ using System.Runtime.InteropServices;
 [assembly: AssemblyFileVersion("<#= this.MajorVersion #>.<#= this.MinorVersion #>.<#= this.BuildNumber #>.<#= this.RevisionNumber #>")]
 <#+
 int MajorVersion = 2;
-int MinorVersion = 17;
```
reddyashish, line 71: Do not commit this file.

dnenov: Reverted

src/Config/upiconfig.xml
```diff
@@ -8,13 +8,13 @@
         <upi_attribute name='id' value='DYN'/>
         <upi_element name='level'>
           <upi_attribute name='name' value='release'/>
-          <upi_attribute name='id' value='2.17.0'/>
+          <upi_attribute name='id' value='2.16.0'/>
```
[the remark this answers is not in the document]
dnenov: reverted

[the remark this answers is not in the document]
dnenov: reverted

dnenov:
Hi @reddyashish - I have updated the PR to contain the Dynamo Example Graphs. Still, no Slack on my side, so no way to talk to you directly, but I did drop a message to Aaron. Would you be able to help me in bringing these changes to https://github.com/DynamoDS/Dynamo/tree/RC2.17.0_master ?

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
```diff
@@ -371,6 +371,11 @@ internal void InsertGraph()
         internal delegate void InsertDocumentationLinkEventHandler(object sender, InsertDocumentationLinkEventArgs e);
         internal event InsertDocumentationLinkEventHandler HandleInsertFile;
 
+        private string DynamoGraphFromMDFilePath(string path)
+        {
+            return path.Replace(".md", ".dyn").Replace("%20", " ");
```
reddyashish, line 376: You can use Path.GetFileNameWithoutExtension to trim the extension name and add .dyn. That would be robust.

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
```diff
@@ -229,19 +237,26 @@ private void GroupInsertedGraph(List<AnnotationViewModel> existingGroups, string
             DoEvents();
 
             var annotation = this.DynamoViewModel.Model.CurrentWorkspace.AddAnnotation(Resources.InsertedGroupSubTitle, Guid.NewGuid());
-            annotation.AnnotationText = graphName;
-            DoEvents();
+            if (annotation != null)
+            {
+                annotation.AnnotationText = graphName;
+                DoEvents();
```
mjkkirschner, line 243: what is up with this use of DoEvents from application code?

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
```diff
@@ -229,19 +237,26 @@ private void GroupInsertedGraph(List<AnnotationViewModel> existingGroups, string
...
-            var styleItem = annotationViewModel.GroupStyleList.First(x => x.Name.Equals(DynamoProperties.Resources.GroupStyleDefaultReview));
-            var groupStyleItem = new GroupStyleItem {Name = styleItem.Name, HexColorString = styleItem.HexColorString};
-            annotationViewModel.UpdateGroupStyle(groupStyleItem);
+                var styleItem = annotationViewModel.GroupStyleList.First(x => x.Name.Equals(DynamoProperties.Resources.GroupStyleDefaultReview));
+                var groupStyleItem = new GroupStyleItem {Name = styleItem.Name, HexColorString = styleItem.HexColorString};
+                annotationViewModel.UpdateGroupStyle(groupStyleItem);
+
+                DynamoSelection.Instance.ClearSelection();
+                DynamoSelection.Instance.Selection.AddRange(annotation.Nodes);
+                if(annotation.HasNestedGroups)
```
mjkkirschner, line 254: braces {}

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
```diff
@@ -267,7 +281,18 @@ private List<ISelectable> GetCurrentSelection()
 
             foreach (var selected in DynamoSelection.Instance.Selection)
             {
+                foreach (var group in this.DynamoViewModel.CurrentSpaceViewModel.Annotations)
+                {
```
mjkkirschner, line 285: deserves a comment and I think can be potentially slow.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -2081,10 +2081,11 @@ private void InsertWorkspace(WorkspaceModel ws, ExtraWorkspaceViewInfo viewInfo
         {
             var nodes = ws.Nodes;
             var connectors = ws.Connectors;
+            var notes = viewInfo?.Annotations;
 
             DynamoSelection.Instance.ClearSelection();
 
-            if (nodes == null || !nodes.Any() || NodesAlreadyLoaded(nodes))
+            if ((nodes == null && notes == null) || (!nodes.Any() && !notes.Any()) || (NodesAlreadyLoaded(nodes) || NotesAlreadyLoaded(notes)))
```
mjkkirschner, line 2088: don't you want Empty here instead of !Any?

dnenov: Totally, but Workspace.Nodes is IEnumerable, it doesn't have .Empty(). I actually wouldn't have thought of using .Empty() in any case, so thanks for the prompt! I will leave it as is, if that's ok.

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
dnenov: I know how you feel about nested for loops and nested if statements! This one had both and a bit of a twisted flow logic, so I split it into two methods and documented it more thoroughly.

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
dnenov: Done.

src/DocumentationBrowserViewExtension/DocumentationBrowserViewExtension.cs
dnenov: I have sprinkled them like Magic dust :D I cleaned up the code and got rid of all but one DoEvents(), which I still need in order to get the desired result.

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
dnenov: Yup, great idea! Done.

src/DocumentationBrowserViewExtension/DocumentationBrowserView.xaml.cs
```diff
@@ -133,25 +137,35 @@ async void InitializeAsync()
...
                 this.viewModel.InsertGraph();
             }
+
+            MessageIsHandled = true;
+            HandleInsert();
+        }
+
+        private async void HandleInsert()
+        {
+            await Task.Delay(MESSAGE_HANDLE_DELAY);
```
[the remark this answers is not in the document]
dnenov: I absolutely understand, and I was looking for an alternative way to solve this, but I cannot find a way around it. What I am observing is that [WebView2.WebMessageReceived event](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.winforms.webview2.webmessagereceived?view=webview2-dotnet-1.0.1418.22) is firing twice. In my case, this triggers the second insertion of a graph inside Dynamo, which is definitely not what we want.  Since I am raising a Warning and not allowing this to happen, the result is a faulty prompt being raised. The only way I managed to suppress this behavior was to introduce a 'timer' and only allow a single run every 2 seconds. The Task.Delay is inside an async method, so it's not halting the main thread, I am just timing out a boolean. I have been puzzling over this for quite a while, I don't know how else to solve it @mjkkirschner  if you have stumbled on this issue, or if you could suggest a more elegant way of handling this, I would do it!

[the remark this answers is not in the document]
mjkkirschner: @dnenov can you give me more info on what message is triggering the handler? 

Maybe you can tell me a bit more about the design here - when a user hits insert what happens? 

I would assume you could skip this handler and use the scriptingObject directly to marshal your call to the c# side - I guess my question is - do you have to use this event or is there an alternative?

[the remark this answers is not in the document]
dnenov: Sure thing. To my understanding, this is what's happening:
- the DocumentationBrowser is an HTML page displayed through the WebView2 control inside the WPF Window
- user engagement happens on the HTML side. Clicking the 'Insert' button triggers a window.chrome.webview.postMessage(message) where the message is an arbitrary string
- we capture that message by subscribing to the 'WebView2.WebMessageReceived'. We use the message string just to guide us to the desired functionality, the rest of it is on the C# side
- when testing, the **postMessage** is only triggered once on the HTML side. However, the **WebMessageReceived** is triggered twice. I cannot find any documentation confirming this redundancy, but it is what I am observing.

What would be an alternative design pattern between the hosted HTML page and the WPF? At the moment, this seems to be resolving the issue, but I will change it to a more robust pattern if possible.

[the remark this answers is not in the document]
mjkkirschner: you can also use `hostObjects` to call a method on a c# object.
https://github.com/DynamoDS/Dynamo/blob/master/src/LibraryViewExtensionWebView2/web/library/library.html#L157
https://github.com/DynamoDS/Dynamo/search?q=bridgeTwoWay

Though I do not believe we do not see the same double execution behavior with the postMessage command in librareieJS which is also an html/js application using webView2 so I would debug the startup code to make sure that you do not accidentally subscribe the same webView2 instance to this event twice - ie do you see two event handlers in the list of subscribers?

you can use quickwatch with `GetInvocationList()` on the event to check this.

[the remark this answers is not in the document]
dnenov: Hi @mjkkirschner - great news, @reddyashish and I finally solved this puzzle! What was happening was that the CoreWebView2 was being initialized every time a new Node help menu was opened. So by the second Node help, we would have subscribed 2, and then 3 times, and so on. 
Thank you for chasing me on this one, I think we solved a bigger problem by not allowing me to 'patch' things up. I will mark this one as resolved.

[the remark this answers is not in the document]
mjkkirschner: awesome, great work guys.

reddyashish (approved):

reddyashish:
All tests passed. Merging this PR. @dnenov I think you can update this PR description as you fixed the issue some of the known issues. Please add any related future improvements also to the description so that they can be tracked.

src/DocumentationBrowserViewExtension/ScriptingObject.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// <summary>
+        /// The method takes a string code and will execute a different program based on that code
+        /// </summary>
+        /// <param name="message"></param>
+        public void Notify(string message)
+        {
+            if (string.Equals(message, "insert"))
+            {
+                // Insert the graph inside the current worskspace
+                this.viewModel.InsertGraph();
```
RobertGlobant20, line 29: @dnenov do you remember if this functionality of inserting the graph from the DocumentationBrowser was tested also with xml dyn files (most of the dyn files in fallback_docs folder are in json format but there are some cases in which we are using xml files but is crashing when doesn't have notes).
FYI: @QilongTang 
image

dnenov: You are right, @RobertGlobant20 , I don't think there was anything to test with regarding the XML files. I do remember trying to find valid XML dynamo files to test with and failing (Dynamo itself wouldn't open them).

PR #13464, opened 2022-11-03
Dyn 5317 check webview2 installed

RobertGlobant20:
### Purpose
Check that WebView2 is installed before starting Dynamo
I added a validation that before starting the DynamoSandbox app we will check if the WebView2 Runtime is installed or not, if is not installed we will show a MessageBox with the link of the WebView2 installer and then after pressing the OK button close Dynamo.
Also in the DynamoMessageBox I added a RichTextBox that will be visible under very specific circumstances, this RichTextBox allows to have a blue hyperlink as a part of the text so the user can click it and open the webview2 installer webpage.

### Declarations

### Release Notes

Check that WebView2 is installed before starting Dynamo

### Reviewers

@QilongTang 

### FYIs

@filipeotero


RobertGlobant20:
GIF showing behavior when WebView2 Runtime is not installed.
DynamoMissingWebView2

RobertGlobant20:
GIF showing behavior when WebView2 is installed
DynamoWebView2Installed

src/DynamoCoreWpf/UI/Prompts/DynamoMessageBox.xaml
```diff
@@ -115,6 +116,15 @@
...
                            TextWrapping="Wrap" />
+                <localui:CustomRichTextBox x:Name="ContentRichTextBox" 
+                                           IsDocumentEnabled="True"
+                                           Visibility="Collapsed"
+                                           DockPanel.Dock="Left"
+                                           Grid.Row="1"
+                                           Margin="20"
+                                           CustomText="{Binding BodyText}"
+                                           Style="{StaticResource CustomRichTextBoxStyle}">
+                </localui:CustomRichTextBox>
```
QilongTang, line 127: Would you also test if normal messages display correctly?

QilongTang: Oh you are adding a RIchTextBox instead of forcing it, I guess it is less impactful then

src/DynamoSandbox/DynamoCoreSetup.cs
```diff
@@ -57,6 +58,10 @@ public void RunApplication(Application app)
         {
             try
             {
+                //This line validates if the WebView2 Runtime is installed in the computer, if is not we return and then exit Dynamo
+                if (!WebView2Utilities.ValidateWebView2RuntimeInstalled())
+                    return;
```
QilongTang, line 63: I think this is fine for now. For integrations, usually host will include WebView2 Runtime, for any integration launching Dynamo as another process, this should hit

QilongTang (approved):

QilongTang:
@RobertGlobant20 Can you check the regression?

src/DynamoSandbox/DynamoCoreSetup.cs
mjkkirschner: do we really need to exit Dynamo?

QilongTang: @mjkkirschner It would be safe to do so since Splash Screen depending on WebView2, unless we want to totally skip that experience when runtime not installed

PR #13471, opened 2022-11-04
fix intermittent issue loading icons (timing)

mjkkirschner:
### Purpose

You can see that in the original MSWebBrowser version of the library extension - we use a single global `currentImage` which gets set to some base64 encoded string as images are loaded.
https://github.com/DynamoDS/Dynamo/blob/RC2.13.0_master/src/LibraryViewExtensionMSWebBrowser/web/library/library.html#L170

When we moved to WebView2 - there is no longer a simple synchronous invoke method - so the `completeReplaceImages` got changed to iterate all images and set them to a single base64 encoded string if they had not already been set to some base64 string value- that was previously requested by `ReplaceImages` . IMO this is not robust (it's also potentially slow) , as we have seen by intermittent failures on 3 machines. I believe it is very prone to timing issues. Kind of just luck of the draw.

This change moves to using the `sync` object proxy that webview2 provides and then awaits it, so the behavior is as before. Images now load for me.
Screen Shot 2022-11-04 at 10 48 49 AM

FYI @QilongTang @hwahlstrom 

### Declarations

### Release Notes

### Reviewers

### FYIs

### DIFF
--- a/src/LibraryViewExtensionWebView2/web/library/library.html
+++ b/src/LibraryViewExtensionWebView2/web/library/library.html
@@ -104,7 +104,7 @@
         LIBPLACEHOLDER
     </script>
 
-<script>
+    <script>
         var replaceImageDelayTime = 100;
         var currentImage;
         let intervalId;
--- a/src/LibraryViewExtensionWebView2/web/library/library.html
+++ b/src/LibraryViewExtensionWebView2/web/library/library.html
@@ -104,7 +104,7 @@
         LIBPLACEHOLDER
     </script>
 
-    <script>
+<script>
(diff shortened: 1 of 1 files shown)


src/LibraryViewExtensionWebView2/web/library/library.html
```diff
@@ -104,7 +104,7 @@
         LIBPLACEHOLDER
     </script>
 
-    <script>
+<script>
```
QilongTang, line 107: Is this misplaced?

QilongTang (approved):
One comment then LGTM and tested

PR #13473, opened 2022-11-04
Dyn 5358 friendly splash screen

filipeotero:
### Purpose

This PR includes the following improvements for the splash screen:

* Close window before launching Dynamo
* Increase the size of the window

### Declarations

### Reviewers

@QilongTang @aabi

### FYIs

@RobertGlobant20 @jesusalvino


src/DynamoCoreWpf/Views/SplashScreen/SplashScreen.xaml.cs
```diff
@@ -89,6 +92,7 @@ internal async void SetLoadingDone()
             loadingTimer.Stop();
             loadingTimer = null;
             await webView.CoreWebView2.ExecuteScriptAsync($"window.setLoadingDone()");
+            await webView.CoreWebView2.ExecuteScriptAsync($"window.setTotalLoadingTime('Total loading time: {totalLoadingTime}ms')");
```
QilongTang, line 95: Do we need to localize this?

filipeotero: Yes, I will create a resource for it. Thanks!

QilongTang (approved):

src/DynamoSandbox/DynamoCoreSetup.cs
```diff
@@ -67,7 +67,7 @@ public void RunApplication(Application app)
 
                 DynamoModel.RequestUpdateLoadBarStatus += DynamoModel_RequestUpdateLoadBarStatus;
 
-                splashScreen = new Dynamo.UI.Views.SplashScreen();
+                splashScreen = new Dynamo.UI.Views.SplashScreen(Program.DynamoCorePath);
```
QilongTang, line 70: This does not seem the right path? FYI: @RobertGlobant20

RobertGlobant20: @filipeotero the right path should be something like this:
C:\Users\myuser\AppData\Roaming\Dynamo\Dynamo Core\2.17\
Please verify if this is the expected path.

filipeotero: Aaron is right, the path is wrong. But to get the user path, we need to load Dynamo model first. Is it ok @QilongTang?

QilongTang: Hi @filipeotero If we load Dynamo Model first, then a few steps from the dynamic splash screen will be done by the time we display it. For now, would you hard code the path in CoreSetup?

filipeotero: now its like this: `C:\Users\myuser\AppData\Roaming\Dynamo\Dynamo Core\2.17`

PR #13481, opened 2022-11-07
AGD-2808 : add data folder option for cli

BogdanZavu:
### Purpose
Allow the DynamoCLI to work with specific userData and commonData folder.
We need more control over these locations for a few reasons :
- we want to be able to use a custom location ( in a way similar to the way D4R does it )
- for our usage ( in parallel with D4R ) we want to avoid using common packages location to avoid potential conflicts or execution mismatch when running the same graph in D4R.

### Declarations

### Reviewers
@mjkkirschner @saintentropy @QilongTang

### DIFF
--- a/src/DynamoApplications/PathResolvers.cs
+++ b/src/DynamoApplications/PathResolvers.cs
@@ -1,4 +1,4 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 using System.IO;
 using Dynamo.Interfaces;
 
@@ -73,7 +73,7 @@ namespace Dynamo.Applications
         private readonly List<string> additionalNodeDirectories;
         private readonly List<string> preloadedLibraryPaths;
 
-        public CLIPathResolver(string preloaderLocation)
+        public CLIPathResolver(string preloaderLocation, string userDataFolder, string commonDataFolder)
         {
             // If a suitable preloader cannot be found on the system, then do 
             // not add invalid path into additional resolution. The default 
@@ -103,6 +103,8 @@ namespace Dynamo.Applications
                 "GeometryColor.dll"
             };
 
+            UserDataRootFolder = userDataFolder;
+            CommonDataRootFolder = commonDataFolder;
         }
 
         public IEnumerable<string> AdditionalResolutionPaths
@@ -120,14 +122,8 @@ namespace Dynamo.Applications
             get { return preloadedLibraryPaths; }
         }
 
-        public string UserDataRootFolder
-        {
-            get { return string.Empty; }
-        }
+        public string UserDataRootFolder { get; private set; }
 
-        public string CommonDataRootFolder
-        {
-            get { return string.Empty; }
-        }
+        public string CommonDataRootFolder { get; private set; }
     }
 }
--- a/src/DynamoApplications/StartupUtils.cs
+++ b/src/DynamoApplications/StartupUtils.cs
@@ -262,7 +262,7 @@ namespace Dynamo.Applications
         public static DynamoModel MakeModel(bool CLImode, string asmPath = "", string hostName ="")
         {
             var isASMloaded = PreloadASM(asmPath, out string geometryFactoryPath, out string preloaderLocation);
-            var model = StartDynamoWithDefaultConfig(CLImode, "", "", geometryFactoryPath, preloaderLocation, new HostAnalyticsInfo() { HostName = hostName });
+            var model = StartDynamoWithDefaultConfig(CLImode, string.Empty, string.Empty, geometryFactoryPath, preloaderLocation, new HostAnalyticsInfo() { HostName = hostName });
             model.IsASMLoaded = isASMloaded;
             return model;
         }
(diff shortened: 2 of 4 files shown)


src/DynamoApplications/StartupUtils.cs
```diff
@@ -133,6 +135,8 @@ public static CommandLineArguments Parse(string[] args)
                 .Add("gp=|GP=|geometrypath=|GeometryPath=", "relative or absolute path to a directory containing ASM. When supplied, instead of searching the hard disk for ASM, it will be loaded directly from this path.", gp => asmPath = gp)
                 .Add("k|K|keepalive", "Keepalive mode, leave the Dynamo process running until a loaded extension shuts it down.", k => keepAlive = k != null)
                 .Add("nc|NC|noconsolecli", "Don't rely on the console window to interact with CLI in Keepalive mode", nc => noConsoleCli = nc != null)
+                .Add("ud=|UD=|userdatacli", "Specify user data folder to be used by PathResolver with CLI", ud => userDataFolderCli = ud)
+                .Add("cd=|CD=|commondatacli", "Specify common data folder to be used by PathResolver with CLI", cd => commonDataFolderCli = cd)
```
pinzart90, line 139: do we need to complicate these options with the suffix `cli` ?

src/DynamoApplications/PathResolvers.cs
```diff
@@ -73,7 +73,7 @@ internal class CLIPathResolver : IPathResolver
         private readonly List<string> additionalNodeDirectories;
         private readonly List<string> preloadedLibraryPaths;
 
-        public CLIPathResolver(string preloaderLocation)
+        public CLIPathResolver(string preloaderLocation, string userDataFolder, string commonDataFolder)
```
mjkkirschner, line 76: @BogdanZavu you cannot break a public signature like this in DynamoCore.

src/DynamoApplications/StartupUtils.cs
```diff
@@ -227,6 +235,23 @@ private static IUpdateManager InitializeUpdateManager()
             return um;
         }
 
+        /// <summary>
+        /// Use this overload to construct a DynamoModel in CLI context when the location of ASM to use is known and host analytics info is known.
```
mjkkirschner, line 239: ...and you want to set data paths.

src/DynamoApplications/StartupUtils.cs
```diff
@@ -227,6 +235,23 @@ private static IUpdateManager InitializeUpdateManager()
...
+        /// </summary>
+        /// <param name="asmPath">Path to directory containing geometry library binaries</param>
+        /// <param name="userDataFolder">Path to be used by PathResolver for UserDataFolder</param>
+        /// <param name="commonDataFolder">Path to be used by PathResolver for CommonDataFolder</param>        
+        /// <returns></returns>
+        public static DynamoModel MakeCLIModel(string asmPath, string userDataFolder, string commonDataFolder, HostAnalyticsInfo info = new HostAnalyticsInfo())
+        {
+            // Preload ASM and display corresponding message on splash screen
+            DynamoModel.OnRequestUpdateLoadBarStatus(new SplashScreenLoadEventArgs(Resources.SplashScreenPreLoadingAsm, 10));
+            var isASMloaded = PreloadASM(asmPath, out string geometryFactoryPath, out string preloaderLocation);
```
mjkkirschner, line 249: any reason this can't call the other makeModel overload?

src/DynamoWPFCLI/Program.cs
```diff
@@ -83,14 +83,10 @@ internal static void Main(string[] args)
...
             DynamoModel model;
-            if (!String.IsNullOrEmpty(cmdLineArgs.ASMPath))
-            {
-                model = Dynamo.Applications.StartupUtils.MakeModel(true, cmdLineArgs.ASMPath, cmdLineArgs.AnalyticsInfo);
-            }
-            else
-            {
-                model = Dynamo.Applications.StartupUtils.MakeModel(true, string.Empty, cmdLineArgs.AnalyticsInfo);
-            }
+            model = Dynamo.Applications.StartupUtils.MakeCLIModel(String.IsNullOrEmpty(cmdLineArgs.ASMPath) ? string.Empty : cmdLineArgs.ASMPath,
```
mjkkirschner, line 86: have you verified these changes don't try to force ASM to load from an empty path string? That would be bad.

mjkkirschner (changes requested):

src/DynamoApplications/PathResolvers.cs
BogdanZavu: @mjkkirschner this class is internal so there shouldn't be any breakage.
I taught it's simpler to keep just 1 constructor but lmk if you still prefer to have another one

mjkkirschner: thanks @BogdanZavu! I overlooked that and just saw the method modifier - 1 constructor is fine with me.

mjkkirschner (approved):
hey @BogdanZavu  https://github.com/DynamoDS/Dynamo/pull/13481 looks good to me except I'm a bit concerned about the change to pass an empty string to the ASM path - can you verify that the CLI will still try to preload ASM by searching the machine in that case - I'm not sure we have tests for that specifically.

src/DynamoWPFCLI/Program.cs
BogdanZavu: Works as expected, nothing changes here compared to previous behavior

src/DynamoApplications/StartupUtils.cs
```diff
@@ -262,7 +262,7 @@ public static DynamoModel MakeCLIModel(string asmPath, string userDataFolder, st
         public static DynamoModel MakeModel(bool CLImode, string asmPath = "", string hostName ="")
         {
             var isASMloaded = PreloadASM(asmPath, out string geometryFactoryPath, out string preloaderLocation);
-            var model = StartDynamoWithDefaultConfig(CLImode, geometryFactoryPath, preloaderLocation, new HostAnalyticsInfo() { HostName = hostName });
+            var model = StartDynamoWithDefaultConfig(CLImode, "", "", geometryFactoryPath, preloaderLocation, new HostAnalyticsInfo() { HostName = hostName });
```
mjkkirschner, line 265: can we use string.empty in these spots?

src/DynamoApplications/StartupUtils.cs
BogdanZavu: done

mjkkirschner:
LGTM!

PR #13491, opened 2022-11-09
move to a single function with no globals

mjkkirschner:
### Purpose

continues trying to improve icon loading in librariejs - I think this is a good improvement,  we no longer split the method into two halves, and instead now we can use the new webView2 APIs to actually get return values from the c# functions - use this directly and avoid having globally scoped variables.

### Declarations

### Release Notes

### Reviewers

### FYIs

### DIFF
--- a/src/LibraryViewExtensionWebView2/ScriptingObject.cs
+++ b/src/LibraryViewExtensionWebView2/ScriptingObject.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


src/LibraryViewExtensionWebView2/ScriptingObject.cs
```diff
@@ -24,21 +24,21 @@ public ScriptingObject(LibraryViewController controller)
...
-        public async void GetBase64StringFromPath(string iconurl)
+        public string GetBase64StringFromPath(string iconurl)
         {
 
             string ext;
             var iconAsBase64 = controller.iconProvider.GetResourceAsString(iconurl, out ext);
             if (string.IsNullOrEmpty(iconAsBase64))
             {
-                await LibraryViewController.ExecuteScriptFunctionAsync(controller.browser, "completeReplaceImages", "","");
+                return "";
```
RobertGlobant20, line 34: instead of "" can be string.empty?

src/LibraryViewExtensionWebView2/web/library/library.html
```diff
@@ -155,16 +154,13 @@
                      continue;
                 }
                 //request the icon from the extension.
-                await window.chrome.webview.hostObjects.bridgeTwoWay.GetBase64StringFromPath(src).sync();
+                var base64String = await window.chrome.webview.hostObjects.bridgeTwoWay.GetBase64StringFromPath(src);
```
RobertGlobant20, line 157: if GetBase64StringFromPath() method is not async, should we still use await?

RobertGlobant20 (approved):
LGTM with some comments

src/LibraryViewExtensionWebView2/web/library/library.html
mjkkirschner: my understanding from reading this page:
https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.addhostobjecttoscript?view=webview2-dotnet-1.0.1418.22

is that even if the method is not async - the proxy wrapper that webView2 generates to access it returns a promise and needs to be awaited or resolved using `then()` - do you understand these docs differently?

RobertGlobant20: yes, you are right when using AddHostObjectToScript you have to wait a promise. Thanks

QilongTang (approved):
Looks good on my computer, all icons are now displayed correctly

PR #13496, opened 2022-11-10
Integrating ML node autocomplete. 

reddyashish:
### Purpose

Integrating ML node autocomplete. 

Task: https://jira.autodesk.com/browse/DYN-5335

ML node autocomplete

**To-do changes**
1. Using temporary token, change it to use oAuth sign-in.
2. Add tests
3. Some UI improvements.

### Declarations

### Release Notes
Integrating ML node autocomplete. 

### Reviewers

@QilongTang @zeusongit 

### FYIs

### DIFF
--- a/src/DynamoCore/Search/SearchElements/MLNodeSearchElement.cs
+++ b/src/DynamoCore/Search/SearchElements/MLNodeSearchElement.cs
@@ -14,6 +14,7 @@ namespace Dynamo.Search.SearchElements
         {
             Node = new NodeRequest();
             Port = new PortRequest();
+            Host = new HostRequest();
             Context = new ContextRequest();
             Packages = new List<PackageItem>();
         }
--- a/src/DynamoCore/Search/SearchElements/MLNodeSearchElement.cs
+++ b/src/DynamoCore/Search/SearchElements/MLNodeSearchElement.cs
@@ -1,0 +1,208 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.Serialization;
+using Dynamo.Configuration;
+using Dynamo.Graph.Nodes;
+
+namespace Dynamo.Search.SearchElements
+{
+    [DataContract]
+    internal class MLNodeAutoCompletionRequest
+    {
+        internal MLNodeAutoCompletionRequest()
+        {
+            Node = new NodeRequest();
+            Port = new PortRequest();
+            Context = new ContextRequest();
+            Packages = new List<PackageItem>();
+        }
+
+        [DataMember(Name = "node")]
+        internal NodeRequest Node { get; set; }
+
+        [DataMember(Name = "port")]
+        internal PortRequest Port { get; set; }
+
+        [DataMember(Name = "host")]
+        internal HostRequest Host { get; set; }
+
+        [DataMember(Name = "dynamoVersion")]
+        internal string DynamoVersion { get; set; }
+
+        [DataMember(Name = "numberOfResults")]
+        internal int NumberOfResults { get; set; }
+
+        [DataMember(Name = "packages")]
+        internal IEnumerable<PackageItem> Packages { get; set; }
+
+        [DataMember(Name = "context")]
+        internal ContextRequest Context { get; set; }
+    }
+
+    [DataContract]
+    internal class NodeRequest
+    {
+        internal NodeRequest()
+        {
+            Type = new NodeTypeRequest();
+        }
+
+        [DataMember(Name = "id")]
+        internal string Id { get; set; }
+
+        [DataMember(Name = "type")]
+        internal NodeTypeRequest Type { get; set; }
+
+        [DataMember(Name = "lacing")]
+        internal string Lacing { get; set; }
+    }
+
+    [DataContract]
+    internal class NodeTypeRequest
+    {
+        [DataMember(Name = "id")]
+        internal string Id { get; set; }
+    }
+
+    [DataContract]
+    internal class PortRequest
+    {
+        [DataMember(Name = "name")]
+        internal string Name { get; set; }
+
+        [DataMember(Name = "direction")]
+        internal string Direction { get; set; }
+
+        [DataMember(Name = "listAtLevel")]
+        internal int ListAtLevel { get; set; }
+
+        [DataMember(Name = "keepListStructure")]
+        internal string KeepListStructure { get; set; }
+    }
+
+    [DataContract]
+    internal class HostRequest
+    {
+        [DataMember(Name = "name")]
+        internal string Name { get; set; }
+
+        [DataMember(Name = "version")]
+        internal string Version { get; set; }
+    }
+
+    [DataContract]
+    internal class ContextRequest
+    {
+        internal ContextRequest()
+        {
+            Connections = new List<ConnectionsRequest>();
+            Nodes = new List<NodeRequest>();
+        }
+
+        [DataMember(Name = "nodes")]
+        internal IEnumerable<NodeRequest> Nodes { get; set; }
+
+        [DataMember(Name = "connections")]
+        internal IEnumerable<ConnectionsRequest> Connections { get; set; }
+    }
+
+    [DataContract]
+    internal class ConnectionsRequest
+    {
+        [DataMember(Name = "from")]
+        internal ConnectorNodeItem StartNode { get; set; }
+
+        [DataMember(Name = "to")]
+        internal ConnectorNodeItem EndNode { get; set; }
+    }
+
+    [DataContract]
+    internal class ConnectorNodeItem
+    {
+        internal ConnectorNodeItem(string nodeId, string portName)
+        {
+            NodeId = nodeId;
+            PortName = portName;
+        }
+
+        [DataMember(Name = "nodeId")]
+        internal string NodeId { get; set; }
+
+        [DataMember(Name = "portName")]
+        internal string PortName { get; set; }
+    }
+
+    [DataContract]
+    internal class PackageItem
+    {
+        internal PackageItem(string name, string version)
+        {
+            Name = name;
+            Version = version;
+        }
+
+        [DataMember(Name = "name")]
+        internal string Name { get; set; }
+
+        [DataMember(Name = "version")]
+        internal string Version { get; set; }
+    }
+
+    [DataContract]
+    internal class MLNodeAutoCompletionResponse
+    {
+        internal MLNodeAutoCompletionResponse()
+        {
+            Results = new List<ResultItem>();
+        }
+
+        [DataMember(Name = "version")]
+        internal string Version { get; set; }
+
+        [DataMember(Name = "results")]
+        internal IEnumerable<ResultItem> Results { get; set; }
+    }
+
+    [DataContract]
+    internal class ResultItem
+    {
+        [DataMember(Name = "node")]
+        internal NodeResponse Node { get; set; }
+
+        [DataMember(Name = "port")]
+        internal PortResponse Port { get; set; }
+
+        [DataMember(Name = "score")]
+        internal double Score { get; set; }
+    }
+
+    [DataContract]
+    internal class PortResponse
+    {
+        [DataMember(Name = "name")]
+        internal string Name { get; set; }
+    }
+
+    [DataContract]
+    internal class NodeResponse
+    {
+        internal NodeResponse()
+        {
+            Type = new NodeTypeResponse();
+        }
+
+        [DataMember(Name = "type")]
+        internal NodeTypeResponse Type { get; set; }
+    }
+
+    [DataContract]
+    internal class NodeTypeResponse
+    {
+        [DataMember(Name = "nodeType")]
+        internal string NodeType { get; set; }
+
+        [DataMember(Name = "id")]
+        internal string Id { get; set; }
+    }
+}
--- a/src/DynamoCore/Search/SearchElements/NodeSearchElement.cs
+++ b/src/DynamoCore/Search/SearchElements/NodeSearchElement.cs
@@ -309,7 +309,7 @@ namespace Dynamo.Search.SearchElements
         /// <summary>
         /// Rating of confidence
         /// </summary>
-        public int ConfidenceScore { get; set; }
+        public double ConfidenceScore { get; set; }
         /// <summary>
         /// Indicates if the Node is part of ML result per use
         /// </summary>
--- a/src/DynamoCoreWpf/App.config
+++ b/src/DynamoCoreWpf/App.config
@@ -1,5 +1,9 @@
 ﻿<?xml version="1.0"?>
 <configuration>
+  <appSettings>
+    <add key="MLNodeAutocomplete" value="https://dev.autocomplete.dynamobim.org/autocomplete"/>
+	<add key="token" value="eyJhbGciOiJSUzI1NiIsImtpZCI6IlU3c0dGRldUTzlBekNhSzBqZURRM2dQZXBURVdWN2VhIn0.eyJzY29wZSI6WyJkYXRhOnJlYWQiXSwiY2xpZW50X2lkIjoiN3RpUXl2cWtHdmN5M3BYY0NBTW1LbTlldjFHR0FuZGoiLCJhdWQiOiJodHRwczovL2F1dG9kZXNrLmNvbS9hdWQvYWp3dGV4cDYwIiwianRpIjoiQTBrSjZGem8zNTVQTlFtY0hPNFdUYXhleG9ZMDZVM3YzeWZsU25UMkI0OFU3enpSZ3RJczZ1aTQ5UEZZWXRaRiIsImV4cCI6MTY2ODAwNDA3Nn0.Q_yhMnWR9IyQaKgbDyfEXQgRjDlc57MMVLAhWi-5_n905XxGdzqR5KwZbSVirfrJQ70R-9XaaXbOLIdTZmMGcQlU4D8U3anMsg8O7tXns1HmZ3lRTDQSXW_KpSBw9AWA2CUFOogKmoCa_xz0PMw-hQWoX6Uw_ER_wB9UrAR5P_A_fwhkLhAgCtbR2JrQmiuJ-Pl3YxXTCKtW7y7t_trImgo98k2BnsyI20jiVn6bEuVZJcZEWJGqtgawDECzI67lg4ivUE2m22B2lxpljmLq5sCphOu4LueNbmuZunbieDrFxTvWOQqPNi5-733Y-PIfOWoT5GJ0TG_Uy9mMVKBwcA"/>
+  </appSettings>
   <startup>
     <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.8"/>
   </startup>
--- a/src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml
+++ b/src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml
@@ -162,7 +162,7 @@
                         </TextBlock>
                     </StackPanel>
              </StackPanel>
-             <StackPanel x:Name="LowConfidenceSpace" Orientation="Horizontal" VerticalAlignment="Center" HorizontalAlignment="Left" Margin="0,10,0,10"
+             <StackPanel x:Name="LowConfidenceSpace" MouseLeftButtonUp="ShowLowConfidenceResults" Orientation="Horizontal" VerticalAlignment="Center" HorizontalAlignment="Left" Margin="0,10,0,10"
                          Visibility="{Binding Path=DisplayLowConfidence, Converter={StaticResource BooleanToVisibilityConverter}}">
                 <Image HorizontalAlignment="Center"
                      Height="14"     
--- a/src/DynamoCoreWpf/UI/Converters.cs
+++ b/src/DynamoCoreWpf/UI/Converters.cs
@@ -15,7 +15,7 @@ using Dynamo.Graph.Nodes;
 using Dynamo.Graph.Workspaces;
 using Dynamo.Logging;
 using Dynamo.PackageManager;
-using Dynamo.Search.SearchElements;
+using SearchElements = Dynamo.Search.SearchElements;
 using Dynamo.UI;
 using Dynamo.UI.Controls;
 using Dynamo.Updates;
@@ -2722,30 +2722,30 @@ namespace Dynamo.Controls
     {
         public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
         {
-            var type = (ElementTypes)value;
+            var type = (SearchElements.ElementTypes)value;
 
             switch (type)
             {
-                case ElementTypes.Packaged:
+                case SearchElements.ElementTypes.Packaged:
                     return Resources.PackageTypeShortString;
 
-                case ElementTypes.Packaged | ElementTypes.ZeroTouch:
+                case SearchElements.ElementTypes.Packaged | SearchElements.ElementTypes.ZeroTouch:
                     return Resources.PackageTypeShortString;
 
-                case ElementTypes.Packaged | ElementTypes.CustomNode:
+                case SearchElements.ElementTypes.Packaged | SearchElements.ElementTypes.CustomNode:
                     return Resources.PackageTypeShortString;
 
-                case ElementTypes.Packaged | ElementTypes.ZeroTouch | ElementTypes.CustomNode:
+                case SearchElements.ElementTypes.Packaged | SearchElements.ElementTypes.ZeroTouch | SearchElements.ElementTypes.CustomNode:
                     return Resources.PackageTypeShortString;
 
-                case ElementTypes.ZeroTouch:
+                case SearchElements.ElementTypes.ZeroTouch:
                     return Resources.ZeroTouchTypeShortString;
 
-                case ElementTypes.CustomNode:
+                case SearchElements.ElementTypes.CustomNode:
                     return Resources.CustomNodeTypeShortString;
 
-                case ElementTypes.BuiltIn:
-                case ElementTypes.None:
+                case SearchElements.ElementTypes.BuiltIn:
+                case SearchElements.ElementTypes.None:
                 default:
                     return string.Empty;
             }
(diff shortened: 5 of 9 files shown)


QilongTang:
@reddyashish Do we get confidence as part of the response? Since from the gif, it is not there yet, just curious

src/DynamoCoreWpf/App.config
```diff
@@ -1,5 +1,9 @@
 ﻿<?xml version="1.0"?>
 <configuration>
+  <appSettings>
+    <add key="MLNodeAutocomplete" value="https://dev.autocomplete.dynamobim.org/autocomplete"/>
+	<add key="token" value="eyJhbGciOiJSUzI1NiIsImtpZCI6IlU3c0dGRldUTzlBekNhSzBqZURRM2dQZXBURVdWN2VhIn0.eyJzY29wZSI6WyJkYXRhOnJlYWQiXSwiY2xpZW50X2lkIjoiN3RpUXl2cWtHdmN5M3BYY0NBTW1LbTlldjFHR0FuZGoiLCJhdWQiOiJodHRwczovL2F1dG9kZXNrLmNvbS9hdWQvYWp3dGV4cDYwIiwianRpIjoiQTBrSjZGem8zNTVQTlFtY0hPNFdUYXhleG9ZMDZVM3YzeWZsU25UMkI0OFU3enpSZ3RJczZ1aTQ5UEZZWXRaRiIsImV4cCI6MTY2ODAwNDA3Nn0.Q_yhMnWR9IyQaKgbDyfEXQgRjDlc57MMVLAhWi-5_n905XxGdzqR5KwZbSVirfrJQ70R-9XaaXbOLIdTZmMGcQlU4D8U3anMsg8O7tXns1HmZ3lRTDQSXW_KpSBw9AWA2CUFOogKmoCa_xz0PMw-hQWoX6Uw_ER_wB9UrAR5P_A_fwhkLhAgCtbR2JrQmiuJ-Pl3YxXTCKtW7y7t_trImgo98k2BnsyI20jiVn6bEuVZJcZEWJGqtgawDECzI67lg4ivUE2m22B2lxpljmLq5sCphOu4LueNbmuZunbieDrFxTvWOQqPNi5-733Y-PIfOWoT5GJ0TG_Uy9mMVKBwcA"/>
```
zeusongit, line 5: I don't think we should be committing this, this token can be used, maybe use the new Idsdk API to fetch the token.
https://github.com/DynamoDS/Dynamo/blob/8baa17e786dcf506e18c495ec33f164a2c834c53/src/DynamoCore/Core/IDSDKManager.cs#L182

src/DynamoCoreWpf/UI/Converters.cs
```diff
@@ -15,7 +15,7 @@
 using Dynamo.Graph.Workspaces;
 using Dynamo.Logging;
 using Dynamo.PackageManager;
-using Dynamo.Search.SearchElements;
+using SearchElements = Dynamo.Search.SearchElements;
```
zeusongit, line 18: Any specific reason for this way of declaration?

src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml
```diff
@@ -162,7 +162,7 @@
                         </TextBlock>
                     </StackPanel>
              </StackPanel>
-             <StackPanel x:Name="LowConfidenceSpace" Orientation="Horizontal" VerticalAlignment="Center" HorizontalAlignment="Left" Margin="0,10,0,10"
+             <StackPanel x:Name="LowConfidenceSpace" MouseLeftButtonUp="ShowLowConfidenceResults" Orientation="Horizontal" VerticalAlignment="Center" HorizontalAlignment="Left" Margin="0,10,0,10"
```
zeusongit, line 165: Do we also have an option to revert this? That is, hide `Low Confidence` results after the user have expanded the list?

src/DynamoCoreWpf/App.config
reddyashish: Yes, I wil be removing this token and use the oAuth sign -in. Just saw your PR merged.

src/DynamoCoreWpf/Controls/NodeAutoCompleteSearchControl.xaml
reddyashish: Not in that window. When the node autocomplete window is opened again, it will be back to starting behaviour.

reddyashish:
@QilongTang Yes we do get the confidence score, I am adding a field to show that. Will push those changes.

src/DynamoCoreWpf/UI/Converters.cs
reddyashish: Oh, this was related to the other changes I made this file. Don't need it that explicitly. Will undo this change

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -155,67 +163,224 @@ private void InitializeDefaultAutoCompleteCandidates()
...
-            */
+            // Set context info
+            // Considering upstream nodes when the PortType is output and considering downstream nodes when the PortType is input. 
+            var connectors = nodeInfo.AllConnectors;
 
-            // Case 4: Confidence score are under a threshold, assuming the minimum value is 50 and all results nodes are under it
-            /*
-            FilteredResults = DefaultResults.Where(e => e.Name == "Watch 3D" || e.Name == "Python Script").ToList();
-            foreach (var item in FilteredResults)
+            foreach (var connector in connectors)
```
varvaratou, line 211: If I understand well, this would retrieve the fist level of neighbors of the current node as context, right?
It is not very clear in the schema description, and I may have confused you during our discussion - apologies for this. For context, we expect all nodes (and connections) that can be visited from the node that triggers the query (the whole connected component that includes the query node).

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -155,67 +163,224 @@ private void InitializeDefaultAutoCompleteCandidates()
...
+            if (MLresults == null || MLresults.Results.Count() == 0)
+            {
+                DisplayNoRecommendationsLowConfidence = true;
+                DisplayLowConfidence = false;
+                NoRecommendationsOrLowConfidenceTitle = Resources.AutocompleteNoRecommendationsTitle;
+                NoRecommendationsOrLowConfidenceMessage = Resources.AutocompleteNoRecommendationsMessage;
+                return;
+            }
+
+            // ML Results are categorized based on the threshold confidence score before displaying.
```
varvaratou, line 276: Currently the service returns no nodes coming from packages, but we want to enable package support in the next month. If the recommendation list contains a node from a package (that the user has loaded), would the search for the `NodeSearchElementViewModel` still work?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -26,6 +33,7 @@ public class NodeAutoCompleteSearchViewModel : SearchViewModel
         private string noRecommendationsOrLowConfidenceTitle;
         private bool displayNoRecommendationsLowConfidence;
         private bool displayLowConfidence;
+        private double confidenceThreshold = 0.5;
```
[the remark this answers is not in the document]
QilongTang: @varvaratou Which threshold are you referring to? This particular one is more for UI affordance, to hide results with less confidence etc, I would not expect end users to touch this often. I remember there is another threshold more related to service?

src/DynamoCore/Search/SearchElements/MLNodeSearchElement.cs
```diff
@@ -0,0 +1,209 @@
...
+        internal MLNodeAutoCompletionResponse()
+        {
+            Results = new List<ResultItem>();
+        }
+
+        [DataMember(Name = "version")]
+        internal string Version { get; set; }
+
+        [DataMember(Name = "results")]
+        internal IEnumerable<ResultItem> Results { get; set; }
```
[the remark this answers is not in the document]
reddyashish: This property is here: https://github.com/DynamoDS/Dynamo/pull/13496/files#diff-1e35319171f52504cbbdd1bc8238f0e3a9f443b5a87684d8c0754ed11a116263R35

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -155,67 +163,224 @@ private void InitializeDefaultAutoCompleteCandidates()
...
-            NoRecommendationsOrLowConfidenceTitle = Resources.AutocompleteNoRecommendationsTitle;
-            NoRecommendationsOrLowConfidenceMessage = Resources.AutocompleteNoRecommendationsMessage;
+            // Intialize request for the the ML API
+            MLNodeAutoCompletionRequest request = new MLNodeAutoCompletionRequest();
+
+            var nodeInfo = PortViewModel.PortModel.Owner;
+            var portInfo = PortViewModel.PortModel;
+
+            request.DynamoVersion = AssemblyHelper.GetDynamoVersion().ToString();
+            request.NumberOfResults = 10;
```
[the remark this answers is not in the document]
reddyashish: We were thinking to set to to 5 by default. I was using 10 to test some cases. 
@QilongTang Do you want to go with 5 or 10 as default?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -155,67 +165,252 @@ private void InitializeDefaultAutoCompleteCandidates()
...
+        {
+            MLNodeAutoCompletionResponse results = null;
+            var uri = DynamoUtilities.PathHelper.getServiceBackendAddress(this, "MLNodeAutocomplete");
+            var token = DynamoUtilities.PathHelper.getServiceConfigValues(this, "token");
+            var authProvider = dynamoViewModel.Model.AuthenticationManager.AuthProvider;
+
+            if (authProvider is IDSDKManager manager)
+            {
+                /*
+                var client = new RestClient(uri);
```
reddyashish, line 358: Getting 403 forbidden error when using the RestSharp API. Will undo this code, once that is unblocked.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
reddyashish: yes as long as the node info is found in the VM, it should work.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
reddyashish: Yes currently, it is sending only the first level of neighbors. I will look into adding the remaining information.

src/DynamoCore/Search/SearchElements/MLNodeSearchElement.cs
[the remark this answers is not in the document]
varvaratou: I was referring to the one the service sends back with the response, but I saw later in the code that you do not make any use of it.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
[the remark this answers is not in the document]
varvaratou: I was referring to the threshold for the UI affordance - I was thinking more in terms of whether we have the flexibility to test it and see if this number makes sense and change if not. It may be a bit strict. I am thinking for example a case where the service could return 5 recommendations with scores 0.4, 0.2, 0.2, 0.1, 0.1. In this example, none would fall under the high confidence score but there would still be the top recommendations. 

Just an idea, maybe an alternative approach could be including the nodes whose confidence score sum up to a particular threshold.

src/DynamoCore/Search/SearchElements/MLNodeSearchElement.cs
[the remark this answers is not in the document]
reddyashish: Added this property to response.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -155,67 +163,224 @@ private void InitializeDefaultAutoCompleteCandidates()
...
             }
-            DisplayNoRecommendationsLowConfidence = !FilteredResults.Where(n => n.Model.AutoCompletionNodeMachineLearningInfo.IsByRecommendation).Any();
-            DisplayLowConfidence = false;
-            NoRecommendationsOrLowConfidenceTitle = Resources.AutocompleteNoRecommendationsTitle;
-            NoRecommendationsOrLowConfidenceMessage = Resources.AutocompleteNoRecommendationsMessage;
-            */
 
+            // Set port info
+            request.Port.Direction = portInfo.PortType == PortType.Input ? PortType.Input.ToString().ToLower() : PortType.Output.ToString().ToLower();
+            request.Port.Name = portInfo.Name;
```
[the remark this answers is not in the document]
reddyashish: Nice catch. So just to to clear for the variable input nodes, should the port-name be empty only for the ports that are additionally added by user or for all the ports of that node? This is because the inbuilt ports for those nodes can be of different types but the additional ones that can be added will be of same type.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -155,67 +163,224 @@ private void InitializeDefaultAutoCompleteCandidates()
...
-            FilteredResults = DefaultResults.Where(e => e.Name == "Watch 3D" || e.Name == "Python Script").ToList();
+            // Set host info
+            request.Host.Name = dynamoViewModel.Model.HostAnalyticsInfo.HostName;
+            request.Host.Version = dynamoViewModel.Model.HostVersion;
 
-            foreach (var item in FilteredResults)
+            // Set packages info
+            var packageManager = dynamoViewModel.Model.ExtensionManager.Extensions.OfType<PackageManagerExtension>().FirstOrDefault();
+
+            foreach (var pkg in packageManager.PackageLoader.LocalPackages)
```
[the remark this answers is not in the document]
reddyashish: Yes the request will only contain the packages that were successfully loaded into the VM.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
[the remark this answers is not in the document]
mjkkirschner: @varvaratou @reddyashish @QilongTang  it would be kind of interesting to consider pulling this value from launch darkly or a config, then you could roll out different confidence thresholds to % of users or change it whenever you want without a new release...

src/DynamoCore/Search/SearchElements/MLNodeSearchElement.cs
```diff
@@ -0,0 +1,212 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.Serialization;
+using Dynamo.Configuration;
+using Dynamo.Graph.Nodes;
+
+namespace Dynamo.Search.SearchElements
+{
+    [DataContract]
+    internal class MLNodeAutoCompletionRequest
```
mjkkirschner, line 11: seems like you might want a constructor...

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -155,98 +164,296 @@ private void InitializeDefaultAutoCompleteCandidates()
...
+
+            // Set port info
+            // If the node is a Variable-input nodemodel or zero-touch node, then set the port name as empty.
+            request.Port.Name = (nodeInfo is VariableInputNode || nodeInfo is DSVarArgFunction) ? string.Empty : portInfo.Name;
+            request.Port.Direction = portInfo.PortType == PortType.Input ? PortType.Input.ToString().ToLower() : PortType.Output.ToString().ToLower();
+            request.Port.KeepListStructure = portInfo.KeepListStructure.ToString();
+            request.Port.ListAtLevel = portInfo.Level;
 
+            // Set host info
+            request.Host.Name = dynamoViewModel.Model.HostAnalyticsInfo.HostName;
```
mjkkirschner, line 199: I think you need to check if this is null and fallback to `dynamoViewModel.Model.HostName` or you need to make `HostAnalyticsInfo .HostName` fallback itself. (this causes issues for CER as well)

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -155,98 +164,296 @@ private void InitializeDefaultAutoCompleteCandidates()
...
+            request.Host.Version = dynamoViewModel.Model.HostVersion;
 
-            // Case 3: Confidence score are under a threshold, assuming the minimum value is 50 and some result nodes are under it
-            /*
-            FilteredResults = DefaultResults.Where(e => e.Name == "Watch 3D" || e.Name == "Python Script").ToList();
+            // Set packages info
+            var packageManager = dynamoViewModel.Model.ExtensionManager.Extensions.OfType<PackageManagerExtension>().FirstOrDefault();
 
-            foreach (var item in FilteredResults)
+            foreach (var pkg in packageManager.PackageLoader.LocalPackages)
```
mjkkirschner, line 205: `packageManager` might be null, please check.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
mjkkirschner: there should be some code for this floating around for sure.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -155,98 +164,296 @@ private void InitializeDefaultAutoCompleteCandidates()
...
+            string jsonRequest = JsonConvert.SerializeObject(request);
+            MLNodeAutoCompletionResponse MLresults = null;
+
+            // Get results from the ML API.
+            try
             {
-                FilteredResults = new List<NodeSearchElementViewModel>();
+                MLresults = GetMLNodeAutocompleteResults(jsonRequest);
+            }
+            catch(Exception e) {
```
mjkkirschner, line 256: maybe also log e incase it's something unexpected ?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -155,98 +164,296 @@ private void InitializeDefaultAutoCompleteCandidates()
...
+
+            // ML Results are categorized based on the threshold confidence score before displaying. 
+            if (MLresults.Results.Count() > 0)
+            {
+                FilteredHighConfidenceResults = new List<NodeSearchElementViewModel>();
+                FilteredLowConfidenceResults = new List<NodeSearchElementViewModel>();
+
+                foreach (var item in MLresults.Results)
+                {
+                    if (item.Node.Type.NodeType.Equals("FunctionNode"))
```
mjkkirschner, line 287: nameof?

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -155,98 +164,296 @@ private void InitializeDefaultAutoCompleteCandidates()
...
+                            element.AutoCompletionNodeMachineLearningInfo.ConfidenceScore = item.Score;
+                            element.AutoCompletionNodeMachineLearningInfo.ViewConfidenceScoreRecentUse = true;
+                            element.AutoCompletionNodeMachineLearningInfo.IsByRecommendation = true;
+                            matchingResults.Add(element);
+                        }
+                    }
+                    else
+                    {
+                        // Retreive assembly name and full name from type id.
+                        var typeInfo = item.Node.Type.Id.Split(',');
```
mjkkirschner, line 301: I think it would be good to use this constructor and avoid parsing it yourself.
https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assemblyname.-ctor?view=net-7.0#system-reflection-assemblyname-ctor(system-string)

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
```diff
@@ -155,98 +164,296 @@ private void InitializeDefaultAutoCompleteCandidates()
 
         internal void DisplayMachineLearningResults()
         {
-            // Case 1: no results (0 items)
```
mjkkirschner, line 158: @reddyashish this function is gigantic...

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
reddyashish: done.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
reddyashish: The type id we are sending has a different format than the Assembly path and was getting a exception while loading with that name format. I created a class to make the parsing robust.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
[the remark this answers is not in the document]
reddyashish: After discussing with Aaron, going with 10 for now.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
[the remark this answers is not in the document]
reddyashish: Yes, we have an upcoming task to add this control in the preferences panel. Will add this information to that task.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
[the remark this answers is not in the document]
reddyashish: After discussing with Martin and Varvara, we will be parsing the port name to remove the digits at the end.

src/DynamoCore/Search/SearchElements/NodeSearchElement.cs
```diff
@@ -21,10 +21,6 @@ public abstract class NodeSearchElement : ISearchEntry, ISource<NodeModel>
         private bool isVisibleInSearch = true;
 
         internal AutoCompletionNodeElementInfo AutoCompletionNodeElementInfo { get; set; } = new AutoCompletionNodeElementInfo();
-        /// <summary>
-        /// Machine Learning related info
-        /// </summary>
-        public AutoCompletionNodeMachineLearningInfo AutoCompletionNodeMachineLearningInfo { get; set; } = new AutoCompletionNodeMachineLearningInfo();
```
reddyashish, line 27: Moved this property to NodeSearchElementViewModel as setting this info on NodeSearchElement will interfere with the regular search results. This public property was added after 2.16 and should not be a breaking change.

src/DynamoCore/Search/SearchDictionary.cs
```diff
@@ -305,11 +305,11 @@ private static bool MatchWithQueryString(string key, string[] subPatterns)
             foreach (var searchElement in searchDict)
             {
                 var currentElement = (searchElement.Key as SearchElements.NodeSearchElement);
-                var currentElementName = currentElement.FullName;
+                var currentElementName = currentElement.CreationName;
                 foreach (var ele in subset)
                 {
                     //if any element in tagDictionary matches to any element in subset, return true
-                    if (currentElementName.IndexOf(ele.FullName, StringComparison.OrdinalIgnoreCase) != -1)
+                    if (currentElementName.Equals(ele.CreationName, StringComparison.OrdinalIgnoreCase))
```
reddyashish, line 312: Checking for the exact CreationName of Node to match from the subset results.

src/DynamoCoreWpf/ViewModels/Search/SearchViewModel.cs
```diff
@@ -46,7 +46,12 @@ public void OnSearchTextChanged(object sender, EventArgs e)
             if (SearchTextChanged != null)
                 SearchTextChanged(this, e);
 
-            SearchCommand?.Execute(null);
+            // If this is trigged from NodeAutoCompleteSearchViewModel, we don't want to execute this SearchCommand on all nodes.
+            // The Search will later be triggered on a subset of filtered results.
+            if (this is NodeAutoCompleteSearchViewModel is false)
```
reddyashish, line 51: @QilongTang This was triggering the search command twice that's why we were seeing the results change during the window loading. This is because NodeAutoCompleteSearchViewModel is leveraging the SearchViewModel along with the search text bar. To avoid this, we will not trigger the search on all the nodes in library from NodeAutoCompleteSearchViewModel. Instead we will later search the filtered items based on ML or object type.

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
reddyashish: Made changes to the Context property which will have all the information regarding the connected component

QilongTang (approved):

src/DynamoCoreWpf/ViewModels/Search/NodeAutoCompleteSearchViewModel.cs
reddyashish: Changed this to use the RestSharp API and setting the IDSDK token in header if the user is logged in.

QilongTang:
@reddyashish The job reported failed but I dont see any regressions? Can you double check

reddyashish:
Failed while getting the package: Error downloading 'adp-desktop-sdk_win_release_intel64_v140.4.2.0'
Running it again.

reddyashish:
Job passed. Merging this.

PR #13497, opened 2022-11-10
Enable Authentication in Dynamo Sandbox with the new NuGet package

zeusongit:
### Purpose

Uncomment IDSDK APIs and include the IDSDK nuget package.

ezgif com-gif-maker (2)

### Declarations

### Reviewers

@DynamoDS/dynamo


src/DynamoApplications/StartupUtils.cs
```diff
@@ -377,7 +377,7 @@ private static bool PreloadASM(string asmPath, out string geometryFactoryPath, o
                 HostAnalyticsInfo = info,
                 CLIMode = CLImode
             };
-            config.AuthProvider = CLImode ? null : new Core.IDSDKManager();
+            config.AuthProvider = CLImode || !string.IsNullOrEmpty(info.HostName) ? null : new Core.IDSDKManager();
```
QilongTang, line 380: Host info can be passed by CLI or as param in out of process integrations so technically it may not be null all the time, I feel we will need more testing on this. Would you revert this for now, that is more conservative I think

QilongTang (approved):
LGTM with one comment

src/DynamoApplications/StartupUtils.cs
QilongTang: I'm not 100% sure what would be the best check here, maybe combine this with some other check, e.g. process name?

zeusongit:
Before merging this can you take a look at https://github.com/DynamoDS/SplashScreen/pull/19
Would like to get that in before merging this.

PR #13500, opened 2022-11-10
DYN-5326-Import-PreferencesSettings

RobertGlobant20:
### Purpose

Load Preferences xml file located in the ProgramData/Dynamo path.
I added some code that at dynamo startup will check for xml configuration files in the next path: C:\ProgramData\Dynamo, if any xml file is found it will be deserialized, if deserialization is successful then load the settings (it will check for all the xml files in the location until one is deserialized successfully).

### Declarations

### Release Notes

Load Preferences xml file located in the ProgramData/Dynamo path.

### Reviewers

@QilongTang 

### FYIs

@jesusalvino

### DIFF
--- a/src/DynamoCoreWpf/Views/SplashScreen/SplashScreen.xaml.cs
+++ b/src/DynamoCoreWpf/Views/SplashScreen/SplashScreen.xaml.cs
@@ -203,6 +203,10 @@ namespace Dynamo.UI.Views
             // Stop the timer in any case
             loadingTimer.Stop();
             loadingTimer = null;
+
+            //When a xml preferences settings file is located at C:\ProgramData\Dynamo will be read and deserialized so the settings can be set correctly.
+            LoadPreferencesFileAtStartup();
+
             // If user is launching Dynamo for the first time or chose to always show splash screen, display it. Otherwise, display Dynamo view directly.
             if (viewModel.PreferenceSettings.IsFirstRun || viewModel.PreferenceSettings.EnableStaticSplashScreen)
             {
--- a/src/DynamoCoreWpf/Views/SplashScreen/SplashScreen.xaml.cs
+++ b/src/DynamoCoreWpf/Views/SplashScreen/SplashScreen.xaml.cs
@@ -4,6 +4,8 @@ using System.IO;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Windows;
+using System.Xml.Serialization;
+using Dynamo.Configuration;
 using Dynamo.Controls;
 using Dynamo.Core;
 using Dynamo.Logging;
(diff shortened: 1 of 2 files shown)


RobertGlobant20:
This is a GIF showing that the xml file in C:\ProgramData\Dynamo has the value MaxNumRecentFiles = 50 and then is reflected in the Preferences panel.
LoadingDynamoPreferences

src/DynamoSandbox/DynamoCoreSetup.cs
```diff
@@ -154,6 +154,9 @@ private void LoadDynamoView()
             splashScreen.DynamoView = new DynamoView(viewModel);
             splashScreen.OnRequestStaticSplashScreen();
 
+            //When a xml preferences settings file is located at C:\ProgramData\Dynamo will be read and deserialized so the settings can be set correctly.
+            splashScreen.LoadPreferencesFileAtStartup();
```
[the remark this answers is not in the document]
RobertGlobant20: @QilongTang here my comments:
1. No, I was under the idea that this is a external(or different) import process than splashscreen, I mean, If the user puts a xml file in C:\ProgramData\Dynamo, then opens Dynamo the splashscreen will be shown and the settings will be loaded automatically (automatic process), if the user in splashscreen clicks the "Import Settings"button and select a xml file then it will overwrite the settings already set by the automatic process (let me know if this is the expected behavior).

2. Yes, I will move this line to the SplashScreen.xaml.cs event handler.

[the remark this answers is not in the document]
RobertGlobant20: @QilongTang I moved the method call to splashscreen.
commit: https://github.com/DynamoDS/Dynamo/pull/13500/commits/3c6c3805bddf79b9bb586356864c4502983e1233

QilongTang (approved):

src/DynamoCoreWpf/Views/SplashScreen/SplashScreen.xaml.cs
```diff
@@ -340,6 +346,69 @@ internal async void SetLabels()
...
 
+        /// <summary>
+        /// At Dynamo startup process load the preferences settings file located in C:\ProgramData\Dynamo
+        /// </summary>
+        internal void LoadPreferencesFileAtStartup()
+        {
+            if (viewModel.PreferenceSettings.IsFirstRun == true)
+            {
+                //Move the current location two levels up
+                var programDataDir = Directory.GetParent(Directory.GetParent(viewModel.Model.PathManager.CommonDataDirectory).ToString()).ToString();
```
QilongTang, line 357: Would you make it a property not that it should be but that way I imagine we can add unit test for it but not direct splash screen to import setting from a different test location? Anyway, I would like a unit test to guard this behavior if possible.

QilongTang:
LGTM with the request of one unit test

src/DynamoCoreWpf/Views/SplashScreen/SplashScreen.xaml.cs
RobertGlobant20: @QilongTang 
Then my understanding is that you are requesting a unit test which validates that the NewProperty added:

NewProperty = Directory.GetParent(Directory.GetParent(viewModel.Model.PathManager.CommonDataDirectory).ToString()).ToString();

is always pointing to the path "C:\ProgramData\Dynamo".
is that correct?

QilongTang: @RobertGlobant20 Sorry for the confusion, I would like a unit test to valid if initializing Splash Screen with isFirstTime = true, we will try to import a setting for a defined property path, does it make sense?

RobertGlobant20: @QilongTang 
I just did a quick analysis of creating an UI Unit Test for SplashScreen and the problem that I can see is that DynamoTestUIBase class (base class for most of the UI unit tests) is always creating the DynamoView and DynamoModel instances so we need to create a new base UI test class (something like SplashScreenDynamoUIBase class) or modify the current one (that could affect all the UI Unit Tests that derive from DynamoTestUIBase ) to create the SplashScreen instance and programatically click the "Launch Dynamo" button (in the SplashScreen).
Also another thing is that we need to copy/create a xml file to C:\ProgramData\Dynamo (not sure if we have permissions to do that in the CI server) and that needs to happen before the SplashScreen OnStaticScreenReady event handler is executed (we can do that in the new SplashScreenDynamoUIBase class before creating the SplashScreen instance).
Then I think we should consider create a new Jira task for coding the necessary infrastructure for creating SplashScreen tests and analyze all the possible solutions (so we don't affect existing UI Tests).
Please let me know your thoughts about it. Thanks
FYI:  @filipeotero @jesusalvino @reddyashish @zeusongit @avidit

QilongTang: @RobertGlobant20 I think the new UITestBase class is more of an issue than the file path. The second issue you mentioned is exactly why I recommend you moving the path to a property so we can ulter the default behavior when running tests.
I agree on creating a new Jira task. Let's catch up on Monday and I will merge this for now.

PR #13547, opened 2022-11-21
Cross platform DynamoCore local setup (part1)

pinzart90:
Purpose:
Provide a local configuration/system to build DynamoCore for net60 (windows and linux)

What changed:
Added local build configuration for dotnet6.0 (linux and windows supported OSs)
A new solution DynamoCore.net6.sln has been added, to make DEV experience easier. When the dotnet6 ecosystem is mature enough, we can move everything over into a single Dynamo.Core.sln (multi target without UI) and a Dynamo.All.sln (multi target with UI)
Windows specific APIs have been marked with a `[SupportedOSPlatform("Windows")]` attribute.
net48 and net6.0 Packages have been segregated.
Some of the codebase has changed to accommodate net6.0 and non-windows APIs

TODO:
The new PackageReferences  (for dotnet6.0) have not been vetted properly.
Bring in more of the Dynamo Core projects.
Remove windows only dependencies from the Core and Linux build if not needed (ex image resources). Having certain resources means we are stuck with msbuild and cannot simply use dotnet build for our build process
There are dependencies of DynamoCore+Linux which still need to be compiled specifically for linux (Ex. LibG native binaries need to be compiled for Linux or else they will not load properly)

`DynamoCore.net6.sln`:
image


src/Config/CS_SDK.props
```diff
@@ -19,8 +19,6 @@
     <TestProjectDefaults Condition="$(MSBuildProjectDirectory.Contains('test'))">true</TestProjectDefaults>
     <TestProjectDefaults Condition=" '$(TestProjectDefaults)' == '' ">false</TestProjectDefaults>
     <NoWarn>MSB3539</NoWarn><!--The value of the property "BaseIntermediateOutputPath" was modified after it was used by MSBuild-->
-     <!--This is used to set the framework version for icon resource dll generation (lookup of system.drawing.dll)-->
-    <ResourceGeneration_FrameworkVersion>v4.8</ResourceGeneration_FrameworkVersion>
```
mjkkirschner, line 23: ?

src/DynamoCLI/DynamoCLI.csproj
```diff
@@ -44,7 +40,6 @@
     <ProjectReference Include="..\NodeServices\DynamoServices.csproj">
       <Project>{ef879a10-041d-4c68-83e7-3192685f1bae}</Project>
       <Name>DynamoServices</Name>
-      <Private>False</Private>
```
mjkkirschner, line 47: ?

src/Config/CS_SDK.props
pinzart90: Moved it at line number 116

src/DynamoApplications/StartupUtils.cs
```diff
@@ -205,6 +217,9 @@ private static void ShowHelp(OptionSet opSet)
         /// </summary>
         /// <param name="geometryFactoryPath">libG ProtoInterface path</param>
         /// <param name="preloaderLocation">libG folder path</param>
+#if NET6_0_OR_GREATER
+        [System.Runtime.Versioning.SupportedOSPlatform("windows")]
```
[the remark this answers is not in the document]
pinzart90: Not sure if they are completely omitted by the compiler but they do generate build errors if you try to use them with an incompatible runtime identifier (i,e. <RuntimeIdentifier>linux-x64</RuntimeIdentifier> will generate compile error if you try to use a windows only API)

src/DynamoCLI/DynamoCLI.csproj
pinzart90: Will revert

src/Config/CS_SDK.props
```diff
@@ -1,14 +1,15 @@
 <Project>
   <PropertyGroup>
+    <Platforms>AnyCPU;NET60_Windows;NET60_Linux;</Platforms>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <PlatformTarget >AnyCPU</PlatformTarget>
```
pinzart90, line 6: `Platform` can have any value. It is user-defined.
`PlatformTarget` must have a valid value of AnyCPU, x86, x64, ARM,

src/Config/CS_SDK.props
```diff
@@ -19,10 +20,9 @@
     <TestProjectDefaults Condition="$(MSBuildProjectDirectory.Contains('test'))">true</TestProjectDefaults>
     <TestProjectDefaults Condition=" '$(TestProjectDefaults)' == '' ">false</TestProjectDefaults>
     <NoWarn>MSB3539</NoWarn><!--The value of the property "BaseIntermediateOutputPath" was modified after it was used by MSBuild-->
-     <!--This is used to set the framework version for icon resource dll generation (lookup of system.drawing.dll)-->
-    <ResourceGeneration_FrameworkVersion>v4.8</ResourceGeneration_FrameworkVersion>
     <ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>None</ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>
     <DebugSymbols>true</DebugSymbols>
+    <GenerateDependencyFile>false</GenerateDependencyFile>
```
pinzart90, line 25: I turned this off because it is causing issues (could not find assembly references in the app directory) with the existing `Copy local`=`false` settings in Dynamo.  We do not need the deps files at the moment

```
.deps.json - A list of dependencies, as well as compilation context data and compilation dependencies. 
Not technically required, but required to use the servicing or package cache/shared package install features. 
It can be used to configure dynamic linking to assemblies that come from packages. 
As mentioned above, .NET Core can be configured to dynamically load assemblies from multiple locations. These locations include:

App base directory (in the same folder as the entry point application, no config required)
Package cache folders (NuGet restore cache or NuGet fallback folders)
An optimized package cache or runtime packages store
The servicing index (rarely used. For Windows Update purposes.)
Shared framework (configured via runtimeconfig.json)
```

src/Config/CS_SDK.props
```diff
@@ -34,18 +34,38 @@
...
+  <PropertyGroup Condition="$(Platform.Contains('Linux'))" >
+    <RuntimeIdentifier>linux-x64</RuntimeIdentifier>
+  </PropertyGroup>
+  <PropertyGroup>
+    <BaseIntermediateOutputPath>int\$(TargetFramework)\</BaseIntermediateOutputPath>
+    <IntermediateOutputPath>obj\$(TargetFramework)\</IntermediateOutputPath>
+  </PropertyGroup>
+  <PropertyGroup Condition="$(Platform.Contains('NET60'))" >
+    <!--Needed to copy nuget package assemblies to output folder. Anet6 issue-->
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
```
pinzart90, line 52:
```
The CopyLocalLockFileAssemblies property is useful for plugin projects that have dependencies on other libraries. If you set this property to true, any NuGet package dependencies are copied to the output directory. That means you can use the output of dotnet build to run your plugin on any machine.
```

src/DynamoCLI/DynamoCLI.csproj
pinzart90: reverted

.github/workflows/dynamoNet6.0_build.yml
```diff
@@ -0,0 +1,41 @@
+# Build Dynamo using latest VS and DotNET
+name: Dynamo-net6.0-Build
+on: [push,pull_request]
+jobs:
+ build:
+  runs-on: windows-latest
```
mjkkirschner, line 6: should this also build on unbuntu?

src/Config/CS_SDK.props
mjkkirschner: maybe not the time for it, but we don't really support x86 anymore.

src/DynamoApplications/StartupUtils.cs
```diff
@@ -13,19 +13,31 @@
...
 using DynamoShapeManager;
+using DynamoUtilities;
 using Microsoft.Win32;
 using NDesk.Options;
 
 namespace Dynamo.Applications
 {
+#if NET6_0_OR_GREATER
+    // Add the same signature for the OptionSet.Add method as found in the NDesk.Options package for net48
+    internal static class OptionsExtensions
```
mjkkirschner, line 24: maybe a better solution is to switch to https://github.com/DynamoDS/Dynamo/blob/83161e3be55c43aaf0ae0342cae80c66a0edbcbe/src/Tools/NodeDocumentationMarkdownGenerator/NodeDocumentationMarkdownGenerator.csproj#L14

like we use in newer parts of the repo.

src/DynamoUtilities/PathHelper.cs
```diff
@@ -1,16 +1,30 @@
...
 using System.Linq;
+using System.Runtime.Versioning;
 using System.Security.AccessControl;
 using System.Security.Principal;
 using System.Xml;
 using Newtonsoft.Json;
 
 namespace DynamoUtilities
 {
+    internal static class OSHelper
```
mjkkirschner, line 13: kind of an unexpected location I guess - inside path helper?

src/Engine/ProtoCore/ProtoCore.csproj
```diff
@@ -12,21 +12,30 @@
...
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net6.0' ">
+    <PackageReference Include="System.CodeDom" Version="6.0.0" />
+    <PackageReference Include="SoapFormatter" Version="1.0.11" />
+    <PackageReference Include="System.Reflection.Emit" Version="4.7.0" />
+  </ItemGroup>
   <ItemGroup>
+    <None Include="FFI\FFICppFunction.cs" />
+    <None Include="FFI\PInvokeFFI.cs" />
+    <Compile Condition=" '$(TargetFramework)' == 'net6.0' " Remove="FFI\FFICppFunction.cs" />
+    <Compile Condition=" '$(TargetFramework)' == 'net6.0' " Remove="FFI\PInvokeFFI.cs" />
```
mjkkirschner, line 36: are these files actually still used, even in net framework 48?

src/Libraries/CoreNodeModels/CoreNodeModels.csproj
```diff
@@ -20,14 +20,18 @@
   <ItemGroup>
     <None Remove="CoreNodeModelsImages.resources" />
   </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net48' ">
```
mjkkirschner, line 23: I think we should just dump this node - I don't think theres anything it can do that a code block cannot.

src/Libraries/CoreNodes/CoreNodes.csproj
```diff
@@ -15,14 +15,22 @@
...
     <Reference Include="Microsoft.CSharp" />
     <PackageReference Include="DynamoVisualProgramming.LibG_229_0_0" Version="2.16.0.3357"/>
     <PackageReference Include="ncalc" Version="1.3.8">
       <!--Exclude copying the runtime dlls because we need to keep binary compatibility with Revit for now -->
       <ExcludeAssets>runtime</ExcludeAssets>
     </PackageReference>
     <Reference Include="PresentationCore" />
+  </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net6.0' ">
+    <PackageReference Include="CoreCLR-NCalc" Version="2.2.92" />
```
mjkkirschner, line 28: same as above, I don't think it's worth the complexity. Maybe we can even migrate to a codeblock or python.

src/Libraries/DynamoUnits/Units.csproj
```diff
@@ -10,10 +10,18 @@
     <AssemblyName>DynamoUnits</AssemblyName>
     <DocumentationFile>$(OutputPath)\DynamoUnits.xml</DocumentationFile>
   </PropertyGroup>
-  <ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'net48' ">
```
mjkkirschner, line 13: I'm confused by this change, it looks like net48 no longer references forge units?

src/Tools/DynamoInstallDetective/DynamoInstallDetective.csproj
```diff
@@ -9,7 +9,11 @@
...
     <AssemblyName>DynamoInstallDetective</AssemblyName>
   </PropertyGroup>
-  <ItemGroup>
-	<Reference Include="System.Configuration" />
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net48' ">
+	  <Reference Include="System.Configuration" />
+  </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net6.0' ">
+    <PackageReference Include="System.Configuration.ConfigurationManager" Version="5.0.0" />
+    <PackageReference Include="Microsoft.Win32.Registry" Version="4.7.0" />
```
mjkkirschner, line 17: again I'm confused, shouldn't this be under net48 not net6?

src/Tools/DynamoShapeManager/Preloader.cs
```diff
@@ -52,7 +52,9 @@ public class Preloader
         /// typical setup this would be the same directory that contains Dynamo 
         /// core modules. This must represent a valid directory.
         /// </param>
-        ///
```
mjkkirschner, line 55: so how will ASM be loaded on linux builds?

src/DynamoApplications/StartupUtils.cs
sm6srw: Is it strictly necessary to keep combability here? I think we can get close enough. Maybe something we do for 3.0.

src/DynamoUtilities/DynamoUtilities.csproj
```diff
@@ -11,9 +11,22 @@
     <RootNamespace>DynamoUtilities</RootNamespace>
     <AssemblyName>DynamoUtilities</AssemblyName>
   </PropertyGroup>
-  <ItemGroup>
-	<PackageReference Include="AvalonEdit" Version="4.3.1.9430" CopyXML="true" />
-	<PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+  <ItemGroup Label="Common Depenedencies">
```
sm6srw, line 14: Spelling

src/Config/CS_SDK.props
```diff
@@ -91,4 +109,19 @@
     <Copy SourceFiles="@(PDBContent)" DestinationFolder="$(OutDir)$(PackagesOutputSubdirectory)" Condition="Exists('%(FullPath)')" />
     <Copy SourceFiles="@(DLLContent)" DestinationFolder="$(OutDir)$(PackagesOutputSubdirectory)$(Local_CopyToSubdirectory)" Condition="Exists('%(FullPath)')" />
   </Target>
+  <Target Name="ResolveSateliteResDeps" AfterTargets="Build" Condition=" '$(OS)' != 'Unix' ">
```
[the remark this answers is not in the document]
pinzart90: it is built in

src/DynamoApplications/StartupUtils.cs
pinzart90: Compatibility was not the main reason.
The main reason was to minimize the `#if def` code for dotnet6 and maximize common code

src/DynamoApplications/StartupUtils.cs
```diff
@@ -13,19 +13,70 @@
...
+        public bool DisableAnalytics { get; set; }
+        [Option("ParentId", Required = false, HelpText = "Identify Dynamo host analytics parent id.")]
+        public string ParentId { get; set; } = String.Empty;
+        [Option("SessionId", Required = false, HelpText = "Identify Dynamo host analytics session id.")]
+        public string SessionId { get; set; } = String.Empty;
+        [Option("CERLocation", Required = false, HelpText = "Specify the crash error report tool location on disk.")]
+        public string CERLocation { get; set; } = String.Empty;
+    }
+#endif
+
```
pinzart90, line 68: I decided to use a separate `internal` class for the CommandLine options. Hopefully this is less cluttered and more readable.

.github/workflows/dynamoNet6.0_build.yml
mjkkirschner: well you can do `mono msbuild`... 
Why do we need msbuild?

src/Config/CS_SDK.props
```diff
@@ -34,18 +33,37 @@
     <Optimize>true</Optimize>
     <DefineConstants>TRACE</DefineConstants>
   </PropertyGroup>
+  <PropertyGroup Condition="$(Platform.Contains('NET60'))" >
+    <TargetFramework>net6.0</TargetFramework>
+    <!--Needed to copy nuget package assemblies to output folder. Anet6 issue-->
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
```
mjkkirschner, line 39: so... does this turn all references into copy local true?

pinzart90: yes, it copies all referenced assemblies in the output path

mjkkirschner: that sounds dangerous - do we have duplicate copies now of things like newtonsoft.json in the `/nodes` and `/bin` folders?

src/DynamoApplications/StartupUtils.cs
```diff
@@ -13,19 +13,70 @@
...
+            "this option is only supported when run from DynamoSandbox")]
+        public string CommandFilePath { get; set; } = String.Empty;
+        [Option('o', "OpenFilePath", Required = false, HelpText = "Instruct Dynamo to open headless and run a dyn file at this path.")]
+        public string OpenFilePath { get; set; } = String.Empty;
+        [Option('v', "Verbose", Required = false, HelpText = "Instruct Dynamo to output all evalautions it performs to an xml file at this path.")]
+        public string Verbose { get; set; } = String.Empty;
+        [Option('x', "ConvertFile", Required = false, HelpText = "When used in combination with the 'O' flag, opens a .dyn file from the specified path and converts it to .json." +
+            "File will have the .json extension and be located in the same directory as the original file.")]
+        public bool ConvertFile { get; set; }
+        [Option('g', "Geometry", Required = false, HelpText = "Instruct Dynamo to output geometry from all evaluations to a json file at this path.")]
```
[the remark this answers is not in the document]
pinzart90: I did not even try. Obviously something is not going to work

src/DynamoApplications/StartupUtils.cs
```diff
@@ -13,19 +13,70 @@
...
+#else
 using NDesk.Options;
+#endif
 
 namespace Dynamo.Applications
 {
+#if NET6_0_OR_GREATER
+    internal class CMDLineOptions
+    {
+        [Option('l', "Locale", Required = false, HelpText = "Running Dynamo under a different locale setting.")]
```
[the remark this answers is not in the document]
mjkkirschner: yep, bummer, I regret the suggestion - 😢

src/DynamoApplications/StartupUtils.cs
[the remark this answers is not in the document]
pinzart90: compile was fine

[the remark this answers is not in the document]
mjkkirschner: doesn't need to be resolved here but we should probably align these and just document the break between the two targets.

src/DynamoApplications/StartupUtils.cs
```diff
@@ -378,7 +465,7 @@ private static bool PreloadASM(string asmPath, out string geometryFactoryPath, o
                 CLIMode = CLImode
             };
             config.AuthProvider = CLImode ? null : new Core.IDSDKManager();
-            config.UpdateManager = CLImode ? null : InitializeUpdateManager();
+            config.UpdateManager = CLImode ? null : OSHelper.IsWindows() ? InitializeUpdateManager() : null;
```
mjkkirschner, line 468: what is windows only about updatemanager?

src/DynamoCLI/DynamoCLI.csproj
```diff
@@ -8,11 +8,12 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>DynamoCLI</RootNamespace>
     <AssemblyName>DynamoCLI</AssemblyName>
+    <SelfContained>false</SelfContained>
```
mjkkirschner, line 11: just curious - what happens if this is true, the exe also includes all the binaries in the bin folder?

pinzart90: SelfContained = true will copy all the dotnet6 runtime dlls

src/DynamoCoreWpf/DynamoCoreWpf.csproj
```diff
@@ -69,6 +69,7 @@
     <None Remove="Views\GuidedTour\HtmlPages\Resources\ConnectTheNode.gif" />
   </ItemGroup>
   <ItemGroup>
+  <PackageReference Include="AvalonEdit" Version="4.3.1.9430" CopyXML="true"/>
```
mjkkirschner, line 72: why was this needed?

src/DynamoApplications/StartupUtils.cs
pinzart90: Hopefully we can get rid of most of the Windows only APIs from DynamoCore
This can be done in steps

src/DynamoPackages/DynamoPackages.csproj
```diff
@@ -9,7 +9,7 @@
     <RootNamespace>DynamoPackages</RootNamespace>
     <AssemblyName>DynamoPackages</AssemblyName>
   </PropertyGroup>
-  <ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net48' ">
```
mjkkirschner, line 12: so, will net6 builds be able to load config files?

src/DynamoUtilities/PathHelper.cs
```diff
@@ -61,7 +75,10 @@ public static bool IsReadOnlyPath(string filePath)
                 // We mark the path read only when
                 // 1. file read-only
                 // 2. user does not have write access to the folder
-                return Finfo.IsReadOnly || !HasWritePermissionOnDir(Finfo.Directory.ToString());
+
+                // We have no cross platform Directory access writes APIs.
+                bool hasWritePermissionOnDir = OSHelper.IsWindows() ? HasWritePermissionOnDir(Finfo.Directory.ToString()) : true;
```
mjkkirschner, line 80: I guess now wherever this is used should be wrapped in try catch?
Maybe update summary?

src/Config/CS_SDK.props
pinzart90: duplicate copies will not happen because all projects that have the outputPath to `/nodes` also have a redirect attribute
```
    <ReferenceCopyLocalPaths>
	  <!--Copy all assembly references to the OutputPath parent dir (/nodes/../) -->
	  <DestinationSubDirectory>..\</DestinationSubDirectory>
    </ReferenceCopyLocalPaths>
    ```

File: src/DynamoApplications/StartupUtils.cs (hunk 20 above)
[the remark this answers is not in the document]
Comment (pinzart90) [reply src/DynamoApplications/StartupUtils.cs:41]: will add a task for it once we are sure that this is the package we want

File: src/Tools/DynamoShapeManager/Utilities.cs [hunk 27]
```diff
@@ -611,11 +622,28 @@ internal static bool IsASMInstallationComplete(IEnumerable<string> filePaths, in
         /// <returns></returns>
         /// <param name="searchPattern">optional - to be used for testing - default is the ASM search pattern</param>
         /// <returns></returns>
-        public static Version GetVersionFromPath(string asmPath, string searchPattern = "ASMAHL*.dll")
+        public static Version GetVersionFromPath(string asmPath, string searchPattern = "*ASMAHL*.*")
```
mjkkirschner, line 625: can you explain this change?

src/Tools/DynamoShapeManager/Utilities.cs
```diff
@@ -611,11 +622,28 @@ internal static bool IsASMInstallationComplete(IEnumerable<string> filePaths, in
...
         {
             var ASMFilePath = Directory.GetFiles(asmPath, searchPattern, SearchOption.TopDirectoryOnly).FirstOrDefault();
             if (ASMFilePath != null && File.Exists(ASMFilePath))
             {
+#if NET6_0_OR_GREATER
+                if (!OperatingSystem.IsWindows())
+                {
+                    string fileName = Path.GetFileNameWithoutExtension(ASMFilePath);
```
mjkkirschner, line 633: I mean, cool, but ugh - I'm surprised theres nothing cross platform here.

mjkkirschner (approved):

mjkkirschner:
A lot of changes, and still some questions, but it looks good.

src/DynamoCoreWpf/DynamoCoreWpf.csproj
pinzart90: this packageRefecence was moved from DynamoUtilities. DynamoCoreWpf.csproj is the project that actually uses it.
I think it was used by DynamoUtilities only for an indirect reference to `System.Collections.Immutable`.
I refactored it so that we do not need it in DynamoUtilities.

This might cause compile issues...if someone was using `AvalonEdit` APIs by referencing the DynamoUtilities project ...But from what I was able to check...looks like we are ok with the autodesk integrators.

src/DynamoPackages/DynamoPackages.csproj
pinzart90: Yes...there is a separate ConfigurationManager nuget package for that

src/DynamoUtilities/PathHelper.cs
pinzart90: Not sure yet how we should go about it for linux. There are packages that have similar APIs ...(as far as I remember)
Or we can implement a simple version ourselves.

I think we need to do this in a separate task...talk about acceptance criteria and maybe some spikes too

src/Tools/DynamoShapeManager/Utilities.cs
pinzart90: .so (shared objects) do not have the same metadata as dlls.
I think the recommended versioning technique is to bake the version in the name as a prefix (ex libtsplines.so.10)

src/DynamoUtilities/PathHelper.cs
```diff
@@ -72,13 +89,19 @@ public static bool IsReadOnlyPath(string filePath)
...
         {
             try
             {
                 var writeAllow = false;
                 var writeDeny = false;
-                var accessControlList = Directory.GetAccessControl(folderPath);
+                DirectoryInfo dInfo = new DirectoryInfo(folderPath);
+                if (dInfo == null)
+                    return false;
+                var accessControlList = dInfo.GetAccessControl();
```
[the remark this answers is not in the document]
pinzart90: The API `Directory.GetAccessControl` is not available on dotnet6
the new API `dInfo.GetAccessControl()` is available on both net48 and dotnet6

src/DynamoApplications/StartupUtils.cs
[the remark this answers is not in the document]
pinzart90: `PreloadShapeManager` is windows only for now because it currently relies a lot on the windows registry to figure out where ASM is installed.
The entire logic of searching a machine for ASM installations does not really make sense for linux, since that will be mainly for services. In the case of services we are most likely going to just supply the path to ASM

src/DynamoUtilities/DynamoUtilities.csproj
```diff
@@ -11,9 +11,22 @@
...
+    <PackageReference Include="System.IO.FileSystem.AccessControl" Version="5.0.0" />
+  </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net48' ">
+    <PackageReference Include="System.Collections.Immutable" Version="5.0.0">
+      <GeneratePathProperty>true</GeneratePathProperty>
+      <IncludeAssets>none</IncludeAssets>
+    </PackageReference>
+    <Reference Include="System.Collections.Immutable">
+      <HintPath>$(PkgSystem_Collections_Immutable)\lib\portable-net45+win8+wp8+wpa81\System.Collections.Immutable.dll</HintPath>
+    </Reference>
```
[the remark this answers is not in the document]
pinzart90: both what ?
The PackageReference does not bring in any references by itself (`<IncludeAssets>none</IncludeAssets>`) because we want to use a specific assembly (the one in `net45+win8+wp8+wpa81\`)
For net48 we've been using `net45+win8+wp8+wpa81\` because it does not bring any other dependencies.
For net6 the extra deps are already brought in by other 3d party packages...so  I saw no need to use `net45+win8+wp8+wpa81` anymore

aparajit-pratap (approved):

PR #13557, opened 2022-11-22
DYN-5388-SplashScreen-Revit

RobertGlobant20:
### Purpose

Removing line that is making Dynamo crash in Revit.
When Dynamo is opened from Revit it was crashing due that Application.Current is null, so I removed the line, we already validated that the MainWindow is already set when reaching this line (this was validated with @filipeotero ), then is not needed.

### Declarations

### Release Notes

Removing line that is making Dynamo crash in Revit.

### Reviewers

@QilongTang 

### FYIs

@reddyashish

### DIFF
--- a/src/DynamoCoreWpf/Views/SplashScreen/SplashScreen.xaml.cs
+++ b/src/DynamoCoreWpf/Views/SplashScreen/SplashScreen.xaml.cs
@@ -190,7 +190,6 @@ namespace Dynamo.UI.Views
             DynamoModel.RequestUpdateLoadBarStatus -= DynamoModel_RequestUpdateLoadBarStatus;
             StaticSplashScreenReady -= OnStaticScreenReady;
             Close();
-            Application.Current.MainWindow = dynamoView;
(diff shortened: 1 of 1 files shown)


QilongTang:
Thanks, waiting for PR checks.

src/DynamoCoreWpf/Views/SplashScreen/SplashScreen.xaml.cs
```diff
@@ -190,7 +190,6 @@ private void LaunchDynamo(bool isCheckboxChecked)
             DynamoModel.RequestUpdateLoadBarStatus -= DynamoModel_RequestUpdateLoadBarStatus;
             StaticSplashScreenReady -= OnStaticScreenReady;
             Close();
-            Application.Current.MainWindow = dynamoView;
```
QilongTang, line 193: It seems we did this later in DynamoView code but probably indeed not needed here

QilongTang:
@RobertGlobant20 Please cherry-pick

PR #13539, opened 2022-11-18
Fix sign-in button UI

zeusongit:
### Purpose

- Move the sign in button to left of export button
- Refactor shortcut toolbar code

DynamoSandbox_Chry1pEjnw

About Box update, added IDSDK disclaimer

Screen Shot 2022-11-21 at 11 58 06 AM

### Declarations

Check these if you believe they are true

- [x] The codebase is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [x] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.
- [ ] This PR modifies some build requirements and the readme is updated

### Release Notes

(FILL ME IN) Brief description of the fix / enhancement. **Mandatory section** 

### Reviewers

(FILL ME IN) Reviewer 1  (If possible, assign the Reviewer for the PR)

(FILL ME IN, optional) Any additional notes to reviewers or testers.

### FYIs

(FILL ME IN, Optional) Names of anyone else you wish to be notified of

PR #13558, opened 2022-11-22
Fix SignOut button UI

zeusongit:
### Purpose

Fix an issue where signing out wont change the sign out button to its initial state: Sign In

### Declarations

### Release Notes

### Reviewers

@QilongTang


src/DynamoCoreWpf/Controls/ShortcutToolbar.xaml.cs
```diff
@@ -49,7 +49,7 @@ public ShortcutToolbar(DynamoViewModel dynamoViewModel)
 
             InitializeComponent();         
 
-            var shortcutToolbar = new ShortcutToolbarViewModel(dynamoViewModel);
+            var shortcutToolbar = new ShortcutToolbarViewModel(this, dynamoViewModel);
```
QilongTang, line 52: I think this would cause memory leak.. usually we do not keep a reference of the view in view model. because view model will not be able to dispose before view. You may want to again use event. e.g. have the view watch for view model property changed

QilongTang:
One comment, please address

src/DynamoCoreWpf/Controls/ShortcutToolbar.xaml.cs
zeusongit: Ok, so I moved the code to the code behind itself. Please review now. Thanks

zeusongit:
@QilongTang After this PR gets merged I should cherry-pick both PRs this and https://github.com/DynamoDS/Dynamo/pull/13539 as well right?

QilongTang (commented):
Overall looks solid. The only thing left, which is my bad.. I remember in C# it  is hard to unsubscribe from anonymous handler which is another source of memory leak...
you may want to switch to C#7 way
```
EventHandler foo = null;
foo = delegate(object s, EventArgs ev)
    {
        // Your code
        MyEvent -= foo;
    };
MyEvent += foo;
```
Or
```
void foo(object s, MyEventArgs ev)
{
     // Your code
    MyEvent -= foo;
};
MyEvent += foo;
```

QilongTang (approved):

QilongTang:
> @QilongTang After this PR gets merged I should cherry-pick both PRs this and #13539 as well right?

Yes, see if you can, the licence.rtf file may give you trouble (conflict where you need to apply manually)

zeusongit:
> 

Updated to only subscribe to logout event when the option is available, and unsubscribing after completing action.

reddyashish (approved):

PR #13566, opened 2022-11-23
Modify existing events so that a user can be notified before a graph is executed

pinzart90:
https://jira.autodesk.com/browse/DYN-5255

The reported issue:
User needs to gathering information for each node when a Graph is opened/executed. When the `CurrentWorkspaceChanged` event is triggered, the engine controller points to an object that will be replaced before the graph is run.
Side issue:
Clutter and redundant flags in the Graph.Run logic.

The flow for the reported issue :
CurrentWorkspaceChanged is triggered
User code enables profiling on the current engine controller of that workspace.
Dynamo resets the engine controller to a new controller.
Profiling is now disabled.
User is confused.

This is the proposed solution to get the proper events:
```
        /// Some Profiling Extension
        public void Loaded(ViewLoadedParams p)
        {
            // Triggered when a dyn/dyf is opened
            p.CurrentWorkspaceChanged += (ws) =>
            {
                if (ws is HomeWorkspaceModel hwm)
                {
                    hwm.EvaluationStarted += (object sender, EventArgs args) => { ProfilePerformance(hwm.EngineController); };
                    hwm.EvaluationCompleted += (object sender, EvaluationCompletedEventArgs args) => { };
                }
            };
```


test/DynamoCoreWpfTests/ViewExtensions/GraphNodeManagerViewExtensionTests.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
```
pinzart90, line 1: not sure how this keeps happening...

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -831,10 +834,17 @@ public void Run()
             EngineController.ProcessPendingCustomNodeSyncData(scheduler);
             
             var task = new UpdateGraphAsyncTask(scheduler, verboseLogging);
+
+            // Start of graph evaluation (includes task.Initialize)
+            // Set these flags to false so that we do not get into infinite loop scenarios
+            // if listeners of EvaluationStarted decide to call RequestRun
+            RunSettings.RunEnabled = false; // Disable 'Run' button.
+            executingTask = false; // Not yet executing.
+            OnEvaluationStarted(EventArgs.Empty);
```
pinzart90, line 843: moved the `OnEvaluationStarted` event call here so that it can run before `task.Initialize(EngineController, this)` (where the AST is parsed and can affect performance)

mjkkirschner: I find it hard to reason about this change.
Is it important that `OnEvaluationStarted` used to be called *after* `silenceNodeModifications` was set true.
Do we use `OnEvaluationStarted` anywhere else internally or inside integrations?

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -39,11 +39,27 @@ public class HomeWorkspaceModel : WorkspaceModel
...
+        /// </summary>
+        internal bool GraphRunInProgress
+        {
+            get { return graphRunInProgress; }
+            set
+            {
+                if (graphRunInProgress == value) return;
+
+                graphRunInProgress = value;
+                RaisePropertyChanged("GraphRunInProgress");
```
mjkkirschner, line 59: nameof

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -186,9 +205,19 @@ internal bool IsEvaluationPending
         public virtual void OnEvaluationStarted(EventArgs e)
         {
             this.HasRunWithoutCrash = false;
+            GraphRunInProgress = true;
 
-            var handler = EvaluationStarted;
-            if (handler != null) handler(this, e);
+            try
+            {
+                // Do not allow graph runs to be trigered from the EvaluationStarted event.
```
mjkkirschner, line 212: spelling (triggered)

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -186,9 +205,19 @@ internal bool IsEvaluationPending
...
+                // Use a simple internal flag like ForceBlockRun (RunEnabled would notify WPF of changes and that would incur peformance penalties without any benefit)
+                RunSettings.ForceBlockRun = true;
```
mjkkirschner, line 214: couldn't you use `GraphRunInProgress` that you just added?

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -490,11 +521,7 @@ internal override void RequestRun()
 
             if (RunSettings.RunType != RunType.Manual)
             {
-                // TODO for Dynamo 3.0: The boolean "executingTask" that is used here is a make-do fix.
-                // We will be needing a separate variable(boolean flag) to check whether run can be enabled from external applications
-                // and not confuse it with the internal flag RunEnabled which is associated with the Run button in Dynamo. 
-                // Make this RunSettings.RunEnabled private, introduce the new flag and remove the "executingTask" variable. 
-                if ((RunSettings.RunEnabled || executingTask) && !DelayGraphExecution)
```
mjkkirschner, line 497: there were some hairy bugs this flag was related to I think either @reddyashish or @RobertGlobant20 worked on this - did you look into the history of why this and the comments exist?

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -1871,7 +1871,7 @@ private void ShowOpenDialogAndOpenResult(object parameter)
 
         private bool CanShowOpenDialogAndOpenResultCommand(object parameter)
         {
-            return HomeSpace.RunSettings.RunEnabled;
+            return HomeSpace.RunSettings.RunEnabled && !HomeSpace.GraphRunInProgress;
```
mjkkirschner, line 1874: I'm assuming these changes are not really related to this PR but are good anyway to avoid edge cases.

src/DynamoCoreWpf/ViewModels/RunSettingsViewModel.cs
```diff
@@ -266,6 +267,8 @@ public RunSettingsViewModel(RunSettings settings, HomeWorkspaceViewModel workspa
             Model.PropertyChanged += Model_PropertyChanged;
 
             this.workspaceViewModel = workspaceViewModel;
+            workspaceViewModel.Model.PropertyChanged += HomeWorkspaceModel_PropertyChanged;
```
mjkkirschner, line 270: where can you unsubscribe from this handler?

src/DynamoCoreWpf/ViewModels/RunSettingsViewModel.cs
```diff
@@ -292,6 +295,26 @@ public override void Dispose()
 
         #region private and internal methods
 
+        /// <summary>
+        /// Notifies all relevant Dynamo features (UI elements, commands) that the Graph exection has been enabled/disabled. 
+        /// </summary>
+        void NotifyOfGraphRunChanged()
+        {
+            RaisePropertyChanged("RunButtonEnabled");
```
mjkkirschner, line 303: nameof

src/DynamoCoreWpf/ViewModels/RunSettingsViewModel.cs
```diff
@@ -292,6 +295,26 @@ public override void Dispose()
...
+            RaisePropertyChanged("RunButtonToolTip");
+
+            if (Application.Current != null)
+            {
+                Application.Current.Dispatcher.Invoke(new Action(() =>
```
mjkkirschner, line 308: any reason to use invoke instead of begin invoke?

test/DynamoCoreWpfTests/ViewExtensions/GraphNodeManagerViewExtensionTests.cs
mjkkirschner: I think @avidit added an editor config - perhaps it has some line endings preference.

test/DynamoCoreWpfTests/ViewExtensions/ViewExtensionTests.cs
```diff
@@ -346,6 +353,44 @@ public static void RaiseLoadedEvent(FrameworkElement element)
...
+        [Test]
+        public void TestGraphEvaluationEvents()
+        {
+            RaiseLoadedEvent(this.View);
+            var extensionManager = View.viewExtensionManager;
+            extensionManager.Add(viewExtension);
+
+            EngineController controller = null;
+            int counter = 0;
+            viewExtension.OnOpenEvaluationStarted += (HomeWorkspaceModel hwm, EventArgs args) =>
```
mjkkirschner, line 369: unsub these somewhere in the test or in a finally etc.

test/DynamoCoreWpfTests/ViewExtensions/ViewExtensionTests.cs
```diff
@@ -374,6 +422,12 @@ public void Loaded(ViewLoadedParams p)
         {
             p.CurrentWorkspaceChanged += (ws) =>
             {
+                if (ws is HomeWorkspaceModel hwm)
+                {
+                    hwm.EvaluationStarted += (object sender, EventArgs args) => { OnOpenEvaluationStarted?.Invoke(hwm, args); };
```
mjkkirschner, line 427: unsub

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
pinzart90: `GraphRunInProgress` does not block execution.
It is just a flag that indicates whether Dynamo is currently running a graph.

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
pinzart90: Yes. 
Here is the scripture:
1. In the beginning (~2015) the `RunEnabled` flag was introduced as a way to notify UI elements (ex Run button) that the graph was running.  [commit](https://github.com/DynamoDS/Dynamo/commit/41397868fc5a15dfeb272604544fc80cd75a7aad#diff-69f082a38992f2e5879342dcf124febcd476976748fe61a2bc1006041811562cR192)
2. In 2019 the `RunEnabled` flag was extended to also block execution. [commit](https://github.com/DynamoDS/Dynamo/commit/504f53b97deece81aea8f666b03552abd94a4bb8#diff-df9fc71f88d3bf7d4b6700f044ed43c5c6d16e44b80ba2efdef111f75ee16c5fR334)
3. Shortly after that a bug was identified and another flag was introduced `executingTask`. This was again meant to represent that the Graph was currently running or not. [commit](https://github.com/DynamoDS/Dynamo/commit/76c7b618b1668aee9b99571e3e75706c70a14f4a#diff-df9fc71f88d3bf7d4b6700f044ed43c5c6d16e44b80ba2efdef111f75ee16c5fR36)

The requirements (so far) around when a Graph is being run are:
1. **Parts of the UI need to be disabled when a graph is running (ex Run button)**
2. **If a Graph Run has started (i.e GraphUpdate task has been scheduled), other Graph run calls need to be permitted (i.e other GraphUpdate tasks can be created**). An example case would be hooking up a connector to an already connected port. This translates to 2 consecutive Graph runs. 
3. **Host integrators want an API (they are already using` RunEnabled`)  to disable Graphs from being run (not just UI) when the host apps change documents**

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
pinzart90: no need to set `silenceNodeModifications` during/before the `OnEvaluationStarted` event because we are already blocking graph runs with `ForceBlockRun`

pinzart90:
I reviewed the ALIAS, Civil3D, Revit and SteelConnections host integrations
Only Civil3D is using the EvaluationStarted event and it does not seem like they are doing anything special. 
I can request that they run tests if we merge to master. 

Or are there any other suggestions on how to proceed ?

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
pinzart90: These changes are needed because of the refactor I did regarding `RunEnabled`
`RunEnabled` is not set by Dynamo during a graph execution anymore. We introduced the new internal flag `GraphRunInProgress` (previously called `taskExecuting`)

src/DynamoCoreWpf/ViewModels/RunSettingsViewModel.cs
pinzart90: This code was moved (not added). 
I am not sure what the implications would be if we made the change

test/DynamoCoreWpfTests/ViewExtensions/ViewExtensionTests.cs
pinzart90: I am not sure if we need to. Aren't all these objects destroyed anyway at the end of the test?

mjkkirschner: I don't know - it's hard to prove without using a memory profiler.

I guess you can run the master-15 self service and see if the job runs out of memory or if the test time goes up - but it seems faster to just always unsubscribe.

mjkkirschner (approved):

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
pinzart90: The state of things before this PR:
If someone sets the `RunEnabled` flag to false, the Graph Run related UI is disabled and Graph executions are disabled (at least WorkspaceModel::RequestRun is disabled)

Checking for both `HomeSpace.RunSettings.RunEnabled && !HomeSpace.GraphRunInProgress` keeps the same functionality

src/DynamoCoreWpf/ViewModels/RunSettingsViewModel.cs
pinzart90: There is a Dispose method. I unsubscribed there

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -186,9 +205,19 @@ internal bool IsEvaluationPending
         public virtual void OnEvaluationStarted(EventArgs e)
         {
             this.HasRunWithoutCrash = false;
+            GraphRunInProgress = true;
 
-            var handler = EvaluationStarted;
-            if (handler != null) handler(this, e);
+            try
+            {
+                // Do not allow graph runs to be triggered from the EvaluationStarted event.
```
aparajit-pratap, line 212: Why?

test/DynamoCoreWpfTests/ViewExtensions/ViewExtensionTests.cs
pinzart90: done

test/DynamoCoreWpfTests/ViewExtensions/ViewExtensionTests.cs
pinzart90: done

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -1950,7 +1950,7 @@ private void OpenRecent(object path)
 
         private bool CanOpenRecent(object path)
         {
-            return HomeSpace.RunSettings.RunEnabled;
+            return HomeSpace.RunSettings.RunEnabled && !HomeSpace.GraphRunInProgress;
```
aparajit-pratap, line 1953: Since these two conditions are used in several places, how about wrapping them in a new property that is appropriately named to convey its exact meaning?

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
pinzart90: We are now consistently blocking calls to `Run()` (graph run) from the `EvaluationStarted` event for the following reasons:
1. We did this before my PR, only it was not consistent
2. Allowing subsequent calls to Run() from `EvaluationStarted` can lead to infinite loops
3. When the `EvaluationStarted` event is fired, we are already inside the `Run` function. Node modifications are allowed (ex one can add more nodes to be computed in the `EvaluationStarted` event)

Clarifications on 1 (We did this before my PR):
Only calls to RequestRun were blocked by 
```
if ((RunSettings.RunEnabled || executingTask) && !DelayGraphExecution)
                {
                    Run();
                }
```
Depending on the timing of things, we could have had `executingTask` be true or false during the `EvaluationStarted` event, because `executingTask` is set to true before the tasks are scheduled and set back to false after a task is completed (asynchronously)

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
pinzart90: done

pinzart90:
@aparajit-pratap do you need anything else cleared up with this PR ?

src/DynamoCore/Graph/Workspaces/HomeWorkspaceModel.cs
```diff
@@ -204,7 +233,9 @@ public virtual void OnEvaluationStarted(EventArgs e)
         public virtual void OnEvaluationCompleted(EvaluationCompletedEventArgs e)
         {
             this.HasRunWithoutCrash = e.EvaluationSucceeded;
-            
+            GraphRunInProgress = false;
+            EvaluationCount++;
```
aparajit-pratap, line 237: Sometimes `OnEvaluationCompleted` is called even when evaluation is skipped for various reasons. Do you want this counter to increment even for these cases?

pinzart90: I guess I am not sure, I do not have any good reasons to go one way or the other
I guess it is safer to leave the counter as it was...

aparajit-pratap (approved):

pinzart90:
A test failed `WorkspaceContextMenu_TestIfSearchTextClearsOnOpeningContextMenu` but it looks like a fluke
Passed locally
Merging

PR #12687, opened 2022-03-11
Call the save method on the CustomNodeWorkspaceModel, when the custom node is being saved.

reddyashish:
### Purpose

This PR is to fix the: https://jira.autodesk.com/browse/DYN-4460. 

When a new custom node is created and saved to a local path in a dynamo session, the custom node manager doesn't have the path information for that particular custom node. It will only have the path information after dynamo is restarted because Dynamo tries to load all the definitions from the node package paths. This was happening because when the CN is saved, during the InternalSaveAs method in dynamo view model, the workspace model's save function is always called. 

This save method  on the CustomNodeWorkspaceModel model is never called: https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Graph/Workspaces/CustomNodeWorkspaceModel.cs#L298. When the custom node workspace is being saved, this method has to be called and that will make sure that the path information is updated in the custom node manager.

Fix custom node saving issue

### Declarations

Check these if you believe they are true

- [ ] The codebase is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.
- [ ] This PR modifies some build requirements and the readme is updated

### Release Notes

Call the save method on the CustomNodeWorkspaceModel, when the custom node is being saved.

### Reviewers
@QilongTang @mjkkirschner @zeusongit

PR #13585, opened 2022-11-30
Fix CustomNode view layout after it is saved.

reddyashish:
### Purpose

Task: https://jira.autodesk.com/browse/DYN-5423. 

This PR is to fix the custom node layout by serializing the "Views" property when a new file is saved. This is done in the WorkspaceViewModel which will handle both the HomeWorkspaceModel and CustomNodeWorkspaceModel. 

CN saving

The previous check for CustomNodeWorkspaceModel is moved inside WorkspaceViewModel, where we will update the file path info in the custom node manager after a new custom node file is saved (related to the [PR](https://github.com/DynamoDS/Dynamo/pull/12687)).

### Declarations

### Release Notes
Fix CustomNode view layout after it is saved.

### Reviewers
@DynamoDS/dynamo

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
@@ -2001,16 +2001,7 @@ namespace Dynamo.ViewModels
             try
             {
                 Model.Logger.Log(String.Format(Properties.Resources.SavingInProgress, path));
-
-                // If the current workspace is a CustomNodeWorkspaceModel, then call the save method on it.
-                if (CurrentSpaceViewModel.Model is CustomNodeWorkspaceModel)
-                {
-                    CurrentSpaceViewModel.Model.Save(path, false, Model.EngineController);
-                }
-                else
-                {
-                    CurrentSpaceViewModel.Save(path, isBackup, Model.EngineController, saveContext);
-                }
+                CurrentSpaceViewModel.Save(path, isBackup, Model.EngineController, saveContext);
 
                 if (!isBackup) AddToRecentFiles(path);
             }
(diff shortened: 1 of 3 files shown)


QilongTang:
Excellent! Do you know when this regression begin to happen?

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -2339,7 +2330,8 @@ public void ShowSaveDialogAndSaveResult(object parameter)
                     _fileDialog.InitialDirectory = fi.DirectoryName;
                     _fileDialog.FileName = fi.Name;
                 }
-                else if (vm.Model.CurrentWorkspace is CustomNodeWorkspaceModel)
+
```
reddyashish, line 2333: Removed the "else" check here as this condition for CustomNodeWorkspaceModel should be met irrespective of the FileName being null.

reddyashish:
@QilongTang After this https://github.com/DynamoDS/Dynamo/pull/12687 fix by me, which was to store and retrieve the custom node file-path info after it is saved without the need to restart Dynamo.

QilongTang:
> @QilongTang After this #12687 fix by me, which was to store and retrieve the custom node file-path info after it is saved without the need to restart Dynamo.

Gotcha, looks like this has been failing for a while. Would you add a unit test to guard it?

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -2001,16 +2001,7 @@ private void InternalSaveAs(string path, SaveContext saveContext, bool isBackup
...
-                // If the current workspace is a CustomNodeWorkspaceModel, then call the save method on it.
-                if (CurrentSpaceViewModel.Model is CustomNodeWorkspaceModel)
-                {
-                    CurrentSpaceViewModel.Model.Save(path, false, Model.EngineController);
-                }
-                else
-                {
-                    CurrentSpaceViewModel.Save(path, isBackup, Model.EngineController, saveContext);
-                }
+                CurrentSpaceViewModel.Save(path, isBackup, Model.EngineController, saveContext);
```
QilongTang, line 2004: So for none custom node workspace, including saveContext does not matter and would not change the save file result?

reddyashish: yeah no change in result. This save method was called with saveContext before this issue came up.

QilongTang (commented):
LGTM with one question

QilongTang (approved):

PR #13599, opened 2022-12-05
DYN-5405 npm pack dependencies

filipeotero:
### Purpose

This PR is to avoid installing all the npm dependencies of the Notification Center and Splash Screen.
* Use of npm pack instead of npm install
* Extract the downloaded .tgz file
* Update the assembly path of the files 
* Update the .gitignore for the new files

### Declarations

### Reviewers

@QilongTang 

### FYIs

@RobertGlobant20


src/DynamoCoreWpf/DynamoCoreWpf.csproj
```diff
@@ -14,16 +14,33 @@
   </PropertyGroup>
 
   <Target Name="NpmRunBuild" BeforeTargets="BeforeBuild">
-      <Exec Command="npm install --prefix ./ @dynamods/splash-screen@latest" />
+      <Exec Command="npm pack @dynamods/splash-screen@latest" />
   </Target>
+
+    <Target Name="ExtractTGZFile" BeforeTargets="BeforeBuild">
```
QilongTang, line 20: @filipeotero Curious if this is depending on any particular zipping software? If I remember correctly, Windows now comes with unzipping tool OOTB?

src/DynamoCoreWpf/DynamoCoreWpf.csproj
```diff
@@ -14,16 +14,33 @@
...
+        <ItemGroup>
+            <TGZFiles Include="./*.tgz" />
+        </ItemGroup>
+
+        <ItemGroup>
+            <Reversed Include="@(TGZFiles->Reverse())" />
```
QilongTang, line 26: this means?

.gitignore
```diff
@@ -104,11 +104,7 @@ test/core/customast/test.txt
 test/core/files/test.png
 test/core/migration/writetext.txt
 logo.png
-/src/Notifications/node_modules
```
filipeotero, line 107: Removing the old folder hierarchy

src/DynamoCoreWpf/DynamoCoreWpf.csproj
```diff
@@ -14,16 +14,33 @@
   </PropertyGroup>
 
   <Target Name="NpmRunBuild" BeforeTargets="BeforeBuild">
-      <Exec Command="npm install --prefix ./ @dynamods/splash-screen@latest" />
+      <Exec Command="npm pack @dynamods/splash-screen@latest" />
```
filipeotero, line 17: The first step is to download the .tgz file

src/DynamoCoreWpf/DynamoCoreWpf.csproj
```diff
@@ -14,16 +14,33 @@
...
   </Target>
+
+    <Target Name="ExtractTGZFile" BeforeTargets="BeforeBuild">
+        <ItemGroup>
+            <TGZFiles Include="./*.tgz" />
```
filipeotero, line 22: This gets all the files that contain the .tgz suffix

src/DynamoCoreWpf/DynamoCoreWpf.csproj
filipeotero, line 26: Reverts the order of the found files

src/DynamoCoreWpf/DynamoCoreWpf.csproj
```diff
@@ -14,16 +14,33 @@
...
+        </ItemGroup>
+
+        <ItemGroup>
+            <Reversed Include="@(TGZFiles->Reverse())" />
+        </ItemGroup>
+
+        <PropertyGroup>
+            <Last>%(TGZFiles.Identity)</Last>
```
filipeotero, line 30: Gets the last one

src/DynamoCoreWpf/DynamoCoreWpf.csproj
QilongTang: Would you add inline comments, you should be able to use `<!-- comment -->`

src/DynamoCoreWpf/DynamoCoreWpf.csproj
filipeotero: Looks like it's coming with libarchive:
https://learn.microsoft.com/en-us/virtualization/community/team-blog/2017/20171219-tar-and-curl-come-to-windows

src/DynamoCoreWpf/DynamoCoreWpf.csproj
filipeotero: Reverts the order of the files to get the latest version

QilongTang: Would this still work if this project is depending on more than one npm package?

src/DynamoCoreWpf/DynamoCoreWpf.csproj
QilongTang: Should we clean up none latest versioned npm package?

QilongTang (commented):
No doubt this would work, but I have some concern. 
1. If we want to specify the npm package version, similar to versioned Nuget package
2. If we want to clean up the legacy npm package.

Just two thoughts I would like team to think over

filipeotero:
> No doubt this would work, but I have some concern.
> 
> 1. If we want to specify the npm package version, similar to versioned Nuget package. If we stick to latest, it seems hard to get the exact package
> 2. If we want to clean up the legacy npm package.
> 
> Just two thoughts I would like team to think over

1 - If we delete it after being extracted, maybe it won't be an issue since we will not maintain the files locally.

QilongTang (approved):

reddyashish:
The self-service job is running in a loop and the build is failing. Converting this to draft to unblock the jobs.

filipeotero:
We are getting this message from the self-service job:

```
18:27:12    tar -xzf c:\WorkspaceDynamo\src\DynamoCoreWpf\.\dynamods-splash-screen-1.0.5.tgz
18:27:13    'tar' is not recognized as an internal or external command,
18:27:13    operable program or batch file.
18:27:13  c:\WorkspaceDynamo\src\DynamoCoreWpf\DynamoCoreWpf.csproj(38,9): error MSB3073: The command "tar -xzf c:\WorkspaceDynamo\src\DynamoCoreWpf\.\dynamods-splash-screen-1.0.5.tgz" exited with code 9009.
```
Looks like there is no libarchive or something else to extract the tgz file. 
Any ideas on how we can extract the file? 

@QilongTang @reddyashish @RobertGlobant20

QilongTang:
@filipeotero Should we include this dll in the repo, is that enought to leverage it?

mjkkirschner (changes requested):
@filipeotero you cannot include an LGPL license'd component in our repo without consulting legal. I would be hesitant to even reference it, it will only be a continuous source of questions.

filipeotero:
> If this is really the best solution to this problem - then why include the binaries directly instead of referencing nuget packages like: https://www.nuget.org/packages/7-Zip.StandaloneConsole.x64 ?

Here are the things that I tried:
**1 - Extract the file with `tar` and don't insert any binaries.**
Error from the job:
`The command "tar -xzf c:\WorkspaceDynamo\src\DynamoCoreWpf\dynamods-splash-screen-1.0.5.tgz" exited with code 9009.`

**2 - Insert tar.exe from the windows system in the project**
Error from the job:
`The command "../../tar/tar.exe -xzf c:\WorkspaceDynamo\src\DynamoCoreWpf\dynamods-splash-screen-1.0.5.tgz" exited with code -1073741511`

**3 - tar.exe from https://gnuwin32.sourceforge.net/packages/gtar.htm**
Error from the job:
`The command "..\..\extern\tar\tar.exe -xzf c:\WorkspaceDynamo\src\DynamoCoreWpf\dynamods-splash-screen-1.0.6.tgz" exited with code -1073741515.`

**4 - Now from 7zip**

Now I will check this nugget package.

mjkkirschner:
`9009` - is usually something like command not found, how are you pointing to `tar`?
have you tested it locally? Does it work if you use the full path to the tool?

I see a weird combination of / and \ in your commands, maybe those are not exactly what you ran but I would not expect that to work - `\` only on windows.

can you try with `v` for verbose mode, curious if theres any more interested output...

filipeotero:
**how are you pointing to `tar`?** 

On the first attempt, I just called the command directly
Second I pointed to the `extern` folder (To which I added the tar.exe file)

**Have you tested it locally?** 
Yes

**Does it work if you use the full path to the tool?** 
Yes

**From what terminal are you running this?**
Windows command prompt 

 **Have you tried adding tar to your path?**
tar is already included in my PC
check this: https://learn.microsoft.com/en-us/virtualization/community/team-blog/2017/20171219-tar-and-curl-come-to-windows

**can you try with `v` for verbose mode, curious if theres any more interesting output...**
Sure, I will do that.

mjkkirschner:
@filipeotero so - I'm still missing some information.
* does the full build work locally or just your calls to invoke `tar` at the command line? Does this only fail on the build machine? (ie inside docker?)
* IE are you certain that the targets, npm pack steps, and reversal of files are executing in the order you expect?
* I think it would be good to use a message task to log some of the variables you have created, like the last and path etc to see what is getting passed to `tar` - https://learn.microsoft.com/en-us/visualstudio/msbuild/message-task?view=vs-2022 if you don't already see that in the log - you could also try using diagnostic build mode.
*

filipeotero:
**does the full build work locally or just your calls to invoke `tar` at the command line?** 
Both work locally. 

**Does this only fail on the build machine? (ie inside docker?)**
Yes, 

**IE are you certain that the targets, npm pack steps, and reversal of files are executing in the order you expect?**
I guess so because it wouldn't even print the name of the file (dynamods-splash-screen-1.0.6.tgz) if it wasn't executing in the right order. The flow is the following: Download the package -> Find the highest version -> Extract the package -> Delete the package 

**I think it would be good to use a message task to log some of the variables you have created, like the last and path etc to see what is getting passed to `tar` - https://learn.microsoft.com/en-us/visualstudio/msbuild/message-task?view=vs-2022 if you don't already see that in the log - you could also try using diagnostic log mode in visual studio.**
Ok, I will add some messages in the steps. Thanks!

mjkkirschner:
@filipeotero what version of visual studio are you using locally? The builder image in docker is a bit old and based on an old windowsservercore image - perhaps we should update it to something more modern...

mjkkirschner:
@filipeotero @avidit (not sure what investigation you did of the builder image) but
two thoughts:

this post makes it seem like windows doesn't include tar until pretty recently - ... and we're using a build image from 2019.
https://learn.microsoft.com/en-us/virtualization/community/team-blog/2017/20171219-tar-and-curl-come-to-windows

In addition, have you considered using powershell? (which we know works on the builder image)
https://scatteredcode.net/download-and-extract-gzip-tar-with-powershell/

mjkkirschner:
one more thought - if 7zip turns out to be the best solution, another approach is to add it to the builder image - this is nice because we add it for all our jobs that use that image, not just dynamo, in addition it makes it very unlikely we would distribute it - worth checking with legal if that makes things simpler.

filipeotero:
Hi @mjkkirschner, I'm using VS 2019.
What about using the nuget package as you mentioned? I didn't try it yet but if it works we won't need to include binaries, right?

mjkkirschner:
we don't need to include the binaries, but we still need to reference it and still need to go through legal.

avidit:
> @avidit (not sure what investigation you did of the builder image)

filipeotero:
@mjkkirschner I submitted a new commit that deletes the binaries and references the nuget package. Also, I'm still trying to reach the legal team about using 7zip. Thanks!

filipeotero:
@QilongTang Just to let you know, we have to wait for the EngOps to catch up before we merge this. https://jira.autodesk.com/browse/ESC-34251

src/DynamoCoreWpf/DynamoCoreWpf.csproj
```diff
@@ -11,19 +11,50 @@
...
+    <ItemGroup>
+    <None Remove="package\build\index.bundle.js" />
+    <None Remove="package\build\index.html" />
     <None Remove="Views\SplashScreen\WebApp\splashScreenBackground.png" />
   </ItemGroup>
   <ItemGroup>
-    <EmbeddedResource Include="node_modules\@dynamods\splash-screen\build\index.bundle.js" />
-    <EmbeddedResource Include="node_modules\@dynamods\splash-screen\build\index.html" />
+    <EmbeddedResource Include="package\build\index.bundle.js" />
+    <EmbeddedResource Include="package\build\index.html" />
```
QilongTang, line 57: If there is no need for a specific package name, does it mean that we would only support a single npm package per project? Do we need to specify the folder name during unzipping?

QilongTang (commented):
LGTM but with one concern commented

src/DynamoCoreWpf/DynamoCoreWpf.csproj
filipeotero: Is not necessary to specify the folder name during unzipping. By default, the folder created is called _package_. We can create folders named properly using the _--directory_ argument from tar.

QilongTang: @filipeotero Understood, my point is that if in the future `DynamoCoreWpf` is depending on multiple npm packages, they will all be unzipped to the `package` folder and the index.html will be overwriting each other. If possible, I would love to have the dir specified to prevent that case from happening, what do you think?

filipeotero: Makes sense. I will do that!

sm6srw:
@filipeotero I think we can get this PR ready for merging again. The BRE pipeline have been using VS2022 now for a couple of days.

filipeotero:
@zeusongit Just to let you know, this PR is ready to be merged after the checks are completed. Thanks!

PR #13608, opened 2022-12-06
DYN-4810_Dots in Preview Bubble and Watch sublists are rendered off-axis

dnenov:
### Purpose

This is a bug-fix to the [Dots in Preview Bubble and Watch sublists are rendered off-axis](https://jira.autodesk.com/browse/DYN-4810) issue. It fixes the misalignment of the dots in the sub-list node menu.

### Declarations

### Release Notes

- fix to the misalignment of the sub-list dots

### Reviewers

@reddyashish 

### FYIs

@Amoursol

### DIFF
--- a/src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
+++ b/src/DynamoCoreWpf/Views/Preview/WatchTree.xaml
@@ -1,4 +1,4 @@
-﻿<UserControl x:Class="Dynamo.Controls.WatchTree"
+<UserControl x:Class="Dynamo.Controls.WatchTree"
              xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
              xmlns:controls="clr-namespace:Dynamo.Controls"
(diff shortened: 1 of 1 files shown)


Amoursol:
@dnenov - If you zoom in to the fixed version, it looks like it's very slightly still skewed to the left. Is it possible to nudge it a tiny bit more?

dnenov:
> 

Great catch, keen eye! However, the shift that we need to do is half of a pixel, and we cannot do that. What we have here is a rectangle of 1px width and an ellipse with 5px width and height. They cannot get centrally aligned it seems and will be offset by 0.5px in either left or right direction. I spent most of the time trying to find a solution for exactly this issue, but I don't think it exists. These are dynamic lines as well, so it will be tough to replace them with an icon, which is what I also though about doing.

Amoursol:
> > 
> 
> Great catch, keen eye! However, the shift that we need to do is half of a pixel, and we cannot do that. What we have here is a rectangle of 1px width and an ellipse with 5px width and height. They cannot get centrally aligned it seems and will be offset by 0.5px in either left or right direction. I spent most of the time trying to find a solution for exactly this issue, but I don't think it exists. These are dynamic lines as well, so it will be tough to replace them with an icon, which is what I also though about doing.

Ah, gotcha. No worries in that case! I'm cool with it as is :D

reddyashish (approved):

reddyashish:
@QilongTang Do you want to get this in for 2.17?

PR #13662, opened 2023-01-06
Cross platform part2

pinzart90:
This PR does:
Added remaining projects to the dotnet6 build
Added new Pythonnet assemblies targeting netstandard2.0 (will remove when pythonnet PR is merged)
Fixed a bug with the way we search for ASM files on linux vs windows
Some small changes from @saintentropy


extern/Python/CPython_netstd20/Readme.txt
```diff
@@ -0,0 +1 @@
+These assemblies are built from https://git.autodesk.com/Dynamo/pythonnet targeting netstandard2.0
```
pinzart90, line 1: I added new assemblies for netstd20  only because I could not get tests to run on the pythonnet targeting netstd20
Once we get pythonnet CICD to run properly, we can just keep the new netstd20 assmeblies (since they are compatible with net48 and net6.0)

sm6srw: Did the test run properly for you locally? The embedded test part of CICD for PythoNet is unstable (when running in docker) and has been for a while.

sm6srw: Also, we have more changes coming into python net soon so we should resolve this sooner than later.

sm6srw (approved):
LGTM

extern/Python/CPython_netstd20/Readme.txt
pinzart90: I could not get the tests to run locally. 
Something with the nunit version is causing issues.

PR #13663, opened 2023-01-07
Fixing NodewView serializationn properties

jesusalvino:
### Purpose

Fix the bug https://jira.autodesk.com/browse/DYN-5446-serializing-dyn-file

### Declarations

### Reviewers
@QilongTang 

### FYIs
@RobertGlobant20 @filipeotero


jesusalvino:
imagen

QilongTang:
@jesusalvino Would you set the order for Node Model as well?

jesusalvino:
> @jesusalvino Would you set the order for Node Model as well?

Done.

QilongTang:
@jesusalvino Would you share the reason why for NodeModel, you did not use the sequence number as identifier but chose a different implementation?

jesusalvino:
> @jesusalvino Would you share the reason why for NodeModel, you did not use the sequence number as identifier but chose a different implementation?

Sure, because the NodeModel has serialized properties that are not directly mapped to its C# class, like the "ConcreteType" property and the NodeModel serialized properties could have extra properties based on the different kind of nodes like "Code" property or "InputValue", that's why the order of its serialized properties is handable after its json format, not in the tag order as part of its property in its C# Class like the NodeView, the order of serialization of the NodeModel property is in the NodeModel.PropertiesSerializationOrder

src/DynamoCore/Graph/Workspaces/SerializationExtensions.cs
```diff
@@ -95,5 +99,214 @@ internal static string SerializeIntegerSliderAs32BitType(string json)
...
+        /// Orders the properties of each Node according to the specify in the NodePropertiesOrder function
+        /// </summary>
+        /// <param name="metaContent"></param>
+        /// <returns></returns>
+        internal static string customOrderNodes(string metaContent)
+        {
+            try
+            {
+                string metaNodesStartKey = "\"Nodes\": [\r\n   ";
+                string metaNodesEndKey = "\r\n  ]";
```
mjkkirschner, line 115: you should use something cross platform using system environment new line or similar

src/DynamoCore/Graph/Workspaces/SerializationExtensions.cs
```diff
@@ -95,5 +99,214 @@ internal static string SerializeIntegerSliderAs32BitType(string json)
...
+                int flatNodesStartIndex = metaContent.IndexOf(metaNodesStartKey) + metaNodesStartKey.Length;
```
mjkkirschner, line 116: I think this code has the potential to be very slow, have you profiled the serialization performance of a very large graph before and after this change?

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -2265,6 +2265,11 @@ public virtual string PrintExpression()
             return s;
         }
 
+        public static List<string> PropertiesSerializationOrder()
```
mjkkirschner, line 2268: why is this public?

src/DynamoCore/Graph/Workspaces/SerializationExtensions.cs
```diff
@@ -95,5 +99,214 @@ internal static string SerializeIntegerSliderAs32BitType(string json)
...
+            {
+                return metaContent;
+            }            
+        }
+
+        internal static List<string> splitFlatNodes(string flatNodes)
+        {
+            List<string> metaProperties = new List<string>();
+
+            var rxKeyProperty = new Regex(keyProperty);
```
mjkkirschner, line 147: oh boy.

mjkkirschner:
There is a lot of complexity in this new code, I did not finish reviewing it.

I am curious how much benefit there would be to simply using `[JsonProperty(Order = n)]` and avoiding the custom serialization ordering logic all together. 

I'm also concerned about the potential performance impact - we should test it with various size graphs if we're going to proceed with this.

jesusalvino:
@QilongTang @mjkkirschner unfortunatelly I couldn't test the custom approach with a large .dyn file [enclosed here](https://jira.autodesk.com/browse/DYN-5446) , it stucks my Visual Studio when I try to debug it, so in order to keeping simple I switched to the order approach).

QilongTang (approved):

QilongTang:
@jesusalvino Glad you were able to solve this using simple order definition.

PR #13678, opened 2023-01-13
AGD-2857 : cache adp analytics optin

BogdanZavu:
### Purpose
Avoid hang or long startup times when there are issues with ADP connection.
The number of retries in Analytics.NET IsOptedIn. is too high and it takes a failed call 30s to conclude.
In my case the return code was 11 but there could be other codes as well.
On Dynamo startup there is a high number of track event calls which can easily add up to a lot of time.
I believe this is the reason for other reports that we have with Dynamo or Player starting slowly when startup sequence takes too long.
Imho one quick possible solution for this release R2024 is to interrogate once per session for IsOptedIn value and cache it. Also limit the number of retries and overall isOpt check should not tale longer than 3s or other value.
While this will prevent the analytics behavior to be changed during the same session I think we can live with that.
This is somewhat already happening when you are set to false since trackers are only registered on startup if set to true.

### Declarations

### DIFF
--- a/src/DynamoCore/Configuration/PreferenceSettings.cs
+++ b/src/DynamoCore/Configuration/PreferenceSettings.cs
@@ -55,6 +55,8 @@ namespace Dynamo.Configuration
         private bool isNotificationCenterEnabled;
         private bool isStaticSplashScreenEnabled;
         private bool isCreatedFromValidFile = true;
+        private bool isADPChecked = false;
+        private bool isADPOptedIn = false;
 
         #region Constants
         /// <summary>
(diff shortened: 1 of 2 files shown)


mjkkirschner (approved):
I'm fine with this if the potential for missing analytics is acceptable to @QilongTang.

I guess a future improvement could be update this cached value at some interval or in the background/from another thread.

src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -115,7 +117,16 @@ public class PreferenceSettings : NotificationObject, IPreferences, IRenderPreci
         [Obsolete("Setter is obsolete - ADP consent should not be set directly, it should be set using the consent dialog.")]
         public bool IsADPAnalyticsReportingApproved
         {
-            get { return Logging.AnalyticsService.IsADPOptedIn; }
+            get
+            {
```
pinzart90, line 121: pls add some comments here as to why we are only checking a single time.

QilongTang (approved):

PR #13680, opened 2023-01-16
read scientific notation

filipeotero:
### Purpose

The purpose of this PR is to fix scientific notation values that are read from Excel files.

image

### Declarations

### Reviewers

@QilongTang

### DIFF
--- a/src/Libraries/DSOffice/OpenXmlHelper.cs
+++ b/src/Libraries/DSOffice/OpenXmlHelper.cs
@@ -6,7 +6,6 @@ using System.Linq;
 using DocumentFormat.OpenXml;
 using DocumentFormat.OpenXml.Packaging;
 using DocumentFormat.OpenXml.Spreadsheet;
-using ProtoCore.AST.AssociativeAST;
 
 namespace DSOffice
 {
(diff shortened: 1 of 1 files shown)


QilongTang:
There is one reported regression, can you take a look?

filipeotero:
> There is one reported regression [DynamoMSOfficeTests.OpenXmlTests.TestFormula](https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/8200/testReport/junit/DynamoMSOfficeTests/OpenXmlTests/TestFormula/), can you take a look?

Thanks Aaron! I will also add a test containing a scientific notation example.

src/Libraries/DSOffice/OpenXmlHelper.cs
```diff
@@ -345,7 +345,7 @@ private static object GetCellValue(Cell cell, SharedStringTable sharedStringTabl
                     else
                     {
                         // This is a number
-                        if (cell.CellValue.TryGetDouble(out var value))
+                        if (double.TryParse(cell.CellValue.InnerText, NumberStyles.Any, NumberFormatInfo.InvariantInfo, out var value))
```
Mikhinja, line 348: Have we tested change in behavior when digit separator is changed (from OS settings)?

Mikhinja: Possible use case: customer distributes a spreadsheet with per-cell number formatting to at least 2 different users in 2 different locations with different OS settings relating to number formatting (digit separator). Did the previous function work for both OS settings and does the new one not work the same?

PR #13692, opened 2023-01-18
update upi config

mjkkirschner:


reddyashish (approved):

QilongTang:
I dont think release info needs to be updated, the UPI `release` is different than our understanding of `release`. You probably want to refer to previous hotfix definitions, like https://github.com/DynamoDS/Dynamo/blob/RC2.16.1_master/tools/autobuild/build.json

mjkkirschner:
ok @QilongTang thanks for pointing that out - it seems I really only need to update the build.json file?

src/Config/upiconfig.xml
```diff
@@ -8,17 +8,17 @@
         <upi_attribute name='id' value='DYN'/>
         <upi_element name='level'>
           <upi_attribute name='name' value='release'/>
-          <upi_attribute name='id' value='2.17.0'/>
+          <upi_attribute name='id' value='2.17.1'/>
```
pinzart90, line 11: release needs to remain `2.17.0`

Only the build id tag is supposed to be changed when incrementing the build part of the version

mjkkirschner:
PTAL!

pinzart90 (approved):

PR #13696, opened 2023-01-19
Defending code

jesusalvino:
### Purpose

Adding defensive code to deal the bug https://jira.autodesk.com/browse/DYN-5467

### Declarations

### Release Notes

### Reviewers
@mjkkirschner

### DIFF
--- a/src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
+++ b/src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
@@ -137,9 +137,9 @@ namespace Dynamo.WorkspaceDependency
             RestartBanner.Visibility = Visibility.Hidden;
             ws.ForceComputeWorkspaceReferences = forceCompute;
 
-            var packageDependencies = ws.NodeLibraryDependencies.Where(d => d is PackageDependencyInfo).ToList();
-            var localDefinitions = ws.NodeLocalDefinitions.Where(d => d is DependencyInfo).ToList();
-            var externalFiles = ws.ExternalFiles.Where(d => d is DependencyInfo).ToList();
+            var packageDependencies = ws.NodeLibraryDependencies?.Where(d => d is PackageDependencyInfo).ToList();
+            var localDefinitions = ws.NodeLocalDefinitions?.Where(d => d is DependencyInfo).ToList();
+            var externalFiles = ws.ExternalFiles?.Where(d => d is DependencyInfo).ToList();
 
             foreach (DependencyInfo info in localDefinitions)
             {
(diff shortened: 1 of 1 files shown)


src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -187,9 +187,16 @@ internal void DependencyRegen(WorkspaceModel ws, bool forceCompute = false)
...
+                    try
+                    {
+                        (packageDependencies.FirstOrDefault(x => x.Name == package.Name) as PackageDependencyInfo).State =
                         PackageDependencyState.RequiresRestart;
-                    hasPackageMarkedForUninstall = true;
+                        hasPackageMarkedForUninstall = true;
+                    }
+                    catch (Exception ex)
+                    {
+                        dependencyViewExtension.OnMessageLogged(LogMessage.Info(string.Format(Properties.Resources.DependencyViewExtensionErrorTemplate, ex.ToString())));
```
mjkkirschner, line 198: perhaps it would be good to note in the message what failed to happen - so If I am correct about this code, it would be something like "failure to set package uninstall state" ? - I think it's acceptable for the text to be non localized if the message is not intended for users but more for debugging for us.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -187,9 +187,16 @@ internal void DependencyRegen(WorkspaceModel ws, bool forceCompute = false)
                 foreach (var package in dependencyViewExtension.pmExtension.PackageLoader.LocalPackages.Where(x => 
                 x.LoadState.ScheduledState == PackageLoadState.ScheduledTypes.ScheduledForDeletion || x.LoadState.ScheduledState == PackageLoadState.ScheduledTypes.ScheduledForUnload))
                 {
-                    (packageDependencies.FirstOrDefault(x => x.Name == package.Name) as PackageDependencyInfo).State =
```
mjkkirschner, line 190: I notice above that `ws` is never checked for being null.

mjkkirschner: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and-

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
```diff
@@ -200,10 +207,17 @@ internal void DependencyRegen(WorkspaceModel ws, bool forceCompute = false)
...
                 foreach (PackageDependencyInfo packageDependencyInfo in packageDependencies)
                 {
-                    var targetInfo = pmExtension.PackageLoader.LocalPackages.Where(x => x.Name == packageDependencyInfo.Name).FirstOrDefault();
-                    if (targetInfo != null)
+                    try
+                    {
+                        var targetInfo = pmExtension.PackageLoader.LocalPackages.Where(x => x.Name == packageDependencyInfo.Name).FirstOrDefault();
+                        if (targetInfo != null)
+                        {
+                            packageDependencyInfo.Path = targetInfo.RootDirectory;
```
mjkkirschner, line 215: could it cause problems for this to be left null? should it be to set to empty string or something? I don't actually know if it's null - just something to consider.

jesusalvino: it could be null

imagen

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
jesusalvino: Done.

src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs
jesusalvino: Done.

mjkkirschner (approved):

PR #13697, opened 2023-01-20
SPIKE: coercing arrays using linq functions

aparajit-pratap:
### Purpose

This PR attempts at emitting collection coercion code using Linq expressions. E.g. conversion from long[] to IEnumerable<double>:
```
var arr = new long[] { x, y, z };
var coerced_arr = arr.Select(Convert.ToDouble);  
var res = Math.Sum(coerced_arr);    // Math.Sum() takes in an IEnumerable<double>
```
The emitted MSIL for the above conversion looks like:
```
ldnull      
ldftn       System.Convert.ToDouble
newobj  System.Func<System.Int64,System.Double>..ctor
call         System.Linq.Enumerable.Select<Int64,Double>
call        ToList or ToArray
```

Blocker:
~~I'm seeing the returned `IEnumerable` as a `WhereSelectArrayIterator ` when it is wrapped in an `object[]` and passed to `ReplicationLogic`, in other words when it's wrapped in a `CLRStackValue`.~~

Performance measurements:
It was found that even for a large array of 100 million elements (array_coercion.dyn), there was a negligible change (in compilation time) but a significant change in (execution) performance between using the linq approach and the current coercion using for loops.

Before:
image

After:
image

Observations:
As seen above, even though this approach doesn't offer any performance benefits, it reduces the complexity of the codegen logic and that of the MSIL being emitted. Previously there were errors reported in the MSIL that was being generated for coercion (using `for ` and `foreach` loops), which although can be fixed, is hard to do. Now the emitted code is so simple and concise that the decompiler doesn't report any errors in the generated MSIL when decompiling the resulting assembly.
Note that the overall compilation and execution times are almost comparable in most cases except for the last graph, which contains an array of 100 million elements, in which case the execution time is more than double what it was before.
In summary, it can be said that in most common workflows, this approach will not have any significant positive or negative impact but for certain, isolated cases involving large arrays could have a performance hit. Being able to generate error-free MSIL is an added advantage as it avoids more serious errors that can accumulate with large graphs.

TODO:
The last call to `ToList` or `ToArray ` was found to be required only in the case of replicated calls to workaround the `WhereSelectArrayIterator` issue. The `WhereSelectIterator ` is returned from the call to `Enumerable.Select` and didn't pose to be an issue for direct function calls (non-replicated) case. There are 2 things that still could be figured out here:
- Why is this iterator returned instead of an `IEnumerable<T>`?
- Can we somehow use this iterator to still properly marshal it to a `CLRStackValue `in the replication case and thereby achieve some sort of deferred execution?

### Declarations

### Release Notes

### Reviewers

### FYIs

### DIFF
--- a/src/Engine/EmitMSIL/CodeGenIL.cs
+++ b/src/Engine/EmitMSIL/CodeGenIL.cs
@@ -305,18 +305,6 @@ namespace EmitMSIL
             var genericToList = toListMethod.MakeGenericMethod(typeof(Target));
             EmitOpCode(OpCodes.Call, genericToList);
             return typeof(List<Target>);
-
-            //if (typeof(IList).IsAssignableFrom(ienumerableParamType) ||
-            //    typeof(IList<Target>).IsAssignableFrom(ienumerableParamType))
-            //{
-            //    var requiredType = typeof(Target);
-            //    var toListMethod = typeof(Enumerable).GetMethod(nameof(Enumerable.ToList));
-            //    mInfo = toListMethod.MakeGenericMethod(requiredType);
-
-            //    EmitOpCode(OpCodes.Call, mInfo);
-            //    return typeof(List<Target>);
-            //}
-
         }
 
         private IEnumerable<ProtoCore.CLRFunctionEndPoint> FunctionLookup(IList args)
--- a/src/Engine/EmitMSIL/CodeGenIL.cs
+++ b/src/Engine/EmitMSIL/CodeGenIL.cs
@@ -6,10 +6,8 @@ using ProtoCore.Properties;
 using ProtoCore.Utils;
 using ProtoFFI;
 using System;
-using System.CodeDom;
 using System.Collections;
 using System.Collections.Generic;
-using System.Configuration.Assemblies;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
@@ -145,13 +143,11 @@ namespace EmitMSIL
                     DfsTraverse(ast);
                 }
                 EmitOpCode(OpCodes.Ret);
+
                 return (asm, type);
 #if DEBUG
             }
 #endif
-            EmitOpCode(OpCodes.Ret);
-
-            return (asm, type);
         }
 
         // Given a double value on the stack, emit call to Math.Round(arg, 0, MidpointRounding.AwayFromZero);
--- a/src/Engine/EmitMSIL/CodeGenIL.cs
+++ b/src/Engine/EmitMSIL/CodeGenIL.cs
@@ -6,6 +6,7 @@ using ProtoCore.Properties;
 using ProtoCore.Utils;
 using ProtoFFI;
 using System;
+using System.CodeDom;
 using System.Collections;
 using System.Collections.Generic;
 using System.Configuration.Assemblies;
@@ -159,7 +160,8 @@ namespace EmitMSIL
         {
             EmitOpCode(OpCodes.Ldc_I4_0);
             EmitOpCode(OpCodes.Ldc_I4_1);
-            var roundMethod = typeof(Math).GetMethod(nameof(Math.Round), new[] { typeof(double), typeof(int), typeof(MidpointRounding) });
+            var roundMethod = typeof(Math).GetMethod(nameof(Math.Round),
+                new[] { typeof(double), typeof(int), typeof(MidpointRounding) });
             EmitOpCode(OpCodes.Call, roundMethod);
         }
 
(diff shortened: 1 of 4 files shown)


mjkkirschner:
@aparajit-pratap - an initial question - what is up with the 280~ test failures, it looks like they are related to ASM 229 loading failures?

aparajit-pratap:
@mjkkirschner I'm not concerned about the PR checks right now until this branch is up-to-date with master. I'm hoping that they'll go away once this branch is updated.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -418,117 +255,56 @@ private Type EmitCoercionCode(AssociativeNode arg, Type argType, Type param)
...
-            {
-                EmitOpCode(OpCodes.Ldelem_I8);
-            }
-            else if (typeof(Source) == typeof(int))
+            EmitOpCode(OpCodes.Ldnull);
+            MethodInfo convertMethod = null;
+            if (typeof(Target) == typeof(int))
             {
-                EmitOpCode(OpCodes.Ldelem_I4);
+                convertMethod = typeof(Convert).GetMethod(nameof(Convert.ToInt32), new[] { typeof(Source) });
```
mjkkirschner, line 269: perhaps to make compilation faster we could cache these method lookups at compiler startup time instead of looking them up each time we go through this function.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -418,117 +255,56 @@ private Type EmitCoercionCode(AssociativeNode arg, Type argType, Type param)
...
-                EmitOpCode(OpCodes.Stelem_R8);
+                convertMethod = typeof(Convert).GetMethod(nameof(Convert.ToBoolean), new[] { typeof(Source) });
             }
-            // i++;
-            EmitOpCode(OpCodes.Ldloc, counterIndex);
-            EmitOpCode(OpCodes.Ldc_I4_1);
+            EmitOpCode(OpCodes.Ldftn, convertMethod);
 
-            EmitOpCode(OpCodes.Add);
+            var ctor = System.Linq.Expressions.Expression.GetDelegateType(typeof(Source), typeof(Target))
```
mjkkirschner, line 285: so this is creating a delegate that returns the target type and takes the source?
I see that this requires referencing codeDom... and linq expression trees - do you think this additional dependency will cause issues for us? Is it supported in .net 6/7/8?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -418,117 +255,56 @@ private Type EmitCoercionCode(AssociativeNode arg, Type argType, Type param)
...
+                .GetConstructors()[0];
+            EmitOpCode(OpCodes.Newobj, ctor);
             
-            EmitOpCode(OpCodes.Stloc, counterIndex);
-
-            // i < arr.Length;
-            MarkLabel(loopCondLabel.Value,"label:cond");
-
-            EmitOpCode(OpCodes.Ldloc, counterIndex);
+            var selectMethod = typeof(Enumerable).GetMember(nameof(Enumerable.Select)).OfType<MethodInfo>().First();
```
mjkkirschner, line 289: another opportunity for cacheing reflection.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -418,117 +255,56 @@ private Type EmitCoercionCode(AssociativeNode arg, Type argType, Type param)
...
 
-            // Load input array
-            EmitOpCode(OpCodes.Ldloc, currentVarIndex);
-            EmitOpCode(OpCodes.Ldlen);
-            EmitOpCode(OpCodes.Conv_I4);
+            var genericSelect = selectMethod.MakeGenericMethod(typeof(Source), typeof(Target));
```
mjkkirschner, line 291: if we find that something like this is slow, we could likely precompute many of the common versions.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -418,117 +255,56 @@ private Type EmitCoercionCode(AssociativeNode arg, Type argType, Type param)
...
-            EmitOpCode(OpCodes.Ldloc, newArrIndex);
-
-            if(typeof(List<Target>).IsAssignableFrom(ienumerableParamType))
+            // If param is an array
+            if (typeof(Array).IsAssignableFrom(ienumerableParamType))
             {
                 var requiredType = typeof(Target);
-                var toListMethod = typeof(Enumerable).GetMethod(nameof(Enumerable.ToList));
-                var mInfo = toListMethod.MakeGenericMethod(requiredType);
+                var toArrayMethod = typeof(Enumerable).GetMethod(nameof(Enumerable.ToArray));
```
mjkkirschner, line 298: another cache opportunity.

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -418,117 +255,56 @@ private Type EmitCoercionCode(AssociativeNode arg, Type argType, Type param)
...
+                var genericToArray = toArrayMethod.MakeGenericMethod(requiredType);
 
-                EmitOpCode(OpCodes.Call, mInfo);
-                t = typeof(List<Target>);
+                EmitOpCode(OpCodes.Call, genericToArray);
+                return typeof(Target[]);
             }
-            return t;
+            var toListMethod = typeof(Enumerable).GetMethod(nameof(Enumerable.ToList));
```
mjkkirschner, line 304: " "

mjkkirschner:
the reduction in code complexity is great! 

I do think we could get better compiler performance by cacheing all of the reflection calls.

src/Engine/EmitMSIL/CodeGenIL.cs
aparajit-pratap: This doesn't require any special dependencies. I just removed the `using System.Codedom` as it wasn't referenced anywhere in this file.

test/CodeGenILTests/RangeMicroTests.cs
```diff
@@ -9,49 +9,6 @@
 
 namespace CodeGenILTests
 {
-    public class MicroTests
```
aparajit-pratap, line 12: Refactored this class into its own file.

aparajit-pratap:
@mjkkirschner I'll incorporate your function caching suggestions in a follow-up PR.

mjkkirschner:
@aparajit-pratap so do you intend to cherry pick this to emitMSIL-v2? I guess once we close all current prs we should delete emitMSIL? Or make it readonly?

sm6srw:
Yes, please make it read only and keep it around for a while when we are done. Just in case.

PR #13693, opened 2023-01-19
Markdown procedural image location

dnenov:
### Purpose

A small PR with changes to the NodeDocumentationHtmlGenerator. Now identifies the location of the  tag inside markdown help files and correctly injects HTML code for that row.

### Declarations

Check these if you believe they are true

- [x] The codebase is in a better state after this PR
- [x] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [x] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.
- [ ] This PR modifies some build requirements and the readme is updated

### Release Notes

- fixed hard-coded image location
- now identifies image location

### Reviewers

@reddyashish 
@RobertGlobant20 

### FYIs

@Amoursol

PR #13414, opened 2022-10-20
DocumentationBrowserViewExtension Upgrade + Insert

dnenov:
### Purpose

This PR covers the work done in two directions:

- DocumentationBrowserViewExtension Upgrade
- Insert Functionality (new File Menu Item)

####
DocumentationBrowserViewExtension

- Update the UI to modernize and align with the current design language of Dynamo
- New functionality - interactive image panel with zoom in, zoom out and pan function
- New functionality - insert dynamo example graph associated with node help information
- New functionality - Warning notification message displays in the Documentation Browser UI

general usage_part 1
general usage_part 2
general usage_part 3

####
Insert

- New File Menu Item - Dynamo now allows users to insert .dyn or .dyf (to confirm) in the current workspace environment. 

insert

### Declarations

Check these if you believe they are true

- [x] The codebase is in a better state after this PR
- [x] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [x] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.
- [ ] This PR modifies some build requirements and the readme is updated

### Release Notes

(FILL ME IN) Brief description of the fix / enhancement. **Mandatory section** 

### Reviewers

@reddyashish 

### FYIs

@Amoursol

PR #13703, opened 2023-01-23
Dyn 5497 markdown not loading

RobertGlobant20:
### Purpose

Fixing bug about markdown files not loading for node packages documentation.
There was a code change that was preventing for loading md files for packages node documentation, so after this fix was done in a different pull request (https://github.com/DynamoDS/Dynamo/pull/13693), now it is loading the md file but the images were not loading due that WebView2 was using virtual folder fixed to fallback_docs folder then in this fix we are allowing to create the virtual folder dynamic depending if is a dynamo node or package node.
For some nodes of the Orchid package is working there is another bug with some specific nodes (like Math.Addition).

### Declarations

### Release Notes

Fixing bug about markdown files not loading for node packages documentation.

### Reviewers

@QilongTang @reddyashish @dnenov


RobertGlobant20:
Screenshot showing the documentation with images for a package node.
image

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
```diff
@@ -132,6 +140,8 @@ public bool ShowBrowser
 
         internal UIElement DynamoView { get; set; }
 
+        internal NodeInfo CurrentNodeInfo { get; set; }
```
mjkkirschner, line 143: I don't think carrying around this state is ideal - at the minimum it should have some comments...

I'm not sure if it's possible, but it seems simpler to keep track of where this data is used if it's passed in an event or binding (or something similar) to the View.

Just a thought.

mjkkirschner:
Can we write a test for this regression?

mjkkirschner:
what does this mean `For some nodes of the Orchid package is working there is another bug with some specific nodes (like Math.Addition).` Is it tracked and in jira?

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
RobertGlobant20: @mjkkirschner removed the NodeInfo class and I just added a new internal property, check the next commit:
https://github.com/DynamoDS/Dynamo/pull/13703/commits/fdd337c2dd455d53ae51bbd88c17e8a5056ca389

RobertGlobant20:
> Can we write a test for this regression?

@mjkkirschner please confirm that this a list of steps for the test expected for validating this functionality.
With this test we will be only validating that the md file expected will be loaded (no validation on images loaded).
- Open Dynamo
- Install a package with documentation (note that most of the packages currently doesn't have documentation even the Orchid package in production doesn't have documentation).
- Open the DocumentationBrowser
- Validate that the md in the Package/doc folder match with the one loaded in the DocumentationBrowser.

mjkkirschner:
> > Can we write a test for this regression?
> 
> @mjkkirschner please confirm that this a list of steps for the test expected for validating this functionality. With this test we will be only validating that the md file expected will be loaded (no validation on images loaded).
> 
> * Open Dynamo
> * Install a package with documentation (note that most of the packages currently doesn't have documentation even the Orchid package in production doesn't have documentation).
> * Open the DocumentationBrowser
> * Validate that the md in the Package/doc folder match with the one loaded in the DocumentationBrowser.

what about mocking most of that - for example, just set the view model data with some fake data that should come from a package, and then check that the virtual folder is setup correctly using that data - we just want to ensure it does not regress later.

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
```diff
@@ -89,6 +90,22 @@ private set
         private string content;
         private string name;
 
+
+        /// <summary>
+        /// Package Name
+        /// </summary>
+        internal string Name
```
filipeotero, line 97: Maybe a more intuitive name would be PackageName... What do you think?

zeusongit: Or graph name?

RobertGlobant20: @filipeotero yes, I think is a better idea to use PackageName, then updated in the next commit: https://github.com/DynamoDS/Dynamo/pull/13703/commits/0e3c928e47fc6f87b9d35cbe9b58b41d47318eda

RobertGlobant20:
> > > Can we write a test for this regression?
> > 
> > 
> > @mjkkirschner please confirm that this a list of steps for the test expected for validating this functionality. With this test we will be only validating that the md file expected will be loaded (no validation on images loaded).
> > 
> > * Open Dynamo
> > * Install a package with documentation (note that most of the packages currently doesn't have documentation even the Orchid package in production doesn't have documentation).
> > * Open the DocumentationBrowser
> > * Validate that the md in the Package/doc folder match with the one loaded in the DocumentationBrowser.
> 
> what about mocking most of that - for example, just set the view model data with some fake data that should come from a package, open the docs browser, and then check that the virtual folder is setup correctly using that data - we just want to ensure it does not regress later.

@mjkkirschner I've added a unit test for validating that the WebView2 Virtual Directory is created in the right place (so images will be loaded correctly) when displaying documentation for package nodes.

When implementing the unit test I found that the call to await WebView2.EnsureCoreWebView2Async() was returning the execution to the unit test (instead of executing the next code), then I had to move the Virtual Folder code creation before EnsureCoreWebView2Async() call also I've added a property for VirtualDirectory so it can be checked in the unit test., please let me know is the test should be enough 
Thanks

Here is the commit with the changes:
https://github.com/DynamoDS/Dynamo/pull/13703/commits/0e3c928e47fc6f87b9d35cbe9b58b41d47318eda

src/DocumentationBrowserViewExtension/DocumentationBrowserView.xaml.cs
```diff
@@ -130,6 +133,21 @@ protected virtual void Dispose(bool disposing)
 
         async void InitializeAsync()
         {
+            VirtualFolderPath = string.Empty;
+            try
+            {
+                if (viewModel.Link != null && !string.IsNullOrEmpty(viewModel.PackageName))
+                {
+                    VirtualFolderPath = HttpUtility.UrlDecode(viewModel.Link.AbsolutePath).Replace(viewModel.PackageName + ".md", "");
```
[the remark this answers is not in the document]
RobertGlobant20: thanks @pinzart90 I've done the change in the next commit: https://github.com/DynamoDS/Dynamo/pull/13703/commits/ad5209f44adda39a2cd3e0f619a5991c479e3c5a

src/DocumentationBrowserViewExtension/DocumentationBrowserView.xaml.cs
```diff
@@ -130,6 +133,21 @@ protected virtual void Dispose(bool disposing)
...
+            try
+            {
+                if (viewModel.Link != null && !string.IsNullOrEmpty(viewModel.PackageName))
+                {
+                    VirtualFolderPath = Path.GetDirectoryName(HttpUtility.UrlDecode(viewModel.Link.AbsolutePath));
+                }
+                else
+                    VirtualFolderPath = FallbackDirectoryName;
+            }
+            catch (Exception ex)
```
mjkkirschner, line 146: what exception do you think will be caught here?

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
```diff
@@ -87,7 +88,23 @@ private set
...
         private string graphPath;
         private string content;
-        private string name;
+        private string packageName;
+
+
+        /// <summary>
+        /// Package Name
+        /// </summary>
+        internal string PackageName
```
mjkkirschner, line 97: it still feels pretty odd to me that the DocumentationBrowserViewModel has a single piece of state like this that it has to hold on to.

If I am understanding correctly, can we at least name it something like:
`CurrentPackageName` or something like that and add a comment describing when it will be null and when it will be valid?

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -2120,9 +2120,9 @@ internal AssemblyName GetNameOfAssemblyReferencedByNode(NodeModel node)
         {
             AssemblyName assemblyName = null;
             // Get zerotouch assembly
-            if (node is DSFunction)
+            if (node is DSFunction || node is DSVarArgFunction)
```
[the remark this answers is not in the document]
mjkkirschner: I am incorrect, `DSVarArgFunction` inherits from `DSFunctionBase` not `DSFunction` - apologies for the confusion.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -2120,9 +2120,9 @@ internal AssemblyName GetNameOfAssemblyReferencedByNode(NodeModel node)
...
             {
-                var descriptor = (node as DSFunction).Controller.Definition;
+                var descriptor = node is DSFunction function ? function.Controller.Definition :  (node as DSVarArgFunction).Controller.Definition;
```
mjkkirschner, line 2125: I think you can get rid of all these cases and the if statement above - just check if the type is `DSFunctionBase` - it has a `Controller.Definition` property you can access and you don't need to check any types, it's safer an faster - unless you need special behavior for each type.

test/DynamoCoreWpfTests/ViewExtensions/DocumentationBrowserViewExtensionTests.cs
```diff
@@ -103,6 +103,48 @@ public void ShowingStartPageHidesBrowser()
...
+                     Guid.NewGuid().ToString(), nodeName, 0, 0, false, false)
+                 );
+
+            var node = this.ViewModel.Model.CurrentWorkspace.Nodes.FirstOrDefault();
+            node.Name = nodeRename; // Forces original name header to appear 
+            var nodeAnnotationEventArgs = new OpenNodeAnnotationEventArgs(node, this.ViewModel);
+
+            docBrowserviewExtension.HandleRequestOpenDocumentationLink(nodeAnnotationEventArgs);
+            var htmlContent = GetSidebarDocsBrowserContents();
+            htmlContent = htmlContent.Replace(@"%5C", "/");
```
mjkkirschner, line 138: please add a comment

mjkkirschner (changes requested):
please see inline comments

src/DocumentationBrowserViewExtension/DocumentationBrowserView.xaml.cs
RobertGlobant20: @mjkkirschner I found a case that `viewModel.Link = "{}"` so it was crashing when trying the next line `viewModel.Link.AbsolutePath` so I added this try catch because I didn't found any other way to validate the `viewModel.Link` content.
With the fix added related to DSVarArgFunction now I expect that viewModel.Link always will contain something valid but just in case I think we should keep this code or if you have any other idea in how to validate  `viewModel.Link = "{}"` please let me know (Link is Uri type, Uri.IsWellFormedUriString returns true).

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
RobertGlobant20: @mjkkirschner notice that the dependency on PackageName was added in the next PR:
https://github.com/DynamoDS/Dynamo/pull/13414
Renamed to CurrentPackageName in the next commit: https://github.com/DynamoDS/Dynamo/pull/13703/commits/f00db5b919450b0ccdafac4c9f90a8827b5d80c1
image

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
RobertGlobant20: Checking if is DSFunctionBase type I've simplified the code in the next way:
image
Please let me know if it sounds good to you. Thanks
commit: https://github.com/DynamoDS/Dynamo/pull/13703/commits/f00db5b919450b0ccdafac4c9f90a8827b5d80c1

test/DynamoCoreWpfTests/ViewExtensions/DocumentationBrowserViewExtensionTests.cs
RobertGlobant20: Added comment, please check the next commit: https://github.com/DynamoDS/Dynamo/pull/13703/commits/f00db5b919450b0ccdafac4c9f90a8827b5d80c1

src/DocumentationBrowserViewExtension/DocumentationBrowserView.xaml.cs
mjkkirschner: I think it's a good practice to log something including the exception if possible to the dynamo console - if some unanticipated exception starts showing up here atleast we'll know.

mjkkirschner (approved):

mjkkirschner:
thanks @RobertGlobant20 !

src/DocumentationBrowserViewExtension/DocumentationBrowserView.xaml.cs
RobertGlobant20: @mjkkirschner I've added some code for logging the exception in the dynamo console, see the next commit: https://github.com/DynamoDS/Dynamo/pull/13703/commits/5980d151914322d4a1eb6d6965f5e2b116cfc9a4

zeusongit:
Thanks for addressing the review comments @RobertGlobant20 I will merge the PR as soon as the self serve job completes

RobertGlobant20:
> Thanks for addressing the review comments @RobertGlobant20 I will merge the PR as soon as the self serve job completes
@zeusongit 
Just to comment that I keep receiving this emails after I push any commit in this PR (but all the checks seems that are passing):
image

mjkkirschner:
@RobertGlobant20 seems there is a conflicted file.

PR #13712, opened 2023-01-27
Support more direct function call cases - emitv2

mjkkirschner:
### Purpose

move https://github.com/DynamoDS/Dynamo/pull/13436 to emitMSIL-v2

This PR begins to address https://jira.autodesk.com/browse/DYN-5321, but we'll keep this open as there are remaining known cases which are still failing. Unclear how to fix some of these at the moment and they require more discussion.

1. PR attempts to unify custom compile time will replicate logic with DFSTraversal which is already pretty good at determining compile time type information. This ends up simplifying the `WillReplicate` checks. Uses two compile passes to achieve this.
2. When determining if call will replicate also includes the selected `FEP` as a parameter so we can check what attributes each parameter has - these are useful when determining if a FEP will replicate - for example the `arbitraryRankArrayImport` attribute should make it so a parameter is not replicated over except with specific replication guides.
3. PR attempts to replace `ArrayTypeStatistics` logic in `EmitExprListNode` method with DFSTraversal - simplifies that code and uses two compile passes to achieve this.
4. introduces `StrictDirectFunctionCallValidation` on the `MSILRuntime` and CodeGenIL` objects, if these bools are set true exceptions will be thrown when direct function calls should made, but can't be. If these values are false we'll fallback to replicated calls.

### Declarations

### Release Notes

### Reviewers

### FYIs


mjkkirschner:
passed here:
https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/8281
PTAL!

pinzart90 (approved):
LGTM

sm6srw (approved):

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1384,21 +1366,48 @@ private void EmitUnmarshalFunctionArgs(List<AssociativeNode> args, List<Type> pa
...
+        /// <param name="parameterTypes"></param>
+        /// <param name="args"></param>
+        /// <returns></returns>
+        /// <exception cref="Exception"></exception>
+        private bool DoParamsRankMatchArgs(ProtoCore.CLRFunctionEndPoint fep, List<Type> parameterTypes, List<AssociativeNode> args)
         {
+            //TODO this can throw because it's possible that the FEP is incorrect
+            //and if this is the case we should throw or fallback to replication -
+            //in either case log something - we need compile time overload support.
+            if (parameterTypes.Count != args.Count)
```
aparajit-pratap, line 1383: Are the counts for these always expected to be the same irrespective of whether the method is static or instance?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1412,44 +1421,97 @@ private bool DoesParamArgRankMatch(List<Type> parameterTypes, List<AssociativeNo
...
+                // If both have rank 0, it could also be because their type info is ambiguous.
+                // or it could be fine!
+                if (argRank == 0 && paramRank == 0)
+                {
+                    EmitILComment("param and arg are both rank 0, type info may be ambiguous.");
+                    if (argType == typeof(DSASM.CLRStackValue))
+                    {
+                        EmitILComment("argument is from a replicate call, and param is rank 0, fallback to replication.");
+                        return false;
+                    }
```
aparajit-pratap, line 1499: What are the cases, outside of this `if`? Is it the case where both arg and param are reference types and have rank 0? If so, shouldn't we let replication handle fitting the arg to the param type in this case?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1412,44 +1421,97 @@ private bool DoesParamArgRankMatch(List<Type> parameterTypes, List<AssociativeNo
...
+        /// This function besides checking rank directly, also does checks using the FEP attributes.
+        /// </summary>
+        /// <param name="fep">the function end point wrapper</param>
+        /// <param name="paramType">type of param on the fep</param>
+        /// <param name="argType">actual argument type</param>
+        /// <param name="paramIndex">parameter index in list of params to fep.</param>
+        /// <returns></returns>
+        private bool DoesParamRankMatchArg(ProtoCore.CLRFunctionEndPoint fep,Type paramType, Type argType, int paramIndex)
+        {
+            if (!paramType.Equals(argType))
```
aparajit-pratap, line 1469: Minor but it would've been easier to read and less indentation if you inverted this `if`:
```
if(paramType.Equals(argType)) return true;
...
```

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1266,14 +1255,7 @@ private Type EmitFunctionCallNode(AssociativeNode node)
...
                     {
                         Type t = DfsTraverse(arg);
-                        if (isStaticOrCtor)
-                        {
-                            t = EmitCoercionCode(arg, t, parameters[index]);
-                        }
-                        else if (index > 0)
-                        {
-                            t = EmitCoercionCode(arg, t, parameters[index - 1]);
-                        }
```
aparajit-pratap, line 1276: Why?

src/Engine/ProtoCore/Lang/FunctionEndPoint.cs
```diff
@@ -61,6 +61,9 @@ public System.Type CLRReturnType
             }
         }
 
+        
+        internal Attribute[][] ParamAttributes => method.GetParameters().Select(x => x.Info.GetCustomAttributes().ToArray()).ToArray();
```
aparajit-pratap, line 65: Where is this used?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -1412,44 +1421,97 @@ private bool DoesParamArgRankMatch(List<Type> parameterTypes, List<AssociativeNo
...
+                        return false;
+                    }
+                }
+            }
+            return true;
+        }
+
+        private static int GetParamRank(ProtoCore.CLRFunctionEndPoint fep, Type type, int paramIndex)
+        {
+            if (fep.procedureNode.ArgumentTypes.ElementAt(paramIndex).rank == DSASM.Constants.kArbitraryRank)
```
aparajit-pratap, line 1507: Is this equivalent to checking for the `ArbitraryDimensionArrayImport` attribute on the parameter? If so, do we need the `ParamAttributes` property defined in the `CLRFunctionEndPoint` class?

aparajit-pratap (commented):
@mjkkirschner sorry, I missed reviewing this earlier. Have a couple of questions.

src/Engine/ProtoCore/Lang/FunctionEndPoint.cs
mjkkirschner: it's not currently - we could get rid of it as you suggested, I instead used the `procedureNode.ArgumentTypes`... I can send a cleanup PR for this.

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: yes - we can remove the `ParamAttributes` for now, we can always add it back if we decide to reduce our use of protocore type system.

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: not necessarily, either the param or the arg could also be value types. 
I cannot recall which, but making this change allowed some tests start passing.

I think until we find a case where this incorrectly calls a function we should allow this - we may even want to emit a try catch and use it to fallback to replication at runtime.

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: In my tests I found that the number of parameters and arguments were always the same even in the case of static methods.

This caused a test failure, unfortunately I cannot remember which one, but if you revert this change I believe you'll see one.

src/Engine/EmitMSIL/CodeGenIL.cs
mjkkirschner: this is a good question - 
initially I did not think so, but then regarding this:
https://github.com/DynamoDS/Dynamo/pull/13712#discussion_r1096028816 it seems that the number of args/params are always the same...

I did find this code which seems to imply unifying the number of params:
https://github.com/DynamoDS/Dynamo/blob/emitMSIL-V2/src/Engine/EmitMSIL/CodeGenIL.cs#L417

src/Engine/EmitMSIL/CodeGenIL.cs
aparajit-pratap: But we're doing the same over here: https://github.com/DynamoDS/Dynamo/blob/2d53bb411af05a4921a98fcba4bf9f18c960daf0/src/Engine/EmitMSIL/CodeGenIL.cs#L1180

PR #13724, opened 2023-02-03
attempt to fix flaky ui test

mjkkirschner:
attempts to fix intermittent test failures with `DynamoCoreWpfTests.CoreUserInterfaceTests.WorkspaceContextMenu_TestIfSearchTextClearsOnOpeningContextMenu`

could not reproduce locally so this is just a guess - but it should not have negative consequences in any case.


RobertGlobant20 (approved):
LGTM, very weird that the test fail is not reproducible, I guess the context menu is not still opened when reaching the Assert part (I think Popup.IsOpen = true is async).

sm6srw (approved):

PR #13725, opened 2023-02-03
Do not force user login for package search operation

zeusongit:
### Purpose

Do not initiate a login process when user opens the package search window.

### Declarations

### Reviewers

@DynamoDS/dynamo 

### FYIs

@Amoursol

### DIFF
--- a/src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
@@ -740,6 +740,9 @@ namespace Dynamo.PackageManager
 
         internal void RefreshInfectedPackages()
         {
+            //do not call a protected route, if user is not logged in already.
+            if (string.IsNullOrEmpty(PackageManagerClientViewModel.Username)) return;
+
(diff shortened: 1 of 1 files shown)


src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
```diff
@@ -740,6 +740,9 @@ public void RefreshAndSearchAsync()
 
         internal void RefreshInfectedPackages()
         {
+            //do not call a protected route, if user is not logged in already.
+            if (string.IsNullOrEmpty(PackageManagerClientViewModel.Username)) return;
```
mjkkirschner, line 744: wouldn't it be more idiomatic to check if the user is logged in? Or does that call actually start a login process as well?

mjkkirschner:
It might be hard to test, but this seems quite fragile to regress.

src/DynamoCoreWpf/ViewModels/PackageManager/PackageManagerSearchViewModel.cs
zeusongit: switched to `LoginState`

PR #13731, opened 2023-02-07
New CLI options to run MSIL engine

aparajit-pratap:
### Purpose

New CLI options to run MSIL engine - added a `runmode` command line option to specify the following options:
1. compiling and executing with the new engine
2. compile only  - this builds a dynamic assembly from a graph and saves it to disk
3. execute only - this is added to execute a graph given the location to its compiled assembly

Note that any of the above options can be run only in combination with the `eng` command line option set to `false`.

TODOs:
The execute-only option is a placeholder currently, and still needs to be tested.

This task prompted the creation of these 2 follow-up tasks to have fewer dependencies on the old VM (ProtoCore):
https://jira.autodesk.com/browse/DYN-5595
https://jira.autodesk.com/browse/DYN-5596 
https://jira.autodesk.com/browse/DYN-5608

### Declarations

### Release Notes

### Reviewers

### FYIs


src/DynamoCLI/CommandLineRunner.cs
```diff
@@ -48,11 +48,6 @@ private static XmlDocument RunCommandLineArgs(DynamoModel model, StartupUtils.Co
 
             model.HostAnalyticsInfo = cmdLineArgs.AnalyticsInfo;
 
-            cmdLineArgs.ImportedPaths.ToList().ForEach(path =>
-            {
-                ImportAssembly(model, path);
-            });
```
aparajit-pratap, line 54: This is already taking place in the call to `DynamoCLI.Program.StartupDynamo`.

src/DynamoCore/Library/LibraryServices.cs
```diff
@@ -193,8 +193,11 @@ private void PreloadLibraries(IEnumerable<string> preloadLibraries)
         {
             importedLibraries.AddRange(preloadLibraries);
 
-            foreach (var library in importedLibraries)
-                CompilerUtils.TryLoadAssemblyIntoCore(LibraryManagementCore, library);
+            var importNodes = preloadLibraries.Select(lib => new ImportNode { ModuleName = lib });
+            var codeGen = new CodeGenDS(importNodes);
+            string code = codeGen.GenerateCode();
+
+            CompilerUtils.TryLoadAssembliesIntoCore(LibraryManagementCore, code);
```
aparajit-pratap, line 200: Unrelated to this task but this function loads all path manager libraries together instead of one-by-one, which was what was happening earlier. This is now consistent with the loading of libraries in LiveRunner core.

src/DynamoApplications/StartupUtils.cs
```diff
@@ -224,8 +226,8 @@ public static CommandLineArguments Parse(string[] args)
                 .Add("pi=|PI=|parentId", "Identify Dynamo host analytics parent id", pi => parentId = pi)
                 .Add("da|DA|disableAnalytics", "Disables analytics in Dynamo for the process liftime", da => disableAnalytics = da != null)
                 .Add("cr=|CR=|cerLocation", "Specify the crash error report tool location on disk ", cr => cerLocation = cr)
-                .Add("eng=|ENG=|Eng=|engine=", "true if using the old DesignScript engine, false if using .NET execution of graphs", eng => dsExecutionEngine = !(eng != null && eng.ToLower() == "false"))
-                ;
+                .Add("eng=|ENG=|Eng=|engine=", "True if using the old DesignScript engine, false if using .NET execution of graphs", eng => dsExecutionEngine = !(eng != null && eng.ToLower() == "false"))
+                .Add("rm=|RM=|runMode=", "0 for Compile and Execute mode, 1 for compile-only mode, 2 for execution-only mode for running new engine", rm => msilRunMode = int.Parse(rm) == 0 ? DynamoModel.RunMode.CompileAndExecute : int.Parse(rm) == 1 ? DynamoModel.RunMode.CompileOnly : DynamoModel.RunMode.ExecuteOnly);
```
mjkkirschner, line 230: instead of just parsing the string to an int... what about also trying to parse it as an enum directly?
It could be nice for users who were looking just at the client code that invokes the CLI.

`DynamoCLI.exe -eng false -rm 0` ... I won't know what this is doing without using the help docs or finding this source code - but 
`DynamoCLI.exe -eng false -rm CompileAndExecute` 
gives me a fighting chance.

mjkkirschner: also this help doc should probably mention that it's only valid when `eng` is false.

src/DynamoCLI/CommandLineRunner.cs
```diff
@@ -61,8 +56,10 @@ private static XmlDocument RunCommandLineArgs(DynamoModel model, StartupUtils.Co
             // If the desire is to have additional states you can add logic here to build
             // up a list and iterate through each state in the list using the loop below.
             // This must be done after potentially loading states from an external file.
-            var stateNames = new List<String>();
-            stateNames.Add("default");
+            var stateNames = new List<String>
```
mjkkirschner, line 59: wow, I have not seen this code in a long time - you can get rid of this state nonsense, it was for setting presets, back when those existed.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1491,10 +1514,13 @@ private void InitializeNodeLibrary()
             foreach (var type in migrationTypes)
                 MigrationManager.AddMigrationType(type);
 
-            // Import Zero Touch libs
-            var functionGroups = LibraryServices.GetAllFunctionGroups();
             if (!IsTestMode)
+            {
```
mjkkirschner, line 1518: ?

src/Engine/EmitMSIL/CodeGenIL.cs
```diff
@@ -19,6 +19,29 @@ namespace EmitMSIL
...
     {
+        /// <summary>
+        /// Run modes. 
+        /// </summary>
+        public enum RunMode
+        {
+            /// <summary>
+            /// Compile graph and execute compiled assembly in memory.
+            /// </summary>
+            CompileAndExecute,
```
mjkkirschner, line 30: do we really need two of these enums? Is the intention to split the internal and API layer?

src/Engine/ProtoCore/Core.cs
```diff
@@ -770,7 +770,11 @@ public void GenerateExecutable()
                 BfsBuildInstructionStreams(CodeBlockList[n], DSExecutable.instrStreamList);
             }
 
-            GenerateExprExe();
+            if (Options.RunMode == InterpreterMode.Expression)
```
mjkkirschner, line 773: ?

mjkkirschner (approved):

src/Engine/EmitMSIL/CodeGenIL.cs
aparajit-pratap: DynamoApplication, DynamoCore , and EmitMSIL projects need access to this enum and not all of them reference each other.

src/Engine/ProtoCore/Core.cs
aparajit-pratap: `GenerateExprExe()` pertains to the expression interpreter, which was like a watch window for DesignScript debugging during the text IDE days for DS. It's legacy code now, needs to be removed at some point.

src/DynamoCore/Models/DynamoModel.cs
aparajit-pratap: The `functionGroups ` variable is used only inside the `if` condition, we needn't do extra work if it doesn't enter the `if` condition in the first place.

src/DynamoApplications/StartupUtils.cs
aparajit-pratap: > 

I was thinking about this and I felt that CLI users would need to type the exact strings for this to work, e.g. "compileandexecute", etc. I can change it, however, as it improves readability in exchange for verbosity.

pinzart90:
An alternative (as mentioned in the jira task) was to create an executable from the graph. 
But I guess that requires more work/refactor since we still need type information (among other things) to already be processed and loaded.
@aparajit-pratap @mjkkirschner do you imagine any major differences/target workflow between the 2 options ?

aparajit-pratap:
> An alternative (as mentioned in the jira task) was to create an executable from the graph (instead of a cli that load the graph as a dll). But I guess that requires more work/refactor since we still need type information (among other things) to already be processed and loaded. @aparajit-pratap @mjkkirschner do you imagine any major differences/target workflow between the 2 options ? I guess the management of dependencies would be one...

@pinzart90 yes, you're right, even for the execute-only mode for the CLI, we will ideally need to load the precompiled graph assembly (generated at compile time) together with loading some state (namely the type information in MSILRuntimeCore and the method) that is serialized/deserialized from a separate file. I will file a separate task for this. For either option (CLI or graph executable), this last task is necessary.

PR #13751, opened 2023-02-15
Insert unique GUIDs update

dnenov:
### Purpose

This PR addresses an outstanding issue with the Insert functionality introduced a while back. Previously, it wasn't possible for users to insert the same graph or derivatives of the same graph inside the workspace. That was due to the identical GUIDs present in the graphs (attached current state of affairs comparison between 3 derivative dynamo graphs). 

Logic to recreate all element GUIDs that make up the workspace prior to inserting them into the current workspace is introduced with this PR to handle these cases.

#### Side-by-side derivative comparison
Insert

#### Insert multiple of the same
insert_multiple

### Declarations

### Release Notes

- assign new GUIDs for all workspace elements
- now correctly inserts groups and notes
- now correctly inserts pinned notes
- added new tests for Insert functionality
- change to old test to account for the ability to insert multiple of the same graph

### Reviewers

@QilongTang 
@mjkkirschner 
@Amoursol 

### FYIs

@sm6srw

### DIFF
--- a/src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
+++ b/src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
@@ -2530,7 +2530,7 @@ namespace Dynamo.Graph.Workspaces
                 var text = annotationViewInfo.Title;
 
                 var pinnedNode = this.Nodes.
-                    FirstOrDefault(x => x.GUID.ToString() == annotationViewInfo.PinnedNode);
+                    FirstOrDefault(x => x.GUID.ToString("N") == annotationViewInfo.PinnedNode);
 
                 NoteModel noteModel;
                 if (offsetX == 0.0 && offsetY == 0.0)
--- a/src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
+++ b/src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
@@ -2530,7 +2530,7 @@ namespace Dynamo.Graph.Workspaces
                 var text = annotationViewInfo.Title;
 
                 var pinnedNode = this.Nodes.
-                    FirstOrDefault(x => x.GUID.ToString("N") == annotationViewInfo.PinnedNode);
+                    FirstOrDefault(x => x.GUID.ToString() == annotationViewInfo.PinnedNode);
 
                 NoteModel noteModel;
                 if (offsetX == 0.0 && offsetY == 0.0)
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -1971,6 +1971,9 @@ namespace Dynamo.Models
                         {
                             ExtraWorkspaceViewInfo viewInfo = ExtraWorkspaceViewInfoFromJson(fileContents);
 
+                            // Update (assign new) GUIDs for each node, connection, note and group
+                            UpdateWorkspaceGUIDs(ref viewInfo, ref ws);
+
                             InsertWorkspace(ws, viewInfo);
                         }
                     }
@@ -1981,6 +1984,98 @@ namespace Dynamo.Models
             {
                 Console.WriteLine(e.Message);
                 throw e;
+            }
+        }
+
+        private void UpdateWorkspaceGUIDs(ref ExtraWorkspaceViewInfo viewInfo, ref WorkspaceModel ws)
+        {
+            var nodes = ws.Nodes;
+            var connectors = ws.Connectors;
+            var notes = viewInfo?.Annotations;
+
+            Dictionary<Guid, Guid> nodesGuids = new Dictionary<Guid, Guid>();
+            Dictionary<Guid, Guid> portsGuids = new Dictionary<Guid, Guid>();
+            Dictionary<Guid, Guid> connectorsGuids = new Dictionary<Guid, Guid>();
+            Dictionary<Guid, Guid> connectorPinsGuids = new Dictionary<Guid, Guid>();
+            Dictionary<string, string> notesGuids = new Dictionary<string, string>();
+
+            // Assign new GUIDs
+            foreach (var node in nodes)
+            {
+                var newGuid = Guid.NewGuid();
+                nodesGuids[node.GUID] = newGuid;
+                node.GUID = newGuid;
+
+                foreach (var port in node.InPorts)
+                {
+                    newGuid = Guid.NewGuid();
+                    portsGuids[port.GUID] = newGuid;
+                    port.GUID = newGuid;
+                }
+
+                foreach (var port in node.OutPorts)
+                {
+                    newGuid = Guid.NewGuid();
+                    portsGuids[port.GUID] = newGuid;
+                    port.GUID = newGuid;
+                }
+            }
+
+            foreach (var connector in connectors)
+            {
+                var newGuid = Guid.NewGuid();
+                connectorsGuids[connector.GUID] = newGuid;
+                connector.GUID = newGuid;
+
+                // Assign new pin GUIDs
+                foreach (var pin in connector.ConnectorPinModels)
+                {
+                    newGuid = Guid.NewGuid();
+                    connectorPinsGuids[pin.GUID] = newGuid;
+                    pin.GUID = Guid.NewGuid();
+                }
+            }
+
+            // Assign new annotation models new GUIDs
+            // TODO: What about the stuff that were grouped?!
+            // Or attached notes?
+            foreach (var note in notes)
+            {
+                string newGuid = Guid.NewGuid().ToString();
+                notesGuids[note.Id] = newGuid;
+                note.Id = newGuid;
+            }
+
+            if (viewInfo == null) return;
+
+            // Now propagate these changes to the viewInfo object
+            if (viewInfo.NodeViews != null)
+            {
+                foreach (var node in viewInfo.NodeViews)
+                {
+                    node.Id = nodesGuids[new Guid(node.Id)].ToString();
+                }
+            }
+
+            //foreach (var annotation in viewInfo.Annotations)
+            //{
+            //    annotation.Id = notesGuids[annotation.Id];
+            //}
+
+            if (viewInfo.Notes != null)
+            {
+                foreach (var note in viewInfo.Notes)
+                {
+                    note.Id = notesGuids[note.Id];
+                }
+            }
+
+            if (viewInfo.ConnectorPins != null)
+            {
+                foreach (var pin in viewInfo.ConnectorPins)
+                {
+                    pin.ConnectorGuid = connectorsGuids[new Guid(pin.ConnectorGuid)].ToString();
+                }
             }
         }
 
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -1892,7 +1892,6 @@ namespace Dynamo.Models
             }
         }
 
-
         static private DynamoPreferencesData DynamoPreferencesDataFromJson(string json)
         {
             JsonReader reader = new JsonTextReader(new StringReader(json));
--- a/src/DynamoCore/Models/DynamoModel.cs
+++ b/src/DynamoCore/Models/DynamoModel.cs
@@ -1986,175 +1986,7 @@ namespace Dynamo.Models
                 throw e;
             }
         }
-
-        /// <summary>
-        /// Update inserted workspace elements GUIDs 
-        /// </summary>
-        /// <param name="viewInfo">viewInfo</param>
-        /// <param name="ws">workspace</param>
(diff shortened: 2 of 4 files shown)


src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -2119,7 +2121,7 @@ private void InsertWorkspace(WorkspaceModel ws, ExtraWorkspaceViewInfo viewInfo
 
             CurrentWorkspace.UpdateWithExtraWorkspaceViewInfo(viewInfo, offsetX, offsetY);
 
-            InsertAnnotations(viewInfo.Annotations, offsetX, offsetY);
+            //InsertAnnotations(viewInfo.Annotations, offsetX, offsetY);
```
QilongTang, line 2124: Just want to confirm this is intended, if yes, let's delete this.

dnenov: Sure thing! I was testing out running the functionality without it, and it seems we don't need this part. Thank you for reminding me to clean up, all done now.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -2530,7 +2530,7 @@ private void LoadNotesFromAnnotations(IEnumerable<ExtraAnnotationViewInfo> annot
                 var text = annotationViewInfo.Title;
 
                 var pinnedNode = this.Nodes.
-                    FirstOrDefault(x => x.GUID.ToString("N") == annotationViewInfo.PinnedNode);
+                    FirstOrDefault(x => x.GUID.ToString() == annotationViewInfo.PinnedNode);
```
QilongTang, line 2533: Curious why we remove the formatting? Did you find it does not matter in this case and the default format is "N" anyway?

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
```diff
@@ -2530,7 +2530,7 @@ private void LoadNotesFromAnnotations(IEnumerable<ExtraAnnotationViewInfo> annot
                 var text = annotationViewInfo.Title;
 
                 var pinnedNode = this.Nodes.
-                    FirstOrDefault(x => x.GUID.ToString("N") == annotationViewInfo.PinnedNode);
```
mjkkirschner, line 2533: why make this change?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -3531,6 +3519,177 @@ private bool NotesAlreadyLoaded(IEnumerable<ExtraAnnotationViewInfo> notes)
...
         #endregion
 
+        #region Insert Private Methods
+
+        /// <summary>
+        /// Update inserted workspace elements GUIDs 
+        /// </summary>
+        /// <param name="viewInfo">viewInfo</param>
+        /// <param name="ws">workspace</param>
+        private void UpdateWorkspaceGUIDs(ref ExtraWorkspaceViewInfo viewInfo, ref WorkspaceModel ws)
```
mjkkirschner, line 3529: I see no need for the `ref` keyword, you are passing classes, and changes you make will  be reflected on the classes - why do you think you need to use `ref`?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -3531,6 +3519,177 @@ private bool NotesAlreadyLoaded(IEnumerable<ExtraAnnotationViewInfo> notes)
...
+            var connectorsGuids = UpdateConnectorGUIDs(ref connectors);
+            var notesGuids = UpdateNoteGUIDs(nodesGuids, ref notes);
+
+            // Now propagate these changes to the viewInfo object
+            if (viewInfo == null) return;
+
+            PropagateUpdatesToViewInfo(viewInfo, nodesGuids, notesGuids, connectorsGuids);
+        }
+
+        private Dictionary<string, string> UpdateNoteGUIDs(Dictionary<Guid, Guid> nodesGuids, ref IEnumerable<ExtraAnnotationViewInfo> notes)
```
mjkkirschner, line 3545: again, I see no need for `ref` - please explain.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -3531,6 +3519,177 @@ private bool NotesAlreadyLoaded(IEnumerable<ExtraAnnotationViewInfo> notes)
...
+        {
+            Dictionary<string, string> notesGuids = new Dictionary<string, string>();
```
mjkkirschner, line 3547: please use `var` in all cases.

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -3531,6 +3519,177 @@ private bool NotesAlreadyLoaded(IEnumerable<ExtraAnnotationViewInfo> notes)
...
+        #region Insert Private Methods
+
+        /// <summary>
+        /// Update inserted workspace elements GUIDs 
+        /// </summary>
+        /// <param name="viewInfo">viewInfo</param>
+        /// <param name="ws">workspace</param>
+        private void UpdateWorkspaceGUIDs(ref ExtraWorkspaceViewInfo viewInfo, ref WorkspaceModel ws)
+        {
+            var nodes = ws.Nodes;
```
mjkkirschner, line 3531: check for ws being null as well?

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -3531,6 +3519,177 @@ private bool NotesAlreadyLoaded(IEnumerable<ExtraAnnotationViewInfo> notes)
...
+        }
+
+        /// <summary>
+        /// Update GUIDs inside ExtraWorskapceViewInfo
+        /// </summary>
+        /// <param name="viewInfo">ViewInfo object containing information allowing to recreate the workspace</param>
+        /// <param name="nodesGuids">Nodes [old,new] GUID values</param>
+        /// <param name="notesGuids">Annotations [old,new] GUID values</param>
+        /// <param name="connectorsGuids">Connectors [old,new] GUID values</param>
+        private void PropagateUpdatesToViewInfo(ExtraWorkspaceViewInfo viewInfo,
```
QilongTang, line 3641: For this function, it seems we are only replacing Guids, I feel the function maybe can be simplified further. After some deeper look, it seems these elements do not share same interface, we cant really write utility method to avoid repeating code. So good for now.

QilongTang (approved):

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -1971,6 +1970,9 @@ private bool InsertJsonFileFromPath(string fileContents, string filePath, bool f
                         {
                             ExtraWorkspaceViewInfo viewInfo = ExtraWorkspaceViewInfoFromJson(fileContents);
 
+                            // Update (assign new) GUIDs for each node, connection, note and group
```
QilongTang, line 1973: would you also add comments as to why for future reference?

QilongTang:
LGTM with couple comments

src/DynamoCore/Models/DynamoModel.cs
```diff
@@ -3531,6 +3519,177 @@ private bool NotesAlreadyLoaded(IEnumerable<ExtraAnnotationViewInfo> notes)
...
+            PropagateUpdatesToViewInfo(viewInfo, nodesGuids, notesGuids, connectorsGuids);
+        }
+
+        private Dictionary<string, string> UpdateNoteGUIDs(Dictionary<Guid, Guid> nodesGuids, ref IEnumerable<ExtraAnnotationViewInfo> notes)
+        {
+            Dictionary<string, string> notesGuids = new Dictionary<string, string>();
+
+            // Assign new annotation models new GUIDs
+            // We need two separate runs to first create the new GUIDs and second work with the new GUIDS
+            if (notes != null)
```
mjkkirschner, line 3551: I find this entire approach a bit complicated...

I am just curious what you think of an approach that modifies the json. I would usually shy away from that approach but you could select all IDS easily BEFORE deserialization or as jtokens and transform them...

I think it would be much less code.

test/DynamoCoreTests/Models/OpenFileCommandTest.cs
```diff
@@ -116,12 +116,53 @@ public void TestInsertNodesCount()
...
-            // TODO - this part of the test might change if we include a logic
-            // to recreate the GUIDs of the 'same' nodes to allow for duplicate or partial duplicate insertions
+            //Assert 2x
             this.CurrentDynamoModel.InsertFileFromPath(wspath);
             var updatedCount = this.CurrentDynamoModel.CurrentWorkspace.Nodes.Count();
-            Assert.AreEqual(currentCount, updatedCount);
+            Assert.AreEqual(currentCount * 2, updatedCount);
+        }
+
+        [Test]
```
mjkkirschner, line 125: do these tests cover ALL your new methods? It would be good to cover each new method you add with a unit test.

src/DynamoCore/Models/DynamoModel.cs
dnenov: I was thinking the same, but they are ever so slightly different to make me not do it. They do share the same base class, but that class does not have the GUID we are after. Also, the GUID is sometimes a Guid, sometimes a string.

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
dnenov: I honestly don't know why the formatting was there to start with, and how it was functioning before. With the formatting, the code was breaking because it wasn't finding the necessary Node the Note was pinned to.

test/DynamoCoreTests/Models/OpenFileCommandTest.cs
dnenov: There are 3 things to worry about, the way I see it:

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
dnenov: Reverted back after the update. I believe that the extra Guid assignment I was doing was the reason the formatting didn't function. In any case, this change is now removed.

src/DynamoCore/Models/DynamoModel.cs
dnenov: I did not know that's how passing classes worked, Michael! I feel ashamed but also thankful for this comment, I will remember that.

All of this code is now removed in favor of a much simpler method that operates directly on the json string before deserialization, as you prompted.

src/DynamoCore/Models/DynamoModel.cs
dnenov: Yup, got it!

src/DynamoCore/Models/DynamoModel.cs
dnenov: Yes! Great call and while I didn't get the idea immediately, I tested it today and it works great and simplifies this whole thing. [Committed here](https://github.com/DynamoDS/Dynamo/pull/13751/commits/d4cc43095260d490fe875a402be0c4f4d693e9c9).

dnenov: Let me know if that doesn't look good to you and I will rework as needed.

src/DynamoCore/Models/DynamoModel.cs
dnenov: Got it!

src/DynamoCore/Graph/Workspaces/WorkspaceModel.cs
dnenov: Reverted

src/DynamoCore/Models/DynamoModel.cs
dnenov: Updated, let me know if it is still vague or not clear.

src/DynamoCore/Models/DynamoModel.cs
dnenov: All code is now reworked in favor of a much simpler method that operates directly on the json string, as discussed on Friday. It works great to on my side, but let me know if that's not ideal for whatever reason and I can update as needed.

src/DynamoCore/Models/DynamoModel.cs
QilongTang: @mjkkirschner Previously I feel the approach was OK also because we have another task to update Guids while Dynamo users save the current workspace as a new one, then a JSON block of the graph is not guaranteed to be there for a simplified approach. But after second thought, if this approach is preferred, maybe we can do the same for save as  - touch the JSON instead of in-memory workspace representation.

test/DynamoCoreTests/Models/OpenFileCommandTest.cs
QilongTang: I will follow up on save as in a new PR and add test as well

PR #13752, opened 2023-02-15
add net60 linux nuget deployment nuspec pt1.

mjkkirschner:
### Purpose

To support the internal net60 linux deployment job: https://git.autodesk.com/Dynamo/DynamoSelfServe/tree/net60_linux_target-release, and perhaps eventually a public deploy.

* adds a github action that builds the net6.sln on a linux runner.
* builds portable pdbs when the platform is net6
* does not attempt to invoke resgen to pack image/customization dlls when platform is net6 - resgen is not supported in new version of dotnet.
* adds a new nuspec built by the linked internal job.

⚠️ 
### Please note I will be sending a followup to this PR attempting to introduce basic smoketests for the net6 solution. I don't want to leave the deployment untested, but I would like to merge this so the job will start posting nugets without manually assigning a branch to build.

I'd also like to investigate modifying the .nuspec so it uses both `/content` and `/lib` as it appears that might make this package a bit more idiomatic.

### Declarations


.github/workflows/dynamoNet6.0_linux_build.yml
```diff
@@ -0,0 +1,37 @@
...
+on: [push,pull_request]
+jobs:
+ build:
+  runs-on: ubuntu-latest
+  steps:
+    - name: Checkout Dynamo Repo
+      uses: actions/checkout@v2
+      with:
+        path: Dynamo
+        repository: mjkkirschner/Dynamo
```
aparajit-pratap, line 12: ?

aparajit-pratap (approved):

src/Config/CS_SDK.props
```diff
@@ -23,16 +23,27 @@
     <ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>None</ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>
     <DebugSymbols>true</DebugSymbols>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)' == 'Debug'">
-    <DebugType>full</DebugType>
+  <PropertyGroup Condition="'$(Configuration)' == 'Debug' AND $(Platform.Contains('NET60')) ">
+    <DebugType>portable</DebugType>
```
pinzart90, line 27: I think you could just add
```
<DebugType Condition="$(Platform.Contains('NET60')) ">portable</DebugType>
<DebugType Condition="!$(Platform.Contains('NET60')) ">full</DebugType>
```
If you think this is more readable

src/DynamoCore.net6.sln
```diff
@@ -93,6 +93,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Watch3DNodeModels", "Librar
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Migrations", "Migrations\Migrations.csproj", "{06B9E5B0-7C50-4351-9D88-E159DC25755F}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Tests", "Tests", "{15CCC76A-F02A-4EFA-97A2-51A1343E7457}"
```
pinzart90, line 96: what is this ? an empty project ?
Is this for dotnet6 ? maybe specify core or cross platform or something in the name ?

pinzart90 (approved):
A couple of questions
then LGTM

.github/workflows/dynamoNet6.0_linux_build.yml
mjkkirschner: nice catch - that let me test this github action against *this* code - but it will be wrong when I merge it.

src/Config/CS_SDK.props
mjkkirschner: yes, it's nicer, also considering that Debug or Release changes nothing, maybe I can move these elements outside the debug and release prop groups and not repeat them.

src/DynamoCore.net6.sln
mjkkirschner: it's just a folder, I will remove it for now.

PR #13756, opened 2023-02-16
AGD-9999 : allow extension to load in service mode

BogdanZavu:
### Purpose
Allow extension to load in service mode.

### Declarations

### Reviewers
@QilongTang @saintentropy


zeusongit:
I thought we were going to disallow all extensions, and only white-list the ones that we want to use, like Player or PM, right? @QilongTang

QilongTang:
> I thought we were going to disallow all extensions, and only white-list the ones that we want to use, like Player or PM, right? @QilongTang

That was my original plan, but given that we are going to deploy our own build so that we have full control, plus currently there is only one default extension (view extensions are not going to be loaded), this is probably fine to start with. What do you think?

zeusongit:
> > I thought we were going to disallow all extensions, and only white-list the ones that we want to use, like Player or PM, right? @QilongTang
> 
> That was my original plan, but given that we are going to deploy our own build so that we have full control, plus currently there is only one default extension (view extensions are not going to be loaded), this is probably fine to start with. What do you think?

Alright, makes sense 👍🏼

QilongTang (approved):

PR #13766, opened 2023-02-21
Support Net6 target for DynamoCoreTests.dll (WIP for feedback)

mjkkirschner:
### Purpose

This PR starts to move tests in DynamoCoreTests to .NET6.

It's still a WIP - but I think it's ready for feedback.

does the following:
* even though tests now build on linux- testing on github action is still blocked - I filed https://github.com/microsoft/vstest/issues/4330
*  removes duplicate `-g` for geometryPath / geometry mode in CLI and moves to `Geometry` only
*  adds `DisableAnalytics` to unify tests that use CLI.
* adds DynamoCoreTests,TestServices,DSOffice and FFITarget .csprojs to the DynamoCoreNet6.sln.
* unifies on a new version of `Moq` which can be referenced from NET6 - 
    * ⚠️  this brings in new dependencies to the bin folder - ~~I have tried to filter them out during the `install.sln` so they are not delivered~~  These are now copied to the `test/test_dependencies` folder and resolvers added to look here during testing. 
    * Existing issue - I have noticed that some test binaries are already being included since at least 2.8 - thats as far back as I went - `(FFITarget,Moq, Nunit.Framework)` - it appears that exclusions need to be added to the robo copy step in `dynamoinstall.csproj` - hopefully we come up with a more holistic solution before the next GL. https://jira.autodesk.com/browse/DYN-5673
    * updates use of moq to apis in new version.
* make DSOffice.csproj conditional so it does not reference or build exel interop nodes for net6 (open xml is built)
* conditionally updates DynamoCoreTests.csproj to `Nunit3` for NET6
    * ⚠️  - serialization tests fail when comparing dictionaries now - I've created a test only comparison func that seems to work using sequence equals on keys and vals - only for DynamoDictionaries.
    * use `#ifdefs` to move to new `NUnit3`  asserts and attributes.
    *   marks some tests as `Category("FailureNET6")]` - some of these can be fixed with more complex refactoring so we can file a followup.
* DynamoCoreTests had some transitive WPF references - attempted to refactor the easy ones by moving tests to DynamoCoreWFPtests.csproj.
* AssemblyResolve behavior has changed between .net48 and .net6 - runtime better handles resolvers that throw exceptions. (this required test changes)
* unify on `Microsoft.NET.Test.Sdk` 17 between both solutions.
* skip a few migration tests that check excel and legacy unit nodes when targeting .net6.
* Moves PythonMigration tests to DynamoPythonTest project
* TypeForwards `TestExtensions` from `SystemTestServices` to `TestServices` so it can be used in DynamoCoreTests without transitive WPF dep.
* 

### Declarations

### Release Notes

Support Net6 target for DynamoCoreTests.dll

### Reviewers


src/DynamoCore/Properties/AssemblyInfo.cs
```diff
@@ -50,5 +50,7 @@
 [assembly: InternalsVisibleTo("GraphNodeManagerViewExtension")]
 [assembly: InternalsVisibleTo("ExportSampleImagesViewExtension")]
 [assembly: InternalsVisibleTo("DocumentationBrowserViewExtension")]
+[assembly: InternalsVisibleTo("Net60SmokeTests")]
```
mjkkirschner, line 53: mistake - will remove

src/Libraries/DesignScriptBuiltin/Dictionary.cs
```diff
@@ -129,6 +129,13 @@ public override string ToString()
                 result.Append("}");
                 return result.ToString();
             }
+
+            //TODO this could cause unanticipated consequences -
```
mjkkirschner, line 133: I think this is not safe - ideas for alternatives - maybe something like `TestTimeEquals`?

src/DynamoCore.net6.sln
```diff
@@ -93,6 +93,16 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Watch3DNodeModels", "Librar
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Migrations", "Migrations\Migrations.csproj", "{06B9E5B0-7C50-4351-9D88-E159DC25755F}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Tests", "Tests", "{5B9B5B6B-0BA7-4606-B8E5-70C958346D57}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "TestServices", "..\test\Libraries\TestServices\TestServices.csproj", "{6CD0F0CF-8199-49F9-B0EA-0B9598B44419}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DSOffice", "Libraries\DSOffice\DSOffice.csproj", "{9B4FDC96-E2F9-4B8F-894A-4294405D50E7}"
```
pinzart90, line 100: I cannot tell, but did you add dsoffice to the Linux part of the solution?

src/DynamoInstall/Release.xsl
```diff
@@ -47,6 +47,10 @@
     <xsl:key name="moq-search" match="wix:Component[contains(wix:File/@Source, 'Moq')]" use="@Id"/>
     <xsl:template match="wix:Component[key('moq-search', @Id)]" />
 
+    <!--Exclude Caste.Core.dll-->
+    <xsl:key name="castle-search" match="wix:Component[contains(wix:File/@Source, 'Castle.Core')]" use="@Id"/>
+    <xsl:template match="wix:Component[key(castle-search', @Id)]" />
+
```
pinzart90, line 53: What is this about?

pinzart90: oh..the Moq dependency.
I've tried before to solve the issue of test dependencies in the final output path but I could not find a one size fits all solution.
Maybe the easiest way is to put everything in there, then just filter out (but I know some file escaped in the past and we got stuck with them)

src/Libraries/DSOffice/DSOffice.csproj
```diff
@@ -10,11 +10,13 @@
     <AssemblyName>DSOffice</AssemblyName>
     <DocumentationFile>$(OutputPath)\$(UICulture)\DSOffice.xml</DocumentationFile>
   </PropertyGroup>
+    <ItemGroup Condition=" '$(TargetFramework)' == 'net48' ">
```
pinzart90, line 13: I understand why we would avoid Excel on linux but why are we limiting it to netframework only ?
Does it not work on dotnet6+ ?

pinzart90: Here is something I tried before
https://github.com/DynamoDS/Dynamo/pull/12716/files#diff-a9d51e372ddc4731d6c88a4bfbb2dcd01b5dfa880e0d34d34078f09fd4aaf9a7R30

I think I also added a note about this on the wiki

pinzart90: https://www.nuget.org/packages/Microsoft.Office.Interop.Excel#supportedframeworks-body-tab

src/Libraries/DesignScriptBuiltin/Dictionary.cs
pinzart90: what is the problem that required this solution?

src/VisualizationTests/WpfVisualizationTests.csproj
```diff
@@ -16,7 +16,7 @@
     <PackageReference Include="HelixToolkit" Version="2.17.0" />
     <PackageReference Include="HelixToolkit.Wpf" Version="2.17.0" />
     <PackageReference Include="HelixToolkit.Wpf.SharpDX" Version="2.17.0" />
-    <PackageReference Include="Moq" Version="4.2.1507.0118" />
+    <PackageReference Include="Moq" Version="4.18.4" ExcludeAssets="none" />
```
pinzart90, line 19: moq should be excluded here.
I think only DynamoCoreTests should have ExcludeAssets="none"

src/Libraries/DesignScriptBuiltin/Dictionary.cs
mjkkirschner: see above in description:
```
 serialization tests fail when comparing dictionaries now - I've added an Equals override for our Dictionary type - but I think this is quite risky as it can change runtime behavior - probably best to have some kind of test only equals method that we can call from serialization - appreciate some thoughts here. I think this is just a change in assert behavior in Nunit.
 ```

File: src/DynamoApplications/StartupUtils.cs [hunk 7]
```diff
@@ -223,10 +223,9 @@ public static CommandLineArguments Parse(string[] args)
                 .Add("hn=|HN=|hostname", "Identify Dynamo variation associated with host", hn => hostname = hn)
                 .Add("si=|SI=|sessionId", "Identify Dynamo host analytics session id", si => sessionId = si)
                 .Add("pi=|PI=|parentId", "Identify Dynamo host analytics parent id", pi => parentId = pi)
-                .Add("da|DA|disableAnalytics", "Disables analytics in Dynamo for the process liftime", da => disableAnalytics = da != null)
+                .Add("da|DA|disableAnalytics|DisableAnalytics", "Disables analytics in Dynamo for the process liftime", da => disableAnalytics = da != null)
```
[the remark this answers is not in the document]
mjkkirschner: I don't think so - it only adds an uppercase version of the flag to match the upper case version that already exists for Net6.

src/DynamoCore.net6.sln
mjkkirschner: I added it for NET6 in general - so yes. DSOffice also includes openxml and CSV nodes.

src/DynamoInstall/Release.xsl
mjkkirschner: yeah, I mention a few in the pr description - I don't intend to solve it here, but I don't want to make it worse.

src/Libraries/DSOffice/DSOffice.csproj
mjkkirschner: I think I will leave this for another time and file a followup if thats ok.

src/VisualizationTests/WpfVisualizationTests.csproj
mjkkirschner: @pinzart90 can you explain why? I was thinking it would be simpler if these package references all looked the same when possible but open to other approaches for sure.

Do you think it's easier to control if only a single reference brings in the binaries so it's easy to see what project is bringing them in?

src/DynamoCore.net6.sln
```diff
@@ -93,6 +93,16 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Watch3DNodeModels", "Librar
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Migrations", "Migrations\Migrations.csproj", "{06B9E5B0-7C50-4351-9D88-E159DC25755F}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Tests", "Tests", "{5B9B5B6B-0BA7-4606-B8E5-70C958346D57}"
```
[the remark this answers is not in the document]
aparajit-pratap: Ah, I see it's just the category/folder.

src/Libraries/DesignScriptBuiltin/Dictionary.cs
aparajit-pratap: Why not just have a test-only method in this class or in the test framework for now that we use only for tests instead of using Nunit.Assert?

src/DynamoCore.net6.sln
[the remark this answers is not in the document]
mjkkirschner: this is just a solution folder called `Tests`

src/Libraries/DesignScriptBuiltin/Dictionary.cs
mjkkirschner: @aparajit-pratap I will give that a shot - since the issue only happens for dictionaries, I'd like to avoid writing custom comparers for all our types... will take a look if we can inject our own comparer into nunit.

test/DynamoCoreTests/NodeMigrationTests.cs
```diff
@@ -119,7 +118,8 @@ public void TestMigration_File_Directory()
         {
             TestMigration("TestMigration_File_Directory.dyn");
         }
-
+        //TODO_MSIL pull csv nodes into their own file with a partial class?
```
aparajit-pratap, line 121: MSIL?

test/DynamoCoreTests/UnitsOfMeasureTests.cs
```diff
@@ -1045,7 +1044,7 @@ public void ParseExpressionIncompatible()
         {
             Assert.That(() => { DynamoUnits.Utilities.ParseExpression("3mm"); }, Throws.Exception);
         }
-        [Test, Category("UnitTests")]
+        [Test, Category("UnitTests"), Category("FailureNET6")]
```
aparajit-pratap, line 1047: why does this fail for net6?

test/DynamoCoreTests/SerializationTests.cs
```diff
@@ -293,7 +293,8 @@ public static void CompareWorkspaceModels(serializationTestUtils.WorkspaceCompar
                     // When values are geometry, sometimes the creation
                     // of the string representation for forming this message
                     // fails.
-                    Assert.AreEqual(valueA, valueB,
+
+                    Assert.That(valueA, Is.EqualTo(valueB).Using<Dictionary>(DynamoDictionaryEquality),
```
aparajit-pratap, line 297: 👍

aparajit-pratap (approved):

test/DynamoCoreTests/NodeMigrationTests.cs
mjkkirschner: thanks, should be NET_6

test/DynamoCoreTests/UnitsOfMeasureTests.cs
mjkkirschner: there is an access violation in forge units - I think @pinzart90 already found it but I'm not sure if theres a task to fix it. @pinzart90 can you update?

src/DynamoApplications/StartupUtils.cs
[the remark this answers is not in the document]
pinzart90: oh..I did not see it was an addition

src/VisualizationTests/WpfVisualizationTests.csproj
mjkkirschner: well... that sounds reasonable, especially if we will move towards a solution like that. 
I'm not sure it will help for nunit framework / ms test.sdk or the dependencies we are already stuck with in the bin folder - but I will follow that advice for now and only copy from DynamoCore to make it simpler to undo later.

src/DynamoInstall/Release.xsl
mjkkirschner: I ended up continuing with your approach - `Moq`'s new deps are copied to the test_dependencies folder. In addition `Moq` is also copied to the bin folder since we cannot remove it until a GL release.

I've also added a bunch of `SetupFixtures` that hook up the resolver you added for other test assemblies that started failing now that the new moq deps are not in the bin folder.

src/VisualizationTests/WpfVisualizationTests.csproj
mjkkirschner: So I've done this - kind of - now actually none of the moq package references are `excludeAssets="none"` - instead I'm doing manual copy steps as I don't want to pull the transitive deps into the bin folder.

This is kind of a bummer, I wonder if long term it might make sense to setup a different .sln for test projects - OR just set test projects to have a completely different output path.

mjkkirschner: something I hope we can look at in:
https://jira.autodesk.com/browse/DYN-5673
FYI @jasonstratton

src/VisualizationTests/WpfVisualizationTests.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
```
pinzart90, line 1: how are we running  tests for a wpf project ?
Or what is this change about ?
I see there is also a new Setup.cs file ? why ?

test/DynamoCoreTests/DynamoCoreTests.csproj
```diff
@@ -147,12 +147,33 @@
...
-  <Target Name="CopyDisableADPTestDependencies" AfterTargets="Build">
-    <Copy SourceFiles="$(PkgMicrosoft_Diagnostics_Runtime)\lib\net461\Microsoft.Diagnostics.Runtime.dll" DestinationFolder="$(TestDependenciesPath)" />
-    <Copy SourceFiles="$(PkgSystem_Collections_Immutable)\lib\net461\System.Collections.Immutable.dll" DestinationFolder="$(TestDependenciesPath)" />
-    <Copy SourceFiles="$(PkgSystem_Reflection_Metadata)\lib\net461\System.Reflection.Metadata.dll" DestinationFolder="$(TestDependenciesPath)" />
-    <Copy SourceFiles="$(PkgSystem_Runtime_CompilerServices_Unsafe)\lib\net461\System.Runtime.CompilerServices.Unsafe.dll" DestinationFolder="$(TestDependenciesPath)" />
-    <Copy SourceFiles="$(PkgSystem_Memory)\lib\net461\System.Memory.dll" DestinationFolder="$(TestDependenciesPath)" />
-    <Copy SourceFiles="$(PkgSystem_Buffers)\lib\net461\System.Buffers.dll" DestinationFolder="$(TestDependenciesPath)" />
+<Target Name="CopyDisableADPTestDeps">
+    <!--ms diag runtime + deps -->
+    <Message Importance="High" Text="This is a test" />
```
pinzart90, line 152: probably not needed anymore

test/DynamoCoreTests/DynamoCoreTests.csproj
```diff
@@ -147,12 +147,33 @@
...
+    <Copy SourceFiles="$(PkgMicrosoft_Diagnostics_Runtime)\lib\netstandard2.0\Microsoft.Diagnostics.Runtime.dll" DestinationFolder="$(TestDependenciesPath)" />
+    <Copy SourceFiles="$(PkgSystem_Collections_Immutable)\lib\netstandard2.0\System.Collections.Immutable.dll" DestinationFolder="$(TestDependenciesPath)" />
+    <Copy SourceFiles="$(PkgSystem_Reflection_Metadata)\lib\netstandard2.0\System.Reflection.Metadata.dll" DestinationFolder="$(TestDependenciesPath)" />
+    <Copy SourceFiles="$(PkgSystem_Runtime_CompilerServices_Unsafe)\lib\netstandard2.0\System.Runtime.CompilerServices.Unsafe.dll" DestinationFolder="$(TestDependenciesPath)" />
+    <Copy SourceFiles="$(PkgSystem_Memory)\lib\netstandard2.0\System.Memory.dll" DestinationFolder="$(TestDependenciesPath)" />
+    <Copy SourceFiles="$(PkgSystem_Buffers)\lib\netstandard2.0\System.Buffers.dll" DestinationFolder="$(TestDependenciesPath)" />
+</Target>
+
+  <Target Name="MoveSomeTestDepsOutOfBin_net48" AfterTargets="Build" Condition=" '$(TargetFramework)' == 'net48' ">
+      <Message Importance="High" Text="This is a test" />
```
pinzart90, line 162: not needed

test/DynamoCoreTests/DynamoCoreTests.csproj
```diff
@@ -147,12 +147,33 @@
...
+    <Message Importance="High" Text="This is a test" />
+    <Copy SourceFiles="$(PkgMicrosoft_Diagnostics_Runtime)\lib\netstandard2.0\Microsoft.Diagnostics.Runtime.dll" DestinationFolder="$(TestDependenciesPath)" />
+    <Copy SourceFiles="$(PkgSystem_Collections_Immutable)\lib\netstandard2.0\System.Collections.Immutable.dll" DestinationFolder="$(TestDependenciesPath)" />
+    <Copy SourceFiles="$(PkgSystem_Reflection_Metadata)\lib\netstandard2.0\System.Reflection.Metadata.dll" DestinationFolder="$(TestDependenciesPath)" />
+    <Copy SourceFiles="$(PkgSystem_Runtime_CompilerServices_Unsafe)\lib\netstandard2.0\System.Runtime.CompilerServices.Unsafe.dll" DestinationFolder="$(TestDependenciesPath)" />
+    <Copy SourceFiles="$(PkgSystem_Memory)\lib\netstandard2.0\System.Memory.dll" DestinationFolder="$(TestDependenciesPath)" />
+    <Copy SourceFiles="$(PkgSystem_Buffers)\lib\netstandard2.0\System.Buffers.dll" DestinationFolder="$(TestDependenciesPath)" />
+</Target>
+
+  <Target Name="MoveSomeTestDepsOutOfBin_net48" AfterTargets="Build" Condition=" '$(TargetFramework)' == 'net48' ">
```
pinzart90, line 161: so we are moving stuff out even on net48 ? That will not break Revit installers ?

test/Libraries/PackageManagerTests/Setup.cs
```diff
@@ -0,0 +1,43 @@
+using Dynamo.Utilities;
```
pinzart90, line 1: why is a new Setup.cs needed ?

pinzart90 (approved):
A couple of questions, then LGTM

src/VisualizationTests/WpfVisualizationTests.csproj
mjkkirschner: because now `Moq` (v 4.18) has new dependencies which I only copy to the `test/test_dependencies` folder. So we need to setup a setupfixture resolve handler to find them.

test/DynamoCoreTests/DynamoCoreTests.csproj
mjkkirschner: no because these are new dependencies of `Moq` - and I put `Moq` back in the bin folder as the last step:
```
        <Copy SourceFiles="$(PkgMoq)\lib\netstandard2.0\Moq.dll" DestinationFolder="$(OutputPath)" />

```

src/VisualizationTests/WpfVisualizationTests.csproj
pinzart90: It just seemed to me that most of the Visualization tests are derived from SystemTestBase which already uses an AssemblyResolver which we could use

mjkkirschner: yes, but not all of them - then we'd have two places to remember to look for this resolve to occur.

mjkkirschner:
the test failure appears to have come from here:
https://github.com/DynamoDS/Dynamo/pull/13794
merging.

PR #13770, opened 2023-02-23
DYN-5605 Dynamo installs, but fails to pick up package

filipeotero:
### Purpose

This PR fixes the bug that doesn't refresh the status of some packages after being loaded.

refreshPackageDownloadStatus

### Declarations

- [ ] User-facing strings, if any, are extracted into `*.resx` files

### Reviewers

@QilongTang 

### FYIs

@RobertGlobant20 @jesusalvino

### DIFF
--- a/src/DynamoPackages/PackageLoader.cs
+++ b/src/DynamoPackages/PackageLoader.cs
@@ -315,10 +315,12 @@ namespace Dynamo.PackageManager
                 }
 
                 package.SetAsLoaded();
-                PackgeLoaded?.Invoke(package);
+
 
                 PythonServices.PythonEngineManager.Instance.
                     LoadPythonEngine(package.LoadedAssemblies.Select(x => x.Assembly));
+
+                PackgeLoaded?.Invoke(package);
             }
             catch (CustomNodePackageLoadException e)
             {


src/DynamoPackages/PackageLoader.cs
```diff
@@ -315,10 +315,12 @@ private void TryLoadPackageIntoLibrary(Package package)
                 }
 
                 package.SetAsLoaded();
-                PackgeLoaded?.Invoke(package);
+
 
                 PythonServices.PythonEngineManager.Instance.
                     LoadPythonEngine(package.LoadedAssemblies.Select(x => x.Assembly));
+
+                PackgeLoaded?.Invoke(package);
```
QilongTang, line 323: What would be the difference here to execute this slightly later?

filipeotero: There is a [piece of code](https://github.com/DynamoDS/Dynamo/blob/master/src/WorkspaceDependencyViewExtension/WorkspaceDependencyView.xaml.cs#L174) in the WorkspaceDepedencyView that checks the loaded python packages. 

The code was interpreting some packages as missing because, when it was hitting there, they were not included in the loaded assemblies. So, we need to trigger the `PackgeLoaded `action after the assembly is found in the available engines.

QilongTang (approved):

PR #13772, opened 2023-02-23
DYN-5530 library slider tooltip

filipeotero:
### Purpose

It's included in this PR a tooltip to show the value of the library's zoom slider. Also, the fix for the help icon.

sliderTooltip

### Declarations

### FYIs

@QilongTang @avidit

### DIFF
--- a/src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
+++ b/src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
@@ -1273,7 +1273,7 @@
                                         <RowDefinition Height="Auto"/>
                                     </Grid.RowDefinitions>
                                     <StackPanel Margin="0,10,0,5" Grid.Row="0" Orientation="Horizontal">
-                                        <TextBlock  FontSize="12" VerticalAlignment="Center" Text="Library"></TextBlock>
+                                        <TextBlock  FontSize="12" VerticalAlignment="Center" Text="{x:Static p:Resources.DynamoViewSettingLibraryZoomScale}"></TextBlock>
                                         <Label HorizontalAlignment="Left"
                                                    Name="LibraryZoomScaleInfoLabel"
                                                    VerticalAlignment="Bottom"
--- a/src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
+++ b/src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
@@ -1273,16 +1273,17 @@
                                         <RowDefinition Height="Auto"/>
                                     </Grid.RowDefinitions>
                                     <StackPanel Margin="0,10,0,5" Grid.Row="0" Orientation="Horizontal">
-                                        <TextBlock  FontSize="12" Text="Library"></TextBlock>
+                                        <TextBlock  FontSize="12" VerticalAlignment="Center" Text="Library"></TextBlock>
                                         <Label HorizontalAlignment="Left"
                                                    Name="LibraryZoomScaleInfoLabel"
-                                                   VerticalAlignment="Center" 
+                                                   VerticalAlignment="Bottom"
+                                                   Margin="0,2,0,0"
                                                    Height="26" 
                                                    Width="53">
                                             <Image
                                                Width="14"
                                                Height="14"
-                                               Margin="0,0,0,5"
+                                               Margin="0,0,0,0"
                                                Grid.Column="1"
                                                HorizontalAlignment="Center"
                                                VerticalAlignment="Center"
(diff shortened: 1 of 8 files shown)


src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
```diff
@@ -1273,16 +1273,17 @@
                                         <RowDefinition Height="Auto"/>
                                     </Grid.RowDefinitions>
                                     <StackPanel Margin="0,10,0,5" Grid.Row="0" Orientation="Horizontal">
-                                        <TextBlock  FontSize="12" Text="Library"></TextBlock>
+                                        <TextBlock  FontSize="12" VerticalAlignment="Center" Text="Library"></TextBlock>
```
QilongTang, line 1276: I did not notice this before, it seems the text `Library` is missing localization this way

src/LibraryViewExtensionWebView2/LibraryViewController.cs
```diff
@@ -88,7 +88,8 @@ internal LibraryViewController(Window dynamoView, ICommandExecutive commandExecu
 
         private void Browser_ZoomFactorChanged(object sender, EventArgs e)
         {
-            dynamoViewModel.Model.PreferenceSettings.LibraryZoomScale = ((float)browser.ZoomFactor);
+            //Multiplies by 100 so the value can be saved as a percentage
+            dynamoViewModel.Model.PreferenceSettings.LibraryZoomScale = (float)browser.ZoomFactor * 100;
```
QilongTang, line 92: If we use this approach, the number should be int instead of float right?

filipeotero: Yes, changed the variable to int

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
filipeotero: Thanks for noticing, added a resource for that.

QilongTang (approved):

QilongTang:
@filipeotero Please take a look at the regression [TestImportCopySettings](https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/8565/testReport/junit/Dynamo.Tests.Configuration/PreferenceSettingsTests/TestImportCopySettings/)

zeusongit:
Hey @filipeotero, not to extend the scope of this PR too much, if it is not a lot of work (in which case we can create a separate task) can you add the current zoom level indicator, like we have for ML autocomplete confidence level slider?
Screenshot 2023-02-23 at 6 40 21 PM

filipeotero:
> Hey @filipeotero, not to extend the scope of this PR too much, if it is not a lot of work (in which case we can create a separate task) can you add the current zoom level indicator, like we have for ML autocomplete confidence level slider? Was there a reason it was added as a tooltip and not a fixed element? Screenshot 2023-02-23 at 6 40 21 PM

I used the native Autotooltip from the slider. I will adapt it as it is in the ML autocomplete since we discussed in the Jira task that we want it to have the same behavior. Thank you for noticing that!

filipeotero:
fixedTooltipSliderNew

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
```diff
@@ -530,5 +530,20 @@ private void displayConfidenceLevel()
                 this.lblConfidenceLevel.Margin = new Thickness(left, -15, 0, 0);
             }
         }
+
+        private void zoomScaleLevel_ValueChanged(object sender, RoutedPropertyChangedEventArgs<double> e)
+        {
+            Slider slider = (Slider)sender;
+
+            //Since the percentage goes from 10 to 300, the value is decremented by 10 to standardize. 
+            double percentage = slider.Value - 10;
```
QilongTang, line 539: Now that it starts with 10%, thats why we minus 10 as a starting point?

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
```diff
@@ -530,5 +530,20 @@ private void displayConfidenceLevel()
...
+            //This is the relation between the margin in pixels and the value of the percentage
+            double marginValue = (79 * percentage / 29) - 480;
+            if (lblZoomScalingValue != null)
```
QilongTang, line 542: When would this value be null, do you have an example?

QilongTang (commented):
LGTM with two comments

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
filipeotero: Yes, to facilitate the calculation.

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
filipeotero: When the component is initialized, this method is called, but the value of the zoom scale is null.

PR #13813, opened 2023-03-10
Support embedded interop types

pinzart90:
Changes:
Interop types equivalence is checked when importing types from assemblies.

Embedded interop types will appear as distinct when you check type.Equals(other). 
Dotnet provides a new API to check equivalence = IsEquivalentTo
https://learn.microsoft.com/en-us/dotnet/framework/interop/type-equivalence-and-embedded-interop-types

Potential issues:
Even though the main part of dynamo that deals with import of types is now updated, there is a risk that other parts of dynamo still have code that compare type.Equals(other). This sort of code will not work as expected for embedded interop types


src/Engine/ProtoCore/FFI/CLRDLLModule.cs
```diff
@@ -10,6 +10,20 @@
 
 namespace ProtoFFI
 {
+
+    internal class TypeEqualityComparer : IEqualityComparer<Type>
+    {
+        public bool Equals(Type x, Type y)
+        {
+            return x.IsEquivalentTo(y) || x.Equals(y);
```
pinzart90, line 18: dotnet provides this method to check if 2 embedded COM types are equivalent
https://learn.microsoft.com/en-us/dotnet/api/system.type.isequivalentto?view=net-7.0
IsEqual and hash code are different between 2 equivalent COM types

src/Engine/ProtoCore/FFI/CLRDLLModule.cs
```diff
@@ -10,6 +10,20 @@
...
+        }
+
+        public int GetHashCode(Type obj)
+        {
+            return obj.GetHashCode();
```
pinzart90, line 23: The hash code will be different for 2 equivalent embedded COM types.
So this strategy does not work

src/Engine/ProtoCore/FFI/CLRDLLModule.cs
```diff
@@ -58,9 +72,12 @@ public static CLRModuleType GetInstance(Type type, CLRDLLModule module, string a
                     {
                         mtype = new CLRModuleType(type);
                         //Now check that a type with same name is not imported.
+                        bool equivalentComTypes = false;
                         Type otherType;
                         if (mTypeNames.TryGetValue(mtype.FullName, out otherType))
+                        {
                             throw new InvalidOperationException(string.Format("Can't import {0}, {1} is already imported as {2}, namespace support needed.", type.FullName, type.Name, otherType.FullName));
```
pinzart90, line 79: If no other ideas work, I suggest that we only change the message:
if (type.IsEquivalentTo(otherType)) throw "Equivalent Embeded interop types detected..etc. Dynamo does not support ...etc "
ANd maybe remove the `namespace support needed` part of the message.

FYI @mjkkirschner

src/Engine/ProtoCore/FFI/CLRDLLModule.cs
pinzart90: I could implement a custom dictionary that does not use hash codes. but the current dictionaries are already exposed as public. It would be complicated

src/Engine/ProtoCore/FFI/CLRDLLModule.cs
```diff
@@ -58,9 +72,12 @@ public static CLRModuleType GetInstance(Type type, CLRDLLModule module, string a
...
+                        }
 
                         mTypes.Add(type, mtype);
```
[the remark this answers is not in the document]
mjkkirschner: so - does this mean that we cannot even the more general case where 2 packages depend on the same binary?
`If I skip the other type (and only keep the first equivalent type) then we get into stackoverflows (types are expected to get stored in ther type map)`

It seems it must be possible - consider the Revit API, many packages reference it, but none distribute it.

OR are you trying to solve a different problem with his PR?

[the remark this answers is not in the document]
pinzart90: The problem is visible only when you embed types in the managed assembly.
"Normal" references will work just fine (like Revit API references)

src/Engine/ProtoCore/FFI/CLRDLLModule.cs
```diff
@@ -49,22 +49,24 @@ static CLRModuleType()
...
         {
-            CLRModuleType mtype;
-            if (!mTypes.TryGetValue(type, out mtype))
+            if (!mTypes.TryGetValue(type, out CLRModuleType mtype))
             {
                 lock (mTypes)
                 {
-                    if (!mTypes.TryGetValue(type, out mtype))
+                    mtype = new CLRModuleType(type);
+                    //Now check that a type with same name is not imported.
```
mjkkirschner, line 57: is this comment entirely accurate anymore?

test/Engine/FFITarget/FFITarget.csproj
```diff
@@ -22,9 +22,16 @@
         <Reference Include="Microsoft.CSharp" />
         <Reference Include="System.Data" />
         <Reference Include="System.Xml" />
+        <Reference Include="$(PkgMicrosoft_Office_Interop_Excel)\lib\net20\Microsoft.Office.Interop.Excel.dll">
+            <EmbedInteropTypes>True</EmbedInteropTypes>
+            <WrapperTool>tlbimp</WrapperTool>
```
mjkkirschner, line 27: what is this?

test/Engine/FFITarget/FFITarget.csproj
```diff
@@ -22,9 +22,16 @@
         <Reference Include="Microsoft.CSharp" />
         <Reference Include="System.Data" />
         <Reference Include="System.Xml" />
+        <Reference Include="$(PkgMicrosoft_Office_Interop_Excel)\lib\net20\Microsoft.Office.Interop.Excel.dll">
```
mjkkirschner, line 25: should this reference be wrapped in a condition for now?

mjkkirschner: will this make it to the bin folder? I think we purposefully removed it at some point.

test/test_support_projects/EmbeddedInterop.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <OutputType>Library</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>EmbeddedInterop</RootNamespace>
+    <AssemblyName>EmbeddedInterop</AssemblyName>
+  </PropertyGroup>
+    <ItemGroup>
+        <PackageReference Include="Microsoft.Office.Interop.Excel" Version="15.0.4795.1000" GeneratePathProperty="true">
+            <ExcludeAssets>all</ExcludeAssets>
+        </PackageReference>
+        <Reference Include="$(PkgMicrosoft_Office_Interop_Excel)\lib\net20\Microsoft.Office.Interop.Excel.dll">
```
mjkkirschner, line 13: so is there anything in this binary except for the office types?

mjkkirschner (approved):
provided the tests pass, I think it seems great except my question about the test binaries making their way to the bin folder.

I also am still not sure if the normal reference case works fine - but let's table that until we get a report from customers - maybe it was fixed a while ago and I missed it.

test/Engine/FFITarget/FFITarget.csproj
pinzart90: totally forgot about test dlls. Will work on removing them

test/test_support_projects/EmbeddedInterop.csproj
pinzart90: no, just office types and a wrapping class

pinzart90:
> provided the tests pass, I think it seems great except my question about the test binaries making their way to the bin folder.
> 
> I also am still not sure if the normal reference case works fine - but let's table that until we get a report from customers - maybe it was fixed a while ago and I missed it.

maybe we can discuss this tomorrow, maybe I am missing something

test/Engine/FFITarget/FFITarget.csproj
pinzart90: done

PR #13817, opened 2023-03-13
DYN-5369 (WIP)

LongNguyenP:
### Purpose

This PR implements [DYN-5369](https://jira.autodesk.com/browse/DYN-5369) (Allowing the Watch3D node to have variable number of input ports)

### Declarations

### Release Notes

image

This PR makes the Watch3D node into a VariableInputNode. This allows the user to avoid using the List.Create.


src/Libraries/Watch3DNodeModels/Watch3D.cs
```diff
@@ -95,9 +95,9 @@ public Watch3DCamera()
     [OutPortTypes("var")]
     [AlsoKnownAs("Dynamo.Nodes.dyn3DPreview", "Dynamo.Nodes.3DPreview", "Dynamo.Nodes.Watch3D", "DynamoWatch3D.Watch3D")]
     [IsDesignScriptCompatible]
-    public class Watch3D : NodeModel
+    public class Watch3D : VariableInputNode
```
mjkkirschner, line 98: This is an API break. This is a public type, and introducing a new intermediate base class which is abstract forces the implementor to implement two new methods.

Previous inheritors of this class will now likely have broken code at runtime.

Now it's unlikely that someone has inherited from Watch3d - but we have no idea and so this would force us to update to Dynamo 3.x. Anytime we decide to break API compatibility/semantic versioning for some reason a decision/discussion is required.

It's difficult to judge if the change is trivial or who it will affect, we can really use our best judgement.

To avoid those discussions a pattern can be used, like was done in the new `if` node work.
https://github.com/DynamoDS/Dynamo/pull/11728/files

the old node is obsoleted and an `[AlsoKnownAs]` attribute is added to the new node so migration is performed. You can also create more complex migrations using the migration framework to replace multiple nodes etc - like the List.Create.

mjkkirschner (changes requested):
please see comments regarding API compatibility - either way we should have tests that ensure old graphs continue to load and display watch3d data correctly - these may already exist.

src/Libraries/Watch3DNodeModels/Watch3D.cs
LongNguyenP: I am considering this solution: implementing a intermediate non-abstract class between Watch3D and VariableInputNode. This class will take care of the two required abstract methods from VariableInputNode and any other aspects so that the Watch3D node will still behave as it currently does (have not thought this through fully yet). Please Let me know your thoughts.

LongNguyenP: @mjkkirschner I just added a new BasicVariableInputNode class with default implementations for the two abstract methods. Will this be an acceptable solution?

mjkkirschner:
Hi @LongNguyenP unfortunately your branch is out of date/conflicted with master - the history needed to be rewritten.
Please follow these steps:
* pull latest master locally
* create a new branch from this new master
* cherry pick your commits in order
* force push to this branch so you will not need to open a new PR.

I will review when the PR is no longer in conflict.

src/Libraries/Watch3DNodeModels/Watch3D.cs
mjkkirschner: I don't know - did you look into the .net api compatibility documentation ?

https://learn.microsoft.com/en-us/dotnet/core/compatibility/library-change-rules

PR #13820, opened 2023-03-14
Fixing Ghost warning

jesusalvino:
### Purpose

Fix the bug https://jira.autodesk.com/browse/DYN-5553

### Declarations

### Reviewers
@QilongTang 

### FYIs
@RobertGlobant20

### DIFF
--- a/src/DynamoCore/Graph/Nodes/NodeModel.cs
+++ b/src/DynamoCore/Graph/Nodes/NodeModel.cs
@@ -1912,6 +1912,31 @@ namespace Dynamo.Graph.Nodes
             infos.Add(new Info(p, ElementState.AstBuildBroken));
         }
 
+        public void UpdateAssociatedWarning(string fileFromPathNodeName, string imageReadFromFileNodeName)
+        {
+            
+            if (this.Name == imageReadFromFileNodeName && this.InputNodes.Count > 0)
+            {
+                if (this.InputNodes.First().Value != null)
+                {
+                    var previousNode = this.InputNodes.First().Value.Item2;
+                    
+
+                    if (previousNode != null && previousNode.Name == fileFromPathNodeName)
+                    {
+                        if (this.Infos.Count > 0)
+                        {
+                            previousNode.State = ElementState.AssociatedWarning;
+                        }
+                        else
+                        {
+                            previousNode.State = ElementState.Active;
+                        }
+                    }
+                }
+            }
+        }
+
         #endregion
 
         #region Port Management
@@ -2914,6 +2939,7 @@ namespace Dynamo.Graph.Nodes
         Active,
         Warning,
         PersistentWarning,
+        AssociatedWarning,
         Error,
         AstBuildBroken,
         Info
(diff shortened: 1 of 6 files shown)


jesusalvino:
ghost_warning

reddyashish:
Can we add a test for this

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -1912,6 +1912,31 @@ public void NotifyAstBuildBroken(string p)
             infos.Add(new Info(p, ElementState.AstBuildBroken));
         }
 
+        public void UpdateAssociatedWarning(string fileFromPathNodeName, string imageReadFromFileNodeName)
+        {
+            
+            if (this.Name == imageReadFromFileNodeName && this.InputNodes.Count > 0)
```
QilongTang, line 1918: so `AssociatedWarning` only applies to this particular node for now?

src/DynamoCore/Graph/Nodes/NodeModel.cs
```diff
@@ -1912,6 +1912,31 @@ public void NotifyAstBuildBroken(string p)
...
+            {
+                if (this.InputNodes.First().Value != null)
+                {
+                    var previousNode = this.InputNodes.First().Value.Item2;
+                    
+
+                    if (previousNode != null && previousNode.Name == fileFromPathNodeName)
```
QilongTang, line 1925: so AssociatedWarning only applies to this particular node for now?

jesusalvino: I have applied this fix bearing in mind only the person responsible for display or not the ghost warning, its converter, it depended only on the zoom value, but it should be consider the node state also, that's why my fix consider both.

Even I couldn't trust on the status nodes, as we can see, both the FFP node and the IRFF node has Active as state (this second one displays ui alerts even), that's why I consider the info counts.

After checking your PR I agree with that, I can apply it here instead of the new status and update the unit test , or only keep here the unit test and consider your PR first, please let me know.

State doesn't match with warning message
ghost_warning_2

Unlink nodes
unlink


Thoughts:

IMHO from an user perspective, the warning messages would be display in all the nodes that are part of this set because the goal is display an image, the would be treated as a total, even if I unlink any node the image is still displayed.

I consider it would be a good candidate to have a specific node to display an image, we can save space for the draw getting less visual noise.

I would love to know more about the nodes behavior and how they should act as a group in a .dyn file anatomy, thanks for your appreciate feedback.

QilongTang: Thank @jesusalvino Check my other PR for comments and reasoning. I think for some reason, the `file from path` node got a warning during the computation and later the warning state was removed, but the color overlay was never removed due to issues in the code. So I think the existing states should be sufficient, if not, we should ask the Turbo team to look at VM and not remove the warning state after computation. Back to the issue, it seems naturally we would want to refresh the overlay states in this case after computation because that did not happen for this `file from path` node.

I would consider apply the fix in my proposed PR and combine it with your test, what do you think?

jesusalvino: > 

sure, let's do it

QilongTang: @jesusalvino Feel free to use the branch I created or use your own fork, looking forward to it

jesusalvino: Done.

PR #13822, opened 2023-03-14
Update analytics

pinzart90:
Remove all usage of IsOptedIn for ADP (it is considered an internal API by ADP)
Added `USEEXCEPTIONTRACE` option for CER as per this discussion https://autodesk.slack.com/archives/CH8CL3MRR/p1664815775433379

### DIFF
--- a/src/DynamoCoreWpf/UI/GuidedTour/GuidesManager.cs
+++ b/src/DynamoCoreWpf/UI/GuidedTour/GuidesManager.cs
@@ -8,6 +8,7 @@ using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Controls.Primitives;
 using Dynamo.Controls;
+using Dynamo.Logging;
 using Dynamo.ViewModels;
 using Dynamo.Wpf.Properties;
 using Dynamo.Wpf.ViewModels.GuidedTour;
@@ -474,7 +475,7 @@ namespace Dynamo.Wpf.UI.GuidedTour
                         Sequence = jsonStepInfo.Sequence,
                         RatingTextTitle = formattedText.ToString(),
                         StepType = Step.StepTypes.SURVEY,
-                        IsRatingVisible = dynamoViewModel.Model.PreferenceSettings.IsADPAnalyticsReportingApproved,
+                        IsRatingVisible = !Analytics.DisableAnalytics && Analytics.IsEnabled,
                         StepContent = new Content()
                         {
                             FormattedText = formattedText,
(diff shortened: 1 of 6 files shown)


src/DynamoCoreWpf/UI/GuidedTour/GuidesManager.cs
```diff
@@ -474,7 +475,7 @@ private Step CreateStep(Step jsonStepInfo, HostControlInfo hostControlInfo, int
                         Sequence = jsonStepInfo.Sequence,
                         RatingTextTitle = formattedText.ToString(),
                         StepType = Step.StepTypes.SURVEY,
-                        IsRatingVisible = dynamoViewModel.Model.PreferenceSettings.IsADPAnalyticsReportingApproved,
```
pinzart90, line 477: @mjkkirschner @QilongTang any ideas what this is used for ?

QilongTang: This is about the five star rating after guided tour exited, when analytics opt-in, we display the rating for ppl to rate

pinzart90: well now we will no longer know if ADP is opted in or not.
Should I just check for google analytics ? or as it is (if no one explicitly disabled analytics) ?

QilongTang: After this change, it seems only startup params would impact which I think should apply. I would leave it as it is. If user opt-out for GA, I think the client would send nothing right?

src/DynamoCoreWpf/UI/GuidedTour/GuidesManager.cs
```diff
@@ -474,7 +475,7 @@ private Step CreateStep(Step jsonStepInfo, HostControlInfo hostControlInfo, int
...
+                        IsRatingVisible = !Analytics.DisableAnalytics && Analytics.IsEnabled,
```
mjkkirschner, line 478: will `Analytics.IsEnabled` always be true?

src/DynamoCoreWpf/UI/GuidedTour/GuidesManager.cs
pinzart90: Sounds right

src/DynamoCoreWpf/UI/GuidedTour/GuidesManager.cs
pinzart90: IsEnabled is false when the DynamoAnalyticsClient is not created. DynamoAnalyticsClient is created when dynamo model is started with these flags
```
!IsServiceMode && !areAnalyticsDisabledFromConfig && !Dynamo.Logging.Analytics.DisableAnalytics
```

pinzart90: There might be some other conditions depending on how the Dynamo model is created

PR #13834, opened 2023-03-17
ALSO-5603 Add parameter hiding autocomplete options from the UI

spodo99:
Please Note:

### Purpose

The new flag is added to the setting to hide new autocomplete method options, if it will be requested by the users.  Per conversation at https://autodesk.slack.com/archives/C1866AFQB/p1679067914031959

### Declarations

### Release Notes

### Reviewers

### FYIs

### DIFF
--- a/src/DynamoCore/Configuration/IPreferences.cs
+++ b/src/DynamoCore/Configuration/IPreferences.cs
@@ -183,6 +183,18 @@ namespace Dynamo.Interfaces
     }
 
     /// <summary>
+    /// Temporary interface to avoid breaking changes.
+    /// TODO: Merge with StartupConfiguration for 3.0 (DYN-1699)
+    /// </summary>
+    internal interface IHideAutocompleteMethodOptions
+    {
+        /// <summary>
+        /// If true, autocomplete method options are hidden from UI 
+        /// </summary>
+        bool HideAutocompleteMethodOptions { get; set; }
+    }
+
+    /// <summary>
     /// Represents data about active state of preview background
     /// </summary>
     public class BackgroundPreviewActiveState
--- a/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
@@ -655,6 +655,11 @@ namespace Dynamo.ViewModels
             /// If true, Analytics and Usage options are hidden from UI 
             /// </summary>
             public bool HideReportOptions { get; set; }
+
+            /// <summary>
+            /// If true, autocomplete method options are hidden from UI 
+            /// </summary>
+            public bool HideAutocompleteMethodOptions { get; set; }
         }
 
         public static DynamoViewModel Start(StartConfiguration startConfiguration = new StartConfiguration())
@@ -750,6 +755,8 @@ namespace Dynamo.ViewModels
             model.RequestNotification += model_RequestNotification;
 
             preferencesViewModel = new PreferencesViewModel(this);
+            preferencesViewModel.HideAutocompleteMethodOptions = startConfiguration.HideAutocompleteMethodOptions;
+
 
 
             if (!DynamoModel.IsTestMode && !DynamoModel.IsHeadless)
--- a/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
+++ b/src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
@@ -655,11 +655,6 @@ namespace Dynamo.ViewModels
             /// If true, Analytics and Usage options are hidden from UI 
             /// </summary>
             public bool HideReportOptions { get; set; }
-
-            /// <summary>
-            /// If true, autocomplete method options are hidden from UI 
-            /// </summary>
-            public bool HideAutocompleteMethodOptions { get; set; }
         }
 
         public static DynamoViewModel Start(StartConfiguration startConfiguration = new StartConfiguration())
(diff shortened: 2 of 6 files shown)


src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -655,6 +655,11 @@ public struct StartConfiguration
             /// If true, Analytics and Usage options are hidden from UI 
             /// </summary>
             public bool HideReportOptions { get; set; }
+
+            /// <summary>
+            /// If true, autocomplete method options are hidden from UI 
+            /// </summary>
+            public bool HideAutocompleteMethodOptions { get; set; }
```
QilongTang, line 662: @pinzart @mjkkirschner Is this allowed to add new params in this class?

pinzart90: https://learn.microsoft.com/en-us/dotnet/core/compatibility/library-change-rules
looks at `DISALLOWED: Adding an instance field to a struct that has no nonpublic fields`
```
Adding any new fields - public or nonpublic - to such a struct is a source breaking change for these callers, as the compiler will now require the additional fields to be initialized.
```

mjkkirschner:
But doesn’t a bool have a default value ?On Mar 17, 2023, at 3:08 PM, pinzart90 ***@***.***> wrote:﻿
@pinzart90 commented on this pull request.



In src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs:
> @@ -655,6 +655,11 @@ public struct StartConfiguration
             /// If true, Analytics and Usage options are hidden from UI 
             /// </summary>
             public bool HideReportOptions { get; set; }
+
+            /// <summary>
+            /// If true, autocomplete method options are hidden from UI 
+            /// </summary>
+            public bool HideAutocompleteMethodOptions { get; set; }

https://learn.microsoft.com/en-us/dotnet/core/compatibility/library-change-rules
looks at DISALLOWED: Adding an instance field to a struct that has no nonpublic fields
Adding any new fields - public or nonpublic - to such a struct is a source breaking change for these callers, as the compiler will now require the additional fields to be initialized.

pinzart90:
> But doesn’t a bool have a default value ?

It does, but that is not the issue. Try this code in c#:
```
using System;
				
public struct Test {
	
	public int a; // Was added in version 1.0.0
	public double b; // Was added version 1.0.0
	public bool c;// Was added in version 1.0.1 (breakin change)
}

public class Program
{
	public static void Main()
	{
		Test m;
		
		m.a = 3;
		m.b = 0.4;
		Console.WriteLine(m.c); // Compilation error (line 18, col 21): Use of possibly unassigned field 'c 
	}
}
```

pinzart90:
> @pinzart90 Got it, sorry, in C++ it would fail only if you use initializer list... How would be better to pass this parameter then? Using new argument to DynamoViewModel::Start() function and DynamoViewModel constructor?

I don't know strict we've been in the past about this case. I think we've done this before without knowing it is a breaking  change.
The way we've extended existing classes is to add a new Interface to the Preferences class. There are quite a few now, looke at the interfaces here https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoCore/Configuration/IPreferences.cs

pinzart90:
@mjkkirschner any other ideas besides yet another interface ?

mjkkirschner:
> @mjkkirschner any other ideas besides yet another interface ?

no better ideas, except perhaps just modifying the concrete type.

mjkkirschner:
@spodo99 there is a test failure:

[Dynamo.Tests.Configuration.PreferenceSettingsTests.TestImportCopySettings](https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/8726/testReport/junit/Dynamo.Tests.Configuration/PreferenceSettingsTests/TestImportCopySettings/)

src/DynamoCore/Configuration/IPreferences.cs
```diff
@@ -182,6 +182,18 @@ internal interface IDisablePackageLoadingPreferences
         bool DisableCustomPackageLocations { get; set; }
     }
 
+    /// <summary>
+    /// Temporary interface to avoid breaking changes.
+    /// TODO: Merge with StartupConfiguration for 3.0 (DYN-1699)
+    /// </summary>
+    internal interface IHideAutocompleteMethodOptions
```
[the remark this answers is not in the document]
QilongTang: @pinzart90 I don't think we do currently, we intend to use Launchdarkly to control beta features and UI aspects in general.

QilongTang:
> @mjkkirschner any other ideas besides yet another interface ?

@pinzart @mjkkirschner I think startup params should also work for them right? See https://github.com/DynamoDS/Dynamo/blob/master/src/DynamoSandbox/DynamoCoreSetup.cs#L40, I guess when Alias launching DynamoSandbox, they can pass in param to modify certain feature sets. Essentially, this might just be similar to the existing way without an additional interface

mjkkirschner:
> Do you mean a startup configuration param? (which is a struct - and had the issue Tibi identified) ? or did you mean a new CLI argument? Or both?

I think that like Tibi said, it's an API break, but it's unclear to me who would be affected.

IMO The new preference is safer, the startup config option seems more idiomatic at the cost of integrators potentially having runtime or compile tie failures. (unlikely I guess).

spodo99:
@mjkkirschner  Added missing option to the new setting file

mjkkirschner:
Hi @spodo99 it looks good to me. I guess ideally it would be good to add a test that asserts the UI is actually hidden? I know UI tests are sometimes painful to write in wpf.

Perhaps because of the time constraints here you could merge this and send a followup for a UI test if it's straightforward enough to write - maybe something like this test:
https://github.com/DynamoDS/Dynamo/blob/master/test/DynamoCoreWpfTests/CoreUITests.cs#L760

mjkkirschner (approved):
the test failure is a flaky ui test.

QilongTang:
As communicated internally, we will make a 2.17.2 release for this, so this needs to be cherry-picked there

spodo99:
@mjkkirschner , @avidit found that I missed several option controls in the Preferences menu. I created a PR https://github.com/DynamoDS/Dynamo/pull/13859 fixing this. I control visibility of the entire StackPanel containing all ML controls now. Could you please merge it with 2.17.2 as well, if it's ok? 
I will create unit tests later as it was proposed...

PR #13841, opened 2023-03-21
support serializing dynamic python engines

mjkkirschner:
### Purpose

while working on IronPython3 package recharge project I noticed that dynamic engine names were not serializing correctly because of an obsolete public API that requires returning an enum.. This PR works around that.

Screen Shot 2023-03-21 at 3 21 18 PM

### Declarations

### Release Notes

### Reviewers

### FYIs

### DIFF
--- a/src/Libraries/PythonNodeModels/PythonNode.cs
+++ b/src/Libraries/PythonNodeModels/PythonNode.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.ComponentModel;
@@ -54,6 +54,11 @@ namespace PythonNodeModels
                 if (!Enum.TryParse(engine, out PythonEngineVersion engineVersion) ||
                     engineVersion == PythonEngineVersion.Unspecified)
                 {
+                    //if this is a valid dynamically loaded engine, return unknown, and serialize the name.
+                    if (PythonEngineManager.Instance.AvailableEngines.Any(x=>x.Name == engine))
+                    {
+                        return PythonEngineVersion.Unknown;
+                    }
                     // This is a first-time case for newly created nodes only
                     SetEngineByDefault();
                 }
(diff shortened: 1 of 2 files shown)


QilongTang (approved):

src/Libraries/PythonNodeModels/PythonNode.cs
```diff
@@ -54,6 +54,11 @@ public PythonEngineVersion Engine
                 if (!Enum.TryParse(engine, out PythonEngineVersion engineVersion) ||
                     engineVersion == PythonEngineVersion.Unspecified)
                 {
+                    //if this is a valid dynamically loaded engine, return unknown, and serialize the name.
+                    if (PythonEngineManager.Instance.AvailableEngines.Any(x=>x.Name == engine))
+                    {
+                        return PythonEngineVersion.Unknown;
```
pinzart90, line 60: why isn't PythonEngineVersion.Unspecified good enough to be returned ?

mjkkirschner: @pinzart90 - two reasons IMO:
1. If the version is unspecified it makes sense to me that the default version should kick in.
2. Unspecified is not the same as Unknown. In this case the version is definitely specified, it's just not a valid value in the enum.

pinzart90: So all this is to enhance the user experience for an obsoleted API? Is it still used by our integrators ?

pinzart90 (approved):

src/Libraries/PythonNodeModels/PythonNode.cs
mjkkirschner: hah, not quite, I don't know who uses it, and I don't know when it will disappear, and IMO it is about enhancing correctness not user experience - not collapsing two different states into one.

Let us remove this enum and never mention it again, but until then, I don't want to break python node initialization and I also want dynamic engines to serialize correctly.

PR #13851, opened 2023-03-24
use another target to generate resources

mjkkirschner:
### Purpose

@QilongTang reported that dynamocore.customization.dll was not being produced, and he was correct!

It seems that `ItemGroups` - are great for grouping items, but they do not actually run the tasks they contain, instead a target should be used... seems obvious now.

* use a target with a condition
* change the condition to match the more general check that @pinzart90 used which uses `OS != unix` instead of the target framework - this now matches the other csproj files which need to generate resource dlls.

### Declarations

### Release Notes

fix build

### DIFF
--- a/src/DynamoCore/DynamoCore.csproj
+++ b/src/DynamoCore/DynamoCore.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
   <ImportGroup Label="PropertySheets">
     <Import Project="$(SolutionDir)Config\CS_SDK.props" />
   </ImportGroup>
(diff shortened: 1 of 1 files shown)


QilongTang (approved):
awesome!

mjkkirschner:
unfortunately, there is a test failure I have not seen before from this PR - it only occurs on dyn_win_perf build machine - I have no idea how it's connected to these changes or why it only occurs on that machine.

pinzart90:
> unfortunately, there is a test failure I have not seen before from this PR - it only occurs on dyn_win_perf build machine - I have no idea how it's connected to these changes or why it only occurs on that machine - I would hold off merging until we know more.

I suggest you merge this PR and we can create a separate task for it. It does not seem likely that any change in this PR would cause a failure of ASM load

mjkkirschner:
> > unfortunately, there is a test failure I have not seen before from this PR - it only occurs on dyn_win_perf build machine - I have no idea how it's connected to these changes or why it only occurs on that machine - I would hold off merging until we know more.
> 
> I suggest you merge this PR and we can create a separate task for it. It does not seem likely that any change in this PR would cause a failure of ASM load

thanks @pinzart90 - I'm gonna spend a bit more time and will merge tomorrow and file if I can't figure it out.

mjkkirschner:
The last few runs on dyn_perf have all passed - (well not the other flaky UI tests) but I have not seen the preferences one fail with the ASM load exception - I've added some more logging here so hopefully if it does happen again we'll get a better indication of what to try next. 

Merging this for now, but if we see the issue come back please feel free to revert @pinzart90 @QilongTang .

PR #13863, opened 2023-03-31
DYN-5742 add workspace

RobertGlobant20:
### Purpose

Graph is not being added in the workspace because dyn file was not found.
The problem is that when executing DynamoSandbox from a path that contain empty spaces the URI.AbsolutePath is converting them to "%20" so the dyn file is not found.
Then with this fix I'm removing the "%20" from the string path and is found correctly when inserting the graph in the workspace.

### Declarations

### Release Notes

Graph is not being added in the workspace because dyn file was not found.

### Reviewers

@QilongTang 

### FYIs

@reddyashish

### DIFF
--- a/src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
+++ b/src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
@@ -1,18 +1,14 @@
 using System;
 using System.Collections.Generic;
-using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Reflection;
-using System.Web;
 using System.Windows;
-using System.Windows.Markup;
 using Dynamo.Core;
 using Dynamo.DocumentationBrowser.Properties;
 using Dynamo.Logging;
 using Dynamo.Utilities;
 using Dynamo.ViewModels;
-using Dynamo.Wpf.Extensions;
 
 namespace Dynamo.DocumentationBrowser
 {
--- a/src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
+++ b/src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
@@ -4,6 +4,7 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Web;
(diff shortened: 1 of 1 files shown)


src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
```diff
@@ -448,8 +445,7 @@ internal void InsertGraph()
         private string DynamoGraphFromMDFilePath(string path)
         {
             var dynPath = Path.Combine(Path.GetDirectoryName(path), Path.GetFileNameWithoutExtension(path).Replace("%20", " ")) + ".dyn";
-            return dynPath;
-            return path.Replace("%20", " ").Replace(".md", ".dyn");
+            return dynPath.Replace("%20", " ");
```
mjkkirschner, line 448: I think it is more robust to use a url decode method https://learn.microsoft.com/en-us/dotnet/api/system.web.httputility.urldecode?source=recommendations&view=net-7.0

RobertGlobant20:
GIF showing that the graph is added to workspace when Dynamo is executed from a path that contains empty spaces.
AddingGraphToWorkspace_EmptySpaces

src/DocumentationBrowserViewExtension/DocumentationBrowserViewModel.cs
RobertGlobant20: @mjkkirschner even if the DynamoGraphFromMDFilePath method is called from 2 different places I will need to use the UrlDecode method is both places, is that ok?

RobertGlobant20: Fixed in the next commit: https://github.com/DynamoDS/Dynamo/pull/13863/commits/5c3666153466e7fa04b950918fd7f7b46664a79f

reddyashish (approved):

reddyashish:
@RobertGlobant20 Can you add a test for this.

QilongTang:
From our internal discussion https://autodesk.slack.com/archives/C1866AFQB/p1680717374613279, merging this and continuing to work once @RobertGlobant20 is back.

PR #13873, opened 2023-04-05
use timespans to compare benchmarkdotnet results

mjkkirschner:
### Purpose

sometimes results from benchmarkdotnet are in ms, and sometimes they are in us. Use timespans to compare them.

I'm currently running this branch on the perf ci job.

This still needs work, the comparison makes more sense, but it yields NAN and some divide by zero errors, I guess we need more precision when converting us to ms. Looks like we are rounding somewhere to 2 decimals.

### Declarations


sm6srw (approved):
LGTM! I don't know if that test always fails now?

mjkkirschner:
somehow my math is still off, or we're losing precision, lookin into it.

mjkkirschner:
merging this - results seem good.

PR #13824, opened 2023-03-14
Dyn 5676 color picker custom colors popup

RobertGlobant20:
### Purpose

Several changes were done for supporting the Custom Color controls in the CustomColorPicker Popup, here is the list of the most important changes:
- I've added the controls for selecting the custom colors that will be visible if a flag is enabled (otherwise the available color controls will be shown).
- I've modified to current CustomColorPicker to use just one Popup (using 2 Popups was producing more problems).
- I've modified the code for creating the CustomColorPicker so now we will be passing the list of custom colors and the initial color that will be selected (in case it exist in the ColorPicker).
- I've updated the style for the ToggleButton inside the Color Palette node due that was not following the Visual Refresh style used in other ComboBoxes.

### Declarations

Check these if you believe they are true

- [X] The codebase is in a better state after this PR
- [X] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [X] The level of testing this PR includes is appropriate
- [X] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.
- [ ] This PR modifies some build requirements and the readme is updated

### Release Notes

Several changes were done for supporting the Custom Color controls in the CustomColorPicker Popup

### Reviewers

@QilongTang 

### FYIs

@avidit 
@Jingyi-Wen

PR #13895, opened 2023-04-12
Graphic Helpers Update

dnenov:
### Purpose

Add new preference setting to scale helpers in a graph (Grids, planes, coordinate systems etc.) under Preferences à Visual Settings à Display that ranges from 1 (current) to 10000 to unlock metric users.

There is a[ Dynamo Revit counterpart to this PR](https://github.com/DynamoDS/DynamoRevit/pull/2915), which allows Dynamo to pull current document units (on starting Dynamo only) when using Dynamo inside Revit.

#### Grid scale factor
grid scale 1

### Declarations

- [x] This PR contains no files larger than 50 MB 

### Release Notes

- multiplies the grid cell to scale (up)
- added `CustomColorsPopupTitle` resource inside the corresponding Resources.resx previously missed inside this [PR](https://github.com/DynamoDS/Dynamo/pull/13824)

### Reviewers

@reddyashish 
@QilongTang 

### FYIs

@mmisol


src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
```diff
@@ -626,5 +629,32 @@ private void UpdateZoomScaleValueLabel(Slider slider, Label label)
...
+
+        private void UpdateGraphicScaleValueLabel(Slider slider, Label label, bool initialRun = false)
+        {
+            double value = slider.Value;
+            if (label != null)
+            {
+                //The margin value for the label goes from - 480 to 310, resulting in 790 pixels from the starting point to the end.
+                //The value is decreased to 480 because the margin begins at - 480
+                //This is the relation between the margin in pixels and the value
+                double marginValue = (790 * value / 10000) - 480;
```
mjkkirschner, line 647: I think any kind of pixel logic is going to need robust testing in various DPI settings - theres probably a way to factor in wpf scale value here.

QilongTang:
@reddyashish @Amoursol Is this intended for 2.18 release?

Amoursol:
@QilongTang no, this can be 2.19.

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
```diff
@@ -16,6 +16,7 @@
 using Dynamo.Logging;
 using Dynamo.UI;
 using Dynamo.ViewModels;
+using static Dynamo.UI.Controls.DynamoToolTip;
```
reddyashish, line 19: No changes to this file, can remove the static usings.

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -883,6 +975,19 @@ public bool HideAutocompleteMethodOptions
...
+        /// <summary>
+        /// Returns if the current session is Dynamo Revit
+        /// </summary>
+        public bool IsDynamoRevit
+        {
+            get
+            {
+                // HostAnaltyicsInfo is not set when this is invoked??
+                //return this.dynamoViewModel.Model.HostAnalyticsInfo.HostName.Equals("Dynamo Revit");
+                return this.dynamoViewModel.Model.HostName.Equals("Dynamo Revit");
```
reddyashish, line 987: Model.HostName will be deprecated. Isn't HostAnaltyicsInfo providing the host name? Is that null here?

dnenov: That's exactly right. The HostAnalyticsInfo is null when we need it, however the old Model.HostName has the correct value.

reddyashish: Can the HostAnalyticsInfo be null and HostName have the correct host in any case? May be HostAnalyticsInfo is set later in the execution?.

src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml.cs
dnenov: Done!

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
dnenov: Updated as per @mjkkirschner comment - now checks for HostAnalyticsInfo first, if that is `null` goes on to check Model.HostName.

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -883,6 +1020,28 @@ public bool HideAutocompleteMethodOptions
...
+        {
+            get
+            {
+                // HostAnalyticsInfo is not set when this is invoked??
+                //return this.dynamoViewModel.Model.HostAnalyticsInfo.HostName.Equals("Dynamo Revit");
+                var host = this.dynamoViewModel.Model.HostAnalyticsInfo.HostName;
+
+                if (host != null)
+                {
+                    return host.Equals("Dynamo Revit");
```
reddyashish, line 1036: sorry i missed this but i don't think we want to check for the host "Dynamo Revit" by hardcoding it. We have a method GetKnownHosts on PackageManagerClient which can be used to check for the known hosts. @QilongTang What do you think?

reddyashish:
Thanks for updating this Deyan. LGTM with one last comment.

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
mjkkirschner: 1. I don't think that package manager method has the same list of names as the host names in analytics, and 
2. It seems a bit fishy to call a service just to get the name Revit/DynamoRevit which is being hardcoded somewhere else.
What will happen if the service is slow, times out etc, how will that impact Dynamo's performance for this simple check?

src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -150,6 +158,11 @@ public bool IsADPAnalyticsReportingApproved
         /// </summary>
         public bool ShowPreviewBubbles { get; set; }
 
+        /// <summary>
+        /// Indicates if Revit units should be used for graphic helpers for Dynamo Revit
+        /// </summary>
+        public bool UseRevitScaleUnits { get; set; }
```
mjkkirschner, line 164: I think it's strange that Revit should be referenced directly in Dynamo core?

src/DynamoCore/Configuration/Configurations.cs
```diff
@@ -393,6 +393,38 @@ public class Configurations
             }
         }
 
+        public enum Units
```
mjkkirschner, line 396: does this need to be public?

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -2466,11 +2466,26 @@ public void ToggleBackgroundGridVisibility(object parameter)
             }
         }
 
+        /// <summary>
+        /// Updates grapic helpers (grid) inside the background preview VM
+        /// </summary>
+        /// <param name="parameter"></param>
+        public void UpdateGraphicHelpersScale(object parameter)
+        {
+            BackgroundPreviewViewModel.GridScale = PreferenceSettings.GridScaleFactor;
```
mjkkirschner, line 2475: could `BackgroundPreviewViewModel` be null? maybe `?`

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
dnenov: This should be rather simple to achieve. We just need a way to know if we are running under DynamoRevit. It seems it's creating a lot of discussion. Could you, please, point me to a single good solution for it? The goal is to only display a certain user option if Dynamo is used inside Revit. I believe this is the first time something like this is happening (this has been brought up a few times).

To spin this around, if we cannot know if we are under Revit, we will not be able to use this new functionality.

src/DynamoCore/Configuration/PreferenceSettings.cs
dnenov: Hey @mjkkirschner - I am afraid this whole process requires for some form of awareness on the Dynamo side that this is a Revit session. Basically, the task is to make the grid layout behave differently under Revit, that's how I understand it. Maybe this core concept is no good, and then we may have to go back a bit and analyze the approach once again? @Amoursol

src/DynamoCore/Configuration/Configurations.cs
dnenov: I made them internal, would that be OK?

src/DynamoCore/Configuration/Configurations.cs
```diff
@@ -393,6 +393,38 @@ public class Configurations
...
+            Feet,
+            Inches,
+            Miles
+        }
+
+        /// <summary>
+        /// Supported Revit units and their conversion to dynamo units
+        /// Used in graphic helpers scaling (grid, axes)
+        /// </summary>
+        public static Dictionary<Units, double> SupportedUnits
```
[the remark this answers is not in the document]
dnenov: Same as above, internal.

src/DynamoCore/Configuration/PreferenceSettings.cs
```diff
@@ -240,6 +253,42 @@ public bool IsBackgroundGridVisible
...
+                gridScaleFactor = value;
+
+                RaisePropertyChanged(nameof(GridScaleFactor));
+            }
+        }
+
+        /// <summary>
+        /// The current Revit document units. Will be updated the first time Dynamo is started
+        /// </summary>
+        public Configurations.Units CurrentRevitUnits
```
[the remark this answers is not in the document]
dnenov: Same as above, internal.

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
dnenov: Yup, got it! Added the check.

src/DynamoCore/Configuration/PreferenceSettings.cs
Amoursol: @mjkkirschner agree that having a Revit reference in Core isn't ideal. Is there a way to inject this functionality in from D4R instead?

src/DynamoCore/Configuration/PreferenceSettings.cs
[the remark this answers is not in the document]
Amoursol: Same comment as https://github.com/DynamoDS/Dynamo/pull/13895#discussion_r1197898057

src/DynamoCore/Configuration/PreferenceSettings.cs
dnenov: Hi @mjkkirschner - I replaced 'Revit' to 'Host' for a platform-agnostic approach, I hope that addresses the concern around semantics, and sorry if I didn't understand correctly where the issue was coming from. I hope I got it right this time around!

src/DynamoCoreWpf/ViewModels/Menu/PreferencesViewModel.cs
```diff
@@ -791,10 +791,6 @@ private Configurations.Units GetTransformedRevitUnits(Configurations.Units revit
             {
                 return Configurations.Units.Millimeters;
             }
-            else if (revitUnits == Configurations.Units.Inches)
```
reddyashish: Is this not needed?

dnenov: Actually, it turns out that it is not needed. For Imperial units, the direct translation of scale works great. For Metric units, an inverse translation had to take place for things to look 'normal' against the grid. I, then, removed the unnecessary translation for Inches.

reddyashish:
CustomColorPicker_PrePopulateDefaultColors and TestImportCopySettings are failing in the latest job. Please check them once.

reddyashish:
Running the job again now to confirm the failing tests. @mjkkirschner Let us know if you think of any changes needed.

PR #13896, opened 2023-04-12
DYN-5369: variable number of input ports for Watch3D nodes

LongNguyenP:
### Purpose

This PR implements [DYN-5369](https://jira.autodesk.com/browse/DYN-5369) (Allowing the Watch3D node to have variable number of input ports)

### Declarations

### Release Notes

image

This PR makes the Watch3D node into a VariableInputNode. This allows the user to avoid using the List.Create.


LongNguyenP:
@mjkkirschner
After some thoughts, I think the API breaking change problem does not exist. It is true that Watch3d now inherits from VariableInputNode and needs to implement the two abstract method (and it already does), but this will not cause any problem for any existing third-party nodes that inherit from Watch3d because the two abstract methods have already been implemented in the Watch3d. These third-party nodes should just work as they are. Would be great if you can confirm whether I am correct here.

mjkkirschner:
Hi @LongNguyenP,

I tested your question - by creating a custom watch3d node inheriting from `Watch3D` referencing 2.18 binaries, and then loaded it into dynamo built against your branch.

It did load correctly, and the UI was updated to include the new buttons.

the docs do mention your question as well:

```
Introducing a new base class

So long as it does not introduce any new abstract members or change the semantics or behavior of existing members, a type can be introduced into a hierarchy between two existing types. For example, between .NET Framework 1.1 and .NET Framework 2.0, we introduced DbConnection as a new base class for SqlConnection which previously derived from Component.
```

and indicate that it IS an API break... but given my test, I am not sure under what circumstances, maybe only if the type does not override these methods, I am not sure.

@pinzart90 any thoughts on this? I am inclined to overlook it because I doubt watch3d is inherited frequently.

src/DynamoCoreWpf/Controls/DynamoNodeButton.cs
```diff
@@ -54,16 +59,16 @@ public DynamoNodeButton(ModelBase model, string eventName)
...
             // needs the "ModelEventCommand" to be sent when user clicks
             // on the button.
-            // 
+            //
             if (null != this.model && (!string.IsNullOrEmpty(this.eventName)))
             {
                 MessageBoxResult result = MessageBoxResult.None;
 
-                if (eventName.Equals("RemoveInPort"))
+                if (eventName.Equals("RemoveInPort") && ShowWarningForRemovingInPort)
```
mjkkirschner, line 71: what is the deal with this? Were you going to set this true in a previous use case or something?

src/Libraries/Watch3DNodeModels/Watch3D.cs
```diff
@@ -310,5 +286,23 @@ private void DeserializeCamera()
             // No visualization update is required for this node type.
             return false;
         }
+
+        protected override string GetInputTooltip(int index)
+        {
+            return string.Empty;
```
mjkkirschner, line 292: we don't want tooltips or names?

src/Libraries/Watch3DNodeModelsWpf/HelixWatch3DNodeViewModel.cs
```diff
@@ -116,11 +116,11 @@ protected override void PortDisconnectedHandler(PortModel obj)
         {
             if (obj.PortType == PortType.Input  && watchModel == obj.Owner)
             {
-                OnClear();
+                UpdateUpstream();
```
mjkkirschner, line 119: what is this change for?

src/Libraries/Watch3DNodeModelsWpf/Watch3DNodeViewCustomization.cs
```diff
@@ -93,47 +95,63 @@ public void CustomizeView(Watch3D model, NodeView nodeView)
...
+            {
+                VerticalAlignment = VerticalAlignment.Top,
+                HorizontalAlignment = HorizontalAlignment.Center
+            };
+
+            wrapPanel.Children.Add(
+                new DynamoNodeButton(nodeView.ViewModel.NodeModel, "AddInPort")
+                {
+                    Content = "+",
+                    Style = (Style)SharedDictionaryManager.DynamoModernDictionary["AddRemoveButton"]
```
mjkkirschner, line 142: is it possible to use nameof or some other method to have this fail at compile time if the key is changed?

mjkkirschner (changes requested):
1.A few questions and
2. I think this needs tests for the new methods / interactions.

src/DynamoCoreWpf/Controls/DynamoNodeButton.cs
```diff
@@ -40,6 +43,8 @@ private Window Owner
             }
         }
 
+
+
```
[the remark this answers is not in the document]
LongNguyenP: Looks like so. I already removed those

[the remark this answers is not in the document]
LongNguyenP: Added

src/Libraries/Watch3DNodeModels/Watch3D.cs
LongNguyenP: I just added tooltip. Keeping the same as the current tooltip "Incoming Geometry Objects". I think no need for name as the default

src/Libraries/Watch3DNodeModelsWpf/HelixWatch3DNodeViewModel.cs
LongNguyenP: Without this change, the view inside the node will not be updated when a port is disconnected.

src/Libraries/Watch3DNodeModelsWpf/Watch3DNodeViewCustomization.cs
LongNguyenP: I am not sure if nameof is possible here. This part of the code is based on the NodeViewCustomization of the VariableInputNode and there the name "AddInPort" etc.. were actually hardcoded

src/Libraries/Watch3DNodeModels/Watch3D.cs
```diff
@@ -171,34 +171,10 @@ public void SetWatchSize(double w, double h)
 
         public override IEnumerable<AssociativeNode> BuildOutputAst(List<AssociativeNode> inputAstNodes)
         {
-            if (IsPartiallyApplied)
```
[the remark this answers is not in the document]
LongNguyenP: If partially-applied Watch3d would output a function.

Here in this PR's implemenation. 

> 1.A few questions and 2. I think this needs tests for the new methods / interactions.

OK will add the tests :)

src/DynamoCoreWpf/Controls/DynamoNodeButton.cs
```diff
@@ -16,6 +16,9 @@ public class DynamoNodeButton : Button
         private string eventName = string.Empty;
         private ModelBase model = null;
         private DynamoViewModel dynamoViewModel;
+
+        public bool ShowWarningForRemovingInPort { get; set; }
```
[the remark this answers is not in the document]
LongNguyenP: Added

src/DynamoCoreWpf/Controls/DynamoNodeButton.cs
mjkkirschner: > This was from the current Master pull. The `VariableInputNode `from Master now shows a warning dialog box whenver the user click on the minus button to remove a port. For the Watch3D node, we don't need this dialog box to show up, hence I added the `ShowWarningForRemovingInPort `flag to disable it

I'm still confused - I never see you set this true, and the default will be false, so does this affect existing nodes? Did you intend to set it true for existing nodes?

mjkkirschner:
looks like a bunch of failing tests in last build.

LongNguyenP:
Yep it seems that in this PR, Dyn files that contain Watch3D node saved in previous version did not open correctly. Looks like the deseiralization method of the VariableInputNode is not compatible with existing Watch3d implementation  Investigating now. L

src/DynamoCoreWpf/Controls/DynamoNodeButton.cs
LongNguyenP: Yep you are right. I should have give it a "true" default value. Fixed now

pinzart90:
> Hi @LongNguyenP,
> 
> I tested your question - by creating a custom watch3d node inheriting from `Watch3D` referencing 2.18 binaries, and then loaded it into dynamo built against your branch.
> 
> It did load correctly, and the UI was updated to include the new buttons.
> 
> the docs do mention your question as well:
> 
> ```
> Introducing a new base class
> 
> So long as it does not introduce any new abstract members or change the semantics or behavior of existing members, a type can be introduced into a hierarchy between two existing types. For example, between .NET Framework 1.1 and .NET Framework 2.0, we introduced DbConnection as a new base class for SqlConnection which previously derived from Component.
> ```
> 
> and indicate that it IS an API break... but given my test, I am not sure under what circumstances, maybe only if the type does not override these methods, I am not sure.
> 
> @pinzart90 any thoughts on this? I am inclined to overlook it because I doubt watch3d is inherited frequently.

I do not think there will be any binary compatibility breaking changes or source code breaking changes as long as the `Watch3D` class implements all the abstract methods of the new intermediary base class `VariableInputNode`
There might be Behavioral changes if the new Intermediary base class changes the overrides of the `NodeModel` virtual methods

mjkkirschner:
@LongNguyenP please don't update these xml files unless you have a good reason - the fact that you need to illustrates that legacy user graphs that were saved in .xml will potentially be broken. Have you been able to find out why the failures are occurring with xml deserialization.

We still need to support xml deserialization!

LongNguyenP:
> We still need to support xml deserialization!

Understood. I am digging deeper in the issue with XML now.en 
BTW. Do you remember at which point Dynamo swtiched from XML to JSON dyn files?

LongNguyenP:
Ok I think I have got it. It was indeed because the XML deseiralization of the VariableInputNode (which Watch3D now inherits) was expecting a an XML attribute called "inputcount", which did not exist in the existing Watch3D's XML serialized data. 

I have addressed this and the tests now pass on my local machine. Let's see they will pass on Jenkins too.

LongNguyenP:
Looks like all tests now pass on Jenkins! 

Except `CoreUserInterfaceTests.WorkspaceContextMenu_TestIfOpenOnRightClick`. However, this test passes on my machine and does not seem to have anything to do with Watch3d. Investigating...

src/Libraries/Watch3DNodeModels/Watch3D.cs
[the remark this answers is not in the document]
LongNguyenP: I am not sure what IsPartiallyApplied mean? (I am still quite ignorant of many of Dynamo's inner working details) Does this mean the node will output a Function when the input is a Function? If so, this behavior is still maintined in this PR

image

LongNguyenP:
All tests have passed now !

mjkkirschner:
> > We still need to support xml deserialization!
> 
> Understood. I am digging deeper in the issue with XML now.en BTW. Do you remember at which point Dynamo swtiched from XML to JSON dyn files?

this happened when we moved from dynamo 1.x to 2.x

src/Libraries/Watch3DNodeModels/Watch3D.cs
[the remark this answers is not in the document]
mjkkirschner: kind of, partially applied is a functional programming term - it means that a function has been created with some input parameters set and others still free.

Can you please confirm what happened before and after your changes if you actually invoke the function on some input, you can use `map` node - I don't think this is so important but honestly I have no idea what the behavior was, so let's see...

My feeling is that it used to just return the input object, lets see if it still does that?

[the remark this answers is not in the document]
mjkkirschner: your screenshot is not what I mean - 
we want the output of the Watch3D node to be a function - or rather, I am trying to understand if that used to work and now it is broken - which I am guessing is the case because of the code you have removed.

mjkkirschner:
I also notice the watch3d node looks a bit different -FYI @Jingyi-Wen for review.

mjkkirschner:
> Looks like all tests now pass on Jenkins!
> 
> Except `CoreUserInterfaceTests.WorkspaceContextMenu_TestIfOpenOnRightClick`. However, this test passes on my machine and does not seem to have anything to do with Watch3d. maybe just due to Jenkins . Investigating...

yeah this test is just flaky, don't worry about that one, I think it's marked failure now.

LongNguyenP:
> I also notice the watch3d node looks a bit different -FYI @Jingyi-Wen for review.

We made some change to the UI/UX design. We wanted to use per-port context menu for add and removing ports (instead of using the plus or minus button). This way the user can insert a new node in-between existing node and remove a node in the middle if needed. But this require significantly more work so @saintentropy and I thought we should just keep using the plus and minus buttons for now and make another task for the per-port context menu later (I believe this won't break any existing graph that uses plus and minus buttons)

mjkkirschner:
> > I also notice the watch3d node looks a bit different -FYI @Jingyi-Wen for review.
> 
> We made some change to the UI/UX design. We wanted to use per-port context menu for add and removing ports (instead of using the plus or minus button). This way the user can insert a new node in-between existing node and remove a node in the middle if needed. But this require significantly more work so @saintentropy and I thought we should just keep using the plus and minus buttons for now and make another task for the per-port context menu later (I believe this won't break any existing graph that uses plus and minus buttons)

thanks @LongNguyenP - last thing can you please see my clarifying comments on the partial function application (function passing) and check the behavior before and after.

src/Libraries/Watch3DNodeModels/Watch3D.cs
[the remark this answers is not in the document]
LongNguyenP: @mjkkirschner You are right, in the current master branch the Watch3D node will output a function (rather than "null) if the only-input port is free.

So looks like this PR will break any existing graph that relies on this function-output behavior, even thought I doubt if anyone would use Watch3D in such a way.

aparajit-pratap:
@LongNguyenP please give an appropriately descriptive title to your Dynamo PRs, other than just the JIRA ticket number. It helps reviewers both internally and externally what the overall intention of the change is.

mjkkirschner (approved):

PR #13904, opened 2023-04-15
Add Pause Splash Screen Tooltip and Run Preview Tooltip

jesusalvino:
### Purpose
Implementing the improvement https://jira.autodesk.com/browse/DYN-5721

### Declarations

### Reviewers
@QilongTang 

### FYIs
@RobertGlobant20
@Enzo707

### DIFF
--- a/src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
+++ b/src/DynamoCoreWpf/Views/Menu/PreferencesView.xaml
@@ -260,12 +260,10 @@
                                 <Grid Grid.Row="1"
                                   Margin="4,0,0,20">
                                     <Grid>
-                                        <Grid.ToolTip>
-                                            <ToolTip Content="{x:Static p:Resources.PreferencesViewShowStaticSplashScreenTooltip}" Style="{StaticResource GenericToolTipLight}"/>
-                                        </Grid.ToolTip>
                                         <Grid.ColumnDefinitions>
                                             <ColumnDefinition Width="Auto"/>
-                                            <ColumnDefinition Width="*"/>
+                                            <ColumnDefinition Width="Auto"/>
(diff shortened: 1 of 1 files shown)


jesusalvino:
pref-tooltips

zeusongit:
Does this work with localized strings? Can you test that as well?

QilongTang (approved):

QilongTang:
> > > Does this work with localized strings? Can you test that as well?
> > 
> > 
> > Do you mean when the tooltip is localized?
> 
> No I mean when the preference option is localized, I saw some examples where when localized the location of the question mark will not change while the string is now longer than it is in English.

Gotcha, do you recall which ones? I guess for @jesusalvino Please just sanity check other languages to see if the position is good

zeusongit:
> > > > Does this work with localized strings? Can you test that as well?
> > > 
> > > 
> > > Do you mean when the tooltip is localized?
> > 
> > 
> > No I mean when the preference option is localized, I saw some examples where when localized the location of the question mark will not change while the string is now longer than it is in English.
> 
> Gotcha, do you recall which ones? I guess for @jesusalvino Please just sanity check other languages to see if the position is good

German / de-DE

reddyashish:
@Jingyi-Wen Looks like there is extra space between the label name and the icon. Probably reduce that a little? Also not icon is not center aligned vertically in some cases.

jesusalvino:
de-DE lang looks good IMHO

image

zeusongit (approved):

Jingyi-Wen:
> @Jingyi-Wen Looks like there is extra space between the label name and the icon. Probably reduce that a little? Also the icon is not center-aligned vertically in some cases.

Good catch! We definitely should have them center-aligned vertically. And agree with reducing the space, the icon next to the Max Number of Recent Files text box is a good reference

PR #13910, opened 2023-04-18
Fix Civil3D hostname to fix psuedo warning for package users

zeusongit:
### Purpose

https://jira.autodesk.com/browse/DYN-5470
Added `Dynamo Civil 3D` as the hostname for Civil 3D packages that have `Civil 3D` as the host dependency.

### Declarations

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/src/DynamoPackages/PackageManagerExtension.cs
+++ b/src/DynamoPackages/PackageManagerExtension.cs
@@ -44,14 +44,6 @@ namespace Dynamo.PackageManager
         /// </summary>
         private Dictionary<string, List<PackageInfo>> NodePackageDictionary;
 
-        // TODO : Update all packages to use this hostname and replace the hostname in package manager as well.
-        /// <summary>
-        /// This will be used to match Civil 3D specific packages,
-        /// as there is a mismatch between the host dependency name in packages and the host name used by Civil 3D for Dynamo.
-        /// </summary>
-        private readonly string Civil3DHostName = "Dynamo Civil 3D";
-        private readonly string oldCivil3DHostName = "Civil 3D";
-
         public string Name { get { return "DynamoPackageManager"; } }
 
         public string UniqueId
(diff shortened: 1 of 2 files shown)


src/DynamoPackages/PackageManagerExtension.cs
```diff
@@ -44,6 +44,14 @@ public class PackageManagerExtension : IExtension, ILogSource, IExtensionSource
         /// </summary>
         private Dictionary<string, List<PackageInfo>> NodePackageDictionary;
 
+        // TODO : Update all packages to use this hostname and replace the hostname in package manager as well.
+        /// <summary>
+        /// This will be used to match Civil 3D specific packages,
+        /// as there is a mismatch between the host dependency name in packages and the host name used by Civil 3D for Dynamo.
+        /// </summary>
+        private readonly string Civil3DHostName = "Dynamo Civil 3D";
+        private readonly string oldCivil3DHostName = "Civil 3D";
```
QilongTang, line 53: hmm it's a bit weird to define this in the Dynamo code base, should we switch to a more tolerant match based on your earlier comment?

mjkkirschner: just curious - could we do this transformation at runtime on the package manager?

zeusongit: @mjkkirschner Do you mean adding the hostname in the hosts returned from Package Manager?
here: https://git.autodesk.com/Dynamo/PackageManager/blob/c1e6047b5a92419049098c4a478bfb463aed6aae/src/app.js#L239

zeusongit: > 

updated with more tolerant approach

mjkkirschner: no, I mean that when package manager encounters a package with this old hostname to just return the new one instead in the response when the user downloads this package or gets its header... but perhaps this has issues with the pkg.json?

reddyashish:
looks like a sporadic failure. Running the job again: https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/8921/

reddyashish:
Cherrypick this?

PR #13917, opened 2023-04-20
implement new token interface

mjkkirschner:
### Purpose

depends on this PR:
https://github.com/DynamoDS/PackageManagerClient/pull/80

implement new interface

### Declarations

### Release Notes

make single sign on usable from graphs


reddyashish (approved):

mjkkirschner:
pass here:
https://master-5.jenkins.autodesk.com/job/Dynamo/job/DynamoSelfServe/job/pullRequestValidation/9005/

@zeusongit - I notice that this failure occurred AFTER I merged your changes, and I also noticed that these tests actually put real load on the package manager to get the hosts lists - if for some reason the package manager returns an empty list or times out during a test run, the value may now be cached. 

I've only seen it infrequently, but perhaps we should either make the cache short lived, only cache if the result is not garbage, and in addition make these tests use mocks.

Issue #13671, opened 2023-01-10
 Loading .NET library: CompilerInternalException if methods are present with optional parameter of type enum (2nd attempt)


MichaelBuehler:
If this issue is with Dynamo for Revit, please post your issue on the [Dynamo for Revit Issues page](https://github.com/DynamoDS/DynamoRevit/issues).

If this issue is **not** a bug report or improvement request, please check the [Dynamo forum](https://forum.dynamobim.com/), and start a thread there to discuss your issue.

## Dynamo version

ALL up to 2.17

## Operating system

Windows 10

## What did you do? 

Loading a dll into Dynamo Sandbox or Dynamo Revit.
The dynamo load library function will cause 'CompilerInternalException' error message if a method like this is present

myClass.AnyMethod(myCustomEnumType parameter = myCustomEnumType .None);
because "parameter" provides a default value of a custom type.
It looks like all dynamo versions are affected (I found in 2.12.0 but it appears also inside current 'main' code).

This is caused by the Dynamo code parsing the dll to generate nodes:

Engine\ProtoCore\FFI\CLRDLLModule
Method ParseArgumentSignature(MethodBase method)

This is caused by the Dynamo code parsing the dll to generate nodes:

Engine\ProtoCore\FFI\CLRDLLModule
Method ParseArgumentSignature(MethodBase method)
..
foreach (var parameter in parameter)
{
if (parameter.IsOptional
{
var defaultValue = = parameter.DefaultValue;
if (defaultValue != null)
{

**var rhs = AstFactory.BuildPrimitiveNodeFromObject(defaultValue); // EXCEPTION**

}
}
}

which calls

Engine/ProtoCore/Parser/AssociativeAST.cs
Method BuildPrimitiveNodeFromObject(object value)
where value must be of type short, int, long, float, double, string, boolan - otherwise it will throw exception.

## What did you expect to see?

No error on loading a dll.

Also: If error a more meaningful error message would be helpful.
Currently: 'Exception of type 'ProtoCore.Exceptions.CompilerInternalException' was thrown

Including dll name + class name + method name (or at least dll name) would be extreme helpful

quite often errors can be caused due a dependency.. (version conflict for example). Which dependency?

## What did you see instead?

A dialog 'Failed to load library'.

mjkkirschner:
https://jira.autodesk.com/browse/DYN-5506

erfajo:
[...]

PR #13925, opened 2023-04-21
support enum default values in imported zt methods

mjkkirschner:
### Purpose

https://github.com/DynamoDS/Dynamo/issues/13671
https://jira.autodesk.com/browse/DYN-5506

at @aparajit-pratap's suggestion - instead of trying to generate and cast and int to an enum when we encounter an enum default value - instead generate a get method call for the generated getter which represents the specific enum field. This works as `Enum` types are imported into the VM as classes.

a test is added.

### Declarations

### Release Notes

support enum default values in zero touch import


erfajo:
Sorry to interrupt, but this does look like something I have put on the wish list -> https://github.com/DynamoDS/DynamoWishlist/issues/237

I think this can be solved by adding another constructor to the attribute class
Autodesk.DesignScript.Runtime.DefaultArgumentAttribute

```
[AttributeUsage(AttributeTargets.Parameter)]
public sealed class DefaultArgumentAttribute : Attribute
{
    public dynamic ArgumentExpression { get; private set; }

    public DefaultArgumentAttribute(string defaultArgumentExpression)
    {
        ArgumentExpression = defaultArgumentExpression;
    }

    public DefaultArgumentAttribute(string enumValue, Type enumType)
    {
        ArgumentExpression = Enum.Parse(enumType, enumValue);
    }
}
```

PR #13973, opened 2023-05-08
Enabling the Dynamo Environment

jesusalvino:
### Purpose

Enabling the Dynamo Environment based on the IsEnabled properties of their main components instead of the Preferences Deactivated Event for the implementation of the improvement https://jira.autodesk.com/browse/DYN-5656.

### Declarations

### Reviewers
@QilongTang 
@reddyashish 

### FYIs
@RobertGlobant20 
@Enzo707

### DIFF
--- a/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
+++ b/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
@@ -1999,11 +1999,7 @@ namespace Dynamo.Controls
 
         public void EnableEnvironment(bool isEnabled)
         {
-            this.titleBarGrid.IsEnabled = isEnabled;
-            this.shortcutBarGrid.IsEnabled = isEnabled;
-            this.sidebarGrid.IsEnabled = isEnabled;
-            this.WorkspaceTabs.IsEnabled = isEnabled;
-            this.bottomBarGrid.IsEnabled = isEnabled;
+            this.mainGrid.IsEnabled = isEnabled;
         }
 
         /// <summary>
--- a/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
+++ b/src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
@@ -140,6 +140,7 @@ namespace Dynamo.Controls
 
             SizeChanged += DynamoView_SizeChanged;
             LocationChanged += DynamoView_LocationChanged;
+            MouseLeftButtonDown += DynamoView_MouseLeftButtonDown;
 
             // Apply appropriate expand/collapse library button state depending on initial width
             UpdateLibraryCollapseIcon();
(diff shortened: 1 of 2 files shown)


jesusalvino:
pref_enabling

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -1984,7 +1994,16 @@ private void OnPreferencesWindowClick(object sender, RoutedEventArgs e)
             preferencesWindow = new PreferencesView(this);
             dynamoViewModel.OnPreferencesWindowChanged(preferencesWindow);
             preferencesWindow.WindowStartupLocation = WindowStartupLocation.CenterOwner;
-            preferencesWindow.ShowDialog();
+            preferencesWindow.Show();
+        }
+
+        public void EnableEnvironment(bool isEnabled)
+        {
+            this.titleBarGrid.IsEnabled = isEnabled;
```
reddyashish, line 2002: When the pref window is opened, these are already disabled, right? why do we need this method?

jesusalvino: yes, because it was being opened using ShowDialog(), but in this way I couldn't find how to detect when the user click outside the preferences, that's Why I'm using Show() instead so I need to disable/enable the main components of Dynamo explicitly

QilongTang: so you mean `Show()` will set `PreferencesWindow.IsLoaded` to true?

jesusalvino: Using the ShowDialog() didn't allow me detect when the user click outside the preferences view (main goal), when we use it we don't worry about the state because it locks the Dynamo environment.

jesusalvino: @QilongTang with my last commit https://github.com/DynamoDS/Dynamo/pull/13973/commits/14c28d26e24ddfeb02a1aa807ba600ce82bac2d6 we won't worry about  enable/disable specific Dynamo parts if we add a new grid.

QilongTang (approved):
LGTM with one last comment

src/DynamoCoreWpf/Views/Core/DynamoView.xaml.cs
```diff
@@ -1984,7 +1994,12 @@ private void OnPreferencesWindowClick(object sender, RoutedEventArgs e)
             preferencesWindow = new PreferencesView(this);
             dynamoViewModel.OnPreferencesWindowChanged(preferencesWindow);
             preferencesWindow.WindowStartupLocation = WindowStartupLocation.CenterOwner;
-            preferencesWindow.ShowDialog();
+            preferencesWindow.Show();
+        }
+
+        public void EnableEnvironment(bool isEnabled)
```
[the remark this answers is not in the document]
jesusalvino: Done.

PR #13975, opened 2023-05-10
add dispose method for generated enum wrapper class DYN-5506

mjkkirschner:
### Purpose

https://jira.autodesk.com/browse/DYN-5506

@aparajit-pratap noticed an issue when creating and undoing creation of an enum backed ds object in dynamo in a codeblock.
I noticed that the generated wrapper class for each enum did not include a default dispose function which all generated type backed classes have - adding this removes the pointer to the enum wrapper from the `dsobjectmap` and `clrobjectmap` - which avoids exceptions from being thrown when a clr object is first bound to a dswrapper.

### TODO
- [x] tests? 

### Declarations

### Release Notes

Fixes a bug where enum backed dsobjects are not disposed fully.


test/Engine/ProtoTest/FFITests/CSFFIDispose.cs
```diff
@@ -155,15 +155,11 @@ public void Dispose67_PointInScope()
...
        public void DisposeEnumWrapper()
        {
            string code =
                @"
import(""FFITarget.dll"");
x = Days.Monday;
+__GC();
x = Days.Monday;
+__GC();
+x = Days.Monday;
```
aparajit-pratap: This diff is quite hard to read, I wonder if it's got to do with line endings or something?

mjkkirschner: wow - I didn't realize it was that bad, let me take a look to see if I can improve it.

mjkkirschner:
I've made the diff worse ;) , but I think it will be better going forward as the line endings are now consistent for the entire file.

PR #13957, opened 2023-04-28
Python editor docking and undocking

reddyashish:
### Purpose

Python editor docking and undocking. https://jira.autodesk.com/browse/DYN-3393

DockWindowInSideBar() is introduced to dock a window to the right side bar.
UndockWindow() is to undock a tab from the right side bar and call the necessary action to launch the original window.

python docking new

### Declarations

Check these if you believe they are true

- [ ] The codebase is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
- [ ] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions) and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.
- [ ] This PR modifies some build requirements and the readme is updated
- [ ] This PR contains no files larger than 50 MB 

### Release Notes
Python editor docking and undocking

### Reviewers
@QilongTang @zeusongit

PR #13984, opened 2023-05-12
Python editor docking improvements

reddyashish:
### Purpose

A followup PR to https://github.com/DynamoDS/Dynamo/pull/13957

Addressed the points mentioned in this task: https://jira.autodesk.com/browse/DYN-5862

1) Node rename should rename the docked python editor tab name.
2) Change the name SideBar Panel to Sidebar 
3) Engine name is not being set when the editor is opened in docked state.
4)  Changing the engine name and clicking on save & run button should update the engine name.

For the custom node workspace case, the editor doesn't close when the CN workspace is closed. It will close only when the CN is disposed. We will need to discuss on how to handle when the editor has unsaved changes.

python docking improvements

### Declarations

### Release Notes
Python editor docking improvements

### Reviewers
@QilongTang @zeusongit


src/DynamoCoreWpf/Properties/Resources.en-US.resx
```diff
@@ -2143,7 +2143,7 @@ Do you want to install the latest Dynamo update?</value>
     <value>Write note here</value>
   </data>
   <data name="SideBarPanelViewTitle" xml:space="preserve">
-    <value>SideBar Panel</value>
+    <value>SideBar</value>
```
avidit, line 2146: can we make is Sidebar instead of sideBar

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -81,6 +82,22 @@ public partial class DynamoViewModel : ViewModelBase, IDynamoViewModel
         /// </summary>
         internal Dictionary<string, ViewExtensionDisplayMode> NodeWindowsState { get; set; } = new Dictionary<string, ViewExtensionDisplayMode>();
 
+        /// <summary>
+        /// Collection of Right SideBar tab items: view extensions and docked windows.
+        /// </summary>
+        public ObservableCollection<TabItem> SideBarTabItems
```
QilongTang, line 88: Is `TabItem` type of WPF view elements? If yes, do you think we still want to store them under viewModel?

reddyashish: It is a window control element. I moved it to DynamoViewModel so that I can bind it to the view and don't have to set the tabcontrol content everytime the collection is changed. I was seeing combox(for engine value) issues when we were not using binding and have to update it everytime.

reddyashish:
Same flaky tests which are passing locally.

src/DynamoCoreWpf/ViewModels/Core/DynamoViewModel.cs
```diff
@@ -1470,6 +1487,23 @@ internal void AddToRecentFiles(string path)
             }
         }
 
+        // Get the nodemodel if a node is present in any open workspace.
+        internal NodeModel GetDockedWindowNodeModel(string tabId)
```
QilongTang, line 1491: Nice 👍🏻

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
```diff
@@ -1329,8 +1332,37 @@ private void Hide(object parameters)
             }
             else
             {
+                // When closing the custom node workspace, close any docked python script windows if there are no unsaved changes.
+                // Show warning message on the script editor if it has unsaved changes.
+                foreach (var node in Nodes)
```
QilongTang, line 1337: We may want to wrap this section into a utility function in case we want to reuse this part in the future in other workflows.

QilongTang (commented):
Overall solid improvements, one thing I forgot to ask before. Can we add analytics tracking for Python node docking/undocking for usage tracking?

src/DynamoCoreWpf/ViewModels/Core/WorkspaceViewModel.cs
reddyashish: Yeah moved it to a method in DynamoViewModel.

reddyashish:
@QilongTang I have now added the analytics tracking for docking/undocking only for python node windows explicitly.

reddyashish:
@QilongTang good to merge this?

PR #14007, opened 2023-05-18
Adding a GitHub action to perform a Binary Diff job on each PR/push event

zeusongit:
### Purpose

Add a binary diff workflow that will run on each push and PR, this will enable devs to check for updated files (Added, Deleted or Modified) in their PR vs files on the master branch.

Note: caches will be removed automatically after 7 days.

Screenshot 2023-05-19 at 6 44 49 PM

### Declarations

### Release Notes

Add Bin Diff support on PRs

### Reviewers

@DynamoDS/dynamo

### DIFF
--- a/.github/scripts/bin_diff.ps1
+++ b/.github/scripts/bin_diff.ps1
@@ -1,0 +1,225 @@
+#########################################################################
+### USAGE: rdiff path/to/left,path/to/right [-s path/to/summary/dir]  ###
+### ADD LOCATION OF THIS SCRIPT TO PATH                               ###
+#########################################################################
+[CmdletBinding()]
+param (
+  [parameter(HelpMessage="Stores the execution working directory.")]
+  [string]$ExecutionDirectory=$PWD,
+
+  [parameter(Position=0,HelpMessage="Compare two directories recursively for differences.")]
+  [alias("c")]
+  [string[]]$Compare,
+
+  [parameter(HelpMessage="Export a summary to path.")]
+  [alias("s")]
+  [string]$ExportSummary
+)
+
+### FUNCTION DEFINITIONS ###
+
+# SETS WORKING DIRECTORY FOR .NET #
+function SetWorkDir($PathName, $TestPath) {
+  $AbsPath = NormalizePath $PathName $TestPath
+  Set-Location $AbsPath
+  [System.IO.Directory]::SetCurrentDirectory($AbsPath)
+}
+
+# RESTORES THE EXECUTION WORKING DIRECTORY AND EXITS #
+function SafeExit() {
+  SetWorkDir /path/to/execution/directory $ExecutionDirectory
+  Exit
+}
+
+function Print {
+  [CmdletBinding()]
+  param (
+    [parameter(Mandatory=$TRUE,Position=0,HelpMessage="Message to print.")]
+    [string]$Message,
+
+    [parameter(HelpMessage="Specifies a success.")]
+    [alias("s")]
+    [switch]$SuccessFlag,
+
+    [parameter(HelpMessage="Specifies a warning.")]
+    [alias("w")]
+    [switch]$WarningFlag,
+
+    [parameter(HelpMessage="Specifies an error.")]
+    [alias("e")]
+    [switch]$ErrorFlag,
+
+    [parameter(HelpMessage="Specifies a fatal error.")]
+    [alias("f")]
+    [switch]$FatalFlag,
+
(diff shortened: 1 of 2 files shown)


.github/scripts/bin_diff.ps1
```diff
@@ -0,0 +1,258 @@
+################################################################################
```
mjkkirschner, line 1: is this script based on some sample code? Did it have a license, can we link to it?

.github/scripts/bin_diff.ps1
```diff
@@ -0,0 +1,258 @@
...
+# GETS ALL FILES WITH CALCULATED HASH PROPERTY RELATIVE TO A ROOT DIRECTORY RECURSIVELY #
+# RETURNS LIST OF @{RelativePath, Hash, FullName}
+function GetFilesWithHash {
+  [CmdletBinding()]
+  param (
+    [parameter(Mandatory=$TRUE,Position=0,HelpMessage="Path to get directories for.")]
+    [string]$Path,
+
+    [parameter(HelpMessage="The hash algorithm to use.")]
+    [string]$Algorithm="MD5"
```
mjkkirschner, line 154: https://stackoverflow.com/questions/201705/how-many-random-elements-before-md5-produces-collisions
SHA256 perhaps?

.github/scripts/bin_diff.ps1
```diff
@@ -0,0 +1,258 @@
...
+
+
+### END FUNCTION DEFINITIONS ###
+
+### PROGRAM LOGIC ###
+
+if($Compare.length -ne 2) {
+  Print -x "Compare requires passing exactly 2 path parameters separated by comma, you passed $($Compare.length)." -f
+}
+Print "Comparing $($Compare[0]) to $($Compare[1])..." -a 1
```
mjkkirschner, line 226: so I have a bit of a hard time understanding how this code works - but from your screenshot I have a few questions/comments.

1. I think noting that a file is `modified` if it still exists is probably just noise - IMO we don't really care, I expect files to be different when I do a build - they will contain different version numbers etc.
2. What I really care about are new files and deleted files - if we need to keep `modified` files - can we make deleted and added files pop out some how? (color text?)
3. Is a deleted file also considered modified or does it have a separate state? Can you show a screenshot of this?

.github/workflows/dynamoBinDiff.yml
```diff
@@ -0,0 +1,86 @@
+# Build Dynamo using latest VS and DotNET and perform a Bin Diff
+name: Dynamo-BinDiff
+on: [push,pull_request]
+jobs:
+ build-current:
+  runs-on: windows-2022
+  steps:
+    - name: Checkout Dynamo Repo current branch
```
mjkkirschner, line 8: I have a complication for you - we also need to do this same comparison for the DynamoAll.Net6.Sln targeting windows.

Would you prefer to just duplicate this action for net6 or have this job do 4 builds?

.github/scripts/bin_diff.ps1
zeusongit: 1. Agreed, do you suggest removing modified entries? I think we can let them be.
2. I have added colors to make them pop, green for Added, and red for Deleted entries.
3. Yes, deleted files has a separate "Deleted" label. attaching ss below.
Screenshot 2023-05-22 at 7 22 26 PM

.github/scripts/bin_diff.ps1
zeusongit: Done

.github/scripts/bin_diff.ps1
zeusongit: Added source.

.github/workflows/dynamoBinDiff.yml
zeusongit: Added 2 more builds for Net6-Windows (parallel builds for quick results)

Screenshot 2023-05-22 at 8 10 43 PM

mjkkirschner (approved):

.github/scripts/bin_diff.ps1
zeusongit: removed hash and side indicator columns as they were not useful.

Screenshot 2023-05-22 at 8 32 56 PM

QilongTang (approved):

zeusongit:
@mjkkirschner I don't think the coloring worked in Github Action console logs, but it did work in powershell.

